PR #1607, opened 2022-10-29
Send registration info to Fideslog

seanpreston:
Closes <issue>

### Code Changes

* [x] Call `send_registration` when a `UserRegistration` object is created or updated

### Steps to Confirm

* [ ] Run `nox -s dev`
* [ ] Make a `PUT` request to `/api/v1/registration` with the following data:
```
{
  "analytics_id": "this-is-a-test",
  "opt_in": "true"
}
```
  - assert a Fideslog registration record exists for this `analytics_id`.

### Pre-Merge Checklist

* Documentation Updated:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

- The `Registration` class has only been stubbed for now, as it'll need to be imported from Fideslog once [this PR](https://github.com/ethyca/fideslog/pull/87/files) has been merged and the SDK has been bumped. Subsequently the linting steps will fail.
- Another thing to note is that the internal Fides API supports dynamically opting in and out, but until the Fideslog client supports a delete registration function, it makes sense to leave this to only log the first registration request if the user opts in.


src/fides/api/ops/analytics.py
```diff
@@ -41,14 +40,23 @@ async def send_analytics_event(event: AnalyticsEvent) -> None:
...
     try:
-        await analytics_client._AnalyticsClient__send(  # pylint: disable=protected-access
-            event
+        await analytics_client.send_async(event)
+    except AnalyticsError as err:
+        log.warning("Error sending analytics event: {err}")
+    else:
+        log.info(
+            "Analytics event sent: {event.event} with client id: {analytics_client.client_id}"
         )
```
NevilleS, line 49: This feels like `DEBUG` level to me (if even). That'll be a lot of logs otherwise... and `event` might contain PII so probably best to exclude entirely.

src/fides/api/ops/analytics.py
```diff
@@ -41,14 +40,23 @@ async def send_analytics_event(event: AnalyticsEvent) -> None:
...
+
+
+async def send_registration(registration: UserRegistration) -> None:
+    if CONFIG.user.analytics_opt_out:
+        return
```
NevilleS, line 54: OK, so if analytics opt out is set, we won't send either analytics events or registrations. I think that's probably OK until we have a need to do those two things separately (e.g. register servers, but disable analytics). Right now we're expecting a single opt-out though.

NevilleS (approved):
⚠️ One change needed to remove the verbosity of some logs, but otherwise LGTM

src/fides/api/ops/analytics.py
```diff
@@ -41,14 +40,23 @@ async def send_analytics_event(event: AnalyticsEvent) -> None:
...
+    try:
+        await analytics_client.register_async(registration.as_fideslog())
     except AnalyticsError as err:
-        logger.warning("Error sending analytics event: %s", err)
+        log.warning(f"Error sending registration event: {err}")
```
ThomasLaPiana, line 58: there might be PII in the logs?

PR #1612, opened 2022-10-30
View the config as TOML instead of JSON

### LINKED ISSUE #1610: `fides view config` should dump out valid `toml` instead of `json`

Viewing the config shows users the correct key/value, but not in a usable format.

This should instead dump out `toml`, so the user can copy/pasta the config if they want

Give the user the choice, but since we don't accept JSON for configuration this seems unnecessary

ThomasLaPiana:
Closes #1610 

### Code Changes

* [x] print the config as TOML instead of JSON
* [x] create a bug issue for the fideslib `sqlalchemy_uri` misconfiguration (https://github.com/ethyca/fideslib/issues/77)

### Steps to Confirm

* [ ] `fides view config` shows _all_ valid values as TOML

### Pre-Merge Checklist

* Documentation Updated:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

Quick fix to print out the config as a TOML file

FYI: when I ran static fixes on my files locally it produced a diff, so unfortunately those are included as well. `view.py` is really the file to check here


src/fides/api/ctl/migrations/versions/8f1a19465239_adds_userregistration_table.py
```diff
@@ -5,36 +5,48 @@
...
-    op.create_index(op.f('ix_userregistration_id'), 'userregistration', ['id'], unique=False)
     # ### end Alembic commands ###
 
 
 def downgrade():
     # ### commands auto generated by Alembic - please adjust! ###
-    op.drop_index(op.f('ix_userregistration_id'), table_name='userregistration')
-    op.drop_table('userregistration')
+    op.drop_index(op.f("ix_userregistration_id"), table_name="userregistration")
+    op.drop_table("userregistration")
```
seanpreston, line 51: This looks like `black` formatting changes only?

src/fides/api/main.py
```diff
@@ -22,14 +22,7 @@
...
 from fides.api.ctl.database.database import configure_db
-from fides.api.ctl.routes import (
-    admin,
-    crud,
-    datamap,
-    generate,
-    health,
-    validate,
-)
+from fides.api.ctl.routes import admin, crud, datamap, generate, health, validate
```
seanpreston, line 25: Is there something we can set to preserve these import lists? Not sure what the rest of the team thinks, I find it very useful to see diffs per-line in cases like this with imports and method args

src/fides/cli/commands/view.py
```diff
@@ -1,8 +1,9 @@
 """Contains the view group of the commands for fides."""
 
 import click
+from toml import dumps as toml_dumps
 
-from fides.cli.utils import pretty_echo, with_analytics
+from fides.cli.utils import with_analytics, print_divider
```
seanpreston, line 6: Nit: Should these be alphabetical?

seanpreston (approved):

src/fides/api/ctl/migrations/versions/8f1a19465239_adds_userregistration_table.py
ThomasLaPiana: correct, black did some things (i run it as pre-commit)

src/fides/api/main.py
ThomasLaPiana: this is auto-formatting, maybe there's a setting we could dig into but that should be its own issue

src/fides/cli/commands/view.py
ThomasLaPiana: no idea how that slipped through and didn't trigger a CI failure....fixing

Issue #1479, opened 2022-10-18
Config Wizard - Add systems without privacy declarations


mfbrown:
As a user, when I complete a scan, I want to have access to the results of the scan so that I can add privacy information on my own schedule without losing the results of the scan.

**Details**
Today, when you perform an AWS scan, you need to add privacy declarations for each system and if you cancel, the systems from the scan aren't persisted, so you need to re-scan to add in the difference. 

**AC**
- Given that I have completed a scan in the config wizard, when I choose to add selected systems to my map, then they are created with empty privacy declarations.

PR #1634, opened 2022-10-31
1479/systems without declarations default

LKCSmith:
Closes [this issue](https://github.com/ethyca/fides/issues/1479).

https://user-images.githubusercontent.com/99051851/199157988-8f9e9f19-76bd-41f6-9769-009f436389c6.mov

### Code Changes

* [ ] Allow the creation of systems after the scan without going through the privacy declaration step

### Steps to Confirm

* [ ] Go through the config wizard as a plus and non-plus user 
* [ ] In the manual setup, you should still be prompted for privacy declaration setup
* [ ] In system scan, you will skip this setup and create systems immediately to add privacy declarations later
* [ ] Route to /systems in non-plus
* [ ] Route to /datamap in plus

### Pre-Merge Checklist

* Documentation Updated:
* [ ] Update `CHANGELOG.md`

### Description Of Changes


clients/admin-ui/cypress/e2e/config-wizard.cy.ts
```diff
@@ -72,25 +72,6 @@ describe.skip("Config Wizard", () => {
       cy.intercept("PUT", "/api/v1/system*", {
         fixture: "generate/system_to_review.json",
       }).as("putSystem");
-
-      // The form interactions are covered by the system forms tests.
-      cy.contains("Continue").click();
```
LKCSmith, line 77: These could probably be preserved in the future for declaration testing, but it doesn't belong in the scan test anymore.

clients/admin-ui/src/features/config-wizard/ScanResultsForm.tsx
```diff
@@ -7,17 +7,20 @@ import {
...
   ColumnDropdown,
   ColumnMetadata,
 } from "~/features/common/ColumnDropdown";
 import { SystemsCheckboxTable } from "~/features/common/SystemsCheckboxTable";
-import WarningModal from "~/features/common/WarningModal";
 import { System } from "~/types/api";
 
+import { useFeatures } from "../common/features.slice";
+import WarningModal from "../common/WarningModal";
+import { useCreateSystemMutation } from "../system";
```
allisonking, line 23: nit: prefer `~/features/` over relative paths that go outside of the folder (`./component` is fine, but `../component` starts to get confusing, especially when you move files around)

clients/admin-ui/src/features/config-wizard/ScanResultsForm.tsx
```diff
@@ -30,29 +33,48 @@ const ALL_COLUMNS: ColumnMetadata[] = [
...
   } = useDisclosure();
+  const [createSystem] = useCreateSystemMutation();
   const [selectedSystems, setSelectedSystems] = useState<System[]>(systems);
+  const features = useFeatures();
   const [selectedColumns, setSelectedColumns] =
     useState<ColumnMetadata[]>(ALL_COLUMNS);
 
+  const createSystems = async () => {
+    await selectedSystems.forEach((system: System) => createSystem(system));
+  };
```
allisonking, line 57: I don't think we should do a `forEach` here. We should use the system `/upsert` endpoint instead, which allows creating multiple systems at once. You can see [how it's used in plus](https://github.com/ethyca/fidesctl-plus/blob/main/clients/datamap-ui/src/features/scan/scan.slice.ts#L38-L49) for reference

clients/admin-ui/src/features/config-wizard/ScanResultsForm.tsx
```diff
@@ -30,29 +33,48 @@ const ALL_COLUMNS: ColumnMetadata[] = [
...
+
   const confirmRegisterSelectedSystems = () => {
-    dispatch(chooseSystemsForReview(selectedSystems.map((s) => s.fides_key)));
-    dispatch(changeStep());
+    if (manualSystemSetupChosen) {
+      dispatch(chooseSystemsForReview(selectedSystems.map((s) => s.fides_key)));
+      return dispatch(changeStep());
+    }
```
allisonking, line 63: why do we care about if `manulSystemSetupChosen`? doesn't the manual step skip the scan results step (since there are no scan results?)

clients/admin-ui/src/features/config-wizard/ScanResultsForm.tsx
```diff
@@ -30,29 +33,48 @@ const ALL_COLUMNS: ColumnMetadata[] = [
...
-      onWarningOpen();
-    } else {
-      confirmRegisterSelectedSystems();
+      return onWarningOpen();
+    }
+    if (manualSystemSetupChosen) {
+      return confirmRegisterSelectedSystems();
     }
+    createSystems();
+    return features.plus ? router.push(`/datamap`) : router.push(`/system`);
```
allisonking, line 77: is it possible to combine this logic with some of `confirmRegisterSelectedSystems` since they look very similar?

clients/admin-ui/cypress/e2e/config-wizard.cy.ts
allisonking: that's fine, but we should extend this test then, since there is new functionality now which calls the create. we should assert that we do indeed POST the systems that are selected

allisonking (changes requested):

clients/admin-ui/cypress/e2e/config-wizard.cy.ts
LKCSmith: Makes sense! I'll add :)

clients/admin-ui/src/features/config-wizard/ScanResultsForm.tsx
LKCSmith: Thank you! Keeping Upserts in my back pocket again 💃

clients/admin-ui/cypress/e2e/config-wizard.cy.ts
```diff
@@ -1,7 +1,7 @@
 import { stubSystemCrud, stubTaxonomyEntities } from "cypress/support/stubs";
 
 // skipping while configWizardFlag exists
-describe.skip("Config Wizard", () => {
+describe("Config Wizard", () => {
```
allisonking, line 4: you'll probably have to put the `skip` back in to make CI happy :(

clients/admin-ui/src/features/config-wizard/ConfigWizardWalkthrough.tsx
```diff
@@ -69,14 +73,19 @@ const ConfigWizardWalkthrough = () => {
...
+              ) : null}
               {step === 3 ? <AuthenticateScanner /> : null}
               {step === 4 ? (
                 <Box pr={10}>
                   <ScanResultsForm />
                 </Box>
               ) : null}
-              {step === 5 ? (
+              {/* These steps should only apply if you're creating systems manually */}
+              {step === 5 && manualSystemSetupChosen ? (
```
allisonking, line 88: I'm a little confused why `manualSystemSetupChosen` is needed? I think step 4 goes straight to other routes, so is there a path that ends up at step 5 without going the manual route? lemme know if I'm missing something!

clients/admin-ui/src/features/config-wizard/ScanResultsForm.tsx
```diff
@@ -33,26 +36,33 @@ const ALL_COLUMNS: ColumnMetadata[] = [
...
     onClose: onWarningClose,
   } = useDisclosure();
+  const [upsertSystems] = useUpsertSystemsMutation();
   const [selectedSystems, setSelectedSystems] = useState<System[]>(systems);
+  const features = useFeatures();
   const [selectedColumns, setSelectedColumns] =
     useState<ColumnMetadata[]>(ALL_COLUMNS);
 
+  const createSystems = async () => {
+    await upsertSystems(selectedSystems);
```
allisonking, line 52: we should get the result and handle errors, something like this pattern: https://github.com/ethyca/fides/blob/main/clients/admin-ui/src/features/config-wizard/AuthenticateAwsForm.tsx/#L103-L116

and then we'd probably want a toast or something, not sure if the designs have something like that

clients/admin-ui/src/features/config-wizard/ScanResultsForm.tsx
```diff
@@ -33,26 +36,33 @@ const ALL_COLUMNS: ColumnMetadata[] = [
...
+  };
+
   const confirmRegisterSelectedSystems = () => {
     dispatch(chooseSystemsForReview(selectedSystems.map((s) => s.fides_key)));
-    dispatch(changeStep());
+    createSystems();
+    return features.plus ? router.push(`/datamap`) : router.push(`/system`);
```
allisonking, line 58: I don't think you need a `return` statement here?

clients/admin-ui/src/features/config-wizard/ScanResultsForm.tsx
```diff
@@ -33,26 +36,33 @@ const ALL_COLUMNS: ColumnMetadata[] = [
...
   };
 
   const handleSubmit = () => {
     if (systems.length > selectedSystems.length) {
-      onWarningOpen();
-    } else {
-      confirmRegisterSelectedSystems();
+      return onWarningOpen();
     }
+    return confirmRegisterSelectedSystems();
```
allisonking, line 65: same thing here, I don't think you need the `return`. not that it hurts, but it makes me think something other than void is getting returned

clients/admin-ui/cypress/e2e/config-wizard.cy.ts
```diff
@@ -73,24 +73,17 @@ describe.skip("Config Wizard", () => {
...
-      cy.get(`[data-testid="system-needs-review"]`).should("have.length", 2);
+      // assert that the systems do POST
+      cy.intercept({
+        method: "POST",
+        url: "/api/v1/system/upsert",
+      }).as("upsertSystems");
+
+      cy.getByTestId("warning-modal-confirm-btn").click();
+
+      cy.wait("@upsertSystems").then((interception) => {
```
allisonking, line 84: thanks for adding this!!

allisonking (commented):
looking really good! thanks for making those changes. just a few comments

clients/admin-ui/src/features/config-wizard/ScanResultsForm.tsx
LKCSmith: I think the toasts (or at least, response messages feedback to the user) are going to be handled by [this issue](https://github.com/orgs/ethyca/projects/17/views/1?filterQuery=is%3Aissue+sprint%3A%40current+assignee%3ALKCSmith) when I get to it next, but let me know if you think otherwise! I'll add the other things though.

clients/admin-ui/src/features/config-wizard/ScanResultsForm.tsx
LKCSmith: I think I was getting a linting error since I'm using an arrow function here that it needs to return something at the end of the fn 🤔

LKCSmith: Your other comment made me curious, so I'll take another look

clients/admin-ui/cypress/e2e/config-wizard.cy.ts
LKCSmith: Oh, ha, yup

clients/admin-ui/src/features/config-wizard/ConfigWizardWalkthrough.tsx
LKCSmith: You still need it because otherwise, after clicking "register selected systems" on the ScanResultsForm, it'll go to the Describe/Declare/Review page, which we are now skipping for the scanned systems flow.

clients/admin-ui/src/features/config-wizard/ScanResultsForm.tsx
LKCSmith: @allisonking Actually, are you okay with me handling the success/responses from this upsert in that PR related to that issue, since the config wizard is hidden for now? Toasts aren't in the designs, and I think that issue covers your concerns here when I get to it later today.

LKCSmith:
@allisonking - Thank you! Left a comment above for one of your feedback points. Let me know!

clients/admin-ui/src/features/config-wizard/ScanResultsForm.tsx
```diff
@@ -33,26 +36,38 @@ const ALL_COLUMNS: ColumnMetadata[] = [
...
+    router.push(`/system`);
   };
 
   const handleSubmit = () => {
     if (systems.length > selectedSystems.length) {
       onWarningOpen();
-    } else {
-      confirmRegisterSelectedSystems();
     }
+    confirmRegisterSelectedSystems();
```
allisonking, line 70: hmm are you sure we don't need the `else` here? we wouldn't want to open the warning AND confirm selected systems. not sure if that's what happens, or if opening the warning somehow breaks us out of the function?

clients/admin-ui/src/features/config-wizard/ScanResultsForm.tsx
```diff
@@ -33,26 +36,38 @@ const ALL_COLUMNS: ColumnMetadata[] = [
...
+    await upsertSystems(selectedSystems);
+  };
+
   const confirmRegisterSelectedSystems = () => {
     dispatch(chooseSystemsForReview(selectedSystems.map((s) => s.fides_key)));
-    dispatch(changeStep());
+    createSystems();
+
+    if (features.plus) {
+      router.push(`/datamap`);
```
allisonking, line 60: it just occurred to me that this won't work, since the datamap UI will be in a different nextjs zone. I think you'll have to do something like what Sebastian implemented for the navbar. and it'll be a pain to test 😬 

https://github.com/ethyca/fides/blob/1479/systems-without-declarations-default/clients/admin-ui/src/features/common/nav/NavLink.tsx/#L29-L32

clients/admin-ui/src/features/config-wizard/ScanResultsForm.tsx
allisonking: I just tested it out, and confirmed removing the `else` does make clicking "register" just register the systems without first waiting for the warning to be accepted

allisonking (changes requested):

clients/admin-ui/src/features/config-wizard/SuccessPage.tsx
```diff
@@ -49,7 +50,21 @@ const SuccessPage = ({
...
+
+    const datamapRoute = resolveLink({
+      href: "/datamap",
+      basePath,
+    });
+
+    const systemRoute = resolveLink({
+      href: "/system",
+      basePath,
+    });
```
allisonking, line 63: I don't think `systemRoute` needs to be resolved since it'll be within the same zone as the config wizard

clients/admin-ui/src/features/config-wizard/SuccessPage.tsx
```diff
@@ -49,7 +50,21 @@ const SuccessPage = ({
 
   const onFinish = () => {
     dispatch(setActiveSystem(undefined));
-    return features.plus ? router.push("/datamap") : router.push("/system");
+    const basePath = router.basePath || "/";
```
allisonking, line 53: Because this is the admin-ui, the basePath will always be `/`. that means you can simplify this bit of code to:

```typescript
const datamapRoute = resolveLink({
  href: "/datamap",
  basePath: "/"
});

features.plus ? router.push(datamapRoute.href) : router.push('/system')
```

I could also see an argument to do it the way you have now, since then these components would work regardless of if they are in the admin UI or in the datamap UI, but in that case, we should refactor this into a function so that all of the places that need it can use it without copy and pasting code

allisonking (changes requested):
almost there!!

clients/admin-ui/src/features/system/SystemRegisterSuccess.tsx
```diff
@@ -42,10 +42,24 @@ const SystemRegisterSuccess = ({ system, onAddNextSystem }: Props) => {
     : [];
 
   const systemName = system.name ?? system.fides_key;
+  const basePath = router.basePath || "/";
 
   const onFinish = () => {
     dispatch(setActiveSystem(undefined));
-    return features.plus ? router.push("/datamap") : router.push("/system");
```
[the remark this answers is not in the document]
LKCSmith: I can do that, since the ticket really is just related to the wizard specifically

clients/admin-ui/src/features/config-wizard/SuccessPage.tsx
LKCSmith: Ah, right right

allisonking (approved):
💥 go ahead and merge, and I'll sort out the conflicts on my branch 🙌

PR #1569, opened 2022-10-26
[1480] ui/auth: Query current user permissions to detect stale token

ssangervasi:
Closes #1480
Depends on API change: https://github.com/ethyca/fides/pull/1553

### Code Changes

* Query current user permissions to detect stale token
  * The important change!
  * ProtectedRoute will now kick the user to the login page if their client token is invalidated.
* Add & use API models for the use-management feature
  * I let some of the aliases stick around because the names are actually nicer than the API one.
* Remove some lingering code from unified merge
  * The auth slice middleware was [deleted](https://github.com/ethyca/fidesops/pull/1409/files#diff-af758411de81fed87d14012809b227b11c67ce770c998fd5059ea0769fbb5bd4L59) from ops, but made its way back by accident
  * The old ctl user state was unused.

### Steps to Confirm

1. Set up the test env
1. Login as fidestest
1. Navigate to a protected route (DSR)
1. Teardown and recreate the test env
    - Or just delete the user client record from the DB
1. Reload the page 
    - [ ] You should be redirected to the login page
    - [ ] Going through the flow with a non-root user should work as well.

### Pre-Merge Checklist

* [ ] All CI Pipelines Succeeded
* Documentation Updated:
  * [ ] documentation complete, or draft/outline provided (tag docs-team to complete/review on this branch)
  * [ ] documentation issue created (tag docs-team to complete issue separately)
* [ ] Issue Requirements are Met
* [ ] Relevant Follow-Up Issues Created
* [x] Update `CHANGELOG.md`

### Description Of Changes

The user permissions endpoint will allow users to query their own permissions. An invalid (expired) token will result in an error, which means the user is no longer authenticated. In addition to redirecting to the login page, the auth state is cleared from local storage.

We can also build on this in the future to have protected routes check for particular permissions instead of just success/error.

PR #1635, opened 2022-10-31
Restore: [1480] ui/auth: Query current user permissions to detect stale token

Kelsey-Ethyca:
restore https://github.com/ethyca/fides/pull/1569

There is a bug now where root_user cannot access the app and a test skipped that needs to be restored.


chriscalhoun1974 (approved):
lgtm

Issue #1648, opened 2022-11-01
Datamap endpoint `system_dependencies` are being overwritten by N/A when there are no privacy declarations


allisonking:
### Bug Description

When there are no privacy declarations, the `system_dependencies` field of the datamap obj returns `N/A` even if there is actually stuff that should go there.

#### Steps to Reproduce

1. Replace the `privacy_declarations` in https://github.com/ethyca/fides/blob/main/demo_resources/demo_system.yml with `[]`
1. `fides push demo_resources`
2. Query the datamap endpoint
3. `system_dependencies` will show up as `N/A` despite one of them actually having a dependency

### Expected behavior

System dependencies to show up properly

### Screenshots

_If applicable, add screenshots to help explain your problem._

### Environment

* Version:
* OS:
* Python Version:
* Docker Version:

### Additional context

Likely a bug introduced by https://github.com/ethyca/fides/pull/1603

PR #1649, opened 2022-11-01
Fix start index of privacy declarations

allisonking:
Closes https://github.com/ethyca/fides/issues/1648

### Code Changes

* [x] Fixes the start index of privacy declarations

### Steps to Confirm

* [ ] Follow the repro steps in the bug ticket and confirm it doesn't happen anymore

### Pre-Merge Checklist

* Documentation Updated:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

I _think_ this is the change needed, but I'm not sure if other things might be affected. for instance, I would expect since we started too early before (at index 6, `system_dependencies`), we would've ended too early (before `data_qualifier`). but it doesn't look like `data_qualifier` is part of the output so maybe I'm missing something.


src/fides/ctl/core/export.py
```diff
@@ -232,7 +232,7 @@ def generate_system_records(
                 data_protection_impact_assessment["link"],
             ]
             num_privacy_declaration_fields = 12
-            privacy_declaration_start_index = 6
+            privacy_declaration_start_index = 7
```
allisonking, line 235: perhaps your idea to explicitly write out all the N/A's would've been better 😅

SteveDMurphy: We still can!

SteveDMurphy (approved):
Confirmed the bug is now fixed! It was being pushed down into the same position as `system.privacy_declaration.data_qualifier`. The number of fields is the same but the insertion point was just off by 1 - thanks and my apologies @allisonking

Issue #1636, opened 2022-10-31
fix ProtectedRoute test


Kelsey-Ethyca:
when the user is logged in - should render passed children

We need to mock useGetUserPermissionsQuery

ssangervasi:
I'm probably going to fix this as part of https://github.com/ethyca/fidesctl-plus/issues/246 -- many new pages require login checks, so I have spread the mock love.

PR #1106, opened 2022-09-21
Set up cypress component testing

allisonking:
Cypress introduced [component testing](https://docs.cypress.io/guides/component-testing/writing-your-first-component-test) a few versions ago. While thinking about the classify features which don't have endpoints to test with yet, I thought it would be useful to explore component testing. They're easier to build out in isolation, and are much faster than e2e tests

Component testing is kind of like a mix of storybook and jest tests: like storybook, it renders your component in a browser, but like jest, you can actually test things (although you can use cypress commands, which we already know and love). It looks like storybook now also has [interaction testing](https://storybook.js.org/docs/react/writing-tests/interaction-testing) which might be the same thing except for within storybook. However, since we're already using cypress, it was easier to set this up.

### Code Changes

* [x] Configure cypress to be able to use component testing (including tsconfigs and making it use chakra for theming)
* [x] Add a stanza to GHA to run component tests
* [x] Add a small test for `DataCategoryInput`

### Steps to Confirm

* [ ] Run `npm run cy:open`. This will let you choose between E2E testing and component testing. Choose component testing, then watch it do its thing
* [ ] [CI passes](https://github.com/ethyca/fides/actions/runs/3100405570/jobs/5020650252#step:6:84)

### Pre-Merge Checklist

* [x] All CI Pipelines Succeeded
* Documentation Updated:
  - [ ] documentation complete, or draft/outline provided (tag docs-team to complete/review on this branch)
  - [ ] documentation issue created (tag docs-team to complete issue separately)
* [ ] Issue Requirements are Met
* [ ] Relevant Follow-Up Issues Created
* [x] Update `CHANGELOG.md`

### Description Of Changes

Traditionally this sort of test should live next to the component. We don't have a habit of putting our components in their own folders though, so it would make our folder structure kind of messy. We may want to consider putting our components in lil folders for purposes like this (maybe storybook files would also go here, css modules [not that we really have any], unit tests...). I put them in `cypress/components` (the default installation path) for now, but we can always move them if people feel strongly about it.

PR #1670, opened 2022-11-02
ui/layout: Wrap pages in ProtectedRoute by default

ssangervasi:
Contributes to https://github.com/ethyca/fidesctl-plus/issues/246
(The datamap portion lives in Plus and has to be solved on its own.)

### Code Changes

- Wrap pages in ProtectedRoute by default
    - Other than the changes to _app.tsx, most of the lines are whitespace changes due to the removal of <ProtectedRoute> tags that are no longer necessary.
- Cypress: Login before each test suite
    - May fix https://github.com/ethyca/fides/issues/1636
    - Uncovered a problem with how local storage was serialized that should definitely help.

### Steps to Confirm

1. Use the test_env
2. Log out
3. Try to access Ctl route like http://localhost:3000/system
    - [ ] Should be redirected to the login page
4. Log in as the test user
5. Try to access those routes 
    - [ ] Should be able to access

### Pre-Merge Checklist

* Documentation Updated:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

I chose to solve this by having our core app layout automatically wrap `NextPage` components in `ProtectedRoute`. An alternative is to directly wrap every page, like we've been doing, but this adds boilerplate to new pages with the risk of leaving them exposed. The only page that has a reason to be unprotected is `Login` which has a specific bypass.


clients/admin-ui/cypress/e2e/datasets.cy.ts
```diff
@@ -6,11 +6,8 @@ import {
 } from "cypress/support/stubs";
 
 describe("Dataset", () => {
-  before(() => {
```
ssangervasi, line 9: Note that `before` runs once for the describe block, instead of wrapping each test case. That's generally more fragile, and it was in this case because Cypress automatically clears localStorage between test cases.

clients/admin-ui/src/features/user-management/UserManagementLayout.tsx
```diff
@@ -18,29 +18,27 @@ interface Props {
   children: React.ReactNode;
 }
 const Profile = ({ title, children }: Props) => (
-  <ProtectedRoute>
```
ssangervasi, line 21: (whitespace changes)

clients/admin-ui/src/pages/_app.tsx
```diff
@@ -32,7 +35,17 @@ const MyApp = ({ Component, pageProps }: AppProps) => (
       <Provider store={store}>
         <PersistGate loading={null} persistor={persistor}>
           <ChakraProvider theme={theme}>
-            <Component {...pageProps} />
+            {Component === Login ? (
+              // Only the login page is accessible while logged out. If there is
```
chriscalhoun1974, line 39: Nice comment for future reference!

chriscalhoun1974 (approved):
lgtm

clients/admin-ui/__tests__/index.test.tsx
```diff
@@ -1,22 +0,0 @@
-import Home from "../src/pages";
-import { mockNextUseRouter, render, screen } from "./test-utils";
-
-// skipping while configWizardFlag exists
-describe("Home", () => {
```
ssangervasi, line 5: 🧵 Each of the tests I deleted were testing a mocked version of the authentication rules that are now covered by the cypress test. The cypress flow tests how the interaction between localStorage, the redux store, and the actual Next routing will function. In my experience, that's going to a give us a lot more confidence than maintaining these mocks.

(We may get rid of `@testing-library/react` based test altogether, seeing as we have Cypress [component testing](https://github.com/ethyca/fides/pull/1106).)

clients/admin-ui/src/app/store.ts
```diff
@@ -60,26 +60,14 @@ import { reducer as taxonomyReducer, taxonomyApi } from "~/features/taxonomy";
...
-  removeItem() {
-    return Promise.resolve();
-  },
-});
-
 const storage =
   typeof window !== "undefined"
     ? createWebStorage("local")
-    : createNoopStorage();
+    : createWebStorage("noop");
```
ssangervasi, line 70: Stumbled across this util.

chriscalhoun1974 (approved):
Excellent work!

PR #1678, opened 2022-11-03
ui/auth: Remove lingering local storage parsing

ssangervasi:
### Description Of Changes

I was wondering why we need to manually open up localStorage when we have redux-persist configured to do that for us now. Removed this and the log in/out still works as expected. So I think this was just a lingering from changeover.

Similar to this block that was left over from the fidesops merge: https://github.com/ethyca/fides/pull/1569/files#diff-3071c1012e5bece4441d5b7643c083188c22fbfe96caea78da1f09fdeeeb1599L59

### DIFF
--- a/clients/admin-ui/src/app/store.ts
+++ b/clients/admin-ui/src/app/store.ts
@@ -57,7 +57,7 @@ import {
 import { reducer as systemReducer, systemApi } from "~/features/system";
 import { reducer as taxonomyReducer, taxonomyApi } from "~/features/taxonomy";
 
-import { authApi, AuthState, reducer as authReducer } from "../features/auth";
+import { authApi, reducer as authReducer } from "../features/auth";
 
 /**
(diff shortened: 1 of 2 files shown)


clients/admin-ui/src/app/store.ts
```diff
@@ -111,10 +111,12 @@ const reducer = {
   userManagement: userManagementReducer,
 };
 
+export type RootState = StateFromReducersMapObject<typeof reducer>;
+
 const allReducers = combineReducers(reducer);
 
-const rootReducer = (state: any, action: AnyAction) => {
-  let newState = { ...state };
+const rootReducer = (state: RootState | undefined, action: AnyAction) => {
```
ssangervasi, line 118: Note: by moving the RootState definition up we can use the right type here.

chriscalhoun1974 (approved):

PR #1679, opened 2022-11-03
Make all dependencies part of the standard install except for known-dangerous packages

### LINKED ISSUE #1667: Direct dependency `bson` is not explicitly required
### Bug Description

The [bson package](https://pypi.org/project/bson/) is referenced in several places as a direct dependency, but it is not explicitly required.

#### Steps to Reproduce

1. `pip install ethyca-fides`
1. Start the API server
1. See exceptions like `ModuleNotFoundError: No module named 'bson'`

### Expected behavior

Direct dependencies should be explicitly required.

### Additional context

The bson package is a dependency of the "mongo" set of optional dependencies:

https://github.com/ethyca/fides/blob/f684b0ab86730e9efb30587fe28b246e1cd72a98/setup.py#L54

It may be enough to gate it's usage behind a check for `pymongo`.

The bson package is currently referenced in the following places:

1. https://github.com/ethyca/fides/blob/829b941c529088a6d2ca1ffe9292677efea49789/src/fides/api/ops/graph/data_type.py#L6-L7
1. https://github.com/ethyca/fides/blob/829b941c529088a6d2ca1ffe9292677efea49789/tests/ops/graph/test_data_types.py#L1
1. https://github.com/ethyca/fides/blob/829b941c529088a6d2ca1ffe9292677efea49789/tests/ops/integration_tests/test_mongo_task.py#L8
1. https://github.com/ethyca/fides/blob/829b941c529088a6d2ca1ffe9292677efea49789/tests/ops/task/test_filter_results.py#L4
1. https://github.com/ethyca/fides/blob/829b941c529088a6d2ca1ffe9292677efea49789/tests/ops/task/test_graph_task.py#L3

### LINKED ISSUE #1668: Dependency `redis` is optional, but treated as required
### Bug Description

The `redis` dependency is included in `optional-requirements.txt`:

https://github.com/ethyca/fides/blob/829b941c529088a6d2ca1ffe9292677efea49789/optional-requirements.txt#L18-L19

And only installed when using `pip install ethyca-fides[redis]` or `pip install ethyca-fides[all]`.

It is treated as a required dependency, by assuming it is installed, here:

https://github.com/ethyca/fides/blob/829b941c529088a6d2ca1ffe9292677efea49789/src/fides/api/main.py#L18

#### Steps to Reproduce

1. `pip install ethyca-fides`
1. Start the API server
1. Observe that exceptions like `ModuleNotFoundError: No module named 'redis'` are thrown

### Expected behavior

Exceptions should not be thrown.

### LINKED ISSUE #1676: Dependency `boto3` is optional, but treated as required
### Bug Description

The `boto3` dependency is included in `optional-requirements.txt`:

https://github.com/ethyca/fides/blob/f684b0ab86730e9efb30587fe28b246e1cd72a98/optional-requirements.txt#L3-L4

And only installed when using `pip install ethyca-fides[aws]` or `pip install ethyca-fides[all]`.

It is treated as a required dependency, by assuming it is installed, here:

https://github.com/ethyca/fides/blob/f684b0ab86730e9efb30587fe28b246e1cd72a98/src/fides/api/ops/models/datasetconfig.py#L4

#### Steps to Reproduce

1. `pip install ethyca-fides`
1. Start the API server
1. Observe that exceptions like `ModuleNotFoundError: No module named 'boto3'` are thrown

### Expected behavior

Exceptions should not be thrown.

### LINKED ISSUE #1677: Dependency `snowflake-sqlalchemy` is optional, but treated as required
### Bug Description

The `snowflake-sqlalchemy` dependency is included in `optional-requirements.txt`:

https://github.com/ethyca/fides/blob/f684b0ab86730e9efb30587fe28b246e1cd72a98/optional-requirements.txt#L25-L26

And only installed when using `pip install ethyca-fides[snowflake]` or `pip install ethyca-fides[all]`.

It is treated as a required dependency, by assuming it is installed, here:

https://github.com/ethyca/fides/blob/f684b0ab86730e9efb30587fe28b246e1cd72a98/src/fides/api/ops/service/connectors/sql_connector.py#L5

#### Steps to Reproduce

1. `pip install ethyca-fides`
1. Start the API server
1. Observe that exceptions like `ModuleNotFoundError: No module named 'snowflake'` are thrown

### Expected behavior

Exceptions should not be thrown.

ThomasLaPiana:
Closes #1668 
Closes #1667 
Closes #1676 
Closes #1677

### Code Changes

* [x] update the `fides` CI check to also import the webserver
* [x] move the `fides` install check to the static checks file, since no Docker build is required
* [x] more the `optional-requirements.txt` file, moving them all into `requirements.txt`
* [x] remove the `setup.py` logic for the various optional requirements, except for the dangerous `mssql` one
* [x] update all documentation that references this feature
* [x] update documentation to encourage users to use best practices, i.e. `pipx` or `venv` for installing fides
* [x] Remove code related to conditionally checking for packages

### Steps to Confirm

* [ ] installing the package locally installs everything except for `pyodbc` (mssql)

### Pre-Merge Checklist

* Documentation Updated:
  * [x] documentation complete, or draft/outline provided (tag docs-team to complete/review on this branch)
* [x] Update `CHANGELOG.md`

### Description Of Changes

While importing and using `fides` in other repos we've discovered that some of the `optional` dependencies weren't actually `optional` for running the webserver. This PR includes a test for this in CI checks as well as moves _all_ optional dependencies, except for the dangerous one (mssql) into the main requirements file.

#### But why get rid of optional dependencies altogether?

It is a feature no one has asked for (@ThomasLaPiana believed it was a good idea, as some other projects follow this pattern) but in the end without a single cohesive framework for actually handling the careful importation of potentially non-existent packages, the engineering burden is too high.

This should get revisited with a design for a proper `plugin` framework (like those other projects have!) when there becomes a need for it


ThomasLaPiana:
current failing checks are all expected

ThomasLaPiana:
I think these failures are just colliding with the runs on `main` (besides the external `ctl` ones), will rerun to confirm when both are done

ThomasLaPiana:
@sanders41 those 3 ctl failures are known-bad, I'm rerunning the one failed job here to confirm it passes

ThomasLaPiana:
yep, all expected are passing

sanders41 (approved):

.github/workflows/cli_checks.yml
```diff
@@ -28,7 +28,7 @@ jobs:
 
   Fides-Deploy:
     runs-on: ubuntu-latest
-    timeout-minutes: 15
+    timeout-minutes: 20
```
PSalant726, line 31: Why include a timeout at all? Is it possible that this (and other jobs with timeouts set) could hang indefinitely?

.github/workflows/static_code_checks.yml
```diff
@@ -82,6 +82,25 @@ jobs:
...
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ env.DEFAULT_PYTHON_VERSION }}
+
+      - name: Install Nox
+        run: pip install nox>=2022
```
PSalant726, line 99: It's definitely inefficient to keep doing this in multiple jobs, and I feel like there has to be a more DRY workaround. Do you think this repo is in a place where we could realistically address a follow-up issue to optimize our workflows? We may need to at least block an issue like that on consistently passing tests, but it's the type of thing I'd personally _love_ to address.

requirements.txt
```diff
@@ -27,11 +30,17 @@ psycopg2-binary==2.9.3
 pydantic<1.10.0
 pydash==5.1.1
 PyJWT==2.4.0
+pymongo==3.13.0
+PyMySQL==1.0.2
 python-jose[cryptography]==3.3.0
 pyyaml>=5,<6
+redis==3.5.3
+sqlalchemy[asyncio]==1.4.27
```
PSalant726, line 38: Is the version change here required?

PSalant726 (changes requested):
This is definitely the simplest solution to the bug issues that I opened, but there's a problem: We currently have a valid requirement for some dependencies remaining optional. See: https://github.com/ethyca/fidesops-plus/pull/99#pullrequestreview-1168933882.

I would rather have seen a PR that restores our guarantee that dependencies defined as optional actually _are_ optional, and then work to remove dependency optionality as individual feature enhancement PRs. A workflow like that would provide the opportunity to discuss the ramifications of each currently optional dependency becoming required. Per the above linked comment, the fides repo does not currently support extension from fidesops-plus, and unfortunately these changes wouldn't enable that.

.github/workflows/static_code_checks.yml
sanders41: I don't know a way right off for a DRY workaround, but you can cache the dependencies so they don't have to reinstall on every run. Only one dependency in this case so probably not a big difference, but still some improvement.

PSalant726: I have a few ideas, for this and our other workflows. I just don't want to start work on an overhaul until things are stable.

.github/workflows/cli_checks.yml
ThomasLaPiana: Yes, this is a stop-gap until there is time to go find and fix that issue properly.

.github/workflows/static_code_checks.yml
ThomasLaPiana: @PSalant726 design doc or issue? I'd be really curious to see as well, I agree there could be improvements. Actually in this case, it would easily be a matrix

requirements.txt
ThomasLaPiana: @PSalant726 I think I was getting some small version issues since it is now in the same file as `sqlalchemy` base, and the small bump fixed it

ThomasLaPiana:
> This is definitely the simplest solution to the bug issues that I opened, but there's a problem: We currently have a valid requirement for some dependencies remaining optional. See: [ethyca/fidesops-plus#99 (review)](https://github.com/ethyca/fidesops-plus/pull/99#pullrequestreview-1168933882).
> 
> I would rather have seen a PR that restores our guarantee that dependencies defined as optional actually _are_ optional, and then work to remove dependency optionality as individual feature enhancement PRs. A workflow like that would provide the opportunity to discuss the ramifications of each currently optional dependency becoming required. Per the above linked comment, the fides repo does not currently support extension from fidesops-plus, and unfortunately these changes wouldn't enable that.

@PSalant726 I'm not fully understanding how these issues are related. That issue is specifically referring to the inability to run the application without `redis` and the `database` spun up, which has its own ticket. That product also gets packaged as a standalone docker image _only_, so the extra dependencies aren't a huge concern to me.

What _is_ a concern to me is the complexity of testing and verifying each `optional` dependency against a specific feature-set. As stated above, the additional complexity is not worth the value we are getting, and I'd rather ship this change to increase stability and revisit `optional` dependencies later when we can engineer a proper solution from the ground up.

We built checks into `ctl` as we knew this feature was something we wanted to support, but `ops` has a lot of code paths that are non-trivial to rewrite. I feel that this is the best solution for where we are now

ThomasLaPiana:
this looks like it is hitting Shopify rate-limits during normal test runs, this should be addressed

ThomasLaPiana:
I created a bug issue for these failures: #1714 

merging as unrelated

PR #1688, opened 2022-11-03
Adds server-side rate limiter with sane default to Fides

seanpreston:
### Code Changes

* [x] Adds `slowapi` to the project's dependencies
* [x] Adds `CONFIG.security.request_rate_limit` and sets the sane default to 100 requests per rolling minute
* [x] Adds `CONFIG.security.rate_limit_prefix` to append a prefix to any key the rate limiter uses. This is helpful when using Redis as the backend, so that users can guarantee no key conflicts

### Steps to Confirm

* [ ] Set the rate limit to something arbitrarily low e.g. `1/minute`
* [ ] Run the server with `nox -s dev`
* [ ] Try to access `/health` more than once within a minute
* [ ] Assert that the second request returns a 429

### Pre-Merge Checklist

* Documentation Updated:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

- I've chosen to use `slowapi` as the rate limiting library as it's [consistent with `fideslog`'s implementation](https://github.com/ethyca/fideslog/blob/main/fideslog/api/requirements.txt#L4). Overall it's a nice implementation that allows us to override the ratelimit dynamically at route level with a decorator e.g.
```
@router.get("/test")
@limiter.limit("2/minute")
async def test(
    request: Request
):
return "hi"
```
- `slowapi` does come with some [known issues](https://slowapi.readthedocs.io/en/latest/#limitations-and-known-issues), ~for example requiring the `request` to be passed into an endpoint in order to effectively ratelimit this. I'll need to spend some time analysing the extent to which we omit this var~. UPDATE: this is only if we use the `@limiter(...)` decorator, and not when we default for every request.

### To Discuss

- My initial concern is that could this could get in the way of legitimate requests that Fides wants to receive under all circumstances e.g. Webhooks


tests/ops/api/test_ratelimit.py
```diff
@@ -0,0 +1,26 @@
...
+from fides.ctl.core.config import get_config
+
+CONFIG = get_config()
+
+
+def test_requests_ratelimited(api_client, cache):
+    """
+    Asserts that incremental HTTP requests above the ratelimit threshold are
+    rebuffed from the API with a 429 response.
+    """
```
seanpreston, line 13: There's a failure condition of this test where the `CONFIG.security.request_rate_limit` var is set so high that the test client can't send requests quickly enough to beat it — that shouldn't be possible with the default setting of `100/minute`.

I've tried to override the limit to something artificially small for the sake of a more reliable test with the following:
```
REQUEST_LIMIT = 1
DEFAULT_TEST_RATE_LIMIT = [f"{REQUEST_LIMIT}/minute"]


def configure_rate_limits(
    app: FastAPI,
    limits: List[str],
    key_prefix: str = CONFIG.security.rate_limit_prefix,
) -> FastAPI:
    limiter = Limiter(
        default_limits=limits,
        headers_enabled=True,
        key_prefix=key_prefix,
        key_func=get_remote_address,
        retry_after="http-date",
        storage_uri=CONFIG.redis.connection_url,
    )
    app.state.limiter = limiter
    app.add_exception_handler(RateLimitExceeded, _rate_limit_exceeded_handler)
    return app


def rate_limited_api_client():
    """
    Yields a client used to make API requests with an artificially low
    ratelimit set.
    """
    rate_limited_app = configure_rate_limits(
        app=app,
        limits=DEFAULT_TEST_RATE_LIMIT,
    )
    with TestClient(rate_limited_app) as client:
        yield client
```
however, the test seems to hang indefinitely. I believe it needs to be done this way as the rate limit is set on the `FastAPI` application at runtime.

pattisdr: Can you just clarify `There's a failure condition of this test`--this test is passing for me, so I just want to better understand where to focus

ThomasLaPiana: @seanpreston if there are caveats, please also note them in the test docstring too for future developers' sake

It sounds like this test doesn't reliably pass because the client can't reliably overwhelm the 100 requests/rolling minute, and that changing the default at runtime doesn't work either

seanpreston: This is me highlighting a known edge case, I've never seen this test fail due to the reason, but it theoretically could. Will add a note in the docstring.

ThomasLaPiana: Ah, I see! Appreciated for calling it out

tests/ops/api/test_ratelimit.py
```diff
@@ -0,0 +1,30 @@
...
+
+    response = api_client.get(HEALTH)
+    assert response.status_code == 429
+
+    ratelimiter_cache_keys = [key for key in cache.keys() if key.startswith("LIMITER/")]
+    for key in ratelimiter_cache_keys:
+        # Depending on what requests have been stored previously, the ratelimtier will
+        # store keys in the format `LIMITER/fides-/127.0.0.1//health/100/1/minute`
+        assert key.startswith(f"LIMITER/{CONFIG.security.rate_limit_prefix}")
+        # Reset the cache to not interere with any other tests
```
pattisdr, line 29: minor type: interfere

tests/ops/api/test_ratelimit.py
pattisdr: I tried something similar that seemed to work to lower the rate to "2/minute" so we didn't have to make 100+ requests for this test 

```python

def api_client_for_rate_limiting() -> Generator:
    """Return a client used to make API requests"""
    app.state.limiter = Limiter(
        default_limits=["2/minute"],
        headers_enabled=True,
        key_prefix=CONFIG.security.rate_limit_prefix,
        key_func=get_remote_address,
        retry_after="http-date",
        storage_uri=CONFIG.redis.connection_url,
    )
    with TestClient(app) as c:
        yield c
        app.state.limiter = Limiter(
            default_limits=[CONFIG.security.request_rate_limit],
            headers_enabled=True,
            key_prefix=CONFIG.security.rate_limit_prefix,
            key_func=get_remote_address,
            retry_after="http-date",
            storage_uri=CONFIG.redis.connection_url,
        )


def test_requests_ratelimited(api_client_for_rate_limiting, cache):
    """
    Asserts that incremental HTTP requests above the ratelimit threshold are
    rebuffed from the API with a 429 response.

    A theoretical failure condition exists in this test should the container
    running it not be able to execute 100 requests against the client in a
    one minute period.
    """
    for _ in range(0, 2):
        response = api_client_for_rate_limiting.get(HEALTH)
        assert response.status_code == 200

    response = api_client_for_rate_limiting.get(HEALTH)
    assert response.status_code == 429

    ratelimiter_cache_keys = [key for key in cache.keys() if key.startswith("LIMITER/")]
    for key in ratelimiter_cache_keys:
        # Depending on what requests have been stored previously, the ratelimtier will
        # store keys in the format `LIMITER/fides-/127.0.0.1//health/100/1/minute`
        assert key.startswith(f"LIMITER/{CONFIG.security.rate_limit_prefix}")
        # Reset the cache to not interere with any other tests
        cache.delete(key)
```

pattisdr (commented):
neat @seanpreston I hadn't heard of `slowapi`.  I'm not sure what a good limit is - if `100/minute` is too few but since this is configurable I think this is fine

tests/ops/api/test_ratelimit.py
```diff
@@ -0,0 +1,30 @@
...
+    for _ in range(ratelimit):
+        response = api_client.get(HEALTH)
+        assert response.status_code == 200
+
+    response = api_client.get(HEALTH)
+    assert response.status_code == 429
+
+    ratelimiter_cache_keys = [key for key in cache.keys() if key.startswith("LIMITER/")]
+    for key in ratelimiter_cache_keys:
+        # Depending on what requests have been stored previously, the ratelimtier will
```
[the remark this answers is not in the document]
ThomasLaPiana:
```suggestion
        # Depending on what requests have been stored previously, the ratelimiter will
```

tests/ops/api/test_ratelimit.py
ThomasLaPiana:
```suggestion
        # Reset the cache to not interfere with any other tests
```

docs/fides/docs/installation/configuration.md
```diff
@@ -213,6 +213,8 @@ The `fides.toml` file should specify the following variables:
 | `root_password` | string | None | If set, this can be used in conjunction with `root_username` to log in without first creating a user in the database. |
 | `root_user_scopes` | list of strings | All available scopes | The scopes granted to the root user when logging in with `root_username` and `root_password`. |
 | `subject_request_download_link_ttl_seconds` | int | `432000` | The number of seconds that a pre-signed download URL when using S3 storage will be valid. |
+| `request_rate_limit` | str | `100/minute` | The number of requests from a single IP address allowed to hit an endpoint within a rolling 60 second period. |
+| `rate_limit_prefix` | str | `fides-` | The prefix given to keys in the Redis cache used by the rate limiter. |
```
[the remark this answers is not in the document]
seanpreston: We found out with Celery config that users often want very granular control over their caches. If a user was running Fides on an Elasticache that's shared among other projects (a real use case we've encountered) they would want a guarantee here (especially if the cache was shared with another service also using `slowapi`).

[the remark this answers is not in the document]
pattisdr: ah that is good context there, makes sense

seanpreston:
I've QA'd this latest change manually, if an invalid var for the rate limit is specified, the following message shows:

```
fides-fides-1  | pydantic.error_wrappers.ValidationError: 1 validation error for FidesConfig
fides-fides-1  | security -> request_rate_limit
fides-fides-1  |   
fides-fides-1  |             Ratelimits must be specified in the format: [count] [per|/] [n (optional)] [second|minute|hour|day|month|year]
fides-fides-1  |             e.g. 10 per hour
fides-fides-1  |             e.g. 10/hour
fides-fides-1  |             e.g. 10/hour;100/day;2000 per year
fides-fides-1  |             e.g. 100/day, 500/7days
fides-fides-1  |              (type=value_error)
```

If the rate limit is exceeded, the following response is returned:

```
{
error: "Rate limit exceeded: 2 per 1 minute"
}
```

pattisdr (changes requested):
Code changes themselves looks fine but I was doing some more QA on the CLI side (since I learned this week that the CLI calls the API!) and I have concerns about this:

Steps:
1 - Set `request_rate_limit`="2/minute" in your local toml
2 - in the shell (`nox -s dev -- shell`) run `fides get data_category user` multiple times 
3 - Observe KeyError because the healthceck API response is unexpected is `{'error': 'Rate limit exceeded: 2 per 1 minute'}` - no "version" key
```
  File "/usr/local/bin/fides", line 8, in <module>
    sys.exit(cli())
  File "/root/.local/lib/python3.10/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/root/.local/lib/python3.10/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/root/.local/lib/python3.10/site-packages/click/core.py", line 1654, in invoke
    super().invoke(ctx)
  File "/root/.local/lib/python3.10/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/root/.local/lib/python3.10/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/root/.local/lib/python3.10/site-packages/click/decorators.py", line 26, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/fides/src/fides/cli/__init__.py", line 95, in cli
    check_server(VERSION, str(config.cli.server_url), quiet=True)
  File "/fides/src/fides/cli/utils.py", line 80, in check_server
    server_version = health_response.json()["version"]
KeyError: 'version'

```

So we could fix this specific location instead of throwing a KeyError, to raise an exception that the rate limit was exceeded. But more broadly, are there other areas where the CLI is making lots of requests where we'll easily exceed the limits? (Beyond my area of expertise)

src/fides/ctl/core/config/security_settings.py
```diff
@@ -15,7 +17,30 @@ class SecuritySettings(FideslibSecuritySettings):
...
+        except ValueError:
+            message = """
+            Ratelimits must be specified in the format: [count] [per|/] [n (optional)] [second|minute|hour|day|month|year]
+            e.g. 10 per hour
+            e.g. 10/hour
+            e.g. 10/hour;100/day;2000 per year
+            e.g. 100/day, 500/7days
+            """
+            raise ValueError(message)
+        return v
```
pattisdr, line 43: Thank you for adding this nice validation - the example is very helpful 🏆

seanpreston:
> So we could fix this specific location instead of throwing a KeyError, to raise an exception that the rate limit was exceeded. But more broadly, are there other areas where the CLI is making lots of requests where we'll easily exceed the limits? (Beyond my area of expertise)

This is a really good find, thanks. One thing I can think of to immediately address this is to increase the rate limit default, something like 500/minute or 1000/minute still offer us good benefits here, because they still essentially stop brute force attacking of sensitive endpoints.

Another thing I can think to do is remove the limits entirely where there's no sensitive data being exposed.

pattisdr:
OK that sounds good, can you address the possible KeyError too when you get a chance?

pattisdr (approved):

PR #1693, opened 2022-11-04
238/Scan complete success alert

LKCSmith:
Closes [this issue](https://github.com/ethyca/fidesctl-plus/issues/238).

https://user-images.githubusercontent.com/99051851/201233086-ff308050-5c29-4751-8d75-bba949c5107d.mov

Note: I believe error handling is already being taken care of in the system scanner [here](https://github.com/ethyca/fides/blob/main/clients/admin-ui/src/features/config-wizard/AuthenticateAwsForm.tsx#L106). A [separate issue was created for pagination](https://github.com/ethyca/fidesui/issues/28) in tables for reusability. This PR will address just the success alert of the issue ask.

### Code Changes

* [x] Add a success alert after the scanner completes a successful scan

### Steps to Confirm

* [ ] _list any manual steps taken to confirm the changes_

### Pre-Merge Checklist

* [ ] All CI Pipelines Succeeded
* Documentation Updated:
  * [ ] documentation complete, or draft/outline provided (tag docs-team to complete/review on this branch)
  * [ ] documentation issue created (tag docs-team to complete issue separately)
* [ ] Issue Requirements are Met
* [ ] Relevant Follow-Up Issues Created
* [ ] Update `CHANGELOG.md`

### Description Of Changes

_Write some things here about the changes and any potential caveats_

PR #1697, opened 2022-11-04
239/System scan registered systems alert

LKCSmith:
Closes [this issue](https://github.com/ethyca/fidesctl-plus/issues/239).

### Code Changes

* [ ] Add success alert when registering systems.

### Steps to Confirm

### Pre-Merge Checklist

* Documentation Updated:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

### DIFF
--- a/clients/admin-ui/src/features/config-wizard/ScanResultsForm.tsx
+++ b/clients/admin-ui/src/features/config-wizard/ScanResultsForm.tsx
@@ -16,6 +16,7 @@ import {
   ColumnMetadata,
 } from "~/features/common/ColumnDropdown";
 import { useFeatures } from "~/features/common/features.slice";
+import { isErrorResult } from "~/features/common/helpers";
 import { useAlert, useAPIHelper } from "~/features/common/hooks";
 import { resolveLink } from "~/features/common/nav/zone-config";
 import { SystemsCheckboxTable } from "~/features/common/SystemsCheckboxTable";
@@ -23,7 +24,6 @@ import WarningModal from "~/features/common/WarningModal";
 import { useUpsertSystemsMutation } from "~/features/system";
 import { System } from "~/types/api";
 
-import { isErrorResult } from "~/features/common/helpers";
 import {
   changeStep,
   chooseSystemsForReview,
--- a/clients/admin-ui/src/features/config-wizard/ScanResultsForm.tsx
+++ b/clients/admin-ui/src/features/config-wizard/ScanResultsForm.tsx
@@ -22,8 +22,8 @@ import { SystemsCheckboxTable } from "~/features/common/SystemsCheckboxTable";
 import WarningModal from "~/features/common/WarningModal";
 import { useUpsertSystemsMutation } from "~/features/system";
 import { System } from "~/types/api";
-import { isErrorResult } from "../common/helpers";
 
+import { isErrorResult } from "../common/helpers";
 import {
   changeStep,
   chooseSystemsForReview,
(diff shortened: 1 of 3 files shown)


LKCSmith:
Need to look into the error handling -

clients/admin-ui/src/features/system/SystemRegisterSuccess.tsx
```diff
@@ -44,6 +46,12 @@ const SystemRegisterSuccess = ({ system, onAddNextSystem }: Props) => {
     dispatch(setActiveSystem(undefined));
 
     router.push("/system");
+
+    successAlert(
+      `You have successfully added ${allRegisteredSystems?.length} systems to your Data Map`,
+      `${allRegisteredSystems?.length} Systems successfully added to your Data Map`,
+      { isClosable: true }
+    );
```
allisonking, line 54: Are you sure this is the file you meant to change? As we've talked about before, somehow `SuccessPage` has come back, and that's the one that shows up at the end of the config wizard. I think we'd have to put this in both, at least until we figure out how to merge those two files back to one again (can you make a ticket for that?)

clients/admin-ui/src/features/config-wizard/ScanResultsForm.tsx
```diff
@@ -48,19 +50,31 @@ const ScanResultsForm = () => {
...
+      const datamapRoute = resolveLink({
+        href: "/datamap",
+        basePath: "/",
+      });
+
+      successAlert(
+        `You have successfully added ${selectedSystems?.length} systems to your Data Map`,
+        `${selectedSystems?.length} Systems successfully added to your Data Map`,
+        { isClosable: true }
+      );
```
allisonking, line 72: I'm pretty certain this is not going to work for the datamap view because it's technically a different Nextjs app which the toast won't know about. I think the only way to navigate to the other app while also showing a toast might be to pass a query param, or maybe something else sneaky in `window.location`. Maybe our resident inter-zone navigation expert @ssangervasi has an idea

allisonking (commented):
@LKCSmith in the future, could you please record a video or show some screenshots of the changes you're making? helps me a lot to understand the context of the PR, especially now that I'm not working as directly with this code

clients/admin-ui/src/features/system/SystemRegisterSuccess.tsx
LKCSmith: Both of these files are the correct file - I'll add a screenshot to the PR comment, but this alert is meant to show on the `/system` or `/datamap` route  **after** successfully registering (a) system(s). I added the alert here after the re-route to `/system` as well as the `ScanResultsForm.tsx` after successfully registering systems and redirecting to either `/system` or `/datamap`.

clients/admin-ui/src/features/config-wizard/ScanResultsForm.tsx
LKCSmith: Oh, interesting. I'll look into this more.

ssangervasi: Yeah, a query parameter is a pretty classic way of triggering something like this on a redirect. That would Plus would need to check its router on page load, trigger the alert with that text, and clear the query parameter.

One other way we have of communicating between the apps is with localStorage, as of https://github.com/ethyca/fidesctl-plus/pull/303. That's more complicated, though, so probably not worth it for this case.

And then I have one hack idea which is to simply delay the `router.push` for ~1 second, so the toast is visible before the page changes and it goes away. Not quite the experience we're designing for, though.

LKCSmith: Gotcha. I'll work on the plus updates as well in conjunction with this and [238](https://github.com/ethyca/fides/pull/1693).

ThomasLaPiana:
@LKCSmith this seems stale, can you confirm this is ok to close?

LKCSmith:
@ThomasLaPiana - Yes, closed. Thank you 🙏

PR #1712, opened 2022-11-08
ui/taxonomy: CustomSelect isClearable - handle empty string selection

### LINKED ISSUE #1504: Allow single select lists to be cleared in Taxonomy management
There is no "Clear" or "None" option on the special category and legal basis dropdowns in taxonomy management, so I can't correct a mistake.

**Details**
There might be a component that we can use from Fidesops screens on this. 
All existing fidesctl inputs are in a single place
Might want to pass a flag that indicates a field is "clearable"

**AC**
- Given that I make a selection in any single select of the Taxonomy UI, when it is selected, then I can remove my selection to leave a null input.
- Given that I make a selection in any single select of the System Management UI, when it is selected, then I can remove my selection to leave a null input.

Image

ssangervasi:
Closes #1504 

### Code Changes

* Use `isClearable` during `onChange` in forms.
* Apply this to taxonomy Data Use

### Steps to Confirm

* [Video of clearing multiple.](https://user-images.githubusercontent.com/2236777/200447047-c585b93d-42b6-4283-bb28-c1bd0fac81da.mp4)

### Pre-Merge Checklist

* Documentation Updated:
* [x] Update `CHANGELOG.md`

### Description Of Changes

The single-select clearing functionality already existed, but we hadn't tried it anywhere. The only thing it was missing was assigning the value to a "cleared" value. `react-select` doesn't want to use `null` or `undefined` for this state, which is reasonable since the native HTML way of doing it is to set an empty string (whenever the value would not match one of the option elements). 

This empty-string case then needs to be handled when transforming the form into an object to submit to the server.

I looked for other uses of `CustomSelect` where the value should be clearable, but they all seemed required to me. For example, submitting a null/undefined/empty-string `data_qualifier` is not accepted by the server.

clearing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ The types of changes are:
 * Removed local storage parsing that is redundant with redux-persist. [#1678](https://github.com/ethyca/fides/pull/1678)
 * Show a helpful error message if Docker daemon is not running during "fides deploy" [#1694](https://github.com/ethyca/fides/pull/1694)
 * Allow users to query their own permissions, including root user. [#1698](https://github.com/ethyca/fides/pull/1698)
+* Single-select taxonomy fields legal basis and special category can be cleared. [#1712](https://github.com/ethyca/fides/pull/1712)
 
 ### Security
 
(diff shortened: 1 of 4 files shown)


LKCSmith (approved):
👍

chriscalhoun1974 (approved):

Issue #1572, opened 2022-10-26
Merge `fideslib` into `unified-fides`


ThomasLaPiana:
### Is your feature request related to a specific problem?

Now that everything is merged into a single project, the additional friction of updating `fideslib` is no longer worth it since it is no longer enabling code-sharing. Case in point, [PRs like this](https://github.com/ethyca/fides/pull/1523/files) become more complicated as we end up circumventing `fideslib` altogether instead of trying to fix it properly and cut another release.

### Describe the solution you'd like

Copy/pasta, keeping the git history doesn't seem particularly important here given the additional complexity

ThomasLaPiana:
@NevilleS @seanpreston FYI this is on my mind. That repo is now devoid of its original purpose and should be folded back in (the arms of `unified-fides` are ever-open to embrace more code)

PR #1718, opened 2022-11-08
Fix security config race condition

sanders41:
Closes <issue>

### Code Changes

* [x] Pre-load the config in the `__init__.py`
* [x] Load the security setting before loading the full model
* [x] init `SecuritySettings` directly if the section doesn't exist in the `toml` config file
* [x] Added `_FUNCS.clear()` to clear Pydantic's validator cache before loading
* [x] DO NOT include the security defaults in the `fides.toml` file

### Steps to Confirm

* [ ] Build the container `nox -s "build(dev)"`
* [ ] `docker run -it ethyca/fides:local` fails on the missing config items
* [ ] `docker run -e FIDES__USER__ANALYTICS_OPT_OUT=True -e FIDES__SECURITY__APP_ENCRYPTION_KEY=kfkdkslaksldkfjdkslaksldkfjskald -e FIDES__SECURITY__OAUTH_ROOT_CLIENT_SECRET=rootclient -e FIDES__SECURITY__OAUTH_ROOT_CLIENT_ID=someclient ethyca/fides:local` passes

### Pre-Merge Checklist

* Documentation Updated:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

Fixes the issue where the security config is not properly loading from environment variables.

The root cause of this is Pydantic was delaying the loading of nested models when using environment variables. Because of this if the nested value was accessed before it was loaded an error occurred. The fix was to pre-load the nested model and add it to the parent model at initialization when loading from environment variables.

In doing this pre-loading it turned up a new issue where in some environments (CI for example) Pydantic would through an error that validators were being reused, while in other environments (running locally for example) no error occurred. This seems to occur because of a conflict between server cache and Pydantic's cache. I did try removing our `lru_cache` to see if that was involved in the issue, but it made no difference. The solution turned out to be clearing Pydantic's cache before loading the model with `_FUNCS.clear()`. More context for this can be found [here](https://github.com/streamlit/streamlit/issues/3218).

Other changes in this PR, bumping Pydantic and bringing fideslib code directly in, were done as attempts to solve this Pydantic issue before the cache clearing option was found. In the end they didn't make any difference, but we left here since this change is planned as part of #1572.


sanders41:
@ThomasLaPiana these tests are passing when I run them locally. Any chance the failures could be related to the recent CI updates?

ThomasLaPiana:
@sanders41 no, the latest `main` merge shows only 4 tests failing, all known-bad (Shopify errors + the Admin UI Cypress tests)

These are probably from this PR

ThomasLaPiana:
this seems like a pretty sprawling change, simple but touches a ton of code. I'm sure you've already thought about it, but maybe there's a way to add one or two checks somewhere instead of in every function that might need it? This will get tricky to maintain/revert/document

sanders41:
I tried several other ways to do the checks, but mypy was failing any other way. I don’t like the way I did this either, but I couldn’t come up with another way. 

Troubleshooting the failures is also going to be difficult since they pass locally and the error is with the database config which didn’t change. The only difference I can think of right now is dev vs production builds, but don’t see why that would matter.

sanders41:
I'm still struggling to reproduce this error locally. I tried building and testing with the production image and that didn't do it either.

sanders41:
Possibly related https://github.com/nazrulworld/fhir.resources/issues/41. If so it could be a caching issue. It's the same scenario, but in the case of this issue it was with lambda and not github actions.

NevilleS:
Hey @sanders41, I think you can reproduce this pretty easily by doing the following:
```
docker run ethyca/fides:local
```

...and then start layering in minimal ENV variables, like:
```
docker run -e FIDES__USER__ANALYTICS_OPT_OUT=True -e FIDES__SECURITY__APP_ENCRYPTION_KEY=keygoeshere ethyca/fides:local
```

sanders41:
This is what I did and it was running without error locally. I got it working both locally and in ci now, but the docs build is failing still. running a test now where I hope I have that fixed also.

sanders41:
@ThomasLaPiana I finally found a way to get everything passing in CI and running locally. It should be ready for review now.

requirements.txt
```diff
@@ -27,7 +27,7 @@ pandas==1.4.3
 passlib[bcrypt]==1.7.4
 plotly==5.4
 psycopg2-binary==2.9.3
-pydantic<1.10.0
+pydantic<1.10.2
```
ThomasLaPiana, line 30: is there an actual fix here or just got bumped in the attempt to figure out the bug?

Either way it's fine, I'm just curious!

requirements.txt
```diff
@@ -44,4 +44,5 @@ SQLAlchemy-Utils==0.38.3
 toml>=0.10.1
 twilio==7.15.0
 Unidecode==1.3.4
+validators==0.20.0
```
ThomasLaPiana, line 47: Ohh, is this for pydantic specifically?

ThomasLaPiana:
@sanders41 can you add some more details to the PR description here? Just so I/anyone else can look at this PR and understand more clearly _how_ the problem was solved

ThomasLaPiana:
Still getting a failure when I run it locally 😞 

Command: 
`docker run -e FIDES__USER__ANALYTICS_OPT_OUT=True -e FIDES__SECURITY__APP_ENCRYPTION_KEY=kfkdkslaksldkfjdkslaksldkfjskald -e FIDES__SECURITY__OAUTH_ROOT_CLIENT_SECRET=rootclient -e FIDES__SECURITY__OAUTH_ROOT_CLIENT_ID=someclient ethyca/fides:local`

Result:
```
pydantic.error_wrappers.ValidationError: 1 validation error for FidesConfig
security
  field required (type=value_error.missing)
```

src/fides/ctl/core/config/__init__.py
```diff
@@ -213,6 +230,10 @@ def get_config(config_path_override: str = "", verbose: bool = False) -> FidesCo
     except IOError:
         echo_red("Error reading config file")
 
-    config = FidesConfig()
+    # database_settings = DatabaseSettings()
```
ThomasLaPiana, line 233: code comments?

src/fides/ctl/core/config/__init__.py
```diff
@@ -206,13 +223,22 @@ def get_config(config_path_override: str = "", verbose: bool = False) -> FidesCo
             credentials_dict=config_environment_dict
         )
 
+        # This is required to set security settings from the environment
+        # if that section of the config is missing
+        if not settings.get("security"):
```
ThomasLaPiana, line 228: @sanders41 this was the required fix, and more evidence we need a larger config logic refactor...here's what was happening....

When running a CLI command for the first time, it detects whether or not a config file exists. If it doesn't exist, it will create one from defaults. For some reason, this wasn't grabbing the config with the security values initialized (this must be a bug in the `get_config` logic), so it was writing out a `toml` file with no `security` section. It would then read this file in and complain that that section didn't exist (which is correct).

This really needs to get cleaned up...I'd love to tackle this after unified fidesplus, as currently we have logic (I'll take the blame for this!) spread across the CLI entrypoint, CLI utils, and the config module....it makes it really hard to track these bugs down and fix things, and is quickly getting to the point of being unmaintainable (as evidenced by the complexity of the fix)

ThomasLaPiana (changes requested):
Code comment cleanup, better explanation of changes/fixes

ThomasLaPiana:
we also need to add some sort of test for this, as this was basically verified 100% through manual testing

requirements.txt
sanders41: I was trying to solve the validator reuse error in CI (this is going to be the theme to everything here) and I saw an older Pydantic issue where this error started happening after a python upgrade. They got the issue fixed, so I figured it was a long shot but worth a try since the issue had been closed for a long time. Turned out to make no difference, but since everything else was still working I figured I may as well leave the newer version in.

requirements.txt
sanders41: This one we brought into fideslib because one of our users (left unmentioned here for confidentiality) wanted to be able to specify urls as partial urls (https://my-site* and things like this). FastAPI could handle this, but Pydantic's url validator couldn't. Using this validators package let us specify the url as a string for pydantic, and still validate it.

sanders41:
> Weirdly enough I've also confirmed that this specifically breaks on webserver, but running fides view config works just fine and shows the security config as expected.... 🙃

This is the kind of thing that made this so difficult to solve! Things that seemingly should act the same didn't. I would think I had everything going and find another way to break it...and thankfully you found even more in your review.

ThomasLaPiana:
im testing this in the [plus PR](https://github.com/ethyca/fidesctl-plus/pull/308) and still seeing the security errors....still digging

ThomasLaPiana:
If there is a config file, it will fail to init from the environment variables for the security settings

ThomasLaPiana:
fixed the docs checks, next step is to fix the config tests (they need the required env vars now!)

src/fides/ctl/core/config/__init__.py
```diff
@@ -46,7 +43,9 @@ class FidesConfig(BaseModel):
     logging: LoggingSettings = LoggingSettings()
     notifications: NotificationSettings = NotificationSettings()
     redis: RedisSettings = RedisSettings()
-    security: SecuritySettings = {}  # type: ignore
+    # Use 'construct' to avoid running validation
+    # This allows us to use the correct type but populate later
+    security: SecuritySettings = SecuritySettings.construct()
```
ThomasLaPiana, line 48: @sanders41 i think this was the final solution...doing this here and then making sure to populate the settings correctly below

ThomasLaPiana:
@NevilleS @sanders41 I want to call out that now, if a user does a `pip install ethyca-fides`, they will not be able to do _anything!_ until they've set these values, and the failure will happen before we have a chance to tell them

sanders41:
Would deferring the loading of the config help with this? Adding the load to the `__init__.py` was one of the first things done in trouble shooting and it did help, but there have been a lot of other changes since then so we might be able to get away with removing it.

ThomasLaPiana:
@sanders41 I think we can cross check this PR and then merge it, and do that UX improvements in another PR as this one is getting big

TheAndrewJackson:
I think everything is working on my end. It was able to load the config but failed when trying to connect to the db with the provided snippet

```
docker run -e FIDES__USER__ANALYTICS_OPT_OUT=True -e FIDES__SECURITY__APP_ENCRYPTION_KEY=kfkdkslaksldkfjdkslaksldkfjskald -e FIDES__SECURITY__OAUTH_ROOT_CLIENT_SECRET=rootclient -e FIDES__SECURITY__OAUTH_ROOT_CLIENT_ID=someclient ethyca/fides:local
```

I think that is expected because no db config is provided so it used `default-db`

ThomasLaPiana:
I got this branch working with the `plus` PR, so I'm going to merge this. Will immediately open a follow-up ticket around the user experience

ThomasLaPiana (approved):

Issue #1514, opened 2022-10-21
Fix the Stepper in the config wizard


mfbrown:
As a user, I want the Config Wizard Stepper to reflect the steps which are required for the flow that I select, so that I have an accurate preview of the steps required to add new systems.

**AC**
- Given that I select the manual inputs, when the first step loads, then the stepper updates to show 3 scanning steps
- Given that I select the AWS Scanner, when the first step loads, then the stepper updates to show steps specific to the AWS Scanning
- Given that I select the Okta Scanner, when the first step loads, then the stepper updates to show steps specific to the Okta scanner.

**Figma**
- Updated stepper: https://www.figma.com/file/NeFvOOb1RT2fcdqZT0vBF2/Fidesctl-Management-UI?node-id=110%3A6861
- Okta Scanning: https://www.figma.com/file/NeFvOOb1RT2fcdqZT0vBF2/Fidesctl-Management-UI?node-id=108%3A5125
- AWS Scanning: https://www.figma.com/file/NeFvOOb1RT2fcdqZT0vBF2/Fidesctl-Management-UI?node-id=29%3A943
- Manual Input: https://www.figma.com/file/NeFvOOb1RT2fcdqZT0vBF2/Fidesctl-Management-UI?node-id=110%3A6861

TBD: Just need to determine the final number of steps.

Kelsey-Ethyca:
Note to myself to add these details to the ticket https://ethyca.slack.com/archives/C0114BAK160/p1666633875802189

Kelsey-Ethyca:
[My "designs"]( https://www.figma.com/file/NeFvOOb1RT2fcdqZT0vBF2/Fidesctl-Management-UI?node-id=29%3A943) are in the pink box. We should review with Design Team and have a mini planning meeting to finalize

LKCSmith:
@mfbrown - I am still extremely confused by the Figma designs I am meant to follow, and when I click your Figma links, for some reason they don't lead to a specific node or design. There are also multiple flows on the page.

Could you please show me (maybe in screenshots added here for simplicity):
1) The starting state of the stepper, before a scanner or the manual entry is chosen
2) What the stepper looks like with manual entry
3) What the stepper looks like with AWS
4) What the stepper looks like with Okta

LKCSmith:
Moving to blocked until the above is clearer to me ^ Just want to make sure we're all on the same page for the stepper. Thank you.

Kelsey-Ethyca:
[...]

PR #1720, opened 2022-11-08
Trigger classify on onboarding k8s system scan

allisonking:
Closes https://github.com/ethyca/fidesctl-plus/issues/190

### Code Changes

* [x] Moves storing the way the user is adding systems (aws, runtime scanner, okta, etc.) into redux
* [x] Adds a `classify=true` query param to the existing scan endpoint (backend work still [in progress](https://github.com/ethyca/fidesctl-plus/pull/304))
* [x] Refactors the ScanResults page to `RegisterSystems`. This was more useful when I thought `/classify-systems` wasn't going to be its own page and was just going to be a view stored in state. That's no longer the case after talking with design today, but I think the refactor is still a bit cleaner
* [x] Adds a new page `/classify-systems` with a classification table
* [x] Refactors the dataset classification badge into its own component so we can reuse it in this systems classification table
* [x] Renders the classification status in the table 
* [x] Adds cypress tests

### Steps to Confirm

* [ ] The backend endpoints don't quite exist yet, so your best bet is to run the cypress test "Allows calling the runtie scanner with classify". Note that the fixture only has two classification instances, hence why there are only two rows with "Awaiting Review" and everything else is "Unknown"

### Pre-Merge Checklist

* Documentation Updated:
* [x] Update `CHANGELOG.md`

### Description Of Changes

https://user-images.githubusercontent.com/24641006/201762900-7f867111-870b-4dac-bab6-0c36b521cfb1.mov


clients/admin-ui/cypress/e2e/config-wizard-plus.cy.ts
```diff
@@ -71,11 +78,14 @@ describe.skip("Config wizard with plus settings", () => {
         const { body } = interception.request;
         expect(body.length).to.eql(numSystems);
       });
-      /* This will redirect to localhost:3000/datamap instead of localhost:4000/datamap
-       * which is not actually what we want.
-       * However, it is difficult to test across zones, so all we can do is assert the path is right
-       */
-      cy.url().should("contain", "datamap");
+      cy.getByTestId("systems-classify-table");
```
ssangervasi, line 81: I tried running the tests by removing the `describe.skip` and enabling the feature flag to make the page appear. The test fails here, looking like this:
Screen Shot 2022-11-11 at 15 17 11

clients/admin-ui/src/features/config-wizard/AuthenticateRuntimeForm.tsx
```diff
@@ -19,7 +19,9 @@ import ScannerLoading from "./ScannerLoading";
 const AuthenticateRuntimeForm = () => {
   const dispatch = useAppDispatch();
   const toast = useToast();
-  const { data, error: queryError } = useGetScanResultsQuery();
+  const { data, error: queryError } = useGetScanResultsQuery({
```
ssangervasi, line 22: Interesting that this is a GET request. And by interesting I mean "yikes". It kicks off a process that will, among other things, create ClassifyInstances, right? Unless I'm missing something.

Not a UI code problem, if that's case, just raised my eyebrows. Maybe the component shouldn't be called "Authenticate", though, if were really kicking it off just by entering this step.

allisonking: ahh right, that's a good point. we had planned to change it over to PUT as part of https://github.com/ethyca/fidesctl-plus/issues/195 (which will write scan history), but you're right that we should do it now that we're writing classify instances. @SteveDMurphy 's being a hero and said he'll change it to a PUT as part of https://github.com/ethyca/fidesctl-plus/pull/304, and then I'll update the frontend in https://github.com/ethyca/fides/pull/1752

Yeah, it probably shouldn't be called Authenticate. it was the easiest way to fit it into the existing onboarding flow though. I think https://github.com/ethyca/fides/issues/1514 will make the "steps" more true to each flow, and then this component won't have to be stuffed into "Authenticate". I'll rename it for now though 👍

clients/admin-ui/cypress/e2e/config-wizard-plus.cy.ts
ssangervasi: I'm still getting an error just on `Allows calling the runtime scanner with classify`

ssangervasi (approved):
Just one funky test

allisonking:
> Just one funky test

confirmed with @ssangervasi that it works after https://github.com/ethyca/fides/pull/1720/commits/2bf3d53f110992c9201723ee839803a8fd531a55, so merged! thanks for the thorough review 🙏

PR #1731, opened 2022-11-09
Populate Application DB instead of the Test DB

### LINKED ISSUE #1175: [Unified Fides] Populate Application DB instead of Test DB

Currently, Fides has a "Test mode" setting which populates the test db when True, and the application db when False.  Test mode seems to be True by default.  So when running the application locally, it is the test db that is getting populated.  

Fidesops had "test mode" as False by default. It was not meant to be set as True by anyone, it was just a variable that Pytest set to True to have the Test DB be used. 

For the Fidesops side of things, it is useful to have a separate database from the test database when you're testing privacy requests locally, and depending on some resources to persist there without being created and torn down by the testing harness.

- Populate the normal application database during regular development
- Have Pytest set up and tear down the test database for its own testing 
- Get rid of the "Test Mode" variable, the application doesn't need to set this. Pytest itself should just use the test db.

---

*EDIT ON October 20, 2022 by @ThomasLaPiana*:

After more discussion in the initial PR that is now closed, it was determined that the right way to go forward is the following:

* Remove test mode
* Update ctl database access methods to support injecting/swapping database connections after startup
* Verify that both ops and ctl use the test url when running pytest, and test runs leave the application db untouched

pattisdr:
❗ This could create breaking changes in the -plus repos that would need to be handled before they are upgraded to use this 

Closes #1175 

### Code Changes

Mostly ctl-side changes: 

#### Environment Variables

- Don't set `FIDES__TEST_MODE` in the `docker-compose.yml` - this should be **set by pytest only**. So this is _only True when running pytest_. 
- Update incorrect variable name `TESTING` to  `FIDES__TEST_MODE` in `pyproject.toml` to set test mode to true for pytest only.  `TESTING` was the old variable we used in ops.
- Update base credentials > app_postgres to be a connection string pointing to the `fides` db instead of the `fides_test` db in the `.fides/fides.toml`.  Running `fides scan dataset db --credentials-id app_postgres` locally should scan the application db not the test db.

#### Sessions
- Dependency injection: Inject an async session as a dependency into the ctl `crud`, `datamap`, `view`, and `generate` routes.  Pass that session onto the crud methods.  (Ops already injects a session into its routes. ) The admin route doesn't seem to need a session because it is using database_uri's that have test_mode awareness. The `health` endpoint is already injecting the sync session from ops.  The validate routes don't seem to need a session either.
    - Create an `async_db` generator for API endpoint dependency injection.  
- Update the ctl database crud methods to take in an `async_session` as a variable (`api/ctl/database/crud.py`), which is test_mode aware
- Update ctl route utils to take in an `async_session` parameter (`api/ctl/routes/util.py`) that is test mode aware

#### Test refactors

- Add an async_session to the ctl conftest for use in testing methods that need an async_session
- Remove unused user fixtures from the ctl conftest
- Monkeypatch several methods in the `requests` library to use the `TestClient` (which calls functions directly instead of sending a request to a running webserver). This is primarily for the methods in `src/fides/ctl/core/api.py`.  Now all the CLI-related tests and all methods calling `api.get `or `api.pos`t etc. should be usingthe TestClient instead of the requests library.

#### Ops changes
Ops was already using a `TestClient` to test its endpoints, which does not ping a running webserver.

- Update ops test  `setup_db` to reset the test db, not the application db (by using the test client).

#### Misc 
- Fix existing logging tests to remove need for unnecessary `toggle_testing_envvar` fixture
- Have `logging_sessions` log debug logs on dev mode, not test mode (test mode is false now, when running application locally). Running the application had been resulting in fewer logs now that test mode was false.
- Have `load_default_organization` and `load_default_taxonomy` take in an async_session (ctl-side)

### Steps to Confirm

* [x] Run `nox -s dev -- postgres` to bring up the webserver
* [x] Connect to the `fides` application db `psql -h localhost -p 5432 -U postgres -d fides` locally and verify that this application db is up and populated
* [x] Run a privacy request against the postgres db and verify the supporting resources are all created in the application db
* [x] Try to connect to the test db and observe it doesn't exist yet `psql -h localhost -p 5432 -U postgres -d fides_test`
* [x] Now bring up the shell `nox -s dev -- postgres shell`
* [x] Run some tests in pytest `pytest tests/ctl/database/test_crud.py`, put a breakpoint in the test after resources are created
* [x] Connect to the test db ``psql -h localhost -p 5432 -U postgres -d fides_test`.  Observe the test db was created, and the resources exist in the test db
* [x] Query the application db - observe that it is not wiped, and previously created resources still exist
* [x] Run `fides get data_category user` - observe successful return from application db

### Pre-Merge Checklist

* Documentation Updated:
* [x] Update `CHANGELOG.md`

### Description Of Changes

Fides is currently writing to the test db both when running the application and when we're running pytest.  This is difficult for local development - when you run tests in pytest, the test db is wiped and populated with fixture data.  It would be useful to populate the application db when running the application locally and run the test db when running pytest. 

Note that before the merge of fidesctl and fidesops, fidesops already had this division, writing to the application db or the test db depending on the context .  Fidesctl differed in that it spun up a webserver and the tests made requests against that running webserver. Fidesops used test clients.  This refactor primarily injects sessions into ctl-routes, and then uses test clients that have awareness of test mode to talk to the test db during testing.  However, for the CLI tests that are talking to a running webserver, we're monkeypatching the `requests` library to use the TestClient instead.

### Note

I want to make sure we're not being too aggressive with the monkey patching - ops uses the requests library to make connector requests in many places. I am only monkeypatching the requests library ctl side.


tests/ctl/core/test_system.py
```diff
@@ -7,44 +7,47 @@
...
+        test_client.delete(
+            generate_resource_url(
+                test_config.cli.server_url, "system", resource_id=system.fides_key
+            ),
             headers=test_config.user.request_headers,
         )
 
 
 @pytest.fixture(scope="function")
 def create_test_server_systems(
```
pattisdr, line 39: This fixture is now using the `test_client` directly which is aware of test_mode, instead of using the `api.create`/`api.delete` methods so I believe this fixture is creating the resources on the test db.  

Unfortunately I think methods outside of the ctl API (which are injecting sessions) are still talking to the application db, and I'm not sure how to resolve yet - So `test_get_all_server_systems` is failing but I think that it's looking for the resources in the application db.

tests/ctl/core/test_system.py
```diff
@@ -7,44 +7,47 @@
...
+    test_config: FidesConfig, redshift_systems: List[System], test_client
 ) -> Generator:
     systems = redshift_systems
-    delete_server_systems(test_config, systems)
-    create_server_systems(test_config, systems)
+    delete_server_systems(test_config, systems, test_client)
+    create_server_systems(test_config, systems, test_client)
     yield
-    delete_server_systems(test_config, systems)
+    delete_server_systems(test_config, systems, test_client)
```
pattisdr, line 46: This fixture is an example using the`test_client` directly which is aware of test_mode, instead of using the `api.create`/`api.delete` methods so I believe this fixture is creating the resources on the test db.  I made similar changes across several tests.

I've updated the CTL API to inject sessions into the routes, which seems to be working well.  But I think separate methods in src/fides/ctl/core/api.py are still talking to a webserver that's using the application db.   I'm not sure how to resolve yet -  will look more tomorrow - The test `test_get_all_server_systems` is failing because it's probably looking for the resources in the application db because it's calling methods in core/api.  There are a couple of other files that aren't failing per se, but are also writing to the application db.

pattisdr, line 46: This fixture is an example using the`test_client` directly which is aware of test_mode, instead of using the `api.create`/`api.delete` methods so I believe this fixture is creating the resources on the test db.  I made similar changes across several tests.

I've updated the CTL API to inject sessions into the routes, which seems to be working well.  But I think separate methods in src/fides/ctl/core/api.py are still talking to a webserver that's using the application db.   I'm not sure how to resolve yet -  will look more tomorrow - The test `test_get_all_server_systems` is failing because it's probably looking for the resources in the application db because it's calling methods in core/api.  There are a couple of other files that aren't failing per se, but are also writing to the application db.

Do you have any pointers @ThomasLaPiana?

pattisdr: Thanks for weighing in, I also shared that sentiment when looking at the CLI code and helpers. First I'd like to look into untangling the CLI from the API, perhaps having both the CLI and API call the same methods instead of the CLI calling the API directly.

ThomasLaPiana: I might've slightly misunderstood, but to clarify the delineation should be:

* CLI - always uses API calls to talk with the server
* API - DOES NOT use CLI-style API calls to talk to itself, instead uses its own internal functions to communicate with the database directly when needed

pattisdr: I misunderstood you when you said 

> I think refactoring/removing those calls might be viable
> This is a weird (and in my opinion not optimal place) where the CLI code/helpers feel like they're a bit too intertwined with the API code.  

So I explored replacing the API calls in `fides/ctl/core/api.py` with internal functions that communicated with the database directly (that perhaps the API itself could share too, flipping what is what calling what) but I abandoned that after awhile, thinking that even though I didn't like the server calling itself, now wasn't the time to make that refactor. I didn't realize the CLI should be intentionally using API calls to talk with the server, but either way, the CLI is still using API calls to talk to the server.

I spent the remainder of the day working on your other suggestion, trying to create a client, similar to how we do in testing, instead of using the `requests` methods in `fides/ctl/core/api.py`.  Specifically I was trying to use the `httpx` `AsyncClient`.  If I pass in the app to that AsyncClient in testing I get the
RunTimeError: got Future <Future pending> attached to a different loop, I assume the client's trying to create its own loop? If I don't pass in the app, then it makes the request as expected, but against the webserver that's running the application db not the test db.

In summary, I think most of this is working (using a `test_client` is now causing it to use the test db) except when the server tries to call itself. Everything I've tried there still has it querying the application db not the test db in testing.

pattisdr: To make sure we're on the same page, can you give code examples of:

CLI - always uses API calls to talk with the server. 
API - DOES NOT use CLI-style API calls to talk to itself


I want to make sure where we're talking about `cli/commands/core.py` for example versus `ctl/core/api.py` versus `ctl/routes/crud.py`

pattisdr: Update: for the methods in `fides/ctl/core/api.py`, the latest commit is passing in a TestClient for test mode only which is writing to the test db .  Otherwise, it's still using `requests` library to talk to the running webserver and the application db.

ThomasLaPiana: Sorry, let me clarify! Although I think you got it :) 

I don't think the webserver code logic should ever be using `fides.ctl.core.api` functions to call itself...that seems strange and is what is leading to this issue I think?

If code is running on the server, it seems like it should be able to call those functions directly instead of making an API call.

If we managed to solve this without needing to refactor all of that, I'm totally for it! There's a code smell there for sure though, at least to my "code nose" haha

src/fides/ctl/core/api.py
```diff
@@ -1,9 +1,35 @@
...
 import requests
+from starlette.testclient import TestClient
 
 from fides.api.ctl.routes.util import API_PREFIX
+from fides.ctl.core.config import get_config
+
+CONFIG = get_config()
+
+
+def verify_client(func: Callable) -> Callable:
```
ThomasLaPiana, line 14: Directly mixing in test logic with code logic feels suboptimal to me, I think there might be another solution.

I agree that passing in the client feels like a good move here, maybe there is some kind of mocking/testing library specifically for `requests`, so the application code never has to know if it is using a test client or a normal client?

I found the following SO thread that has some potentially useful solutions:

https://stackoverflow.com/questions/9559963/unit-testing-a-python-app-that-uses-the-requests-library

ThomasLaPiana: I might also be totally off-base here and not grasping that potentially the `TestClient` can't be replaced by another library

tests/ctl/core/test_system.py
pattisdr: I'm still not understanding you then - can we chat tomorrow? 

The CLI is the primary item making API calls to talk to the server, and you say above that we should keep it this way because 
> There is no guarantee that the CLI will be running in the same place as the server

But here you say: 
> If code is running on the server, it seems like it should be able to call those functions directly instead of making an API call.

src/fides/ctl/core/api.py
pattisdr: > Directly mixing in test logic with code logic feels suboptimal to me

Agreed, this was a last resort given what I understood to be the constraints! In particular, it sounds like the CLI needs to make API requests


> I found the following SO thread that has some potentially useful solutions:
https://stackoverflow.com/questions/9559963/unit-testing-a-python-app-that-uses-the-requests-library

hmm, I'm hesitant to mock the response from the `requests` calls here, because then we'd be mocking most of the logic used by the CLI


> I might also be totally off-base here and not grasping that potentially the TestClient can't be replaced by another library

The core of the issue is that I want to unit test the calls in `src/fides/ctl/core/api.py` which are primarily used by the CLI (`test_cli`, `test_api`, `test_api_helpers`).  You said above the CLI should always be making [API requests against the webserver ](https://github.com/ethyca/fides/pull/1731#discussion_r1019989425) because there's no guarantee that it is running in the same place as the server.

The `TestClient` doesn't actually make http requests, it simulates making http requests and calls functions directly.  This is great for testing our routes: you don't need a live webserver, you reduce overhead of going through an entire http stack, etc.  

However, on this page, I'm trying to unittest locations which make http requests to our webserver.   I can't inject the test client here, it's not an endpoint per se, and I can't create a client in this context without getting multiple event loops running.  The only workable option I've seen so far is to pass in the TestClient for use in pytest, otherwise, it makes http calls. Open to other ideas!.

.fides/fides.toml
```diff
@@ -7,7 +7,7 @@ db = "fides"
 test_db = "fides_test"
 
 [credentials]
-app_postgres = {connection_string="postgresql+psycopg2://postgres:fides@fides-db:5432/fides_test"}
+app_postgres = {connection_string="postgresql+psycopg2://postgres:fides@fides-db:5432/fides"}
```
pattisdr, line 10: Running `fides scan dataset db --credentials-id app_postgres` locally should now scan the application db not the test db

docker-compose.yml
```diff
@@ -112,7 +111,6 @@ services:
     restart: always
     environment:
       FIDES__CONFIG_PATH: ${FIDES__CONFIG_PATH:-/fides/.fides/fides.toml}
-      FIDES__TEST_MODE: "True"
```
pattisdr, line 115: Removing setting `FIDES__TEST_MODE` in the `docker-compose.yml`.  In the spirit of how fidesops worked, this variable will only be set by pytest.

pyproject.toml
```diff
@@ -149,7 +149,7 @@ extension-pkg-whitelist = ["pydantic", "zlib"]
 ############
 [tool.pytest.ini_options]
 env = [
-  "TESTING=True"
+  "FIDES__TEST_MODE=true"
```
pattisdr, line 152: This is where `FIDES__TEST_MODE` is set to true for pytest only

src/fides/api/ctl/database/crud.py
```diff
@@ -12,23 +12,22 @@
...
 
-from fides.api.ctl.database.session import async_session
 from fides.api.ctl.utils import errors
 
 
 # CRUD Functions
 async def create_resource(
-    sql_model: Base,
-    resource_dict: Dict,
+    sql_model: Base, resource_dict: Dict, async_session: AsyncSession
```
pattisdr, line 23: Adjusted these crud methods to require an `async_session` as an argument which will be test-mode aware, and talk to the correct db in the correct context.

src/fides/api/ctl/database/session.py
```diff
@@ -1,20 +1,33 @@
...
-)
-async_session = sessionmaker(engine, class_=AsyncSession, expire_on_commit=False)
-
-sync_engine = create_engine(config.database.sync_database_uri, echo=False)
-sync_session = sessionmaker(
-    sync_engine,
-    class_=Session,
-    expire_on_commit=False,
-    autocommit=False,
-)
```
pattisdr, line 20: Ops-side already has this concept, so removing here in favor of using the ops one.

src/fides/api/ctl/routes/crud.py
```diff
@@ -60,8 +62,9 @@
         },
     )
     async def create(
-        resource: resource_model,
+        resource: fides_model,
         resource_type: str = get_resource_type(router),
+        db: AsyncSession = Depends(get_async_db),
```
pattisdr, line 67: `get_async_db` is a dependency here that creates a new AsyncSession that is used in this request, and closed when finished. https://fastapi.tiangolo.com/tutorial/sql-databases/#create-a-dependency

This pattern is used frequently on the ops side.

tests/ctl/core/test_system.py
ThomasLaPiana: @pattisdr sure! We can add it to our 1:1 agenda?

To try to sum it up in the mean time:
* The CLI should always use API calls to interact with any server logic
* The server should always call internal functions directly, i.e. if the server needs to reset the database it shouldn't call the `admin` endpoint, it should directly call the `reset_database` function

At least that's how I see it in my head....not saying it's the only way! Just feels like it has the right level of separation and maintainability

pattisdr: OK great let's talk in our 1:1.  This ^ is helpful, it's not a problem to refactor things like the reset calls, but those aren't the core testing problem here, it's the CLI specifically.

src/fides/ctl/core/api.py
pattisdr: Discussed in person - Thomas suggested monkeypatching https://docs.pytest.org/en/7.1.x/how-to/monkeypatch.html so I'll investigate that next!

tests/ctl/cli/test_cli.py
```diff
@@ -28,18 +29,24 @@ def test_cli_runner() -> Generator:
...
 
 @pytest.mark.integration
-def test_init(test_cli_runner: CliRunner) -> None:
+def test_init(test_cli_runner: CliRunner, test_client: TestClient) -> None:
     result = test_cli_runner.invoke(
-        cli, ["init"], env={"FIDES__USER__ANALYTICS_OPT_OUT": "true"}
+        cli,
+        ["init"],
+        env={"FIDES__USER__ANALYTICS_OPT_OUT": "true"},
+        obj={"client": test_client},
```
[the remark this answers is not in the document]
pattisdr: Reverted

src/fides/api/ops/api/deps.py
```diff
@@ -18,22 +18,21 @@ def get_config() -> FidesConfig:
...
+def get_sync_session() -> Session:
     """Gets the shared database session to use for API functionality"""
     global _engine  # pylint: disable=W0603
     if not _engine:
         _engine = get_db_engine(config=CONFIG)
-    SessionLocal = get_db_session(CONFIG, engine=_engine)
-    db = SessionLocal()
-    return db
+    session_maker = get_db_session(CONFIG, engine=_engine)
+    return session_maker()
```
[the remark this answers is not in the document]
pattisdr: Reverted to reduce the diff size

src/fides/api/ctl/database/session.py
pattisdr: Reverted to reduce the diff size

tests/ctl/conftest.py
```diff
@@ -8,33 +8,56 @@
...
+    mpatch.undo()
+
+
+@pytest.fixture(autouse=True, scope="session")
+def monkeypatch_requests(test_client, monkeysession) -> None:
+    monkeysession.setattr(requests, "get", test_client.get)
+    monkeysession.setattr(requests, "post", test_client.post)
+    monkeysession.setattr(requests, "put", test_client.put)
+    monkeysession.setattr(requests, "patch", test_client.patch)
+    monkeysession.setattr(requests, "delete", test_client.delete)
```
pattisdr, line 60: @ThomasLaPiana your monkeypatching suggestion worked very well! My main lingering concern would be that I don't want to be too aggressive with this monkeypatching - so I'm just doing this on the ctl-side. The ops side uses the `requests` library a lot to talk to connectors.

pattisdr:
Issues still existing from the merge ^ will address

ThomasLaPiana:
@pattisdr amazing work here, seriously!

It's looking like this is ready for a final review?

pattisdr:
hello @ThomasLaPiana yes this is ready for review, I wanted to get your take on this, since most of the changes are ctl-related!

src/fides/api/ctl/routes/crud.py
ThomasLaPiana: I believe this is a much better pattern, thank you for bringing this over!

src/fides/cli/utils.py
```diff
@@ -269,7 +273,7 @@ def send_init_analytics(opt_out: bool, config_path: str, executed_at: datetime)
     try:
         client = AnalyticsClient(
             client_id=analytics_id or generate_client_id(FIDESCTL_CLI),
-            developer_mode=bool(getenv("FIDES_TEST_MODE") == "True"),
+            developer_mode=get_dev_mode(),
```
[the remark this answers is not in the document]
ThomasLaPiana: Agreed this is more logical!

tests/ctl/conftest.py
ThomasLaPiana: @pattisdr that is correct, it monkeypatches _everywhere_ (due to how Python works under the hood I guess) so you then need to explicitly un-patch it, but this model works fine for our current use!

ThomasLaPiana (approved):
Tested and confirmed working!

Despite being a relatively large PR, it was a breeze to review thanks to your comments and documentation, thank you!

This also hugely improves the codebase on the ctl side 🙏

pattisdr:
Thanks so much for your review @ThomasLaPiana, we'll wait to merge this until fidesplus 2.0 is out

pattisdr:
Latest commit https://github.com/ethyca/fides/pull/1731/commits/f5ddef8db8aae5e2d20e50b01adfcc2c9c222fa6 needed to get up to date with recent changes in https://github.com/ethyca/fides/pull/1860 and https://github.com/ethyca/fides/pull/1861.  These tests were failing because they were talking to the application db instead of the test db. We're now populating the test db for tests.

pattisdr:
^ I still have one test hanging for the FidesConnector work: tests/ops/service/connectors/test_fides_connector.py::TestFidesConnectorIntegration::test_retrieve_data

The FidesConnector/FidesClient use their own requests.Sessions and httpx.AsyncClient()s to make requests which make requests against the running webserver.  I'm trying to monkeypatch similar to what we did on the -ctl side to use various TestClients instead.  This test is currently creating/queueing a privacy request but the celery task is staying in pending and the privacy request status is not getting updated.

ThomasLaPiana:
> ^ I still have one test hanging for the FidesConnector work: tests/ops/service/connectors/test_fides_connector.py::TestFidesConnectorIntegration::test_retrieve_data after merging in main
> 
> The FidesConnector/FidesClient use their own requests.Sessions and httpx.AsyncClient()s to make requests which make requests against the running webserver. I'm trying to monkeypatch similar to what we did on the -ctl side to use TestClients instead. This test is currently creating/queueing a privacy request but the privacy request status is not getting updated.
> 
> EDIT Going back into draft mode

Sounds good, lmk if I can help!

src/fides/api/ops/service/privacy_request/request_service.py
```diff
@@ -75,13 +75,18 @@ def cache_data(
...
+
+
 async def poll_server_for_completion(
     privacy_request_id: str,
     server_url: str,
     token: str,
     *,
     poll_interval_seconds: int = 30,
-    timeout_seconds: int = 1800,  # 30 minutes
+    timeout_seconds: int = 60,  # 30 minutes
```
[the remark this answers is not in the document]
pattisdr: Don't worry I'll revert this! it's a WIP commit https://github.com/ethyca/fides/pull/1731/commits/a4123b2a7f01ec8fa4913a1a1ecef9d3d92de7a0 that I just pushed to show what I'm talking about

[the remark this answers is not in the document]
adamsachs: great, thanks for clarifying @pattisdr! figured it hadn't slipped by you, but just wanted to be extra sure :)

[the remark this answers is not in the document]
pattisdr: always happy to have extra eyes, thanks Adam

pattisdr:
OK I think we're ready to go again!

pattisdr:
@ThomasLaPiana this PR has sat for awhile, as I deprioritized to work on unified-fides-resources, but with this last commit https://github.com/ethyca/fides/pull/1731/commits/834aae16bc214b89ac76e2a0f5f656ca5d24307d, I think it's good to go.

ThomasLaPiana:
@pattisdr you're a hero, thank you! I've been anxiously awaiting! I'll re-review/retest 🙂

pattisdr:
hooray!  I think this functionality https://github.com/ethyca/fides/commit/0882943768fcdfc7890bd7aa607327e25380acd8 that Adam merged while this was sitting might have helped tests pass too. Previously in trying to monkeypatch async clients in the fides connector test I had been running into errors with multiple event loops.

src/fides/api/ops/service/connectors/fides/fides_client.py
```diff
@@ -145,7 +146,11 @@ def create_privacy_request(
 
     @sync
     async def poll_for_request_completion(
-        self, privacy_request_id: str, timeout: int, interval: int
+        self,
+        privacy_request_id: str,
+        timeout: int,
+        interval: int,
+        async_client: AsyncClient | None = None,
```
ThomasLaPiana, line 153: Will this work in older versions of Python?

tests/ctl/conftest.py
```diff
@@ -35,6 +37,31 @@
...
+@pytest.fixture(scope="session")
+def monkeysession():
+    """monkeypatch fixture at the session level instead of the function level"""
+    mpatch = MonkeyPatch()
+    yield mpatch
+    mpatch.undo()
+
+
+@pytest.fixture(autouse=True, scope="session")
+def monkeypatch_requests(test_client, monkeysession) -> None:
```
ThomasLaPiana, line 59: A quick doc string here would be good for future clarify

tests/ctl/conftest.py
```diff
@@ -8,13 +8,15 @@
...
 import os
 from datetime import datetime
-from typing import Any, AsyncGenerator, Dict, Generator, Union
-from uuid import uuid4
+from typing import Dict, Generator, Union
 
 import pytest
+import requests
 import yaml
+from _pytest.monkeypatch import MonkeyPatch
```
ThomasLaPiana, line 16: what is the underscore for?

ThomasLaPiana (approved):
Fantastic work here! Excited to see this merged and making development easier 🙂

src/fides/api/ops/service/connectors/fides/fides_client.py
sanders41: It works on 3.7+ with `from __future__ import annotations`

tests/ctl/conftest.py
pattisdr: Great idea

src/fides/api/ops/service/connectors/fides/fides_client.py
pattisdr: While we've largely been using the older format, folks have been starting to use this newer format in places for awhile now too.

pattisdr:
Appreciate the re-review @ThomasLaPiana!

PR #1737, opened 2022-11-10
230/Login defaults

LKCSmith:
Closes [this issue](https://github.com/ethyca/fidesctl-plus/issues/230).

Config wizard flag off:

https://user-images.githubusercontent.com/99051851/201433118-e701b740-c438-4e15-b6a0-29a48e562b38.mov

Config wizard flag on:

https://user-images.githubusercontent.com/99051851/201433142-e4feb6d4-ac22-4681-8076-4d72200eb8e5.mov

To do:
* Tests
* Text updates

### Code Changes

* [ ] On checking token, redirect to the correct login route based on plus/non-plus status and if systems already exist or not

### Steps to Confirm

* [ ] Log-in as root user with and without systems and
* [ ] With and without config wizard flag on and
* [ ] Plus and non plus

### Pre-Merge Checklist

* Documentation Updated:
* [ ] Update `CHANGELOG.md`

### Description Of Changes


clients/admin-ui/src/pages/login.tsx
```diff
@@ -71,7 +82,26 @@ const useLogin = () => {
   };
 
   if (token) {
-    router.push(INDEX_ROUTE);
+    const configWizardFlagIsActive = Flags.some(
```
LKCSmith, line 85: Since we're already doing the routing on login here, this seemed like the most logical place to me to put this logic.

clients/admin-ui/src/features/config-wizard/setup.tsx
```diff
@@ -36,7 +36,7 @@ const Setup = () => {
           width="50%"
         >
           <Heading as="h3" size="lg">
-            Get Started
+            Generate your DataMap
```
chriscalhoun1974, line 39: Should this be Data Map?

clients/admin-ui/src/pages/login.tsx
chriscalhoun1974: Should this redirection be encapsulated in a custom hook?

clients/admin-ui/src/constants.ts
```diff
@@ -119,6 +119,9 @@ export const LOGIN_ROUTE = "/login";
 export const USER_MANAGEMENT_ROUTE = "/user-management";
 export const CONNECTION_ROUTE = "/connection";
 export const CONNECTION_TYPE_ROUTE = "/connection_type";
+export const DATAMAP_ROUTE = "/datamap";
+export const SYSTEM_ROUTE = "/system";
+export const CONFIG_WIZARD_ROUTE = "/config-wizard";
```
ssangervasi, line 124: I think this block of constants is for API routes, `api/v1/{FOO_ROUTE}`, though I think I've seen them mixed around with the "UI ROUTES" below. @chriscalhoun1974 this is from the Ops codebase, right?

chriscalhoun1974: @ssangervasi is correct in that this file is used within defining API routes in addition to root UI routes. It was a carryover from FidesOps.

clients/admin-ui/src/pages/login.tsx
```diff
@@ -71,7 +82,26 @@ const useLogin = () => {
...
+
+        return features.plus
+          ? router.push(datamapRoute.href)
+          : router.push(SYSTEM_ROUTE);
+      } 
+        router.push(CONFIG_WIZARD_ROUTE);
+      
+    } else {
+      router.push(INDEX_ROUTE);
+    }
```
ssangervasi, line 104: The logic of this block has gotten pretty complex, and the way that it "returns" is by calling `router.push` which can have unintended consequences. I think this could benefit from a helper function that returns the route that should come next (if any), and then `router.push` is only called in one place.

```js
const getRedirectRoute = (systems, ...etc) => {
  if (!token) {
    return undefined
  }
  const configWizardFlagIsActive = ...;
  if (configWizardFlagIsActive) {
    ...
  }
  return INDEX_ROUTE
}

// In useLogin
const redirectRoute = getRedirectRoute(...)
if (redirectRoute) {
  router.redirect(redirectRoute)
}
// maybe other code doesn't need to run if a redirect is going to occur...
```

clients/admin-ui/src/pages/login.tsx
```diff
@@ -71,7 +82,26 @@ const useLogin = () => {
   };
 
   if (token) {
-    router.push(INDEX_ROUTE);
+    const configWizardFlagIsActive = Flags.some(
+      (flag) => flag.name === "configWizardFlag" && flag.isActive
+    );
+
+    if (configWizardFlagIsActive) {
+      if (systems && systems.length > 0) {
```
ssangervasi, line 90: This will be false if the systems query is still loading, which I think means the page will render. The loading state should take this query into account.

clients/admin-ui/src/features/config-wizard/setup.tsx
LKCSmith: The text was copied directly from the issue

clients/admin-ui/src/pages/login.tsx
LKCSmith: From what I can tell, in NextJS, router.push is how we've been redirecting routes in the code. 👀

clients/admin-ui/src/pages/login.tsx
```diff
@@ -70,8 +81,33 @@ const useLogin = () => {
...
+    if (!token) {
+      return undefined;
+    }
+
+    const configWizardFlagIsActive = Flags.some(
+      (flag) => flag.name === "configWizardFlag" && flag.isActive
+    );
+
+    if (configWizardFlagIsActive) {
+      if (!isLoading && systems && systems.length > 0) {
```
LKCSmith, line 94: Is this enough to count for the loading state?

clients/admin-ui/src/features/config-wizard/setup.tsx
ssangervasi: The figma has a space though, I think it was a copy probly.

ssangervasi (approved):

chriscalhoun1974 (approved):

Issue #1744, opened 2022-11-11
Associate scan history with classification instances


allisonking:
### Is your feature request related to a specific problem?

We currently have no way of associating a classification instance with a specific scan instance. This makes it difficult to show the classification results after a scan because we don't know if the results are from the scan that just ran, or some long ago one. We would end up with potentially confusing results.

### Describe the solution you'd like

Perhaps in our classification table, we add a foreign key relationship to a scan history ID. Then, the frontend can query for all instances from the latest scan (instead of using fides keys as a workaround)

### Describe alternatives you've considered, if any

_A description of any alternative solutions or features you've considered._

### Additional context

_Add any other context or screenshots about the feature request here._

PR #1752, opened 2022-11-11
Integrate backend with classify UI

allisonking:
Closes https://github.com/ethyca/fides/issues/1784 (integration portion of https://github.com/ethyca/fides/pull/1720).

https://github.com/ethyca/fidesctl-plus/pull/304 added query params on GET `/plus/classify`: `resource_type` and `fides_keys`. It also converted the GET `/plus/scan` endpoint to a PUT, so https://github.com/ethyca/fides/pull/1752/commits/957ea54bf557858133b229538e3d2dbb90c3789c handles that conversion

### Code Changes

* [x] Update typescript types after https://github.com/ethyca/fidesctl-plus/pull/304. At first I thought we would keep plus models separate, but I think those actually already exist in the repo? So I merged them, but it was kind of a pain, as I think there are some types from maybe ops so couldn't just run the script and then commit everything
* [x] Update `plus.slice.ts` to make sure we can take the new query params. Updates selectors too, hopefully I did it right 🤞 
* [x] Only show the systems that are in review in the classify screen. There's a tricky bit here, which is captured in https://github.com/ethyca/fides/issues/1744
* [x] Convert the GET `/plus/scan` to a PUT. Because of the way the onboarding flow works/is structured, this ends up being handled in a `useEffect` which might be weird... open to other thoughts!
* [x] Cypress tests

### Steps to Confirm

Below steps are for running everything locally. You'll have to configure system scanning though, which takes some coordination. The Cypress tests should also cover most of this though
* [ ] Run fidesctl+ via `make server`
* [ ] Disable protected routes (comment it out in _app.tsx) since we [can't log into fidesctl+ yet](https://ethyca.slack.com/archives/CBKK8TS74/p1667926862630059) 🥲 
* [ ] Enable the config wizard flag
* [ ] Go through the config wizard and trigger a runtime scan
* [ ] Only register a few of the systems
* [ ] You'll be brought to the classify screen and each system will probably say "Processing". Refresh the page until it's done (~2-3 minutes) and it'll say "Awaiting review"

### Pre-Merge Checklist

* Documentation Updated:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

image


clients/admin-ui/src/features/common/plus.slice.ts
```diff
@@ -61,9 +68,19 @@ export const plusApi = createApi({
...
-      void
+      ClassifyInstancesParams
     >({
-      query: () => `classify/`,
+      query: (params) => {
+        const urlParams = new URLSearchParams();
+        urlParams.append("resource_type", params.resource_type);
+        params.fides_keys?.forEach((key) => {
+          urlParams.append("fides_keys", key);
+        });
```
allisonking, line 78: I couldn't find an easier way to do this but I'm all ears...

clients/admin-ui/src/features/dataset/DatabaseConnectForm.tsx
```diff
@@ -139,7 +133,7 @@ const DatabaseConnectForm = () => {
         error: string;
       }
     | {
-        classifyInstances: ClassifyInstanceResponseValues[];
+        classifyInstances: Record<string, string>[];
```
allisonking, line 136: This started erroring because of the change in type of `ClassifyDatasetResponse`. I tested it locally, and it turns out this does not return a `ClassifyInstanceResponseValues`. it returns a list that looks like 
```
[
  {
    dataset_name:  "bestpizzaco"
    id : "cls_f555ae90-25a7-4cdb-9d03-8bf644e6375f"
    status: "Created"
  },
...
]
```

I guess that's not explicitly typed in the backend though, so we just have `Record<string, string`. better than `any`, I guess?

clients/admin-ui/src/features/plus/plus.slice.ts
```diff
@@ -61,9 +67,19 @@ export const plusApi = createApi({
...
-      void
+      ClassifyInstancesParams
     >({
-      query: () => `classify/`,
+      query: (params) => {
+        const urlParams = new URLSearchParams();
+        urlParams.append("resource_type", params.resource_type);
+        params.fides_keys?.forEach((key) => {
+          urlParams.append("fides_keys", key);
+        });
```
allisonking, line 77: I couldn't find an easier way to do this but I'm all ears...

allisonking:
oh oof I just realized I need to invalidate tags... sorry for the fake out again, putting this back into draft

allisonking:
okay _now_ it should be ready to review 🤞

clients/admin-ui/src/features/dataset/DatabaseConnectForm.tsx
ssangervasi: Sounds like a bug ticket for the API then, even if the fix is just to give a type definition to the route.

For now, `Pick<ClassifyInstanceResponseValues, 'dataset_name', 'id', 'status'>` might work?

clients/admin-ui/cypress/e2e/config-wizard-plus.cy.ts
```diff
@@ -124,10 +113,20 @@ describe.skip("Config wizard with plus settings", () => {
...
       });
+
+      // Make sure there are only two systems in this table
+      cy.getByTestId("systems-classify-table");
+      cy.getByTestId("status-vzmgr-service");
+      cy.getByTestId("status-kube-dns");
+      cy.get("table")
+        .find("tr")
+        .then((rows) => {
+          expect(rows.length).to.eql(3); // +1 for the header row
```
ssangervasi, line 124: `find("tbody > tr")` could spare us the math.

clients/admin-ui/src/features/plus/plus.slice.ts
ssangervasi: This is the canonical way to have an "array" in search/query params, and yes it strange. Looks like RTK supports a passing a simple object out of the box, but not a _fun_ case like this. https://redux-toolkit.js.org/rtk-query/api/fetchBaseQuery#paramsserializer

clients/admin-ui/src/features/plus/plus.slice.ts
```diff
@@ -98,27 +133,46 @@ export const useHasPlus = () => {
...
-    if (map.size === 0) {
-      return emptyClassifyInstanceMap;
-    }
-
-    return map;
+
+const instancesToMap = (instances: ClassifyInstanceResponseValues[]) => {
+  const map = new Map<string, ClassifyInstanceResponseValues>();
+  instances.forEach((ci) => {
+    map.set(ci.dataset_key, ci);
```
ssangervasi, line 158: This code uses the `dataset_key` key, so it _shouldn't_ work for the systems map which [looks up](https://github.com/ethyca/fides/blob/0da90bcb8efba4ac10cf181b757481bbd5c816a3/clients/admin-ui/src/features/system/ClassifySystemsTable.tsx#L24) by `system.fides_key`. I hope this API isn't putting the system key on the dataset key field 🤞😑🤞

clients/admin-ui/src/pages/classify-systems/index.tsx
```diff
@@ -1,26 +1,55 @@
...
-  const { isLoading: isLoadingSystems, data: systems } = useGetAllSystemsQuery(
-    undefined,
-    { skip: !hasPlus }
-  );
+  /**
+   * TODO: fides#1744
+   * Because there is currently no way to associate a scan with its classification,
+   * we attempt to get the classification results for only the systems in review.
+   * However, because this is a page that is waiting for results and so will likely
+   * be refreshed, we can't necessarily rely on knowing the systems we put into review
```
ssangervasi, line 30: Maybe "we can't necessarily rely on knowing _which_ systems were put into review."

Though I think we actually don't have to worry about this case because the redux state is persisted to local storage. The config wizard state is only cleared with an [explicit cancel](https://github.com/ethyca/fides/blob/ed6bc3b59ea1c5af482db03b3cf2c4b9c64e0d5a/clients/admin-ui/src/features/config-wizard/ConfigWizardWalkthrough.tsx#L38). If this case came up, something else is going wrong.

clients/admin-ui/src/types/api/models/ClassificationStatus.ts
```diff
@@ -3,7 +3,14 @@
 /* eslint-disable */
 
 /**
- * The model housing statuses for a classify instance
+ * The model housing statuses for a classify instance.
+ *
+ * CREATED: The initial state of an attempt to classify
+ * PROCESSING: Attempting to classify dataset(s)
+ * COMPLETED: The classify process completed successfully
```
ssangervasi, line 10: Nit: guessing this came from autogeneration but "COMPLETED" ≠ "COMPLETE" LOL

clients/admin-ui/cypress/e2e/config-wizard-plus.cy.ts
allisonking: oh nice tip, I'm all about sparing myself math

clients/admin-ui/src/features/plus/plus.slice.ts
allisonking: it is... 😬 @SteveDMurphy and I talked about changing the key to `fides_key` instead but I don't think we ever made an issue for it. mostly didn't want to break the existing dataset classification (the original API was written in a way that never expected to do anything other than datasets, but here we are 😄 ) 

I just made a ticket https://github.com/ethyca/fidesctl-plus/issues/326

clients/admin-ui/src/pages/classify-systems/index.tsx
allisonking: mm that's true. I was worried about somebody going directly to the `/classify-systems` URL without having gone through the config wizard first. but maybe in that case we should just render an empty state since scan wouldn't have happened yet.

clients/admin-ui/src/types/api/models/ClassificationStatus.ts
allisonking: fixed here, the source comment (on the backend in fidesctl+) will be done separately

ssangervasi (approved):

clients/admin-ui/src/features/dataset/DatabaseConnectForm.tsx
allisonking: https://github.com/ethyca/fidesctl-plus/issues/331

allisonking:
confirmed that the `fides-deploy` check passes locally, so I'm gonna merge this

PR #1754, opened 2022-11-11
treat  response from adobe jwt appropriately, as milliseconds

### LINKED ISSUE #1753: adobe auth strategy treats ms expiration duration as seconds
### Bug Description

in our adobe auth override, we're mistakenly treating the `expires_in` value we get from the adobe jwt exchange API response as if it were a seconds duration. in reality, it's a milliseconds duration, so we should treat it accordingly.

#### Steps to Reproduce

1. run a DSR that uses the adobe connector. it will retrieve and store a temporary auth key in our app db.
2. with same auth config (i.e. don't reset your db), run another DSR that uses the same adobe connector, more than 24 hours after step 1. 
3. you should hit an auth error (401) coming from adobe in your DSR execution. the token has expired, but our app will think it's still valid (since it thought it was expiring in 999 days, not .99 😄 )

### Expected behavior

we should be refreshing and getting a new token when the existing token actually expires, i.e. what seems to be .99 days after it was created. in any case, it's expressed in ms, so as long as we handle that properly, we will be refreshing when we need to!

### Screenshots

### Environment

* Version:
* OS:
* Python Version:
* Docker Version:

### Additional context

adamsachs:
Closes #1753 

### Code Changes

correct treatment of `expires_in` in adobe JWT response as milliseconds value, not seconds.

### Steps to Confirm

### Pre-Merge Checklist

* Documentation Updated:
* [x] Update `CHANGELOG.md`

### Description Of Changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@ The types of changes are:
 * Show a helpful error message if Docker daemon is not running during "fides deploy" [#1694](https://github.com/ethyca/fides/pull/1694)
 * Allow users to query their own permissions, including root user. [#1698](https://github.com/ethyca/fides/pull/1698)
 * Single-select taxonomy fields legal basis and special category can be cleared. [#1712](https://github.com/ethyca/fides/pull/1712)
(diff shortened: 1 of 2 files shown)


galvana (approved):
Nice fix!

PR #1786, opened 2022-11-15
Generated dataset fides_keys are unique

SteveDMurphy:
Closes [fidesplus-licensing#62](https://github.com/ethyca/fidesplus-licensing/issues/62)

### Code Changes

* [x] Added function to create repeatable hash
* [x] Update generated datasets to have a repeatable uuid as suffix of fides_key. Also include hash components in `meta`

### Steps to Confirm

* [x] Generated datasets both through the API and CLI successfully
* [x] Validate the args passed to classify are updated to reflect new fides_key values as well

### Pre-Merge Checklist

* Documentation Updated:
  * [x] documentation complete, or draft/outline provided (tag docs-team to complete/review on this branch)
  * [x] documentation issue created (tag docs-team to complete issue separately)
* [x] Update `CHANGELOG.md`

### Description Of Changes

Currently this is geared heavily towards datasets but could be more agnostic to allow for uniqueness in other resources (i.e. Systems) as well

### DIFF
--- a/src/fides/ctl/core/dataset.py
+++ b/src/fides/ctl/core/dataset.py
@@ -123,20 +123,19 @@ def create_db_dataset(schema_name: str, db_tables: Dict[str, List[str]]) -> Data
     return dataset
 
 
-def make_dataset_fides_key_unique(
-    datasets: List[Dataset], database_host: str, database_name: str
-) -> List[Dataset]:
+def make_dataset_key_unique(
+    dataset: Dataset, database_host: str, database_name: str
+) -> Dataset:
     """
-    Ensure each generated dataset has a repeatable unique ID appended to the end
+    Ensure the dataset has a repeatable unique ID appended to the end
     to avoid naming collisions.
     """
 
-    for dataset in datasets:
-        dataset.fides_key = generate_unique_fides_key(
-            dataset.fides_key, database_host, database_name
-        )
-        dataset.meta = {"database_host": database_host, "database_name": database_name}
-    return datasets
+    dataset.fides_key = generate_unique_fides_key(
+        dataset.fides_key, database_host, database_name
+    )
+    dataset.meta = {"database_host": database_host, "database_name": database_name}
+    return dataset
 
 
 def find_uncategorized_dataset_fields(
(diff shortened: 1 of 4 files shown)


SteveDMurphy:
Wanted to bump this if you have time @ThomasLaPiana, also feel free to reassign!

ThomasLaPiana:
sorry @SteveDMurphy , getting to it now!! This is a great change 🙂

src/fides/ctl/core/utils.py
```diff
@@ -136,6 +137,19 @@ def sanitize_fides_key(proposed_fides_key: str) -> str:
     return sanitized_fides_key
 
 
+def generate_unique_fides_key(
+    proposed_fides_key: str, database_host: str, database_name: str
+) -> str:
+    """
+    Uses host and name to generates a UUID to be
```
ThomasLaPiana, line 144:
```suggestion
    Uses host and name to generate a UUID to be
```

tests/ctl/core/test_utils.py
```diff
@@ -113,3 +113,12 @@ def test_new_file_is_dirty(self) -> None:
             file.write("test file")
         assert utils.git_is_dirty()
         os.remove(test_file)
+
+
+@pytest.mark.unit
+def test_repeatable_unique_key() -> None:
```
ThomasLaPiana, line 119: are there any edge cases you can think of that might break this functionality?

The functionality is straight-forward and implemented in a transparent way, but still wondering if there might be anything else to consider for testing here

src/fides/ctl/core/dataset.py
```diff
@@ -319,4 +344,7 @@ def generate_bigquery_datasets(bigquery_config: BigQueryConfig) -> List[Dataset]
     bigquery_engine = get_bigquery_engine(bigquery_config)
     bigquery_schemas = get_db_schemas(engine=bigquery_engine)
     bigquery_datasets = create_db_datasets(db_schemas=bigquery_schemas)
-    return bigquery_datasets
+    unique_bigquery_datasets = make_dataset_fides_key_unique(
```
ThomasLaPiana, line 347: This is a design comment/nit, but generally I think it is better to loop _outside_ of the lower-level functions and instead do it in the main codepath.

As an example, this is how I would rewrite this line

`unique_bigquery_datasets = [make_dataset_key_unique(dataset) for dataset in bigquery_datasets]`

The above function would then be updated to only take a single dataset instead of a list of them.

Does that make sense/do you agree with that opinion?

ThomasLaPiana (approved):
Approving, but left some nits/food for thought :) 
Great fix here, very valuable feature!

src/fides/ctl/core/dataset.py
SteveDMurphy: Thank you for taking the time to lay that out! I am definitely ok with that. I think it is sort of an existing pattern currently elsewhere but am happy to start making an effort to standardize around this 👍🏽

tests/ctl/core/test_utils.py
SteveDMurphy: I was thinking about it and the problem would really be if this were used with something other than datasets (or anything that doesn't have a host/name for us to use (which is out of scope for this change) and I couldn't really imagine how to test that. I liked this because it achieved the idea of the uuid being appended as well as being repeatable so two tests in one almost 👍🏽

Issue #1760, opened 2022-11-13
Refactor the `config` loading flow


ThomasLaPiana:
### Is your feature request related to a specific problem?

Right now the flow of checking for/writing out/loading/injecting variables into the config has gotten quite complex and extremely difficult (for me at least, and I wrote most of it) to follow. 

It is hard to know what gets loaded where, when, and finding/squashing bugs in the config flow is becoming increasingly difficult

### Describe the solution you'd like

This will be a multi-part fix, and should probably include the following:
* A refactor of the `fides.ctl.core.config` module itself, simplifying those functions, etc.
* no longer having `CONFIG` available anywhere as a constant. Functions should call `get_config` as needed, not instantly call it when the module gets imported. This feels like a suboptimal design for sure
* There is currently a lot of `config`-related code in the `fides.cli` module, this needs to get refactored out into the `fides.ctl.core.config` module instead
* As part of the above, the logic/order of operations for loading the config need to be greatly clarified/simplified within the main code path
* Better documentation/patterns around required configuration variables 

### Additional context

This is intended to be its own small epic, not done in a single PR

PR #1788, opened 2022-11-15
Add helper function to debug required config vars

### LINKED ISSUE #1787: Improve UX for missing required config variables

Currently, if there is a required config var that is missing, the app crashes _hard_

There should be a friendly error message to the user that there are certain variables that are required to run. This might end up being a bigger change, or might require some finesse about catching `MissingConfig` exceptions and making them more pretty

### Additional context

This is a follow up from #1718, as that fixed the bug but another UX pass is needed here

TheAndrewJackson:
Closes #1787 

### Code Changes

* [ ] Add helper function to `fides.ctl.core.config.utils` that checks if all the required config variables are configured

### Steps to Confirm

* [ ] Run `nox -s "build(dev)" && docker run -it ethyca/fides:local` and it should display the message and exit with error code 1

### Pre-Merge Checklist

* Documentation Updated:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

In the current draft state this PR serves as a starting point to get a conversation going. This solution is harder to maintain for developers but _hopefully_ offers a great user experience by immediately displaying all of the missing required configs vars. The output should look like this when no config is provided.

Screen Shot 2022-11-16 at 08 43 44

It works by running before the pydantic code tries to load everything in. It circumvents catching `MissingConfig` exceptions by manually checking for config vars. It also makes it very easy to display helpful error messages.

I'm open to a more automated solution but I want to minimize how much work is done for this increment when fides config is about to receive a big overhaul (#1760). I think this adds a good amount of value while being simple to implement.

### DIFF
--- a/src/fides/ctl/core/config/utils.py
+++ b/src/fides/ctl/core/config/utils.py
@@ -92,8 +92,10 @@ def check_if_required_config_vars_are_configured() -> None:
         for missing_var in missing_required_config_vars:
             print(f"fides.toml: {missing_var[0]} or ENV VAR: {missing_var[1]}")
 
-        print("\nVisit the Fides deployment documentation for more information: "
-              "https://ethyca.github.io/fides/deployment/")
+        print(
+            "\nVisit the Fides deployment documentation for more information: "
+            "https://ethyca.github.io/fides/deployment/"
+        )
         sys.exit(1)
 
 
--- a/src/fides/ctl/core/config/utils.py
+++ b/src/fides/ctl/core/config/utils.py
@@ -52,13 +52,18 @@ def check_if_required_config_vars_are_configured() -> None:
         "FIDES__SECURITY__OAUTH_ROOT_CLIENT_SECRET"
     )
     try:
-        app_encryption_key = get_config_from_file("", "security", "app_encryption_key")
-        oauth_root_client_id = get_config_from_file(
-            "", "security", "oauth_root_client_id"
-        )
-        oauth_root_client_secret = get_config_from_file(
-            "", "security", "oauth_root_client_secret"
-        )
+        if not app_encryption_key:
+            app_encryption_key = get_config_from_file(
+                "", "security", "app_encryption_key"
+            )
+        if not oauth_root_client_id:
+            oauth_root_client_id = get_config_from_file(
+                "", "security", "oauth_root_client_id"
+            )
+        if not oauth_root_client_secret:
+            oauth_root_client_secret = get_config_from_file(
+                "", "security", "oauth_root_client_secret"
+            )
     except FileNotFoundError:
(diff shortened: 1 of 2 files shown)


src/fides/ctl/core/config/utils.py
```diff
@@ -40,6 +41,53 @@ def get_config_from_file(
...
+        "FIDES__SECURITY__OAUTH_ROOT_CLIENT_SECRET"
+    )
+    try:
+        app_encryption_key = get_config_from_file("", "security", "app_encryption_key")
+        oauth_root_client_id = get_config_from_file(
+            "", "security", "oauth_root_client_id"
+        )
+        oauth_root_client_secret = get_config_from_file(
+            "", "security", "oauth_root_client_secret"
+        )
```
sanders41, line 61: These will run any time there is a config file and could lead to false positives since some values could be set in the environment variables and others in the toml file. Environment variables are supposed to take precedence here so you could only check the toml file if the variable is `None` after checking the env var.

```py
try:
    if not app_encryption_key:
        app_encryption_key = get_config_from_file("", "security", "app_encryption_key")
    ...
```

TheAndrewJackson: Good catch. Updated!

sanders41:
The changes look good. Is the plan for this to stay long term, or to be a stop gap until the overhaul is done? If the plan is for long term I would also add some tests to make sure future changes don't break things.

src/fides/ctl/core/config/utils.py
```diff
@@ -40,6 +41,58 @@ def get_config_from_file(
...
+        missing_required_config_vars.append(
+            (
+                "security.oauth_root_client_secret",
+                "FIDES__SECURITY__OAUTH_ROOT_CLIENT_SECRET",
+            )
+        )
+
+    if len(missing_required_config_vars) > 0:
+        print(
+            "There are missing required config variables. Please add the following to start Fides: "
```
ThomasLaPiana, line 89:
```suggestion
            "There are missing required config variables. Please add the following config variables to either the `fides.toml` file or your environment variable to start Fides: "
```

ThomasLaPiana: also maybe a link to our docs page that describes configuration?

ThomasLaPiana:
@TheAndrewJackson I like this solution, gets the job done and is implemented without extra complexity

src/fides/ctl/core/config/utils.py
TheAndrewJackson: Good idea!

sanders41 (approved):

sanders41:
The failed test looks like it just timed out and didn't actually fail so I restarted it.

PR #1803, opened 2022-11-16
Fix EditDrawer styling and move save/cancel buttons to DrawerFooter

allisonking:
This has been bothering me for a while and I've finally leveled up enough to fix it 😅 

The EditDrawers always had weird styling that I was only able to get around by using `vh`. This caused some inconsistency though and made some elements cut off unexpectedly at times. 

In addition, it was annoying that the Cancel/Save buttons were not included in the `EditDrawer` wrapper. This was mostly because I thought the "submit" buttons had to be inside a `<form> `element. It turns out you can just [associate a submit button with a form via ID](https://stackoverflow.com/questions/52577141/how-to-submit-form-from-a-button-outside-that-component-in-react/53573760#53573760) though and it'll figure itself out regardless of hierarchy 🎉 

### Code Changes

* [x] Fix styling error by removing an unneeded Box component
* [x] Remove styling `vh` workarounds
* [x] Move cancel/edit buttons into EditDrawer

### Steps to Confirm

* [ ] Try editing a dataset, field, or collection. Things should work the same

### Pre-Merge Checklist

* Documentation Updated:
* [x] Update `CHANGELOG.md`

### Description Of Changes

### DIFF
--- a/clients/admin-ui/src/features/dataset/EditDrawer.tsx
+++ b/clients/admin-ui/src/features/dataset/EditDrawer.tsx
@@ -25,7 +25,8 @@ interface Props {
   deleteTitle: string;
   deleteMessage: ReactNode;
   children: ReactNode;
-  /* Associates the submit button with a form, which is useful for when the button
+  /**
+   * Associates the submit button with a form, which is useful for when the button
    * does not live directly inside the form hierarchy
    */
   formId?: string;
--- a/clients/admin-ui/src/features/dataset/EditDrawer.tsx
+++ b/clients/admin-ui/src/features/dataset/EditDrawer.tsx
@@ -4,6 +4,7 @@ import {
   Drawer,
   DrawerBody,
   DrawerContent,
+  DrawerFooter,
   DrawerHeader,
   DrawerOverlay,
   IconButton,
@@ -24,6 +25,10 @@ interface Props {
   deleteTitle: string;
   deleteMessage: ReactNode;
   children: ReactNode;
+  /* Associates the submit button with a form, which is useful for when the button
+   * does not live directly inside the form hierarchy
+   */
+  formId?: string;
 }
 
 const EditDrawer = ({
@@ -35,6 +40,7 @@ const EditDrawer = ({
   deleteTitle,
   deleteMessage,
   children,
+  formId,
 }: Props) => {
   const {
     isOpen: deleteIsOpen,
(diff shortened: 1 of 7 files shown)


clients/admin-ui/src/features/dataset/EditDrawer.tsx
```diff
@@ -24,6 +25,10 @@ interface Props {
   deleteTitle: string;
   deleteMessage: ReactNode;
   children: ReactNode;
+  /* Associates the submit button with a form, which is useful for when the button
+   * does not live directly inside the form hierarchy
+   */
+  formId?: string;
```
ssangervasi, line 31: Tip: block comments matching
```js
/**
 * ...
 */
```
will be picked up by JS documentation tools better, like VS code intellisense:
Screen Shot 2022-11-17 at 12 33 46

LKCSmith (approved):
👍

clients/admin-ui/src/features/dataset/EditDrawer.tsx
```diff
@@ -46,30 +52,42 @@ const EditDrawer = ({
...
+          <DrawerFooter justifyContent="flex-start">
+            <Button onClick={onClose} mr={2} size="sm" variant="outline">
+              Cancel
+            </Button>
+            <Button
+              type="submit"
+              colorScheme="primary"
+              size="sm"
+              data-testid="save-btn"
+              form={formId}
```
ssangervasi, line 86: Yep, form id comes in real handy here, and these changes are an improvement.

I think this also illustrates that EditDrawer is mixing two responsibilities: tidying up the layout markup we don't want to repeat, and managing the close/delete/submit state. It can be helpful to break a component like this into smaller pieces that are focused on the styling so that there are fewer layers to pass props through. Sketch of an example:

```jsx
export const EditCollectionDrawer = () => {
  // ...

  return (
    // maybe hoisting the <form> tag up here doesn't seem as crazy anymore
    <EditDrawer>
      <EditDrawerHeader>{/*...*/}</EditDrawerHeader>

      <EditDrawerBody>{/*...*/}</EditDrawerBody>

      <EditDrawerFooter>
        <CancelButton>
        <SaveButton /* or maybe form={id} could be passed directly ...*//>
      </EditDrawerFooter>
    </EditDrawer>
  );
};
```

ssangervasi (approved):
Workin 👍

clients/admin-ui/src/features/dataset/EditDrawer.tsx
allisonking: oh that's a great tip, thanks!

clients/admin-ui/src/features/dataset/EditDrawer.tsx
allisonking: oh I see what you mean! I'll leave it as-is for now, but will keep that separation in mind for future components (and maybe this one too, if things start getting messier)

PR #1811, opened 2022-11-17
Only check for required variables when the webserver is run

### LINKED ISSUE #1810: Cannot run any CLI commands without specifying mandatory security config settings
### Bug Description

Thanks to _several_ very useful improvements to loading our config settings, we're now throwing errors and showing helpful messages about the lack of some mandatory security settings:
* `FIDES__SECURITY__APP_ENCRYPTION_KEY`
* `FIDES__SECURITY__OAUTH_ROOT_CLIENT_ID`
* `FIDES__SECURITY__OAUTH_ROOT_CLIENT_SECRET`

These settings are critical for running the webserver (aka `fides webserver`) but not necessary for most CLI commands. However, we've made them necessary, which unfortunately breaks a vanilla, Python-based `fides` install and setup.

Where this is most obvious is the new `fides deploy` flow:
```
(venv) ~/fides-2-1-0-test% pip install ethyca-fides
...<snip>...
(venv) ~/fides-2-1-0-test% fides deploy up --no-pull
There are missing required config variables. Please add the following config variables to either the `fides.toml` file or your environment variables to start Fides:

fides.toml: security.app_encryption_key or ENV VAR: FIDES__SECURITY__APP_ENCRYPTION_KEY
fides.toml: security.oauth_root_client_id or ENV VAR: FIDES__SECURITY__OAUTH_ROOT_CLIENT_ID
fides.toml: security.oauth_root_client_secret or ENV VAR: FIDES__SECURITY__OAUTH_ROOT_CLIENT_SECRET

Visit the Fides deployment documentation for more information: https://ethyca.github.io/fides/deployment/
```
#### Steps to Reproduce

To reproduce this locally in the repo, you need to force yourself to install the `pip` package locally to a separate virtual env, because otherwise you'll incidentally load up our development `fides.toml` files or `FIDES__*` ENV vars.

So, on my machine, I've cloned the `fides` repo to `~/git/fides` and I can repro this as follows:

1. `nox -s "build(sample")`
2. `mkdir ~/fides-2-1-0-test && cd ~/fides-2-1-0-test`
3. `python3 -m venv venv && source venv/bin/activate`
4. `pip install -e ~/git/fides`
5. `fides deploy up --no-pull`

### Expected behavior

This should run `fides deploy up` successfully and load the sample project.

### Environment

* Version: 2.1.0
* OS: macOS-12.6.1-x86_64-i386-64bit
* Python Version: Python 3.9.6
* Docker Version: 20.10.17

ThomasLaPiana:
Closes #1810 

### Code Changes

* [x] make the "required" vars optional
* [x] add a check before the webserver starts
* [x] refactor the config var check a bit

### Steps to Confirm

* [ ] in a blank directory, run `fides webserver` and confirm you get the error
* [ ] run `nox -s "build(sample)"`, and then in a black directory `fides deploy up --no-pull --no-init` and confirm you get no errors
* [ ] in the `fides` project dir, running `fides webserver` works (but fails if no db is up)
* [ ] `docker run -it ethyca/fides:local /bin/bash`, set some env vars manually and verify the error (we want to make sure its loading env vars correctly)

### Pre-Merge Checklist

* Documentation Updated:
* [x] Update `CHANGELOG.md`

### Description Of Changes

Quick fix to make the CLI usable without the "required" webserver variables

Note that this will only fail only `fides webserver`, our compose setup that runs the app directly will not throw this error

Note that this also doesn't have tests, so let's be diligent in our manual review 🙂 

---

#### Error message when running `fides webserver` without the required variables:

```
test> fides webserver
Loading config from: .fides/fides.toml
Directory './.fides' already exists.
Configuration file already exists: ./.fides/fides.toml
To learn more about configuring fides, see:
        https://ethyca.github.io/fides/installation/configuration/
C:\Users\tlapiana\anaconda3\lib\site-packages\snowflake\connector\options.py:96: UserWarning: You have an incompatible version of 'pyarrow' installed (6.0.1), please install a version that adheres to: 'pyarrow<8.1.0,>=8.0.0; extra == "pandas"'
  warn_incompatible_dep(
INFO:fideslib.core.config:Loading file .fides/fides.toml from .
INFO:fideslib.core.config:Loading file .fides/fides.toml from .
INFO:fideslib.core.config:Loading file .fides/fides.toml from .  

There are missing required configuration variables. Please add the following config variables to either the `fides.toml` file or your environment variables to start Fides:

- app_encryption_key
- oauth_root_client_id
- oauth_root_client_secret

Visit the Fides deployment documentation for more information: https://ethyca.github.io/fides/deployment/
```

### DIFF
--- a/src/fides/ctl/core/config/security_settings.py
+++ b/src/fides/ctl/core/config/security_settings.py
@@ -6,7 +6,6 @@ from typing import Dict, List, Optional, Tuple, Union
 import validators
 from fideslib.core.config import FidesSettings
 from fideslib.cryptography.cryptographic_util import generate_salt, hash_with_salt
-from fideslib.exceptions import MissingConfig
 from pydantic import validator
 from slowapi.wrappers import parse_many  # type: ignore
 
--- a/src/fides/ctl/core/config/security_settings.py
+++ b/src/fides/ctl/core/config/security_settings.py
@@ -45,6 +45,11 @@ class SecuritySettings(FidesSettings):
         cls, v: Optional[str], values: Dict[str, str]
     ) -> Optional[str]:
         """Validate the encryption key is exactly 32 characters"""
+
+        # If the value is the default value, return immediately to prevent unwanted errors
+        if v == "":
+            return v
+
         if v is None or len(v.encode(values.get("encoding", "UTF-8"))) != 32:
             raise ValueError(
                 "APP_ENCRYPTION_KEY value must be exactly 32 characters long"
@@ -77,17 +82,16 @@ class SecuritySettings(FidesSettings):
     @classmethod
     def assemble_root_access_token(
         cls, v: Optional[str], values: Dict[str, str]
-    ) -> Tuple:
+    ) -> Optional[Tuple]:
         """
         Sets a hashed value of the root access key.
         This is hashed as it is not wise to return a plaintext for of the
         root credential anywhere in the system.
         """
-        value = values.get("oauth_root_client_secret")
+        value = values.get("oauth_root_client_secret", "")
+
         if not value:
-            raise MissingConfig(
-                "oauth_root_client_secret is required", SecuritySettings
-            )
+            return None
 
         encoding = values.get("encoding", "UTF-8")
 
--- a/src/fides/ctl/core/config/security_settings.py
+++ b/src/fides/ctl/core/config/security_settings.py
@@ -24,7 +24,7 @@ class SecuritySettings(FidesSettings):
     rate_limit_prefix: str = "fides-"
     aes_encryption_key_length: int = 16
     aes_gcm_nonce_length: int = 12
-    app_encryption_key: str
+    app_encryption_key: str = ""
     drp_jwt_secret: Optional[str] = None
     root_username: Optional[str] = None
     root_password: Optional[str] = None
@@ -32,8 +32,8 @@ class SecuritySettings(FidesSettings):
     encoding: str = "UTF-8"
 
     cors_origins: List[str] = []
-    oauth_root_client_id: str
-    oauth_root_client_secret: str
+    oauth_root_client_id: str = ""
+    oauth_root_client_secret: str = ""
     oauth_root_client_secret_hash: Optional[Tuple]
     oauth_access_token_expire_minutes: int = 60 * 24 * 8
     oauth_client_id_length_bytes = 16
--- a/src/fides/ctl/core/config/utils.py
+++ b/src/fides/ctl/core/config/utils.py
@@ -4,6 +4,7 @@ from typing import Any, Dict, Union
 from click import echo
 from fideslib.core.config import load_file
 from toml import dump, load
+
 from fides.ctl.core.utils import echo_red
 
 DEFAULT_CONFIG_PATH = ".fides/fides.toml"
@@ -59,13 +60,11 @@ def check_required_webserver_config_values() -> None:
     }
 
     missing_required_config_vars = []
-    for key in required_config_dict:
+    for key, value in required_config_dict.items():
         try:
-            config_value = getenv(
-                required_config_dict[key]["env_var"]
-            ) or get_config_from_file(
+            config_value = getenv(value["env_var"]) or get_config_from_file(
                 "",
-                required_config_dict[key]["config_subsection"],
+                value["config_subsection"],
                 key,
             )
         except FileNotFoundError:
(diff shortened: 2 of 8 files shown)


NevilleS:
So, a couple things I was noting while poking around a bit and the first thing I noticed is that the `CONFIG.security.app_encryption_key` is used directly in the various SQL classes, so ideally we'd just lazy-load that when the database is needed.

NevilleS:
Similarly, the `oauth_root_client_id` and `oauth_root_client_secret` would only be needed when an attempt is made to either create an oauth token or login as a root user.

So if we can add existence checks at that time, I'd think we'd be good there too. I suspect that'll start getting mixed up with `fideslib` though, so this'll get messy...

ThomasLaPiana:
static checks are still running but can I get 👀 ?

TheAndrewJackson:
This is possibly unrelated to this change but I get an error complaining about redis not being installed when running `fides webserver` so I'm not able to test that it's fixed.

Here is the error
```
ModuleNotFoundError: No module named 'redis'
```


I installed fides with `pipx install ethyca-fides`. I'm surprised that all the dependencies aren't installed.

src/fides/ctl/core/config/security_settings.py
```diff
@@ -45,6 +44,11 @@ def validate_encryption_key_length(
         cls, v: Optional[str], values: Dict[str, str]
     ) -> Optional[str]:
         """Validate the encryption key is exactly 32 characters"""
+
+        # If the value is the default value, return immediately to prevent unwanted errors
+        if v == "":
```
TheAndrewJackson, line 49: I like this workaround.

src/fides/ctl/core/config/utils.py
```diff
@@ -41,62 +42,50 @@ def get_config_from_file(
...
-            )
-        if not oauth_root_client_secret:
-            oauth_root_client_secret = get_config_from_file(
-                "", "security", "oauth_root_client_secret"
-            )
-    except FileNotFoundError:
-        pass
+def check_required_webserver_config_values() -> None:
+    """Check for required env vars and print a user-friendly error message."""
+    required_config_dict = {
```
TheAndrewJackson, line 47: Much more concise. I like it 🙂

src/fides/ctl/core/config/security_settings.py
```diff
@@ -24,16 +23,16 @@ class SecuritySettings(FidesSettings):
     rate_limit_prefix: str = "fides-"
     aes_encryption_key_length: int = 16
     aes_gcm_nonce_length: int = 12
-    app_encryption_key: str
+    app_encryption_key: str = ""
```
sanders41, line 26: With this being set to an empty string won't we need a check any time it is used to make sure it is set to something other than `""`? This seems the same as setting it to a default string which is what we wanted to remove, the difference being now we could check for `""` with this.

src/fides/ctl/core/config/security_settings.py
```diff
@@ -24,16 +23,16 @@ class SecuritySettings(FidesSettings):
...
     root_username: Optional[str] = None
     root_password: Optional[str] = None
     identity_verification_attempt_limit: int = 3  # 3 attempts
     encoding: str = "UTF-8"
 
     cors_origins: List[str] = []
-    oauth_root_client_id: str
-    oauth_root_client_secret: str
+    oauth_root_client_id: str = ""
+    oauth_root_client_secret: str = ""
```
sanders41, line 35: Same question about blank strings.

ThomasLaPiana:
the `fides deploy` error seems legit, troubleshooting locally

TheAndrewJackson:
image

Thanks, it would definitely help to install the right version 😅 Trying that now

src/fides/ctl/core/config/security_settings.py
ThomasLaPiana: if all of the values get checked for existence before the webserver exists, technically no right?

We're moving where we're doing the checking from the config itself to that function that gets run before the webserver

I agree this feels janky, but I think it is reasonable as a fix for now

ThomasLaPiana: we're using boolean checks (`if value:` == 'if "":`), which will come up as `False` when doing the checks (in theory?)

sanders41 (approved):
I got the expected result from each test. I also tested to be sure that if some of the required variables are set and not others with `fides webserver` the list displays the correct missing variables.

PR #1831, opened 2022-11-21
Update copy on `Scan for Systems` page

seanpreston:
Closes https://github.com/ethyca/fidesctl-plus/issues/339

### Code Changes

* [ ] Rename `Scan for Systems` page to `Add Systems`
* [ ] Updates description copy
* [ ] Updates tests

### Steps to Confirm

* [ ] Run `nox -s dev`
* [ ] Navigate to the `/config-wizard` flow
* [ ] Assert that `Scan for Systems` page now displays the correct copy

Screenshot 2022-11-21 at 11 45 03

### Pre-Merge Checklist

* Documentation Updated:
* [ ] Update `CHANGELOG.md`

### DIFF
--- a/clients/admin-ui/src/features/datastore-connections/add-connection/constants.ts
+++ b/clients/admin-ui/src/features/datastore-connections/add-connection/constants.ts
@@ -62,7 +62,7 @@ export const STEPS: AddConnectionStep[] = [
     label: "Choose your connection",
     href: `${DATASTORE_CONNECTION_ROUTE}/new?step=1`,
     description:
-      "The building blocks of your data map are the list of systems that exist in your organization. Think of systems as as anything that might store or process data in your organization, from a web application, to a database, or data warehouse.",
+      "Systems describe any services that store or process data for your organization, including third-party APIs, web applications, databases, and data warehouses. System discovery allows Fides to identify and build a data map from the list of systems that exist within your organization.",
     parentStepId: 0,
   },
   {
--- a/clients/admin-ui/src/features/datastore-connections/add-connection/constants.ts
+++ b/clients/admin-ui/src/features/datastore-connections/add-connection/constants.ts
@@ -1,5 +1,6 @@
 import { ItemOption } from "common/dropdown/types";
 
+import { ADD_SYSTEM_DESCRIPTION } from "~/features/system/constants";
 import { SystemType } from "~/types/api";
 
 import { DATASTORE_CONNECTION_ROUTE } from "../../../constants";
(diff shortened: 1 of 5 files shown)


clients/admin-ui/src/features/datastore-connections/add-connection/constants.ts
```diff
@@ -62,7 +62,7 @@ export const STEPS: AddConnectionStep[] = [
     label: "Choose your connection",
     href: `${DATASTORE_CONNECTION_ROUTE}/new?step=1`,
     description:
-      "The building blocks of your data map are the list of systems that exist in your organization. Think of systems as as anything that might store or process data in your organization, from a web application, to a database, or data warehouse.",
+      "Systems describe any services that store or process data for your organization, including third-party APIs, web applications, databases, and data warehouses. System discovery allows Fides to identify and build a data map from the list of systems that exist within your organization.",
```
allisonking, line 65: just double checking that we do indeed want to change this one? (it's not directly related to the ticket, but I can see wanting to keep all descriptions about systems consistent. though in that case we should probably pull it out into a variable that both files can use)

allisonking (approved):

clients/admin-ui/src/features/datastore-connections/add-connection/constants.ts
allisonking: we don't really have a great place for it right now. maybe a new `constants.ts` file in the `systems` directory? similar to the one we have in the `dataset` folder https://github.com/ethyca/fides/blob/main/clients/admin-ui/src/features/dataset/constants.ts

another option would be [`/common/constants.ts`](https://github.com/ethyca/fides/blob/main/clients/admin-ui/src/features/common/constants.tsx)

seanpreston: I've made this change and actually I think we'll want to use some other copy entirely in the connection creation flow. While QA'ing I've found it doesn't make a whole lot of sense:

Screenshot 2022-11-22 at 18 28 12

@rsilvery / @conceptualshark — any ideas here?

rsilvery: What have you removed? Trying to follow. What I think I'm seeing is that we need different copy for the Connections creation flow, correct?

seanpreston: I've removed the `Systems describe any services ...` copy, because this page is for adding Connections and not Systems.

PR #1833, opened 2022-11-21
Fix parent user creation bug

### LINKED ISSUE #1832: Uncaught error on parent user creation
### Bug Description

I just approved and merged https://github.com/ethyca/fides/pull/1819, but then soon after merging and doing some more development based on the feature, i realized there's bug :(

On startup, if the parent user already exists and its password has _not_ been updated in `fides.toml`, we try to [create the user again](https://github.com/ethyca/fides/pull/1819/files#diff-a3fe914840ef0720f85edf5f47bd86e59a23fdbed3c93dbd6e8062777e10e85aR83), which leads to a duplicate key DB error. Since the exception goes uncaught, we actually end up totally erroring out on startup. The error isn't even logged or reported anywhere (an issue in its own right, likely due to https://github.com/ethyca/fides/issues/1228?) and the application just hangs with this as the last message in the logs:
```
fides-fides-1  | 2022-11-21 16:25:33.213 [INFO] (seed:create_or_update_parent_user:83): Creating parent user
```

#### Steps to Reproduce

1. Add a `parent_server_username` and `parent_server_password` to the fides.toml file.
2. Start `fides` once
3. Shut it down, start it again
4. Watch it hang

### Expected behavior

- This is the base case for parent user creation, there shouldn't be any error or problem starting up in this case
- We should also take the opportunity to make sure we are catching any exceptions that are raised from the parent user creation, and at a minimum providing some information to the user about an issue that's encountered
    - I think we should also cleanly exit startup, because if a parent user is specified but there's an error in creation, then the app should not continue to boot up; it should call attention to the error so that it can be resolved

sanders41:
Closes #1832

### Code Changes

* [ ] Checks if the user exists with no changes to the password and exits gracefully if so

### Steps to Confirm

* [ ] Add a `parent_server_username` and `parent_server_password` to the fides.toml
* [ ] Run `nox -s dev`
* [ ] Stop the server
* [ ] Run `nox -s dev` again. The servers should not freeze and report healthy.

### Pre-Merge Checklist

* Documentation Updated:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

### DIFF
--- a/src/fides/api/ctl/database/seed.py
+++ b/src/fides/api/ctl/database/seed.py
@@ -79,6 +79,9 @@ def create_or_update_parent_user() -> None:
                 user.update_password(db_session, CONFIG.security.parent_server_password)
                 return
 
+            # User already exists and the password hasn't changed to nothing to do.
+            return
+
         log.info("Creating parent user")
         user = FidesUser.create(
             db=db_session,


sanders41:
Closing in favor of #1834

Issue #1827, opened 2022-11-18
Enable rescanning using the scan history diff endpoint


allisonking:
### Is your feature request related to a specific problem?

If a user has already done a scan before (there is something in the scan history), then we should render only the **added** systems in the registration flow. These systems should still go through the classify flow the same way as before. 

### Describe the solution you'd like

* [ ] Check if this is the first scan (check if `GET /api/v1/plus/scan/latest` 404s). Assuming it does not 404...
* [ ] After our normal `PUT /api/v1/plus/scan`, do a `GET /api/v1/plus/scan/latest?diff=true` to get the diff
* [ ] Render just the `added_systems` part of the response in the registration table
* [ ] Flow continues as normal

### Describe alternatives you've considered, if any

_A description of any alternative solutions or features you've considered._

### Additional context

Requires https://github.com/ethyca/fidesctl-plus/pull/333.

PR #1844, opened 2022-11-22
Rescan UI

allisonking:
Closes https://github.com/ethyca/fides/issues/1827

### Code Changes

* [x] Update typescript types with the new scan history and diff types
* [x] Update `LoadDataFlowScanner` so that it checks whether or not this is a rescan 
* [x] If it is a rescan, only show the added systems (as given to us by the `/scan/latest?diff=true` endpoint)
* [x] Adds a state for `systemsToClassify` in redux so that we aren't reusing `systemsToReview` anymore
* [x] Updates nextjs to the latest, which has a [fix for proxies timing out when they take over 30 seconds](https://github.com/vercel/next.js/issues/36251) (the data flow scan API sometimes does. eventually, this will be replaced with an async flow, but for now, we have to live with it)
* [x] When we navigate away from the scan results page (the last step of the onboarding flow), add some logic to reset the config wizard state so it's easy to go back to and run a scan again
* [x] Cypress tests

### Steps to Confirm

* [ ] Comment out the protected routes in `_app.tsx`
* [ ] Run `make server` in the fidesctl+ repo with scanning enabled
* [ ] Clear your local db of scan history and systems that may already be around
 
```
$ psql postgresql://postgres:fidesctl@localhost:5432/fidesctl
psql (14.6 (Homebrew), server 12.12 (Debian 12.12-1.pgdg110+1))
Type "help" for help.

fidesctl=# DELETE FROM plus_system_scans ;
DELETE 4
fidesctl=# DELETE FROM ctl_systems;
DELETE 45
```

* [ ] Now go through the add systems flow. Uncheck a few systems but register the rest.
* [ ] Click on "Add Systems" again to be prompted to scan again
* [ ] This time the scan should return the systems you didn't register earlier. Register them all
* [ ] Click on "Add Systems" again to be prompted to scan again
* [ ] This time the scan should find no new systems to register

### Pre-Merge Checklist

* Documentation Updated:
* [x] Update `CHANGELOG.md`

### Description Of Changes

(video edited to cut out some of the load screen times)

https://user-images.githubusercontent.com/24641006/204342261-4fdededd-70da-427d-a715-aa05b4b8c54b.mov


clients/admin-ui/next.config.js
```diff
@@ -20,6 +20,9 @@ const nextConfig = {
 
         return config;
     },
+    experimental: {
```
ssangervasi, line 23: Could benefit from a comment with your PR explanation.

clients/admin-ui/src/features/system/system.slice.ts
```diff
@@ -108,8 +108,9 @@ export const {
 export interface State {
   activeSystem?: System;
   activeClassifySystemFidesKey?: string;
+  systemsToClassify: System[];
 }
-const initialState: State = {};
+const initialState: State = { systemsToClassify: [] };
```
ssangervasi, line 113: Not sure if it applies here, but `systemsForReview` initialize as undefined to differentiate between "nothing selected _yet_" and "zero things were selected". The ScanResults component might, for example, distinguish between "no results came back yet" and "empty results" (but maybe the preceeding steps mean it doesn't matter).

Also, does this state ever get cleared? I don't see a `setSystemsToClassify([])` anywhere.

clients/admin-ui/src/types/api/models/SystemsDiff.ts
```diff
@@ -0,0 +1,17 @@
+/* istanbul ignore file */
+/* tslint:disable */
+/* eslint-disable */
+
+import type { System } from "./System";
+
+/**
+ * A schema representing the diff between the scan and saved systems
+ */
+export type SystemsDiff = {
```
ssangervasi, line 10: Fancy API 🎩

ssangervasi (approved):

clients/admin-ui/src/features/system/system.slice.ts
allisonking: hmm that's a good point about it not getting cleared. this page kind of lives in a weird place in the onboarding flow in that it is part of it, but is also its own page so that people can feasibly go back to it (although we don't include it in the nav bar yet). so I'm not sure when the use case for clearing it would be, but I'll keep noodling on it as we add to the feature

and yeah that makes sense about empty vs undefined—I'll update!

PR #1850, opened 2022-11-23
Update wording on scan page

seanpreston:
### Description Of Changes

This PR updates copy on the scan results page from 

> Below are search results for {region}. Please select and register the systems you would like to maintain in your mapping and reports.

to

> Below are the results of your infrastructure scan. To continue, select the systems you would like registered in your data map and reports.


allisonking (approved):
💥

PR #1859, opened 2022-11-28
Move `fideslib` files into `fides`

### LINKED ISSUE #1572: Merge `fideslib` into `unified-fides`

ThomasLaPiana:
Closes #1572 

### Code Changes

* [ ] copy pasta the source code into `src/fides/lib`
* [ ] update code paths
* [ ] move the tests over and make sure they get included in the test runs

### Steps to Confirm

### Pre-Merge Checklist

* Documentation Updated:
* [ ] Update `CHANGELOG.md`

### Description Of Changes


src/fides/lib/db/base_class.py
```diff
@@ -0,0 +1,290 @@
...
+from fideslib.exceptions import KeyOrNameAlreadyExists, KeyValidationError
+from fideslib.utils.text import to_snake_case
+from sqlalchemy import Column, DateTime, String
+from sqlalchemy.ext.declarative import declarative_base, declared_attr
+from sqlalchemy.orm import Query, Session
+from sqlalchemy.sql import func
+from sqlalchemy.sql.expression import BinaryExpression, BooleanClauseList
+
+T = TypeVar("T", bound="OrmWrappedFidesBase")
+ALLOWED_CHARS = re.compile(r"[A-z0-9\-_]")
```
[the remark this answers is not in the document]
ThomasLaPiana: @sanders41 thoughts here?

ThomasLaPiana:
currently got everything working, but need to go in and fix some tests/checks that are now failing due to the code move

src/fides/lib/db/base_class.py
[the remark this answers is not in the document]
sanders41: Based on this message it looks like we could improve the regex, but it is also validating the key and not a password or anything security related so it doesn't seem like huge issue.

ThomasLaPiana:
tests should all be passing now, next step is to fix static checks and then should be good to go

ThomasLaPiana:
@sanders41 any idea where these failures are coming from?

https://github.com/ethyca/fides/actions/runs/3586465322/jobs/6035688944

`fideslib` shouldn't exist at all, so how is it getting imported here? do you know what the fides_client is?

sanders41:
This isn't helpful other than to let you know you aren't going crazy...the tests pass locally for me also and I can't find any reference to `fideslib` anywhere. I think the Docker image gets cached in order to share between tests. Is there any change it's using an old image that still has references to `fideslib` and that is causing the error? Seems unlikely, but maybe?

ThomasLaPiana:
> This isn't helpful other than to let you know you aren't going crazy...the tests pass locally for me also and I can't find any reference to `fideslib` anywhere. I think the Docker image gets cached in order to share between tests. Is there any chance it's using an old image that still has references to `fideslib` and that is causing the error? Seems unlikely, but maybe?

Every little bit of confirmation helps! haha glad to know I was right in that locally everything is working...at least I've narrowed the problem down to CI and potentially caching, I'll take a look there

ThomasLaPiana:
so I fixed this problem by merging in main, but I'm still not sure why it was happening in the first place, this could be an issue

ThomasLaPiana:
mypy is now a major issue here, as it appears `fideslib` had some typing issues/our usage of fideslib had some typing issues

sanders41:
Looking at the mypy errors, I think it is because mypy is set to ignore missing imports from fideslib. Fideslib has a py.typed file so it shouldn’t have had missing imports, but it looks like it wasn’t getting included for some reason making all mypy errors related to fideslib ignored. 

So this _should_ mean all the errors are in places fides was calling fideslib and not in the fideslib code itself.

ThomasLaPiana:
> Looking at the mypy errors, I think it is because mypy is set to ignore missing imports from fideslib. Fideslib has a py.typed file so it shouldn’t have had missing imports, but it looks like it wasn’t getting included for some reason making all mypy errors related to fideslib ignored.
> 
> So this _should_ mean all the errors are in places fides was calling fideslib and not in the fideslib code itself.

Yep! Exactly what I thought

There are quite a few errors though (77 across 39 files I think) so it'll take me a bit to chew through them, but otherwise everything is looking pretty good

ThomasLaPiana:
I'm slowly chipping away at these, but there are some fundamental type errors here. Will end up ignoring most of them, or fixing where it isn't too invasive

ThomasLaPiana:
Failing external tests are expected (note `main` is also failing)

src/fides/lib/oauth/api/deps.py
```diff
@@ -0,0 +1,47 @@
...
+    """Return our database session.
+
+    This should be overridden by the installing package.
+    """
+    try:
+        SessionLocal = get_db_session(get_config())
+        db = SessionLocal()
+        yield db
+    finally:
+        db.close()
```
sanders41, line 25: There is a separate version of `get_db` in `src/fides/ops/api/api/deps.py` that used to be used to override this one. Now that they are using the same config we should be able to remove the one from ops.

src/fides/lib/oauth/api/routes/user_endpoints.py
```diff
@@ -0,0 +1,256 @@
+import logging
```
sanders41, line 1: The extra/overriding endpoints from `src/fides/api/ops/api/v1/endpoints/user_endpoints.py` could be moved into here if they were extra, or removed if overriding.

src/fides/lib/oauth/api/urn_registry.py
```diff
@@ -0,0 +1,16 @@
+# Prefix
```
sanders41, line 1: We could move this into the larger urn_registry now that everything is together, or the ones from ops could move here.

src/fides/lib/oauth/scopes.py
```diff
@@ -0,0 +1,164 @@
+AUTHORIZE = "authorize"
```
sanders41, line 1: These can move into `src/fides/api/ops/api/v1/scopes_registry.py` or the ones from ops could move here. They should already be close to the same, but got out of sync over time.

src/fides/lib/oauth/api/deps.py
```diff
@@ -0,0 +1,47 @@
...
+    authorization: str = Security(oauth2_scheme()),
+    db: Session = Depends(get_db),
+) -> ClientDetail:
+    """Calls oauth_util.verify_oauth_client.
+
+    This is here because config values are needed, this dependency should be overridden
+    by the installing library.
+    """
+    config = get_config()
+    return verify(security_scopes, authorization, db=db, config=config)
```
pattisdr, line 47: Can we rework so this isn't needed? This was something that was added to fideslib when we moved authentication to fideslib.  We should be able to use https://github.com/ethyca/fides/pull/1859/files#diff-e1a49d5e530f5afa7a461c486f5a216ad3fe94131d8130872dcafbdbd2337b89R33 directly

src/fides/lib/oauth/oauth_util.py
```diff
@@ -0,0 +1,80 @@
...
+    return (datetime.now() - issued_at).total_seconds() / 60.0 > token_duration_min
+
+
+def verify_oauth_client(
+    security_scopes: SecurityScopes,
+    authorization: str,
+    *,
+    db: Session,
+    config: FidesConfig,
+) -> ClientDetail:
```
pattisdr, line 39: Very similar code exists in src/fides/api/ops/util/oauth_util.py and I think that is the version we should be using, not this one.  Ops was overriding the fideslib version.

.fides/fides.toml
```diff
@@ -49,3 +49,8 @@ task_retry_delay = 1
 
 [admin_ui]
 enabled = true
+
+[celery]
+event_queue_prefix = "fides_worker"
+task_default_queue = "fides"
+task_always_eager = true
```
pattisdr, line 56: Can you explain these changes - getting rid of the celery.toml and adding this to the config?

pattisdr (commented):
good type hinting improvements here

.fides/fides.toml
ThomasLaPiana: @pattisdr this was an additional upstream PR, having `celery.toml` as a separate config file didn't feel like a good UX, so now the user can directly define a `celery` subsection in their `fides.toml` and it will get passed through as expected. Additionally it is not validated, so the user can pass through whatever they want (same functionality as before, just in a single location)

pattisdr: thanks

.github/workflows/backend_checks.yml
```diff
@@ -128,9 +126,6 @@ jobs:
 
       - name: Set Up Python
         uses: actions/setup-python@v4
-        with:
-          python-version: ${{ env.DEFAULT_PYTHON_VERSION }}
-          cache: "pip"
```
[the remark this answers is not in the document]
ThomasLaPiana: I was throwing smores at the wall trying to get something to stick with that weird caching issue we saw. I need to add this back in! good catch

.gitignore
```diff
@@ -115,7 +117,6 @@ develop-eggs/
 dist/
 downloads/
 eggs/
-lib/
```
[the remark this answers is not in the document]
ThomasLaPiana: ah yes, I believe this was an errant erasure

src/fides/ctl/core/config/helpers.py
```diff
@@ -0,0 +1,254 @@
...
+    raises FileNotFound if none is found
+    """
+
+    possible_directories = [
+        os.getenv("FIDES__CONFIG_PATH"),
+        os.curdir,
+        os.pardir,
+        os.path.abspath(os.path.join(os.curdir, "..", "..")),
+        os.path.join(os.pardir, ".fides"),
+        os.path.expanduser("~"),
```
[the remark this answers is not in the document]
ThomasLaPiana: it does! I guess python is smart enough to figure it out in a platform-agnostic way

src/fides/ctl/core/config/helpers.py
```diff
@@ -0,0 +1,254 @@
...
+    :param updates: A nested `dict`, where top-level keys correspond to configuration sections and top-level values contain `dict`s whose key/value pairs correspond to the desired option/value updates.
+    :param config_path: The path to the current configuration file.
+    """
+
+    try:
+        config_path = config_path_override or load_file(
+            file_names=[DEFAULT_CONFIG_PATH]
+        )
+    except FileNotFoundError as error:
+        raise error
```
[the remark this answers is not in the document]
ThomasLaPiana: agreed!

src/fides/lib/oauth/api/deps.py
ThomasLaPiana: I'll take a look at consolidating these further, thanks!

ThomasLaPiana:
@pattisdr given the already large amount of code changes here, I decided that it is probably better/safer to create a separate issue/PR for further consolidation so that it will be easier to review those changes and verify that they didn't cause any breaks

https://github.com/ethyca/fides/issues/2032

src/fides/lib/oauth/api/urn_registry.py
ThomasLaPiana: I've created a new issue for further refactoring, so it'll be easier to track these smaller changes when reviewing

src/fides/lib/oauth/scopes.py
ThomasLaPiana: see above comment, created another issue to further deduplicate

sanders41 (approved):
lgtm. Thanks for this, it will make things easier since we merged repos.

TheAndrewJackson (approved):
:shipit:

PR #1894, opened 2022-11-29
Add endpoint to retrieve child access request

### LINKED ISSUE #1750: Parent Fides can create a combined response of all results of child Fides nodes
### Description
Child nodes must still write to their own s3 bucket and the parent node must download from all the child nodes and create an archive.

User Story - As a Fides admin, I want Fides to automatically create a combined response of all the results of child nodes, from multiple Fides instances, once I know that all the child nodes have completed their DSR response.

### Acceptance Criteria

- Fides can create a combined response of all results of child Fides nodes

### Implementation Notes/Details
will need to zip all different results into one

sanders41:
Closes #1750

### Code Changes

* Added an endpoint to retrieve the access request data from the child

### Steps to Confirm

### Pre-Merge Checklist

* Documentation Updated:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

### DIFF
--- a/src/fides/api/ops/api/v1/endpoints/privacy_request_endpoints.py
+++ b/src/fides/api/ops/api/v1/endpoints/privacy_request_endpoints.py
@@ -42,6 +42,7 @@ from fides.api.ops.api.v1.scope_registry import (
     PRIVACY_REQUEST_CREATE,
     PRIVACY_REQUEST_READ,
     PRIVACY_REQUEST_REVIEW,
+    PRIVACY_REQUEST_TRANSFER,
     PRIVACY_REQUEST_UPLOAD_DATA,
     PRIVACY_REQUEST_VIEW_DATA,
 )
@@ -56,6 +57,7 @@ from fides.api.ops.api.v1.urn_registry import (
     PRIVACY_REQUEST_RESUME,
     PRIVACY_REQUEST_RESUME_FROM_REQUIRES_INPUT,
     PRIVACY_REQUEST_RETRY,
+    PRIVACY_REQUEST_TRANSFER_TO_PARENT,
     PRIVACY_REQUEST_VERIFY_IDENTITY,
     PRIVACY_REQUESTS,
     REQUEST_PREVIEW,
@@ -1284,6 +1286,40 @@ def upload_manual_webhook_data(
         access_manual_webhook,
         privacy_request,
     )
+
+
+@router.get(
+    PRIVACY_REQUEST_TRANSFER_TO_PARENT,
+    status_code=HTTP_200_OK,
+    dependencies=[Security(verify_oauth_client, scopes=[PRIVACY_REQUEST_TRANSFER])],
+    response_model=Dict[str, Optional[List[Row]]],
+)
+def privacy_request_data_transfer(
+    *,
+    privacy_request_id: str,
+    db: Session = Depends(deps.get_db),
+    cache: FidesopsRedis = Depends(deps.get_cache),
+) -> Dict[str, Optional[List[Row]]]:
+    """Transfer access request iinformation to the parent server."""
+    privacy_request = PrivacyRequest.get(db=db, object_id=privacy_request_id)
+
+    if not privacy_request:
+        raise HTTPException(
+            status_code=HTTP_400_BAD_REQUEST,
+            detail=f"No privacy request with id {privacy_request_id} found",
+        )
+
+    value_dict = cache.get_encoded_objects_by_prefix(
+        f"{privacy_request_id}__access_request"
+    )
+
+    if not value_dict:
+        raise HTTPException(
+            status_code=HTTP_400_BAD_REQUEST,
+            detail=f"No access request information found for privacy request id {privacy_request_id}",
+        )
+
+    return {k.split("__")[-1]: v for k, v in value_dict.items()}
 
 
 @router.get(
--- a/src/fides/api/ops/api/v1/scope_registry.py
+++ b/src/fides/api/ops/api/v1/scope_registry.py
@@ -26,6 +26,7 @@ PRIVACY_REQUEST_CALLBACK_RESUME = (
     "privacy-request:resume"  # User has permission to resume a privacy request
 )
 PRIVACY_REQUEST_REVIEW = "privacy-request:review"
+PRIVACY_REQUEST_TRANSFER = "privacy-request:transfer"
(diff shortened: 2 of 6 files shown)


sanders41:
@adamsachs I have what we talked about in place here. I'm still trying to find a good way to test it.

When I run a privacy request manually in the privacy center it shows that the request completed successfully, but I see an error in the logs when I'm running with a child server `nox -s dev -- child`. The real error is getting masked by the `RuntimeError: this event loop is already running.` error. I tried commenting out the code that retrieves and adds the child data to the upload and the log error still shows up so it seems to happen somewhere before I try to gather results.

My current guess is it is in the connector polling, but that is only because I know `@sync` is getting used there so it seems like a likely spot.

src/fides/api/ops/service/privacy_request/request_runner_service.py
```diff
@@ -516,3 +531,74 @@ def generate_id_verification_code() -> str:
...
+
+    if response.status_code != 200:
+        logger.error(
+            "Error retrieving data from child server for privacy request %s: %s",
+            fides_connector[0],
+            response.json(),
+        )
+        return None
+
+    return {fides_connector[1].key: response.json()}
```
adamsachs, line 604: i'm not sure this is the right key? at first glance i'm not 100% sure on the shape of the response data, so i may be mistaken - but i don't think we'll want to put the connector key as the key for this data. shouldn't the response data be "flattened" onto the aggregate results?

will step through it with real data on my end locally tomorrow to get a firmer understanding here..

src/fides/api/ops/service/privacy_request/request_runner_service.py
```diff
@@ -516,3 +531,74 @@ def generate_id_verification_code() -> str:
     Generate one-time identity verification code
     """
     return str(random.choice(range(100000, 999999)))
+
+
+def _retrieve_child_results(  # pylint: disable=R0911
+    fides_connector: Tuple[str, ConnectionConfig], rule_key: str
+) -> Optional[Dict[str, Optional[List[Row]]]]:
+    """Get child access request results to add to upload."""
+    secrets = fides_connector[1].secrets
```
adamsachs, line 540: not a huge deal, but on first glance i'm thinking you may want to just instantiate a real `FidesConnector` and use some of its methods to initialize and authenticate a `FidesClient`? just may help to DRY this up a bit over the codebase so you don't have to go through the secrets parsing and login/client instantiation again

src/fides/api/ops/service/privacy_request/request_runner_service.py
```diff
@@ -205,6 +212,13 @@ def upload_access_results(
             manual_data
         )  # Add manual data directly to each upload packet
 
+        if fides_connectors_by_dataset:
+            for child in fides_connectors_by_dataset:
+                child_data = _retrieve_child_results(child, rule.key)
```
adamsachs, line 217: nice - i like this in its own method. keeps things clean up here.

src/fides/api/ops/api/v1/endpoints/privacy_request_endpoints.py
```diff
@@ -1286,6 +1290,77 @@ def upload_manual_webhook_data(
...
+    privacy_request = PrivacyRequest.get(db=db, object_id=privacy_request_id)
+
+    if not privacy_request:
+        raise HTTPException(
+            status_code=HTTP_404_NOT_FOUND,
+            detail=f"No privacy request with id {privacy_request_id} found",
+        )
+
+    rule = Rule.filter(db=db, conditions=(Rule.key == rule_key)).first()
+    print(rule.__dict__)
```
adamsachs, line 1316: assuming this is for debugging, fine for now but just don't want to forget about it before we merge

src/fides/api/ops/service/privacy_request/request_runner_service.py
```diff
@@ -516,3 +531,74 @@ def generate_id_verification_code() -> str:
...
+            "Error logging into to child server for privacy request %s: %s",
+            fides_connector[0],
+            e,
+        )
+        return None
+
+    try:
+        request = client.authenticated_request(
+            method="get",
+            path=f"{V1_URL_PREFIX}{PRIVACY_REQUEST_TRANSFER_TO_PARENT.format(privacy_request_id=fides_connector[0], rule_key=rule_key)}",
```
adamsachs, line 584: `fides_connector[0]` isn't going to be the privacy request ID, it's going to be the address corresponding to the dataset/collection for the given `FidesConnector`. i think you'll essentially need to do a look up on the full access results with this string, to grab the `PrivacyRequestResponse` dict, and look up its `"id"` field.

i realize we didn't go into this while talking earlier. i can take a closer look tomorrow morning and could probably take a stab at writing what's needed, if what i said above makes no sense :)

adamsachs:
@sanders41 this sounds good, and generally this looks like it's pretty close. tomorrow AM i'm going to take a closer look and try to set up at least a manual smoke test so that we can step through things.

i did leave some initial comments from a quick glance, just to see if they spark anything that you want to take a look at quickly. i think [this comment](https://github.com/ethyca/fides/pull/1894#discussion_r1035490977) may be a (the?) spot where a fix is needed.

src/fides/api/ops/service/privacy_request/request_runner_service.py
sanders41: I _think_ I updated the function to do what you are saying here, but if I got it wrong let me know.

src/fides/api/ops/service/privacy_request/request_runner_service.py
```diff
@@ -516,3 +548,75 @@ def generate_id_verification_code() -> str:
...
+        logger.error(
+            "Error create client for child server %s: %s", fides_connector[0], e
+        )
+        return None
+
+    results = []
+
+    for key, rows in access_result.items():
+        if key.startswith(
+            f"{fides_connector[0]}:"
```
adamsachs, line 571: just seeing if i can quickly address this

adamsachs: should be addressed in https://github.com/ethyca/fides/pull/1894/commits/d0d6e76483b817d103fb392f100dfe8643215f8b

adamsachs (approved):
this all looks good to me now, thanks for all the iterations to cover the edge cases!

PR #1964, opened 2022-12-01
Update how `config` gets loaded into each module

### LINKED ISSUE #1959: Update how the `config` is loaded in each module
no longer having `CONFIG` available anywhere as a constant. Functions should call `get_config` as needed (as a function parameter with a default value of `get_config()`), not instantly call it when the module gets imported.

ThomasLaPiana:
Closes #1959 

### Code Changes

* [x] update every instance of `CONFIG = get_config()`

### Steps to Confirm

### Pre-Merge Checklist

* Documentation Updated:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

The changes here are designed to eliminate the use of a global `config` file, and instead treat it as something that always gets passed in to whatever function needs it, or in some cases call `get_config` directly when passing in isn't possible

#### OLD
```python
CONFIG = get_config()
...
   def get_masking_request(self) -> Optional[SaaSRequest]:

        if not CONFIG.execution.masking_strict:
            gdpr_delete = self.data_protection_request
            delete = self.get_erasure_request_by_action("delete")
```

#### NEW
```python
def get_masking_request(
    self, masking_strict: bool = get_config().execution.masking_strict
) -> Optional[SaaSRequest]:
    if not CONFIG.execution.masking_strict:
        gdpr_delete = self.data_protection_request
        delete = self.get_erasure_request_by_action("delete")
```


ThomasLaPiana:
hitting some difficulties here, not sure why the ops tests are failing. Only thing I can think of is that the global config is getting updated somewhere but it isn't propagating for some reason like it did before

ThomasLaPiana:
this is turning into quite the headache, a lot of this stuff is intertwined to the point that undoing this ball of yarn in a single PR doesn't feel good. I think doing it slowly, maybe first ctl and then ops, would be more prudent but I'll keep poking at this

sanders41:
The ctl `conftest.py` has a fixture that clears the `get_config` cache between tests. Maybe adding it to ops also will make a difference here. 

```py
@pytest.fixture(autouse=True)
def clear_get_config_cache() -> None:
    get_config.cache_clear()

```

ThomasLaPiana:
> The ctl `conftest.py` has a fixture that clears the `get_config` cache between tests. Maybe adding it to ops also will make a difference here.
> 
> ```python
> @pytest.fixture(autouse=True)
> def clear_get_config_cache() -> None:
>     get_config.cache_clear()
> ```

that's a good idea, that was going to be my last shot, I'll give it a try and see what happens! Thanks 😄

ThomasLaPiana:
I'm going to revisit this once the other config changes are done, so this PR is outdated

PR #1982, opened 2022-12-05
Refactor the logic that hydrates the config

### LINKED ISSUE #1962: Simplify config loading/logic within the main code path

ThomasLaPiana:
Closes #1962 

### Code Changes

* [x] add a new `build_config` function that elegantly handles hydrating the config
* [x] remove `security`-specific workarounds
* [x] make all config subsections required for more safety (a default config can still be generated using `build_config`

### Steps to Confirm

* [ ] play with changing local env vars and config file vars and seeing the changes propagate

### Pre-Merge Checklist

* Documentation Updated:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

Simplifies hydrating the config by programmatically instantiating each subsection to prevent bugs caused by env vars not being respected.

Lots of workarounds/trickery has been removed and should make understanding the code flow generally much less complex and consistent :) 

__NOTE:__ This is branchception

### DIFF
--- a/src/fides/ctl/core/config/__init__.py
+++ b/src/fides/ctl/core/config/__init__.py
@@ -26,7 +26,12 @@ from .notification_settings import NotificationSettings
 from .redis_settings import RedisSettings
 from .security_settings import SecuritySettings
 from .user_settings import UserSettings
-from .utils import CONFIG_KEY_ALLOWLIST, DEFAULT_CONFIG_PATH, get_test_mode
+from .utils import (
+    CONFIG_KEY_ALLOWLIST,
+    DEFAULT_CONFIG_PATH,
+    DEFAULT_CONFIG_PATH_ENV_VAR,
+    get_test_mode,
+)
 
 
 class FidesConfig(FidesSettings):
@@ -34,6 +39,7 @@ class FidesConfig(FidesSettings):
     Composite class that encapsulates all of the config subsections.
     """
 
+    # Setting Subsections
     admin_ui: AdminUISettings = AdminUISettings()
     cli: CLISettings = CLISettings()
     credentials: Dict[str, Dict] = {}
@@ -47,6 +53,7 @@ class FidesConfig(FidesSettings):
     security: SecuritySettings = SecuritySettings.construct()
     user: UserSettings = UserSettings()
 
+    # Root Settings
     test_mode: bool = get_test_mode()
     is_test_mode: bool = test_mode
     hot_reloading: bool = getenv("FIDES__HOT_RELOAD", "").lower() == "true"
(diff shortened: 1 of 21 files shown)


src/fides/ctl/core/config/__init__.py
```diff
@@ -108,6 +108,45 @@ def censor_config(config: FidesConfig) -> Dict[str, Any]:
     return filtered
 
 
+def build_config(config_dict: Dict[str, Any]) -> FidesConfig:
```
ThomasLaPiana, line 111: @sanders41 I think I finally found the solution to our config woes...the tests pass at least

sanders41 (approved):
Hopefully if something does go wrong this also give a better error that makes it easier to track down the problem.

TheAndrewJackson (approved):

src/fides/ctl/core/config/__init__.py
```diff
@@ -108,6 +108,45 @@ def censor_config(config: FidesConfig) -> Dict[str, Any]:
...
+    """
+    This function builds and instantiates a FidesConfig.
+
+    It will pull from the provided config_dict when possible for subsections,
+    otherwise instantiating the settings with default values.
+
+    The settings_map will need to be kept up-to-date manually with the
```
NevilleS, line 118: This'll be hard to remember and maintain, can you do a pytest that checks these stay in sync that would fail helpfully when someone adds to the settings model without updating the build config function?

NevilleS (approved):
One ask for maintainability, but otherwise 👍

NevilleS:
Thanks! Two questions:

1. I worry that it'll be too easy to forget to keep the build_config function up to date, can you add a test or something similar to ensure that happens?
2. You say "all subsections are required" but is that a breaking change? Right now TOMLs can be sparsely populated with only the subsections that are needed

src/fides/ctl/core/config/__init__.py
ThomasLaPiana: @NevilleS I'm not sure that is necessary. If someone adds a new config section but forgets to add it here, it will not show up in the config at all and presumably the new functionality would also not work.

I'm also not sure right off-hand what an automated way to do that would be, otherwise I would've definitely avoided this manual solution (not a fan of these kinds of solutions, but I documented it at least!) 

Given the rarity with which we add new config subsections and the unlikelihood of causing any errors in its fail-state, I am ok with it as-is, but lmk if you disagree

ThomasLaPiana:
> Thanks! Two questions:
> 
> 1. I worry that it'll be too easy to forget to keep the build_config function up to date, can you add a test or something similar to ensure that happens?
> 2. You say "all subsections are required" but is that a breaking change? Right now TOMLs can be sparsely populated with only the subsections that are needed

I disagree on 1, commented above, lmk what your thoughts are after reading

Sorry, let me clarify on point 2. If you try to instantiate the config directly (`FidesConfig()`) it will break. But using the new `build_config` function, it will handle explicitly instantiating each subsection to avoid this problem.

From a user-facing perspective, there are zero changes here (note the zero test changes), it would only matter if someone tried to do `FidesConfig()` as opposed to `build_config()` or `get_config()`

ThomasLaPiana:
my upstream celery branch was approved and merged, so I will make some more docs updates here to account for that as well

conceptualshark (approved):

PR #2000, opened 2022-12-07
1999 connectors connection test endpoints can clog up the whole server

### LINKED ISSUE #1999: Connector's "connection test" endpoints can clog up the whole server
### Bug Description

When testing a connector's connection to its external resource, if our test request to the external resource hangs for whatever reason, the whole fides webserver becomes unresponsive. This is because our `/connection/{connection_key}/test` endpoint is defined as an `async` function (`async def test_connection_config_secrets`), but we generally are not properly using `async` clients to communicate with our external resources. By doing this, we're effectively telling FastAPI we'll give up control of the event loop as soon as we can, but in reality we're not upholding our end of the promise.

Specifically, this problem came up when testing the connection of a `fides connector` and the "child" fides is _stuck_ (for whatever reason). The problem manifested as the "parent" fides becoming totally unresponsive (and actually failing to respond to its own health checks).

#### Steps to Reproduce

In the wild, you'd need to set up a parent and child fides and configure the parent to connect to the child via a `fides connector`. You'd then need to make the child fides somehow hang on API requests issued against it - note that simply removing the child fides, or configuring the parent's `fides connector` to point to a bogus URL, won't reproduce the issue because the parent's connection test will just quickly error out in this case. The child fides must _hang_ in responding to the parent fides' connection tests.

You could also reproduce this with any other type of connector that connects to a system that will _hang_ in responding to the configured test requests.

Locally, i reproduced the issue with the following setup:

1. Update the `docker-compose.child-env.yml` to allow for remote debugging:
    a.  update the `fides-child.command` to be: `python -m debugpy --listen 0.0.0.0:5678 -m uvicorn --host 0.0.0.0 --port 8080 --reload --reload-dir src fides.api.main:app`
    b. update the `fides-child.ports` to include an additional port mapping entry, e.g.`"5679:5678"`
2. run `nox -s dev -- remote_debug child`
3. update the vscode debug config `Remote Attach Fides` to have a `connect.port` of `5679` instead of `5678`
4. put a breakpoint close to the top of `privacy_request_endponits.get_request_status` , as this is the endpoint that the parent fides hits on the child fides. this will allow us to pause execution on the child here, effectively mimicking a hung child server
5. configure a fides connector on the parent to talk to the child
6. start your remote debugger that connects to the child fides
7. explicitly hit your connector's `/connection/{connection_key}/test` endpoint on the parent
8. your breakpoint should be hit on the child, meaning that the connection test on the parent is effectively hung waiting for a response
9. try to hit your parent webserver with any other request, e.g. `curl localhost:8080/health`

### Expected behavior

Even if a connection test gets hung up, our server's core functionality and responsiveness should not be severely impacted

### Screenshots

@RobertKeyser not sure if you want to drop your loom in here, not sure how public that is...

### Additional context

As noted above - this problem came up when testing the connection of a `fides connector` and the "child" fides is _stuck_ (for whatever reason). The problem manifested as the "parent" fides becoming totally unresponsive (and actually failing to respond to its own health checks).

adamsachs:
Closes #1999 

### Code Changes

* removes the `async` keyword for the methods that define our connection endpoints that actually test a connection
* adds a timeout into the fides client's connection test call, so that unexpected hanging from the "child" fides will be more gracefully handled

### Steps to Confirm

* see issue for detailed repro steps. i went through those and confirmed we no longer hold up the web server in the case where the connection test hangs
* i also confirmed that the new 5 second timeout is respected in these cases, and we get a clean error logged on the parent (as well as a clear connection test response that indicates a faiulre)

### Pre-Merge Checklist

* Documentation Updated:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

Note that another option to resolve this issue would be to keep the endpoints as `async` functions and instead ensure that all of our connection tests are effectively `async` down the stack, and ceding control of the event loop when they reach out to make their calls to the external resources. that felt like a big change that would require a fair amount of regression testing, if it's even feasible.

I feel more comfortable going down this route, especially given some of the general guidance found on the fastAPI [docs](https://fastapi.tiangolo.com/async/):

> If you are using a third party library that communicates with something (a database, an API, the file system, etc.) and doesn't have support for using await, (this is currently the case for most database libraries), then declare your path operation functions as normally, with just def, like:

### DIFF
--- a/src/fides/api/ops/api/v1/endpoints/connection_endpoints.py
+++ b/src/fides/api/ops/api/v1/endpoints/connection_endpoints.py
@@ -357,7 +357,7 @@ def connection_status(
     dependencies=[Security(verify_oauth_client, scopes=[CONNECTION_CREATE_OR_UPDATE])],
     response_model=TestStatusMessage,
 )
-async def put_connection_config_secrets(
+def put_connection_config_secrets(
     connection_key: FidesOpsKey,
     *,
     db: Session = Depends(deps.get_db),
@@ -392,7 +392,7 @@ async def put_connection_config_secrets(
     dependencies=[Security(verify_oauth_client, scopes=[CONNECTION_READ])],
     response_model=TestStatusMessage,
 )
-async def test_connection_config_secrets(
+def test_connection_config_secrets(
     connection_key: FidesOpsKey,
     *,
     db: Session = Depends(deps.get_db),
--- a/src/fides/api/ops/service/connectors/fides/fides_client.py
+++ b/src/fides/api/ops/service/connectors/fides/fides_client.py
@@ -177,7 +177,7 @@ class FidesClient:
             if privacy_request_id
             else None,
         )
-        response = self.session.send(request)
+        response = self.session.send(request, timeout=5)
         if not response.ok:
             log.error(
                 f"Error retrieving status of privacy request [{privacy_request_id}] on remote Fides {self.uri}",


sanders41 (approved):
Looks good. This seems to be by far the quickest fix right now.

PR #2018, opened 2022-12-09
asserts test_mode is true in task config

seanpreston:


pattisdr (approved):
Thanks for verifying test mode in this context @seanpreston 

For those following along, this came out of a discussion about https://github.com/ethyca/fides/pull/1731/commits/f5ddef8db8aae5e2d20e50b01adfcc2c9c222fa6.

We have a single end-to-end test failing in the work where we're trying to switch to populating the application db in application context and the test db for pytest.   When standing up a worker to run pytest against (just to figure out the issue), the worker is not in test mode, but this verifies that it is in test mode in this different context, at least.

ThomasLaPiana:
@seanpreston is this still relevant?

PR #2023, opened 2022-12-12
Do not always create a `fides.toml` by default

### LINKED ISSUE #1661: Support the ability for users to run `fides` without a `.fides` dir or file-based config

Currently `fides` will _always_ try to write out a config file if it can't find one. This doesn't really make sense in the context of deploying an application via a container.

A lot of UX thought will need to be put in here, but generally the CLI should be able to run by itself unless maybe a command gets called that requires further configuration?

### Additional context

Suggesting that users run the `init` command if no config file is found also seems helpful, and figuring out a way for user analytics prompt to maybe not run unless they run `init`?

We shouldn't need to run the `create_config` function every single time, as it also outputs logging that is noisy on every command

ThomasLaPiana:
Closes #1661

### Code Changes

* [x] remove the call to create the config from within the main CLI codepath
* [x] only run `check_and_update_analytics_config` when `fides init` is run

### Steps to Confirm

* [ ] head into a fresh directory and run `fides`, confirm no file was written and no `.fides` dir exists
* [ ] run `fides init` and confirm that the flow works, as well as prompting about user analytics

### Pre-Merge Checklist

* Documentation Updated:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

The side-effect of having the `fides.toml` _always_ be created whether it exists or not is a bit redundant and potentially error-prone (see the attached issue for more discussion). As such, the config file should only be created when a user specifically runs the `fides init` command

Due to this change though, we also needed to update how `analytics_opt_out` is handled, given that before we always expected the user to have a file-based configuration file to update. The behaviour is now that we assume analytics is false if there is no config file, but will prompt the user for it when they run `fides init`. This means we defaulting to `opt-out` and are no longer prompting the user every time they use the CLI to update this. I think this is no longer a major issue, as we should be relying more heavily on the `user registration` flow for valuable usage analytics.

### DIFF
--- a/src/fides/cli/utils.py
+++ b/src/fides/cli/utils.py
@@ -59,16 +59,17 @@ FIDES_ASCII_ART = """
 """
 
 
-def check_server_health(server_url: str) -> requests.Response:
+def check_server_health(server_url: str, verbose: bool = True) -> requests.Response:
     """Hit the '/health' endpoint and verify the server is available."""
 
     healthcheck_url = server_url + "/health"
     try:
         health_response = check_response(_api.ping(healthcheck_url))
     except requests.exceptions.ConnectionError:
-        echo_red(
-            f"Connection failed, webserver is unreachable at URL:\n{healthcheck_url}."
-        )
+        if verbose:
+            echo_red(
+                f"Connection failed, webserver is unreachable at URL:\n{healthcheck_url}."
+            )
         raise SystemExit(1)
     return health_response
 
@@ -227,7 +228,7 @@ def check_and_update_analytics_config(ctx: click.Context, config_path: str) -> N
         if ctx.obj["CONFIG"].user.analytics_opt_out is False:
             server_url = ctx.obj["CONFIG"].cli.server_url
             try:
-                check_server_health(server_url)
+                check_server_health(server_url, verbose=False)
                 should_attempt_registration = not is_user_registered(ctx)
             except SystemExit:
                 should_attempt_registration = False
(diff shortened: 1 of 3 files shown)


ThomasLaPiana:
@NevilleS can you take a look at the logic changes here and lmk if you're comfortable with these changes?

Particularly around only doing analytics opt-ins when someone runs `fides init`. My thinking here is that most of that heavy lifting should actually be done by user registration now

ThomasLaPiana:
it looks like there is an error when running `fides init` and not choosing to opt-out of analytics

src/fides/cli/utils.py
```diff
@@ -47,6 +47,8 @@
 from fides.ctl.core.config.utils import get_config_from_file, update_config_file
 from fides.ctl.core.utils import check_response, echo_green, echo_red
 
+APP = fides.__name__
+PACKAGE = "ethyca-fides"
```
SteveDMurphy, line 51: Nice catch on these 👍🏽 do we want to just default to using the `ethyca-fides` package name for now or is there still value in the `fides` name by itself for these logs? Might be able to avoid doing `import fides` if so

SteveDMurphy (approved):
Tested in a local install of the branch and looks sharp! Added one question/comment earlier

Also, these are my favorite type of tongue twisters 😂 
```
    if not no_analytics:
```

src/fides/cli/utils.py
ThomasLaPiana: I think generally using `fides` is more accurate still for various reasons (branding, the actual name of the code module) but good thought here. Definitely something we can revisit if we feel the need in the future :)

PR #2031, opened 2022-12-13
Update all logging references to use loguru

### LINKED ISSUE #1842: Update all `logger.` references to use `loguru` to stop logs getting suppressed
When `fidesops` and `fidesctl` joined, `fidesctl` brought with it the library `loguru`. `loguru` is now suppressing any vanilla `logger.` log output. We should investigate converting all logging to use `loguru` as it's a [more performant library](https://loguru.readthedocs.io/en/stable/resources/migration.html).

adamsachs:
Update to use `loguru` instead of built-in `logging`

Closes #1842

### Code Changes

* Update all logging references to use `loguru`
* Switch string formatting to use {}-style loguru formatting for lazy evaluation
* Update Pii class to work well with loguru lazy formatting
* Include a new, proper `logging.log_pii` config property

### Steps to Confirm

* run your server
* notice (hopefully) all of our logs are back

### Follow-up issues
* https://github.com/ethyca/fides/issues/2036

### Pre-Merge Checklist

* Documentation Updated:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

Most of this is pretty straightforward find-replace. 

The only change that required some thought was the `Pii` updates, since before we'd been creating our own `LogRecordFactory` - there may have been a straightforward way of mimicking this functionality in `loguru`, but it wasn't immediately apparent to me, so i decided to switch up the implementation a bit. What I arrived at should be functionally equivalent, and I think it's quite a straightforward implementation.

I also decided to add in a proper `logging.log_pii`  app config property, since we were just relying on a environment var beforehand outside of the general config framework. perhaps there was a reason for having it done that way? i couldn't see one, so i figured it made more sense to properly account for it as an app config property/setting.

A couple of notes:
- I removed the `get_task_logger` reference in `request_runner_service.py` because I couldn't find a quick way to integrate that with `loguru`. this seems to be some functionality we may want, so i think it probably deserves some further attention. but i didn't want to hold up this PR on that, since we need to get overall logging back ASAP. (note that the logging in this file is not coming through _at all_ on the current `main`, so this still represents an improvement). I can create a follow up ticket to handle this if we think it's necessary.
- I did not update our `logging` references in our tests. I wasn't sure if this was in scope - I could see it going either way. regardless, it's not as high priority as the app code updates, and we want to get this out ASAP, so I decided to hold off. I can also create a follow up ticket for those updates if we want.


.fides/fides.toml
```diff
@@ -11,6 +11,7 @@ app_postgres = {connection_string="postgresql+psycopg2://postgres:fides@fides-db
 
 [logging]
 level = "INFO"
+log_pii = true
```
adamsachs, line 14: decided we should put this as a proper app config setting/property

src/fides/api/ops/service/privacy_request/request_runner_service.py
```diff
@@ -1,13 +1,12 @@
-import logging
 import random
 from datetime import datetime, timedelta
 from typing import Any, Dict, List, Optional, Set, Tuple
 
 import requests
-from celery.utils.log import get_task_logger
```
adamsachs, line 7: i couldn't figure out a quick way to integrate this with `loguru`. i can create a follow up ticket if we think this is important functionality

pattisdr:
Thanks @adamsachs will review shortly!

pattisdr:
@adamsachs just starting to review, but I'm still seeing a lot of instances of where the old built-in `logging` is still being used on this branch

adamsachs:
> @adamsachs just starting to review, but I'm still seeing a lot of instances of where the old built-in `logging` is still being used on this branch

whoops, yeah i think i missed a few. the uses in `tests/` and `scripts/` are expected, fwiw. but yeah, definitely some uses in `src/` that somehow slipped by me! i'll address shortly, sorry bout that.

src/fides/api/ctl/routes/health.py
```diff
@@ -19,21 +19,17 @@
 
 router = APIRouter(tags=["Health"])
 
-logger = logging.getLogger(__name__)
-# stops polluting logs with sqlalchemy / alembic info-level logs
```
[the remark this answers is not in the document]
pattisdr: I think we need to restore this.

This was to suppress sqlalchemy logs - the SQL query being generated by SQL Alchemy can often have PII in it.  When running a privacy request without this, you see this:

```
2022-12-13 17:08:26,881 INFO sqlalchemy.engine.Engine SELECT address_id,email,id,name FROM employee WHERE email = %(email)s
2022-12-13 17:08:26,881 INFO sqlalchemy.engine.Engine [generated in 0.00018s] {'email': ('customer-1@example.com',)}
INFO  [sqlalchemy.engine.Engine] SELECT address_id,email,id,name FROM employee WHERE email = %(email)s
INFO  [sqlalchemy.engine.Engine] [generated in 0.00018s] {'email': ('customer-1@example.com',)}
```

src/fides/api/main.py
```diff
@@ -315,22 +307,22 @@ def read_other_paths(request: Request) -> Response:
 
     # If any of those worked, serve the file.
     if ui_file and ui_file.is_file():
-        log.debug(
+        logger.debug(
             f"catchall request path '{path}' matched static admin UI file: {ui_file}"
         )
```
pattisdr, line 312: Should we reformat all of these f string instances while we're here to postpone the formatting of this error message like you're doing in some of the other logs:  At one point, we'd gotten rid of all of these f strings in the logs on the ops side.

```
logger.info("Finding connection configuration with key '{}'", connection_key)
```

pattisdr (changes requested):
thanks for taking this on Adam -

src/fides/api/main.py
adamsachs: yeah, i saw this and sort of deferred. but this nudge has given me the motivation to do it! i'll go for it.

pattisdr: On the ops side we'd done this https://github.com/ethyca/fidesops/issues/837#issuecomment-1178934542 to find all instances and correct - you could probably flip a rule in pylint to catch all these

adamsachs: not sure that's going to work (as a way to identify the instances) for us now though since we're not using the standard `logging` library? i don't see that line in the `pyproject.toml` and my pylint doesn't complain about anything...maybe i'm missing something though?

that being said, i think i can probably identify the instances with a regex

pattisdr: sounds good you're right I don't see that line here in fides

src/fides/api/ctl/routes/health.py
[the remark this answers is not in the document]
pattisdr: Could we also improve this code comment? - I didn't communicate its necessity with the current one.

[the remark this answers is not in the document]
adamsachs: confirmed that we're seeing these problematic log statements even on `main`, when running `nox -s test_env` and processing a privacy request. so it seems like there's been a regression in `main` some point previously (likely the repo unification?). i'll create a follow up issue to fix that, but @pattisdr and i have decided to not let that hold up this PR.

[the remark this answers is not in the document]
adamsachs: created https://github.com/ethyca/fides/issues/2036 as a follow up

ThomasLaPiana:
I think the big fideslib merge caused these issues, I'm going to take a look

ThomasLaPiana:
@adamsachs I'm overnighting a cape to you, thank you for tackling this!

ThomasLaPiana (approved):

pattisdr:
looking now!

src/fides/api/ctl/routes/health.py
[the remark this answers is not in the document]
pattisdr: thank you!

pattisdr (approved):
thanks so much for taking this on @adamsachs!  I added a last commit I was going through your changes, let me know if that looks ok -

src/fides/api/ctl/routes/health.py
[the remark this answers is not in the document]
adamsachs: ah, definitely seems like as good a place as any to start testing in the follow up

adamsachs:
thank you @patt

> thanks so much for taking this on @adamsachs! I added a last commit I was going through your changes, let me know if that looks ok -

more than ok! thank you for that last bit of cleanup!

i think we're good to go now :)

PR #2042, opened 2022-12-14
2041 update logging references in `tests/` and `scripts/`

### LINKED ISSUE #2041: update `logging` references in `tests/`

https://github.com/ethyca/fides/pull/2031 removed all the `logging` uses in our code in `src/` to use `loguru` instead. it did not update the `logging` uses in our test code, i.e. under `tests/`.

To keep things consistent, we should either update all those `logging` references to use `loguru` instead, or simply remove the log messages in our `tests/` (@seanpreston thinks they may be superfluous and flooding our CI output unnecessarily).

For each `logging` references in `tests/`, either:
- update it to to use `loguru`
- simply remove it

### Additional context

This isn't critical, but may as well avoid having two logging standards across our codebase

adamsachs:
Closes #2041 

### Code Changes

* remove `logging` references from `tests/`
    * pretty much just remove all logging in tests, since we don't think they're useful/necessary and may just flood output
* update `logging` to `loguru` in `scripts/` 

### Steps to Confirm

* make sure all our tests and scripts still work as expected and we haven't lost any useful logging info!

### Pre-Merge Checklist

* Documentation Updated:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

@seanpreston and i thought that logging in tests is generally superfluous and not used, so we made the decision to remove most of the logging from our tests. if people are indeed using this logging, then it should be straightforward to push it back in and use `loguru` instead!


tests/ops/test_helpers/vault_client.py
```diff
@@ -1,13 +1,11 @@
-import logging
 import os
 from typing import Any, Dict, Optional
 
 from hvac import Client
+from loguru import logger
```
adamsachs, line 5: this was the only logging reference i did not get rid of in `tests`. the reason is because [here](https://github.com/ethyca/fides/blob/main/tests/ops/test_helpers/vault_client.py#L44) we actually are catching an exception that occurs on many of our tests, i.e it's an expected state in many of our test runs. we log that exception, but i wasn't seeing it at all in any of my test runs locally, so i'm not sure it was ever getting through. i did switch it to use `loguru` and still did not see the message, but figured i'd keep it here since we do need to catch the exception anyway

adamsachs:
@seanpreston just wanted to bump this for you - not high priority, and not major changes, but cleaning up some tech debt we spoke about, and it will become more difficult to merge as it drifts from `main`. let me know if you want to pass it on to someone else for review!

sanders41 (approved):

PR #2045, opened 2022-12-14
avoid error if access results key is not found for erasures

### LINKED ISSUE #2048: Disabled connectors cause erasure requests to error
### Bug Description

When running an erasure request that traverses through a dataset associated with a disabled connector, that erasure request will appear in an errored state (even if it does successfully process erasures for non-disabled connectors).

The underlying error is a `KeyError` that's being thrown because there is no cache entry for access results for the disabled connector's dataset, and the erasure execution is expecting there to be a cache entry, even if an empty one.

#### Steps to Reproduce

1. get an environment up with `nox -s test_env`
2. disable one or both of the Postgresql and MongoDB connectors that are configured by default
3. try running a sample erasure request. you can use a fake email identity, e.g.:
```
curl --location --request POST 'http://0.0.0.0:8080/api/v1/privacy-request' \
--data-raw '[
    {
        "identity": {"email": "foo@bar.com"},
        "policy_key": "default_erasure_policy"
    }
]'
```
4. navigate and log in to the admin UI and approve the request that's been queued. you should see it quickly move to an `ERROR` state

### Expected behavior

- The privacy request should not error, but should go to a `COMPLETE` state, even if it's not erasing any data

### Environment

`test_env`, or any other fides environment really

### Additional context

In some testing with @Roger-Ethyca and @RobertKeyser, we'd noticed some strange behavior where it seemed like a disabled saas connector was throwing an error depending on what the `MASKING_STRICT` config property was set to. i've been unable to reproduce that specific symptom locally and i can't see any way in which the `MASKING_STRICT` property will actually matter in cases where a connector is disabled.

In doing that investigation, though, I noticed the issue here. It's _possible_ that this is actually what we'd been seeing in our testing and we just weren't keeping track of the symptoms closely enough, but i'm not sure about that.

adamsachs:
Closes #2048 

### Code Changes

* default to an empty list `[]` if, in the beginning of erasure request execution, no access result cache entry is found for the given collection

### Steps to Confirm

* See issue description steps to reproduce, and ensure that requests no longer error in this scenario, but instead go to `COMPLETE`
* Also can confirm that the `Event Details` for the step associated with the disabled connector show all of its events with `skipped` status

### Pre-Merge Checklist

* Documentation Updated:
* [x] Update `CHANGELOG.md`

### Description Of Changes

Passing in the empty list is effectively a no-op here, just a way to prevent an error. This is because the list doesn't actually get used, as the erasure request execution will not proceed because the connector is disabled.

### DIFF
--- a/src/fides/api/ops/task/graph_task.py
+++ b/src/fides/api/ops/task/graph_task.py
@@ -724,13 +724,13 @@ async def run_erasure(  # pylint: disable = too-many-arguments
         dsk: Dict[CollectionAddress, Any] = {
             k: (
                 t.erasure_request,
-                access_request_data[
-                    str(k)
-                ],  # Pass in the results of the access request for this collection
+                access_request_data.get(
+                    str(k), []
+                ),  # Pass in the results of the access request for this collection
                 *[
-                    access_request_data[
-                        str(upstream_key)
-                    ]  # Additionally pass in the original input data we used for the access request. It's helpful in
+                    access_request_data.get(
+                        str(upstream_key), []
+                    )  # Additionally pass in the original input data we used for the access request. It's helpful in
                     # cases like the EmailConnector where the access request doesn't actually retrieve data.
                     for upstream_key in t.input_keys
                 ],
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ The types of changes are:
 ### Fixed
 
 * Timing issues with bulk DSR reprocessing, specifically when analytics are enabled [#2015](https://github.com/ethyca/fides/pull/2015)
+* Error caused by running erasure requests with disabled connectors [#2045](https://github.com/ethyca/fides/pull/2045)
 
 ## [2.2.2](https://github.com/ethyca/fides/compare/2.2.1...2.2.2)
 


src/fides/api/ops/task/graph_task.py
```diff
@@ -726,12 +726,14 @@ def termination_fn(*dependent_values: int) -> Tuple[int, ...]:
...
                 access_request_data[
                     str(k)
-                ],  # Pass in the results of the access request for this collection
-                *[
-                    access_request_data[
-                        str(upstream_key)
-                    ]  # Additionally pass in the original input data we used for the access request. It's helpful in
+                ]  # Pass in the results of the access request for this collection
+                if str(k) in access_request_data
+                else [],  # handle case with no read entry, i.e. disabled connector
```
pattisdr, line 731: Good catch avoiding this key error, could be written as this instead, a get with a default if it isn't found.
```
access_request_data.get(str(upstream_key), [])
```

pattisdr (commented):
nice catch @adamsachs

src/fides/api/ops/task/graph_task.py
adamsachs: ah yeah, that's a lot cleaner! soon enough `.get()` will become instinct for me :)

pattisdr (approved):

adamsachs:
integration test [failure](https://github.com/ethyca/fides/actions/runs/3705511773/jobs/6279655233#step:7:2419) seems like a blip and unrelated to changes, so i'll go ahead and merge if that looks right to you @pattisdr ?

pattisdr:
I agree @adamsachs this is likely unrelated and resolved by re-running!

adamsachs:
CI is being stubborn with the integration tests. the tests are passing locally and there's no connection between the failure reported here and the changed code. Merging rather than fighting this more, i think we're good!

Issue #1764, opened 2022-11-14
Add dataset migration CLI command


TheAndrewJackson:
❗ Depends on #1763 ❗ 

Now that the `dataset` column on `datasetconfig` is no longer being used users need a way to merge all of the datasets they have stored in `datasetconfig` into `ctl_datasets`.

A high level idea of what this flow would like is:
- A new `dataset-migration` command is added to the cli
- When a user runs that command the `datasetconfig` table is scanned and every row that contains config data stored on the `dataset` column is converted into a yaml file and merged into the corresponding `ctl_datasets` config yaml file locally on the users machine based on the `fides_key`. 
- The user would then look at the git diffs and manually merge them together. 
- Once everything looks good they can check the merged files into source control and push the changes up to the fides server

This flow would also handle renaming the `fidesops_meta` to `fides_meta`. It would also need a way to mark the datasets that have been converted as such so they aren't scanned and merged again. 

This command won't delete any data. It's just simplifying the merge process. The legacy data will stored be stored for now but not accessible through the api.

TheAndrewJackson:
@ThomasLaPiana Does this match what you had in mind when this was discussed in the design doc review?

ThomasLaPiana:
@TheAndrewJackson yep!

Still some tricky implementation details here, but those can wait for the PR :)

seanpreston:
Whether we action this ticket will depend on the following question:

Are there live Fides deployments where the same dataset exists between both tables, such that they might conflict?

pattisdr:
https://github.com/ethyca/fides/pull/2046 solves this in a simpler way by adding an automated data migration that copies over all DatasetConfig.datasets into the ctl_dataset table, creating new records for each.  If there is a fides_key conflict, the migration will error and the user should manually resolve.

Issue #2016, opened 2022-12-09
Ops-side Tests Clear DB Tables: Removes Resources Loaded by Default 


pattisdr:
### Bug Description

Ops tests can't rely on resources that are automatically added to the database by the application on startup because the database is intentionally wiped before the tests run.

We have a handy `clear_db_tables` "autouse" fixture that wipes the database in between tests to prevent any resources created by tests from sticking around.  However, now that ops and ctl code have merged, there are several additional resources that are automatically added to the db on startup in `load_default_resources`: `load_default_organization`, `load_default_taxonomy`, and `load_default_dsr_policies`.

This came up here: https://github.com/ethyca/fides/pull/2009/files#r1044804685

I was adjusting data category validation on the ops-side to compare against data categories in the database instead of the static default taxonomy list.  But in tests, there are no data categories in the database ops-side. 

#### Steps to Reproduce

1. Bring the shell up. Verify ctl_data_categories table is populated
2. Run a test in pytest - verify the ctl_data_categories table gets wiped.

### Expected behavior

It would be useful if these resources were not removed - or if they were, then we re-ran `load_default_resources`. Removing this fixture altogether does cause several tests to fail where we're not cleaning up resources in tests.

### Screenshots

_If applicable, add screenshots to help explain your problem._

### Environment

* Version:
* OS:
* Python Version:
* Docker Version:

### Additional context

_Add any other context about the problem here._

Issue #1763, opened 2022-11-14
Remove dataset yaml storing code from `datasetconfig` endpoint


TheAndrewJackson:
❗ Depends on #1762 ❗ 

Now that dataset yaml configs will only be stored on `ctl_datasets` all of the code that handles storing them on `datasetconfig` needs to be removed.

Acceptance criteria
- Remove legacy code from `datasetconfig` endpoint that stored data on the `dataset`column
- Port all of the code that handled validating `fidesops_meta` and saas configs outside of pydantic to the `ctl_datasets` endpoint.
- Dataset and saas config endpoints still work as expected

pattisdr:
Also make sure ctl_dataset endpoints are validating data categories -

PR #2046, opened 2022-12-14
Add Non Nullable DatasetConfig.ctl_dataset_id Field

### LINKED ISSUE #1762: Add `ctl_datasets` foreign key to `datasetconfig` table
❗ Depends on #1761 ❗ 

Add `ctl_datasets` foreign key to `datasetconfig` table. This change will allow the `dataset` column on `datasetconfig` to be deprecated and for datasets to be stored in one place.

This ticket is meant to be an incremental change. The code that handles storing datasets on `datasetconfig` will be removed in a future ticket.

Acceptance criteria:
- `ctl_datasets` foreign key is added to `datasetconfig` table
- Only valid dataset configurations can be linked. This consists of:
  - Database `ConnectionConfig`s can have multiple `DatasetConfig`s
  - SAAS `ConnectionConfig`s can only have one `DatasetConfig`
  - A `DatasetConfig` is only linked to one `ConnectionConfig` at a time
- Update the `datasetconfig` endpoint to accept the new `ctl_datasets` foreign key

pattisdr:
❗ Contains multiple migrations (schema and data)
👉  Note this PR is against a feature branch

Closes #1762, https://github.com/ethyca/fides/issues/1764

### Code Changes

* [ ] Added a non-nullable `datasetconfig.ctl_dataset_id` column
* [ ] Added a schema and data migration.  This is an attempt to also resolve https://github.com/ethyca/fides/issues/1764
    * [ ] First, create a `datasetconfig.ctl_dataset_id` column that is nullable
    * [ ] Then, attempt to copy the contents of DatasetConfig.dataset into new `ctl_dataset` records, and then link that `ctl_dataset` as the DatasetConfig.ctl_dataset_id.  If there's a conflict with an existing `ctl_datasets.fides_key`, I error instead of attempting to upsert.  The user should manually resolve.
    * [ ] Adds a follow-up schema migration to then make the `datasetconfig.ctl_dataset_id` field non-nullable
* [ ] Added a new API endpoint `PATCH {{host}}/connection/{{connection_key}}/datasetconfig` that takes in a `fides_key` (for the DatasetConfig) and an existing `ctl_dataset_fides_key`. It upserts a DatasetConfig, links to the existing CTLDataset, then copies the CTLDataset contents back to the DatasetConfig.dataset.  Soon, the DatasetConfig.dataset field is going away.
* [ ] Updated existing PATCH dataset config JSON and YAML variants to temporarily still work. A raw dataset passed in will attempt to upsert both the DatasetConfig and CTLDataset object.  I didn't want the UI to be broken on this feature branch.
    * [ ]  Creating a saas connector from a template still works. The dataset in the template currently upserts the DatasetConfig and CTL Dataset record.
* [ ] In the locations where we retrieve a DatasetConfig.dataset (such as building a graph), return the ctl_dataset record instead of DatasetConfig.dataset. 
* [ ] Lots of test fixtures needed to be changed to create a ctl_dataset before creating a datasetconfig and then linking the two.

### Steps to Confirm

Migration verification
* [ ] Checkout main.  Create a `datasetconfig` in your application database and then switch to this branch without dropping that database.  If you switch and bring up the shell, migrations should run.  Verify that the existing `datasetconfig` table has a `ctl_dataset_id` row.  Verify that this row is non-nullable and populated.  Locate the corresponding `ctl_datasets` record. Compare each column and match to what was in `datasetconfig.dataset`.  `DatasetConfig.dataset` itself should be untouched for now.  Any `fidesops_meta` fields are likely converted to `fides_meta` fields.'

Run a privacy request
* [ ] Run `nox -s test_env`
* [ ] Go to the privacy center, run a privacy request
* [ ] Return to admin UI, approve
* [ ] Verify json file in `fides_uploads` folder and contents are roughly correct.

Existing endpoint parity (will soon be deprecated, but the UI still works here)
* [ ] Run `nox -s test_env`
* [ ] Go to http://localhost:3000, login, then http://localhost:3000/datastore-connection. 
* [ ] Click on the three horizontal dots of the Postgres connector > Configure. Jump to Dataset configuration tab. Edit the description of the dataset. Verify API request is successful (`PATCH http://0.0.0.0:8080/api/v1/connection/postgres_connector/dataset`)
* [ ] Locate the updated datasetconfig.dataset field in the db. Verify the description has changed. Similarly note that the ctl_dataset.description column has changed (`select description from ctl_datasets where id = 'xxxxxx';`)

Test creating saas connectors from a template
* [ ] - http://localhost:3000/datastore-connection > Create new connection
* [ ] Select mailchimp. Enter in connector parameters - because we're not going to use this, your secrets can be fake
* [ ] On Dataset configuration tab click > Save Yaml System.
* [ ] Verify successful API requests
* [ ] Verify that new `datasetconfig` and `ctl_datasets` records and the DatasetConfig has a ctl_dataset_id FK. The dataset should exist in both places.
 
New endpoint
* [ ] In Postman or similar, create a new connection config resource `PATCH {{host}}/connection/`
* [ ] In Postman or similar, add a dataset config with the new endpoint: `PATCH {{host}}/connection/{{existing connection key}}/datasetconfig`.  Create a new fides_key (this will be the identifier for the DatasetConfig, but select an existing ctl_dataset fides_key.
```json
[{
    "fides_key": "new_dataset_config",
    "ctl_dataset_fides_key": "postgres_example_test_dataset"
}]
```
* [ ] Verify a new DatasetConfig was created and the contents of the existing ctl_dataset were ported back into the new DatasetConfig.dataset .
* [ ] Visit this new connector and dataset in the UI

Existing CTL datasets tab
* [ ] Go to http://localhost:3000/dataset. Our newly created datasets (mailchimp) and the postgres one created directly in the API should be there.

### Pre-Merge Checklist

* Documentation Updated:
* [x] Update `CHANGELOG.md`

### Description Of Changes

We are trying to move to storing the bulk of the contents of a Dataset solely on the `ctl_datasets` table. Right now, similar concepts exist in both the `ctl_datasets` table and the `DatasetConfig.dataset` column.  

The idea with this increment is to add a _non-nullable_ `DatasetConfig.ctl_dataset_id` field. DSR's can't run without an associated dataset so I think we should keep this a constraint from the beginning.  I take the contents of existing `DatasetConfig.datasets` and attempt to create new `ctl_dataset` records and then link them to existing DatasetConfig. 

The next step is to keep writing to both places - `DatasetConfig.dataset` AND making the same changes to the `ctl_dataset` record.   This work also starts _reading_ from the `DatasetConfig.ctl_dataset` record instead of `DatasetConfig.dataset`.  

Follow-up work will deprecate some existing endpoints and stop writing to the DatasetConfig.dataset column.

### DIFF
--- a/src/fides/api/ctl/sql_models.py
+++ b/src/fides/api/ctl/sql_models.py
@@ -6,6 +6,7 @@ Contains all of the SqlAlchemy models for the Fides resources.
 
 from typing import Dict
 
+from fideslang.models import Dataset as FideslangDataset
 from sqlalchemy import (
     ARRAY,
     BOOLEAN,
@@ -19,6 +20,7 @@ from sqlalchemy import (
     type_coerce,
 )
 from sqlalchemy.dialects.postgresql import BYTEA
+from sqlalchemy.orm import Session
 from sqlalchemy.sql import func
 from sqlalchemy.sql.sqltypes import DateTime
 
@@ -199,6 +201,16 @@ class Dataset(Base, FidesBase):
     joint_controller = Column(PGEncryptedString, nullable=True)
     retention = Column(String)
     third_country_transfers = Column(ARRAY(String))
+
+    @classmethod
+    def create_from_dataset_dict(cls, db: Session, dataset: dict) -> "Dataset":
+        """Add a method to create directly using a synchronous session"""
+        validated_dataset: FideslangDataset = FideslangDataset(**dataset)
+        ctl_dataset = cls(**validated_dataset.dict())
+        db.add(ctl_dataset)
+        db.commit()
+        db.refresh(ctl_dataset)
+        return ctl_dataset
 
 
 # Evaluation
--- a/src/fides/api/ops/api/v1/endpoints/dataset_endpoints.py
+++ b/src/fides/api/ops/api/v1/endpoints/dataset_endpoints.py
@@ -182,13 +182,12 @@ def patch_dataset_configs(
     """
     created_or_updated: List[Dataset] = []
     failed: List[BulkUpdateFailed] = []
-    logger.info("Starting bulk upsert for %s Dataset Configs", len(dataset_pairs))
+    logger.info("Starting bulk upsert for {} Dataset Configs", len(dataset_pairs))
 
     for dataset_pair in dataset_pairs:
         logger.info(
-            "Finding ctl_dataset with key '%s'", dataset_pair.ctl_dataset_fides_key
+            "Finding ctl_dataset with key '{}'", dataset_pair.ctl_dataset_fides_key
         )
-
         ctl_dataset: CtlDataset = (
             db.query(CtlDataset)
             .filter_by(fides_key=dataset_pair.ctl_dataset_fides_key)
@@ -199,7 +198,7 @@ def patch_dataset_configs(
         data = {
             "connection_config_id": connection_config.id,
             "fides_key": dataset_pair.fides_key,
-            "dataset": fetched_dataset.dict(),  # TODO Unified Fides Resources: Remove writing to this field.
+            "dataset": fetched_dataset.dict(),
             "ctl_dataset_id": ctl_dataset.id,
         }
 
@@ -266,7 +265,7 @@ def patch_datasets(
             dataset,
             db,
             failed,
-            DatasetConfig.create_or_update_with_ctl_dataset,
+            DatasetConfig.upsert_with_ctl_dataset,
         )
     return BulkPutDataset(
         succeeded=created_or_updated,
@@ -320,7 +319,7 @@ async def patch_yaml_datasets(
                 Dataset(**dataset),
                 db,
                 failed,
-                DatasetConfig.create_or_update_with_ctl_dataset,
+                DatasetConfig.upsert_with_ctl_dataset,
             )
     return BulkPutDataset(
         succeeded=created_or_updated,
@@ -342,7 +341,7 @@ def create_or_update_dataset(
             _validate_saas_dataset(connection_config, dataset)  # type: ignore
         # Try to find an existing DatasetConfig matching the given connection & key
         dataset_config = create_method(db, data=data)
-        created_or_updated.append(dataset_config.dataset)
+        created_or_updated.append(dataset_config.ctl_dataset)
     except (
         SaaSConfigNotFoundException,
         ValidationError,
--- a/src/fides/api/ops/api/v1/endpoints/dataset_endpoints.py
+++ b/src/fides/api/ops/api/v1/endpoints/dataset_endpoints.py
@@ -1,4 +1,4 @@
-from typing import List
+from typing import Callable, List
 
 import yaml
 from fastapi import Depends, HTTPException, Request
@@ -22,6 +22,7 @@ from starlette.status import (
     HTTP_422_UNPROCESSABLE_ENTITY,
 )
 
+from fides.api.ctl.sql_models import Dataset as CtlDataset  # type: ignore[attr-defined]
 from fides.api.ops.api import deps
 from fides.api.ops.api.v1.scope_registry import (
     DATASET_CREATE_OR_UPDATE,
@@ -30,6 +31,7 @@ from fides.api.ops.api.v1.scope_registry import (
 )
 from fides.api.ops.api.v1.urn_registry import (
     DATASET_BY_KEY,
+    DATASET_CONFIGS,
     DATASET_VALIDATE,
     DATASETS,
     V1_URL_PREFIX,
@@ -51,6 +53,7 @@ from fides.api.ops.models.datasetconfig import (
 from fides.api.ops.schemas.api import BulkUpdateFailed
 from fides.api.ops.schemas.dataset import (
     BulkPutDataset,
+    DatasetConfigCtlDataset,
     DatasetTraversalDetails,
     ValidateDatasetResponse,
     validate_data_categories_against_db,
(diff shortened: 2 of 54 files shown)


src/fides/api/ops/api/v1/endpoints/dataset_endpoints.py
```diff
@@ -198,7 +259,13 @@ def patch_datasets(
             "dataset": dataset.dict(),
```
pattisdr, line 259: Validation on this endpoint makes sure data categories on the dataset exist in the database.  Because we're accessing the database, it's done outside of a typical pydantic validator.  If this endpoint goes away, we need a new place to stick this. Does the existing ctl_datasets endpoint have this validation?

src/fides/api/ops/api/v1/endpoints/dataset_endpoints.py
```diff
@@ -172,10 +233,10 @@ def patch_datasets(
     Given a list of dataset elements, create or update corresponding Dataset objects
```
pattisdr, line 233: I believe this endpoint `{{host}}/connection/{{connection_key}}/dataset` should be deprecated once the UI has been updated to use the new endpoint above. 

Added some functionality here to make it still usable.  If a raw dataset is passed in, I write it to both the DatasetConfig.dataset field and the ctl_dataset record.

src/fides/api/ops/models/datasetconfig.py
```diff
@@ -40,12 +41,84 @@ class DatasetConfig(Base):
...
+            conditions=(
+                (DatasetConfig.connection_config_id == data["connection_config_id"])
+                & (DatasetConfig.fides_key == data["fides_key"])
+            ),
+        ).first()
+
+        if dataset:
+            upsert_ctl_dataset(
+                dataset.ctl_dataset
+            )  # Update existing ctl_dataset first.
```
pattisdr, line 106: Here, I know the specific ctl_dataset_id because I got it off the existing DatasetConfig. However, if there is no dataset config, I'm looking up a ctl dataset by fides key.

So there's a little extra code here, sometimes I want to update the CTLDataset by id, others by fides_key.

pattisdr:
Test failures are just timescale related that we're seeing on other branches ^

tests/ops/models/test_datasetconfig.py
```diff
@@ -200,41 +205,214 @@ def test_validate_dataset_reference_invalid(db: Session, dataset_config: Dataset
...
+
+        dataset = dataset_config.dataset
+        assert dataset["description"] == postgres_dataset["description"]
+        assert (
+            dataset.get("organization_fides_key") is None
+        ), "Existing behavior, DatasetConfig.dataset not validated by Fideslang Dataset first"
+        assert dataset["collections"] is not None
+
+        dataset_config.delete(db)
+        ctl_dataset.delete(db)
```
TheAndrewJackson, line 284: Are these required? IIRC we have a fixture that runs after every test that clears out all of the tables.

https://github.com/ethyca/fides/blob/e77d6dc6fc7e56b269b77bf94491fa41ad7678a7/tests/ops/conftest.py#L107-L132

tests/ops/fixtures/postgres_fixtures.py
```diff
@@ -40,16 +41,21 @@ def postgres_example_test_dataset_config(
...
+
+    ctl_dataset = CtlDataset.create_from_dataset_dict(db, postgres_dataset)
+
     dataset = DatasetConfig.create(
         db=db,
         data={
             "connection_config_id": connection_config.id,
             "fides_key": fides_key,
-            "dataset": postgres_dataset,
+            "dataset": postgres_dataset,  # TODO Unified Fides Resources - Stop writing to this field
```
TheAndrewJackson, line 52: Will this TODO be addressed in a future PR?

src/fides/api/ops/api/v1/endpoints/dataset_endpoints.py
```diff
@@ -263,16 +331,17 @@ def create_or_update_dataset(
...
     data: dict,
-    dataset: dict,
+    dataset: Dataset,
     db: Session,
     failed: List[BulkUpdateFailed],
+    create_method: Callable,
 ) -> None:
     try:
         if connection_config.connection_type == ConnectionType.saas:
             _validate_saas_dataset(connection_config, dataset)  # type: ignore
```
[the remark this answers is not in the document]
TheAndrewJackson: Does it makes sense to add a comment explaining that this validation only needs to happen here and not in `ctl_datasets`?

src/fides/api/ops/api/v1/endpoints/dataset_endpoints.py
TheAndrewJackson: CC: @ThomasLaPiana. You might be the right person to ask about the ctl side of this

src/fides/api/ops/api/v1/endpoints/dataset_endpoints.py
TheAndrewJackson: Should there be a follow up ticket keeping track of all of the soon to be removed/deprecated routes?

src/fides/api/ctl/sql_models.py
```diff
@@ -200,6 +202,16 @@ class Dataset(Base, FidesBase):
...
 
+    @classmethod
+    def create_from_dataset_dict(cls, db: Session, dataset: dict) -> "Dataset":
+        """Add a method to create directly using a synchronous session"""
+        validated_dataset: FideslangDataset = FideslangDataset(**dataset)
+        ctl_dataset = cls(**validated_dataset.dict())
+        db.add(ctl_dataset)
+        db.commit()
+        db.refresh(ctl_dataset)
+        return ctl_dataset
```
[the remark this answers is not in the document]
TheAndrewJackson: This makes sense. IIRC the ctl endpoints are fairly generic and constructed differently

tests/ops/api/v1/endpoints/test_dataset_endpoints.py
```diff
@@ -412,6 +416,413 @@ def test_put_validate_dataset(
...
+    ):
+
+        old_ctl_dataset_id = dataset_config.ctl_dataset.id
+        assert dataset_config.ctl_dataset == ctl_dataset
+        updated = dataset_config.updated_at
+
+        db.expunge(ctl_dataset)
+        make_transient(
+            ctl_dataset
+        )  # http://docs.sqlalchemy.org/en/rel_1_1/orm/session_api.html#sqlalchemy.orm.session.make_transient
```
TheAndrewJackson, line 597: Interesting, I've never seen transient objects before. That's a cool way to reuse the object

src/fides/api/ops/api/v1/endpoints/dataset_endpoints.py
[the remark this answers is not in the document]
pattisdr: good call

tests/ops/models/test_datasetconfig.py
pattisdr: Not technically, but I think this fixture that clears all the tables is too aggressive because there are resources that are expected to be in the database. 

 I filed an issue to investigate this further and in the meantime trying to make my new tests more self-sufficient generally https://github.com/ethyca/fides/issues/2016

src/fides/api/ops/api/v1/endpoints/dataset_endpoints.py
pattisdr: Good question Andrew, here's the follow-up ticket, we can wait to deprecate until the UI has been pointed to use the new endpoints https://github.com/ethyca/fides/issues/2092

src/fides/api/ops/api/v1/endpoints/dataset_endpoints.py
pattisdr: OK so it looks like the existing crud endpoints don't do this.  The tricky bit is that the crud endpoints are very generic, it's blocks of code that applies to updating an entire set of resources.  Added a note to look into the best place to put this in the next ticket https://github.com/ethyca/fides/issues/1763

pattisdr:
Failing tests are still timescale-related that have been fixed on main.  This will be resolved after this is merged and I get `unified-fides-resources` up-to-date with main.

TheAndrewJackson (approved):
:shipit:

ThomasLaPiana:
im taking a look through this now, will take me a bit

PR #2057, opened 2022-12-15
Add log before raising error

### LINKED ISSUE #2054: Fix SlowAPI support for Redis SSL connections to Elasticache
### Bug Description

When using SlowAPI with Redis as a backend it is currently not working when we are using an SSL connection to Elasticache. All endpoints are responding with a 500 server error.

To fix:
1. Disable SlowAPI by default.
1. Trouble shoot how SlowAPI connects to Redis using SSL and test against Elasticache specifically.
1. Once the issue is identified and fixed re-enable SlowAPI.

#### Steps to Reproduce

1. Set the redis config in Fides to use Elasticache with ssl
1. Start the Fides server
1. On startup the server will report it is able to connect to the cache, but requests that go through SlowAPI will return a 500 error.

### Expected behavior

Endpoints should be reachable.

### Screenshots

### Environment

* Version:
* OS:
* Python Version:
* Docker Version:

### Additional context

sanders41:
Partially fixes #2054

### Code Changes

* Add a log before raising the exception in the logging middleware

### Steps to Confirm

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

### DIFF
--- a/src/fides/api/main.py
+++ b/src/fides/api/main.py
@@ -52,6 +52,7 @@ from fides.api.ops.service.connectors.saas.connector_registry_service import (
 )
 from fides.api.ops.tasks.scheduled.scheduler import scheduler
 from fides.api.ops.util.cache import get_cache
+from fides.api.ops.util.logger import Pii
 from fides.ctl.core.config import FidesConfig, get_config
 from fides.ctl.core.config.helpers import check_required_webserver_config_values
 from fides.lib.oauth.api.routes.user_endpoints import router as user_router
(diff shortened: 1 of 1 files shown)


adamsachs:
merging this now because it's critical we get it into `main` ASAP. a couple of points we may want to follow up on:

- i'm noticing that calls to our health endpoint aren't going to be caught and logged here because we pass them along before getting to the `try` block - https://github.com/ethyca/fides/pull/2057/files#diff-26e95484ee5605f0b1cdfa9dee5e94b082bb7340e1754dbd8b5b1a805c88f633L94
- relatedly, i think we should dig into why our raised exception was not making its way to the logs _anywhere_ before this fix. my initial thinking is that maybe we should be including another FastAPI exception handler into our router to generically catch otherwise uncaught exceptions and report them more properly. i've got no idea if that's the missing piece here, but it's a thought that we should be able to (in)validate pretty quickly.
    - whether it's what i'm suggesting or something else, ensuring we've got a good generic exception handler for at least some minimal logging of uncaught exceptions feels like a must-have.

PR #2082, opened 2022-12-19
Fix user creation as root

### LINKED ISSUE #2077: Creating a user with the root user returns a 403
### Bug Description

Cannot create additional users as the root user

#### Steps to Reproduce

1. nox -s test_env
1. Login with root user instead (see toml for creds)
1. Go to user management
1. Try to create a user

### Expected behavior

A user should be created normally

ThomasLaPiana:
Closes #2077 

### Code Changes

* [x] add a test to catch the 403 caused by creating a user as root
* [x] add the fix

### Steps to Confirm

* [ ] Run `nox -s dev`
* [ ] Login as the root user
* [ ] Create a user, assert the user is created
* [ ] Edit a user, assert the user's data is updated
* [ ] Login as the created user, assert that the user is able to act in some way

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

This regression was caused by the removal of the `fideslib` references from our codebase. `fideslib` was designed to have a common override pattern (with safe backstop) for the `verify_oauth_client` API dependency — which is not being overridden correctly now that the code is in the main codebase. This PR switches out the version of that method designed to be overloaded with the current method in use by the rest of the codebase.

### DIFF
--- a/src/fides/lib/oauth/api/routes/user_endpoints.py
+++ b/src/fides/lib/oauth/api/routes/user_endpoints.py
@@ -17,12 +17,12 @@ from starlette.status import (
     HTTP_404_NOT_FOUND,
 )
 
+from fides.api.ops.util.oauth_util import verify_oauth_client
 from fides.ctl.core.config import FidesConfig, get_config
 from fides.lib.models.client import ADMIN_UI_ROOT, ClientDetail
 from fides.lib.models.fides_user import FidesUser
 from fides.lib.models.fides_user_permissions import FidesUserPermissions
 from fides.lib.oauth.api import urn_registry as urls
-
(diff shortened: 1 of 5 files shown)


ThomasLaPiana:
@sanders41 I believe I broke this with the `fideslib` merge. What I _think_ might be happening here is that the root users isn't getting created/seeded, but I'm not sure

adamsachs (approved):
looks good to me, and i verified that things are working now on `nox -s test_env` just to be sure. thanks for the updated PR description to explain the changes!

sanders41 (approved):

PR #2086, opened 2022-12-19
pc/consent: Include default consent values in through build-time configuration

### LINKED ISSUE #2065: Set default values for consent from Fides privacy center config
We need to set consent options on a browser with a default state. Since the privacy center configuration is very customizable w/r/t the cookie keys that we create for each use case, when the privacy center is built, we should build the fides-consent.js script to set the default values of consent, so that a new user has the appropriate consent set the first time they visit.

Details 
* You will have to use the privacy center and build it for this to work.

**AC**
- When the privacy center is built then the fides-consent.js script sets defaults based on the privacy center config.

ssangervasi:
Closes #2065 

### Code Changes
- Include default consent values in through build-time configuration.
- Update Readme to describe consent

### Steps to Confirm

1. Build and serve the Privacy Center.
2. In a private window, navigate to `/fides-consent-demo.html` on your PC domain (localhost:3000)
3. The default config should show that `data_sales` is opted-in by default.
4. Customize some `cookieKeys` and their default values.
5. Rebuild the PC
6. The defaults should be reflected.

For testing real-ish domains, I recommend updating `/etc/hosts`. Ex:
```
127.0.0.1 ethyca.example.com
127.0.0.1 sub.ethyca.example.com
```

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

This change introduces a config file, `consent-config.json`, to the fides-consent package. This file is auto-generated, derived from the Privacy Center's config at build time. Note that we don't want to bring in the _whole_ PC config because that would mostly be useless and it would balloon the size of the script that gets inserted into external pages.


NevilleS:
Hey @ssangervasi - this makes sense but how do we ensure that this `fides-consent.js` doesn't get permanently cached in someone's Cloudflare/Cloudfront/etc.?

I'd want either something like:
* setting `ETag` to a hash or something
* changing the filename...?

Basically I think we need to help users avoid a situation where they do the right thing (cache an asset like this!) but don't have an easy way to do cache-busting

NevilleS (changes requested):
Hey @ssangervasi - this makes sense but how do we ensure that this `fides-consent.js` doesn't get permanently cached in someone's Cloudflare/Cloudfront/etc.?

I'd want either something like:
* setting `ETag` to a hash or something
* changing the filename...?

Basically I think we need to help users avoid a situation where they do the right thing (cache an asset like this!) but don't have an easy way to do cache-busting

ssangervasi:
`fides-consent.js` is served under `public/` which Next does not cache:
https://github.com/vercel/next.js/issues/1791
A feature, in our case.

NevilleS:
OK, we'll need to provide some clear documentation on how this works so users don't footgun themselves here - some kind of cache is going to be absolutely necessary for `fides-consent.js` at scale.

A couple more useful links from the NextJs docs include:
* Production Caching: https://nextjs.org/docs/going-to-production#caching
* How `/public` folder works for static files: https://nextjs.org/docs/basic-features/static-file-serving 
* Note about `cache-control` header for other routes: https://nextjs.org/docs/api-reference/next.config.js/headers

NevilleS (approved):
OK, I definitely went down the rabbit-hole here, but given that there are definitely some options for how you can configure a cache to work for this, let's not conflate cache performance with the useful feature of consent defaults! I created https://github.com/ethyca/fides/issues/2130 as a follow-up issue for what I noticed here (a lack of caching) and that's good enough for me!

NevilleS:
See https://github.com/ethyca/fides/issues/2130 for the separate caching issue. Thanks @ssangervasi 👍

PR #2089, opened 2022-12-19
specifically handle and report decryption errors on server bootup

### LINKED ISSUE #2088: better handle a change to the `app_encryption_key`

When doing some testing on deployed instances, we put the server into a state where it was throwing `500`s when trying to retrieve all DSR policies (`GET /api/v1/dsr/policy`) or when specifically trying to `GET` the `default_erasure_policy`. We were able to reproduce similar behavior locally when intentionally changing the `fides.security.app_encryption_key` config property and restarting the application (being sure _not_ to clear the database).

Although changing app encryption keys is not a straightforward feature to support, and certainly not a _standard_ workflow, we should at a minimum report this error more clearly - previously, we'd seen only a small `InvalidCiphertextError` in the logs on server spin up.

https://github.com/ethyca/fides/issues/1842 has helped here - when I've reproduced the problem on current `main`, I do get some more information on the `500`s as they're returned. But I think there's still some room for improvement on the error handling on server startup, to try to warn people specifically about a key change, before they get too far. Additionally, I wonder if we should look to enhance our documentation to make it clear that the app encryption key cannot easily change without resetting the DB, and perhaps a follow-up issue to outline a key rotation process, in cases where that's necessary?

- clear error message on server bootup that indicates the issues decrypting and to look at the `app_encryption_key`
- update docs to warn against any changes to the `app_encryption_key` config property (?)
- follow-up issue to test and document a process for properly rotating an `app_encryption_key` (?)

adamsachs:
Closes #2088 

### Code Changes

* Better handle decryption errors on server bootup to prevent people from getting too far with a bad encryption key

### Steps to Confirm

* run `nox -s dev `
* update your `fides.toml` to change a character in the `app_encryption_key` property
* trigger a restart of your server by trivially changing a source code file, by re-running `nox -s dev`, etc.
* look for the new logging around the decryption problems on startup

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

See the issue for some ideas around _further_ guardrails/handling of this. I'm a bit unsure about how far we should go, and how important it is. But this change here seems like some trivial low hanging fruit that will hopefully make things a bit smoother if this happens to come up in the wild.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ The types of changes are:
 ### Added
 
 * Nav redesign with sidebar groups. Feature flagged to only be visible in dev mode until release. [#2030](https://github.com/ethyca/fides/pull/2047)
+* Improved error handling for incorrect app encryption key [#2089](https://github.com/ethyca/fides/pull/2089)
 
 ### Fixed
 
--- a/src/fides/api/ctl/database/database.py
+++ b/src/fides/api/ctl/database/database.py
@@ -9,6 +9,7 @@ from alembic.runtime import migration
 from loguru import logger as log
 from sqlalchemy.orm import Session
 from sqlalchemy_utils.functions import create_database, database_exists
+from sqlalchemy_utils.types.encrypted.encrypted_type import InvalidCiphertextError
 
 from fides.api.ctl.utils.errors import get_full_exception_name
 from fides.ctl.core.config import get_config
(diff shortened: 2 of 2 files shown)


sanders41 (approved):
LGTM. Ultimately it would be good to be able to rotate the key, but that will be a much bigger lift and this is a good improvement in the meantime to let people know what is happening.

Issue #1853, opened 2022-11-23
Support SMS-based ID verification for consent


mfbrown:
As an end user of a service, I want verify my identity when opting out of some consent cases via my phone, so that I can still verify my identity when email isn't supported or available.

**Details**
- The privacy center configuration doesn't seem to support a config option which allows and requires SMS verification

**AC**
- Given that I am configuring the privacy center, when I set up consent then I can choose to require an identity input of either email or phone
- When I choose to make consent changes, when phone is required as the identity input, then the input indicates that the field is required
- Given that I input a phone number to verify my identity when managing consent, when I submit, then I am sent a verification code via SMS to the submitted phone number
- Given that I input a phone number to verify my identity when managing consent, when I submit, then I am shown an input to submit my verification code

mfbrown:
@eastandwestwind says this might just be a front end change, but still confirming.

Right now these are different components on the FE, but we can see what can be shared.

eastandwestwind:
Testing steps:

* [ ] In `fides.toml` set `subject_identity_verification_required = true` and `notification_service_type = "twilio_text"`
* [ ] `nox -s dev`
* [ ] Add twilio sms config/secret using postman (follow https://docs.ethyca.com/fides/dsr_quickstart/dsr_support/messaging#twilio to set up your twilio acct)
* [ ] Ensure your phone number has been added to the twilio account's verified phone numbers (https://console.twilio.com/us1/develop/phone-numbers/manage/verified)
* [ ] In a new terminal, `cd` into `clients/privacy-center`, run `npm install`, `npm build`, then `npm run dev`
* [ ] Submit consent request, enter verification code sent to phone
* [ ] Please also regression test email id verification for consent

PR #2094, opened 2022-12-20
1853 id verification consent sms

eastandwestwind:
Closes https://github.com/ethyca/fides/issues/1853

### Code Changes

* [ ] Adjust back-end to support both email and SMS id verification for consent
* [ ] Adds SMS input field to privacy center
* [ ] Adds unit tests

### Steps to Confirm

* [ ] In `fides.toml` set `subject_identity_verification_required = true` and `notification_service_type = "twilio_text"`
* [ ] `nox -s dev`
* [ ] Add twilio sms config/secret using postman (follow https://docs.ethyca.com/fides/dsr_quickstart/dsr_support/messaging#twilio to set up your twilio acct)
* [ ] Ensure your phone number has been added to the twilio account's verified phone numbers (https://console.twilio.com/us1/develop/phone-numbers/manage/verified)
* [ ] In a new terminal, `cd` into `clients/privacy-center`, run `npm install`, `npm run build`, then `npm run dev`
* [ ] Submit consent request, enter verification code sent to phone
* [ ] Please also regression test email id verification for consent

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### Description Of Changes

### DIFF
--- a/src/fides/api/ops/api/v1/endpoints/consent_request_endpoints.py
+++ b/src/fides/api/ops/api/v1/endpoints/consent_request_endpoints.py
@@ -11,8 +11,8 @@ from starlette.status import (
     HTTP_400_BAD_REQUEST,
     HTTP_403_FORBIDDEN,
     HTTP_404_NOT_FOUND,
-    HTTP_500_INTERNAL_SERVER_ERROR,
     HTTP_422_UNPROCESSABLE_ENTITY,
+    HTTP_500_INTERNAL_SERVER_ERROR,
 )
 
 from fides.api.ops.api.deps import get_db
--- a/src/fides/api/ops/api/v1/endpoints/consent_request_endpoints.py
+++ b/src/fides/api/ops/api/v1/endpoints/consent_request_endpoints.py
@@ -296,7 +296,9 @@ def _get_or_create_provided_identity(
     identity_data: Identity,
     identity_type: str,
 ) -> ProvidedIdentity:
-    """Based on desired identity type, retrieves or creates associated ProvidedIdentity"""
+    """Based on target identity type, retrieves or creates associated ProvidedIdentity"""
+    identity_type: str = infer_target_identity_type(identity_data, identity_type)
+
     if identity_type == ProvidedIdentityType.email.value and identity_data.email:
         identity = ProvidedIdentity.filter(
             db=db,
@@ -352,6 +354,31 @@ def _get_or_create_provided_identity(
             detail="Identity type not found in identity data",
         )
     return identity
+
+
+def infer_target_identity_type(
+    identity_data: Identity,
+    identity_type: str,
+) -> str:
+    """
+    Consent requests, unlike privacy requests, only accept 1 identity type- email or phone.
+    If both identity types are provided, use identity type if defined in CONFIG.notifications.notification_service_type, else default to email
+    """
+    if identity_data.email and identity_data.phone_number:
+        messaging_method = get_messaging_method(
+            CONFIG.notifications.notification_service_type
+        )
+        if messaging_method == MessagingMethod.EMAIL:
+            identity_type = ProvidedIdentityType.email.value
+        elif messaging_method == MessagingMethod.SMS:
+            identity_type = ProvidedIdentityType.phone_number.value
+        else:
+            identity_type = ProvidedIdentityType.email.value
+    elif identity_data.email:
+        identity_type = ProvidedIdentityType.email.value
+    elif identity_data.phone_number:
+        identity_type = ProvidedIdentityType.phone_number.value
+    return identity_type
 
 
 def _get_consent_request_and_provided_identity(
--- a/src/fides/api/ops/api/v1/endpoints/consent_request_endpoints.py
+++ b/src/fides/api/ops/api/v1/endpoints/consent_request_endpoints.py
@@ -330,7 +330,9 @@ def infer_target_identity_type(
 ) -> str:
     """
     Consent requests, unlike privacy requests, only accept 1 identity type- email or phone.
-    If both identity types are provided, use identity type if defined in CONFIG.notifications.notification_service_type, else default to email
+    These identity types are configurable as optional/required within the privacy center config.json.
+    If both identity types are provided, we'll use identity type if defined in
+    CONFIG.notifications.notification_service_type, else default to email.
     """
     if identity_data.email and identity_data.phone_number:
         messaging_method = get_messaging_method(
--- a/src/fides/api/ops/api/v1/endpoints/consent_request_endpoints.py
+++ b/src/fides/api/ops/api/v1/endpoints/consent_request_endpoints.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import Optional
+from typing import Optional, Dict, Any
 
 from fastapi import Depends, HTTPException, Security
 from loguru import logger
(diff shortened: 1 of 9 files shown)


src/fides/api/ops/api/v1/endpoints/consent_request_endpoints.py
```diff
@@ -69,26 +72,46 @@ def create_consent_request(
...
+                identity_data=data,
+                identity_type=ProvidedIdentityType.phone_number.value,
+            )
+        else:
+            identity = _get_or_create_provided_identity(
+                db=db,
+                identity_data=data,
+                identity_type=ProvidedIdentityType.email.value,
+            )
+    elif data.email:
```
adamsachs, line 105: this may be a bit deeper than we want to go in this PR, but i think there's an edge case here that i'm not sure we're handling very cleanly. specifically, if the client (whether it's the privacy center UI or theoretically some other API client) here submits _only_ an `email` or `phone_number` while their app is configured for an incompatible notification service  -- e.g. if their `notification_service_type` property is set to `twilio_text` but the request here contains only an `email`.

i think/assume we'll probably end up erroring out much further downstream in the messaging dispatcher - but i think we could probably be a bit more defensive here (and any other endpoints that go through similar dispatching logic), and just cut things off before we get that far to give what's likely a clearer error message to the user (probably a `422` here?)

but maybe i'm missing something. either way, i don't think it's a must-have at the moment, but curious to get your thoughts.

src/fides/api/ops/api/v1/endpoints/consent_request_endpoints.py
```diff
@@ -69,26 +72,46 @@ def create_consent_request(
...
-        }
-        identity = ProvidedIdentity.create(db, data=provided_identity_data)
+    # consent requests, unlike privacy requests, only accept 1 identity type- email or phone
+    # if both identity types are provided, use identity type if defined in CONFIG.notifications.notification_service_type, else default to email
+    if data.email and data.phone_number:
+        messaging_method = get_messaging_method(
+            CONFIG.notifications.notification_service_type
+        )
+        if messaging_method == MessagingMethod.EMAIL:
+            identity = _get_or_create_provided_identity(
```
adamsachs, line 88: a bit of a nit, but i wonder if it'd make things a bit cleaner to move the `get_messaging_method` check and the different branches here _within_ the `_get_or_create_provided_identity` utility function. it may lead to a bit less code, and it also may make the top-level endpoint method here a bit more readable - all the details about getting the `identity` instance would be fully abstracted away from the top-level method, which you've already started to do here (and i think is a good direction to go in generally).

that being said, it's not too hard to follow as is, so not a huge deal. it just feels like some of the conditional branching here could be moved out of the top-level function, and possibly consolidated a bit if done more closely to the identity retrieval.

adamsachs (approved):
the BE here generally looks good to me, pretty clean updates! just a couple of minor points that i don't think are _necessary_ to address, but wanted to call attention to them.

a couple of other asides that i noticed along the way, which we may want to create follow ups for:
- i found the setup [docs](https://docs.ethyca.com/fides/dsr_quickstart/dsr_support/messaging#twilio-sms-config) generally good, but lacking a crucial detail that indicates the actual URL/path that needs to be hit for the given steps. am i missing something obvious? there do seem to be other points in the docs where we've also omitted the URL/path, so perhaps this is a deliberate decision? from the POV of a client, i think i'd find this pretty confusing. what do you think?
- i also noticed what seems to be a bug in our caching of verification codes - i don't think we're actually setting our verification codes to expire in the cache using the `identity_verification_code_ttl_seconds` setting, i think we're just relying on the `default_ttl_seconds` expiration time. i'll create a follow up ticket for this one and can probably get up a PR pretty quick to fix, just leaving a note here.

src/fides/api/ops/api/v1/endpoints/consent_request_endpoints.py
eastandwestwind: @adamsachs you're correct that in this use case:

1. A customer wants to enable notifications
2. Their `notification_service_type` set to `twilio_text`
3. Their privacy-center `config.json` allows for only email to be sent via the API
4. A privacy request OR consent request is submitted with only an email identity
5. The app will raise a `500` with message ` Error sending the verification code message: No phone identity supplied.`. 

I'd prefer to keep this logic within the messaging service, since it's a central place to handle all things messaging and erroring. Otherwise, we'll have to implement the checks on each caller, which will be more of a hassle to manage (e.g. if we implement a new caller, we have to remember to implement this check).

For now, I'll open a new follow-up ticket to do the following:

1. move this check earlier on in the messaging service, and
2. add new Exception type to the dispatcher to raise `422` code instead, using this new type to handle this err case, and 
3. catch this new exception in every place where we call our message dispatcher, and
2. update the message to be clearer: ` Error sending the verification code message: Message service expecting sms identity, but no sms identity was supplied.

src/fides/api/ops/api/v1/endpoints/consent_request_endpoints.py
eastandwestwind: hey, this is a good point, I'll give it a go!

src/fides/api/ops/api/v1/endpoints/consent_request_endpoints.py
eastandwestwind: ticket here- https://github.com/ethyca/fides/issues/2106

eastandwestwind:
Thanks so much for the thorough testing and review @adamsachs ! A couple things:

1. Based on your feedback, I've updated comments in the code to reflect he configurability of the PC: 
```
Consent requests, unlike privacy requests, only accept 1 identity type- email or phone. 
    These identity types are configurable as optional/required within the privacy center config.json.
    If both identity types are provided, we'll use identity type if defined in
    CONFIG.notifications.notification_service_type, else default to email.
```
2. As discussed in Slack, now that I know where the new docs live, I'll make the appropriate updates to Twilio configuration docs for clarity- https://github.com/ethyca/fides/issues/2107
3. Follow-up ticket for exception handling- https://github.com/ethyca/fides/issues/2106
4. Thanks so much for working on https://github.com/ethyca/fides/pull/2105, I'll take a look today!

src/fides/api/ops/api/v1/endpoints/consent_request_endpoints.py
adamsachs: nice, your point about keeping the logic within the messaging service does make sense. and so moving the check earlier in the messaging service makes a lot of sense to me.

it's a detail, but i'm not sure if i like raising the `422` from within the dispatch service - but i do get that you want to distinguish it as a specific type of exception and not just a generic `MessageDispatchException`, so you can then handle it appropriately in the places that use the dispatcher (which i also think is a great approach). so maybe just another subtype of `MessageDispatchException` that clearly indicates what's going wrong here, so it can be properly handled by the endpoint code and reported as a `422`? i guess i'm just a bit worried that if we introduce `HTTPException`s into the `message_dispatch_service`, it breaks the separation of concerns. feels a bit cleaner to have the underlying service just describe what went wrong, independent of the HTTP endpoint and how it's called; and then the specific pieces of endpoint code figure out which `HTTPException` to throw back to the user based on what's thrown up by the service underneath.

anyway, i'm splitting hairs a bit here, so feel free to ignore :) in general i think where you landed makes a lot of sense!

eastandwestwind: ah I didn't mean literally raise the 422 as an `HTTPException`, but instead use a different error class, maybe `MessageDispatchValueException` so that the endpoint can handle throwing 422. 

But I like the idea of using a subtype of `MessageDispatchException` instead of a new Exception class altogether.

eastandwestwind: But yes, you're totally right! We shouldn't be involving HTTP concepts in the messaging service

eastandwestwind: updated requirements in ticket `add new Exception type (perhaps a subtype of `MessageDispatchException`) to the dispatcher such that the endpoint can catch and raise 422 code`

adamsachs: ah ok great, we're totally on the same page then  👍

adamsachs:
nice, this all looks good to me on the backend pieces  - thanks for addressing everything i brought up in a thoughtful way!

not sure if we want to go ahead and merge without further FE code review? things certainly seem functional, but can't comment much on the code itself :)

PR #2096, opened 2022-12-20
Remove DatasetConfig.dataset field + New Get DatasetConfig Endpoints [#1763]

### LINKED ISSUE #1763: Remove dataset yaml storing code from `datasetconfig` endpoint

pattisdr:
❗ Contains migration
👉  This PR is against the `unified-fides-resources` branch, not main. 

Closes #1763 

### Code Changes

- [ ] Add a data migration to delete the DatasetConfig.dataset column. It should have been previously migrated to a ctl dataset resource
- [ ] Remove the "dataset" key from the data sent to `DatasetConfig.create_or_update` call in the new `patch_dataset_configs` endpoint
    - [ ] Leave the `dataset` key in the data sent to `DatasetConfig.upsert_with_ctl_dataset` because this method currently upserts both the CTL Dataset first and then the DatasetConfig.  This method is currently used in the old patch dataset config endpoints (json and yaml, soon to be deprecated) as well as when you're creating a connection config from a template.
- [ ] Update fixtures to not pass in a "dataset" key to data passed to DatasetConfig.create
- [ ] Throw a 404 if the ctl_dataset_id does not existing when creating a DatasetConfig through the new patch_dataset_configs endpoint
- [ ] Also add validation when linking an existing CTL Dataset to a DatasetConfig that that existing dataset has valid data categories that exist in the db.
- [ ] Add a fix to prevent existing upsert from attempting to update the id of an existing resource. Datasets are referenced by DatasetConfigs now, we need them to stay the same.
- [ ] Add new GET datasetconfig list and detail endpoints that include both the fides_key and nested ctl_dataset.  The DatasetConfig and CTLDataset are different resources and their fides_keys can differ, so both keys need to be in the response.

### Steps to Confirm

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

Get rid of the DatasetConfig.dataset field.

The previous PR stopped reading from DatasetConfig.dataset field and started writing to the new DatasetConfig.ctl_dataset resource.

Nothing should be using this field so let's get rid of it.


src/fides/api/ops/api/v1/endpoints/dataset_endpoints.py
```diff
@@ -193,12 +195,23 @@ def patch_dataset_configs(
...
         )
-        fetched_dataset: Dataset = Dataset.from_orm(ctl_dataset)
+        if not ctl_dataset:
+            raise HTTPException(
+                status_code=HTTP_404_NOT_FOUND,
+                detail=f"No ctl dataset with key '{dataset_pair.ctl_dataset_fides_key}'",
+            )
+
+        try:
+            fetched_dataset: Dataset = Dataset.from_orm(ctl_dataset)
```
pattisdr, line 205: A lot of existing ops datasets don't have an `organization_fides_key` or a `data_qualifier`, but they are not optional fields on the Dataset model.  If a ctl dataset has been created from an ops dataset, these values will be None, and then throw a validation error when parsed here.

This change is to prevent a 500 Internal server error, and catch and throw a 422 instead.  The user should then return to the ctl_dataset and add the organization_fides_key and data_qualifier.

src/fides/api/ops/api/v1/endpoints/dataset_endpoints.py
```diff
@@ -193,12 +195,23 @@ def patch_dataset_configs(
...
+        except PydanticValidationError as e:
+            raise HTTPException(
+                status_code=HTTP_422_UNPROCESSABLE_ENTITY, detail=e.errors()
+            )
+        validate_data_categories(fetched_dataset, db)
```
pattisdr, line 210: Since we have to fetch the dataset anyway to validate for saas connectors, I"m going ahead and checking that its data categories are also in the database like we do on existing dataset config endpoints. The fix is to adjust the ctl dataset though.

It's easier to add this check here than to a generic crud endpoint.

The reason this isn't part of a standard validator is that it needs access to a database session.

src/fides/api/ops/api/v1/endpoints/dataset_endpoints.py
```diff
@@ -463,6 +482,62 @@ def get_dataset(
...
+    logger.info(
+        "Finding all dataset configs for connection '{}' with pagination params {}",
+        connection_config.key,
+        params,
+    )
+    dataset_configs = DatasetConfig.filter(
+        db=db, conditions=(DatasetConfig.connection_config_id == connection_config.id)
+    ).order_by(DatasetConfig.created_at.desc())
+
+    return paginate(dataset_configs, params)
```
pattisdr, line 506: This is identical to `get_datasets` except the response has both the DatasetConfig fides_key and the CTL Dataset fides_key, which can differ.  The other parallel endpoint will eventually be deprecated.

src/fides/api/ops/api/v1/endpoints/dataset_endpoints.py
```diff
@@ -463,6 +482,62 @@ def get_dataset(
...
+    DATASETCONFIG_BY_KEY,
+    dependencies=[Security(verify_oauth_client, scopes=[DATASET_READ])],
+    response_model=DatasetConfigSchema,
+)
+def get_dataset_config(
+    fides_key: FidesKey,
+    db: Session = Depends(deps.get_db),
+    connection_config: ConnectionConfig = Depends(_get_connection_config),
+) -> DatasetConfig:
+    """Returns the specific Dataset Config linked to the Connection Config."""
```
pattisdr, line 519: This is identical to `get_dataset` except the response has both the DatasetConfig fides_key and the CTL Dataset fides_key, which can differ.  The other parallel endpoint will eventually be deprecated.

src/fides/api/ops/models/datasetconfig.py
```diff
@@ -117,6 +113,7 @@ def upsert_ctl_dataset(ctl_dataset_obj: Optional[CtlDataset]) -> CtlDataset:
                 fetched_ctl_dataset
             )  # Create/update existing ctl_dataset first
             data["ctl_dataset_id"] = ctl_dataset.id
+            data.pop("dataset", None)
```
pattisdr, line 116: This method allows both the CTL Dataset and the DatasetConfig to be upserted. The `dataset` is passed in for the CTL Dataset, but we don't want this field when we update the DatasetConfig.

src/fides/api/ctl/database/crud.py
```diff
@@ -163,10 +163,13 @@ async def upsert_resources(
...
                     )
 
+                    excluded = dict(insert_stmt.excluded.items())  # type: ignore[attr-defined]
+                    excluded.pop("id", None)  # If updating, don't update the "id"
+
                     result = await session.execute(
                         insert_stmt.on_conflict_do_update(
                             index_elements=["fides_key"],
-                            set_=insert_stmt.excluded,
+                            set_=excluded,
```
pattisdr, line 172: Allison ran into trying to upsert a CTL Dataset that was already linked to a DatasetConfig and ran into:

```
fides-fides-1  | 2022-12-22 17:00:45.731 [DEBUG] (crud:upsert_resources:182): (sqlalchemy.dialects.postgresql.asyncpg.IntegrityError) <class 'asyncpg.exceptions.ForeignKeyViolationError'>: update or delete on table "ctl_datasets" violates foreign key constraint "datasetconfig_ctl_dataset_id_fkey" on table "datasetconfig"
fides-fides-1  | DETAIL:  Key (id)=(ctl_8ca1f08e-53de-4954-a56a-73123145946e) is still referenced from table "datasetconfig".
```

I believe upserting is also trying to update the `id` of an existing resource, but we need this to stay consistent, especially since DatasetConfigs now have a FK to this table.

src/fides/api/ops/api/v1/endpoints/dataset_endpoints.py
allisonking: should we make a 🎫 for this issue?

src/fides/api/ops/api/v1/endpoints/dataset_endpoints.py
allisonking: my instinct would be on resource creation since that's the closest layer to the db so it'd catch the most things. but yes, the generic CRUD endpoint is nice code cleanliness wise, but makes changes like this more difficult. maybe something to discuss with the team and make a follow up ticket for?

allisonking (approved):
such thorough tests! 🤩

src/fides/api/ops/api/v1/endpoints/dataset_endpoints.py
pattisdr: It's tricky though because the data categories could change later! They may be valid at the time of creation and not at the time of use!

src/fides/api/ops/api/v1/endpoints/dataset_endpoints.py
pattisdr: yep will do!

src/fides/api/ops/api/v1/endpoints/dataset_endpoints.py
pattisdr: will make a ticket!

pattisdr: https://github.com/ethyca/fides/issues/2115

src/fides/api/ops/api/v1/endpoints/dataset_endpoints.py
pattisdr: https://github.com/ethyca/fides/issues/2115

PR #2111, opened 2022-12-22
Switch off dynamic top nav

seanpreston:
This PR switches off requirements for privacy requests to require connections, and datamap to require systems.


ThomasLaPiana:
@seanpreston is this still relevant?

PR #2112, opened 2022-12-22
Add Create and Upsert CTL Dataset YAML Routes [#1767]

### LINKED ISSUE #1767: Move `datasetconfig` yaml file endpoint over to `ctl_datasets`
❗ Depends on https://github.com/ethyca/fides/issues/1763 ❗

There is a `datasetconfig` [endpoint](https://github.com/ethyca/fides/blob/4408279869ce49f19ea39c6b8a60d4a0992d7b57/src/fides/api/ops/api/v1/endpoints/dataset_endpoints.py#L199) that allows users to upload yaml files directly without converting them to json. Now that datasets are entirely managed `ctl_datasets` it would be helpful to port that endpoint to the `ctl_datasets` endpoints.

Acceptance criteria
- Users are able to upload yaml files directly to `/yml/dataset/`
- deprecating the existing dataset config yaml endpoint

pattisdr:
👉 Note, this PR is against `unified-fides-resources` not `main`

Closes #1767 

### Code Changes

* [x] Add new endpoint `POST {{host}}/yml/dataset`.  
    * [x] This endpoint is similar to `POST {{host}}/yml/dataset` except your request body should be yaml.
    * [x] Content-type should be `application/x-yaml`
    * [x] Only one dataset can be submitted through this endpoint
* [x] Add new endpoint `POST {{host}}/yml/dataset/upsert`
    * [x] This endpoint is similar to `POST {{host}}/yml/dataset/upsert` except your request body should be yaml.
    * [x] Content-type should be `application/x-yaml`
    * [x] Multiple datasets can be submitted through this endpoint.  If a `fides_key` matches an existing `fides_key`, we update the matching resource, otherwise, we create a new record.

### Steps to Confirm

* [ ] See new endpoints in committed postman collection.  There's no UI for these endpoints but you can test them out via API calls.

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`. <- We're going to have to update the CHANGELOG anyway before merging unified-fides-resources

### Description Of Changes

The old DatasetConfig yaml endpoint `PATCH {{host}}/yml/connection/{{connection_key}}/dataset` will likely be deprecated soon, as it is upserting CTL Datasets before linking the CTL Dataset to a DatasetConfig.  We'd like to the CTL Dataset to be modified directly first without this happening magically behind the scenes.

Anyway, what was nice about this existing endpoint is that you could upload yaml directly instead of having to convert a yaml file to JSON first. We're porting this over to the CTL Dataset endpoints in case it's useful.


ThomasLaPiana:
@pattisdr I think I'm just missing the context here, but what was the primary reasoning for wanting to enable loading datasets via yaml? I think theoretically it is an extra step, but considering the `CLI` abstracts this, is this an additional maintenance burden that feels necessary? We've had no requests from customers to add this functionality that I'm aware of

ThomasLaPiana (commented):
Changes look good, but not sure what the motivation for this change is

pattisdr:
I found this functionality useful on the ops side and I was maintaining parity with an API endpoint that existed ops side that we are planning on deprecating.  If that's not a strong enough case, can you instead explain what is available on the ctl side?   Can you only create datasets via json or is there a yaml option?

ThomasLaPiana:
> I found this functionality useful on the ops side and I was maintaining parity with an API endpoint that existed ops side that we are planning on deprecating. If that's not a strong enough case, can you instead explain what is available on the ctl side? Can you only create datasets via json or is there a yaml option?

I thought I had opened an issue semi-related to this but I can't find it, so maybe I didn't 🤔

On the `ctl` side, we only support JSON in the API endpoints, and the primary reason for this is that we _always_ expect the user to interact with the webserver via the CLI. So the CLI ingests the YAML, does the parsing and validity checks, and then sends a JSON object. This is a difference between ops and ctl, so in the end maybe this doesn't/shouldn't apply to ops, but I think having two different user experiences detracts from the synergy we're trying to establish.

I had planned on opening an issue (I still should) that enables the same thing experience for ops. Right now, as you know, a user has to make API calls, most likely by setting up scripts of some kind, to configure ops programmatically. My idea was that the CLI should also handle YAML versions of the ops objects so that the users aren't needing to make API calls directly, and instead can store their ops configuration as YAML objects and with a single `fides push` configure everything. On second thought I'm not sure if that's a good idea considering there would be some secrets there potentially and we wouldn't want users storing them in version control.

That being said, this is a huge tangent and not directly related to this PR, so I don't see a blocker here! It just triggered my brain to remember that I had planned on establishing that functionality, but there's no harm in supporting YAML as well, it just didn't seem strictly necessary when I glanced at it

pattisdr:
thanks for the background @ThomasLaPiana, given your eventual plans to have the CLI handle yaml versions of ops objects, would you prefer if we just closed?

ThomasLaPiana:
> thanks for the background @ThomasLaPiana, given your eventual plans to have the CLI handle yaml versions of ops objects, would you prefer if we just closed? I'm now leaning towards closing.

If it is beneficial to you in the short-term to have this functionality, I'd say merging is the right move for sure. However if you're only doing this to update/expand existing functionality, it is worth considering that eventually I'd like to remove this in favor of handling the ops objects via the CLI

pattisdr:
OK thanks Thomas, I'm just going to close.

Issue #1766, opened 2022-11-14
Update Admin UI to support new dataset config changes


TheAndrewJackson:
❗ Depends on https://github.com/ethyca/fides/issues/1762 ❗

Now that the backend no longer supports storing dataset yaml files on the `datasetconfig` model the admin ui will need to be updated. 

When configuring connectors users are able to upload yaml configs like so
<img width="1216" alt="Screen Shot 2022-11-14 at 10 08 02" src="https://user-images.githubusercontent.com/17103888/201694729-a39b0c3f-6d15-4855-8e1b-68505f4b148e.png">
Whenever edits are made they're sent to `/connection/{connection_key}/dataset/`.

This form needs to be updated to send yaml configs to `/dataset` and use the returned ID to link the `ctl_datasets` model to the `datasetconfig` model using the new `ctl_dataset_id` foreign key. Now only that ID will be sent in to `/connection/{connection_key}/dataset/`

So the new form flow will be:

Creating new config
- Send config yaml to `/dataset` and save the return id in memory for the next request
- Send nearly the same data to  `/connection/{connection_key}/dataset/` but leave out the `dataset` field and instead include the stored in in `ctl_datasets_id`

Editing config
- Post updates to `/dataset`
  - This assumes that the `ctl_datasets` model is already linked `datasetconfigs`

This should get current form to feature parity with how it currently works. It still needs some additional updates.

Now that all the datasets are managed in the same place it's possible that a user will configure a connector and want to link it with a dataset that has already been created from the datasets UI. They will need a way to select from datasets that already exist. 

This means that users will be able to either create datasets from this page or reuse one from the list of datasets that already exist.

Extra considerations
- There are two separate yaml forms in the admin ui. This might be a good time to unify them
- It looks like the admin ui may not call the dataset validation endpoints. It would be good to double check and make sure they're called when creating and editing datasets

pattisdr:
@allisonking The `unified-fides-resources` branch is ready to work with this new UI.  A few changes from above.

**Creating new config:**
-  Save dataset: Send config yaml to /dataset and save the fides key in memory for the next request (POST {{host}}/dataset/upsert)
- Link dataset to datasetconfig: "PATCH /connection/{connection_key}/datasetconfig"

request body:
```json
[
        {
            "fides_key": "<Connection Identifier from previous UI screen - for the DatasetConfig>",
            "ctl_dataset_fides_key": "<fides key from previous step - for the CTL Dataset>",
        }
    ]

```

[...]

PR #2113, opened 2022-12-22
Update admin UI for dataset config changes

allisonking:
Closes https://github.com/ethyca/fides/issues/1766

### Code Changes

* [x] Update typescript types to include connection/dataset config types
* [x] Update slices to stop using deprecated endpoints and using the new endpoints as described [here](https://github.com/ethyca/fides/issues/1766#issuecomment-1362174155)
* [x] We now have to do two requests to update the datasets associated with a connection—to upsert on datasets, then to patch the datasetconfig 
* [x] Small fix to the demo dataset included in `demo_resources`
* [x] Cypress tests 

### Steps to Confirm

* [ ] Navigate to "Privacy Requests" then "Connection Manager"
* [ ] Either click "Create new connection" or open the ellipsis menu on an existing connection and click "Configure"
* [ ] Open the "Dataset configuration" tab 
* [ ] Either make edits to the existing yaml, or paste in a dataset yaml (note: if copying from `demo_resources`, use the one on this branch, and remove the top level `dataset` field)
* [ ] Save and verify the dataset updated and that when you open the connection you see the right dataset(s)

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### Description Of Changes

🚨 Important caveats 🚨 

When editing a dataset configuration, we cannot currently handle adding or deleting a dataset in a very good way. Consider:

My mongo connection has two dataset configurations with `fides_key`s `one` and `two`. The first time we save this, they get saved with `dataset_pairs` of

```
[
  { fides_key: 'one', ctl_dataset_fides_key: 'one' }, 
  { fides_key: 'two', ctl_dataset_fides_key: 'two' }
]
```

Now to delete one of these, we go into the yaml editor and delete the dataset with a `fides_key` of `two`. In order to do this, we'd have to call a DELETE and wouldn't be able to do a normal upsert. That's not the worst, but things get hairy because the `fides_key`s themselves are editable. So I could change `two` to `three`, but then should we be sending a pair of `{fides_key: 'two', ctl_dataset_fides_key: 'three'}` or `{fides_key: 'three', ctl_dataset_fides_key: 'three'}` (i.e. the user cannot, in the current UI, view the config `fides_key`, only the `ctl_dataset_fides_key`)

Adding has a similar problem, where we can add a `fides_key: 'three', ctl_dataset_fides_key: 'three'}` but it starts to involve keeping track of before and after. Even switching order is difficult, since if you have the yaml field with datasets `one` and `two` you start off with:

```
[
  { fides_key: 'one', ctl_dataset_fides_key: 'one' }, 
  { fides_key: 'two', ctl_dataset_fides_key: 'two' }
]
```

And then let's say you switch the order of the dataset yaml blocks. Should the new pair be

```
[
  { fides_key: 'one', ctl_dataset_fides_key: 'two' }, 
  { fides_key: 'two', ctl_dataset_fides_key: 'one' }
]
```
or
```
[
  { fides_key: 'two', ctl_dataset_fides_key: 'two' }, 
  { fides_key: 'one', ctl_dataset_fides_key: 'one' }
]
```

It's not possible to tell right now since we don't actually expose `fides_key` in the UX (perhaps for good reason). 

Because there have been other indications that we are [moving away from yaml forms](https://github.com/ethyca/fidesplus/issues/472), and also because it seems like this form could really be a multi select on datasets instead (see [Andrew's comment](https://github.com/ethyca/fides/issues/1766#issue-1448286949)) of a yaml form, and also because there's already a separate dataset yaml form, I thought it didn't make sense to go down this thorny route at the moment.

I was about to go ahead and implement a multiselect for datasets, but UX started to get weird if we want to keep the yaml form as well and I think would need design help if we do want both (for example, should the yaml text editor update as the user selects different datasets?). I think the thing that might make the most sense is to get rid of the yaml form here, and just use a multiselect on datasets. Then add a link like "don't see your dataset? go to the dataset management page"

All that to say, this PR mostly gets things working as they did before, with the caveat about the dataset pairs above. If we do a multiselect only instead, then that problem should go away


pattisdr (approved):
This is a nice increment that preserves the current UI while switching to new endpoints!

The current UI does have some tricky cases as you pointed out.  I like your suggestion of adding a followup ticket that gets rid of the yaml form here and adds a multi-select on datasets.  The danger with the current UI is that an existing customer might not realize these datasets are unified under the hood, and they might inadvertently overwrite an existing dataset.

One issue that I need to figure out the best way to solve on my end, if you send in a dataset without a `default_organization` or a `data_qualifier`, the CTL Dataset will create properly, but then we'll get a hard-to-debug error when we try to link that to a DatasetConfig because those are required fields.

allisonking:
> One weird thing on the backend, a CTL Dataset can be created with a null default_organization or a null data_qualifier, but the Fideslang Dataset schema requires those fields to exist. So if a CTL Dataset doesn't have that, it will error when we try to link it to the DatasetConfig. Different ways we could go with fixing this - either we make them required for CTL Datasets, or we make them not required for Fideslang Datasets.

Ah yes, I noticed that too when I was trying to fix the data qualifier stuff. I'm guessing the upsert endpoint needs to do some validation that it isn't doing... Actually, this is kind of familiar: https://github.com/ethyca/fidesplus/pull/197#pullrequestreview-1139256078

pattisdr:
Reviewing now @allisonking!

pattisdr (approved):
Thanks for adding all the great cypress tests!

PR #2116, opened 2022-12-23
implement endpoints to retrieve DSR rules and rule_targets

### LINKED ISSUE #1741: `GET` endpoints to retrieve DSR `rule_target` (and `rule`) configuration data

We currently have endpoints to get information about DSR `policy` configuration objects, and we also have endpoints to _update_ the `rule` and `rule_target` config objects associated with the policies. we don't, however, have `GET` endpoints to simply retrieve the information about the `rule` and `rule_target` objects. this is specifically problematic for `rule_target` data, since that data is not returned by any other related endpoints; `rule` data can be retrieved as part of the response payload in `GET` calls retrieving their associated `policy` objects.

i assume at some point we'll want to create a management UI to help manage rules and rule targets, and so then these endpoints will likely be _necessary_ then. until that point, they may not be strictly _necessary_, but when configuring a system, it is undoubtedly cumbersome and problematic that we do not have any API support (at least that I can tell) for retrieving rule target config data. without this, admins are unable to determine which data categories are targeted by their policies!

implementation of the following endpoints:
- `GET /dsr/policy/{policy_key}/rule/{rule_key}/target`
- `GET /dsr/policy/{policy_key}/rule/{rule_key}/target/{rule_target_key}`
- `GET /dsr/policy/{policy_key}/rule`
- `GET /dsr/policy/{policy_key}/rule/{rule_key}`

not really sure, the implementation here seems pretty straightforward. as noted, the `rule_target` implementation is more important/pressing IMO.

### Additional context

I came across this need when trying to configure my local server on bootstrap for a particular use case. i wanted to know about the default policies that come loaded on the system, and i was unable to find any information about the rule targets through the API. I had to query the postgres db directly with psql!

adamsachs:
Closes #1741 

### Code Changes

* `GET` endpoint for listing DSR rules associated with a policy (`/dsr/policy/{policy_key}/rule`)
* `GET` endpoint for retrieving an individual DSR rule with a given key (`/dsr/policy/{policy_key}/rule/{rule_key}`)
* `GET` endpoint for listing DSR rule targets associated with a given rule (and policy) (`/dsr/policy/{policy_key}/rule/{rule_key}/target`)
* `GET` endpoint for retrieving an individual DSR rule target with a given key (`/dsr/policy/{policy_key}/rule/{rule_key}/target/{rule_target_key}`)

### Steps to Confirm

* `nox -s dev` or any other basic fides env
* go to swagger docs UI, e.g. http://localhost:8080/docs
* authenticate with root client ID and secret using authenticate button
* go to new endpoints, e.g. http://localhost:8080/docs#/DSR%20Policy/get_rule_list_api_v1_dsr_policy__policy_key__rule_get
* use the default policies and rules to test out endpoint functionality, e.g.:
    *  policy = `default_access_policy` OR policy = `default_erasure_policy`
    * rule = `default_access_policy_rule` OR rule = `default_erasure_policy_rule`
* alternatively, you can hit these endpoints on your local fides server using cURL, e.g.: 
```
curl -X 'GET' \
  'http://localhost:8080/api/v1/dsr/policy/default_erasure_policy/rule?page=1&size=50' \
  -H "Authorization: Bearer $BEARER_TOKEN"
```

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### Description Of Changes

Added new endpoints and corresponding test coverage


adamsachs:
no rush at all on reviewing, just decided to knock this out before break and then cleaned it up and got it ready for review!

src/fides/api/ops/api/v1/endpoints/policy_endpoints.py
```diff
@@ -297,6 +366,82 @@ def delete_rule(
...
+        )
+
+    return rule_target
+
+
+@router.get(
+    urls.RULE_TARGET_LIST,
+    status_code=HTTP_200_OK,
+    response_model=Page[schemas.RuleTarget],
+    dependencies=[Security(verify_oauth_client, scopes=[scope_registry.RULE_READ])],
```
pattisdr, line 400: Note to self: existing target endpoints also use the rule read scope, there's no separate scope for targets

src/fides/api/ops/api/v1/endpoints/policy_endpoints.py
```diff
@@ -297,6 +366,82 @@ def delete_rule(
...
+    response_model=schemas.RuleTarget,
+    dependencies=[Security(verify_oauth_client, scopes=[scope_registry.RULE_READ])],
+)
+def get_rule_target(
+    *,
+    policy_key: FidesOpsKey,
+    rule_key: FidesOpsKey,
+    rule_target_key: FidesOpsKey,
+    db: Session = Depends(deps.get_db),
+) -> schemas.RuleTarget:
```
pattisdr, line 438: I think this return type should be the RuleTarget model, not the schema, because that's what this method literally returns.  It gets serialized with the RuleTarget schema due to the `response_model` declaration above, then hopefully you can get rid of `# type: ignore[return-value]`

Existing endpoints in this file may make this same mistake, but compare to the connection config endpoints for example.

src/fides/api/ops/api/v1/endpoints/policy_endpoints.py
```diff
@@ -148,6 +148,75 @@ def create_or_update_policies(
...
+    status_code=HTTP_200_OK,
+    response_model=schemas.RuleResponseWithTargets,
+    dependencies=[Security(verify_oauth_client, scopes=[scope_registry.RULE_READ])],
+)
+def get_rule(
+    *,
+    policy_key: FidesOpsKey,
+    rule_key: FidesOpsKey,
+    db: Session = Depends(deps.get_db),
+) -> schemas.Rule:
```
pattisdr, line 213: I'd update return type to the Rule sqlalchemy model and then hopefully you cam remove your type ignore below.

pattisdr (changes requested):
Very clear and easy to follow!  I appreciate you following existing conventions for how we were laying out these ops API endpoints and how we were testing them. Nice pagination test.

All comments are minor, the primary one I'd like fixed are the return types.

src/fides/api/ops/api/v1/endpoints/policy_endpoints.py
```diff
@@ -148,6 +148,75 @@ def create_or_update_policies(
...
+    logger.info("Finding rule with key '{}'", rule_key)
+    policy = get_policy_or_error(db, policy_key)
+    rule = Rule.filter(
+        db=db,
+        conditions=((Rule.policy_id == policy.id) & (Rule.key == rule_key)),
+    ).first()
+    if not rule:
+        raise HTTPException(
+            status_code=HTTP_404_NOT_FOUND,
+            detail=f"No Rule found for key {rule_key}.",
```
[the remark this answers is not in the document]
adamsachs: good call out, thanks! while we're on it, do you think this is the preferred implementation that i have here? i.e. since the rule key is technically a unique column, we could just do a rule lookup only on its `key` and not look at the `policy` in this endpoint...but i decided on making it `policy`-specific, since we can only define rules via the `policy` they're attached to, and that relationship is one-to-many (one `policy` to many `rule`s, but a `rule` can only be associated with one `policy`)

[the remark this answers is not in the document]
pattisdr: I like what you have here, nesting it under the policy, for reasons you described

pattisdr (approved):
bravo, looks good to me when tests are passing. there was also a change merged to main related to tests. It shouldn't affect these ops tests, but that's something additional you could check.

adamsachs:
thanks for the review and attention to detail @pattisdr! i don't think there were any test failures before, just the known timeout that can occur on `CLI Command Checks / Fides-Deploy`, unrelated to code changes.

anyway, i've merged in `main` just to be sure, i'll resolve comments and confirm the next CI run looks good before merging.

Issue #2122, opened 2023-01-03
Configure the Privacy Center with which (if any) consent type should be “executable” using connectors


seanpreston:
In the privacy center `config.json`, we should add a boolean field to `consentOptions` such as `executable`. this should be added to the http request to the backend to let the server know if we should propagate the consent request.

This will indicate to the backend API endpoint which data [within the consent dictionary](https://github.com/ethyca/fides/pull/2125/files#diff-2258a0f00bfcec8daae12dfa45ce6e620bccc2fd943597d5633b73e9098d2110R1618-R1619) to filter out, and which to propagate.

pattisdr:
This ticket is confusing to me, I'd suggest structuring more similar to access/erasure configuration, in that the `config.json` defines _which_ Policy to use, but the _details_ of those Policies are stored in the db.   The issue comment above instead seems to add all the details to the `config.json`

My proposal:
- the `config.json` specifies which default policy to use for `consent` (already present in the POC branch)
- A separate UI screen that we need to build loads this policy from the `consent.json` which should have a single default Consent Rule and three default RuleUses attached to it.  The `RuleUses` have a `key` field which can be used as the `consentOption` and an `executable` boolean field.
- Saving what is executable will send an API request to create/update rule uses.

pattisdr:
Note that there's still some details being worked out as to whether we'll create a new `RuleUse` table, or we'll just add new columns and conditional logic around the existing `RuleTarget` table.  If we use `RuleTargets`, the API's will already exist which is a bonus.

pattisdr:
I ended up needing to add `executable` to `consentOptions` since I ripped out the `RuleUse` table I was previously using to store this https://github.com/ethyca/fides/pull/2125/files/749a752b48fe3c5dccb14d2724a2e804615cb5e4#r1065240917. I had concerns about there being multiple sources of truth for this information.

I am now loading these options directly from config.json.  More testing would be useful to verify this is a workable path forward for now.

One thing that would be useful to add to this ticket would be passing in the `policy_key` from the `config.json` into the `set_consent_preferences` API request from the frontend.

pattisdr:
Latest updates to what I think this ticket should do:

Work should be based off of https://github.com/ethyca/fides/pull/2125/ branch or `consent-propagation` if we can get that branch merged quickly

- Set default consent preferences in privacy center `config.json` to be executable = False, they are currently True
[...]

PR #2125, opened 2023-01-04
[POC][Draft]Executing Consent Requests via the Execution Layer

pattisdr:
Illustrating that we can execute consent requests as privacy requests and share the same execution layer.

Closes #<issue>

### Code Changes

### Steps to Confirm

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes


src/fides/api/ctl/database/seed.py
```diff
@@ -278,6 +288,44 @@ async def load_default_dsr_policies() -> None:
...
+                "action_type": ActionType.consent.value,
+                "name": "Default Consent Rule",
+                "key": DEFAULT_CONSENT_RULE,
+                "policy_id": consent_policy.id,
+                "client_id": client_id,
+            },
+        )
+
+        log.info("Creating: Data Category Access Rules...")
+        for use in ["advertising", "advertising.first_party", "improve"]:
```
pattisdr, line 316: I just added the same consent options that I see in src/fides/data/test_env/privacy_center_config/config.json as defaults.

src/fides/api/ops/api/v1/endpoints/consent_request_endpoints.py
```diff
@@ -264,15 +270,48 @@ def set_consent_preferences(
...
+        data=[
+            PrivacyRequestCreate(
+                identity=identity,
+                policy_key=DEFAULT_CONSENT_POLICY,
+                consent_preferences=[
+                    consent.dict() for consent in consent_preferences.consent if consent
+                ],
+            )
+        ],
+        authenticated=True,
```
pattisdr, line 295: We've authenticated with the existing consent request workflow, so now we can say this is authenticated when we create the privacy request to not send another email.

src/fides/api/ops/api/v1/endpoints/consent_request_endpoints.py
```diff
@@ -264,15 +270,48 @@ def set_consent_preferences(
...
+    if privacy_request_results.failed or not privacy_request_results.succeeded:
+        raise HTTPException(
+            status_code=HTTP_400_BAD_REQUEST,
+            detail=privacy_request_results.failed[0].message,
+        )
+
+    consent_request.privacy_request_id = privacy_request_results.succeeded[0].id
+    consent_request.save(db=db)
+
+    return consent_preferences
```
pattisdr, line 307: Currently returning the same consent preferences response as before, but now a privacy request has also been created.  We might return a new response with a combination of both resources.

src/fides/api/ops/task/graph_task.py
```diff
@@ -553,6 +554,30 @@ def erasure_request(self, retrieved_data: List[Row], *inputs: List[Row]) -> int:
...
+                None,
+                ActionType.erasure,
+                ExecutionLogStatus.error,
+            )
+            return False
+
+        output: bool = self.connector.run_consent_request(
+            self.traversal_node, self.resources.policy, self.resources.request, identity
+        )
+        self.log_end(ActionType.consent)
```
pattisdr, line 578: The parallel GraphTask.erasure_request (and access_request) also caches that the erasure was completed on this node so if we restart from failure, we don't re-attempt this node.   This is used to update the graph on retry.

```
self.resources.cache_erasure(
            f"{self.key}", output
        ) 
```

There's a lot of good reasons to not try to re-run an erasure request on a collection if it's already been completed, but I don't know if we need that level of detail here, beyond the execution logs.  I think if a consent request fails, we could run the entire step in its entirety. `cache_failed_checkpoint_details` caches the step where it failed, but I don't think we need to build a modified graph on retry.

src/fides/api/ops/models/policy.py
```diff
@@ -469,6 +487,95 @@ def update(self, db: Session, *, data: Dict[str, Any]) -> FidesBase:
         return super().update(db=db, data=data)
 
 
+class RuleUse(Base):
```
pattisdr, line 490: Copy-pasted and adapted from RuleTarget, this is just for consent.  RuleUse may not be the best name - I'm not well-versed in the nomenclature here.

src/fides/api/ops/models/privacy_request.py
```diff
@@ -190,6 +191,7 @@ class PrivacyRequest(IdentityVerificationMixin, Base):  # pylint: disable=R0904
 
     cancel_reason = Column(String(200))
     canceled_at = Column(DateTime(timezone=True), nullable=True)
+    consent_preferences = Column(MutableList.as_mutable(JSONB), nullable=True)
```
pattisdr, line 194: We need somewhere to store this, so I'm storing as a List of Consent objects.  These pertain to the preferences submitted for a given request which is why we need to persist them separately from the current saved preferences.

src/fides/api/ctl/database/seed.py
```diff
@@ -278,6 +288,44 @@ async def load_default_dsr_policies() -> None:
...
+            try:
+                RuleUse.create(
+                    db=db_session,
+                    data={
+                        "key": use,
+                        "data_use": use,  # TODO: get rid of data_use, and have this be the key.
```
pattisdr, line 322: Redundant pieces of data here, a RuleUse might just have a `key` which is the `use` like `advertising`, the `rule_id, and whether it's `executable`

src/fides/api/ops/api/v1/endpoints/consent_request_endpoints.py
```diff
@@ -264,15 +270,48 @@ def set_consent_preferences(
...
+    consent_preferences: ConsentPreferences = _prepare_consent_preferences(
+        db, provided_identity
+    )
+
+    privacy_request_results: BulkPostPrivacyRequests = create_privacy_request_func(
+        db=db,
+        data=[
+            PrivacyRequestCreate(
+                identity=identity,
+                policy_key=DEFAULT_CONSENT_POLICY,
```
pattisdr, line 289: API endpoint should be adjusted to allow the policy as an argument

src/fides/api/ops/api/v1/endpoints/consent_request_endpoints.py
```diff
@@ -264,15 +270,48 @@ def set_consent_preferences(
                 raise HTTPException(
                     status_code=HTTP_400_BAD_REQUEST, detail=Pii(str(exc))
                 )
+    identity = Identity()
+    setattr(
+        identity,
+        provided_identity.field_name.value,  # type:ignore[attr-defined]
+        provided_identity.encrypted_value["value"],  # type:ignore[index]
+    )
```
pattisdr, line 278: Pulling out the identity to feed this back into creating a PrivacyRequest.
Note that this creates a separate ProvidedIdentity for the PrivacyRequest. Not sure how I feel about this.

So both the ConsentRequest and PrivacyRequest are linked to different ProvidedIdentity records for this POC.

src/fides/api/ops/service/connectors/saas_connector.py
```diff
@@ -404,6 +404,51 @@ def mask_data(
         self.unset_connector_state()
         return rows_updated
 
+    def run_consent_request(
+        self,
+        node: TraversalNode,
+        policy: Policy,
+        privacy_request: PrivacyRequest,
+        identity_data: Dict[str, Any],
+    ) -> bool:
```
pattisdr, line 413: This is where Saas connector experts would step in to start working on how to make a consent request to a third party. This method doesn't currently make a request.

src/fides/api/ops/service/privacy_request/request_runner_service.py
```diff
@@ -415,6 +421,22 @@ async def run_privacy_request(
...
+            action_type=ActionType.consent
+        ) and can_run_checkpoint(
+            request_checkpoint=CurrentStep.consent,
+            from_checkpoint=resume_step,
+        ):
+
+            await run_consent_request(
+                privacy_request=privacy_request,
+                policy=policy,
+                graph=dataset_graph,
```
pattisdr, line 434: I'm currently passing in the same dataset graph under the assumption that we'll create saas ConnectionConfigs where the details to talk to the third party are expressed in the saas config, and that there's an attached DatasetConfig.

Current logic will skip all collections that aren't attached to a saas connector.  The saas-connector logic can then determine if the type of saas connector can manage a consent request.

src/fides/api/ops/service/connectors/saas_connector.py
```diff
@@ -404,6 +404,51 @@ def mask_data(
...
+            Consent(**pref) for pref in privacy_request.consent_preferences or []
+        ]
+
+        for preference in consent_preferences:
+            if preference.data_use not in [use.key for use in executable_uses]:
+                logger.info(
+                    "Skipping updating preference: {} as it is not executable",
+                    preference.data_use,
+                )
+
```
pattisdr, line 443: This sort of logic is probably better moved up on one level to `GraphTask.consent_request` as it is applicable for all connectors.  This filters the consent preferences to just the executable ones.

pattisdr:
Note: commit https://github.com/ethyca/fides/pull/2125/commits/4fc9ec71b75b0ac4e68d78ef1778963c85c1b8bd experiments with how to create a simpler sort of graph for ConsentRequests.  It creates a graph with at most one stubbed collection per dataset type, as I imagine consent requests will largely be requests against the larger service instead of to multiple specific collections.

I may be wrong on these assumptions so the easiest way forward here is to understand where we're going to define how to talk to a third party service.  Once I know how/where that's defined, then I can build a graph with that information.  The larger point here is that we're capable of building a graph with just one "node" per dataset.

seanpreston:
Some initial comments:

>  PATCH {{host}}/consent-request/{{consent_request_id}}/preferences now creates and queues a separate PrivacyRequest to execute those consent preferences

This will need to happen for specific consent prefs when a flag is set in the post data.

> [where we're going to define how to talk to a third party service](https://github.com/ethyca/fides/pull/2125#issuecomment-1371250688)

I think it's safe to say this will be inside the a saas connector yaml, whether that's consent only for one connector, or interlinked with an existing saas connector for now.

src/fides/api/ops/models/datasetconfig.py
```diff
@@ -96,6 +96,29 @@ def get_graph(self) -> Dataset:
...
+    def get_graph_with_stubbed_collection(self) -> Dataset:
+        """
+        Return a Dataset that only has a single collection with the same
+        name as the Dataset, provided at least one collection exists on that Dataset.
+
+        This is for building graphs of the system where we only want one Node
+        per Dataset, and there are no dependencies between Nodes. Additionally,
+        code executed cannot depend on any information at the Collection level, it is a graph
+        whose nodes represent the dataset.
+        """
```
seanpreston, line 111: This is a very useful util

src/fides/api/ops/models/policy.py
seanpreston: Did you consider updating the `RuleTarget` table to include `.data_use` as an option vs. adding another table that's conditionally used alongside `ConsentRule`s? Might be cleaner if all "targets" are grouped together in some way.

pattisdr: I did look into this but my first impression was that it was a little messier and more conditional logic would be needed than was desirable!  

I'd have to change `RuleTarget.data_category` to make it nullable but then add code-level logic everywhere to make sure we had data categories for access/erasure rules but not consent rules.

Definitely worth another look though given your opinion here, so I can explore how much additional validation is needed.

pattisdr:
> I'm not sure where I stand with this — if the collection is skipped on one hand it's good to show that it was visited, but in a pure consent request these aren't relevant to the overall traversal and so if the user is expecting them to be skipped entirely, it could be more confusing to be showing them.

@seanpreston totally agree, it is not a problem to make the graph more targeted.  Adam and I have been talking about different ways to do this, what can immediately be eliminated from the graph.

pattisdr:
> PATCH {{host}}/consent-request/{{consent_request_id}}/preferences now creates and queues a separate PrivacyRequest to execute those consent preferences

> This will need to happen for specific consent prefs when a flag is set in the post data.

@seanpreston currently this is checked at the GraphTask level - when we go to execute the task, we filter out which preferences are not executable.

src/fides/api/ctl/database/seed.py
pattisdr: This was updated from the original POC. Instead of creating RuleUses that mimic the individual uses in the `config.json` file, I read whether the rules are executable from the `config.json` file directly. The immediate concern was that these two sources of truth would get out of sync.

src/fides/api/ops/api/v1/endpoints/consent_request_endpoints.py
pattisdr: Updating the API response would fall under this PR https://github.com/ethyca/fides/issues/2142 as we decide what the PrivacyCenter should display after Consent preferences are saved in Fides and then a follow-up Privacy Request is queued.

src/fides/api/ops/models/policy.py
pattisdr: Latest commits remove the `RuleUse` table entirely and just load the `config.json` file directly to see whether the customer has said a given `consentOption` is executable.

src/fides/api/ops/api/v1/endpoints/consent_request_endpoints.py
pattisdr: The API endpoint now accepts an optional `policy_key` as an argument, and falls back to the default_consent_policy if none is expected. FE should be adjusted to pass in the policy_key it pulls off of the `config.json`

src/fides/api/ctl/database/seed.py
```diff
@@ -278,6 +281,30 @@ async def load_default_dsr_policies() -> None:
...
+            db=db_session,
+            data={
+                "action_type": ActionType.consent.value,
+                "name": "Default Consent Rule",
+                "key": DEFAULT_CONSENT_RULE,
+                "policy_id": consent_policy.id,
+                "client_id": client_id,
+            },
+        )
+
```
pattisdr, line 307: Here I just seed the database with a new default Policy with a Consent Rule attached. The primary purpose of this resource is to let the privacy request execution workflow know that this is a consent request!  No RuleTargets are required for ConsentRules.  Relevant configuration is pulled from the `config.json` file.

src/fides/api/ctl/migrations/versions/de456534dbda_add_privacyrequest_consent_preferences_.py
```diff
@@ -0,0 +1,43 @@
...
+down_revision = "3caf11127442"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column(
+        "privacyrequest",
+        sa.Column(
+            "executable_consent_preferences",
```
pattisdr, line 23: This optional list of consent preferences persists to the PrivacyRequest what consent options we're updating, and in particular, just the executable ones.  That way the execution workflow doesn't have to worry about what is and isn't executable, it can use the whole list.

src/fides/api/ctl/migrations/versions/de456534dbda_add_privacyrequest_consent_preferences_.py
```diff
@@ -0,0 +1,43 @@
...
+            postgresql.JSONB(astext_type=sa.Text()),
+            nullable=True,
+        ),
+    )
+
+    op.add_column(
+        "consentrequest", sa.Column("privacy_request_id", sa.String(), nullable=True)
```
pattisdr, line 30: If a privacy request is created to propagate consent request preferences, we link it on the ConsentRequest for record-keeping.  A privacy request may not always be created, for example, if none of the consent preferences ended up being executable, or something failed with privacy request creation.

src/fides/api/ops/api/v1/endpoints/consent_request_endpoints.py
```diff
@@ -216,6 +225,86 @@ def get_consent_preferences(
...
+            data_use: str = consent.get("fidesDataUseKey")
+            if not data_use:
+                continue
+
+            if consent.get("executable"):
+                executable_consent_options.append(data_use)
+            else:
+                logger.info("Consent option: '{}' is not executable.", data_use)
+
+        return executable_consent_options
```
pattisdr, line 248: Here's where I load whether consent options are executable directly from the `config.json` file. This file_path may need to be configurable. Additionally, Allison raised this may not work with the built image.

src/fides/api/ops/api/v1/endpoints/consent_request_endpoints.py
```diff
@@ -258,7 +347,27 @@ def set_consent_preferences(
...
+        provided_identity,
+        data.policy_key or DEFAULT_CONSENT_POLICY,
+        consent_preferences,
+    )
+
+    if privacy_request_creation_results:
+        consent_request.privacy_request_id = privacy_request_creation_results.succeeded[
+            0
+        ].id
+        consent_request.save(db=db)
```
pattisdr, line 368: This is the primary change of this PR. In the existing PATCH `{{host}}/consent-request/{{consent_request_id}}/preferences` endpoint, After we save consent preferences, we queue a privacy request that attempts to propagate those preferences to third party systems.

src/fides/api/ops/models/datasetconfig.py
```diff
@@ -96,6 +96,24 @@ def get_graph(self) -> Dataset:
...
+        doesn't require information stored at the collection-level.
+
+        The single Collection will be the resource that gets practically added to the graph, but the intent
+        is that this single node represents the overall Dataset, and will execute Dataset-level requests,
+        not Collection-level requests.
+        """
+        dataset_graph: Dataset = self.get_graph()
+        stubbed_collection = Collection(name=dataset_graph.name, fields=[], after=set())
+
+        dataset_graph.collections = [stubbed_collection]
```
pattisdr, line 116: For now, I am creating a Dataset with a single stubbed Collection, assuming all the information to create the request will exist at the Dataset level, and I don't need information on Collections.

This is what allows us to later build a graph with one node per dataset.

src/fides/api/ops/models/policy.py
```diff
@@ -97,7 +99,7 @@ def _validate_rule(
             raise common_exceptions.RuleValidationError(
                 "Access Rules must have a storage destination."
             )
-    if action_type in [ActionType.consent.value, ActionType.update.value]:
+    if action_type in [ActionType.update.value]:
```
pattisdr, line 102: We already technically had consent rules, we're just allowing ourselves to use them!

src/fides/api/ops/service/connectors/saas_connector.py
```diff
@@ -404,6 +404,35 @@ def mask_data(
...
+        logger.info(
+            "Demo only! Testing available consent params! identity_data: {}, executable_preferences: {}",
+            identity_data,
+            [
+                {"use": preference.data_use, "opt_in": preference.opt_in}
+                for preference in executable_preferences
+            ],
+        )
+        return True
+
```
pattisdr, line 435: Currently I am adding a `run_consent_request` method to the SaasConnector. This is where we'd make requests to third-party endpoints to propagate consent.  Extra logging is added here to demonstrate what is available but should be removed before release.

src/fides/api/ops/service/privacy_request/request_runner_service.py
```diff
@@ -415,6 +421,22 @@ async def run_privacy_request(
...
+
+            await run_consent_request(
+                privacy_request=privacy_request,
+                policy=policy,
+                graph=build_consent_dataset_graph(datasets),
+                connection_configs=connection_configs,
+                identity=identity_data,
+                session=session,
+            )
+
```
pattisdr, line 439: Adding an additional step to the existing request runner, similar to how we have steps above for access and erasure.

src/fides/api/ops/api/v1/endpoints/consent_request_endpoints.py
pattisdr: One question I have are if data uses are hierarchical like data targets.  If advertising is not executable, is advertising.first_party executable?

src/fides/api/ops/api/v1/endpoints/consent_request_endpoints.py
```diff
@@ -216,6 +225,86 @@ def get_consent_preferences(
...
+
+    Only propagate consent preferences which are considered "executable" by the current system. If none of the
+    consent preferences are executable, no Privacy Request is queued.
+    """
+    identity = Identity()
+    setattr(
+        identity,
+        provided_identity.field_name.value,  # type:ignore[attr-defined]
+        provided_identity.encrypted_value["value"],  # type:ignore[index]
+    )  # Pull the information on the ProvidedIdentity for the ConsentRequest to pass along to create a PrivacyRequest
```
pattisdr, line 268: Note that a separate `ProvidedIdentity` is created for the PrivacyRequest for now

src/fides/api/ops/api/v1/endpoints/consent_request_endpoints.py
adamsachs: hmmm. i'll need to think about this all a bit more holistically (still need to make my way through the rest of the changes), but loading privacy center config on the backend just feels very unsettling to me from an initial reaction. unless our backend is already aware of our privacy center config for other purposes, or unless we've agreed that the backend should work with the privacy center config, my gut is telling me we should try to keep the privacy center and its configuration as purely front-end artifacts. we can discuss further offline though.

pattisdr: We can explore creating database artifacts for these, that was in the original POC.  But my main goal here was to have a single source of truth - it seems like this is the only place where the customer currently configures the data_uses that they're going to use for consent.

src/fides/api/ops/models/policy.py
```diff
@@ -269,7 +276,25 @@ def save(self, db: Session) -> FidesBase:
...
+            )
+
+        policy = Policy.get_by(db=db, field="id", value=policy_id)
+        if not policy:
+            raise common_exceptions.RuleValidationError(
+                "Policy id must be specified on Rule create."
+            )
+        existing_consent_rules = policy.get_rules_for_action(ActionType.consent)
+
+        if existing_consent_rules:
```
adamsachs, line 294: i must be missing something obvious - but this seems like it's going to do this check on _any_ `Rule` that's being created? i would have thought we only wanted to run this check if the rule being added is a consent rule, i.e. if its `action_type == ActionType.consent`

pattisdr: Thank you good catch, that was my intent here!

clients/privacy-center/config/config.json
```diff
@@ -63,7 +66,8 @@
         "url": "https://example.com/privacy#analytics",
         "default": true,
         "highlight": false,
-        "cookieKeys": []
+        "cookieKeys": [],
+        "executable": true
```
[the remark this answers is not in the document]
adamsachs: - agreed offline that we should be setting `executable:false` as a default
- also agreed offline that the backend will _not_ load this file, and instead the front-end (i.e. privacy center) will fully determine which data uses are executable

src/fides/api/ops/api/v1/endpoints/consent_request_endpoints.py
pattisdr: thank you Adam - added this to the UI ticket to resolve there in the interest of getting this base branch merged quickly https://github.com/ethyca/fides/issues/2122#issuecomment-1377517996

adamsachs:
i left some comments inline that mainly summarize what's been discussed offline as next steps. generally, though, this looks really good to me! i think once those follow up points have been addressed, i'll re-review, but i expect this should be just about ready to go.

really nice work finding a relatively quick/simple framework integration here that gives us a solid foundation for future consent propagation support.

clients/privacy-center/config/config.json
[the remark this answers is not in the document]
pattisdr: thanks for summarizing Adam 👍

pattisdr:
Discussed with Adam, we're going to go ahead and merge this, as it is the foundation for the consent propagation work.  (merging into `consent-propagation`, not main.) Further changes needed are reticketed.

Issue #2115, opened 2022-12-23
Add Increased CTL Dataset Validation on Create/Update/Upsert


pattisdr:
### Is your feature request related to a specific problem?

CTL Datasets could use some increased validation before they are saved to the database.   This is coming up when we try to link an existing CTL Dataset to a DatasetConfig and it is missing expected elements. 

### Describe the solution you'd like

- Before writing a CTL Dataset to the db, verify that supplied data categories match those in the database
- Validate against the Fideslang `Dataset` schema, which requires that an `organizational_fides_key` and a `data_qualifier` exist.

Context:
https://github.com/ethyca/fides/pull/2096#discussion_r1055057301
https://github.com/ethyca/fides/pull/2113#issuecomment-1363975333

### Describe alternatives you've considered, if any

Note this will be a little tricky because this is a crud endpoint which is written in a generic way to work with numerous ctl-resources, beyond just Datasets. 

### Additional context

_Add any other context or screenshots about the feature request here._

pattisdr:
Doing some digging here - 

I see in the crud routes that are programatically defined, it does look like there's a mapping for the fides lang models. The create route for example will have the `resource` validated against the appropriate Fideslang model, so Datasets are being validated through this endpoint.

However, the `upsert` endpoint just has resources as a list of Dictionaries, I assume this is where the issue is. Hopefully this is just a quick change.

pattisdr:
That's it! It's upsert specifically that has the issue.

PR #2134, opened 2023-01-05
Add Increased CTL Dataset Validation [#2115]

pattisdr:
Closes https://github.com/ethyca/fides/issues/2134
👉  Note branch is against `unified-fides-resources`

### Code Changes

- [ ] Fix crud upsert endpoints so they validate the request against the appropriate pydantic model.
- [ ] When creating, updating, or upserting Datasets, validate that the supplied data categories align with those currently in the db. (Note that this has to happen outside of typical Pydantic validation because we need access to a db session)

### Steps to Confirm

* [ ] Verify pasting a yaml dataset in the create connector workflow that is missing an `organization_fides_key` or `data_qualifier` throws an error when the `dataset` is upserted, not on the second request where the dataset is then linked to the DatasetConfig
* [ ] Verify that pasting a yaml dataset with a bad data category fails when you upsert a dataset in the create connector workflow

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

CTL Datasets could use some increased validation before they are saved to the database.  These are things I'm running into when testing `unified-fides-resources`

1) Upserting crud resources are not getting validated against the relevant pydantic model.  So when we upsert a CTL Dataset, it is very possible it is missing something, and this breaks later when we try to use it once it's linked to a DatasetConfig. 
2) On the ops side, we were validating data categories on the dataset before we saved to the db.  Now that datasets are no longer saved on DatasetConfigs, add this validation before saving a CTL Dataset.

### DIFF
--- a/src/fides/api/ctl/routes/crud.py
+++ b/src/fides/api/ctl/routes/crud.py
@@ -48,7 +48,9 @@ async def get_data_categories_from_db(async_session: AsyncSession) -> List[Fides
     return data_categories
 
 
-async def validate_data_categories(resource: FidesModel, async_session: AsyncSession):
+async def validate_data_categories(
+    resource: FidesModel, async_session: AsyncSession
+) -> None:
     """Validate data categories defined on Datasets against data categories in the db"""
     if not isinstance(resource, Dataset):
         return
--- a/src/fides/api/ctl/routes/crud.py
+++ b/src/fides/api/ctl/routes/crud.py
@@ -8,9 +8,12 @@ generated programmatically for each resource.
 
 from typing import Dict, List
 
-from fastapi import Depends, Response, status
-from fideslang import model_map
+from fastapi import Depends, HTTPException, Response, status
+from fideslang import Dataset, FidesModel, model_map
+from fideslang.validation import FidesKey
+from pydantic import ValidationError as PydanticValidationError
 from sqlalchemy.ext.asyncio import AsyncSession
+from starlette.status import HTTP_422_UNPROCESSABLE_ENTITY
 
 from fides.api.ctl.database.crud import (
     create_resource,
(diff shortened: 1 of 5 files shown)


src/fides/api/ops/schemas/dataset.py
```diff
@@ -15,17 +15,21 @@
 )
 
 
-def validate_data_categories_against_db(dataset: Dataset, db: Session) -> None:
+def get_data_categories_from_db(db: Session) -> List[FidesKey]:
+    """Query for existing data categories in the db using a synchronous session"""
+    return [cat[0] for cat in db.query(DataCategory.fides_key).all()]
```
pattisdr, line 20: Refactored so I could share the `validate_data_categories_against_db` method on the ops side by passing in a list of data categories since one endpoint has access to an async session and the other has access to a sync session right now.

tests/ctl/core/test_api.py
```diff
@@ -142,6 +144,41 @@ def test_api_upsert(
...
+
+    async def test_data_categories_validated_against_db(
+        self, test_config: FidesConfig, resources_dict: Dict, async_session
+    ):
+        endpoint = "dataset"
+        manifest: Dataset = resources_dict[endpoint]
+        dict_manifest = manifest.dict()
+        dict_manifest["collections"][0]["data_categories"] = ["bad_category"]
+
+        result = _api.upsert(
```
allisonking, line 174: do you think it's worth testing the create and update too?

src/fides/api/ctl/routes/crud.py
```diff
@@ -171,7 +205,7 @@ async def update(
         },
     )
     async def upsert(
-        resources: List[Dict],
+        resources: List[fides_model],
```
allisonking, line 208: ooh wow such a small change really makes a big difference!

allisonking (approved):
very clean change! 💯

tests/ctl/core/test_api.py
pattisdr: sure can, thanks Allison!

pattisdr:
Remaining failing fides db scan test has nothing to do with this branch.  Can address on `unified-fides-resources` branch if this is truly an issue.

PR #2148, opened 2023-01-05
Add missing permission for user deletion in Admin UI

RobertKeyser:
Fixes the list of permissions per user to include the ability to delete other users

https://github.com/ethyca/fides/blob/1af7a3201a761586585079903c99afe883309232/src/fides/api/ops/api/v1/scope_registry.py\#L68

### Code Changes

* [x] _list your code changes here_
* add `user:delete` to list of possible scopes to manage
* Set router dependencies to include the user deletion scope and remove the conditional to allow only the user to delete themselves or the root user to delete users.
* Rework unit test to check for user A/user B deletion (see below) instead of Admin/root or self

### Steps to Confirm

* [ ] Create user (`user A`) in Admin UI
* [ ] Log in as `user A`
* [ ] Create user (`user B`) in admin UI
* [ ] Try to delete `user B`

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### Description Of Changes

Before this, users created in the admin UI could not delete other users, because no such checkbox was available.


seanpreston:
Hey @RobertKeyser, can I confirm you've tested this?

I'm still seeing a 403 on my end here despite running your branch on a user with the delete permission:
Screenshot 2023-01-05 at 16 49 28

RobertKeyser:
🤦🏻 @seanpreston this time around I also removed the conditional that only allowed self or root user deletions. This prevents people from deleting their _own_ accounts, but I think it makes more sense to handle it this way instead.

RobertKeyser:
Moving to draft until I address the unit tests.

tests/ops/api/v1/endpoints/test_user_endpoints.py
```diff
@@ -340,10 +355,6 @@ def test_delete_user_as_root(self, api_client, db, generate_auth_header, user):
         )
         assert permissions_search is None
 
-        # Deleted user's client is also deleted
-        client_search = ClientDetail.get_by(db, field="id", value=client_id)
-        assert client_search is None
```
seanpreston, line 345: If this test was failing here that looks like a bug — we still want the deleted user's client removed.

src/fides/lib/oauth/api/routes/user_endpoints.py
```diff
@@ -107,12 +108,6 @@ def delete_user(
             status_code=HTTP_404_NOT_FOUND, detail=f"No user found with id {user_id}."
         )
 
-    if not (client.fides_key == ADMIN_UI_ROOT or client.user_id == user.id):
-        raise HTTPException(
-            status_code=HTTP_403_FORBIDDEN,
-            detail="Users can only remove themselves, or be the Admin UI Root User.",
-        )
-
     logger.info("Deleting user with id: '{}'.", user_id)
```
seanpreston, line 111: Now we're loosening the rules for this, it'd be useful to log which user performed this action, consider "User with id {} deleted by user with id {}".

tests/ops/api/v1/endpoints/test_user_endpoints.py
```diff
@@ -218,35 +218,41 @@ def test_create_user_wrong_scope(self, url, api_client, generate_auth_header, db
...
-        assert HTTP_403_FORBIDDEN == response.status_code
-
     def test_delete_nonexistent_user(self, api_client, db, generate_auth_header, user):
         auth_header = generate_auth_header([USER_DELETE])
         url = f"{V1_URL_PREFIX}{USERS}/nonexistent_user"
         response = api_client.delete(url, headers=auth_header)
         assert HTTP_404_NOT_FOUND == response.status_code
 
-    def test_delete_self(self, api_client, db, generate_auth_header):
+    def test_delete_user(self, api_client, db, generate_auth_header):
```
seanpreston, line 227: We still want a test for the case where a user deletes themselves. Consider moving these changes into a new test entirely, that tests explicitly what we want to test here (that a user can be deleted by another with the correct scope).

seanpreston (dismissed):
Thanks for looping back around here @RobertKeyser — just a couple more minor changes then it's good to go!

tests/ops/api/v1/endpoints/test_user_endpoints.py
RobertKeyser: Confused about what these lines do compared to the exact same thing a few lines above.

tests/ops/api/v1/endpoints/test_user_endpoints.py
RobertKeyser: As it stands, the user needs `user:delete` in order to delete themselves. This makes sense to me.

tests/ops/api/v1/endpoints/test_user_endpoints.py
seanpreston: Apologies — the diff didn't show that the code you removed was duplicate.

RobertKeyser:
@seanpreston, is that failing Frontend Cypress test something I should address?

ssangervasi:
> @seanpreston, is that failing Frontend Cypress test something I should address?

Nah looks like a flake

src/fides/lib/oauth/api/routes/user_endpoints.py
```diff
@@ -89,6 +89,7 @@ def create_user(
 @router.delete(
     urls.USER_DETAIL,
     status_code=HTTP_204_NO_CONTENT,
+    dependencies=[Security(verify_oauth_client, scopes=[USER_DELETE])],
```
ssangervasi, line 92: Just making sure the root user still passes this scope check -- the rooter user gets _every_ scope, right?

ssangervasi (approved):

Issue #2114, opened 2022-12-23
Remove the connector configuration dataset YAML upload form


allisonking:
### Is your feature request related to a specific problem?

There are two YAML forms that can be used to create datasets right now, one from the dataset management UI, and one from the connector configuration management UI. We should only have one now that we are [unifying the ops and ctl datasets](https://github.com/orgs/ethyca/projects/18/views/5?filterQuery=is%3Aopen+epic%3A%22Unified+Fides+Resources%22)

### Describe the solution you'd like

IMO it would make sense to remove the connector configuration YAML form:

<img width="1216" alt="Screen Shot 2022-11-14 at 10 08 02" src="https://user-images.githubusercontent.com/17103888/201694729-a39b0c3f-6d15-4855-8e1b-68505f4b148e.png">

And replace it with a multiselect that allows you to connect existing datasets to this connector. If they need to create a new one, then we should redirect them to the dataset management UI to create a dataset.

### Describe alternatives you've considered, if any

Leave as-is, but there are some caveats described in https://github.com/ethyca/fides/pull/2113 which make this UX a bit strange if we leave it.

### Additional context

Follow up to https://github.com/ethyca/fides/issues/1766

I'd like to get @rsilvery 's thoughts on how/if we should go about this. If we do, we may also need designs.

rsilvery:
I know we'd like to keep the more featured one. Working with @mfbrown to figure out how to sunset one of them.

rsilvery:
Most of this is being handled in https://github.com/ethyca/fides/issues/1531. We need to make sure that there's some protection to prevent overwriting datasets.

rsilvery:
Suggestion from @mfbrown : check for the existence of the fides key for the dataset and, if found, confirm user wants to overwrite.

allisonking:
https://github.com/ethyca/fides/issues/1531 will handle upgrading the dataset YAML form to use the same YAML editor as the connector one.

This ticket should focus on replacing the existing YAML form with a select dropdown of existing ctl datasets. If the dataset the user wants to link does not exist yet, we should redirect them to the adding a dataset flow.

allisonking:
Decision:

* Add a drop down to the Connectors dataset UI to select an existing dataset.
* If a dataset is selected in this menu, the editor is disabled. If it's unselected, the editor returns
* Once we've implemented this and can show it to design, get feedback and adapt as needed.

pattisdr:
Nice path forward here! ^

Issue #1386, opened 2022-10-11
Combine our RTK API's into one API instance for better cache re-fetching and tags


allisonking:
### Is your feature request related to a specific problem?

RTK [recommends](https://redux-toolkit.js.org/tutorials/rtk-query#create-an-api-service) only having one call to `createApi` and then injecting more endpoints off of the one instance. 

> Typically, you should only have one API slice per base URL that your application needs to communicate with. For example, if your site fetches data from both /api/posts and /api/users, you would have a single API slice with /api/ as the base URL, and separate endpoint definitions for posts and users. This allows you to effectively take advantage of [automated re-fetching](https://redux-toolkit.js.org/rtk-query/usage/automated-refetching) by defining [tag](https://redux-toolkit.js.org/rtk-query/usage/automated-refetching#tags) relationships across endpoints.
> 
> For maintainability purposes, you may wish to split up endpoint definitions across multiple files, while still maintaining a single API slice which includes all of these endpoints. See [code splitting](https://redux-toolkit.js.org/rtk-query/usage/code-splitting) for how you can use the injectEndpoints property to inject API endpoints from other files into a single API slice definition.

### Describe the solution you'd like

We should refactor our endpoints to this pattern so that it's easier to control cache invalidation across different endpoints. We have ~14 slice files that all create their own API instance.

### Describe alternatives you've considered, if any

_A description of any alternative solutions or features you've considered._

### Additional context

I came across this because I wanted the `/plus/scan` endpoint to invalidate the existing `datamap` cache, but they would need to be the same API instance first.

allisonking:
The unified-fides-resources branch already does some of this, so let's work on this after that feature branch is merged in.

allisonking:
> @allisonking is this still needed?

Yes! I've been picking at it as other features are built but there are still a good number of instances where we aren't doing this yet. This shouldn't be difficult to do though

allisonking:
Leaving a note to myself that I ran into this https://github.com/reduxjs/redux-toolkit/issues/2484, should be able to work around it though

PR #2162, opened 2023-01-06
Connector dataset dropdown

allisonking:
Closes https://github.com/ethyca/fides/issues/2114

### Code Changes

* [x] Adds a dataset dropdown next to the existing yaml editor
* [x] Allows the dataset dropdown to link a dataset to a connector config
* [x] Additional confirmation button for overwriting datasets in the yaml editor
* [x] Fix copy (system yaml --> dataset YAML)
* [x] Cypress tests

### Steps to Confirm

* [ ] `nox -s test_env`
* [ ] Configure a connector dataset

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### Description Of Changes

Using the dropdown

https://user-images.githubusercontent.com/24641006/211687577-d15544a6-4ed1-4fd8-b27b-91840857f52b.mov

Using the yaml editor

https://user-images.githubusercontent.com/24641006/211687670-72ddcd58-c9b9-4319-b864-659f38f8e692.mov

### DIFF
--- a/clients/admin-ui/src/features/datastore-connections/add-connection/DatasetConfiguration.tsx
+++ b/clients/admin-ui/src/features/datastore-connections/add-connection/DatasetConfiguration.tsx
@@ -55,7 +55,7 @@ const DatasetConfiguration: React.FC = () => {
   >(undefined);
 
   useEffect(() => {
-    if (data) {
+    if (data && data.items.length) {
       setSelectedDatasetKey(data.items[0].ctl_dataset.fides_key);
     }
   }, [data]);
--- a/clients/admin-ui/src/features/datastore-connections/add-connection/DatasetConfiguration.tsx
+++ b/clients/admin-ui/src/features/datastore-connections/add-connection/DatasetConfiguration.tsx
@@ -17,7 +17,7 @@ import {
 } from "datastore-connections/datastore-connection.slice";
 import { PatchDatasetsConfigRequest } from "datastore-connections/types";
 import { useRouter } from "next/router";
-import React, { ChangeEvent, useState } from "react";
+import React, { ChangeEvent, useEffect, useState } from "react";
 import { DATASTORE_CONNECTION_ROUTE } from "src/constants";
 
 import { useAppSelector } from "~/app/hooks";
@@ -45,7 +45,7 @@ const DatasetConfiguration: React.FC = () => {
   const { data, isFetching, isLoading, isSuccess } = useGetDatasetConfigsQuery(
     connection!.key
   );
-  const [patchDataset] = usePatchDatasetConfigsMutation();
+  const [patchDatasetConfig] = usePatchDatasetConfigsMutation();
   const [upsertDatasets] = useUpsertDatasetsMutation();
   const { data: allDatasets, isLoading: isLoadingAllDatasets } =
     useGetAllDatasetsQuery();
@@ -53,6 +53,45 @@ const DatasetConfiguration: React.FC = () => {
   const [selectedDatasetKey, setSelectedDatasetKey] = useState<
     string | undefined
   >(undefined);
+
+  useEffect(() => {
+    if (data) {
+      setSelectedDatasetKey(data.items[0].ctl_dataset.fides_key);
+    }
+  }, [data]);
+
+  const handlePatchDatasetConfig = async (
+    datasetPairs: DatasetConfigCtlDataset[]
+  ) => {
+    const params: PatchDatasetsConfigRequest = {
+      connection_key: connection?.key as string,
+      dataset_pairs: datasetPairs,
+    };
+    const payload = await patchDatasetConfig(params).unwrap();
+    if (payload.failed?.length > 0) {
+      errorAlert(payload.failed[0].message);
+    } else {
+      successAlert("Dataset successfully updated!");
+    }
+    router.push(DATASTORE_CONNECTION_ROUTE);
+  };
+
+  const handleLinkDataset = async () => {
+    if (selectedDatasetKey) {
+      try {
+        let fidesKey = selectedDatasetKey;
+        if (data && data.items.length) {
+          fidesKey = data.items[0].fides_key;
+        }
+        const datasetPairs: DatasetConfigCtlDataset[] = [
+          { fides_key: fidesKey, ctl_dataset_fides_key: selectedDatasetKey },
+        ];
+        handlePatchDatasetConfig(datasetPairs);
+      } catch (error) {
+        handleError(error);
+      }
+    }
+  };
 
   const handleSubmitYaml = async (value: unknown) => {
     try {
@@ -85,17 +124,7 @@ const DatasetConfiguration: React.FC = () => {
         }));
       }
 
-      const params: PatchDatasetsConfigRequest = {
-        connection_key: connection?.key as string,
-        dataset_pairs: pairs,
-      };
-      const payload = await patchDataset(params).unwrap();
-      if (payload.failed?.length > 0) {
-        errorAlert(payload.failed[0].message);
-      } else {
-        successAlert("Dataset successfully updated!");
-      }
-      router.push(DATASTORE_CONNECTION_ROUTE);
+      handlePatchDatasetConfig(pairs);
     } catch (error) {
       handleError(error);
     } finally {
@@ -133,6 +162,7 @@ const DatasetConfiguration: React.FC = () => {
                   width="fit-content"
                   placeholder="Select"
                   onChange={handleSelectDataset}
+                  value={selectedDatasetKey}
                 >
                   {allDatasets.map((ds) => (
                     <option key={ds.fides_key} value={ds.fides_key}>
@@ -146,6 +176,7 @@ const DatasetConfiguration: React.FC = () => {
                 colorScheme="primary"
                 alignSelf="start"
                 disabled={!datasetSelected}
+                onClick={handleLinkDataset}
               >
                 Save
               </Button>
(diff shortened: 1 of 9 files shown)


allisonking:
Failing test seems unrelated?

pattisdr:
yep this is unrelated @allisonking, I saw this on my last `unified-fides-resources` PR but I don't think it's on main. After we merge your branch in, I'll get `unified-fides-resources` up-to-date with main and investigate there.

pattisdr (changes requested):
This feels a little buggy to me in places. This branch isn't up-to-date with `unified-fides-resources` that has the [improved upsert validation](https://github.com/ethyca/fides/pull/2134), but that is handy here so it's easy to see where we might add better error handling: 


1) Start from scratch, delete datasetconfigs, connectionconfigs, and ctl_datasets. Refresh.  It makes sense that there's no dropdown to connect my first connection config to an existing dataset, but the save button stacked on the cancel button looks off
Screen Shot 2023-01-11 at 12 39 39 PM
2)  Copy and paste a dataset (mysql is a good one) that doesn't have an `organization_fides_key` or `data_qualifier` in it.  You could also send in a bad data category.  We get failed requests but the UI looks like we never submitted a request at all. (In the background a ctl_dataset has been created, but the ops datasetconfig was not created)

https://user-images.githubusercontent.com/9755598/211892518-7d8c0b07-ce39-4262-a2c9-f6b0159d6cac.mov

3) Editing an existing dataset that was good to something invalid (add bad data category) also puts the UI in a weird state
Screen Shot 2023-01-11 at 12 57 39 PM

4) When saving existing dataset yaml you've added a helpful confirmation that you're about to overwrite a dataset.  This is not necessarily true if the user is also editing the `fides_key`.
Below, the only edit I made was to change `postgres_example_test_dataset` fides_key to `updated_fides_key`.  This will create a new `ctl_dataset` titled `updated_fides_key`, and link it to the DatasetConfig.


Screen Shot 2023-01-11 at 1 03 11 PM


5) When configuring an existing dataset in the connector workflow, it took me awhile to figure out how to edit directly.  I needed to unselect the current dataset.

allisonking:
Thanks for the thorough testing @pattisdr ! 🙏 

1. This should be fixed. It's a little weird code-wise, but the new designs don't have a cancel button anymore so I don't think we have to worry too much about it since it'll go away soon.
image

2 and 3. I'm having trouble reproducing these—I just merged in `unified-fides-resources` which like you said throws the proper errors now, so I wonder if that's letting me see good errors instead of getting stuck in weird states? Can you let me know if you're still seeing these problems, and if so, with what dataset? (I tried out `mysql_example_test_dataset.yml` which seemed to work ok)

4. Good catch, should be fixed!
5. Yeah, this was the decision we came to when we were trying to decide on behavior without design. The good news is this will go away with the new design that's currently in the works

clients/admin-ui/src/features/datastore-connections/add-connection/forms/YamlEditorForm.tsx
```diff
@@ -93,7 +93,12 @@ const YamlEditorForm: React.FC<YamlEditorFormProps> = ({
     // Only need the confirmation if we are overwriting, which only happens when
     // there is already data
     if (data.length) {
-      warningDisclosure.onOpen();
+      const value: string = (monacoRef.current as any).getValue();
+      // Check if the fides key that is in the editor is the same as the one that already exists
+      // If so, then it is an overwrite and we should open the confirmation modal
+      if (data.some((d) => value.includes(`fides_key: ${d.fides_key}\n`))) {
+        warningDisclosure.onOpen();
+      }
```
pattisdr, line 101: I think the check is really does this `fides_key` match any fides_key in the ctl_datasets table? Because we could still update it to match a different ctl_dataset and that would be even more unintuitive. 

Regardless, if you update the key, it now looks like you don't have an option to save anything at all, you just click the save button and nothing happens.   It's probably safer if they can't update the fides_key without a warning, but the UI is in a weird state still.

pattisdr:
2 and 3. I'm having trouble reproducing these—I just merged in unified-fides-resources which like you said throws the proper errors now, so I wonder if that's letting me see good errors instead of getting stuck in weird states? Can you let me know if you're still seeing these problems, and if so, with what dataset? (I tried out mysql_example_test_dataset.yml which seemed to work ok)


OK after some more digging I found what bad request is causing this, it's if a `GET http://localhost:3000/api/v1/dataset` throws an error because it's returning a bad dataset.   It's harder to create a bad dataset after you've merged in `unified-fides-resources`.  

To reproduce, you could force the backend to throw an error, for example:
1) Go to src/fides/api/ctl/routes/crud.py
2) In the list endpoint `async def ls`:, raise an exception first thing Exception()

This mimics what would happen if a user had a bad dataset already saved.

clients/admin-ui/src/features/datastore-connections/add-connection/forms/YamlEditorForm.tsx
```diff
@@ -115,20 +118,21 @@ const YamlEditorForm: React.FC<YamlEditorFormProps> = ({
...
+            isLoading={isSubmitting}
+            loadingText="Saving"
+            onClick={handleConfirmation}
+            type="submit"
+            data-testid="save-yaml-btn"
+            width="fit-content"
+          >
+            Save
+          </Button>
+        </ButtonGroup>
```
pattisdr, line 140: Bravo this looks great

clients/admin-ui/src/features/datastore-connections/add-connection/forms/YamlEditorForm.tsx
allisonking: ah I see, okay yes there are two things going on here, thanks for catching that!

pattisdr:
wonderful, testing now!

clients/admin-ui/src/features/datastore-connections/add-connection/DatasetConfiguration.tsx
```diff
@@ -155,6 +162,11 @@ const DatasetConfiguration: React.FC = () => {
 
   return (
     <VStack alignItems="left">
+      {loadAllDatasetsError ? (
+        <Copy mb={4}>
+          There was a problem loading existing datasets, please try again.
```
pattisdr, line 167: OK great improvements if there are errors loading bad datasets!

I didn't really notice the line of text though: `There was a problem loading existing datasets, please try again.` as it didn't stand out, it's not clear there was an error.  If you're going to add text directly to the page, could it be red, or perhaps a toast message instead?
Screen Shot 2023-01-12 at 11 07 40 AM

clients/admin-ui/src/features/datastore-connections/add-connection/forms/YamlEditorForm.tsx
```diff
@@ -91,17 +93,19 @@ const YamlEditorForm: React.FC<YamlEditorFormProps> = ({
...
+    // there are already datasets
+    if (allDatasets && allDatasets.length) {
       const value: string = (monacoRef.current as any).getValue();
-      // Check if the fides key that is in the editor is the same as the one that already exists
+      // Check if the fides key that is in the editor is the same as one that already exists
       // If so, then it is an overwrite and we should open the confirmation modal
-      if (data.some((d) => value.includes(`fides_key: ${d.fides_key}\n`))) {
+      if (
+        allDatasets.some((d) => value.includes(`fides_key: ${d.fides_key}\n`))
+      ) {
```
pattisdr, line 103: The message in the warning modal may not always be technically correct: the ctl_dataset that we're about to upsert, is going to match whatever the `fides_key` is in the first line of the yaml file. 

So here, if the dataset was originally `my_other_mongo_test` and I have just pasted yaml that has `mongo_test`, it is `mongo_test` that is going to be either created or updated. 

Screen Shot 2023-01-12 at 11 12 43 AM

pattisdr (commented):
OK this UI is feeling pretty good now and holding up under a lot of scenarios I'm testing. Thanks for your nice changes here! There's just a couple of things I saw on this last round but I think it's about ready to go.

clients/admin-ui/src/features/datastore-connections/add-connection/DatasetConfiguration.tsx
allisonking: sure thing!

image

pattisdr:
@allisonking one more thing I just thought of I forgot to test: creating saas connectors.  One big difference with saas connectors is that we create them from a template.  After the user enters their connector parameters and clicks save, the connection config, dataset config, and dataset are all created in the background for them. 

There's a small thing where it's hard to save the dataset when you create multiple saas connectors.
- Create a mailchimp connector
- Enter connector parameters (just fake data is fine)
- On configure dataset click save in the dropdown
- Create a segment connector
- Enter connector parameters
- On configure dataset, the segment dataset is not in the dropdown.  I can click "Save" here, but it's confusing because there's only "Select" in the dropdown. I can't edit the yaml either: by toggling to another dataset, then back to Select, you can edit the yaml.  I suspect the "Select" is tied to the actual dataset here behind the scenes. 


https://user-images.githubusercontent.com/9755598/212152362-2ac27733-b698-444d-b2a5-7233d2091c2e.mov

allisonking:
Oh man, that's a good catch I never would have looked for. [This should fix it](https://github.com/ethyca/fides/pull/2162/commits/19551b0cdb4a51da0969bcd12e48460b12c95450). The change really should have only been the one line (to add "Datasets" to the list of resources in the cache to be invalidated when a new saas connector is created), but [due to some tech debt](https://github.com/ethyca/fides/issues/1386) I had to combine the relevant redux slices to use the same base API first.

pattisdr (approved):
Alright this looks really good to me @allisonking thanks for your hard work here 🏆  Failing test is unrelated like we discussed and I'll look into this after you merge this branch into `unified-fides-resources`.

Issue #2135, opened 2023-01-05
[Epic] Pytest Improvements


ThomasLaPiana:
## Objective

The goal of this Epic is to refactor our backend automated testing, specifically our `pytest` suite, to a point where we can both guarantee quality now as well as make that level of quality maintainable into the future.

## Links

* [Original Confluence Design Doc](https://ethyca.atlassian.net/wiki/spaces/EN/pages/2571370641/2022-12-20+-+Pytest+Automation+Refactor)
* [Code Climate](https://codeclimate.com/quality) (tool for tracking testing coverage, free for OSS)

## Issues (in order of recommended execution)
* [x] #2179
* [x] #2174
* [ ] #2175
* [x] #2176
* [ ] #2177
* [ ] #2178
* [x] #2180
* [x] #2181
* [x] #2182
* [x] #2183

sanders41:
I did a little looking into why tests are running slow. The slowest unit test is currently `test_get_filtered_users` which takes ~ 38 seconds. I ran a cProfile on the tests and a couple things stand out to me. First it looks like in general the collection/setup is relatively slow so we could probably make improvements there that would benefit all tests.

For this specific test it looks like the slowest part is acquiring thread locks for post requests. My current theory here (but unproven) is the lock is acquired because the endpoint is synchronous which FastAPI runs in a thread pool requiring the lock. Maybe if we make the endpoint async and use the async db connection this would remove the need for a lock and speed things up?

I'm attaching the cProfile graph. I'm not sure if it will be zoomable here or not, if not and anyone wants it I can send it in slack and/or show you how I created it. 

![output](https://user-images.githubusercontent.com/25045024/211554145-a999cb15-1cc1-4937-a4ed-13898e72e41f.png)

sanders41:
I have been thinking about our xdist discussion and the database issues around that. One thing we could explore is adding new `database` mark for tests that access the database, then run them in their own job without xdist while using xdist for all other tests. This would allow us to run at least some of the tests in parallel, I'm just not sure yet if it would be a tiny percentage of the tests and therefore not make much difference.

ThomasLaPiana:
@sanders41 thanks for doing that analysis!

I think we want everything to be async, right? Any reason not to? I remember it being some work to get the `ctl` stuff all async but now that we have that done I don't think it would be too hard to port that to the other endpoints? If so, it sounds like we could fix our application performance and get faster tests for free! 

[...]

PR #2172, opened 2023-01-09
Exclude untestable code from coverage

sanders41:
Closes #<issue>

### Code Changes

* Added `pragma: no cover` to `if __name__ == "__main__:"` code
* Excluded migrations from coverage 

### Steps to Confirm

* [ ] Run `pytest tests/ctl -m "unit"` and verify `migrations` and `if __name__ == "__main__:"` sections are not included in the `term-missing` section

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

In thinking about #2135 I am adding some exclusions from coverage here. The `migrations` are used by alembic and not by fides code directly, the only reason we have them in a directory under coverage is so plus can access them.

We could probably jump through some hoops and find a way to test the ``if __name__ == "__main__:"`` sections, but doing so would be for the sake of coverage and not meaningful.

### DIFF
--- a/src/fides/api/ops/graph/data_type.py
+++ b/src/fides/api/ops/graph/data_type.py
@@ -265,7 +265,7 @@ def get_data_type(value: Any) -> Tuple[Optional[str], bool]:
     return data_type, is_array
 
 
-if __name__ == "__main__":
+if __name__ == "__main__":  # pragma: no cover
     v = DataType.no_op.value
     for x in v.__dict__:
         print(x)
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -204,3 +204,6 @@ markers = [
     "unit_saas"
 ]
 asyncio_mode = "auto"
+
+[tool.coverage.run]
+omit = ["src/fides/api/ctl/migrations/*"]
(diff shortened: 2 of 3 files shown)


src/fides/api/ops/graph/data_type.py
```diff
@@ -265,7 +265,7 @@ def get_data_type(value: Any) -> Tuple[Optional[str], bool]:
     return data_type, is_array
 
 
-if __name__ == "__main__":
+if __name__ == "__main__":  # pragma: no cover
```
sanders41, line 268: Instead of adding `no cover` to each `if __name__ == "__main__":` we should be able to add it to the `pyproject.toml` to automatically ignore all of them, but in doing this it was only ignoring the `if __name__ == "__main__":` line and not the code inside the if block. I've never had this happen before, and couldn't figure out why it was happening here so I went with a `pragma: no cover` as an alternative.

```toml
[tool.coverage.report]
exclude_lines = ["if __name__ == .main__.:", "pragma: no cover"]
```

ThomasLaPiana (approved):
Thanks for this!

PR #2187, opened 2023-01-10
[DRAFT] Mandrill consent connector

### LINKED ISSUE #2150: Hard-coded Mailchimp Transactional (Mandrill) consent request execution
In https://github.com/ethyca/fides/issues/2144 we identify a format for specifying consent request generation and execution. Before we implement that, we'd like to hook in a string of requests off to Mandril in the [`run_consent_request`](https://github.com/ethyca/fides/pull/2125/files#diff-35fed73a3a47b2a6c6e7d3d413edf8c493406cb51aceb07a28025574fd293c3fR407) function. The benefits of this are that it:
- buys us some time to finalise the spec so that it fits into the existing SaaS connector DDL well
- gives us a higher chance of having something demo-able by the end of this sprint

pattisdr:
Closes #2150
### Code Changes

- Made this a type of saas connector
- Experimental format explored in mandrill saas config, mandrill dataset is mostly empty
- SaasConnector.run_consent_request has special casing just for mandrill.  Any other consent requests for other saas connector types are skipped.

### Steps to Confirm

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

Exploring Mandrill consent connector

### DIFF
--- a/data/saas/config/mandrill_config.yml
+++ b/data/saas/config/mandrill_config.yml
@@ -1,0 +1,44 @@
+saas_config:
+  fides_key: <instance_fides_key>
+  name: Mandril SaaS Config
+  type: mandrill
+  description: A sample schema representing the Mandril connector for Fided
+  version: 0.0.1
+
+  connector_params:
+    - name: domain
+    - name: api_key
+
+  client_config:
+    protocol: https
+    host: <domain>
+
+  consent_requests:
+    opt_in:
+      - method: POST
+        path: /allowlists/add
+        body: |
+          {
+            "key": "<api_key>",
+            "email": "<email>"
+          }
+        data_uses: [advertising.first_party]
+    opt_out:
+      - method: POST
+        path: /allowlists/delete
+        body: |
+          {
+            "key": "<api_key>",
+            "email": "<email>"
+          }
+        data_uses: [ advertising.first_party]
+
+  test_request:
+    method: GET
+    path: /users/ping
+    body: |
+      {
+        "key": "<api_key>",
+      }
+
+  endpoints: []
--- a/data/saas/dataset/mandrill_dataset.yml
+++ b/data/saas/dataset/mandrill_dataset.yml
@@ -1,0 +1,5 @@
+dataset:
+  - fides_key: <instance_fides_key>
+    name: Mandril Dataset
+    description: A sample dataset representing the Mandril connector for Fides
+    collections: []
--- a/data/saas/saas_connector_registry.toml
+++ b/data/saas/saas_connector_registry.toml
@@ -52,6 +52,12 @@ dataset = "data/saas/dataset/mailchimp_dataset.yml"
 icon = "data/saas/icon/mailchimp.svg"
 human_readable = "Mailchimp"
 
+[mandrill]
+config = "data/saas/config/mandrill_config.yml"
+dataset = "data/saas/dataset/mandrill_dataset.yml"
+icon = "data/saas/icon/default.svg"
+human_readable = "Mandrill"
+
 [outreach]
 config = "data/saas/config/outreach_config.yml"
 dataset = "data/saas/dataset/outreach_dataset.yml"
--- a/src/fides/api/ops/schemas/saas/saas_config.py
+++ b/src/fides/api/ops/schemas/saas/saas_config.py
@@ -104,6 +104,7 @@ class SaaSRequest(BaseModel):
     grouped_inputs: Optional[List[str]] = []
     ignore_errors: Optional[bool] = False
     rate_limit_config: Optional[RateLimitConfig]
+    data_uses: Optional[List[str]] = []
 
     class Config:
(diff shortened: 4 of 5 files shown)


data/saas/config/mandrill_config.yml
```diff
@@ -0,0 +1,44 @@
...
+        data_uses: [advertising.first_party]
+    opt_out:
+      - method: POST
+        path: /allowlists/delete
+        body: |
+          {
+            "key": "<api_key>",
+            "email": "<email>"
+          }
+        data_uses: [ advertising.first_party]
```
pattisdr, line 34: To me it seems like the saas config needs to hold the distinction between:
1) Whether the endpoint is for an opt_in or opt_out request
2) Which data_uses apply

data/saas/dataset/mandrill_dataset.yml
```diff
@@ -0,0 +1,5 @@
+dataset:
+  - fides_key: <instance_fides_key>
+    name: Mandril Dataset
+    description: A sample dataset representing the Mandril connector for Fides
+    collections: []
```
pattisdr, line 5: The dataset here is a placeholder to borrow Adam's terminology.  I think the fastest path forward with this is making Mandrill a saas connector so we can still utilize the same connector workflow + database tables.

So Mandrill here doesn't need collections outlined, but saas connectors that have both access and consent for example, would outline collections.

data/saas/config/mandrill_config.yml
```diff
@@ -0,0 +1,44 @@
...
+  consent_requests:
+    opt_in:
+      - method: POST
+        path: /allowlists/add
+        body: |
+          {
+            "key": "<api_key>",
+            "email": "<email>"
+          }
+        data_uses: [advertising.first_party]
```
pattisdr, line 25: advertising.first_party = 'Email marketing'?

Do we have to worry about customers using different consent preferences in the `config.json` that differ from data uses we spell out in the saas config (or elsewhere)?

data/saas/config/mandrill_config.yml
pattisdr, line 34: Are data uses hierarchical? For example, if advertising is deemed not executable, would advertising.first_party automatically not being executable? 

(That is not how my POC treats it, I assume they are separate data uses)

data/saas/config/mandrill_config.yml
```diff
@@ -0,0 +1,44 @@
...
+    - name: domain
+    - name: api_key
+
+  client_config:
+    protocol: https
+    host: <domain>
+
+  consent_requests:
+    opt_in:
+      - method: POST
```
pattisdr, line 18: Still broader unknowns about how many individual requests about how many requests will be needed.  Here, I'm allowing a list of opt-in requests for different data uses and a list of opt out requests for different data uses.

data/saas/config/mandrill_config.yml
```diff
@@ -0,0 +1,44 @@
+saas_config:
+  fides_key: <instance_fides_key>
+  name: Mandril SaaS Config
+  type: mandrill
```
NevilleS, line 4: One thing right up front, which is annoying but we should do now: let's start calling this Mailchimp 😄 

Mandrill doesn't actually exist any more, all the API docs are "Mailchimp Transactional Email" like this: https://mailchimp.com/developer/transactional/api/

data/saas/config/mandrill_config.yml
NevilleS: I think I'd prefer to avoid encoding any `fideslang` taxonomy types in the SaaS config, so it can remain purely, un-opinionated HTTP details.

Instead, let's think about how we'd customize the connector itself with whether it should or shouldn't be executed. The most primitive version of this, for example, would be a config setting on the connector config.

data/saas/config/mandrill_config.yml
NevilleS: Yes, they are. We should assume that, for a "policy"-type matching rule, any child nodes of a parent node in a `fideslang` taxonomy should match

pattisdr: leave opt in and opt out sections remove data uses

data/saas/config/mandrill_config.yml
pattisdr: That makes sense @NevilleS thanks for making this distinction.

data/saas/config/mandrill_config.yml
pattisdr: yikes, what does that mean if someone says they want to opt-out of `advertising` but opt-in to `advertising.first_party`, that's a conflict. luckily this first version won't be looking at data uses while we evaluate this further.

data/saas/config/mandrill_config.yml
pattisdr: That's perfect, fewer files to create, we can just add the consent section to existing mailchimp config

PR #2193, opened 2023-01-10
Executable consent types for the privacy center

allisonking:
Closes https://github.com/ethyca/fides/issues/2122

### Code Changes

* [x] Set default `config.json` `executable`s to `false`, both in the privacy center directory, and the `test_env` privacy center directory
* [x] Remove `load_executable_consent_options` and references
* [x] Pass `policy_key` and a list of `executable_options` from the frontend to `/consent-request/{{consent_request_id}}/preferences`
* [x] Update `queue_privacy_request_to_propagate_consent` to make PrivacyRequest objs based on the list of passed in `executable_options`
* [x] Add an error message when there are multiple `executable: true` in the `config.json`

### Steps to Confirm

* [ ] Update `test_env/privacy_center_config/config.json` to have one consent option be `executable: true`
* [ ] `nox -s test_env` 
* [ ] At localhost:3001, change your consent preferences. You should see the a new payload with `executable_options` sent to the endpoint if you look in the Network tab
* [ ] The privacy request should be saved with all of the consent preferences listed in config.json
* [ ] Update `test_env/privacy_center_config/config.json` to have more than one consent option be `executable: true`
* [ ] `nox -s test_env`
* [ ] Upon loading the privacy center, you should see an error message that you can't have more than one consent option set to `executable`

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### Description Of Changes

Endpoint now takes `executable_options` and privacy center passes those in
image

And there's also this error message, though I'm not quite sure it belongs in a toast?
image

I think the changes I made are mostly covered by existing test cases, but lemme know if there are other tests I should add!


src/fides/api/ops/api/v1/endpoints/consent_request_endpoints.py
```diff
@@ -340,6 +316,7 @@ def set_consent_preferences(
         else:
             preference_dict = dict(preference)
             preference_dict["provided_identity_id"] = provided_identity.id
+            preference_dict.pop("executable")
```
allisonking, line 319: I felt like I had to do some weird things because `executable` was in `ConsentSchema` but not the ORM `Consent` object

src/fides/api/ops/api/v1/endpoints/consent_request_endpoints.py
```diff
@@ -502,22 +479,37 @@ def _get_consent_request_and_provided_identity(
...
     ).all()
 
-    if not consent:
+    if not consents_matching_identity:
         return ConsentPreferences(consent=None)
 
-    return ConsentPreferences(
-        consent=[
+    consent_list = []
+    for consent in consents_matching_identity:
```
allisonking, line 500: this is the part I'm most unsure about. it seems like we want to merge what the client gives us (which has `executable`) with the Consent objects from the db (which don't have `executable`). it feels a lil weird to me, so I would love some feedback!

allisonking:
@pattisdr Still working on this (especially the backend tests) but would appreciate if you think this is the right approach so far!

src/fides/api/ops/api/v1/endpoints/consent_request_endpoints.py
pattisdr: Yes, good point, this is not feeling ideal given the number of locations you've had to change.  

So a new approach: What about if we instead did something like:

- Left the `ConsentSchema` untouched
- Modified the `ConsentPreferencesWithVerificationCode` schema to have a new `executable_options` key
- `executable_options` might just be a list of new `ExecutableConsent` (new schema) with a data_use and executable (bool) fields.
- Pass in the executable options to`queue_privacy_request_to_propagate_consent` 
- Use those instead of what I was pulling from the config.json to filter which Consent options are saved on the privacy request
- I'm thinking we still just save the executable preferences to the PrivacyRequest for now.
- Still don't queue the privacy request if we don't have executable preferences

Then hopefully the processing of saving Consent records to the db is untouched and _prepare_consent_preferences is untouched and the PrivacyRequest still has just executable preferences.

src/fides/api/ops/api/v1/endpoints/consent_request_endpoints.py
pattisdr: I agree, let's try to adjust to touch less code. Maybe https://github.com/ethyca/fides/pull/2193/files#r1067069764

pattisdr (changes requested):
Nice changes Allison.  A few bugs and questions here.

Responding to:
> And there's also this error message, though I'm not quite sure it belongs in a toast?

Hm, this is tricky.  I was hoping we could get an invalid `config.json` to throw an error on next build so the customer is made aware.  As-is, this doesn't prevent all executables=true from sending a request.

If we can't prevent it from building, or similar, perhaps they get a message that there's something wrong with the system, and we don't allow the consent request to be saved from the UI in the first place.  Just brainstorming here.

src/fides/api/ops/api/v1/endpoints/consent_request_endpoints.py
```diff
@@ -267,15 +242,15 @@ def queue_privacy_request_to_propagate_consent(
         provided_identity.encrypted_value["value"],  # type:ignore[index]
     )  # Pull the information on the ProvidedIdentity for the ConsentRequest to pass along to create a PrivacyRequest
 
-    executable_consent_options: List[str] = load_executable_consent_options(
-        CONFIG_JSON_PATH
-    )
+    executable_data_uses = [ec.data_use for ec in executable_consents or []]
```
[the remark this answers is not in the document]
allisonking: OH okay this is just me totally misinterpreting the requirements. will add that check back 👍

src/fides/api/ops/api/v1/endpoints/consent_request_endpoints.py
```diff
@@ -267,15 +242,15 @@ def queue_privacy_request_to_propagate_consent(
...
     executable_consent_preferences: List[Dict] = [
         pref.dict()
         for pref in consent_preferences.consent or []
-        if pref.data_use in executable_consent_options
+        if pref.data_use in executable_data_uses
     ]
 
+    logger.info(executable_consent_preferences)
```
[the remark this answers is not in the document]
allisonking: whoops 🧹

allisonking:
Okay the backend fixes are in. Still mulling over how best to show an error when there are multiple `executable=true`s

allisonking:
Okay, I ended up putting some logic that gets called during the build stage and I think it works! To test:

`npm run dev` and `npm run build` should still work

But if you change the `config.json` such that there is more than one consent option with `executable=True`, when you `npm run dev` or `npm run build` you should see:

`npm run dev`:
image


`npm run build`:
image

The build one isn't as clean unfortunately, but it's nice that it runs through the same logic as `npm run dev` so we don't need separate cases for both.

The only drawback is you won't get immediate feedback when you change your `config.json` (if you're hot reloading in dev) that it's wrong, but maybe that's okay. If we really want that, we can restore the toast I had earlier

PR #2198, opened 2023-01-11
Send pytest coverage reports to code climate

### LINKED ISSUE #2179: Automate generation of a single number that represents total test coverage

### LINKED ISSUE #2180: Every PR includes an automated comment of test coverage
This is waiting on #2179 and then needs to be verified and confirmed that `Code Climate` will provide an accurate number on each commit.

ThomasLaPiana:
Closes #2179 

### Code Changes

* [x] refactor nox sessions to use a single parent `pytest` session that calls the other test functions
* [x] run `isort(fix)` and `black(fix)` first so that `nox -s static_checks` will try to fix files before checking them
* [x] separate the strictly `pytest` CI checks from the general backend checks to streamline the jobs that upload code coverage
* [x] update the test functions to write out coverage files based on their name
* [x] add a script that fixes the coverage files for Code Climate consumption
* [x] update CI jobs to upload those artifact files
* [x] send coverage reports to `Code Climate` in CI ([CC Docs](https://docs.codeclimate.com/docs/configuring-test-coverage))
  * [x] store each coverage report in a temporary storage destination ([Github Artifacts](https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts))
  * [x] pull them all in in a final job and send to Code Climate 
* [x] add badges to the README

### Steps to Confirm

* [ ] coverage reports show up in Code Climate
* [ ] _safe_ `pytest` jobs produce and upload a coverage file (4 tests)

### Pre-Merge Checklist

* [ ] All CI Pipelines Succeeded (external checks are known-failing)
* Documentation:
* [x] Update `CHANGELOG.md`

### Description Of Changes

__Disclaimer__: This change was meant to be relatively small, but in true "give a mouse a cookie" fashion it snow-balled into something much larger. Generally the changes are quite safe and is mostly all new functionality as opposed to modifying existing functionality. 

__Disclaimer 2__: This change only handles uploading the coverage reports to `Code Climate`, there is still a bug preventing the code comment from updating on every push to the PR that will need to be solved in a PR for #2180 

This PR adds the functionality that allows for continual tracking of overall code coverage via a 3rd party. This required a lot of engineering in terms of CI/scripting and has also touched how tests are run via `nox` sessions.

__As a caveat, we are not including external test coverage as part of this work.__

#### Nox changes

The `nox` suite of test sessions now looks like this:

```
- pytest(ctl-unit) -> Runs Pytests.
- pytest(ctl-not-external) -> Runs Pytests.
- pytest(ctl-integration) -> Runs Pytests.
- pytest(ctl-external) -> Runs Pytests.
- pytest(ops-unit) -> Runs Pytests.
- pytest(ops-integration) -> Runs Pytests.
- pytest(ops-external-datastores) -> Runs Pytests.
- pytest(ops-saas) -> Runs Pytests.
- pytest(lib) -> Runs Pytests.
```

Everything has been converted to a single `pytest` session with individual testing groups as a parameter.


ThomasLaPiana:
At this point I've gotten the coverage reports writing out successfully, and am now toying with the proper CI syntax to get everything to upload

One potential stumbling block is how to handle the fact that we only _sometimes_ run the unsafe CI checks...I'm not sure yet if this will throw false positives about code coverage if those don't get run

ThomasLaPiana:
After some more testing and using [this comment](https://github.com/paambaati/codeclimate-action/issues/269#issuecomment-839050964) as an example, it appears to be an issue caused by running the tests within a docker file.

Going to try some things locally such as setting a specific source file and seeing if that helps things

ThomasLaPiana:
I noticed that it was saying that `fides` didn't exist in the python 3.8 python path, which seemed odd because I had specifically installed python 3.10. It dawned on me that maybe the third-party github action we're using wouldn't respect that, so I removed that CI task and want to see if it works just using the default python path

ThomasLaPiana:
I'm getting things working directly in the docker container, so I'm going to ditch the github action and try it directly as a `nox` session that will upload from within the container

ThomasLaPiana:
got it working locally with the following observations:
* ran it first with just the `lib` test coverage file existing, the report shows 16% coverage (sounds about right)
* ran the full `upload_coverage` session again after also running ctl unit tests and coverage jumped to 56%, so this at least shows that it does properly format/combine the test files to get that larger number

.github/workflows/backend_checks.yml
```diff
@@ -282,4 +317,34 @@ jobs:
...
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Download Coverage Artifacts
+        uses: actions/download-artifact@v3
+        with:
+          path: ${{ github.workspace }}/coverage/
+
+      - name: Display Downloaded Artifacts
+        run: ls -R coverage/
```
sanders41, line 342: Looking at the output here I wonder if part of the problem is the file/structure. There are more files here than I would expect and the ones I do expect look to be in subdirectories.

```console
coverage/:
ctl-not-external
lib
ops-integration
ops-unit
python-3.10.7
python-3.8.14
python-3.9.14

coverage/ctl-not-external:
ctl-not-external.lcov

coverage/lib:
lib.lcov

coverage/ops-integration:
ops-integration.lcov

coverage/ops-unit:
ops-unit.lcov

coverage/python-3.10.7:
python-3.10.7.tar

coverage/python-3.8.14:
python-3.8.14.tar

coverage/python-3.9.14:
python-3.9.14.tar
```

noxfiles/ci_nox.py
```diff
@@ -357,3 +221,32 @@ def python_build(session: nox.Session, dist: str) -> None:
         dist,
         external=True,
     )
+
+
+@nox.session()
+def upload_coverage(session: nox.Session) -> None:
+    """
+    Format, combine and upload the test coverage files.
```
sanders41, line 229: Should we add a note here that this won't work locally because of the need for the upload key? I think it's fine for it to not work locally because we wouldn't want testing locally to modify our baseline test number, it would just make it clear in the help output.

```suggestion
    Format, combine and upload the test coverage files. Note that this command only works in CI.
```

.github/workflows/backend_checks.yml
```diff
@@ -282,4 +317,34 @@ jobs:
...
+          name: ops-saas
+          path: coverage/ops-saas.lcov
+
+  Coverage-Upload:
+    needs: Safe-Tests
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
```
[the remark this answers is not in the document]
ThomasLaPiana: good catch!

.github/workflows/backend_checks.yml
ThomasLaPiana: I moved things around a bit, this has been updated

ThomasLaPiana:
the job is passing and everything is getting uploaded. However it seems that when running in CI it is not properly getting the branch name and commit sha, so I'm working on debugging that

ThomasLaPiana:
confirmed once again that it is working locally but not in CI....

The branch now shows correctly but is showing up as an "invalid state" in the Code Climate UI

ThomasLaPiana:
Going to try something around this solution: https://github.com/codeclimate/test-reporter/issues/375

ThomasLaPiana:
I got it to break locally as well to match the CI breaks, and it definitely is related to the path. I believe what is happening is that because we are installing `fides` as a proper Python package when we run tests, the prefix path that `Code Coverage` is using is not valid, since it uses the current working directory as the prefix, when it should actually be `/usr/local/lib/python3.10/site-packages` or something similar

ThomasLaPiana:
after adding some custom logic in the last commit, I got a valid report uploaded, hopefully it works in CI as well

ThomasLaPiana:
@sanders41 any idea on this error? It's a CI-specific one, and seems arbitrary

sanders41:
The 3 most common reasons I have seen for this are:

1. It really is a permission issues. This one seems unlikely since you could write to it before. Could be ruled out by running with `sudo`
2. Trying to open a directory instead of a file. It sure looks like a file to me, but maybe GitHub actions is doing something weird here and making a directory. Usually this would be debugged by checking that the path is a file before trying to open, but this should be the first creation of the file. May be worth trying though just to see what we get. 
3. Something else already has the file open.

ThomasLaPiana:
The test reports are now working! https://codeclimate.com/repos/63be6d666de62000f3e44a88/settings/test_reports/63c7791b56b68a1d0b82b1a4

ThomasLaPiana:
The (for real this time though) final bug I've discovered is that the branch name in GitHub Actions is given as `2198/merge` instead of `ThomasLaPiana-send-coverage...`

Due to this mismatch, the auto-updating code comment is not seeing what is actually happening in this PR and was instead only reflecting results from my local runs (where everything is working as intended)

I will update the CI checks with hopefully the proper github action var and should be set

ThomasLaPiana:
using the head ref appeared to fix the previous issue

image

ThomasLaPiana:
> using the head ref appeared to fix the previous issue
> 
> image

Even though it matches in the UI, it doesn't appear to be working because the comment still isn't updating...

noxfiles/ci_nox.py
```diff
@@ -162,182 +166,156 @@ def minimal_config_startup(session: nox.Session) -> None:
...
-        session.run(*run_command, external=True)
+def sum_coverage(session: nox.Session) -> None:
+    """
+    Combine multiple Code Climate coverage files.
+
+    7 files are expected for summation, one from each of the test groups run in CI.
+    """
+    download_cc_reporter = f"curl -L {TEST_REPORTER_URL} > ./cc-test-reporter && chmod +x ./cc-test-reporter && "
+    sum_test_coverage = (
+        "./cc-test-reporter sum-coverage --parts 4 coverage/**/codeclimate.json ;"
```
sanders41, line 283: The comment says 7 files, but `--parts` is set to 4. Should this be 7?

ThomasLaPiana: the comment is wrong, I updated it to _not_ expect the external tests and forgot to update the comment

noxfiles/ci_nox.py
```diff
@@ -162,182 +166,156 @@ def minimal_config_startup(session: nox.Session) -> None:
...
-            OPS_TEST_DIR,
-            "-m",
-            "integration_saas",
-        )
-        session.run(*run_command, external=True)
+def sum_coverage(session: nox.Session) -> None:
+    """
+    Combine multiple Code Climate coverage files.
+
+    7 files are expected for summation, one from each of the test groups run in CI.
```
ThomasLaPiana, line 279:
```suggestion
    4 files are expected for summation, one from each of the safe test groups run in CI.
```

noxfiles/ci_nox.py
ThomasLaPiana: fixed! good catch

sanders41 (approved):

PR #2205, opened 2023-01-11
Add custom metadata tables

sanders41:
Closes #<issue>

### Code Changes

* Added new tables for custom metadata

### Steps to Confirm

* [ ] Run `nox -s dev` and verify the new tables are created

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes


src/fides/api/ctl/sql_models.py
```diff
@@ -313,3 +315,64 @@ class SystemScans(Base):
...
+    field_definition = Column(String, index=True)
+    custom_field = relationship(
+        "CustomField",
+        back_populates="plus_custom_field_definition",
+        cascade="delete, delete-orphan",
+    )
+    UniqueConstraint("name", "resource_type")
+
+
+class CustomField(Base):
```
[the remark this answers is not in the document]
adamsachs: my inclination tells me to stick with the "soft" foreign key reference and enforce integrity in the python code, in order to (as you mention) keep things flexible going forward. it would allow us to apply custom fields to more resource types going forward, without needing to create new tables and deal with database migrations.

that being said, there's a risk i'm trying to over-genericize the use case here - maybe we can get some input from product (cc @mfbrown) on how likely it'd be that we'd want to eventually apply custom fields to other resource types beyond what we've scoped out here initially? to me it feels like a good future-proofing, but i'm not sure if i'm just going overboard with this.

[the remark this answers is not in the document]
mfbrown: I agree that there might be cases for custom fields on other types and if we can plan for that, we should. Right off the bat, I can plausibly see a world where someone needs a custom field on organizations, for example, which we keep pretty light today, but as we expand to support multiple orgs, could be useful.

Just one example, but I do see the application here growing over time.

src/fides/api/ctl/sql_models.py
```diff
@@ -313,3 +315,65 @@ class SystemScans(Base):
...
+        nullable=False,
+    )
+    allow_list = Column(
+        ARRAY(String)
+    )  # Ulitmately this will come the the CustomList table
+    resource_type = Column(EnumColumn(ResourceTypes), nullable=False)
+    field_definition = Column(String, index=True)
+    custom_field = relationship(
+        "CustomField",
+        back_populates="plus_custom_field_definition",
```
adamsachs, line 355: i think this should be the column/field name, rather than the table name, right?

adamsachs: made a potential fix in https://github.com/ethyca/fides/pull/2212/files

sanders41: You are right, the change you did in #2212 looks correct so we can use the update there when we merge back here.

sanders41:
@adamsachs these failures didn't make sense to me so I checked out main and the same ones failed for me there. Does the same happen for you? I'm trying to figure out if it is something specific to this PR or not.

.fides/db_dataset.yml
```diff
@@ -1514,4 +1611,4 @@ dataset:
       description: 'The name of the organization this Fides deployment belongs to'
       data_categories:
       - user.workplace
-      data_qualifier: aggregated.anonymized.unlinked_pseudonymized.pseudonymized.identified
```
adamsachs, line 1517: nit: probably want to keep the newline at EOF here?

src/fides/api/ctl/sql_models.py
```diff
@@ -313,3 +315,82 @@ class SystemScans(Base):
...
+    """Custom metadata for resources."""
+
+    __tablename__ = "plus_custom_field"
+
+    resource_type = Column(EnumColumn(ResourceTypes), nullable=False)
+    resource_id = Column(String, index=True, nullable=False)
+    custom_field_definition_id = Column(
+        String, ForeignKey(CustomFieldDefinition.id), nullable=False
+    )
+    value = Column(String)
```
adamsachs, line 389: just thought about this - how are we supporting a string array for a given custom field instance? i feel like we'll either need another column with type `ARRAY(String)` or we'd need to adjust the `UniqueConstraint` to allow more than one `CustomField` with the same `custom_field_definition_id` on the same `resource_id`.

happy to discuss more offline, this is a bit complex.

sanders41: Discussed in slack. We are going to change `value` to be an `ARRAY(String)`

.fides/db_dataset.yml
adamsachs: oh whoops! got it reversed 🤦

adamsachs (approved):
nice, thanks for making the quick adjustments - this looks good for what we need so far! we may end up needing to come back to make adjustments but let's get this cut merged into a release and go from there

PR #2207, opened 2023-01-11
update firebase auth connector to support phone number identities

### LINKED ISSUE #2206: update firebase connector to support phone number identities

for a client use case, we need our firebase integration (i.e. the saas connector) to support DSR processing with phone number identities

- the firebase API supports looking up users by phone number, so we should be able to leverage this functionality
- there are some considerations in terms of using custom overrides with multiple identities, as we will need to here. i think we should have what we need already, but it's a detail we should key in on

not any come to mind

### Additional context

- a client priority item
- the `update` path may have a few nuances to consider, the `read` and `delete` should be pretty straightforward
- we also may switch to using `delete` as the default erasure path, based on client feedback

adamsachs:
Closes #2206 

### Code Changes

* update firebase auth config yaml:
    *  two distinct `read` requests, one for `email` identities and another for `phone_number`
    *  allow for `update` request to receive either an `email` or `phone_number` identity
    * bump the version so as to trigger an update to the template on a running server
* update firebase auth overrides to support `email` or `phone_number` identities for `read`, `update` and `delete` functionality
   * in line with the general saas framework, functionality is still either/or, using multiple identities is not currently supported (though it could be in the future)
* test coverage of above functionality
 

### Steps to Confirm

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes


src/fides/api/ops/service/connectors/saas_query_config.py
```diff
@@ -85,7 +86,7 @@ def _requests_using_identity(
             if any(
                 param_value
                 for param_value in request.param_values or []
-                if param_value.identity == self._get_identity()
+                if param_value.identity == get_identity(self.privacy_request)
```
galvana, line 89: Nice re-use here!

src/fides/api/ops/service/saas_request/override_implementations/firebase_auth_request_overrides.py
```diff
@@ -126,13 +146,31 @@ def firebase_auth_user_delete(
...
         auth.delete_user(user.uid, app=app)
         rows_updated += 1
     return rows_updated
 
 
+def _retrieve_user_record(
+    privacy_request: PrivacyRequest, row_param_values: Dict[str, Any], app: App
+) -> UserRecord:
+    """
+    Utility that erasure functions can use to retrieve a Firebase `UserRecord
```
galvana, line 159:
```suggestion
    Utility that erasure functions can use to retrieve a Firebase `UserRecord`
```

tests/ops/fixtures/saas/firebase_auth_fixtures.py
```diff
@@ -114,6 +115,7 @@ def firebase_auth_user(firebase_auth_secrets) -> Generator:
         email_verified=False,
         display_name="John Doe",
         photo_url="http://www.example.com/12345678/photo.png",
+        phone_number="+1" + str(randint(1000000000, 9999999999)),
```
galvana, line 118: I'm guessing there's no validation on the phone number other than it has to be unique within the system?

galvana (commented):
This change makes sense to me! The refactor of the `get_identity` function is nice and I'm glad to see we can continue to expand on these custom functions without having to change the framework.

Let's settle on the expected behavior for multiple identities soon!

tests/ops/fixtures/saas/firebase_auth_fixtures.py
adamsachs: not that has come up yet in testing, at least :)

i'm a little bit unsure even on the uniqueness constraints, i.e. if the user has both an `email` and a `phone_number`, i'm not even sure if `phone_number` must be unique. but i'm guessing it probably is?

Kelsey-Ethyca (approved):

PR #2222, opened 2023-01-12
Update seed script to not overwrite existing default policy resources

### LINKED ISSUE #2191: `default_access_policy_rule`'s `storage_destination` is reverted on server reboot
### Bug Description

in order to update where access results are stored, our implementation guide encourages users to create a new storage config and  then `PATCH` the `default_access_policy_rule` to point its `storage_destination` property at the newly created storage config. it seems as if this change to the `default_access_policy_rule` is reverted on each server reboot, as reported by @RobertKeyser based on a customer use case. 

upon a quick look at the code, it seems as if our `load_default_resources` method in `seed.py` is generally executing upserts to the default resources, i.e. with `create_or_update`method calls. that would likely explain the behavior. one potential path to resolve this may be to update `seed.py` to instead check if a given default resource already exists, and only if it doesn't should we `create` the default resource with seed values.

#### Steps to Reproduce

i haven't repro'd yet locally, taking @RobertKeyser word about what happened on a client env. to try and repro we could:

1. `nox -s dev -- ui`
2. add a storage config and update the `default_access_policy_rule` to point to this storage config using our API
3. cancel the env/teardown. do not clean/remove the db
4. run `nox -s dev --ui` again, and check whether `default_access_policy_rule` has persisted the updates or not. if we're reproducing the bug, it should have reverted back to its original state, pointing instead to the default storage config. 

### Expected behavior

- changes to `default_access_policy_rule` should be persisted across app restarts
- more broadly, i'd expect changes to _all_ default resources to be persisted across app restarts. i think we should look to accomplish this behavior.

### Additional context

found in a client env, and they were relying on our implementation guides to make these updates. so this should be considered a standard use case, not an edge case.

cc @RobertKeyser @NevilleS

adamsachs:
Prevents any manual changes that have been made to the default policy resources from being overwritten on a server restart.

Closes #2191 

### Code Changes

* update `seed.py` to check for existence of the default policy resources before creating them, rather than upserting the resources
* include test coverage of basic default policy resource creation as well as non-overwrite behavior

### Steps to Confirm

1. nox -s dev -- ui
2. add a storage config and update the default_access_policy_rule to point to this storage config using our API
3. cancel the env/teardown. do not clean/remove the db
4. run nox -s dev --ui again, and check whether default_access_policy_rule has persisted the updates or not

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes


src/fides/api/ctl/database/seed.py
```diff
@@ -278,6 +296,46 @@ async def load_default_dsr_policies() -> None:
             except KeyOrNameAlreadyExists:
                 # This rule target already exists against the Policy
                 pass
+    else:
+        log.info(
+            f"Skipping {DEFAULT_ERASURE_POLICY_RULE} creation as it already exists in the database"
+        )
+
+
+async def load_default_dsr_policies() -> None:
```
sanders41, line 305: There is no `await` so this should be a `def` instead of `async def`. Or should it be using the `async_session`?

adamsachs: not sure on the async session, it had never been using it before so i'd prefer not to change that in this PR.

i'll update to remove the `async`, i was also confused why it was `async` but just decided to leave it alone. just to be clear, i didn't actually change the function signature (just changed the signature's spot in the file so it comes up here in the diff), but i'll make this update (assuming it passes tests) because i agree with you that it shouldn't be `async`

sanders41 (approved):

NevilleS (approved):
Overall this makes sense to me, thanks. I didn't review the code in detail though to be clear!

Just so I understand, I feel like the only major caveat is that this'll prevent us from making changes to the default policies going forward - if it's only creating and not updating, that also means if we decide to modify something (e.g. add/remove a default data category), then that won't update existing installs when they upgrade.

Is that right? I'm personally OK with that - these defaults are reasonable, and the pros outweigh the cons here, but just wanted to confirm my understanding

adamsachs:
yeah that's correct @NevilleS. i think in order to accommodate that, we'll need some versioning scheme. could probably come up with something pretty lightweight, but let's do that outside of this PR.

like you said, i think the pros outweigh the cons here. for the scenarios where we need default policies to be updated, i think we'd either need a standard migration script or to come up with something bespoke for the particular use case. i feel like that sort of situation would require some care and review, regardless, with how things stand now.

anyway, it's a good limitation to keep in mind, but given we feel that this improves things overall, i'll go ahead and merge.

Issue #2092, opened 2022-12-20
Deprecate patch_datasets and patch_yaml_datasets endpoints


pattisdr:
❗ Depends on https://github.com/ethyca/fides/issues/1766 ❗ 

### Is your feature request related to a specific problem?

Currently datasets are stored in two places for different use cases, on the DatasetConfig.dataset table and in the ctl_datasets table.  Our unified fides resources work is moving to storing this in `ctl_datasets`.

A new endpoint, `patch_dataset_configs` is being added.  After the UI is in to use this endpoint, the other endpoints can be deprecated.

### Describe the solution you'd like

- Deprecate `PATCH {{host}}/connection/{{connection_key}}/dataset`. 
- Deprecate `PATCH {{host}}/yml/connection/{{connection_key}}/dataset`
- Deprecate `GET {{host}}//connection/{{connection_key}}/dataset`
- Deprecate `GET {{host}}//connection/{{connection_key}}/dataset/{{datasetconfig_key}}`
    - Have endpoint return a message that the endpoint is deprecated and direct them to their replacement (`/datasetconfig` endpoints)
   

### Describe alternatives you've considered, if any

_A description of any alternative solutions or features you've considered._

### Additional context

_Add any other context or screenshots about the feature request here._

pattisdr:
Also switch any other remaining instances in the codebase to use the new patterns (upsert to CTL Dataset then PATCH to Dataset Config).

`nox -s test_env` for example, calls the PATCH /connection/connection_key/dataset endpoint we'd like to deprecate

pattisdr:
This doesn't block the release of `unified-fides-resources`.

pattisdr:
Also need to discuss what our deprecation policy is.  

Motivations to deprecate: These endpoints on the `unified-fides-resources` branch now potentially have surprising updates, in that they not only update the datasetconfig but the associated ctl dataset.  

While we can change all of our code to not make calls to these endpoints, is anyone relying on these?

PR #2225, opened 2023-01-13
[Unified Fides Resources] Update Test Env Setup and Quickstart

pattisdr:
Closes N/A

### Code Changes

* [ ] Adjust test env setup and quickstart to call a different set of endpoints when creating datasetconfigs (upsert the ctl dataset first and then link to the datasetconfig) instead of the old endpoint (patch a datasetconfig)

### Steps to Confirm

* [ ] Run `nox -s test_env`
* [ ] Verify everything came up as expected
* [ ] `nox -s clean`
* [ ] Run `nox -s quickstart` - run quickstart through to the end

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

Get tests passing on `unified-fides-resources` branch after update with main.

Also: Remove remaining uses of ops dataset endpoints in fides that we'd like to eventually deprecate (not a blocker for `unified-fides-resources`): deprecation will happen as a follow-up in the future: https://github.com/ethyca/fides/issues/2092.  These endpoints still work, but are riskier because a dataset is upserted in the background, and we don't want people to unintentionally override their datasets.

-  Instead of a patch to /connection/connection_key/dataset do a post to /dataset/upsert and then a patch to /connection/connection_key/datasetconfig in both the test env setup and the quickstart

### DIFF
--- a/.fides/fides.toml
+++ b/.fides/fides.toml
@@ -7,7 +7,7 @@ db = "fides"
 test_db = "fides_test"
 
 [credentials]
-app_postgres = {connection_string="postgresql+psycopg2://postgres:fides@fides-db:5432/fides_test"}
+app_postgres = {connection_string="postgresql+psycopg2://postgres:fides@fides-db:5432/fides"}
 
 [logging]
 level = "INFO"
--- a/data/saas/dataset/recharge_dataset.yml
+++ b/data/saas/dataset/recharge_dataset.yml
@@ -91,7 +91,6 @@ dataset:
             fidesops_meta:
               data_type: string
           - name: shopify_customer_id
-            data_categories: [system.operations]
             fidesops_meta:
               data_type: object
           - name: status
--- a/scripts/quickstart.py
+++ b/scripts/quickstart.py
@@ -236,7 +236,7 @@ def validate_dataset(connection_key: str, yaml_path: str):
 
 def create_dataset(connection_key: str, yaml_path: str):
     """
-    Create a dataset in fidesops given a YAML manifest file.
+    Create a dataset and a datasetconfig in fides given a YAML manifest file.
     Requires the `connection_key` for the PostgreSQL connection, and `yaml_path`
     that is a local filepath to a .yml dataset Fides manifest file.
     Returns the response JSON if successful, or throws an error otherwise.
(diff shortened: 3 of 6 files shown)


pattisdr:
@ThomasLaPiana wanted to see if you had any ideas here.  The `unified-fides-resources` feature branch has been consistently failing the `fides_db_scan` checks with:

```
sqlalchemy.exc.OperationalError: (psycopg2.OperationalError) connection to server at "fides-db" (172.18.0.3), port 5432 failed: FATAL:  database "fides_test" does not exist`.  
```

Both Allison and I have run into this on respective PR's against `unified-fides-resources`.  I've been updating the branch regularly with main, and recently updated it today.

.fides/fides.toml
```diff
@@ -7,7 +7,7 @@ db = "fides"
 test_db = "fides_test"
 
 [credentials]
-app_postgres = {connection_string="postgresql+psycopg2://postgres:fides@fides-db:5432/fides_test"}
+app_postgres = {connection_string="postgresql+psycopg2://postgres:fides@fides-db:5432/fides"}
```
pattisdr, line 10: I'm not sure why I made this change on this branch, that was a mistake - this should be looking at the running application db not the test db, and it was why we've had a failing test on this branch for awhile.

scripts/quickstart.py
```diff
@@ -324,7 +344,7 @@ def create_dsr_policy(key: str):
     if response.ok:
         policies = (response.json())["succeeded"]
         if len(policies) > 0:
-            logger.info("Created fidesops policy with key=%s via {url}", key)
+            logger.info("Created fidesops policy with key={} via {}", key, url)
```
pattisdr, line 347: This had been causing the quickstart to break FYI, not sure if people are using the quickstart

ThomasLaPiana: I believe @seanpreston said it was still being used for manual testing. If that's the case I think we should also be running it during CI checks (another ticket for another time)

pattisdr:
@allisonking the main reason I put up this PR was to get tests passing on `unified-fides-resources` after I got it up to date with main.  It fixes that failing fides db scan issue we were both running into, and gets some of the newly added saas connectors adjusted so they work with unified-fides-resources.  The remaining failing tests are unrelated to this work.

data/saas/dataset/recharge_dataset.yml
```diff
@@ -91,7 +91,6 @@ dataset:
             fidesops_meta:
               data_type: string
           - name: shopify_customer_id
-            data_categories: [system.operations]
```
allisonking, line 94: how come this had to be deleted?

allisonking (approved):
🚀

pattisdr:
Thanks for the review @allisonking

PR #2226, opened 2023-01-13
Classify Dataset UX Updates

SteveDMurphy:
Closes [fidesplus#404](https://github.com/ethyca/fidesplus/issues/404)

### Code Changes

* [x] Return only top 1 suggestion instead of `top n` in data table
* [x] rename button to state `approve classifications`
* [x] Update dataset classify heading text
* [x] Update to only approve collection at one time vs entire dataset
* [x] Update from returning the first classification to ensure returning the top 1 in the same way the dropdown does
* [x] Create a common function to get the `topClassification`
* [x] Exclude `Identifiability` from  rendering as part of the `Dataset` view (Closes [fidesplus#513](https://github.com/ethyca/fidesplus/issues/513))

### Steps to Confirm

* [ ] Start up `fidesplus` in the background
* [ ] Navigate to the admin-ui and start up the frontend using `npm run dev`
* [ ] Classify the fides db using `postgresql+psycopg2://postgres:fides@fidesplus-db:5432/fides` (it's actually a bit collection-heavy so feel free to use a [smaller one you can find in 1Password such as this](https://start.1password.com/open/i?a=5LPILPRDUBGWBPJWYS6QHHHLRE&v=p3bpsusghc3t45pzpsebehrd4m&i=ck4j4jubqxyyajj6q4agmrxmly&h=ethyca.1password.com))
* [ ] Click into the dataset and run through each collection. On the last collection, validate the status is updated

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### Description Of Changes

Currently, these changes feel like a lot is left to be desired as far as the UX for classification. The approve classification button is not informative and should likely change on click or load the next collection to be completed.


SteveDMurphy:
As an early criticism, check out this Loom -> https://www.loom.com/share/963dbf44bd044668aa57da96875badb6

It feels like the UX doesn't seem to quite align with what we want here, some more work around the button to approve classifications is likely required. I'd love to see something around loading the next collection to be reviewed as well 🤷🏽

clients/admin-ui/cypress/e2e/datasets-classify.cy.ts
```diff
@@ -88,12 +88,6 @@ describe("Datasets with Fides Classify", () => {
       cy.wait("@getDatasets");
 
       cy.url().should("match", /dataset$/);
-
-      // The combination of Next routing and a toast message makes Cypress get weird
-      // when re-running this test case. Introducing a delay fixes it.
-      // eslint-disable-next-line cypress/no-unnecessary-waiting
-      cy.wait(100);
-      cy.getByTestId("toast-success-msg");
```
allisonking, line 96: hmm what happened to the toast?

clients/admin-ui/src/features/dataset/DatasetFieldCell.tsx
```diff
@@ -29,20 +29,31 @@ const DatasetFieldCell = ({
...
+    const topClassification = classifyField?.classifications.reduce(
+      (maxClassification, next) => {
+        if (
+          (maxClassification.aggregated_score ?? 0) <
+          (next.aggregated_score ?? 0)
+        ) {
+          return next;
+        }
+        return maxClassification;
+      }
```
SteveDMurphy, line 41: Previously was just grabbing the first (top) classification but a failed Cypress test pointed me in the direction of copying the same functionality here to get the top 1 as used in https://github.com/ethyca/fides/blob/5a485387d8af247ec6479e4115088cbbb8394d77/clients/admin-ui/src/features/dataset/helpers.ts#L102 (there may be a better way for me to reuse this)

clients/admin-ui/src/features/dataset/ApproveClassification.tsx
```diff
@@ -43,25 +43,40 @@ const ApproveClassification = () => {
...
+          ) {
+            uncategorizedCount += 1;
+          }
+        });
       });
-      if (isErrorResult(statusResult)) {
-        toast(errorToastParams(getErrorMessage(statusResult.error)));
-        return;
-      }
+      if (uncategorizedCount === 0) {
```
allisonking, line 67: ah okay it took me a while to figure out what was happening here. might be helpful to add a comment along the lines of `Only update the dataset as classified when all fields have been categorized` ?

clients/admin-ui/src/features/dataset/ApproveClassification.tsx
```diff
@@ -43,25 +43,40 @@ const ApproveClassification = () => {
...
       const updateResult = await updateDataset(updatedDataset);
       if (isErrorResult(updateResult)) {
         toast(errorToastParams(getErrorMessage(updateResult.error)));
         return;
       }
-
-      const statusResult = await updateClassifyInstance({
-        dataset_fides_key: dataset.fides_key,
-        status: ClassificationStatus.REVIEWED,
+      toast(successToastParams("Collection classified and approved"));
```
allisonking, line 55: hm I wonder why this isn't showing up in the video. I tried it locally and it's not showing up there either. I'll look into it a bit more...

allisonking: hm it looks like none of our toasts are working anymore, even on `main` 😱

clients/admin-ui/src/features/dataset/DatasetFieldCell.tsx
allisonking: hmm yeah, we may want to pull that `.reduce` out into its own func, also in `dataset/helpers.ts` and use it in both places?

allisonking (commented):
I think the UX is currently weird because toasts seem not to be working even on main. will investigate further, but that's not your PR's fault!

clients/admin-ui/src/features/dataset/ApproveClassification.tsx
TheAndrewJackson: I wonder if it's related to this. I think I ran into this before in the privacy center. Doing what this comment describes fixed my problem. 

[https://github.com/chakra-ui/chakra-ui/issues/5839#issuecomment-1278883908](https://github.com/chakra-ui/chakra-ui/issues/5839#issuecomment-1278883908)

allisonking: heh yeah I remember you fixing this before! I wonder if it came up with the dependabot bump of chakra, or something else...

@TheAndrewJackson want to open a quick PR off of main so we can fix since you've already found the solution? and if there's a release going on, we probably want this in before that...

SteveDMurphy: Oof, thank you both! Late Friday night I was losing my mind trying to get the toast messages to work anywhere and resorted to this 🥴  - thanks again for putting some brain power into it

clients/admin-ui/src/features/dataset/DatasetFieldCell.tsx
SteveDMurphy: fixed! Thank you for the input 🙌🏽

clients/admin-ui/src/features/dataset/ApproveClassification.tsx
SteveDMurphy: comments added 👍🏽

clients/admin-ui/src/features/dataset/ApproveClassification.tsx
TheAndrewJackson: @allisonking Yes, I'll take a look at that today. If that's the issue it should be a quick fix 🤞

clients/admin-ui/cypress/e2e/datasets-classify.cy.ts
allisonking: Since it looks like the toasts are broken [only in the dev environment](https://github.com/ethyca/fides/issues/2268#issuecomment-1387449219), let's restore these tests and make sure they pass in CI (or instead of `npm run dev`, use `npm run cy:start` and then `npm run cy:run` to make sure things pass that way)

clients/admin-ui/src/features/dataset/ApproveClassification.tsx
```diff
@@ -43,25 +43,42 @@ const ApproveClassification = () => {
...
+      // Validate if any fileds still require category approval
```
allisonking, line 56:
```suggestion
      // Validate if any fields still require category approval
```

clients/admin-ui/src/features/dataset/DatasetCollectionView.tsx
```diff
@@ -67,8 +67,9 @@ const DatasetCollectionView = ({ fidesKey }: Props) => {
   const activeCollection = useSelector(selectActiveCollection);
   const activeEditor = useSelector(selectActiveEditor);
 
-  const [columns, setColumns] =
-    useState<ColumnMetadata<DatasetField>[]>(ALL_COLUMNS);
+  const [columns, setColumns] = useState<ColumnMetadata<DatasetField>[]>(
+    ALL_COLUMNS.filter((column) => column.name !== "Identifiability")
```
allisonking, line 71: perhaps we can just remove `Identifiability` from `ALL_COLUMNS`? I don't think anything else is using it, and we wouldn't want it in the column dropdown either

allisonking (changes requested):
looks great!! just requesting changes to the tests, and also a column thing

clients/admin-ui/src/features/dataset/DatasetCollectionView.tsx
SteveDMurphy: Yep! That's a much easier solution. I wasn't sure if it was helpful to keep around just in case but didn't catch the filter so seems to be cleaner that way 👍🏽

SteveDMurphy:
Not sure but I may have introduced a new issue somewhere causing an empty state seizure 🤔 found when retesting https://www.loom.com/share/5b57f5b506cb445b9fa0766b2e640409

allisonking:
> Not sure but I may have introduced a new issue somewhere causing an empty state seizure 🤔 found when retesting 

Hmm what steps did you take to get here? I've seen this once before but I think it's because my backend wasn't working properly. I'll try to repro on main

clients/admin-ui/src/features/dataset/DatasetCollectionView.tsx
SteveDMurphy: Fixed in [0822761](https://github.com/ethyca/fides/pull/2226/commits/082276118aef34f1954aa644505531027d63bcec)!

SteveDMurphy:
> > Not sure but I may have introduced a new issue somewhere causing an empty state seizure 🤔 found when retesting
> 
> Hmm what steps did you take to get here? I've seen this once before but I think it's because my backend wasn't working properly. I'll try to repro on main

It looks like it only happens when using `npm run cy:start` - I noticed the `classify dataset` button was missing as well so thinking maybe something wrong with backend too. Flipped back to `npm run dev` since Cypress seemed to be fine and no more stuttering / `classify dataset` button visible again

clients/admin-ui/cypress/e2e/datasets-classify.cy.ts
```diff
@@ -211,15 +210,8 @@ describe("Datasets with Fides Classify", () => {
...
-      // The mutations should run.
+      // The mutation should run.
       cy.wait("@putDataset");
-      cy.wait("@putClassify");
-
-      // The instance should be re-fetched.
-      cy.wait("@getClassify");
-
-      // The updated status should make the button disappear.
-      cy.getByTestId("approve-classification-btn").should("not.exist");
```
allisonking, line 222: ah, perhaps we still want the button to disappear when the collection has already been approved? I guess there isn't an easy way to tell that it's been approved vs just that all of its fields have been assigned data categories

allisonking (approved):
yay nice job Steve!! I left one suggestion, though it might make sense to make a separate ticket for it, if we want to do it at all

TheAndrewJackson (approved):
Great job! I like your use of `reduce` :shipit:

SteveDMurphy:
> > > > Not sure but I may have introduced a new issue somewhere causing an empty state seizure 🤔 found when retesting
> > > 
> > > 
> > > Hmm what steps did you take to get here? I've seen this once before but I think it's because my backend wasn't working properly. I'll try to repro on main
> > 
> > 
> > It looks like it only happens when using `npm run cy:start` - I noticed the `classify dataset` button was missing as well so thinking maybe something wrong with backend too. Flipped back to `npm run dev` since Cypress seemed to be fine and no more stuttering / `classify dataset` button visible again
> 
> Ohhhh okay yeah `npm run cy:start` should only be used against cypress (which I'm only just realizing now too!). This is because we set a `NODE_ENV=test` there, and in our [next config we don't reroute backend requests](https://github.com/ethyca/fides/blob/SteveDMurphy-404-classify-ui-updates/clients/admin-ui/next.config.js/#L30-L34) if we are in the test environment (since these cypress tests should work without a backend (they're all stubbed))
> 
> I believe to run a built js, you can `npm run build` and then `npm run start`

Ahh, interesting! Thank you once again for the detailed explainer 🙌🏽 makes sense now and I'm not afraid that I did something wonky somewhere 😂

clients/admin-ui/cypress/e2e/datasets-classify.cy.ts
SteveDMurphy: So glad you saw that! I actually brought it up as well [last week in standup](https://ethyca.slack.com/archives/C04484SLXK7/p1673626308868609?thread_ts=1673625498.778729&cid=C04484SLXK7) and @rsilvery opened up a UX Fix ticket that is linked in there. I think there may have been some design help required 🤔

SteveDMurphy: Personally I'd love to have the button be replaced to move to the next collection (if more exist)

clients/admin-ui/src/features/dataset/ApproveClassification.tsx
SteveDMurphy: Thank you both for taking this on along with everything else!

Issue #1293, opened 2022-10-09
Add a new Cypress E2E test runner that executes against a fully configured environment


NevilleS:
### Description

In some previous releases, we worked on building out test setup scripts (#1291) and a local manual test environment (#1292), which gives us the ability to do true E2E manual tests to confirm all features are working as expected when running together.

Our existing Cypress suites for the `admin-ui` and `privacy-center` are great, but they are isolated to their specific applications and tend to stub out their backends for performance. Therefore, there's still an open need for an E2E test runner to use for regression tests on an ongoing basis.

Once the runner is configured, let's put it to work by defining a core smoke test for releases. We should ensure this can run against our local test env first, but I want to immediately start using this against some hosted staging & demo environments in the short-term, so let's ensure that any URLs are easy to configure and override (e.g. http://localhost:3000 will quickly become something like https://fides.fides-staging.ethyca.com/)

In addition, let's signup for Cypress Cloud and publish results there, it's a great product!

### Acceptance Criteria

* MUST implement a new Cypress runner to run E2E tests, targeting the local "manual test" environment (see #1292)
* MUST support targeting different hosts via environment variables, so we can run against staging / demo / etc.
* MUST have a `nox` session (e.g. `nox -s e2e_test`) that runs the Cypress suite in headless mode
* MUST configure the test reporter to publish results to our Cypress Cloud account
* MUST write a single Cypress E2E test that runs a simple smoke test covering:
  1. Confirm can login to the Admin UI
  2. Confirm that the Postgres & Mongo connectors are configured
  3. Confirm that the Privacy Center can be accessed
  4. Submit an access request via the Privacy Center
  5. Approve the access request via the Admin UI
  6. Confirm the access request succeeds

ssangervasi:
One note from this convo:
> not running against the live backend in those E2E tests is a bug, not a feature. we’ve scoped out work to fix that here

I am a fan of FE+BE integrated regression test. However, I would say the 80/20 rule for cypress tests is that 80% should be using stubbed endpoints. Developers should be live-running the Cypress tests for a feature when changing it. An actual BE is usually bad for this because:
- It will be slow
- Create/Deletes are going to get out of sync
- The data that is useful for regressions is a separate concern from what is useful for covering a UI feature
- The BE might not even be updated yet (developing features in tandem like Classify UI).

[...]

Issue #928, opened 2022-07-19
Add a command that allows starting up fidesctl in prod mode


allisonking:
### Is your feature request related to a specific problem?

Right now it's difficult to actually see the built UI since our docker-compose overwrites the built static files with the user's local environment.

### Describe the solution you'd like

We should add a command that mirrors the real deployment as much as possible.

### Describe alternatives you've considered, if any

_A description of any alternative solutions or features you've considered._

### Additional context

https://github.com/ethyca/fides/pull/757#discussion_r901296645

This would also help us out with https://github.com/ethyca/fidesctl-plus/issues/30

I suspect @ThomasLaPiana has some good ideas on how to do this :)

ThomasLaPiana:
so you're saying add a `nox -s prod` in addition to a `nox -s dev`? Makes sense to me! I'm sure I could cook something up :)

allisonking:
Closing this one in favor of https://github.com/ethyca/fides/issues/2385

PR #2241, opened 2023-01-13
Cypress end to end smoke test

allisonking:
A good chunk of https://github.com/ethyca/fides/issues/1293

### Code Changes

* [x] New folder `cypress-e2e` within `clients` with Cypress 12 set up
* [x] Test for logging in
* [x] Test for access request
* [x] Test for confirming mongo and postgres connectors are configured

### Steps to Confirm

* [ ] `nox -s test_env` 
* [ ] In the new folder `cypress-e2e`, install dependencies via `npm i`
* [ ] `npm run cy:run` to run all the cypress tests
* [ ] Alternatively, use `npm run cy:open` to watch the tests run in their UI 

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

This is the first chunk of https://github.com/ethyca/fides/issues/1293. The new shiny thing is `cy.origin` which was introduced in Cypress 12 and makes it possible for us to write tests that go between different origins (i.e. privacy center and admin ui) without cross origin errors. 

The original requirement wanted one single test to test all the things, though I thought it made sense to break them up and wasn't sure if there was a benefit to them all being in one test. (mostly checking mongo + postgres seemed independent of approving an access request)

I'd like to get this reviewed and discussed first, and then I can break out the remaining work in https://github.com/ethyca/fides/issues/1293 to separate tickets.

One thing I'd like to discuss is if it is okay that we are running these tests against the development UI servers (`nox -s test_env` spins up `localhost:3000` and `localhost:3001` for the web apps). I think ideally we'd test against `localhost:8080` (the built frontend for admin-ui at least) to more mimic the production environment. But then we'd likely have to change some configurations with `test_env`, perhaps to not volume mount our local built front end?

Also, do we want this running in CI?


NevilleS:
Thanks for doing this - on first blush this is precisely what we need, so what I want to do now is figure out how to structure this best for maintainability.

I'll give it some more time this week. I tried running locally and the tests failed because of timeouts... because it was having to build the admin UI pages first before it could navigate to them 🙃. A production build of the Admin UI feels more appropriate for Cypress E2E...

clients/cypress-e2e/cypress/support/constants.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Based off of fides.toml
```
ssangervasi, line 2: Looks like Cypress some [other ways](https://docs.cypress.io/guides/guides/environment-variables#Option-3-CYPRESS_) of sharing environment info with a test runner. I don't know if environment variables would actually be convenient, but it seems likely we'd outgrow copying into this file quickly.

clients/cypress-e2e/cypress/e2e/smoke_test.cy.ts
```diff
@@ -0,0 +1,71 @@
...
+    cy.getByTestId("privacy-request-form").within(() => {
+      cy.get("input#name").type("Jenny");
+      cy.get("input#email").type("jenny@example.com");
+
+      cy.get("input#phone").type("555 867 5309");
+      cy.get("button").contains("Continue").click();
+    });
+
+    // Approve the request in the admin UI
+    cy.visit("localhost:3000");
```
ssangervasi, line 23: Probably would be good to extract the base URLs to constant/env setting.

clients/cypress-e2e/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Running
+
+First, start up the test environment. This will spin up all relevant servers and frontend services. From the root directory of this repo:
+
+```
+nox -s test_env
+```
+
+Admin UI will be found at `localhost:3000` and Privacy Center at `localhost:3001`.
```
ssangervasi, line 13: This is maybe more of a limitation of `test_env`, but testing against `localhost:3000` is less than ideal:
 - Pages will have inconsistent (slow) load time because it runs the dev server, so the dev server could introduce failures (false negative).
 - API requests will be happening through Next's redirects, so the dev server could hide something that's really a failure (false positive).
 - Feature flags will come from `NEXT_PUBLIC_APP_ENV="development"`.

If we're requiring a build before running the tests, then the static app served on `localhost:8080` should 🤞 be the Production™️ version we really want to test against.

clients/cypress-e2e/cypress/fixtures/example.json
```diff
@@ -0,0 +1,5 @@
+{
+  "name": "Using fixtures to represent data",
+  "email": "hello@cypress.io",
+  "body": "Fixtures are a great way to mock data for responses to routes"
```
ssangervasi, line 4: We won't have fixture for mocking requests. Maybe we'll find them useful for defining a bunch of test cases, but for now this might be misleading.

ssangervasi (approved):
🚬  Most of my comments could be spun out as enhancement tickets, maybe TODO once we have more cases in mind.

clients/cypress-e2e/cypress/fixtures/example.json
allisonking: good point, will delete!

clients/cypress-e2e/cypress/support/constants.ts
allisonking: hmm okay I'd like to punt on this one until this becomes inconvenient. mostly because I'm not sure how best to use env info in a convenient way yet, but maybe that'll become clear once we arrive at a real problem

clients/cypress-e2e/README.md
allisonking: okay it took me a bit to wrap my head around how `test_env` works, but it looks like we can make this happen with [two small but potentially consequential changes](https://github.com/ethyca/fides/compare/aking/1293/cypress-test-env...aking/1293/cypress-prod-test-env).

1. Change the `test_env` image to build the prod image instead of the dev image. This will build the admin UI such that it can be served on `localhost:8080`
2. Add an exception to our volume mount so that it does not overwrite the built frontend assets

However, I imagine there was a reason to use the `dev` image for `test_env`. I think using the prod image instead of the dev image makes it so the package won't actually reflect dev changes (maybe because it doesn't get the `-e` editable flag?), so making this change would probably mess with other developers' flows.

I _think_ 2 is safe to do regardless, and might even prevent some common mistakes, but I haven't thought through all the use cases yet.

One thing we could do is add a totally new command, something like `nox -s prod_test_env` which uses the prod image instead of the dev image, and then we use that one for cypress instead of the current `test_env`. Might also address this old issue https://github.com/ethyca/fides/issues/928

This might be more of a question of what `test_env` _should_ be. Should it mimic a prod env as much as possible, or should it be a developer test environment? @NevilleS , any thoughts on how to move forward here?

NevilleS: I'd prefer to have `test_env` be a production environment myself. You can get some additional feedback from the eng team to see how they are using it, but the more realistic that environment is the better it'll work for us.

Note that it'll also be faster to test with and should be less of a resource hog if you tighten up that test environment to use fewer containers and production images, which is a bonus.

NevilleS: One note about this commit though @allisonking: https://github.com/ethyca/fides/compare/aking/1293/cypress-test-env...aking/1293/cypress-prod-test-env

That `build(test)` session is good but note that it doesn't force a rebuild of the `fides-privacy-center` image, which you'll also want for this.

NevilleS: See https://github.com/ethyca/fides/blob/67a50d16d97030d9bd7afb18b417db68209dc8be/noxfiles/docker_nox.py#L102

That could just be changed to also include the `test` target, and probably already should have!

allisonking: thanks! made a separate ticket for changing all that up so we can do it separate from this PR: https://github.com/ethyca/fides/issues/2385

clients/cypress-e2e/cypress/support/constants.ts
allisonking: never mind, I thought through it and realized how to do it now 😄 https://github.com/ethyca/fides/pull/2241/commits/3fd039668b51db8a54323aa09abced0b826a594c

that should satisfy the req
> MUST support targeting different hosts via environment variables, so we can run against staging / demo / etc.

clients/cypress-e2e/README.md
allisonking: ah, it loads so much faster now! hopefully will be more consistent, too. ez [change](https://github.com/ethyca/fides/pull/2241/commits/b765637d1746bddccbea56812e007547ec04f56a) once https://github.com/ethyca/fides/pull/2399 went in

allisonking:
privacy center check passed locally. the CI test seems to be failing across multiple branches, will need to look into it separately...

PR #2249, opened 2023-01-16
Support connectors in datamap api

TheAndrewJackson:
Closes https://github.com/ethyca/fidesplus/issues/498

### Code Changes

* [ ] Add `ctl_systems` foreign key to the `connnectionconfig` table
* [ ] Added new connection patch route that is identical other than requiring a system fides key in order to call it. It links the connections to that system
* [ ] Added new connection get route that returns all of the connections related to a particular system
* [ ] Updated the base [`get_connections`](https://github.com/ethyca/fides/blob/2607a789d87274a7cc8a1d6578bbfd14fddf9f00/src/fides/api/ops/api/v1/endpoints/connection_endpoints.py#L107) route to have a filter for connections not linked with a system

### Steps to Confirm

**New routes**
* [ ] Start server with `nox -s dev`
* [ ] Create a system through postman, pgadmin, etc
* [ ] Call the new patch route and create new connections linked with that system
* [ ] Call the new get route and check that it return all of the connections linked with that system

**Updated route**
* [ ] Start server with `nox -s dev`
* [ ] Create some connections with the original connections patch route
* [ ] Call the base [`get_connections`](https://github.com/ethyca/fides/blob/2607a789d87274a7cc8a1d6578bbfd14fddf9f00/src/fides/api/ops/api/v1/endpoints/connection_endpoints.py#L107) route with `orphaned_from_system=true` as the query param and make sure it returns all of the connections without linked systems.

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

These changes are for future UI updates that will have connectors integrate with the datamap UI. This PR will also support future changes to system and connector creation. Eventually the UI will enforce that all new connections are linked to a system upon creation. The new patch route will enforce that from the backend. 

The new `orphaned_from_system` query param is to give us a way to help users migrate their orphaned connections and link them with systems.


src/fides/api/ctl/routes/crud.py
```diff
@@ -33,7 +33,7 @@
 from fides.api.ctl.utils.api_router import APIRouter
 
 # CRUD Endpoints
-routers = []
+routers: List[APIRouter] = []
```
TheAndrewJackson, line 36: I noticed in my testing that the typing wasn't properly recognized so I went ahead and updated this.

src/fides/api/ctl/routes/system.py
```diff
@@ -0,0 +1,211 @@
...
+from pydantic.types import conlist
+from sqlalchemy.orm import Session
+from starlette.status import (
+    HTTP_200_OK,
+    HTTP_400_BAD_REQUEST,
+    HTTP_404_NOT_FOUND,
+    HTTP_422_UNPROCESSABLE_ENTITY,
+)
+
+from fides.api.ctl.sql_models import System  # type: ignore[attr-defined]
```
TheAndrewJackson, line 17: Mypy was having a false positive saying that `System` didn't exist in that module.

src/fides/api/ctl/routes/system.py
```diff
@@ -0,0 +1,211 @@
...
+)
+from fides.api.ops.service.connectors.saas.connector_registry_service import (
+    create_connection_config_from_template_no_save,
+    load_registry,
+    registry_file,
+)
+from fides.api.ops.util.oauth_util import verify_oauth_client
+from fides.lib.exceptions import KeyOrNameAlreadyExists
+
+router = APIRouter(tags=["System"], prefix=f"{V1_URL_PREFIX}{SYSTEM_CONNECTIONS}")
```
TheAndrewJackson, line 49: These new routes still show up under the same `System` section in the swagger docs despite being attached to a completely separate router.

src/fides/api/ops/models/connectionconfig.py
```diff
@@ -13,6 +13,7 @@
     StringEncryptedType,
 )
 
+from fides.api.ctl.sql_models import System  # type: ignore[attr-defined]
```
TheAndrewJackson, line 16: Same mypy issue here.

src/fides/api/ops/service/connectors/saas/connector_registry_service.py
```diff
@@ -93,18 +94,20 @@ def create_connection_config_from_template_no_save(
...
+    data = {
+        "name": template_values.name,
+        "key": template_values.key,
+        "description": template_values.description,
+        "connection_type": ConnectionType.saas,
+        "access": AccessLevel.write,
+        "saas_config": config_from_template,
+    }
+
+    if system_id:
```
TheAndrewJackson, line 106: This function needed a small update to support the new patch route.

src/fides/api/ctl/routes/system.py
```diff
@@ -0,0 +1,211 @@
...
+        raise HTTPException(
+            status_code=HTTP_404_NOT_FOUND,
+            detail="A valid system must be provided to create or update connections",
+        )
+    return system
+
+
+@router.get(
+    "",
+    dependencies=[Security(verify_oauth_client, scopes=[CONNECTION_READ])],
```
pattisdr, line 64: It's an interesting side effect of combining ctl and ops-like routes, but the base system routes are not authenticated but you need a token with scopes to get the system connections.  I think Thomas is working on adding authentication to some of the other ctl routes.

src/fides/api/ops/api/v1/endpoints/connection_endpoints.py
```diff
@@ -167,6 +168,9 @@ def get_connections(
             ConnectionConfig.last_test_succeeded.is_(test_status.str_to_bool())
         )
 
+    if orphaned_from_system:
+        query = query.filter(ConnectionConfig.system_id.is_(null()))
```
pattisdr, line 172: This could give unexpected results on `?orphaned_from_system=false`, I might make the default be None above, so you have three states:

If no query param value: all connections returned
If ?orphaned_from_system=true:  only connections with no system_id returned
If ?orphaned_from_system=false: only connections *with* a system id are returned

Can you add unit tests for this filter?

src/fides/api/ctl/routes/system.py
```diff
@@ -0,0 +1,211 @@
...
+    "",
+    dependencies=[Security(verify_oauth_client, scopes=[CONNECTION_CREATE_OR_UPDATE])],
+    status_code=HTTP_200_OK,
+    response_model=BulkPutConnectionConfiguration,
+)
+def patch_connections(
+    fides_key: str,
+    configs: conlist(CreateConnectionConfigurationWithSecrets, max_items=50),  # type: ignore
+    db: Session = Depends(deps.get_db),
+) -> BulkPutConnectionConfiguration:
```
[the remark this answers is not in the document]
pattisdr: I guess I also don't understand the reasoning behind creating a separate route here - could the `system_id` be added as an optional param to the other route? Or is it that the other route will eventually be deprecated?

src/fides/api/ops/api/v1/endpoints/connection_endpoints.py
```diff
@@ -112,6 +112,7 @@ def get_connections(
     disabled: Optional[bool] = None,
     test_status: Optional[TestStatus] = None,
     system_type: Optional[SystemType] = None,
+    orphaned_from_system: Optional[bool] = False,
     connection_type: Optional[List[str]] = Query(default=None),  # type: ignore
 ) -> AbstractPage[ConnectionConfig]:
```
pattisdr, line 117: Would it be useful to add `system_id` to the response here, or the response to GET /connection, when you're accessing connections directly and not through a system?

tests/ops/api/v1/endpoints/test_system.py
```diff
@@ -0,0 +1,120 @@
...
+        assert resp.status_code == 401
+
+    def test_get_connections_wrong_scope(
+        self, api_client: TestClient, generate_auth_header, connection_config, url
+    ) -> None:
+        auth_header = generate_auth_header(scopes=[STORAGE_DELETE])
+        resp = api_client.get(url, headers=auth_header)
+        assert resp.status_code == 403
+
+    def test_get_connection_configs(
```
pattisdr, line 45: Nice test - another test to add would be when the system id is invalid to test your `validate_system` code

pattisdr (changes requested):
Sorry for the delay reviewing this @TheAndrewJackson I kept getting sidetracked on other things.  Good set of changes here to link connectors to systems in a non-aggressive way.  Requested changes should be quick to address.

src/fides/api/ctl/routes/system.py
TheAndrewJackson: AFAIK yes. I tried adding them to the original router from the crud file and that sort of worked but it was a little buggy and complicated. I tried this second and worked first try. I guess the swagger docs code is smart enough to group routes with the same tags together

src/fides/api/ctl/migrations/versions/5b4b9c2d1c93_add_system_foreign_key_to_connection_.py
```diff
@@ -0,0 +1,40 @@
...
+    )
+    op.create_foreign_key(
+        None, "connectionconfig", "ctl_systems", ["system_id"], ["id"]
+    )
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_constraint(None, "connectionconfig", type_="foreignkey")
```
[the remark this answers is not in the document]
TheAndrewJackson: Good catch. I'll make sure to take a closer look at migrations for this in the future. 

I think we used to have a PR checklist step on the ops team to test the up/down migrations whenever a new one is added.

CHANGELOG.md
```diff
@@ -35,6 +35,7 @@ The types of changes are:
 * Adds support for Twilio email service (Sendgrid) [#2154](https://github.com/ethyca/fides/pull/2154)
 * Access and erasure support for Recharge [#1709](https://github.com/ethyca/fides/pull/1709)
 * Access and erasure support for Friendbuy Nextgen [#2085](https://github.com/ethyca/fides/pull/2085)
+* Add new connection config routes that couple them with systems [#2249](https://github.com/ethyca/fides/pull/2249)
```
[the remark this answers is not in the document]
TheAndrewJackson: It's all good. This work isn't used anywhere currently. I'll start working on the UI for this in the next sprint or so. The next release is fine by me 👍

src/fides/api/ctl/routes/system.py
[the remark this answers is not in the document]
TheAndrewJackson: Good question. It could have been implemented that way. The main reason I implemented it this way is the UI is going to be updated to only allow connections to be created if they're associated with a system. So orphaned systems will go away.

I decided to do it this way so that would be enforced from the backend. Forming the route this way made sense to me at the time because that would match the flow they were describing. I modeled it after the connection/dataset routes. For example, `/connection/{connection_key}/dataset`. `/system/{system_fides_key}/connection` follows that.

I think it would make sense to deprecate the original method once the UI has been updated. That wasn't talked about though.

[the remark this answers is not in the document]
TheAndrewJackson: If we were to support both ways long term then the query param would have been better. I'm _pretty_ sure were moving solely to this new way so I made this new route instead

src/fides/api/ctl/routes/system.py
```diff
@@ -0,0 +1,211 @@
...
+    load_registry,
+    registry_file,
+)
+from fides.api.ops.util.oauth_util import verify_oauth_client
+from fides.lib.exceptions import KeyOrNameAlreadyExists
+
+router = APIRouter(tags=["System"], prefix=f"{V1_URL_PREFIX}{SYSTEM_CONNECTIONS}")
+
+
+def validate_system(db: Session, fides_key: str) -> System:
```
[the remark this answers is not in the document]
TheAndrewJackson: Sure 👍

src/fides/api/ctl/routes/system.py
TheAndrewJackson: There are some interesting differences like sync vs async db sessions.

src/fides/api/ops/api/v1/endpoints/connection_endpoints.py
TheAndrewJackson: Good idea. I'll add a unit test for this.

src/fides/api/ops/api/v1/endpoints/connection_endpoints.py
pattisdr: that's a good point about the id versus fides_key, I forgot about that.

TheAndrewJackson:
@pattisdr Thanks for the awesome review! All of your feedback should be implemented now. I'm not sure why so many of the CI jobs are failing. It looks like none of the tests are actually running. I'll look into it

src/fides/api/ops/util/connection_util.py
```diff
@@ -0,0 +1,243 @@
+from typing import List, Optional
```
TheAndrewJackson, line 1: Code climate is pretty unhappy about this file. All of this code was cut/pasted from `connection_endpoints.py` to avoid circular reference errors. None of this code is actually new so it should be safe to merge.

TheAndrewJackson:
@pattisdr Everything is fixed now 👍 Now the only issue is the code climate stuff I mentioned

pattisdr:
thanks @TheAndrewJackson will take a look

src/fides/api/ctl/routes/system.py
pattisdr: oh gosh yes that distinction between ctl and ops endpoints is causing headaches

src/fides/api/ctl/routes/system.py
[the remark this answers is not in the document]
pattisdr: ok that's fine

CHANGELOG.md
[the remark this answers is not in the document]
pattisdr: This is still in the last release's changelog -

src/fides/api/ctl/migrations/versions/5b4b9c2d1c93_add_system_foreign_key_to_connection_.py
```diff
@@ -0,0 +1,46 @@
+"""add_system_foreign_key_to_connection_config
+
+Revision ID: 5b4b9c2d1c93
+Revises: 3caf11127442
```
pattisdr, line 4: You might also bump the downrev in the code comment too while we're here since the changelog needs to be updated.

pattisdr (approved):
great work @TheAndrewJackson thanks for those refactors.  just some housekeeping and it's good to go

Issue #1852, opened 2022-11-23
Ensure that oauth root client can perform all API actions


adamsachs:
### Is your feature request related to a specific problem?

While the root oauth client is technically able to invoke all API endpoints, there are some db constraints/oddities that prevent the client from effectively being able to perform the API actions. The root client should be able to effectively use all of our APIs.

One notable instance of this is that the root client is unable to update the default DSR access policy (or any policy?) rule, i.e. using the `PATCH /api/v1/dsr/poilcy/default_accesss_policy/rule` endpoint. In this case, a 500 is returned with this underlying exception: `ForeignKeyViolation('insert or update on table "rule" violates foreign key constraint "rule_client_id_fkey"\nDETAIL:  Key (client_id)=(fidesadmin) is not present in table "client".\n')`

https://github.com/ethyca/fides/pull/1698 was another instance of the problem that has already been resolved.

There may be (likely are) more instances of this problem, so we should try to identify and fix them all. 

### Describe the solution you'd like

- Fix whatever DB constraints are preventing the root client from being able to perform an API action.

### Describe alternatives you've considered, if any

- existing workaround is to[ create a new client with all scopes](https://ethyca.github.io/fides/2.1.0/guides/oauth/#create-additional-clients), and to perform API actions as that client. as far as we know, that client doesn't encounter any problems.

### Additional context

- We hit this issue when walking through steps for a client deployment, and attempting to complete them as the root client
- cc @NevilleS

NevilleS:
Thanks for writing this up @adamsachs - this is a footgun that'll sneak up on us in other places too, so I'd like to understand how difficult this would be to tackle

adamsachs:
wanted to bump this for priority as it's come up multiple times in the wild, most recently for @nicolas-ethyca on a presales effort. specifically - updating policies and rules is table-stakes functionality for our app,  we currently tend to use the oauth root user for most of our setup, and the workaround is pretty cumbersome. so this has a pretty big impact.

cc @Kelsey-Ethyca @mfbrown @NevilleS

PR #2267, opened 2023-01-18
Add option for users to protect all API endpoints

### LINKED ISSUE #2165: Add authentication to all unauthenticated API endpoints

There are unauthenticated endpoints related to `ctl` functionality

Once we have the ability to authenticate as a user and store credentials locally with `fides user login` (#2153 ) we can add authentication to all endpoints and make authentication a required part of the CLI application

ThomasLaPiana:
Closes #2165 

### Code Changes

* [x] add a `FIDES__SECURITY__ENV` option to lockdown the API to various degrees
* [x] add required scopes for the CLI-related endpoints
* [x] add logic to `main.py` to enable the authentication
  * [x] `dev` should not lock down CLI-related commands (current behavior, also the now default behavior)
  * [x] `demo` should remove all auth
  * [x] `production` should include auth for virtually all endpoints
* [x] add tests to verify functionality
* [x] cheeky addition: add the fides version to the FastAPI application version

### Steps to Confirm

* [ ] Verify `demo` env (**removed in the latest commit**)
  * [ ] set the `security.env` value in `fides.toml` to "demo"
  * [ ] `nox -s teardown -- volumes`
  * [ ] start the webserver with `nox -s dev`
  * [ ] navigate to `localhost:8080/docs` and hit every authenticated endpoint you can find (they have the padlock icon next to them). Confirm they all work!
* [ ] Verify `dev` env (the default behavior that matches current behavior)
  * [ ] set the `security.env` value in `fides.toml` to "dev"
  * [ ] `nox -s teardown -- volumes`
  * [ ] start the webserver with `nox -s dev`
  * [ ] navigate to `localhost:8080/docs` and hit the ops-related endpoints such as:
    * `users`
    * `cryptography/encryption_key`
  * [ ] Authenticate yourself via the Swagger page and confirm the authenticated endpoints work as expected
  * [ ] run various CLI commands to verify that still functions:
    * `fides ls data_category`
    * `fides push`
* [ ] Verify `production` env
  * [ ] set the `security.env` value in `fides.toml` to "prod"
  * [ ] `nox -s teardown -- volumes`
  * [ ] start the webserver with `nox -s dev`
  * [ ] navigate to `localhost:8080/docs` and confirm _ALL_ authenticated endpoints fail
  * [ ] authenticate via the Swagger page and verify that all endpoints now work
  * [ ] run commands from the CLI and confirm that they fail.
  * [ ] run `fides user login` using the root username and password
  * [ ] confirm the CLI commands now work

### Pre-Merge Checklist

* Documentation:
  * [x] documentation [issue created in fidesdocs](https://github.com/ethyca/fidesdocs/issues/new/choose) (https://github.com/ethyca/fidesdocs/issues/46)
* [x] Update `CHANGELOG.md`

### Description Of Changes

The changes in this PR are a significant step forward for our application security, but as a result are important to understand in the context of how it affects how users interact with fides.

There is a new `env` settings under the `security` section of the config. It is an Enum with 3 possible options, `dev`, `prod`, and `demo`. They function as follows:
* `demo` - bypasses all oauth for all endpoints. Zero security and is specifically meant for demoing/zero-risk environments.
* `dev` - This mirrors the application's current behavior and is the default value. Nothing changes here.
* `prod` - This adds auth checks to every single endpoint (where it makes sense), which for all intents and purposes is the previously unprotected endpoints that were used primarily by the CLI/ctl logic. Users will have to leverage the new `fides user login` flow and make sure permissions are assigned accordingly.

This is accomplished in a bit of a tricky manner. There are two `verify_oauth_client` functions essentially, with the new one being called `verify_oauth_client_cli`. It is a literal programmatic copy of `verify_oauth_client`. All of the newly protected CLI-related endpoints use this new copied function for oauth verification.

The reason for this is in `main.py`, where we can now override the FastAPI security dependencies depending on the value of `config.security.env`.  

#### Fixed API Swagger Version

image


ThomasLaPiana:
The changes specifically request by @NevilleS here:

* Add a new setting var, FIDES__SECURITY__ENV
* Default it to "development" but have another option for "production"
* When env is production, wrap all API calls inside oauth scopes, with the exception of a list of "privacy center" endpoints needed for the privacy center
* When env is "development", keep the current setup for all the fidesctl APIs being open
* When env is "demo" remove auth from everything - including all the fidesops ones

ThomasLaPiana:
At this point the PR is feature-complete, but given the major changes here I'll need to fix/update tests

ThomasLaPiana:
@NevilleS is there any reason now to also use `prod` for our own development? I think this would be a good way to drink our own champagne, since we expect our users to also be using that mode most of the time.

with `fides user login` it is also trivial to store the credentials file for reuse. Thoughts?

src/fides/api/ctl/routes/admin.py
```diff
@@ -16,7 +20,13 @@ class DBActions(str, Enum):
...
 
-@router.post("/admin/db/{action}", tags=["Admin"])
+@router.post(
+    "/admin/db/{action}",
+    tags=["Admin"],
+    dependencies=[
+        Security(verify_oauth_client_cli, scopes=[scope_registry.DATABASE_RESET])
+    ],
+)
 async def db_action(action: DBActions) -> Dict:
```
seanpreston, line 30: It's conceivable that a user would want to disable the ability to reset a DB on a running server entirely so I've added a ticket to implement that [here](https://github.com/ethyca/fides/issues/2319).

src/fides/api/main.py
```diff
@@ -81,6 +86,18 @@
     view.router,
 ]
 
+if CONFIG.security.env.value == "dev":
```
seanpreston, line 89: Some explicit logging in these blocks would go a long way for debugging

src/fides/api/ops/api/v1/scope_registry.py
```diff
@@ -34,34 +68,20 @@
 PRIVACY_REQUEST_UPLOAD_DATA = "privacy-request:upload_data"
 PRIVACY_REQUEST_VIEW_DATA = "privacy-request:view_data"
 
-WEBHOOK_CREATE_OR_UPDATE = "webhook:create_or_update"
```
seanpreston, line 37: Thanks for alphabetizing this!

src/fides/api/main.py
ThomasLaPiana: Wouldn't that be part of the `config` though? Since these values are pulled directly from the config, I think looking at the config is enough?

ThomasLaPiana:
@NevilleS Are you strongly against not changing the default behavior? I think `prod` should be the default, as that is going to be the most secure for our users, but will also be the most disruptive for our current users which is the primary downside.

This might be a band-aid worth ripping off though, unless we want to add some warning messaging to the logs if the user is using the default value.

ThomasLaPiana:
I'm running into a specific problem related to async tests with the test client and am going to try this other package as per starlette's own recommendations: https://github.com/encode/starlette/issues/487

tests/ops/api/v1/test_main.py
```diff
@@ -1,13 +1,120 @@
...
+        test_app = configure_security_env_overrides(
+            application=test_app, security_env=security_env
+        )
+        with TestClient(test_app) as test_client:
+            response = test_client.get(f"/api/v1/{endpoint}")
+            print(response.url)
+            print(response.text)
+            assert response.ok
+
+    def test_dev_endpoints_protected(self) -> None:
```
ThomasLaPiana, line 82: @sanders41 @allisonking any guidance here on what might be going wrong? I'm using the test client, and the first test _always_ passes, but the rest fail and complain about the event loop. Any idea how to potentially solve this? I've googled around but haven't found anything that's worked

ThomasLaPiana:
@NevilleS do you have any objections to using `prod` for ourselves, even during development? We can leave `dev` as the application default, but I think we should develop from the place that we would recommend to users. We can add a follow-up ticket to make that change since it would be disruptive to tests in some places

src/fides/api/main.py
```diff
@@ -92,6 +89,72 @@
...
+        application.dependency_overrides[verify_oauth_client] = get_root_client
+        application.dependency_overrides[verify_oauth_client_cli] = get_root_client
+    elif security_env == "prod":
+        # This is the most secure, so all security deps are maintained
+        pass
+    return application
+
+
+app = create_fides_app()
+app = configure_security_env_overrides(application=app)
```
ThomasLaPiana, line 155: should these actually just be a single function since we're doing all other configuration in `create_fides_app`?

tests/ops/api/v1/test_main.py
sanders41: The `event_loop` fixture should be putting everything in the same loop and preventing this. https://github.com/ethyca/fides/blob/502d689676d5bd77b14875e1e1c3e80b5feb2143/tests/ops/conftest.py#L131

If you remove `@pytest.mark.asyncio` do you still get the error? That decorator lets the test itself be async so `async def test_demo_endpoints_unpretected...```. I wonder if the error is that the test doesn't get awaited since it isn't async.

ThomasLaPiana: yeah I do, I added that as an attempt to fix things...but its not necessary when using `TestClient` technically since it runs things synchronously

ThomasLaPiana:
After talking with @daveqnet we decided to ship the PR as-is with the current working tests to make it into the `2.6` release. We will then immediately open another PR with the regression tests and get those shipped for the `2.7` release

tests/ops/api/v1/test_main.py
sanders41: Are you able to successfully run this endpoint outside of tests? I tried isolating one of the tests with `pytest -k  test_dev_endpoints_protected`and not using fixtures but still got the same error. The error always says postgres has the event loop issue.

I think the answer here is going to be yes, I just want to make sure so we know which path to go down.

```py
from httpx import AsyncClient

...

    @pytest.mark.asyncio
    async def test_dev_endpoints_protected(self) -> None:
        """
        Test that the non-cli-related endpoints are protected in
        a 'dev'-tier security environment.
        """
        security_env = "dev"
        endpoint = "data_category"
        test_app = create_fides_app()
        test_app = configure_security_env_overrides(
            application=test_app, security_env=security_env
        )
        async with AsyncClient(
            app=test_app, base_url="http://0.0.0.0:8080", follow_redirects=True
        ) as client:
            response = await client.get(f"/api/v1/{endpoint}")
            print(response.url)
            print(response.text)
            assert response.status_code == 200
```

NevilleS:
Heya, sorry for the lack of feedback on this. The main question directed at me was what the _default_ should be -- I'd like to stick to my original rec that we default to `dev` in this release (to not ship a breaking change) but do believe that a plan to update that default to `prod` in a future release (maybe a 3.0.0?) does make sense.

I'm very OK with setting the env to `prod` in all our local development

ThomasLaPiana:
> Heya, sorry for the lack of feedback on this. The main question directed at me was what the _default_ should be -- I'd like to stick to my original rec that we default to `dev` in this release (to not ship a breaking change) but do believe that a plan to update that default to `prod` in a future release (maybe a 3.0.0?) does make sense.
> 
> I'm very OK with setting the env to `prod` in all our local development

Sounds good. Let's not disrupt (for now) and I'll update our project default in a follow-up PR

tests/ops/api/v1/test_main.py
ThomasLaPiana: @sanders41 I'm going to remove these tests from this PR and add them to #2463 since we can manually verify these changes here, and then those tests will help prevent regressions in the future

ThomasLaPiana: yep, I've confirmed the endpoints work via manual testing

pattisdr (commented):
Really slick way this was implemented. Nice work - just a few questions

src/fides/api/main.py
```diff
@@ -92,6 +89,62 @@
...
+    fastapi_app.include_router(api_router)
+
+    if security_env == "dev":
+        # This removes auth requirements for CLI-related endpoints
+        # and is the default
+        fastapi_app.dependency_overrides[verify_oauth_client_cli] = get_root_client
+    elif security_env == "demo":
+        # This remove all auth requirements
+        fastapi_app.dependency_overrides[verify_oauth_client] = get_root_client
+        fastapi_app.dependency_overrides[verify_oauth_client_cli] = get_root_client
```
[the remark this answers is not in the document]
seanpreston: Will factor this in when [consolidating the methods](https://github.com/ethyca/fides/issues/2320), thanks for flagging this @pattisdr

src/fides/api/ops/api/v1/scope_registry.py
```diff
@@ -1,14 +1,28 @@
+"""
+This module contains every scope used in the application.
+
+The format for defining a scope is:
+    <SCOPE_SECTION>_<SCOPE_NAME> = "<scope_section>:<scope_name>"
+    CLIENT_CREATE = "client:create"
+
+`SCOPE_REGISTRY` is intended as a comprehensive list of all available scopes.
+
+There is a redundant `fides/lib/oauth/scopes.py` file that should not be used.
```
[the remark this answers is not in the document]
seanpreston: I've added a [follow-up ticket](https://github.com/ethyca/fides/issues/2471).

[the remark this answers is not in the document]
ThomasLaPiana: good call @pattisdr , and thanks @seanpreston !

src/fides/api/main.py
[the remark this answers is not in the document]
ThomasLaPiana: good call @pattisdr I'll add that

src/fides/core/annotate_dataset.py
```diff
@@ -109,7 +109,7 @@ def annotate_dataset(
     resources = api_helpers.list_server_resources(
         url=str(config.cli.server_url),
         resource_type=resource_type,
-        headers=config.user.request_headers,
+        headers=config.user.auth_header or {"Authorization": "Bearer fakekey"},
```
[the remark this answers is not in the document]
ThomasLaPiana: mypy thing, will note in a comment

src/fides/api/ops/util/oauth_util.py
```diff
@@ -115,6 +132,19 @@ def verify_callback_oauth(
     return webhook
 
 
+async def get_root_client(
+    db: Session = Depends(get_db), client_id: str = CONFIG.security.oauth_root_client_id
```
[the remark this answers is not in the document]
ThomasLaPiana: I see 🤔 

It doesn't feel good to me that we would be shipping something that we know is broken in a specific state, but here would be my justification:
* We have a ticket open for it, this is a known bug across the application and isn't specific to this use-case
* `Demo` mode should primarily be used by us to demo, and in that case we could take the extra step of adding a client (but then at that point why bother having a `demo` env?)
* Figuring out a solution that instead entirely removes auth as opposed to overwriting it would be required

I think we have two potential solutions here:
1. Remove the `demo` mode altogether, which also removes the bug. This is a new feature so we wouldn't be "losing" functionality
2. Ship as-is and document the caveat.

I'm leaning towards #1, but could understand #2 as well given that we are aware of this bug and are working to quash it.

Thoughts @pattisdr @seanpreston ?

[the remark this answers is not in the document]
pattisdr: I'd lean towards `#1` too personally and we prioritize getting https://github.com/ethyca/fides/issues/1852 fixed first.  It would be worth revisiting if we needed client id persisted on these resources in the first place with the direction the product is going.

ThomasLaPiana:
I went ahead and removed the `demo` option in case that is the route we go with. Can revert the latest commit if needed but otherwise will be ready to go

pattisdr (approved):
Great @ThomasLaPiana

pattisdr:
Merging on your behalf @ThomasLaPiana

PR #2315, opened 2023-01-21
Consent SaaS Config: Allow definition of multiple opt-in or opt-out requests

pattisdr:
Unticketed

### Code Changes

* [ ] Make the `consent_requests` section on the saas config have `opt_in` and `opt_out` types of `Union[SaaSRequest, List[SaaSRequest]] = []`, so the dev can define a single request or a list of requests.  Validators turn everything into a list of requests
* [ ] Adjust the SaaSConnector to fire either the list of opt in or the list of opt out requests, instead of the single request
* [ ] Adjust MailchimpTransactional connector to fire two opt-out requests - the current requests being fired I don't think have the desired effect when opting out.

https://mailchimp.com/developer/transactional/api/allowlists/remove-email-from-allowlist/
> POST
/allowlists/delete
Removes an email address from the allowlist.

https://mailchimp.com/developer/transactional/api/rejects/add-email-to-denylist/
> Add email to denylist
POST
/rejects/add
Adds an email to your email rejection denylist. Addresses that you add manually will never expire and there is no reputation penalty for removing them from your denylist. Attempting to denylist an address that has been added to the allowlist will have no effect.

Note that opt-in, adding the user to the allowlist, still requires just one request:
https://mailchimp.com/developer/transactional/api/allowlists/add-email-to-allowlist/
> Add email to allowlist
Collapse
POST
/allowlists/add
Adds an email to your email rejection allowlist. If the address is currently on your denylist, that denylist entry will be removed automatically.

### Steps to Confirm

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

The current consent_requests saas config section was set up to allow one opt-out or one opt-in request to be defined, but I think we're already hitting a scenario where a string of requests may need to be fired.

Allow optionally multiple opt in or opt out requests to be defined, that will be fired in sequence for consent requests.

- Adjust the Mailchimp Transactional connector to first remove the user from the allow list, before adding them to the denylist when the user opts out. On a closer read-through of the docs, I think we need to fire two requests.


src/fides/api/ops/service/connectors/saas_connector.py
```diff
@@ -429,32 +429,33 @@ def run_consent_request(
 
         for consent in consent_preferences:
             request_action: str = "opt_in" if consent.opt_in else "opt_out"
-            consent_request: Optional[
+            consent_requests: List[
                 SaaSRequest
             ] = self._get_consent_requests_by_preference(consent.opt_in)
 
-            if not consent_request:
+            if not consent_requests:
```
seanpreston, line 436: Love the attention to detail 🙏

src/fides/api/ops/schemas/saas/saas_config.py
```diff
@@ -214,8 +214,28 @@ class SaaSRequestMap(BaseModel):
...
+
+    @validator("opt_out")
+    def validate_opt_out(
+        cls,
+        opt_out: Union[SaaSRequest, List[SaaSRequest]],
+    ) -> List[SaaSRequest]:
+        """Convert all opt out request formats to a list of opt out requests"""
+        if isinstance(opt_out, SaaSRequest):
+            return [opt_out]
+        return opt_out
```
seanpreston, line 238: Consider combining both of these into one validator, I think `pydantic` allows you to specify multiple fields in the decorate definition like:

```
    @validator("opt_in", "opt_out")
    def validate_list_field(
        cls,
        field_value: Union[SaaSRequest, List[SaaSRequest]],
    ) -> List[SaaSRequest]:
        if isinstance(field_value, SaaSRequest):
            return [field_value]
        return field_value
```

pattisdr: thank you Sean, I hadn't used this variant before, either single validators, or root validators, this seems like a good use case for specifying multiple fields.

src/fides/api/ops/service/connectors/saas_connector.py
pattisdr: thank you!

pattisdr:
Mailchimp transactional opt in and opt out requests both verified locally, integration tests passing locally -

PR #2321, opened 2023-01-23
[#2319] Don't allow db resets unless `config.dev_mode` is `True`

### LINKED ISSUE #2319: Add config. var enabling user to prevent `db_action`s that could harm a running server
The Fides `DBActions` class supports the ability to [`reset`](https://github.com/ethyca/fides/blob/main/src/fides/api/ctl/routes/admin.py#L16) the application DB. While this is a very useful, it's also a footgun if used against a server currently serving privacy requests. We should add a config option to turn this off, or perhaps disable the functionality entirely outside of `DEV_MODE` in the API view [here](https://github.com/ethyca/fides/blob/main/src/fides/api/ctl/routes/admin.py#L20-L30).

seanpreston:
Closes #2319 

### Code Changes

* [ ] Raise a `FunctionalityNotConfigured` exception if a DB reset is called via the API against a server running outside of `dev_mode`

### Steps to Confirm

* [ ] Run `nox -s test_env`
* [ ] Send a `POST` request to `/admin/db/reset`
* [ ] Assert that the request fails

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

I opted to use `dev_mode` here instead of adding a new configuration variable to work against config. bloat, but I'm happy to change if anyone has a compelling reason to.

### DIFF
--- a/src/fides/api/ctl/routes/admin.py
+++ b/src/fides/api/ctl/routes/admin.py
@@ -3,6 +3,7 @@ from typing import Dict
 
 from fides.api.ctl.database import database
 from fides.api.ctl.routes.util import API_PREFIX
+from fides.api.ctl.utils import errors
 from fides.api.ctl.utils.api_router import APIRouter
 from fides.core.config import FidesConfig, get_config
 
@@ -24,6 +25,11 @@ async def db_action(action: DBActions) -> Dict:
 
     action_text = "initialized"
     if action == DBActions.reset:
+        if not CONFIG.dev_mode:
+            raise errors.FunctionalityNotConfigured(
+                "unable to reset fides database outside of dev_mode."
+            )
+
         database.reset_db(CONFIG.database.sync_database_uri)
         action_text = DBActions.reset
     await database.configure_db(CONFIG.database.sync_database_uri)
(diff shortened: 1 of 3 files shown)


seanpreston:
I'm guessing that since this hasn't broken any tests this endpoint is untested — I'll add some and leave the PR in draft state until that's done.

NevilleS:
I like this @seanpreston! Totally fine to ship this as is, but IMO the `FIDES__DEV_MODE` setting is (effectively) an unknown, largely undocumented feature 😢

As a follow-up: I'd like us to deprecate `FIDES__DEV_MODE` in favour of the newly proposed `FIDES__SECURITY__ENV ` setting here: https://github.com/ethyca/fides/pull/2267. That one lets us support multiple environments (`dev`, `demo`, `production`) which makes it more recognizable to new users - see `FLASK_ENV`, `RAILS_ENV`, `NODE_ENV`, etc.

seanpreston:
> As a follow-up: I'd like us to deprecate FIDES__DEV_MODE in favour of the newly proposed FIDES__SECURITY__ENV  setting here: https://github.com/ethyca/fides/pull/2267. That one lets us support multiple environments (dev, demo, production) which makes it more recognizable to new users - see FLASK_ENV, RAILS_ENV, NODE_ENV, etc.

Absolutely, thanks for flagging.

allisonking (approved):

Issue #2228, opened 2023-01-13
Listen to GPC Signal and set initial consent properties accordingly


mfbrown:
As a user, when I have a global privacy control signal, I want to be automatically opted out of any browser based tracking which serves the function of data sales or data sharing, so that I don’t need to verify my identity with the privacy center to have my GPC choice reflected.

**AC**

- Given that I have an opt-out signal set on my global privacy control, when I access a site with Fides consent, then any client side record of my consent is automatically updated to reflect my GPC signal.
- Given that I have a global privacy control opt-out signal set on my browser, when I access a site with Fides consent for the first time, then I am opted out of systems designated as supporting data sales or sharing.
- Given that I have a global privacy control opt-out signal set on my browser, when I access a site with Fides consent, then my consent object includes an indication that my choices are influenced by GPC
- Given that I have a global privacy control opt-out signal set on my browser, when I access a site with Fides consent, then my consent cookie includes an indication that my choices are influenced by GPC

ssangervasi:
This should be covered by the code changes in:
-  https://github.com/ethyca/fides/pull/2341#issue-1553979189
- https://github.com/ethyca/fides/pull/2596

There are cypress tests that cover the with/without GPC, and with/without the user having already stored their consent, using the static page at `/fides-consent-demo.html`.

I want to confirm the functionality more completely by building `fides-consent.js` and putting it into an external page: - 
- [ ] My [place for GTM testing](https://ethyca.atlassian.net/wiki/spaces/EN/pages/2563637409/j1407b+fides-consent.js+Google+Tag+Manager) is disabled at the moment, so I can't do that yet: https://ethyca.slack.com/archives/C01TUEAGCR1/p1676511482711419
- [ ] I should be able [test the Shopify page](https://ethyca.atlassian.net/wiki/spaces/EN/pages/2579038241/Shopify+fides-consent.js) though.

PR #2300, opened 2023-01-19
Add Redux Toolkit to the Privacy Center

ssangervasi:
Closes #2298 

### Code Changes

Setup:
- pc/package.json: Install compatible RTK, Redux Types, and Redux Persist
- pc/app: Initial app store, api, and config
- pc/types: OpenAPI generate types
  - Another stopgap until we solve https://github.com/ethyca/fides/discussions/2297

Payoff:
- pc: IdVerfication slice
- pc/index: Use RTK to query IdVerification config
- pc/consent: Config slice with consent option overrides
- pc/cy: Dispatch config overrides from tests

### Steps to Confirm

* [ ] Modify the id-verification requirements from the fides API. The PC should reflect the changes.
* [ ] Modify the consent options in config.json. The cypress tests should use the options defined within the test instead.

### Pre-Merge Checklist

* [ ] All CI Pipelines Succeeded
* Documentation:
  * [ ] documentation complete, [PR opened in fidesdocs](https://github.com/ethyca/fidesdocs/pulls)
  * [ ] documentation [issue created in fidesdocs](https://github.com/ethyca/fidesdocs/issues/new/choose)
* [ ] Issue Requirements are Met
* [ ] Relevant Follow-Up Issues Created
* [ ] Update `CHANGELOG.md`

### Description Of Changes

The "Setup" code changes add boilerplate and are basically invisible at runtime.

The "Payoff" changes make actual use of RTK. The id-verification code was a clean example of a GET request that can be swapped to RTK Query from a plain `fetch`.

The config slice its consent overrides should be familiar redux code. The Cypress usage is more interesting: we need to expose the store from the app, access the window variable from a cypress command, and actually try that out in a test. What I have here is just a starting point.

In Cypress, the "test" consent options are used by the app:
test-consent-config

Issue #2230, opened 2023-01-13
Reflect GPC choices in the privacy center


mfbrown:
As a user, when I have a global privacy control signal, then I want my opt-out status reflected the first time a visit the privacy center, so that I can confirm the signal is being respected.

**AC**

- Given that I have a global privacy control opt-out signal set on my browser, when I access a Fides privacy center, then I am automatically opted out of any use cases which are indicated as use cases in the privacy center.
- Given that I have a global privacy control opt-out signal set on my browser, when I access a Fides privacy center and the default consent choices are set according to GPC, then I see an indicator on each use case telling me that I am opted out because of GPC.

**Designs**
https://www.figma.com/file/dmEwdK3xZwjKfGVQ9t66xe/Fides-v.2-Master-Working-Files?node-id=1680%3A33346

1/11/23 - Linked designs are WIP.

mfbrown:
For the first time user - we opt follow the GPC signal and opt them out of use cases automatically.
For returning users that have a consent preference saved with their identity - we show the saved consent and then update the use case indicators appropriately if there is a conflict with GPC.

Indicators: 
![Image](https://user-images.githubusercontent.com/4807009/212379884-d74a4922-b002-41ae-ac64-6f05725f7e0c.png)

ssangervasi:
Pausing to work on some bugs, left off on this branch: https://github.com/ethyca/fides/compare/ssangervasi/fides/2230/pc-gpc-ui

Roger-Ethyca:
enabled 


![Image](https://user-images.githubusercontent.com/115798183/221278712-c60d5178-84d9-4942-9a41-8258f768beb7.png)


disabled


![Image](https://user-images.githubusercontent.com/115798183/221278784-be77cffc-c5bc-48b0-b55a-d7514f763faf.png)

PR #2341, opened 2023-01-23
pc/fides-consent: ConsentValue with ConsentContext to support GPC

### LINKED ISSUE #2227: Configure individual data use cases to adhere to GPC
As an engineer deploying the privacy center and setting consent options, when I’ve created my consent options, I want to specify whether they should adhere to a global privacy control signal, so that I can comply with GPC for sales and sharing use cases while still allowing opt-out consent for non sales/sharing.

**AC**
- Given that I am configuring the privacy center, when I add a data use case, then I can specify whether or not that use case is subject to GPC
- Given that I have changed the GPC settings of consent options in my privacy center config, when the privacy center is next built, then fides-consent.js is updated to apply the a GPC signal to each use case according to the config.

ssangervasi:
Closes #2227 
This might cover #2228 but I'll need to run through the AC more before closing it.

### Code Changes

- pc/fides-consent: ConsentValue with ConsentContext to support GPC
- pc/consent: Resolve ConsentValue with ConsentContext.
- pc/config: Default data_sales consent to false for GPC

### Steps to Confirm

* [ ] You can enable GPC in [Firefox](https://blog.mozilla.org/netpolicy/2021/10/28/implementing-global-privacy-control/) or fake it by navigating to the demo page with the query parameter set: http://localhost:3000/fides-consent-demo.html?globalPrivacyControl=true. Also make sure you don't have consent saved (this PR only covers the default behavior).
* [ ] With GPC enabled, the demo page should show `"data_sales": false`
* [ ] With GPC disabled, the demo page should show `"data_sales": true`

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### Description Of Changes

I'm going to comment on some relevant sections of the refactor.

### DIFF
--- a/.github/workflows/frontend_checks.yml
+++ b/.github/workflows/frontend_checks.yml
@@ -105,11 +105,11 @@ jobs:
       - name: Format
         run: npm run format:ci
 
-      - name: Jest test
-        run: npm run test:ci
-
       - name: Build
         run: npm run build
+
+      - name: Jest test
+        run: npm run test:ci
 
   Privacy-Center-Cypress:
     runs-on: ubuntu-latest
--- a/clients/privacy-center/config/config.json
+++ b/clients/privacy-center/config/config.json
@@ -44,7 +44,10 @@
         "name": "Data Sales or Sharing",
         "description": "We may use some of your personal information for behavioral advertising purposes, which may be interpreted as 'Data Sales' or 'Data Sharing' under regulations such as CCPA, CPRA, VCDPA, etc.",
         "url": "https://example.com/privacy#data-sales",
-        "default": true,
+        "default": {
+          "value": true,
+          "globalPrivacyControl": false
+        },
         "highlight": false,
         "cookieKeys": ["data_sales"],
         "executable": false
@@ -54,7 +57,10 @@
         "name": "Email Marketing",
         "description": "We may use some of your personal information to contact you about our products & services.",
         "url": "https://example.com/privacy#email-marketing",
-        "default": true,
+        "default": {
+          "value": true,
+          "globalPrivacyControl": false
+        },
         "highlight": false,
         "cookieKeys": ["tracking"],
         "executable": false
--- a/clients/privacy-center/cypress/e2e/consent.cy.ts
+++ b/clients/privacy-center/cypress/e2e/consent.cy.ts
@@ -233,5 +233,18 @@ describe("Consent settings", () => {
           ]);
       });
     });
+
+    describe("when globalPrivacyControl is enabled", () => {
+      it("uses the globalPrivacyControl default", () => {
+        cy.visit("/fides-consent-demo.html?globalPrivacyControl=true");
+        cy.get("#consent-json");
+        cy.window().then((win) => {
+          expect(win).to.have.nested.property("Fides.consent").that.eql({
+            data_sales: false,
+            tracking: false,
+          });
+        });
+      });
+    });
   });
 });
--- a/clients/privacy-center/packages/fides-consent/rollup.config.js
+++ b/clients/privacy-center/packages/fides-consent/rollup.config.js
@@ -23,40 +23,26 @@ const generateConsentConfig = () => {
    * @type {import('../../types/config').Config}
    */
   const privacyCenterConfig = require("../../config/config.json");
-  const consentOptions = privacyCenterConfig.consent?.consentOptions ?? [];
+  const privacyCenterOptions =
+    privacyCenterConfig.consent?.consentOptions ?? [];
 
-  if (consentOptions.length === 0) {
+  if (privacyCenterOptions.length === 0) {
     console.warn(
       "Privacy Center config.json has no consent options configured."
     );
   }
 
+  const options = privacyCenterOptions.map((pcOption) => ({
+    fidesDataUseKey: pcOption.fidesDataUseKey,
+    default: pcOption.default,
+    cookieKeys: pcOption.cookieKeys,
+  }));
+
   /**
-   * @type {import('./src/lib/cookie').CookieKeyConsent}
+   * @type {import('./src/lib/consent-config').ConsentConfig}
    */
-  const defaults = {};
-  consentOptions.forEach(({ cookieKeys, default: current }) => {
-    if (current === undefined) {
-      return;
-    }
-
-    cookieKeys.forEach((cookieKey) => {
-      const previous = defaults[cookieKey];
-      if (previous === undefined) {
-        defaults[cookieKey] = current;
-        return;
-      }
-
-      if (current !== previous) {
-        console.warn(`Conflicting configuration for cookieKey "${cookieKey}".`);
-      }
-
-      defaults[cookieKey] = previous && current;
-    });
-  });
-
   const consentConfig = {
-    defaults: defaults,
+    options,
   };
 
   fs.writeFileSync(
--- a/clients/privacy-center/packages/fides-consent/src/lib/consent-value.ts
+++ b/clients/privacy-center/packages/fides-consent/src/lib/consent-value.ts
@@ -1,0 +1,43 @@
+import { ConsentContext } from "./consent-context";
+
+export type ConditionalValue = {
+  value: boolean;
+  globalPrivacyControl: boolean;
+};
+
+/**
+ * A consent value can be a boolean:
+ *  - `true`: consent/opt-in
+ *  - `false`: revoke/opt-out
+ *
+ * A consent value can also be context-dependent, which means it will be decided based on
+ * information about the user's environment (browser). The `ConditionalValue` object maps the
+ * context conditions to the value that should be used:
+ *  - `value`: The default value if no context applies.
+ *  - `globalPrivacyControl`: The value to use if the user's browser has Global Privacy Control
+ *    enabled.
+ */
+export type ConsentValue = boolean | ConditionalValue;
+
+export type ConsentTypeToValue = {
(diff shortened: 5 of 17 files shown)


clients/privacy-center/config/config.json
```diff
@@ -43,7 +43,10 @@
         "name": "Data Sales or Sharing",
         "description": "We may use some of your personal information for behavioral advertising purposes, which may be interpreted as 'Data Sales' or 'Data Sharing' under regulations such as CCPA, CPRA, VCDPA, etc.",
         "url": "https://example.com/privacy#data-sales",
-        "default": true,
+        "default": {
+          "value": true,
+          "globalPrivacyControl": false
+        },
```
ssangervasi, line 49: This is the crux of the changes. [Here's the type definition](https://github.com/ethyca/fides/pull/2341/files#diff-103fdb8585246217bff5eb7f5eec4f86a0c93947dc122941b7952615c17dfce2R20) for the new config option. Note that a plain boolean is also valid and would be handled as if GPC doesn't matter, as before.

An alternative implementation would be to keep `default: boolean` and add a sibling config like `defaultWhenGlobalPrivacyControlEnabled: boolean` or something. That version seems more self-contradictory: "default" wouldn't really be the default 😬. With the `ConsentValue` object, we'll be able to add more options if/when we need them.

clients/privacy-center/cypress/e2e/consent.cy.ts
```diff
@@ -140,4 +140,16 @@ describe("Consent settings", () => {
       });
     });
   });
+
+  describe("when globalPrivacyControl is enabled", () => {
```
ssangervasi, line 144: Note that https://github.com/ethyca/fides/pull/2300 still wouldn't give us a way to dynamically change the config used by `fides-consent.js` because the browser script gets its config at build time. I'm considering adding a mechanism for the script to load a config at runtime, maybe by checking for a `window.Fides.config` object.

clients/privacy-center/packages/fides-consent/rollup.config.js
```diff
@@ -23,40 +23,26 @@ const generateConsentConfig = () => {
...
     console.warn(
       "Privacy Center config.json has no consent options configured."
     );
   }
 
+  const options = privacyCenterOptions.map((pcOption) => ({
+    fidesDataUseKey: pcOption.fidesDataUseKey,
+    default: pcOption.default,
+    cookieKeys: pcOption.cookieKeys,
+  }));
```
ssangervasi, line 39: The build-time transformation is actually much simpler now.

.github/workflows/frontend_checks.yml
```diff
@@ -105,12 +105,12 @@ jobs:
...
 
-      - name: Jest test
-        run: npm run test:ci
-
       - name: Build
         run: npm run build
 
+      - name: Jest test
+        run: npm run test:ci
+
```
ssangervasi, line 113: The fides-consent package needs to be built. And if build fails the tests are going to fail anyway.

clients/privacy-center/packages/fides-consent/src/lib/cookie.ts
```diff
@@ -25,8 +28,37 @@ const CODEC: Types.CookieCodecConfig<string, string> = {
...
+    const value = resolveConsentValue(current, context);
+
+    cookieKeys.forEach((cookieKey) => {
+      const previous = defaults[cookieKey];
+      if (previous === undefined) {
+        defaults[cookieKey] = value;
+        return;
+      }
+
+      defaults[cookieKey] = previous && value;
```
LKCSmith, line 53: I'm a little confused on what this value ends up being when setting `defaults[cookieKey]` like this. Does this simply return the boolean?

clients/privacy-center/packages/fides-consent/src/lib/consent-value.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export const resolveConsentValue = (
+  value: ConsentValue | undefined,
+  context: ConsentContext
+): boolean => {
+  if (value === undefined) {
+    return false;
+  }
+
+  if (typeof value === "boolean") {
```
LKCSmith, line 34: Is it possible for `value` to be a boolean and also `context.globalPrivacyControl === true`? I'm wondering if you set this condition to be read first purposefully. Reading the comment above, it seems like it's not something we need to account for, but curious.

LKCSmith (approved):

clients/privacy-center/packages/fides-consent/src/lib/consent-value.ts
```diff
@@ -0,0 +1,43 @@
...
+    return value;
+  }
+
+  if (context.globalPrivacyControl === true) {
```
saiatethyca, line 38: I'm confused about what this means.

My understanding of what it should be - If the GPC is enabled by the user and GPC is enabled for the consent option, then we set the value of the consent option to False (opted-out).
What's happening here -  If the GPC is enabled by the user, then we set the value of the consent option to if GPC applies to the consent or not. 

Example - So let's say the user has GPC as true (User has opted out) and in the config `consentValue.globalPrivacyControl ` is true. Then we set the consent value to be true (the user has opted in) but actually we should set it to false or opted-out.

My JS knowledge is a bit rusty please excuse me if what I'm saying doesn't make sense.

clients/privacy-center/config/config.json
saiatethyca: Does this configuration mean the default value for this consent option is true (Opted-in) and GPC signal doesn't apply to this consent option?

If GPC applies to this consent option, should the engineer configure this option as true?

clients/privacy-center/packages/fides-consent/src/lib/cookie.ts
ssangervasi: It's possible* to configure two consent options (data uses) to target the same key, where one defaults to `true` while the other defaults to `false`. This ensures that the more restrictive default is used so that we don't opt the user into the consent that should be opted out.

\* Possible, but probably a bad idea because it's hard to reason about.

clients/privacy-center/config/config.json
ssangervasi: https://github.com/ethyca/fides/pull/2341/files/05e70196abb4bcd153033d412646820d9a4d1922#r1084674181

> Does this configuration mean the default value for this consent option is true (Opted-in) and GPC signal doesn't apply to this consent option?

If GPC applies to this consent option, should the engineer configure this option as true?

There are two ways we could make GPC configurable:

**1: "Is this affected by GPC?"**

This sounds like what you were expecting the true/false to indicate. I think this is the intuitive question when looking for a quick-fix for respecting GPC, but it has some downsides:

1. It's hard to come up with a name that makes it clear how this config would work: `respectsGlobalPrivacyControl` or `affectedByGPC` ... Let me know if you have any ideas. Let's call it `affectedByGPC` for now. 
2. It is confusing setting `affectedByGPC: true` would mean _setting the default consent to `false`_. It is very easy to introduce confusing double-negatives in these features.
3. It seems likely that we will have _other_ settings that affect the default consent value. For example, location-based checks for EU, California, or Virginia. If we put `affectedByGPC` on the [ConfigConsentOption](https://github.com/ethyca/fides/blob/b4cadae4cb15794718508857ca8ee0bbbae5ec4f/clients/privacy-center/types/config.ts#L28), then adding another config that can mess with the `default` config is even more confusing. That's why my current implementation puts these conditions under the `default` config itself: this makes it clear (I hope) that the conditions only affect how the default consent is set.

**2: "If GPC is active, what should the default be?"**

This is how we should think of the GPC config question (IMO). This helps us address each of the downsides listed above:

1. The config called `globalPrivacyControl` refers to the situation when GPC is active. Maybe this could still benefit from a more detailed name.
2. Reading the config file now makes it clear that the _default_ value changes based on GPC:
  ```json
  "default": {
    "value": true,
    "globalPrivacyControl": false
  }
  ```
3. This also opens up other ways to changes the default consent based on other data-subject/browser information (what I'm calling "consent context"). For example:
  ```json
  "default": {
    "value": true,
    "globalPrivacyControl": false,
    "location": {
      "europe": false,
      "virginia": false,
      "california": true,
    }
  }
  ```

That's just hypothetical, but I think it makes clearer how the default consent is chosen based on the user's browser.

saiatethyca: Thanks @ssangervasi for the explanation. It's clear for me and I agree this is the best way to make it extensible for the future and manage code. 
Separately, we'll have to make this super clear in the documentation.

saiatethyca (approved):
Looks good to me, Thanks!

clients/privacy-center/packages/fides-consent/src/lib/consent-value.ts
ssangervasi: In addition to our other thread, I've updated the readme(s) to describe how this works:
https://github.com/ethyca/fides/blob/93098f0bf4162117a5842e8fdc300b970ffc4471/clients/privacy-center/packages/fides-consent/README.md#consent-context

clients/privacy-center/packages/fides-consent/src/lib/consent-value.ts
ssangervasi: If `value` is not `undefined`, then the user has made an explicit choice, which takes precedence over the "default" choice based on the browser context. https://github.com/ethyca/fides/issues/2230 will cover the UI indication that their explicit choice conflicts with the expected default.

PR #2337, opened 2023-01-23
Pass GA cookie from privacy center

allisonking:
Closes https://github.com/ethyca/fidesplus/issues/519

### Code Changes

* [x] Adds a function `inspectForBrowserIdentities` which looks through cookies on the page and pulls out relevant values
  * This is meant to be extended and potentially refactored once there are other services we'd like to integrate with besides GA. For now, it is hard coded to look for GA cookie values, but it could be extended to look through `config.json` for more cookie values/regexes
* [x] Adds jest test for this function
* [x] Calls this function before sending a consent request PATCH
* [x] Adds cypress test which tests that we are indeed sending the cookie via this PATCH

### Steps to Confirm

* [ ] _list any manual steps for reviewers to confirm the changes_

### Pre-Merge Checklist

* [ ] All CI Pipelines Succeeded
* Documentation:
  * [ ] documentation complete, [PR opened in fidesdocs](https://github.com/ethyca/fidesdocs/pulls)
  * [ ] documentation [issue created in fidesdocs](https://github.com/ethyca/fidesdocs/issues/new/choose)
* [ ] Issue Requirements are Met
* [ ] Relevant Follow-Up Issues Created
* [ ] Update `CHANGELOG.md`

### Description Of Changes

_Write some things here about the changes and any potential caveats_

PR #2304, opened 2023-01-20
Pass `browser_identity` on to privacy requests from consent request patches

allisonking:
Backend part of https://github.com/ethyca/fidesplus/issues/519

### Code Changes

* [x] Update the Identity model to support a new `user_id` 
* [x] Update the `providedidentitytype` enum to also support `user_id`
* [x] Update the patch endpoint to take a `browser_identity`
* [x] Update the `queue_privacy_request_to_propagate_consent` to merge identities 

### Steps to Confirm

* [ ] _list any manual steps for reviewers to confirm the changes_

### Pre-Merge Checklist

* [x] All CI Pipelines Succeeded
* Documentation:
  * [ ] documentation complete, [PR opened in fidesdocs](https://github.com/ethyca/fidesdocs/pulls)
  * [ ] documentation [issue created in fidesdocs](https://github.com/ethyca/fidesdocs/issues/new/choose)
* [ ] Issue Requirements are Met
* [ ] Relevant Follow-Up Issues Created
* [x] Update `CHANGELOG.md`

### Description Of Changes

_Write some things here about the changes and any potential caveats_

PR #2356, opened 2023-01-24
Rename `user_id` --> `ga_client_id`

allisonking:
Patches up https://github.com/ethyca/fides/pull/2304 after some more conversation with @pattisdr 

### Code Changes

* [x] Rename all references of `user_id` to `ga_client_id`. Since this was merged to a feature branch, should be safe to make these changes directly instead of making another alembic migration

### Steps to Confirm

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

Will also need to update https://github.com/ethyca/fides/pull/2337

### DIFF
--- a/src/fides/api/ops/models/privacy_request.py
+++ b/src/fides/api/ops/models/privacy_request.py
@@ -767,7 +767,7 @@ class ProvidedIdentityType(EnumType):
 
     email = "email"
     phone_number = "phone_number"
-    user_id = "user_id"
+    ga_client_id = "ga_client_id"
 
 
 class ProvidedIdentity(Base):  # pylint: disable=R0904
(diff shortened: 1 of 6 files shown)


pattisdr (approved):
this is great thank you

src/fides/api/ops/models/privacy_request.py
```diff
@@ -767,7 +767,7 @@ class ProvidedIdentityType(EnumType):
 
     email = "email"
     phone_number = "phone_number"
-    user_id = "user_id"
+    ga_client_id = "ga_client_id"
```
pattisdr, line 770: one more question, are we sure that it is the `client_id` and not the `app_instance_id`?

My options in the API request are:
Type of user. One of APP_INSTANCE_ID, CLIENT_ID or USER_ID.

allisonking: oh! I know that the cookie only has the `client_id`. I'm guessing the `app_instance_id` might be the GA value customers have to put in their javascript, but let me take a look at the docs and get back to you.

pattisdr: okay great 👍

allisonking: okay yes, looking at this, https://developers.google.com/analytics/devguides/config/userdeletion/v3, it _should_ be `client_id` we want here

pattisdr: excellent

Issue #2208, opened 2023-01-11
Admin should be able to reset password without the knowledge of old password


saiatethyca:
### Description
Currently, in the user management tab. Admin can change a user's password only if they are aware of the previous password. 
This is not a good experience for users as the trigger for a password change is generally when they forget their password. We need to support this use case.

### Use case
- User forgot their password, and they requested their admin to change it.

### Acceptance Criteria
- Admin can change the password without adding an older password.
- Admin can view the status of the change request. Whether it's successful or Failed.

### Mockups

Current
![Screen Shot 2023-01-11 at 5 36 09 PM](https://user-images.githubusercontent.com/117193073/211934434-76e3f94d-7218-4234-bf2a-ba4fdcfb458a.png)

Expected
![Screen Shot 2023-01-11 at 5 46 54 PM](https://user-images.githubusercontent.com/117193073/211934433-3ea65a98-01f0-4743-a07c-db8ba8b69534.png)

allisonking:
After some slack discussion, how about:

1. Remove the current code that checks for the scope `user:reset-password`. We will no longer use a scope to check for if a user can update their own password—they should always be allowed to
2. Create a migration that removes the `user:reset-password` scope from all users' scopes _except_ for the root user's
3. Add an endpoint which takes a `user_id` and a `password` which is scoped to `user:reset-password`. This is the endpoint a root user can hit if they need to reset someone else's password
4. Add a button the user's page that only renders if the user has the `user:reset-password` scope that says something like "Reset this user's password" which brings up a modal with two form fields: new password, and confirm new password. This will hit the endpoint in #3 

Does that sound ok to everyone? The only thing I worry about is if there's any danger in reusing `user:reset-password` like this (i.e. is it possible we could accidentally give permission to someone to reset other user's passwords who wouldn't have it?)

saiatethyca:
Thanks @allisonking, what you've outlined is clear to me!

One other concern I have is if we can restrict adding reset-password scope to a non-root user.

seanpreston:
> One other concern I have is if we can restrict adding reset-password scope to a non-root user.

I don't think this is necessary, since the scope has to be explicitly given.

daveqnet:
For my own reference, the specific use case in this issue is for an authenticated “fides admin” (although we don’t have roles yet) to reset a password for any other fides user.

[...]

Issue #2007, opened 2022-12-08
All users should have `user:reset-password` scope by default


seanpreston:

allisonking:
After the changes described in https://github.com/ethyca/fides/issues/2208 are done, this will effectively be true, where instead of all users needing this scope, there just won't be a scope necessary to change your own password

PR #2373, opened 2023-01-25
Add a scope to allow resetting other users' passwords

allisonking:
Closes https://github.com/ethyca/fides/issues/2208
Closes https://github.com/ethyca/fides/issues/2007

### Code Changes

* [x] Removes the old `user:reset-password` scope in favor of loosening up the endpoint to just always allow users to change their own passwords (as opposed to only allowing users to change their passwords in addition to needing this scope)
* [x] Adds an alembic migration to remove `user:reset-password` from every user
  * I'm not actually sure if this is the right place for it, as we can't really do a down revision for this. It's more a data migration than a db migration 
* [x] Adds a new scope `user:password-reset` (I know, the name is confusing if the old scope still exists!! I had some trouble with naming in this PR. `password-reset` / `reset-password` just seems like the right name for this action!) which allows the user with this scope to reset passwords without knowing the old one (including other users'!)
* [x] Adds a new endpoint `/user/{user_id}/force-reset-password` to allow resetting a password without knowing the old one which is scoped on `user:password-reset`
  * The [Forgot Password Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Forgot_Password_Cheat_Sheet.html) says to return a consistent message for both existent and non-existent accounts. I was going to do that for this endpoint, but thought it would cause confusion if an admin reset a user's password but the user no longer existed, so it should get a 404, but instead they get a success and then they try to tell that person to log in with the new password. Overall, I think this pattern of "admin making a password to reset a user's password" doesn't really follow best practices (as we've already discussed) and is temporary, but I could also make it always return a success instead if we want. Our other user/password related endpoints also do not give a consistent message, so maybe this is just something we need to do as a chunk once we have a proper password reset flow
* [x] Tests

### Steps to Confirm

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### Description Of Changes


src/fides/api/ctl/migrations/versions/7e218e880eaf_remove_user_reset_password_scope.py
```diff
@@ -0,0 +1,28 @@
...
+# revision identifiers, used by Alembic.
+revision = "7e218e880eaf"
+down_revision = "d6c6c6555c86"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.execute(
+        "UPDATE fidesuserpermissions SET scopes = array_remove(scopes, 'user:reset-password')"
```
seanpreston, line 20: @allisonking and I spoke on a call about when `array_remove(...)` was introduced to Postgres — it was introduced in version 11. Fides has only ever shipped with Postgres 12 and beyond so using this function shouldn't be an issue.

src/fides/api/ops/api/v1/endpoints/user_endpoints.py
```diff
@@ -113,6 +118,32 @@ def update_user_password(
...
+    status_code=HTTP_200_OK,
+)
+def update_other_user_password(
+    *,
+    db: Session = Depends(deps.get_db),
+    user_id: str,
+    data: UserForcePasswordReset,
+) -> FidesUser:
+    """
+    Update a user's password given a `user_id`.
```
seanpreston, line 133: Please can you add more context to the docstring here — it's important to mention why this exists separately to the endpoint above.

seanpreston (approved):
Thanks @allisonking — just one minor comment on the docstring :)

allisonking:
Going to merge, as the failing pylint test is unrelated to this PR

PR #2376, opened 2023-01-26
pc/api: Add common headers to base query

ssangervasi:
### Description Of Changes

The request done by the index page [wasn't](https://github.com/ethyca/fides/pull/2300/files#diff-8a8754ffd864b1b78f7740b3bab4da785a47767ec93102493ff4feef95283544L53) adding common headers, so when I used it as my first example for RTK queries I also left out that part. The most important header is probably the one that allows cross-origin requests, which probably too liberal but is necessary with Fides on a different domain.


allisonking (approved):
oooh good catch

Issue #1499, opened 2022-10-19
Field level overrides


Kelsey-Ethyca:
we can't mask a field by just replacing it with the work mask so we are adding a special function 

an example is domo erasure

pattisdr:
From Neville:

> One thing to add: in my opinion we'll need to allow "field-level" overrides as the escape hatch here. For example, if I know I have a field in my database that simply cannot be hashed, despite being a string, I could specify something inline like:

```
fides_meta:
  - data_type: string
    masking_strategy: null_rewrite
```

PR #2377, opened 2023-01-26
Masking Strategies: Handle Null Values and Non-Strings

pattisdr:
Closes: #2391

### Code Changes

* [ ] Hash, Hmac, and AES Encrypt masking strategies adjusted to better handle null values and non-strings (here we try to coerce to a string instead of failing)

### Steps to Confirm

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

Several of our masking strategies are set up to expect values to be strings.  There are some early exits if the input value is None, but since most inputs are now a list of values, even a list of one value, so that's not likely to be True.

If an array of values are passed in, if the original index is None, do nothing to that index.  Also try to coerce the original value to a string.

### DIFF
--- a/src/fides/api/ops/service/masking/strategy/masking_strategy_aes_encrypt.py
+++ b/src/fides/api/ops/service/masking/strategy/masking_strategy_aes_encrypt.py
@@ -40,6 +40,7 @@ class AesEncryptionMaskingStrategy(MaskingStrategy):
     ) -> Optional[List[str]]:
         if values is None:
             return None
+
         if self.mode == AesEncryptionMaskingConfiguration.Mode.GCM:
             masking_meta: Dict[
                 SecretType, MaskingSecretMeta
@@ -59,10 +60,14 @@ class AesEncryptionMaskingStrategy(MaskingStrategy):
 
             masked_values: List[str] = []
             for value in values:
+                if value is None:
+                    masked_values.append(None)
+                    continue
+
                 nonce: bytes | None = self._generate_nonce(
-                    value, key_hmac, request_id, masking_meta  # type: ignore
+                    str(value), key_hmac, request_id, masking_meta  # type: ignore
                 )
-                masked: str = encrypt(value, key, nonce)  # type: ignore
+                masked: str = encrypt(str(value), key, nonce)  # type: ignore
                 if self.format_preservation is not None:
                     formatter = FormatPreservation(self.format_preservation)
                     masked = formatter.format(masked)
--- a/src/fides/api/ops/service/masking/strategy/masking_strategy_aes_encrypt.py
+++ b/src/fides/api/ops/service/masking/strategy/masking_strategy_aes_encrypt.py
@@ -37,7 +37,7 @@ class AesEncryptionMaskingStrategy(MaskingStrategy):
 
     def mask(
         self, values: Optional[List[str]], request_id: Optional[str]
-    ) -> Optional[List[str]]:
+    ) -> Optional[List[Optional[str]]]:
         if values is None:
             return None
 
@@ -58,7 +58,7 @@ class AesEncryptionMaskingStrategy(MaskingStrategy):
             # and therefore the same masked val through the aes strategy. This is called convergent encryption, with this
             # implementation loosely based on https://www.vaultproject.io/docs/secrets/transit#convergent-encryption
 
-            masked_values: List[str] = []
+            masked_values: List[Optional[str]] = []
             for value in values:
                 if value is None:
                     masked_values.append(None)
--- a/src/fides/api/ops/service/masking/strategy/masking_strategy_hash.py
+++ b/src/fides/api/ops/service/masking/strategy/masking_strategy_hash.py
@@ -47,6 +47,7 @@ class HashMaskingStrategy(MaskingStrategy):
         is None"""
         if values is None:
             return None
+
         masking_meta: Dict[
             SecretType, MaskingSecretMeta
         ] = self._build_masking_secret_meta()
@@ -58,7 +59,11 @@ class HashMaskingStrategy(MaskingStrategy):
 
         masked_values: List[str] = []
         for value in values:
-            masked: str = self.algorithm_function(value, salt)  # type: ignore
+            if value is None:
+                masked_values.append(None)
+                continue
+
+            masked: str = self.algorithm_function(str(value), salt)  # type: ignore
             if self.format_preservation is not None:
                 formatter = FormatPreservation(self.format_preservation)
                 masked = formatter.format(masked)
--- a/src/fides/api/ops/service/masking/strategy/masking_strategy_hash.py
+++ b/src/fides/api/ops/service/masking/strategy/masking_strategy_hash.py
@@ -42,7 +42,7 @@ class HashMaskingStrategy(MaskingStrategy):
 
     def mask(
         self, values: Optional[List[str]], request_id: Optional[str]
-    ) -> Optional[List[str]]:
+    ) -> Optional[List[Optional[str]]]:
         """Returns the hashed version of the provided values. Returns None if the provided value
         is None"""
         if values is None:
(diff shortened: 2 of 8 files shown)


tests/ops/service/masking/strategy/test_masking_strategy_aes_encrypt.py
```diff
@@ -49,7 +51,7 @@ def test_mask_all_aes_modes(mock_encrypt: Mock):
 
 def cache_secrets() -> None:
     secret_key = MaskingSecretCache[bytes](
-        secret=b"\x94Y\xa8Z",
+        secret=b"y\xc5I\xd4\x92\xf6G\t\x80\xb1$\x06\x19t/\xc4",
```
pattisdr, line 54: Original AESGCM key in test needed to be 128, 192, or 256 bits - this came up when I tried to use it below intest_mask_arguments_date

src/fides/api/ops/service/masking/strategy/masking_strategy_hash.py
```diff
@@ -58,7 +59,11 @@ def mask(
 
         masked_values: List[str] = []
         for value in values:
-            masked: str = self.algorithm_function(value, salt)  # type: ignore
+            if value is None:
+                masked_values.append(None)
+                continue
+
+            masked: str = self.algorithm_function(str(value), salt)  # type: ignore
```
pattisdr, line 66: Seeking opinions regarding how masking strategies should behave when used in DSR's.  The hash masking strategy is intended to work on strings, but when applied in a DSR context, there are many values that could be potentially passed here that are not strings.

Should we fail or turn them into strings?  Turning them into a string may cause the masking database query to fail if the type can't be coerced.  But is that the responsibility of the masking strategy or the responsibility of the connector/query config?

(As an aside, these bring up good use cases for our detailed Policy > Rule > Rule Target framework.  The customer could configure so string fields are targeted by the hash strategy and non string fields are targeted by null masking for example)

pattisdr:
@adamsachs wanted to get your thoughts on the strategy here

src/fides/api/ops/service/masking/strategy/masking_strategy_hash.py
adamsachs: very good questions, my opinion is that we should try to make it work, for now? the only benefit i'd see for failing explicitly is to get a better error to the user, so i wonder if we can _try to make it work_ in a way that we're still going to get as clear error messages as possible to the user if our attempt fails because of a restriction on the DB side (which is admittedly likely in many of the cases, i'm sure).

i agree with your point that this brings up the need to allow for more fine-grained rule management in our framework. that being said, the current capability to stratify masking rules pretty finely on data categories may not be too bad, in that certain data categories should _generally_ correspond to certain data types. but that's definitely not going to be flawless, so i agree that the framework could use some improvement to give the user more control in this area.

adamsachs (approved):
this looks good to me @pattisdr! nice quick fix.

the `None` handling is exactly what i'd expect, and that's the most important piece.

the non-string-type coercion is a tricky one, and i agree with the approach of _trying_ -- essentially, why not? my only question is whether there's anything we can do to make a somewhat-expected type coercion error from the DB clearly communicated to the end user. maybe it's already going to be clear in how it's set up now, i'm not sure -- just wanted to raise point up to ensure it's considered.

src/fides/api/ops/service/masking/strategy/masking_strategy_hash.py
pattisdr: Thanks so much for your comments @adamsachs !

I think we generally try to make the query - there's something here https://github.com/ethyca/fides/blob/main/src/fides/api/ops/service/connectors/query_config.py#L162 , but that only gets hit in limited cases if I recall.

But using this change as an example: if there's a date field in mongo that we were trying to mask with the hash masking strategy, now the hash strategy will convert the date to a string, and hash it and try to write it back.  With mongo, I think this request succeeds but that wouldn't hold for every db. 

More work to do in this area for sure: looking into what kinds of errors we surface to the user, expanding type support beyond basic field types, adding increased type coercion (in a DSR, say you need to use a field of one type to locate data of another type), and more fine-grained rule management as discussed.

NevilleS: Totally agree with the decisioning here, thanks both!

One thing to add: in my opinion we'll need to allow "field-level" overrides as the escape hatch here. For example, if I know I have a field in my database that simply cannot be hashed, despite being a string, I could specify something inline like:
```
fides_meta:
  - data_type: string
    masking_strategy: null_rewrite
```

adamsachs: agreed @NevilleS - i was actually considering linking to this (vague) issue we'd made a while back here https://github.com/ethyca/fides/issues/1499 - which i think gets at what you're saying.

(the refactoring of the policy/rule framework to allow for more control feels at least like an epic-level initiative, and this issue would be one piece of it. i think we still would need to sketch out a bunch of the other work, i don't know if we've got that documented anywhere!)

src/fides/api/ops/service/masking/strategy/masking_strategy_aes_encrypt.py
```diff
@@ -59,10 +60,14 @@ def mask(
 
             masked_values: List[str] = []
             for value in values:
+                if value is None:
+                    masked_values.append(None)
+                    continue
+
                 nonce: bytes | None = self._generate_nonce(
-                    value, key_hmac, request_id, masking_meta  # type: ignore
+                    str(value), key_hmac, request_id, masking_meta  # type: ignore
```
seanpreston, line 68: If `value` is a member of `List[str]` why do we need this? Looks like sometimes it's explicitly passed as `None`?

I know we're handling it above explicitly but a common gotchya is that `str(None)` actually yields `"None"` FYI

src/fides/api/ops/service/masking/strategy/masking_strategy_aes_encrypt.py
```diff
@@ -40,6 +40,7 @@ def mask(
     ) -> Optional[List[str]]:
```
seanpreston, line 40: Should we change the `values:` typehint to `Optional[List[Optional[str]]]`?

src/fides/api/ops/service/masking/strategy/masking_strategy_aes_encrypt.py
pattisdr: Yes, correct, when executing a DSR, the occurs when if the masking strategy targets a non-string field.

Yep, I didn't want str(None) to be turned in a "None" so I handle that separately.

src/fides/api/ops/service/masking/strategy/masking_strategy_hash.py
pattisdr: @NevilleS Is this a feature that needs to added to this body of work?

src/fides/api/ops/service/masking/strategy/masking_strategy_aes_encrypt.py
pattisdr: good catch, yes, there can be Nones in this list

src/fides/api/ops/service/masking/strategy/masking_strategy_hash.py
seanpreston: > Should a DSR fail if a hash masking strategy was specified to be used on a date field for example, or should we try to make it work?

It's a good question! In this example specifically I think it's a config. overhead, given otherwise we'd have to examine the meaning of hashing to say "if we set a timestamp to 00000... does that count as hashed?"

pattisdr: Added Neville's comments to https://github.com/ethyca/fides/issues/1499 so field level overrides can be prioritized separately

NevilleS: Yes, absolutely - my comment was more pontificating about the larger feature we could add to really give ourselves a lot more configuration flexibility, not something that needed to be added to this PR. Sorry!

pattisdr: Oh no thanks for clarifying and for adding that suggestion! I think a field level override is going to be needed sooner rather than later

PR #2302, opened 2023-01-20
Google Analytics Consent Connector [Backend]

pattisdr:
Closes https://github.com/ethyca/fidesplus/issues/516

❗ Against `consent-propagation` feature branch
❗ To test end-to-end, needs https://github.com/ethyca/fidesplus/issues/519 which passes in a user id from the browser. Or you can fake it by hardcoding adding a `user_id` to the `identity` data when you queue the privacy request 

### Code Changes

* [ ] New Google Analytics SaaS Connector
* [ ] Tests are skipped in CI.  Can run locally by following the OAuth2.0 steps, and then adding the access_token directly to the google analytics secrets fixture.

### Steps to Confirm

* [ ] In your privacy center `config.json`, set `advertising.first_party` as executable=True
* [ ] Run `nox -s test_env`
* [ ] [Follow Internal docs to setup google analytics connector with OAuth2.0](https://ethyca.atlassian.net/wiki/spaces/EN/pages/2593587397/Setting+up+the+Google+Analytics+Consent+Connector) 
* [ ] Go to the Privacy Center and opt out of email marketing
* [ ] Hardcode `identity_data["user_id"]="test_fides_user_id"` in the early stages of [run_privacy_request](https://github.com/ethyca/fides/blob/main/src/fides/api/ops/service/privacy_request/request_runner_service.py#L336) until Allison's work gets in
* [ ] Return to the admin UI and approve the latest privacy request that was created to propagate consent preferences to third party systems
* [ ] Privacy request should complete quickly with a `completed` status
* [ ] Verify there are execution logs for this connector
* [ ] Check the docker logs in the fides container.  As we're in dev mode, you should see evidence of the request made to the Google Analytics

### Pre-Merge Checklist

* [ ] All CI Pipelines Succeeded
* Documentation:
  * [ ] documentation complete, [PR opened in fidesdocs](https://github.com/ethyca/fidesdocs/pulls)
  * [ ] documentation [issue created in fidesdocs](https://github.com/ethyca/fidesdocs/issues/new/choose)
* [ ] Issue Requirements are Met
* [ ] Relevant Follow-Up Issues Created
* [ ] Update `CHANGELOG.md`

### Description Of Changes

Adds the SaaS Config for the Google Analytics Consent Connector.  

Key Items here:
- Currently set up to work with GA4 not UA
- Currently requires a `user_id` to be in the identity map
- Uses OAuth 2.0 for authentication - this can take a little while to set up with fides
[...]

PR #2384, opened 2023-01-26
Skip Missing Param Values

### LINKED ISSUE #2375: [Consent propagation] Option to skip requests without param values
### Description

We have a new type of identity data: the `ga_client_id`.  On the `consent-propagation` branch, we retrieve this google analytics client id from the browser when the user saves their consent preferences.   This client id doesn't always exist.

Later, when a Privacy request is queued to propagate these consent preferences, the Google Analytics Connector will try to build a request using the client id where applicable and fail if it doesn't exist, which will also stop execution of the entire privacy request.

We need a way to indicate in the SaaS config that some requests should be skipped instead of failing if we don't have all the identity data.

### Acceptance Criteria

- Add an option to the SaaS Config to be able to skip sending a request if all the placeholders can't be populated for a given request.  (This is similar to the `ignore_errors` flag).
- Update the Google Analytics opt out SaaSRequest with this flag
- Update `SaaSConnector.run_consent_request` to skip building a request if all param values can't be populated, instead of failing.

pattisdr:
Closes #2375 

### Code Changes

* [ ] Add a new `SaaSRequest.skip_missing_param_values` attribute with a default value of False 
* [ ] Set the google analytics opt out request to have skip_missing_param_values=true
* [ ] Update SaaSConnector.run_consent_request to skip sending a request if we can't replace all the placeholder values in the request rather than fail, provided skip_missing_param_values=true

### Steps to Confirm

* [ ] Update privacy center test env config.json to have advertising.first_party be executable=true
* [ ] Run `nox -s test_env`
* [ ] Set up the GA consent connector locally
* [ ] Go to the privacy center, delete your ga cookie `_ga="GA1.1.XXXXXXXXX.XXXXXXXXX"`from the browser
* [ ]  In the privacy center, update your consent preferences: Specifically opt out of Email Marketing
* [ ] Go to admin app, approve the privacy request
* [ ] Verify privacy request completes instead of errors
* [ ] Verify in docker container that no request was made to GA, instead you see a log: `Skipping optional consent request on node google_analytics:google_analytics: Unable to replace placeholders in body for the 'opt_out' request of google_analytics`

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

Add a new SaasRequest.skip_missing_param_values to indicate whether we should skip an http request when we can't map all the param values in the request body rather than fail.  

For example, we pull the google analytics client id from the browser for the google analytics consent connector but it isn't always guaranteed to exist.  Rather than having that request fail, we should just skip it.

Functionality added for retrieve_data, mask_data, and run_consent_request, although this is unlikely to be needed for read requests.  (It affects when we can't place param values in the body only)

### DIFF
--- a/data/saas/config/google_analytics_config.yml
+++ b/data/saas/config/google_analytics_config.yml
@@ -92,6 +92,7 @@ saas_config:
     opt_out:
       method: POST
       path: /analytics/v3/userDeletion/userDeletionRequests:upsert
+      skip_missing_param_values: true
       param_values:
         - name: ga_client_id
           identity: ga_client_id
--- a/src/fides/api/ops/schemas/saas/saas_config.py
+++ b/src/fides/api/ops/schemas/saas/saas_config.py
@@ -104,7 +104,9 @@ class SaaSRequest(BaseModel):
     grouped_inputs: Optional[List[str]] = []
     ignore_errors: Optional[bool] = False
     rate_limit_config: Optional[RateLimitConfig]
-    skip_missing_param_values: Optional[bool] = False
+    skip_missing_param_values: Optional[
+        bool
+    ] = False  # Skip instead of raising an exception if placeholders can't be populated in body
 
     class Config:
         """Populate models with the raw value of enum fields, rather than the enum itself"""
--- a/src/fides/api/ops/schemas/saas/saas_config.py
+++ b/src/fides/api/ops/schemas/saas/saas_config.py
@@ -104,6 +104,7 @@ class SaaSRequest(BaseModel):
     grouped_inputs: Optional[List[str]] = []
     ignore_errors: Optional[bool] = False
     rate_limit_config: Optional[RateLimitConfig]
+    skip_missing_param_values: Optional[bool] = False
 
     class Config:
         """Populate models with the raw value of enum fields, rather than the enum itself"""
--- a/src/fides/api/ops/service/connectors/saas_connector.py
+++ b/src/fides/api/ops/service/connectors/saas_connector.py
@@ -399,15 +399,25 @@ class SaaSConnector(BaseConnector[AuthenticatedClient]):
             cast(Optional[List[PostProcessorStrategy]], masking_request.postprocessors),
         )
 
-        prepared_requests = [
-            query_config.generate_update_stmt(row, policy, privacy_request)
-            for row in rows
-        ]
         rows_updated = 0
         client = self.create_client()
-        for prepared_request in prepared_requests:
+        for row in rows:
+            try:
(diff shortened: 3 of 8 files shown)


data/saas/config/google_analytics_config.yml
```diff
@@ -92,6 +92,7 @@ saas_config:
     opt_out:
       method: POST
       path: /analytics/v3/userDeletion/userDeletionRequests:upsert
+      skip_missing_param_values: true
```
pattisdr, line 95: This is a patch for the google analytics connector you reviewed previous @adamsachs - wanted to get your thoughts on this approach. Also open to better names!

src/fides/api/ops/schemas/saas/saas_config.py
```diff
@@ -104,6 +104,7 @@ class SaaSRequest(BaseModel):
     grouped_inputs: Optional[List[str]] = []
     ignore_errors: Optional[bool] = False
     rate_limit_config: Optional[RateLimitConfig]
+    skip_missing_param_values: Optional[bool] = False
```
pattisdr, line 107: I was toggling between having this param be more specifically related to missing identity data, but thought we might make this more broad to indicate any param values were missing.

pattisdr:
I need to retrigger tests too

adamsachs:
@pattisdr this looks good to me generally, i like the feature! only question is - since you've put the property on the `SaaSRequest` model generally (which makes sense), should we not also respect it during access/erasure request execution?

pattisdr:
@adamsachs thanks so much for taking a look. I've had blinders on trying to get the `consent-propagation` branch ready to merge.  But in the spirit of trying to keep [access, erasure, and consent aligned](https://github.com/ethyca/fides/pull/2302#discussion_r1084538844) we could definitely look into extending this to access/erasure or creating a follow-up ticket at the very least.

pattisdr:
@adamsachs I've added a commit to add similar behavior to retrieve_data and mask_data - to have a configurable option to skip the request instead of fail if we can't populate placeholders in the body.

Read requests would be unlikely to hit this error, as this is specifically for **missing placeholders in the body**.  I've added some contrived tests too for read/update requests as we don't have examples yet. 

Let me know what you think!

pattisdr:
Realize tests are failing here, but they seem unrelated.  My goal would be to merge this PR into `consent-propgation`, then get `consent-propagation` up-to-date with `main` and I'd re-run tests there.

adamsachs (approved):
@pattisdr what you added looks great - thanks for addressing that! from my POV, this is good to go 👍

PR #2388, opened 2023-01-26
Default PII to hidden on page load

### LINKED ISSUE #2286: PII toggle state between request detail page and privacy request manager page are not maintained
### Bug Description

The PII toggle state between the request detail page and the privacy request manager page is not maintained as expected.

#### Steps to Reproduce

1. Go to privacy requests page and click to reveal PII. 
2. Now go to a detail page for any privacy request. PII toggle is disabled but the PII is visible.
3. Click to enable and go back to the privacy requests page. PII is visible for all the requests and the toggle is off.

### Expected behavior
- PII toggle for the privacy requests page and request detail page should be independent and the default state should have no PII visible.

### Screenshots
Screen Shot 2023-01-19 at 10 52 16 AM

Screen Shot 2023-01-19 at 10 52 06 AM

TheAndrewJackson:
Closes #2286 

### Code Changes
* [ ] Default PII to hidden on pageload

### Steps to Confirm

* [ ] Run fides and create a privacy request. 
* [ ] Toggle the PII switch to reveal PII and navigate to the details page
* [ ] confirm that the switch is set to hide PII and the PII is hidden
* [ ] Toggle PII on and navigate back to the privacy requests table and confirm that revealing PII is turned off and PII is hidden

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### Description Of Changes

The PII state wasn't being set properly for the PII toggle. Now it will correctly set it to false on page load.

https://user-images.githubusercontent.com/17103888/214919829-497f66d8-4a37-4e15-8c9d-e3d88a673f9b.mov


TheAndrewJackson:
Hold please. I may have done the inverse of the ticket. Double checking

clients/admin-ui/src/features/common/PIIToggle.tsx
```diff
@@ -1,12 +1,21 @@
...
   const dispatch = useDispatch();
   const handleToggle = (event: ChangeEvent<HTMLInputElement>) =>
     dispatch(setRevealPII(event.target.checked));
+
+  useEffect(() => {
+    /*
+      PII should default to hidden on page load.
+      This ensures the state isn't reused as the user navigates around
+   */
+    dispatch(setRevealPII(false));
```
allisonking, line 16: if the PII toggle is indeed supposed to be independent from one another, perhaps we don't want to store this state in redux at all, and just use a regular ol' `useState` here?

reading the ticket, it seems like there are two issues:
1. This toggle wasn't showing on/off properly based off of the PII state. this is definitely a bug which could be fixed by setting an initial state based on a selector `selectRevealPII` from redux, though your change here also works because it just always sets it to false on load, ignoring whatever state there initially was
2. The toggle on the list page and the details page should be independent. It seems like this PII toggle was initially set up so that revealing PII would _not_ be independent, and that if you revealed once, then it's revealed everywhere (until you toggle again). This is def a product decision, and it seems like the decision was to make these totally independent, so if you want to unreveal PII you have to do it everywhere, but the code doesn't _quite_ reflect that. Again, your change definitely works since it always sets it to not reveal on load, but it makes things a little weird.

For example, we won't run into this exact use case since it'd be weird for two toggles to be on the same page, but if we had a page that had

```typescript
<PIIToggle id="one" />
<PIIToggle id="two" />

<PII data="name" />
```

And the user clicked `one`, then `<PII data="name" />` would reveal itself since it changes the global PII redux state. But `two` would stay untoggled since they're not linked. To me, the data model makes more sense as

```typescript
const [revealPII, setRevealPII] = useState(false)
<PIIToggle id="one" reveal={revealPII} onChange={setRevealPII} />
<PIIToggle id="two" reveal={revealPII} onChange={setRevealPII} />

<PII data="name" reveal={revealPII} />
```

In this way, it's clearer that it's not a global state that is being changed, but just the page's state, and all the components will respect that.

That involves changing the way `<PII />` works too though and also taking this state out of redux which is a bigger change. And the example I gave above is pretty contrived, so we likely won't run into it, but I can see it getting us in the future.

As an aside, are we sure that we actually want that second bullet point I listed above? I can see it getting annoying if people want to just keep PII revealed but they have to keep setting it. I know it's a product q, but if the answer is "no let's keep them linked", then we don't have any of the problems I just mentioned, and can just change the `<Switch />` component to properly take an initial state :)

allisonking (commented):
this definitely works! I left a comment about a potential drawback, but if we don't want to deal with that now, we don't have to. mostly just wanted to raise it now so that we're aware of it

clients/admin-ui/src/features/common/PIIToggle.tsx
TheAndrewJackson: I completely agree that this is kind of jank solution and it probably shouldn't be in the redux state. Based off what the ticket is saying it seems like they're intended to be independent on each page. 

I did it this way because it was really quick but your proposed solution makes sense. I'll take a look real quick and see how much of a lift it would be to update it to work that way. If it's not too bad I'll update it to work that way in this PR.


Also, I think there may be another hidden bug included in #1. The persisted redux store that gets reloaded may have been setting the state causing the PII to be shown when the user visits the page again in the future. 

The current solution does fix that issue but switching over to a `useState` makes a lot more sense

clients/admin-ui/src/features/privacy-requests/PrivacyRequestsContainer.tsx
```diff
@@ -14,6 +14,7 @@ const ActionButtons = dynamic(
 
 const PrivacyRequestsContainer: React.FC = () => {
   const { processing } = useDSRErrorAlert();
+  const [revealPII, setRevealPII] = useState(false);
```
TheAndrewJackson, line 17: @allisonking It wasn't too bad to convert it over to a `useState` so I went ahead and did it. There is a little prop drilling required for the table page due to the PII components being split between multiple pages. The details page was simple because both PII components are located in the same component.

PR #2402, opened 2023-01-27
Switch to codecov

sanders41:
Closes #<issue>

### Code Changes

* Switch code climate to codecov

### Steps to Confirm

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes


codecov.yaml
```diff
@@ -0,0 +1,10 @@
+coverage:
+  status:
+    project:
+      default:
+        target: 79%
```
sanders41, line 5: I set the minimum coverage at 79% to make sure we don't drop below our current number for now. As we move to our target of 85% we can increase this.

codecov.yaml
```diff
@@ -0,0 +1,10 @@
...
+    patch:
+      default:
+        target: 100%
+notify:
+  after_n_builds: 12
```
sanders41, line 10: This has codecov wait until it receives 12 report files before reporting the coverage back. If I counted correctly this should be the total number of reports that get uploaded.

Without this everything works, but can be confusing because coverage is sent back with each file received. This results in low, failing numbers slowly increasing until all files are received.

sanders41: Decided to change this to 4. There are 4 tests run for 3 different python version resulting in 12 total. So I think we can report at 4 and have a full picture. If this turns out to be wrong we can adjust.

codecov.yaml
sanders41: After getting the migrations excluded from the coverage count (since we can't test these) our coverage is just under 89% so I went ahead and bumped this up to 85%

SteveDMurphy (commented):
It looks like there may be some remaining code climate items in `constants.py` lines 41-43,  `ci_nox.py` lines 187-189, and all of `fix_coverage.py`

SteveDMurphy (approved):
Left a couple of comments around some lingering code climate items but this looks great! 
Excited to see how it looks in PRs once we get some other things merged to `main`. Last thing might just be to add something to `changelog.md` as well (since we mention code climate in there too for the previous PR)

sanders41:
Thanks @SteveDMurphy for finding the extra code coverage spots I missed! 🔍

PR #2417, opened 2023-01-27
Cypress cloud setup

allisonking:
Closes https://github.com/ethyca/fides/issues/1293

### Code Changes

* [x] New nox command `nox -s e2e_test` which spins up the test env then runs the cypress tests against it
* [x] New GHA workflow which calls this command. It's working! 🎉 https://github.com/ethyca/fides/actions/runs/4048522137/jobs/6963867568 Not sure if we want it running on all PR's though. It's a little slow because of setting up the test env, but our other tests do that too 
* [x] Record cypress runs to Cypress cloud

### Steps to Confirm

* [ ] `nox -s e2e_test` works locally
* [ ] CI for `cypress_e2e` workflow passes
* [ ] Uploads to cypress cloud!

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### Description Of Changes


allisonking:
Some notes from this afternoon's tinkering:

1. The GHA I put together doesn't work because `nox -s test_env` exits before the cypress tests can run. Need to look into if I can make it run in the background while cypress does its thing
2. Looked into a `nox` command that'll both start `nox -s test_env` and run the cypress e2e test. Following the pattern of our other nox commands, I think we'd want to run cypress from within a docker container as well. This can be done with the official cypress image via

```sh
docker run -it -v $PWD:/e2e -w /e2e -e CYPRESS_API_URL=http://docker.for.mac.localhost:8080/api/v1 -e CYPRESS_ADMIN_UI_URL=http://docker.for.mac.localhost:3000/ -e CYPRESS_PRIVACY_CENTER_URL=http://docker.for.mac.localhost:3001/ cypress/included:12.3.0
```
However this causes CORS problems since the url's here use `docker.for.mac.localhost` instead of regular ol' `localhost` (this might not be an issue on linux). Not sure how to work around this yet

ThomasLaPiana:
> Some notes from this afternoon's tinkering:
> 
> 1. The GHA I put together doesn't work because `nox -s test_env` exits before the cypress tests can run. Need to look into if I can make it run in the background while cypress does its thing
> 2. Looked into a `nox` command that'll both start `nox -s test_env` and run the cypress e2e test. Following the pattern of our other nox commands, I think we'd want to run cypress from within a docker container as well. This can be done with the official cypress image via
> 
> ```shell
> docker run -it -v $PWD:/e2e -w /e2e -e CYPRESS_API_URL=http://docker.for.mac.localhost:8080/api/v1 -e CYPRESS_ADMIN_UI_URL=http://docker.for.mac.localhost:3000/ -e CYPRESS_PRIVACY_CENTER_URL=http://docker.for.mac.localhost:3001/ cypress/included:12.3.0
> ```
> 
> However this causes CORS problems since the url's here use `docker.for.mac.localhost` instead of regular ol' `localhost` (this might not be an issue on linux). Not sure how to work around this yet

the host issue is definitely a symptom of docker on Mac. As for how to fix it, I think running them with a shared network would fix it? Basically doing explicitly what something like docker compose does implicitly. Creating a docker network and attaching both containers to it

allisonking:
> the host issue is definitely a symptom of docker on Mac. As for how to fix it, I think running them with a shared network would fix it? Basically doing explicitly what something like docker compose does implicitly. Creating a docker network and attaching both containers to it

ah yes, passing a `--network=host` works 💥 thanks for the tip!

noxfiles/dev_nox.py
```diff
@@ -72,6 +73,17 @@ def dev(session: Session) -> None:
         )
 
 
+@nox_session()
+def e2e_test(session: Session) -> None:
+    """
+    Spins up the test_env session and runs Cypress E2E tests against it.
+    """
+    session.log("Setting up the test environment...")
+    test_env(session)
```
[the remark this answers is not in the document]
ThomasLaPiana: you're well on your way to becoming a nox legend @allisonking! I'll take a peek

clients/cypress-e2e/cypress.config.ts
```diff
@@ -1,6 +1,7 @@
 import { defineConfig } from "cypress";
 
 export default defineConfig({
+  projectId: "bauzeh",
```
ThomasLaPiana, line 4: "bowser" said with an accent?

.github/workflows/cypress_e2e.yml
```diff
@@ -0,0 +1,35 @@
+name: Cypress E2E Tests
+
+on:
+  push:
+    paths-ignore:
+      - "**.md"
+
+env:
+  CI: true
```
ThomasLaPiana, line 9: FYI I think this is already set by default: https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables

clients/cypress-e2e/cypress.config.ts
allisonking: 😈

allisonking:
[This commit](https://github.com/ethyca/fides/pull/2417/commits/9ff5de221eb6923d0aaa40e2f4ba4465a22c71fc) gets things working to run in CI via `nox -s e2e_test` but the output isn't great for Cypress Cloud, I'm guessing because it's running in a docker container so it doesn't have access to things like the last git commit. For example

image

I think using the cypress github action would let us get around that, but I haven't been able to get it to play well with `nox -s fides_env(test)` yet. [This](https://github.com/ethyca/fides/pull/2417/commits/0a1fb2109fb2f7dde2b650dc01e144ff9866ef47) should be pretty close but it appears that the server isn't fully coming up and I'm not getting logs once I pass the `&` in 😵

allisonking:
okay it seems like the server might just not be coming up at all and then meets the cypress time out which I set to 20 minutes (the other tests seemed to spin up `fides_env(test)` within 15 min, so it _should_ be enough time). I'm not sure how to debug though...

any chance you have any ideas @ThomasLaPiana ?

ThomasLaPiana:
> okay it seems like the server might just not be coming up at all and then meets the cypress time out which I set to 20 minutes (the other tests seemed to spin up `fides_env(test)` within 15 min, so it _should_ be enough time). I'm not sure how to debug though...
> 
> any chance you have any ideas @ThomasLaPiana ?

have we confirmed that `fides_env(test)` does spin up the environment and leaves it up as expected locally?

allisonking:
> The GHA I put together doesn't work because nox -s test_env exits before the cypress tests can run. Need to look into if I can make it run in the background while cypress does its thing

okay all of my nox knowledge (noxledge?) is coming together and I realized [this line](https://github.com/ethyca/fides/blob/main/noxfiles/dev_nox.py#L113) was causing my problems, where issuing the `notify` to teardown queues it up at the end, and so whenever the process exits, it'll teardown.

I added a new posarg `keep_alive` which won't tear it down, and that lets things work in CI as we want! 🎉 

https://cloud.cypress.io/projects/bauzeh/runs/9/overview

image

And we get some good metadata 💥 
image

ThomasLaPiana:
@allisonking the amount of noxledge on dislpay here is impressive 😄 this looks awesome!

ThomasLaPiana:
The code all looks good to me, but the Cypress check didn't run on the latest PR 🤔 it did on the one before, and there weren't any changes except for the Changelog

ThomasLaPiana (approved):
LGTM! Great work, just need to confirm why the job didn't run before merging

allisonking:
> The code all looks good to me, but the Cypress check didn't run on the latest PR 🤔 it did on the one before, and there weren't any changes except for the Changelog

Ah good catch! the `on:` for this workflow was a little different than the other ones. I changed it to match 👍

PR #2487, opened 2023-02-02
Prevent duplicate usernames

### LINKED ISSUE #2371: Usernames should not be case-sensitive
### Bug Description

Users can be created with the same username but minor differences in uppercase/lowercase. For example, `jane` and `Jane` can be created as separate users. This is confusing and leads to lots of unexpected username issues!

#### Steps to Reproduce

1. Go to user management and create a user with username "sai"
2. Create another user with username "Sai" (S) capital
3. Both users are created.

### Expected behavior

* Username should be treated as case-insensitive value for creating users (e.g. creating `sai`, `SAI`, or `Sai` are all the same username)
* Username should be treated as case-insensitive when logging in; e.g. it should be possible to login as `sai`, `SAI`, `Sai`, etc.

### Implementation Notes

When implementing this, we should enforce this in the database with case-insensitivity. To do this migration we'll have to do an automated migration and delete duplicate users. When there are duplicates, we should only keep the oldest user (e.g. the earliest creation date) instead of the others.

TheAndrewJackson:
Closes #2371 

> **Warning**
> This PR contains a migration that automatically deletes duplicate users from the DB. It keeps the first record that was created for a given username.

### Code Changes

* [ ] Add migration that deletes duplicate users
* [ ] Update user creation endpoint to correctly prevent duplicate usernames
* [ ] Add test that validates that duplicate usernames can't be used.
* [ ] Update login endpoint to be case insensitive and a test to validate that

### Steps to Confirm

Testing this is a little convoluted because fides no longer allows admins to created duplicate users. You need to revert it a little bit to get fides into a bad state. The previous bugged query needs to be restored. 

Revert
```python
user = (
    FidesUser.query(db)
    .filter(func.lower(FidesUser.username) == user_data.username.lower())
    .first()
)
```

to 
```python
user = FidesUser.get_by(db, field="username", value=user_data.username)
```
Then after that go into the `docker-compose.yml` and update the `command` to this
```
command: uvicorn --host 0.0.0.0 --port 8080 fides.api.main:app
```
This makes it so the server doesn't hot reload anymore. You don't want that on because if you make any code changes as you're creating duplicate users the migration may run and and delete your test duplicate users 😬 .

After that you should be ready to test
  

* [ ] Run fides `nox -s dev`
* [ ] Run the admin ui 
* [ ] Exec into the container `docker exec -it fides-fides-1 /bin/bash` and run `cd src/fides/api/ctl/`
* [ ] Run `alembic downgrade 392992c7733a` to go to previous migration
* [ ] Go to the user management page and create duplicate users (ex: same username but with different casing. `johnsmith`-> `JOHNSMITH`)
  * [ ] Mark the first one as the original in the first name column
  * [ ] Mark the duplicates as duplicate
* [ ] In the container run `alembic upgrade head`. This deletes the duplicates
* [ ] Go back to the management page and verify that the duplicate users are gone
* [ ] Revert the query back the working duplicate checking query and restart fides
* [ ] Try to create duplicate users
* [ ] Then logout from the admin role and try logging into one of the users you created and change the casing to something different from the original

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### Description Of Changes

Removing duplicates with the migration

https://user-images.githubusercontent.com/17103888/216435270-64c4eeab-b1ad-4bfe-8fea-62e07b6b00e2.mp4

Preventing duplicate usernames

https://user-images.githubusercontent.com/17103888/216429193-a1910a0d-ff1c-4506-ae0e-fad6cb31783b.mov

Case insensitive login

https://user-images.githubusercontent.com/17103888/216621423-9ae12f6a-2c74-4d6f-9763-6e4902c561bf.mov


src/fides/api/ctl/migrations/versions/643249f65453_case_insensitive_usernames_delete_.py
```diff
@@ -0,0 +1,48 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+
+    # Found here in the 24.2.2.4. Nondeterministic Collations section:
+    # https://www.postgresql.org/docs/current/collation.html
+    op.execute(
+        "CREATE COLLATION IF NOT EXISTS case_insensitive (provider = icu, locale = 'und-u-ks-level2', deterministic = false);"
```
TheAndrewJackson, line 27: This collation is needed it order to make the `DISTINCT ON` function work. Without it the DB will think every username is unique.

src/fides/api/ctl/migrations/versions/643249f65453_case_insensitive_usernames_delete_.py
```diff
@@ -0,0 +1,48 @@
...
+    )
+    session = Session(bind=op.get_bind())
+    conn = session.connection()
+    statement = text(
+        """SELECT DISTINCT ON (username) username collate case_insensitive, created_at, id
```
TheAndrewJackson, line 32: I ran this query using raw sql because I was having a pretty hard time getting the ORM to generate it correctly. It was always pretty close but never worked quite right.

src/fides/api/ctl/migrations/versions/643249f65453_case_insensitive_usernames_delete_.py
```diff
@@ -0,0 +1,48 @@
...
+    )
+
+    result = conn.execute(statement)
+    original_user_ids = [row[2] for row in result]
+    session.query(FidesUserPermissions).filter(
+        FidesUserPermissions.user_id.not_in(original_user_ids)
+    ).delete()
+    session.query(FidesUser).filter(FidesUser.id.not_in(original_user_ids)).delete()
+
+    op.execute("DROP COLLATION IF EXISTS case_insensitive")
```
TheAndrewJackson, line 44: The collation is removed because it's only needed for one query.

CHANGELOG.md
```diff
@@ -75,6 +75,7 @@ The types of changes are:
   * Phone inputs no longer request country SVGs from external domain. [#2378](https://github.com/ethyca/fides/pull/2378)
   * Input validation errors no longer change the height of modals. [#2379](https://github.com/ethyca/fides/pull/2379)
 * Patch masking strategies to better handle null and non-string inputs [#2307](https://github.com/ethyca/fides/pull/2377)
+* Usernames are case insensitive now and prevent all duplicates [#2487](https://github.com/ethyca/fides/pull/2487)
```
ThomasLaPiana, line 78: maybe note here that duplicated will be deleted?

TheAndrewJackson: That should be called out. Good catch

pattisdr:
Starting review, Andrew!

src/fides/api/ctl/migrations/versions/643249f65453_case_insensitive_usernames_delete_.py
```diff
@@ -0,0 +1,48 @@
...
+        """SELECT DISTINCT ON (username) username collate case_insensitive, created_at, id
+FROM fidesuser
+ORDER BY username, created_at ASC"""
+    )
+
+    result = conn.execute(statement)
+    original_user_ids = [row[2] for row in result]
+    session.query(FidesUserPermissions).filter(
+        FidesUserPermissions.user_id.not_in(original_user_ids)
+    ).delete()
```
pattisdr, line 41: Also look into deleting associated user_ids from the `client` table - that will prevent a migration.  But I wouldn't delete the `client` record itself, since several other resources reference it.

src/fides/api/ctl/migrations/versions/643249f65453_case_insensitive_usernames_delete_.py
pattisdr: You probably don't need a collation - something like upper should work?

```psql
SELECT DISTINCT ON (UPPER(username)) username, created_at, id
FROM fidesuser
ORDER BY UPPER(username), created_at ASC
```

src/fides/lib/oauth/api/routes/user_endpoints.py
```diff
@@ -191,8 +196,10 @@ def user_login(
             created_at=datetime.utcnow(),
         )
     else:
-        user_check: Optional[FidesUser] = FidesUser.get_by(
-            db, field="username", value=user_data.username
+        user_check = (
+            FidesUser.query(db)
+            .filter(func.lower(FidesUser.username) == user_data.username.lower())
+            .first()
```
pattisdr, line 202: Looks like we could DRY this up and share it with L75!

src/fides/lib/oauth/api/routes/user_endpoints.py
```diff
@@ -71,7 +72,11 @@ def create_user(
             status_code=HTTP_400_BAD_REQUEST, detail="Username already exists."
         )
 
-    user = FidesUser.get_by(db, field="username", value=user_data.username)
+    user = (
+        FidesUser.query(db)
+        .filter(func.lower(FidesUser.username) == user_data.username.lower())
+        .first()
+    )
```
pattisdr, line 79: from the ticket:
> When implementing this, we should enforce this in the database with case-insensitivity.

These changes are good at preventing duplicate users at the code level but there's nothing at the database level.

Some ideas [here](http://shuber.io/case-insensitive-unique-constraints-in-postgres/) include changing the type of Fidesuser.username to a `citext` type or adding a unique index instead of a constraint.  I'd look more into the pros and cons of these approaches.

pattisdr (changes requested):
Thanks for taking the time to type out those detailed testing instructions.  Back to you!

src/fides/api/ctl/migrations/versions/643249f65453_case_insensitive_usernames_delete_.py
ThomasLaPiana: ^^^ Have experienced this before! Especially when Fideslang changes

src/fides/api/ctl/migrations/versions/643249f65453_case_insensitive_usernames_delete_.py
TheAndrewJackson: Interesting. I'll give that a try and report back

src/fides/api/ctl/migrations/versions/643249f65453_case_insensitive_usernames_delete_.py
TheAndrewJackson: That makes a lot sense. I'll keep using raw sql in the future

src/fides/api/ctl/migrations/versions/643249f65453_case_insensitive_usernames_delete_.py
TheAndrewJackson: @pattisdr I deleted those records because I was running into foreign key constraint errors when deleting the users. It turns out that the `user_id` column on `FidesUserPermissions` can't be null

https://github.com/ethyca/fides/blob/56b888e39e0fd013860aec0bb492ccfb994aac28/src/fides/lib/models/fides_user_permissions.py#L15

We would need to update the model and maybe a migration to allow that field to be nullable.

Does it make sense to delete the permission records with that in mind?

src/fides/lib/oauth/api/routes/user_endpoints.py
TheAndrewJackson: I converted it to use citext instead. Thanks for the suggestion!

src/fides/lib/oauth/api/routes/user_endpoints.py
TheAndrewJackson: Now that citext is being used I reverted this query to the original.

requirements.txt
```diff
@@ -39,6 +39,7 @@ sendgrid==6.9.7
 slowapi==0.1.5
 snowflake-sqlalchemy==1.4.3
 sqlalchemy[asyncio]==1.4.27
+sqlalchemy-citext==1.8.0
```
TheAndrewJackson, line 42: This packages adds a `citext` type that sqlalchemy recognizes.  It's a simple package with 1-2 classes and is MIT licensed.

https://github.com/mahmoudimus/sqlalchemy-citext

src/fides/api/ctl/migrations/versions/643249f65453_case_insensitive_usernames_delete_.py
```diff
@@ -0,0 +1,49 @@
...
+    original_user_ids = [row[2] for row in result]
+    op.execute(
+        f"""DELETE FROM fidesuserpermissions
+    WHERE NOT(user_id=ANY(ARRAY[{original_user_ids}]))"""
+    )
+
+    op.execute(
+        f"""DELETE FROM fidesuser 
+    WHERE NOT(id=ANY(ARRAY[{original_user_ids}]))"""
+    )
```
TheAndrewJackson, line 42: @pattisdr I converted all of the queries to use raw sql. Now this whole migration should capture this state in time 👍

TheAndrewJackson:
It looks like the tests are having an issue with either the `citext` extension or the 3rd party python package. Looking into it now

src/fides/api/ctl/migrations/versions/643249f65453_case_insensitive_usernames_delete_.py
pattisdr: I am talking about a separate model Andrew, I'm fine with what you did with the FidesUserPermissions records, that makes sense.

I was talking about the separate `Client` model that has a FK to `fidesuser`  I think you need to remove associated user_ids. That column is already nullable, so I'd just remove the data.

pattisdr: oh and it'd also be good to double check there are no other tables referencing user_id

tests/conftest.py
```diff
@@ -16,3 +17,8 @@ def loguru_caplog(caplog):
     handler_id = logger.add(caplog.handler, format="{message}")
     yield caplog
     logger.remove(handler_id)
+
+
+def create_citext_extension(engine: Engine) -> None:
```
TheAndrewJackson, line 22: This is required for the test db to work properly. Without it CI fails because the `citext` extension is missing from the test db

src/fides/api/ctl/migrations/versions/643249f65453_case_insensitive_usernames_delete_.py
TheAndrewJackson: I understand now. I'll look into that now

src/fides/api/ctl/migrations/versions/643249f65453_case_insensitive_usernames_delete_.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+def upgrade():
+
+    session = Session(bind=op.get_bind())
+    conn = session.connection()
+    statement = text(
+        """SELECT DISTINCT ON (LOWER(username)) username, created_at, id
+FROM fidesuser
+ORDER BY LOWER(username), created_at ASC"""
```
TheAndrewJackson, line 29: I went with `LOWER` because that's what `citext` does under the hood 

https://www.postgresql.org/docs/current/citext.html

src/fides/api/ctl/migrations/versions/643249f65453_case_insensitive_usernames_delete_.py
```diff
@@ -0,0 +1,49 @@
...
+        f"""DELETE FROM fidesuser 
+    WHERE NOT(id=ANY(ARRAY[{original_user_ids}]::varchar[]))"""
+    )
+
+    op.execute("CREATE EXTENSION IF NOT EXISTS citext;")
+    op.execute("ALTER TABLE fidesuser ALTER COLUMN username TYPE citext;")
+
+
+def downgrade():
+    pass
```
seanpreston, line 49: While we can't restore the deleted data, it would be good to remove the extension here to maintain consistency of the DB across the migration history.

seanpreston:
Great work, thanks @TheAndrewJackson. I'm not sure it would be the right move anyway, since with `citext` we preserve support for Postgres before v12, but did we consider creating a [custom case insensitive collation](https://www.postgresql.org/docs/current/collation.html) for this?

TheAndrewJackson:
@seanpreston We did consider using a collation. I used that because I was under the impression that I needed to. Dawn suggested a solution that worked without needing a collation so I removed it.

src/fides/api/ctl/migrations/versions/643249f65453_case_insensitive_usernames_delete_.py
```diff
@@ -0,0 +1,55 @@
...
+    )
+
+    result = conn.execute(statement)
+    original_user_ids = [row[2] for row in result]
+    op.execute(
+        f"UPDATE client SET user_id = NULL WHERE NOT(user_id=ANY(ARRAY[{original_user_ids}]::varchar[]))"
+    )
+    op.execute(
+        f"UPDATE auditlog SET user_id = NULL WHERE NOT(user_id=ANY(ARRAY[{original_user_ids}]::varchar[]))"
+    )
```
TheAndrewJackson, line 39: I searched the codebase and these were the only other records that had a  `user_id` column

src/fides/api/ctl/migrations/versions/643249f65453_case_insensitive_usernames_delete_.py
```diff
@@ -0,0 +1,55 @@
...
+        f"""DELETE FROM fidesuser 
+    WHERE NOT(id=ANY(ARRAY[{original_user_ids}]::varchar[]))"""
+    )
+
+    op.execute("CREATE EXTENSION IF NOT EXISTS citext;")
+    op.execute("ALTER TABLE fidesuser ALTER COLUMN username TYPE citext;")
+
+
+def downgrade():
+    op.execute("DROP EXTENSION IF EXISTS citext;")
```
pattisdr, line 55: Before you'd be able to downgrade here you'd have to revert the fideruser.username column type like: `op.execute("ALTER TABLE fidesuser ALTER COLUMN username TYPE varchar;")`

src/fides/api/ctl/migrations/versions/643249f65453_case_insensitive_usernames_delete_.py
pattisdr: Great.

One other suggestion, what do you think about replacing this user_id with something like "user_deleted" here? instead of leaving empty for the AuditLogs?

tests/conftest.py
pattisdr: nice one, @TheAndrewJackson

src/fides/api/ctl/migrations/versions/643249f65453_case_insensitive_usernames_delete_.py
pattisdr: ah great

pyproject.toml
```diff
@@ -60,6 +60,7 @@ module = [
   "twilio.*",
   "uvicorn.*",
   "validators.*",
+  "citext.*",
```
pattisdr, line 63: Can you remind me what this does?

pattisdr (commented):
OK looking good @TheAndrewJackson just a couple cleanup items, we're about there.

src/fides/api/ctl/migrations/versions/643249f65453_case_insensitive_usernames_delete_.py
TheAndrewJackson: 🤦‍♂️ Good catch!

src/fides/api/ctl/migrations/versions/643249f65453_case_insensitive_usernames_delete_.py
TheAndrewJackson: I like that. I'll update it to "user_deleted"

src/fides/api/ops/service/privacy_request/request_runner_service.py
```diff
@@ -508,7 +508,10 @@ def initiate_privacy_request_completion_email(
     :param access_result_urls: list of urls generated by access request upload
     :param identity_data: Dict of identity data
     """
-    if not (identity_data.get(ProvidedIdentityType.email.value) or identity_data.get(ProvidedIdentityType.phone_number.value)):
+    if not (
+        identity_data.get(ProvidedIdentityType.email.value)
```
TheAndrewJackson, line 512: No idea how this wasn't caught earlier by `black` 🤷‍♂️

pyproject.toml
pattisdr: great thanks

pattisdr (approved):

PR #2489, opened 2023-02-02
2482 privacy requests approved in quick succession error when not running worker

### LINKED ISSUE #2482: Privacy requests approved in quick succession error when not running worker
### Bug Description

when not running a celery worker, i.e. with `celery.task_always_eager` set to `true`, and when requiring request approval, i.e. `execution.require_manual_request_approval` is set to `true`, if a user of the admin UI clicks "approve" on two privacy requests in quick succession, then one or both of the requests will almost always error out. 

note that with the latest admin UI on `main`, this is actually rather difficult to achieve, since when a worker isn't running, the "approval" call blocks the UI until the request has completed processing and therefore prevents a user from approving another request until the first request is complete. but we can trigger this behavior/bug by opening two browser tabs and approving two pending requests in quick succession.

even if the current UI on `main` makes this race condition less likely to occur in normal application use, it's ultimately a pretty fundamental bug in our non-worker environments that needs to be fixed.

specifically -- non-worker environments (`task_always_eager = true`) use a celery emulation that does _not_ prevent the celery task from running two requests concurrently, and we make assumptions in our code that requests will not be running concurrently in the same celery task. a given task has only one session across multiple privacy requests, since we assume that those multiple privacy requests will not be executed on the same task concurrently.

also, we hope to address the UI blocking behavior in https://github.com/ethyca/fides/issues/2466 - and at that point, the race condition will once again be easy to hit in normal operation within the admin UI.

#### Steps to Reproduce

1. switch `require_manual_request_approval = true` in your `fides.toml`
2. run `nox -s dev -- ui` or `nox -s 'fides_env(dev)'`
3. make sure you've got at least one connector setup, so that request execution doesn't complete instantaneously. `fides_env` should already have the connectors set up, `dev` will not.
4. submit two privacy requests, which will be pending, awaiting approval in the UI. e.g. 
```
curl --location --request POST 'http://0.0.0.0:8080/api/v1/privacy-request' \
--data-raw '[
    {
        "identity": {"email": "test@ethyca.com"},
        "policy_key": "default_access_policy"
    },{
        "identity": {"email": "test2@ethyca.com"},
        "policy_key": "default_access_policy"
    }
]'
```
5. open up two browser tabs to the admin UI to review privacy requests (`http://localhost:3000/privacy-requests`)
6. approve one privacy request in one tab, and quickly approve another privacy request in the other tab. 
7. look for errors in logs, look for at least one of the privacy requests to be in a non-complete state.

### Expected behavior

multiple privacy requests that have been approved in quick succession should not cause problems, even when running without a worker. 

### Screenshots

### Environment

- manual request approval enabled, i.e. `execution.require_manual_request_approval` set to `true`
- non-worker environment, i.e.  `celery.task_always_eager` set to `true`

### Additional context

- we introduced the shared session across a task, which is ultimately the cause for the problem here, in https://github.com/ethyca/fidesops/pull/944, and with good motivation. if we want to change this shared session behavior, we need to be careful not to introduce regressions in the number of connections we're opening up against our app db.
    - some preliminary sqlalchemy research does indicate to me that it's wouldn't necessarily be a bad pattern to create new sessions for each _invocation_ of a privacy request: https://docs.sqlalchemy.org/en/20/orm/session_basics.html#when-do-i-construct-a-session-when-do-i-commit-it-and-when-do-i-close-it, as long as we ensure we've got only a single `Engine` and `sessionmaker` for the lifecycle of the celery task. but this is just my initial understanding, it requires some more testing/input.
- the fix in https://github.com/ethyca/fides/pull/2015 was also adjacent to this problem, but it punted on resolving this core issue, instead just putting in some upstream changes in our endpoints to make the race condition less likely. what we've now uncovered is another way in normal application use to reliably get to the race condition, so we should likely now address the race condition head-on.

### LINKED ISSUE #2486: saas overrides are not loaded into celery worker runtime
### Bug Description

SaaS overrides are not loaded into celery worker runtime, since the celery worker process is not started via our `main.py`, which contains the `from fides.api.ops.service.saas_request.override_implementations import *` that imports the override functions into our general app runtime.

This manifests with runtime errors that prevent privacy request execution when using a celery worker (`nox -s dev -- worker`,  `celery.task_always_eager = false` and attempting to execute a privacy request that hits a connector that leverages an override

#### Steps to Reproduce

1. set `celery.task_always_eager = false` in `fides.toml`
2. set `FIDES__DEV_MODE: "True"` within the `worker` service in our `docker-compose.yml`, in order to get detailed logs on the worker
3. `nox -s dev -- worker`
4. setup a connector that requires an override, e.g. adobe campaign or firebase
5. tail the worker logs in a terminal, e.g. `docker logs -f fides-worker-1`
6. run a privacy request and see the error

### Expected behavior

- request overrides should work when using a worker
    - request overrides should be loaded into the worker runtime

### Additional context

found this while doing testing for https://github.com/ethyca/fides/issues/2482

adamsachs:
Closes #2482 
Closes #2486 

### Code Changes

* create a new db `Session` per celery task _invocation_
* ensure a common `Engine` and `sessionmaker` are reused across invocations to keep db connection resources in check
* update worker "main" module to ensure saas request overrides are added to worker runtime
    * the important line here is https://github.com/ethyca/fides/pull/2489/files#diff-8ad3421e67c70da2822e09434a416a583e1b18d9dd58e87938820f9a07466542R3 - `from fides.api.ops.service.saas_request.override_implementations import *`
    * also needed to split out worker "main" into its own module to avoid circular dependency issues from `tasks/__init__.py` once the above import was added

### Steps to Confirm

- for #2482 fix, run _without_ a worker (e.g. `nox -s 'fides_env(dev)'`) and approve multiple privacy requests in quick succession using multiple browser tabs on the admin UI.
    - ensure privacy requests are executing concurrently, i.e. a second request is approved before the first request has gone to completion. you may need to add longer-running connectors to your local environment to get into this state easily. or you can use a script to send "approve" API calls in rapid succession. (will likely upload a reference script shortly)
- for #2486, run _with_ a worker (e.g. `nox -s dev -- worker ui`) and switch `celery.task_always_eager = false` in `fides.toml`
    - you'll also need to configure a connector that uses request overrides. examples are ACS (adobe campaign standard) or Firebase
    - run a privacy request. before this fix, you'd get errors immediately. with this fix, privacy request execution should proceed normally on the worker.

- see individual issues for more detailed steps to repro

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

These are two distinct fixes that are coupled together here because they touch similar parts of the codebase and came up when testing each other.

Sill pending some local testing to ensure db connections are kept in check even with the changes to session management here

### DIFF
--- a/src/fides/api/ops/tasks/__init__.py
+++ b/src/fides/api/ops/tasks/__init__.py
@@ -1,4 +1,4 @@
-from typing import Any, ContextManager, Dict, List, MutableMapping, Optional, Union
+from typing import Any, ContextManager, Dict, List, Optional
 
 from celery import Celery, Task
 from loguru import logger
@@ -86,20 +86,3 @@ def get_worker_ids() -> List[Optional[str]]:
         logger.critical(exception)
         connected_workers = []
     return connected_workers
-
-
-def start_worker() -> None:
-    logger.info("Running Celery worker...")
-    default_queue_name = celery_app.conf.get("task_default_queue", "celery")
-    celery_app.worker_main(
-        argv=[
-            "worker",
-            "--loglevel=info",
-            "--concurrency=2",
-            f"--queues={default_queue_name},{MESSAGING_QUEUE_NAME}",
-        ]
-    )
-
-
-if __name__ == "__main__":  # pragma: no cover
-    start_worker()
--- a/src/fides/api/ops/tasks/__init__.py
+++ b/src/fides/api/ops/tasks/__init__.py
@@ -3,26 +3,38 @@ from typing import Any, ContextManager, Dict, List, MutableMapping, Optional, Un
 from celery import Celery, Task
 from loguru import logger
 from sqlalchemy.orm import Session
-from toml import load as load_toml
 
 from fides.core.config import FidesConfig, get_config
-from fides.lib.db.session import get_db_session
+from fides.lib.db.session import get_db_engine, get_db_session
 
 CONFIG = get_config()
 MESSAGING_QUEUE_NAME = "fidesops.messaging"
 
 
 class DatabaseTask(Task):  # pylint: disable=W0223
-    _session = None
+    _task_engine = None
+    _sessionmaker = None
 
-    @property
-    def session(self) -> ContextManager[Session]:
-        """Creates Session once per process"""
-        if self._session is None:
-            SessionLocal = get_db_session(CONFIG)
-            self._session = SessionLocal()
+    def get_new_session(self) -> ContextManager[Session]:
+        """
+        Creates a new Session to be used for each task invocation.
 
-        return self._session
+        The new Sessions will reuse a shared `Engine` and `sessionmaker`
+        across invocations, so as to reuse db connection resources.
+        """
+        # only one engine will be instantiated in a given task scope, i.e
+        # once per celery process.
+        if self._task_engine is None:
+            _task_engine = get_db_engine(config=CONFIG)
+
+        # same for the sessionmaker
+        if self._sessionmaker is None:
+            self._sessionmaker = get_db_session(config=CONFIG, engine=_task_engine)
+
+        # but a new session is instantiated each time the method is invoked
+        # to prevent session overlap when requests are executing concurrently
+        # when in task_always_eager mode (i.e. without proper workers)
+        return self._sessionmaker()
 
 
 def _create_celery(config: FidesConfig = get_config()) -> Celery:
--- a/src/fides/cli/commands/util.py
+++ b/src/fides/cli/commands/util.py
@@ -94,7 +94,7 @@ def worker(ctx: click.Context) -> None:
     Starts a celery worker.
     """
     # This has to be here to avoid a circular dependency
-    from fides.api.ops.tasks import start_worker
+    from fides.api.ops.worker import start_worker
 
     start_worker()
 
--- a/tests/ctl/cli/test_cli.py
+++ b/tests/ctl/cli/test_cli.py
@@ -58,6 +58,17 @@ def test_webserver() -> None:
 
 
 @pytest.mark.unit
+def test_worker() -> None:
+    """
+    This is specifically meant to catch when the worker command breaks,
+    without spinning up an additional instance.
+    """
+    from fides.api.ops.worker import start_worker  # pylint: disable=unused-import
+
+    assert True
+
+
+@pytest.mark.unit
 def test_parse(test_config_path: str, test_cli_runner: CliRunner) -> None:
     result = test_cli_runner.invoke(
         cli, ["-f", test_config_path, "parse", "demo_resources/"]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -77,6 +77,8 @@ The types of changes are:
 * Patch masking strategies to better handle null and non-string inputs [#2307](https://github.com/ethyca/fides/pull/2377)
 * Renamed prod pushes tag to be `latest` for privacy center and sample app [#2401](https://github.com/ethyca/fides/pull/2407)
 * Update firebase connector to better handle non-existent users [#2439](https://github.com/ethyca/fides/pull/2439)
+* Fix errors when privacy requests execute concurrently without workers [#2489](https://github.com/ethyca/fides/pull/2489)
+* Enable saas request overrides to run in worker runtime [#2489](https://github.com/ethyca/fides/pull/2489)
 
 
 ## [2.5.1](https://github.com/ethyca/fides/compare/2.5.0...2.5.1)
--- a/src/fides/api/ops/service/messaging/message_dispatch_service.py
+++ b/src/fides/api/ops/service/messaging/message_dispatch_service.py
@@ -99,7 +99,7 @@ def dispatch_message_task(
     A wrapper function to dispatch a message task into the Celery queues
     """
     schema = FidesopsMessage.parse_obj(message_meta)
-    with self.session as db:
+    with self.get_new_session() as db:
         dispatch_message(
             db,
             schema.action_type,
--- a/src/fides/api/ops/service/privacy_request/request_runner_service.py
+++ b/src/fides/api/ops/service/privacy_request/request_runner_service.py
@@ -291,7 +291,7 @@ async def run_privacy_request(
     if from_step:
         logger.info("Resuming privacy request from checkpoint: '{}'", from_step)
 
-    with self.session as session:
+    with self.get_new_session() as session:
         privacy_request = PrivacyRequest.get(db=session, object_id=privacy_request_id)
 
         privacy_request.cache_failed_checkpoint_details()  # Reset failed step and collection to None
--- a/src/fides/api/ops/worker/__init__.py
+++ b/src/fides/api/ops/worker/__init__.py
@@ -1,0 +1,21 @@
+from loguru import logger
+
+from fides.api.ops.service.saas_request.override_implementations import *
+from fides.api.ops.tasks import MESSAGING_QUEUE_NAME, celery_app
+
+
+def start_worker() -> None:
+    logger.info("Running Celery worker...")
+    default_queue_name = celery_app.conf.get("task_default_queue", "celery")
+    celery_app.worker_main(
+        argv=[
+            "worker",
+            "--loglevel=info",
+            "--concurrency=2",
+            f"--queues={default_queue_name},{MESSAGING_QUEUE_NAME}",
+        ]
+    )
+
+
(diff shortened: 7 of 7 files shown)


adamsachs:
running some tests locally, both using a worker or not, and requiring request approval or not, i feel reasonably confident that we haven't introduced significant regressions in terms of database connection resources, especially when looking at test results that @pattisdr had gathered in https://github.com/ethyca/fidesops/pull/944


i tested locally by using the attached testing script, tweaking number of requests and approval concurrency. when running the tests, i had another terminal running `SELECT sum(numbackends) FROM pg_stat_database;` against the app postgres db, to track how many open connections we had at any given moment.

here were my general findings:
- running a worker, we stayed between 6-8 connections consistently, even executing 20 requests at a time many times over. everything was stable.
- without workers, we need to turn on manual request approval in order to get any sort of concurrency - otherwise, requests that are created just get queued up naturally when bulk creating. 
    - consistently approving batches of 5-10 requests one after the other, our active connections did rise up to 25, but they also would drop down after requests finished processing.
    - there was no evidence of a significant connection "leak" as was found in https://github.com/ethyca/fidesops/pull/944 testing
    - after enough batches of 10 concurrent approvals, i  _did_ start seeing `QueuePool limit of size 5 overflow 10 reached` errors pop on the server logs
        - when this happened, privacy requests that were processing continued to completion successfully, while some never actually got "approved" and stayed in a "new"/pending state, able to be approved and then processed successfully after the server got back to normal functioning
    - my reading of this is that we haven't introduced a problem, but seem to just be running into a limit on our connection pool. looking more closely at the stack trace actually seems to be indicating that the problem is occurring not in our task's `Engine`, but instead in the `Engine` used generally by the API endpoints. (stack track attached)
    - before this fix, we were not able to run _any_ concurrent requests when not running a worker - we'd error almost immediately. in that sense, it's hard to really determine whether this is a regression or we're just hitting new edge cases, because we couldn't even get more than one request running in parallel before this fix
    - in my testing so far, these symptoms seem a lot easier to deal with than the problem we were facing before, and from what i can tell this seems to be a separate problem
- if i hit the privacy request approval endpoint with 20 clients at the same time, it actually does cause the server to lock up pretty consistently. this does not seem to occur with normal application use, but did occur if i sent 20 approval requests in parallel from my testing script.
    - i confirmed that this occurs both when running a worker and when not running a worker. it also occurs even without any of the changes in this PR, so i think it is a separate issue that likely needs to be investigated - it may have the same root cause as the connection pool limit being reached with batches of ~10 concurrent approval requests, i.e. what's mentioned above.

[QueuePool_limit_stack.txt](https://github.com/ethyca/fides/files/10575423/QueuePool_limit_stack.txt)
[execute_requests_concurrently.py.txt](https://github.com/ethyca/fides/files/10575425/execute_requests_concurrently.py.txt)

pattisdr:
Starting review...

src/fides/cli/commands/util.py
```diff
@@ -94,7 +94,7 @@ def worker(ctx: click.Context) -> None:
     Starts a celery worker.
     """
     # This has to be here to avoid a circular dependency
-    from fides.api.ops.tasks import start_worker
+    from fides.api.ops.worker import start_worker
```
seanpreston, line 97: Thanks this is a nice semantic change

tests/ctl/cli/test_cli.py
```diff
@@ -57,6 +57,17 @@ def test_webserver() -> None:
...
 
+@pytest.mark.unit
+def test_worker() -> None:
+    """
+    This is specifically meant to catch when the worker command breaks,
+    without spinning up an additional instance.
+    """
+    from fides.api.ops.worker import start_worker  # pylint: disable=unused-import
+
+    assert True
```
seanpreston, line 68: Can you describe further the kind of breakage this would catch beyond issues with the imports?

src/fides/api/ops/tasks/__init__.py
```diff
@@ -1,28 +1,40 @@
...
+        across invocations, so as to reuse db connection resources.
+        """
+        # only one engine will be instantiated in a given task scope, i.e
+        # once per celery process.
+        if self._task_engine is None:
+            _task_engine = get_db_engine(config=CONFIG)
+
+        # same for the sessionmaker
+        if self._sessionmaker is None:
+            self._sessionmaker = get_db_session(config=CONFIG, engine=_task_engine)
```
seanpreston, line 32: Now that two sessions can concurrently change the database's state we should consider changing the `autocommit` and `autoflush` defaults to `True`. Currently in the execution model a privacy request's execution is locked to one worker, but the app server and a worker could theoretically be editing the same privacy request concurrently.

src/fides/cli/commands/util.py
adamsachs: 👍 actually needed to split it out into its own module for dependency reasons, but i do think the organization now is a bit cleaner semantically.

pattisdr (approved):
This seems like a necessary change for customers _not_ running workers: it doesn't seem even workable without this.  As an aside, it would be useful for us to better understand what is actually happening when `task_always_eager` is True.   In my opinion, the preferred mode for running fides should be with workers.  If customers are not doing that, we need to investigate if `task_always_eager=True` is the best way to actually execute privacy requests without workers and if there are other side effects we're not aware of.   

----

**Background from last PR**

This PR https://github.com/ethyca/fidesops/pull/944 assumed we were running celery workers and it makes sense to me why this is starting to breakdown when we're not using workers. 

The biggest benefit from the PR was sharing _engines_ across a celery process.  I did add some smaller changes that also seemed to improve session management alongside: like passing along the same session for execution logs and sharing sessions, but sharing a session wasn't a requirement so it makes sense to investigate that here.

**Testing setup:**

I used Adam's script (thanks for attaching!) to "Make 20 requests with concurrency 5 to approve privacy requests", comparing with a worker / without a worker, to sharing a sessionmaker (new behavior) versus sharing a session (previous behavior).  I also increased the `pool_size` and `max_overflow` to 50 each to avoid hitting these limits early and be able to properly compare.  I had two connectors set up to increase the duration of a privacy request.  I monitored the number of connections open and verified privacy request statuses and execution logs.

**Takeaways**

- It would be useful to add a separate PR sooner rather than later to allow the engine `pool_size` and `max_overflow` to be configurable.  
- Running a worker: Didn't see huge differences between sharing a session versus only sharing the sessionmaker.  In my previous testing in the last PR, I had noticed slightly better session management when sharing a session, but wasn't seeing that in this particular test setup. 
- No worker:  Confirmed huge improvements when not sharing a session.  With the pool_size and max_overflow increased, session management seemed similar (just looking at connections used, max idle connections, etc).. However, sharing a session was causing major errors in privacy requests being able to complete successfully.  97% of privacy requests were failing with Session-related errors, like `NoneType' object has no attribute 'dispatch'"` or `"Instance '<PrivacyRequest at 0xffff7d23eb60>' is not persistent within this Session"`, or `NoneType' object has no attribute 'twophase'`

src/fides/api/ops/tasks/__init__.py
seanpreston: @adamsachs and I discussed this synchronously! This can be addressed separately.

adamsachs: per @seanpreston's slack
>  I don’t think it’s an immediate issue given it’s unlikely for the app server and worker to be saving the same privacy request at the same time

we will create a follow up to look into this further (GH issue coming soon)

tests/ctl/cli/test_cli.py
seanpreston: Generally we can assume the `celery` dependency is tested too, they provide [`pytest` bindings](https://docs.celeryq.dev/en/stable/userguide/testing.html) to virtualise workers.

What's more important is that our application logic within both `start_worker` and individual tasks is tested. Since `start_worker` is only invoking `celery`, the other place to look would be `_create_celery`, which is what loads the config given our rules:
- default config (ie. Redis instance)
- overridden with anything explicitly specified in the `celery` config file
which is [tested here](https://github.com/ethyca/fides/blob/main/tests/ops/tasks/test_celery.py#L19-L34).

seanpreston (approved):

src/fides/api/ops/tasks/__init__.py
adamsachs: created https://github.com/ethyca/fides/issues/2508 for follow up

adamsachs:
> It would be useful to add a separate PR sooner rather than later to allow the engine pool_size and max_overflow to be configurable.

created https://github.com/ethyca/fides/issues/2507 for follow up

PR #2500, opened 2023-02-03
Remove ui-build after building wheel

sanders41:
Closes #2491

### Code Changes

* Removes the `ui-build` directory after the wheel build is completed

### Steps to Confirm

* [ ] Run `nox -s "fides_env(test)"`
* [ ] Navigate to `http://localhost:8080` and log in
* [ ] Navigate to any page that has a route after the `/`, such as `http://localhost:8080/add-systems`
* [ ] Refresh the browser and verify the page loads with no error.

### Pre-Merge Checklist

* [ ] All CI Pipelines Succeeded
* Documentation:
  * [ ] documentation complete, [PR opened in fidesdocs](https://github.com/ethyca/fidesdocs/pulls)
  * [ ] documentation [issue created in fidesdocs](https://github.com/ethyca/fidesdocs/issues/new/choose)
* [ ] Issue Requirements are Met
* [ ] Relevant Follow-Up Issues Created
* [ ] Update `CHANGELOG.md`

### Description Of Changes

_Write some things here about the changes and any potential caveats_

Issue #2491, opened 2023-02-02
browser refresh results in item not find 


Roger-Ethyca:
### Bug Description

_A description of what the bug is._

1. Pull down the lates fides deploy 2.6.0
2. Run deploy up
3. Log in and navigate to any page other than the home page.
4. Click refresh 

### Expected behavior

page should refresh however it errors out with a item not found. it should be noted that this doesn't happen when running test_env

![Screen Shot 2023-02-02 at 5 33 12 PM](https://user-images.githubusercontent.com/115798183/216465009-a1d33ccc-092b-4d9b-a5fb-02f3185855e6.png)
![Screen Shot 2023-02-02 at 5 33 26 PM](https://user-images.githubusercontent.com/115798183/216465030-bbf0a6d7-2f26-4426-a379-18a320c07956.png)

### Screenshots

_If applicable, add screenshots to help explain your problem._

### Environment

* Version:
* OS:
* Python Version:
* Docker Version:

### Additional context

_Add any other context about the problem here._

PR #2502, opened 2023-02-03
Consolidate paths

allisonking:
A more thorough fix for https://github.com/ethyca/fides/issues/2491

### Code Changes

* [x] Removes the local file map check which doesn't need to exist anymore since the package file map check can handle both cases (installed at `src/fides` or installed at `/usr/local/lib/python/site-packages`

### Steps to Confirm

The hotfix in https://github.com/ethyca/fides/pull/2500 fixes things, but not if you have your volume mount active. This should fix it whether or not you have your volume mount around.

To repro in the dev environment
* [ ] Have a local build via `npm run prod-export`
* [ ] `nox -s "fides_env(dev)"`
* [ ] Navigate to `localhost:8080/dataset`. it should load now!

To repro in the prod environment
* [ ] Have a local build via `npm run prod-export`
* [ ] Navigate to `localhost:8080/dataset`. it should load now!

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### Description Of Changes


src/fides/api/main.py
```diff
@@ -321,14 +320,10 @@ def read_other_paths(request: Request) -> Response:
     # check first if requested file exists (for frontend assets)
     path = request.path_params["catchall"]
 
-    # First search in the local (dev) build for for a matching route.
-    ui_file = match_route(get_local_file_map(), path)
```
allisonking, line 325: it seemed like `get_package_file_map` when running via `fides_env(dev)` returned `/fides/src/fides` while the `get_local_file_map` returned `src/fides` (relative path). the relative path caused problems when doing the comparison we do further down to see if it has the same parent as the admin UI path.

but since `get_package_file_map` already returns `/fides/src/fides` (or `/usr/local/lib/python3.10/site-packages/fides` for built packages) it seems like we can just use that in lieu of `get_local_file_map`? I may very well be missing something, but this seems to work

src/fides/api/main.py
```diff
@@ -321,14 +320,10 @@ def read_other_paths(request: Request) -> Response:
...
+    # search for matching route in package
+    ui_file = match_route(get_package_file_map(), path)
 
-    # Next, search for a matching route in the packaged files.
-    if not ui_file:
-        ui_file = match_route(get_package_file_map(), path)
-
-    # Finally, try to find the exact path as a packaged file.
+    # if not, try to find the exact path as a packaged file.
```
allisonking, line 326: I actually don't remember why we'd hit this case vs the previous case?

allisonking: okay I get it—the previous line is for _routes_ and this line is for everything else (i.e. css, images, etc.). I'm updating the comment to reflect that

TheAndrewJackson (approved):
Sorry for the delayed review. It works great! 🚀

PR #2511, opened 2023-02-06
provide config overrides for db engine pool params

### LINKED ISSUE #2507: provide config properties to set db engine `pool_size` and `max_overflow`

based on testing and investigation in https://github.com/ethyca/fides/pull/2489 and related issues, we've realized it would be useful to allow our sql alchemy db `Engine` `pool_size` and `max_overflow` [settings](https://docs.sqlalchemy.org/en/20/core/pooling.html) to be configurable as application config properties.

this would be useful for both testing purposes and also for fine-tuning application deployment in the wild - the optimal values for these settings are likely dependent on both expected application workload as well as constraints on the database side - i.e. environmental factors that we can't easily predict!

- see https://docs.sqlalchemy.org/en/20/core/pooling.html
- we should be able to expose these as application configs - likely in our `DatabaseSettings` module?

- one (the only?) open question i have is whether we should expose different configurable values for the different `Engine`s we create, namely:
    - we have one `Engine` used for API operations
    - and we create an `Engine` per celery task (for privacy request execution and for message dispatch). if using workers, this will be a separate fides instance; if not, though this will be using the same fides instance, and would therefore read from the same config

### Additional context

See comments on https://github.com/ethyca/fides/pull/2489 and https://github.com/ethyca/fidesops/pull/944 for some additional context

adamsachs:
Closes #2507 

Doc updates are in https://github.com/ethyca/fidesdocs/pull/52

### Code Changes

* provide `database` config properties to allow user configurability of `pool_size` and `max_overflow` tunings for the app db engines used by fides
    * a separate pair of settings for both types of engines that our application creates: the engine used generally by the fides API (`database.api_engine_pool_size` and `database.api_engine_max_overflow`), and the engine used by specific celery tasks/processes (`database.task_engine_pool_size` and `database.task_engine_max_overflow`)

### Steps to Confirm

* set respective config values in your `fides.toml` or via environment variable and confirm the db engine that's created is using the set values
* the test script mentioned here https://github.com/ethyca/fides/pull/2489#issuecomment-1414872864 can be a good way to evaluate that raising these settings (in this case, specifically the `task_engine_*` settings) is effective

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### DIFF
--- a/src/fides/api/ops/api/deps.py
+++ b/src/fides/api/ops/api/deps.py
@@ -30,7 +30,11 @@ def get_api_session() -> Session:
     """Gets the shared database session to use for API functionality"""
     global _engine  # pylint: disable=W0603
     if not _engine:
-        _engine = get_db_engine(config=CONFIG)
+        _engine = get_db_engine(
+            config=CONFIG,
+            pool_size=CONFIG.database.api_engine_pool_size,
+            max_overflow=CONFIG.database.api_engine_max_overflow,
+        )
     SessionLocal = get_db_session(CONFIG, engine=_engine)
     db = SessionLocal()
     return db
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,9 @@ The types of changes are:
 
 ## [Unreleased](https://github.com/ethyca/fides/compare/2.6.1...main)
 
+### Added
+* Added config properties to override database Engine parameters [#2511](https://github.com/ethyca/fides/pull/2511)
+
 ## [2.6.1](https://github.com/ethyca/fides/compare/2.6.0...2.6.1)
 
 ## [2.6.0](https://github.com/ethyca/fides/compare/2.5.1...2.6.0)
--- a/src/fides/api/ops/tasks/__init__.py
+++ b/src/fides/api/ops/tasks/__init__.py
@@ -25,7 +25,11 @@ class DatabaseTask(Task):  # pylint: disable=W0223
         # only one engine will be instantiated in a given task scope, i.e
         # once per celery process.
         if self._task_engine is None:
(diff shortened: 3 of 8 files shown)


src/fides/api/ops/api/deps.py
```diff
@@ -30,7 +30,11 @@ def get_api_session() -> Session:
     """Gets the shared database session to use for API functionality"""
     global _engine  # pylint: disable=W0603
     if not _engine:
-        _engine = get_db_engine(config=CONFIG)
+        _engine = get_db_engine(
+            config=CONFIG,
+            pool_size=CONFIG.database.api_engine_pool_size,
+            max_overflow=CONFIG.database.api_engine_max_overflow,
```
pattisdr, line 36: I think this will work well for the API endpoints ops-side but the ctl side creates their own synchronous and asynchronous engines, that aren't taking in these settings - alternatively this might be a good time to unify if we're trying to decrease the number of engines created overall:

The ctl side has a separate [get_db_engine](https://github.com/ethyca/fides/blob/main/src/fides/core/utils.py#L65) method that a few of their API endpoints call, and a lot of their other endpoints use the [async or sync engines created here](https://github.com/ethyca/fides/blob/main/src/fides/api/ctl/database/session.py)

I stumbled on this in my updates  to populate the application db instead of the test db when running the application locally https://github.com/ethyca/fides/pull/1731, but didn't go all the way through with trying to unify this between ctl/ops.

pattisdr (approved):
looks good, we can add a followup for the rest

adamsachs:
merging while @pattisdr creates a follow up for the valid feedback 🙏

pattisdr:
followup here: https://github.com/ethyca/fides/issues/2557

Issue #495, opened 2022-04-14
Bump Python to version 3.9


ThomasLaPiana:
### Is your feature request related to a specific problem?

Fidesops and Fidescls are currently both on Python 3.9. As they are both currently relying on fidesctl and we want our ecosystem to play nicely, we should update to version 3.9

### Describe the solution you'd like

Bump the Dockerfile Python version

### Describe alternatives you've considered, if any

Leaving as-is but testing to make sure our features work on both 3.8 and 3.9 using something like `tox` (this would be a good idea anyway though probably?)

### Additional context

There are various changes in 3.9 that might necessitate us updating some of our requirements as well

sanders41:
Related to the tox question, fidesctl is pip installable from pypi with Python 3.7 through 3.10 as supported versions, from setup.py `python_requires=">=3.7, <4"`. However, only Python 3.8 running in Docker is tested in CI. It seems like it would be beneficial to have all supported Python version tested in CI.

ThomasLaPiana:
agreed, but if we're covering 3.8 -> 3.10 we might as well stay on 3.8 so we don't accidentally use breaking features from 3.9

sanders41:
If you are testing all supported Python versions in CI you can be confident all supported version work. At that point using any of the supported Python versions as the base in docker is fine, I don't think there is a right or wrong answer there (including saying with 3.8).

Either way I would suggest bumping the dependencies. Because the dependencies here are hard pined to to specific version, the current pined versions seem to be causing issues in newer Pythons. For example the current pinned version of Pydantic (I am assuming this is coming from the pinned version of FastAPI) starts breaking somewhere around Python 3.9.4.

ThomasLaPiana:
going to investigate this as part of #518 , which would include multi-version python testing and thus a reworking of the pinned dependencies

ThomasLaPiana:
Going to close this and instead open up an issue to make sure that we're testing across multiple python versions

PR #2531, opened 2023-02-07
Add custom fields to datamap

sanders41:
Closes [#495](https://github.com/ethyca/fidesplus/issues/495)

### Code Changes

* Add a query to retrieve resources with their custom fields
* Update processing of datamap to allow custom fields
* Update the processing of the datamap return type to retain custom fields
* Update processing of datamap to work with both the CLI and API route

### Steps to Confirm

* [x] Add resources
* [x] Run `fides export datamap --output-dir data_map/` and verify the resulting datamap looks as expected
* [x] Get the datamap through the API using `default_organization` and verify the resulting datamap looks as expected
* [x] Add some custom fields. Because the routes for this are in plus the fields here have to be entered directly into the database.
* [x] Get the datamap through the API using `default_organization` and verify the custom fields are present.

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes


src/fides/api/ctl/routes/datamap.py
```diff
@@ -113,7 +117,7 @@ async def export_datamap(
     # load resources from server, filtered by organization
     try:
         try:
-            organization = await get_resource(
+            organization = await get_resource_with_custom_fields(
```
SteveDMurphy, line 120: Initial thoughts around that test failure are that maybe we are looking for custom fields on the wrong resource type (although I could be mistaken) - can custom fields go anywhere or focused on `System` currently?

Here they are specifically pulling in the `Organization` where a few lines below we pull in the remaining resource types (`System`, `Dataset`, etc.) - maybe if we use that query on those other objects it gets us what we are looking for?

sanders41:
@SteveDMurphy I believe this is ready for review now

src/fides/core/export_helpers.py
```diff
@@ -194,9 +223,17 @@ def format_data_subjects(
     ]
 
     formatted_data_subjects: Dict[FidesKey, Dict[str, str]] = {}
+    excluded_attributes = (
+        "name",
+        "rights_available",
+        "automated_decisions_or_profiling",
+    )
```
SteveDMurphy, line 230: So I think what I am seeing is there is an attribute (`system.privacy_declaration.data_subjects.tags`) that wasn't being included when exporting previously as we were being exclusive as to what was allowed in an export. When adding a `custom_field` it is now inclusive of everything in the base model judging the remainder to be custom fields. When adding tags in here as an `excluded_attribute` and into `generate_system_records()` it appears to lead to a bit of whack-a-mole with other attributes.

I think it may be a question of if we want to trace down all the attributes and maintain them inclusively or if we want 
 to treat custom fields as a separate defined object. You've been deep into this lately as well, any other ideas potentially?

SteveDMurphy (changes requested):
Hit a snag with the API output that is referenced with more detail in a comment. Sorry it took me a bit to trace down exactly what I was seeing!

src/fides/core/export_helpers.py
SteveDMurphy: It turns out this had already been hit with `DataUse` and had the fields explicitly applied. I replicated the same style for `DataSubject` and everything was able to succeed as expected

src/fides/core/utils.py
```diff
@@ -102,9 +102,14 @@ def get_all_level_fields(fields: list) -> Iterator[DatasetField]:
     """
     for field in fields:
         yield field
-        if field.fields:
-            for nested_field in get_all_level_fields(field.fields):
-                yield nested_field
+        if isinstance(field, dict):
+            if field["fields"]:
+                for nested_field in get_all_level_fields(field["fields"]):
+                    yield nested_field
```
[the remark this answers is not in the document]
SteveDMurphy: I believe it is dict over models due to variability in custom fields

SteveDMurphy (approved):

Issue #528, opened 2022-04-27
Include db url string formats for convenience


iamkelllly:
### Docs Update Description

> As a developer that wants to generate a dataset yaml, I want to run `generate dataset` in the CLI using a MySQL database connection URL I already have.

Even though it might be obvious to some to look up what the db connection string format is for the various sqlalchemy connectors, it was really not easy to find in our documentation. 

Steps to reproduce:
1. Go to docs site, search for "generate", get the [Generate Resources page](https://ethyca.github.io/fides/1.5.2/guides/generate_resources/)
2. See an example for a postgres database
```
./venv/bin/fidesctl generate dataset db \
  postgresql://postgres:postgres@localhost:5432/flaskr \
  fides_resources/flaskr_postgres_dataset.yml
```

3. Search in the docs site for "mysql", get the [INstallation from Pypi](https://ethyca.github.io/fides/1.5.2/installation/pypi/) documentation. 
4. Still don't see db connection URL format string. 
5. Search code (in my IDE) for "mysql, found in setup.py: `mysql_connector = "pymysql==1.0.0"`
6. looked up in sqlalchemy documentation what the connection string should look like: https://docs.sqlalchemy.org/en/14/dialects/mysql.html#dialect-mysql-pymysql-connect

5 minutes later, I had a working connection string

### Recommended documentation
On the docs site, under the `generate dataset` documentation.  we should be including either:
1. a link to the sqlalchemy documentation for each of the currently supported db connectors 
2. an example connection string for each of the currently supported db connectors

Issue #526, opened 2022-04-27
pre-commit does not find files after project restructure


sanders41:
### Bug Description

After the restructuring of the project directories pre-commit no longer finds the files to check. Prior to this update I had a separate issue with the pre-commit config throwing errors. Pre-commit uses caching to speed things up and I think this was not playing well with the use of Docker to run the checks. After one run I would always have to clean the pre-commit cache in order to run again without errors.

Pre-commit is an awesome tool that I would like to use so I am going to open a PR with a proposal for an update to the `pre-commit-config.yaml` that will make it both much faster to run the checks and run without error. It will only take me a couple minutes to put the proposal together so if no one else likes it and would rather close it, it won't be a big loss.

#### Steps to Reproduce

1. Make any change to the code
1. Run `pre-commit run -a`

### Expected behavior

pre-commit should run checks

### Screenshots

![Screenshot from 2022-04-27 08-28-25](https://user-images.githubusercontent.com/25045024/165517984-a80d49e3-3543-4d9d-b75a-414d396a2673.png)

### Environment

* Version: 1.5.3
* OS: Linux
* Python Version: N/A
* Docker Version: N/A

### Additional context

The current `pre-commit-config.yaml` file is looking for the top level `fidesctl` directory, but this directory no long exists so no files are checked.

PR #525, opened 2022-04-27
Fix docstring quotes

sanders41:
Closes N/A

### Code Changes

* [x] Changed doc string for using one set of quotes to using 3.

### Steps to Confirm

* [x] View the docstring for `check_server`

### Pre-Merge Checklist

* [x] All CI Pipelines Succeeded
* [ ] Documentation Updated
* [ ] Issue Requirements are Met
* [ ] Relevant Follow-Up Issues Created
* [ ] Update `CHANGELOG.md`

### Description Of Changes

I noticed the docstring on `check_server` only had a single set of quotes so it isn't really treated as a docstring by Python `"Runs a health check and a version check against the server."`

PR #2536, opened 2023-02-07
Custom metadata and reporting

### LINKED ISSUE #524: Remove fideslang code, instead importing from fideslang via pip

Fideslang now has its own repo and releases on pypi/docker. Fidesctl should remove all fideslang code and instead rely on the pip package

Described above

n/a

### Additional context

This should be relatively painless, and will remove lots of code/testing. There is a single function that will need to get copied into fidesctl before everything functions 100% in the `relationship.py` module

chriscalhoun1974:
~~PLEASE DOT NOT MERGE this till https://github.com/ethyca/fidesplus/pull/581 has been merged into `main`~~
[Sebastian edit: this is feature flagged folks have been [informed](https://ethyca.slack.com/archives/C043URV1KLL/p1675892546709599?thread_ts=1675803915.275239&cid=C043URV1KLL) of the dependency for testing]

Closes #524 #525 #526 #528
### Code Changes

* All of my code changes are behind the `customFields` feature flag. Currently, this feature flag is only turned on for DEV environment. You can update this directly in the `fides > clients > admin-ui > src > flags.json` file.

Screenshot 2023-02-07 at 10 06 04 PM

* Only a FidesPlus user can access the custom fields functionality
* Allow a user to create a custom field and assign it to an edited resource such as a Data Category, Data Subject, Data Use, and System.
* Clicking on the `Add a custom` field button will trigger the `Add custom field `modal
* Within the `Add custom field` modal, user can create one or more custom list(s) and assign this list to a newly created custom field.
* Custom field definition supports `Single select `and `Multiple select` functionality
* When a user clicks the `Save` button within the modal, a successful toast notification is displayed. The tab form is always in a create mode. 
* In the event the user clicks the `Close` or `Cancel` button or clicks outside the modal with unsaved data changes, a prompt will be displayed asking if the user would like to continue editing or discard the data changes.

### Steps to Confirm

1. Execute the `make server `command via FidesPlus terminal window
2. Run the Fides Admin UI via `npm run dev` command via Fides Admin UI terminal window
3. Open a browser tab and navigate to http://localhost:3000
4. Login as `root_user`
5. Click the `Management` top level nav link
6. Edit a Data Category, Data Uses, or Data Subjects record by clicking on the **Edit** button
7. Click the `Add a custom field` button
8. Create a custom list
9. Create a custom field
10. Close the modal
11. Select one or more custom field values
12. Click the `Update entity` button to assign the custom field to a given resource

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### Screen Shots

Screenshot 2023-02-07 at 10 25 15 PM
Screenshot 2023-02-07 at 10 25 32 PM
Screenshot 2023-02-07 at 10 25 59 PM
Screenshot 2023-02-07 at 10 30 33 PM
Screenshot 2023-02-07 at 10 31 39 PM
Screenshot 2023-02-07 at 10 33 30 PM
Screenshot 2023-02-07 at 10 35 45 PM


chriscalhoun1974:
@adamsachs @sanders41 When invoking the API endpoint http://localhost:3000/api/v1/plus/custom-metadata/custom-field PUT, I am receiving the following exception: 

`Invalid resource type`

NOTE: This is happening on the create custom field for a given resource.

Screenshot 2023-02-07 at 11 44 05 PM
Screenshot 2023-02-07 at 11 44 23 PM

adamsachs:
@chriscalhoun1974 i think the issue there may be that you're passing a `resource_type` field at all in this call. 

the `PUT /custom-field` endpoint just needs to have the following fields: `resource_id`, `custom_field_definition_id` and `value`. the resource type is inferred on the backend based on the associated `custom_field_definition_id` and so it's not needed in the request payload here. are we able to try without that field and see if it resolves things?

clients/admin-ui/src/features/common/custom-fields/CustomFieldsList.tsx
```diff
@@ -0,0 +1,224 @@
...
+
+const CustomFieldsList = ({
+  resourceId,
+  resourceType,
+}: CustomFieldsListProps) => {
+  const { errorAlert, successAlert } = useAlert();
+
+  const { flags, plus } = useFeatures();
+  const skip = !(flags.customFields && plus);
+  const [isLoading, setIsLoading] = useState(!skip);
```
ssangervasi, line 39: We should move away from storing `isLoading` in state. When I opened this page the API returned an error, but the spinner stuck around even though the app wasn't waiting for anything.

`isLoading` can always be derived some other state, such as the loading state of multiple queries:
https://github.com/ethyca/fides/blob/c0866f3f5f8c68c7eaaae0fc7c3ce294397bf095/clients/admin-ui/src/features/dataset/DatasetCollectionView.tsx#L53

Additional logic like the `useEffect` below could be a `useMemo` instead, but usually it's just boolean operations that can done inline without worrying about render optimization (and actually introducing memo checks introduces _more_ comparison operations).

ssangervasi:
Ok, here's my testing results:

Server commit: https://github.com/ethyca/fidesplus/commits/c48ac26
UI commit: https://github.com/ethyca/fides/commits/23fd3e2d9 (Chris's PR#2536 squashed)

**Main checks**
- [x] No errors seemed to be the API's fault. (No 500; validation errors were correct.)
- [x] With the feature flag (`customFields`) disabled, editing taxonomy entities still works.

**Functionality status**
- Defining custom lists
  - [x] Create
  - [x] Use for a field definition
- Defining custom fields
  - [x] Create single select
  - [x] Create multi select
  - [x] Cancel button warns about discarding changes
- [ ] "Show in forms and reports"
  - How do we verify this?
- Using single select
  - [x] Initial selection
  - [ ] Removing selection
  - [ ] Changing selection
  - [ ] Initial load when data has been saved
- Using multi select
  - [x] Initial selection
  - [ ] Removing selection
  - [ ] Changing selection
  - [ ] Initial load when data has been saved
- [ ] When there are multiple metadata fields
- [ ] Switching between taxonomy entities
- [x] Switching between resource types
  - Does seem to reset the form correctly


**Not good**
- On initial load the spinners don't go away
- The requests 404 because no fields have been added to a category
  - The API is correct to return 404
  - The UI needs to understand the "nothing exists yet" case
  - Probably causes the spinners problem
- Clicking "edit" on another category doesn't update the custom fields form
  - Inconsistent
  - Might only happen if you made change in the form
  - Wait, unless custom fields are supposed to show up for all data categories? In which case the problem is that the form sometimes forgets which fields to show.
- Removing a item from the field select list doesn't remove the item
  - But if you "Update entity" and refresh the page it will have been saved
- Clicking "save" doesn't close the add field modal
- Clicking "save" doesn't refresh the custom field form
- Having two fields backed by the same select list causes the form fields to get mixed up
  - This should be fine. They should have the same list of choices, but the values should be independent.
  - Actually, it might not matter which custom list the fields use. Values seem to attach to the wrong fields sometimes regardless.
  - Can result in API error response because a single select was passed an array or a multi was passed a string
    - http://localhost:3000/api/v1/plus/custom-metadata/custom-field
    ```json
    {"custom_field_definition_id":"plu_a5f493f8-cbb7-435d-a94c-f28ec281f4c1","id":"plu_5ee31d1d-7ed2-44fe-9cfb-6970295f0d39","resource_id":"system.operations","value":["a"]}
    ```
    ```json
    {"detail":"The value for the specified custom field type must be a string"}
    ```
- Fields reorder after updating their values
- "Update" button stays disabled after request completed
- Form fields flicker after update.

ssangervasi (approved):
I'm approving this so we can can merge and test against the server updates. The FF-disabled state is good.

Issue #2435, opened 2023-01-30
New manual system creation page


allisonking:
### Is your feature request related to a specific problem?

Instead of our current form, let's replace filling in system info with tabs while also linking it to DSRs

### Describe the solution you'd like

https://github.com/ethyca/fides/issues/2434 will bring us to this page with some context about what type of system we are adding.

@TheAndrewJackson is currently building out components we can use for System information and Privacy declarations in  https://github.com/ethyca/fidesplus/issues/522, so for now, let's keep those blank until we can reuse those components.

We need to render 3 tabs:
* [ ] System Information - leave blank for now
* [ ] Privacy declarations - leave blank for now
* DSR configuration
  * [ ] Render connection information
  * [ ] Saving connection information should create a new saas connection config
  * [ ] Render datasets to connect or create a new dataset (yaml)
  * [ ] If no predefined system type was chosen, show a message that the connection needs to be configured

### Describe alternatives you've considered, if any

_A description of any alternative solutions or features you've considered._

### Additional context

Design: https://www.figma.com/file/dmEwdK3xZwjKfGVQ9t66xe/Fides-v.2-Master-Working-Files?node-id=2781%3A95251&t=F020IMrWRAfmtPLN-0

allisonking:
After talking to @rsilvery and @seanpreston , we are going to change the scope of this to:

* System information form tab
* Privacy declarations form tab

and punt on DSR while designs are still being finalized. We should do this in a way where the datamap UI can reuse these same components.

allisonking:
Talked to @rsilvery some more about this and how it will relate to https://github.com/ethyca/fidesplus/issues/598

We've got two places where we can input/edit system data:
* `/add-systems/new` (this ticket)
* `/system/configure` (https://github.com/ethyca/fidesplus/issues/598)

The problem is that they use the same components to render their content. The add systems flow has in the past rendered an abridged version of the configure flow (the idea being when you add, you just want to get the minimum down, and when you edit you can fill in more info)

We're going to continue that pattern here. This ticket should:

* totally redo the privacy declaration tab in both places
* update system info properties per new design: name, description, key
  * however, keep the extra fields while product/design figure out where those should go so that customers can still fill in those fields

That would make https://github.com/ethyca/fidesplus/issues/598 really only have to do the data flow section.

allisonking:
Changing scope again, turns out this was harder than I initially thought! privacy declarations will now be handled by https://github.com/ethyca/fidesplus/issues/598

PR #2542, opened 2023-02-08
More form input types

allisonking:
Needed for https://github.com/ethyca/fides/issues/2435

The new designs have a new type of input that looks like

image

whereas our existing inputs look like

image

This PR adds a new `variant` prop to a few of the custom form fields that can either be `stacked` or `inline` in order to support both input types.

### Code Changes

* Split form components into 3 common parts:
  * Label
  * Input
  * Error message
* Refactor all components to use at least the common Label and Error message
* Add `stacked` and `inline` variants. By splitting the components into subcomponents, it makes it easier to make the component return the same thing, just visually different
* Refactored `CustomSelect` and `CustomMultiSelect` to be one component that takes an `isMulti` option

### Steps to Confirm

* [ ] Click through various forms and make sure they still look the same as before. Nothing uses the new `stacked` variant yet, and the forms default to `inline`, so nothing should look different.

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### Description Of Changes

Inline vs stacked CustomTextInput
image

Inline vs stacked CustomTextArea
image

Inline vs stacked CustomSelect
image


clients/admin-ui/src/features/common/form/inputs.tsx
```diff
@@ -41,66 +54,208 @@ interface InputProps {
...
+
+  return (
+    <Select
+      options={options}
+      onBlur={(e) => {
+        setTouched({ ...touched, [field.name]: true });
+        field.onBlur(e);
+      }}
+      // @ts-ignore dynamic isMulti makes typing this complicated
+      onChange={isMulti ? handleChangeMulti : handleChangeSingle}
```
allisonking, line 164: could not figure out the typescript for this one :(

ssangervasi: Looks like it's possible to do it with generics and/or overloads, or split out the component call to another function so that `<Select isMulti={false}` and `<Select isMulti={true}` infer the exact type.

But I'd probably go with one of these:

```ts
 // Using `as` slightly better then ts-ignore. Technically unsafe, but only as unsafe
  // as react-select's runtime checking.
  const handleChange = (newValue: MultiValue<Option> | SingleValue<Option>) =>
    isMulti
      ? handleChangeMulti(newValue as MultiValue<Option>)
      : handleChangeSingle(newValue as SingleValue<Option>);

  // Adding type guards that also have runtime checks works. May be redundant.
  const handleChange = (newValue: MultiValue<Option> | SingleValue<Option>) =>
    narrow([{ label: "string", value: "string" }], newValue)
      ? handleChangeMulti(newValue)
      : narrow({ label: "string", value: "string" }, newValue)
      ? handleChangeSingle(newValue)
      : undefined; // Throw error?
```

And then `onChange={handleChange}`

clients/admin-ui/src/features/common/form/inputs.tsx
```diff
@@ -41,66 +54,208 @@ interface InputProps {
 type StringField = FieldHookConfig<string | undefined>;
 type StringArrayField = FieldHookConfig<string[] | undefined>;
 
+const Label = ({
+  title,
+  ...labelProps
+}: {
+  title: string;
```
ssangervasi, line 61: Making this an actual `children: ReactNode` could save us some work if/when we need build this out to match Chakra customizability.

clients/admin-ui/src/features/common/form/inputs.tsx
```diff
@@ -424,21 +505,62 @@ export const CustomTextArea = ({
...
-  return (
-    <FormControl isInvalid={isInvalid}>
-      {label ? (
+  // When there is no label, it doesn't matter if stacked or inline
+  // since we only render the text field
+  if (!label) {
+    return (
+      <FormControl isInvalid={isInvalid}>
+        <Flex>
+          {InnerTextArea}
```
ssangervasi, line 514: Looks like this was already named `InnerTextArea` but in general naming jsx variables with InitCaps can cause confusion because then it looks like it should be used like a component:`<InnerTextArea />`

clients/admin-ui/src/features/common/form/inputs.tsx
```diff
@@ -125,150 +280,64 @@ export const CustomSelect = ({
   isSearchable,
   isClearable,
   size = "sm",
+  isMulti,
+  variant = "inline",
   ...props
-}: SelectProps & StringField) => {
```
ssangervasi, line 129: I'm probably going to introduce merge conflicts for you: https://github.com/ethyca/fides/pull/2568/files#diff-f59681eb1681d3b25307f6abef05b10c5926bafe6543643fcb07a95e5cd4e55fR115
Whether we even need that comes down to the overrides convo: https://ethyca.slack.com/archives/C043URV1KLL/p1676054836626409

ssangervasi (approved):
I value your `Input`

clients/admin-ui/src/features/common/form/inputs.tsx
allisonking: okay I think I solved the merge conflicts, but you may want to keep an eye on it...

allisonking:
Failing test is unrelated

PR #2544, opened 2023-02-09
pc/format: Configure lint-staged and run prettier on api models

ssangervasi:
Looks like in https://github.com/ethyca/fides/pull/2341 the API models didn't git hit by prettier. `lint-staged` is great for covering all modified files, so I've added that here as well.

And yeah, these model files could just be prettier-ignored, but it looks like we're formatting them in the admin-ui too.

### DIFF
--- a/clients/privacy-center/package-lock.json
+++ b/clients/privacy-center/package-lock.json
@@ -55,6 +55,7 @@
         "eslint-plugin-react-hooks": "^4.3.0",
         "identity-obj-proxy": "^3.0.0",
         "jest": "^27.5.1",
+        "lint-staged": "^13.0.3",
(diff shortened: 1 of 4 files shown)


allisonking (approved):
I think it'd actually be great to remove types from prettier in both places. I tried to see if I could get the openapi:generate command to do double quotes in the past so we wouldn't have to worry about it but I didn't find the option. as [described here](https://github.com/ethyca/fides/discussions/2297#discussioncomment-4731678) adding types has become quite a pain for me, since I manually go in and grab the ones I want to avoid large diffs. but running that command generates large diffs because of the prettier/vs not prettier'd version

but anyway, this PR is good and would be great to be merged 😄

PR #2551, opened 2023-02-09
API for messaging-related settings

### LINKED ISSUE #2456: Set the default messaging provider based on what user selects in the UI: BE

adamsachs:
Closes #2456 

This enables the `PATCH /config` endpoint to support the following messaging-related config properties to be specified:

```
  "notifications": {
    "send_request_completion_notification": true,
    "send_request_receipt_notification": true,
    "send_request_review_notification": true,
    "notification_service_type": "string"
  },
  "execution": {
    "subject_identity_verification_required": true
  }
}
```
in addition to the
```
{
  "storage": {
    "active_default_storage_type": "s3"
  }
```
property that was already supported via this API.

The application now leverages the values set for these messaging-related properties, too, using a new `ConfigProxy` construct.

*Any API-set values for these config properties will take precedence over the values set via  "traditional" config mechanisms, i.e. `fides.toml` or ENV var.*

https://github.com/ethyca/fidesdocs/issues/56 has been created to track doc updates related to this and other functionality updates. 

### Code Changes

* updates `PATCH /config` endpoint to allow for updating messaging-related settings (i.e. `notifications` settings and `execution.subject_identity_verification_required` setting)
* creates a `ConfigProxy` construct that allows for config property resolution, which is required now that we are allowing API overwrites of config properties set via "traditional" config means.
    * update all references to the messaging-related config settings in application code to leverage the new ConfigProxy construct, so that API updates to the config properties are actually used in normal application runtime
    * For now, `ConfigProxy` only supports the config properties needed for storage and messaging use cases. Going forward, we may look to expand its use for more, or even _all_ of our config properties. It will likely need refactors to support that.
* updates tests accordingly

### Steps to Confirm

* `nox -s dev -- ui`
* use openAPI UI to set the messaging config settings via API, rather than updating their values in the `fides.toml` or via env var
* ensure sure messaging/notifications work as expected

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

-- Note --

I think I've got pretty good unit test coverage, and I've tested the functionality relatively thoroughly with a live app locally. But we could still used some improved end-to-end automated test coverage for the new functionality. What I am thinking is just leveraging some of our existing privacy request notification tests that use config property changes, but instead of updating the "traditional" config objects, have fixtures that update the API-set config objects instead.


src/fides/api/ops/models/application_config.py
```diff
@@ -82,3 +112,86 @@ def get_api_set_config(cls, db: Session) -> Dict[str, Any]:
...
+                    config_record.config_set, config_property, default_value
+                )
+            return api_prop
+        logger.warning("No config record found!")
+        return default_value
+
+
+def _get_property(
+    properties: Any, property_path: str = None, default_value: Any = None
+) -> Any:
```
sanders41, line 181: According to the instance check `properties` has to be a `dict`. Can we change the type from `Any` to `dict`? Also from my other comment, can we remove `default_value`?. If the answer to both of these is yes, and if the `properties` dict is a hashable dict we could add an LRU cache here so it only needs to do the recursive lookup one time.

```py
from functools import lru_cache

@lru_cache(maxsize=1028)  # I'm not sure what a good maxsize would be here. depends on how many different properties there could be.
def _get_property(properties: dict, property_path: str | None = None) -> dict:
    ...
```

Either way `property_path` should be `property_path: str | None = None` instead of just `str`

src/fides/api/ops/schemas/application_config.py
```diff
@@ -28,6 +30,46 @@ def validate_storage_type(cls, storage_type: StorageType) -> StorageType:
...
+    send_request_receipt_notification: Optional[bool]
+    send_request_review_notification: Optional[bool]
+    notification_service_type: Optional[str]
+
+    @validator("notification_service_type", pre=True)
+    @classmethod
+    def validate_notification_service_type(cls, value: Optional[str]) -> Optional[str]:
+        """Ensure the provided type is a valid value."""
+        if value:
+            valid_values = ["MAILGUN", "TWILIO_TEXT", "TWILIO_EMAIL"]
```
sanders41, line 55: Slightly better performance with a tuple since the valid values are immutable. 

```suggestion
            valid_values = ("MAILGUN", "TWILIO_TEXT", "TWILIO_EMAIL")
```

adamsachs:
test failures _look_ unrelated? some are just cancelled...?

tests/ops/models/test_application_config.py
```diff
@@ -83,3 +210,301 @@ def test_get_api_config(self, db: Session, example_config_record: Dict[str, Any]
...
+            db_config["notifications"]["send_request_completion_notification"]
+            == CONFIG.notifications.send_request_completion_notification
+        )
+        assert (
+            db_config["execution"]["masking_strict"] == CONFIG.execution.masking_strict
+        )
+        assert db_config["redis"]["port"] == CONFIG.redis.port
+        assert db_config["security"]["cors_origins"] == CONFIG.security.cors_origins
+        # and perform a general check on the config objects
+        assert dumps(db_config) == CONFIG.json()
```
sanders41, line 232: I think if `assert dumps(db_config) == CONFIG.json()` would cover all the previous asserts right? So you could remove the other asserts and only have this one?

tests/ops/models/test_application_config.py
```diff
@@ -83,3 +210,301 @@ def test_get_api_config(self, db: Session, example_config_record: Dict[str, Any]
...
+        assert (
+            db_config["notifications"]["send_request_completion_notification"]
+            == CONFIG.notifications.send_request_completion_notification
+        )
+        assert (
+            db_config["execution"]["masking_strict"] == CONFIG.execution.masking_strict
+        )
+        assert db_config["redis"]["port"] == CONFIG.redis.port
+        assert db_config["security"]["cors_origins"] == CONFIG.security.cors_origins
+        assert dumps(db_config) == CONFIG.json()
```
sanders41, line 268: Same as previous question, does `assert dumps(db_config) == CONFIG.json()` cover all the previous asserts?

sanders41:
> test failures look unrelated? some are just cancelled...?

I agree, unrelated. I think they will resolve on a re-run.

src/fides/api/ops/models/application_config.py
```diff
@@ -82,3 +112,86 @@ def get_api_set_config(cls, db: Session) -> Dict[str, Any]:
...
+        """
+        Utility method to set the `api_set` column on the `applicationconfig`
+        db record with the provided dictionary of data.
+
+        Updates are *merged* with any existing `api_set` data in the db.
+        """
+        return cls.create_or_update(db, data={"api_set": api_set_dict})
+
+    @classmethod
+    def set_config_set_config(
```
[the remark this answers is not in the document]
adamsachs: 😆 it is, though i definitely paused while writing it. i call the column `config_set` because it contains the config properties set by traditional "config" mechanisms. but i really hate the name. any suggestions?

even if i don't rename the column, i could rename this method. let me think about something a bit less of a mouthful.

src/fides/api/ops/models/application_config.py
```diff
@@ -82,3 +112,86 @@ def get_api_set_config(cls, db: Session) -> Dict[str, Any]:
...
+def _get_property(
+    properties: Any, property_path: str = None, default_value: Any = None
+) -> Any:
+    """
+    Internal helper to recursively traverse a property dict/value with a
+    dot-separated `property_path` string
+    """
+    if property_path:
+        if not isinstance(properties, dict):
+            raise ValueError("Invalid path specified")
```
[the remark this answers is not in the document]
adamsachs: well the idea here is just that if we have both a non-empty/null `property_path` _and_ `properties` isn't a `dict`, then  something's gone wrong. but i think you're right that it's not necessarily the `property_path`, it could also just be a bogus `properties` object, if that was given to the first call of the function. so i'll adjust the error to make it a bit less specific.

really no code should be getting here, since we're not exposing this function directly. but definitely want to make it as robust as possible!

src/fides/api/ops/models/application_config.py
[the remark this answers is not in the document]
sanders41: Now I see your struggle, the other name that comes to mind is `set_set_config` which is no better :smile:

src/fides/api/ops/models/application_config.py
```diff
@@ -82,3 +112,86 @@ def get_api_set_config(cls, db: Session) -> Dict[str, Any]:
...
+        path_split = property_path.split(".", maxsplit=1)
+        index = path_split.pop(0)
+        try:
+            return _get_property(
+                properties[index], path_split[0] if path_split else None
+            )
+        except KeyError:
+            return default_value
```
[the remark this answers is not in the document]
adamsachs: i'm catching the possible `KeyError` with the `properties[index]` lookup, which is possible. in that case, we're essentially looking up a non-existent property, which can happen if e.g. it hasn't been set by API. and in that case, we want to return a default value.

i just put `default_value` as an arg in case someone wants to use this function for another use later. all our uses now just leverage the default `None` but maybe there'd be another use for the `get_resolved_config_property` going forward.

src/fides/api/ops/models/application_config.py
sanders41: Based on your above descriptions it made me realize a cache won't work here because the value of the dictionary could be an object which is fine. So you can ignore the `lru_cache` suggestion.

So now the remaining questions are should `properties` be `dict` (with more specific types included if possible) and is `Any` correct for `default_values`?

Then `property_path` should be `property_path: str | None = None`

adamsachs: agreed that we could make `property_path: str | None = None`, but `properties` doesn't actually have to be a `dict`. when it gets to the end of the recursion, that will be the returned value, which isn't necessarily a `dict`. the instance check is only if `property_path` isn't `None`. 

i agree there may be ways to optimize - with an lru_cache or otherwise, both to make the code more efficient and more readable/reusable. but maybe we could take some of those up in a follow-up?  i'm sure we'll be taking a look again soon enough at this class to expand its use :) 

unless you think we may get unexpected behavior with how it is now.

tests/ops/models/test_application_config.py
adamsachs: ha, good call. that looks kinda silly in retrospect :)

tests/ops/models/test_application_config.py
adamsachs: yeah, i think what i was trying to do here was make sure we don't have any deserialization issues? but that's probably not needed, may be dumb.

src/fides/api/ops/models/application_config.py
sanders41: Got it, that makes sense. I agree, lets go with it as is now and can optimize later if needed.

pattisdr:
Reviewing now for the purposes of thinking about extending this in fidesplus

sanders41 (approved):

src/fides/api/ops/api/v1/endpoints/config_endpoints.py
```diff
@@ -33,7 +33,7 @@ def get_config(
     logger.info("Getting the exposable Fides configuration")
     if api_set:
         logger.info("Retrieving api-set application settings")
-        return ApplicationConfig.get_api_set_config(db)
+        return ApplicationConfig.get_api_set(db)
```
pattisdr, line 36: Do we also need to censor sensitive config that was set in the API?

src/fides/api/ops/api/v1/endpoints/config_endpoints.py
```diff
@@ -33,7 +33,7 @@ def get_config(
...
     config = censor_config(get_app_config())
```
pattisdr, line 37: This is pulling the config in the more traditional way instead of from the database which is fine, this is just making me think more about the dangers of having two sources of truth when something gets out of sync.

src/fides/api/ops/models/application_config.py
```diff
@@ -82,3 +112,84 @@ def get_api_set_config(cls, db: Session) -> Dict[str, Any]:
...
+        """
+        Utility method to set the `config_set` column on the `applicationconfig`
+        db record by serializing the given `FidesConfig` to a JSON blob.
+        """
+        # need to do this deserialization round-trip to get us a dict
+        # that's serializable as JSON in the db:
+        # pydantic's `.dict()` is NOT serializable as JSON --
+        # see https://github.com/pydantic/pydantic/issues/1409
+        config_dict = loads(config.json())
+        return cls.create_or_update(db, data={"config_set": config_dict})
```
pattisdr, line 151: note to self this is just called on startup and in tests, and it would be nice if we could do away with this -

src/fides/core/config/config_proxy.py
```diff
@@ -0,0 +1,86 @@
...
+    subject_identity_verification_required: bool
+
+
+class StorageSettingsProxy(ConfigProxyBase):
+    prefix = "storage"
+
+    active_default_storage_type: StorageType
+
+
+class ConfigProxy:
```
pattisdr, line 62: Understand we can't refactor everything at once, but I think we need to make sure that there's good communication around this in that some things should be loaded from the ConfigProxy and some should be loaded from our traditional CONFIG.  

A nice state to aim for might be where we load from the ConfigProxy directly everywhere, and only the ConfigProxy itself (and supporting code) loads the traditional CONFIG resources and reconciles

src/fides/core/config/config_proxy.py
```diff
@@ -0,0 +1,86 @@
...
+class ExecutionSettingsProxy(ConfigProxyBase):
+    prefix = "execution"
+
+    subject_identity_verification_required: bool
+
+
+class StorageSettingsProxy(ConfigProxyBase):
+    prefix = "storage"
+
+    active_default_storage_type: StorageType
```
pattisdr, line 59: This also feels like more duplication here, we have to add every piece of data that's a ConfigProxy here.  So on the plus side we couldn't use the config proxy without adding onto it -

src/fides/api/ops/models/application_config.py
```diff
@@ -82,3 +112,84 @@ def get_api_set_config(cls, db: Session) -> Dict[str, Any]:
...
+        if not isinstance(properties, dict):
+            raise ValueError("Invalid property lookup")
+        path_split = property_path.split(".", maxsplit=1)
+        index = path_split.pop(0)
+        try:
+            return _get_property(
+                properties[index], path_split[0] if path_split else None
+            )
+        except KeyError:
+            return default_value
```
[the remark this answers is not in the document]
adamsachs: ah yes, thank you for pointing this out @pattisdr !!! i'd thought i'd remembered seeing a library for this somewhere in our codebase but couldn't find it.

src/fides/core/config/config_proxy.py
adamsachs: agreed, i think that's where i see this going eventually - i was just reluctant to make that big of a code change when trying to get this first increment across.

src/fides/api/ops/models/application_config.py
```diff
@@ -36,6 +39,18 @@ class ApplicationConfig(Base):
...
+            StringEncryptedType(
+                JSONTypeOverride,
+                CONFIG.security.app_encryption_key,
+                AesGcmEngine,
+                "pkcs5",
+            )
+        ),
+        nullable=False,
+        default={},
+    )  # store as encrypted JSON blob since config settings may have sensitive data
```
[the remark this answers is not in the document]
adamsachs: yeah - i think you're right to question it. my motivations were broadly around:

- trying not to assume too much about _how_ we resolve config properties at runtime, since i think there's still some question about the best way to do that. i.e. it felt like having both config records (i.e. api-set and config-set) in the db would make it simpler for us to update, moving forward, how we resolve between the two; or to allow someone to resolve in an alternative manner.
- trying to reduce dependency complexity, i.e. when thinking about the config proxy, i thought it simplified things to _just_ have it depend on the db record, to avoid potential conflicts in depending on both the `CONFIG` and the db, while the db itself depends on the `CONFIG`.

but looking back at these concerns, knowing where we've ended up with it, i don't really think storing the config-set in the db gains us a whole lot, and i think you're right to point out that it may be burdensome and a source of synchronization problems moving forward. additionally, i think it'd be easier to add this back in later if we do see a good utility for it, whereas it'd be a bit harder to take it out. so let's give ti a try!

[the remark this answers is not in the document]
pattisdr: hey @adamsachs after this code review, I started to think more about the plus side, and I've changed my stance here!  I think your original plan was spot-on from the perspective of being able to re-use in Plus. The fact that you're storing both the api-set config and the traditionally set config in the database should make it easier to pull from one source in plus, regardless of whether that traditional config was a FidesConfig object or a FidesplusConfig object. 

And I was making a big deal about the "dangers of things getting out of sync" but that seems overblown in retrospect: it is largely just tests where we're updating the traditional config after the fact.


[Here's my thoughts on the plus side](https://ethyca.atlassian.net/wiki/spaces/EN/pages/2615083278/2023-02-10+Overriding+Plus+Config+without+Re-Deploying), but I think if both the api config and typical config are stored in the database as you describe, it should be fairly straightforward.

src/fides/api/ops/models/application_config.py
```diff
@@ -82,3 +112,84 @@ def get_api_set_config(cls, db: Session) -> Dict[str, Any]:
...
+
+        Api-set values get priority over config-set, in case of conflict.
+        """
+        config_record = db.query(cls).first()
+        if config_record:
+            api_prop = _get_property(config_record.api_set, config_property)
+            if api_prop is None:
+                logger.info(f"No API-set {config_property} property found")
+                return _get_property(
+                    config_record.config_set, config_property, default_value
```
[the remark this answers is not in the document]
pattisdr: Changed my opinion here, my comment wouldn't be as reusuable then from a plus perspective.  The original plan of pulling this from the database makes it easier to use across both repos.

src/fides/api/ops/api/v1/endpoints/config_endpoints.py
pattisdr: Changed my mind, we're largely only changing the config in tests though, it's not a big deal.

src/fides/api/ops/models/application_config.py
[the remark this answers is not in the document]
adamsachs: ha, i guess i should trust my gut a bit more! 😄 we'll keep it as-is then, thanks for taking the time to think through this and consider how we'll extend it for your use case!

ThomasLaPiana:
@adamsachs I'm totally just rubber ducking here so forgive the potentially dumb question, but did you try this using a `CONFIG` object that is shared everywhere as opposed to our current methodology? I need to figure out a way to test it, but it would look something like:

```
In [1]: from fides.core.config import get_config
Loaded config from: .fides/fides.toml

In [2]: CONFIG = get_config()

In [3]: CONFIG.dev_mode
Out[3]: False

In [4]: CONFIG.dev_mode = True

In [5]: CONFIG = get_config()

In [6]: CONFIG.dev_mode
Out[6]: True
```

`CONFIG` is referring to the same object so it preserves settings changes. Although given the parent/child fides instance stuff maybe it needs to be in a database so that it is accessible from outside of that single python application?

src/fides/core/config/config_proxy.py
```diff
@@ -0,0 +1,86 @@
...
+
+    To instantiate a `ConfigProxy`, you must have an active db `Session`,
+    since the `ConfigProxy` will rely on the db for property resolution.
+
+    Instantiating a `ConfigProxy` is itself a cheap operation,
+    i.e. there is minimal resource overhead with instantiating a new `ConfigProxy`.
+    The `ConfigProxy` is a thin wrapper above the given db `Session` -
+    it relies on the given db `Session` and for all of its state.
+
+    Lookups (i.e. attribute access) with the `ConfigProxy` do leverage
```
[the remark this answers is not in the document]
adamsachs: actually i meant `do` here! :)

src/fides/api/ops/models/application_config.py
```diff
@@ -82,3 +112,84 @@ def get_api_set_config(cls, db: Session) -> Dict[str, Any]:
...
+    def get_config_set(cls, db: Session) -> Dict[str, Any]:
+        """
+        Utility method to get the config_set config dict
+
+        An empty `dict` will be returned if no config settings have been set through config
+        """
+        config_record = db.query(cls).first()
+        if config_record:
+            return config_record.config_set
+        return {}
```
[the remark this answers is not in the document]
adamsachs: that's true, but figure it's still a useful utility to have around?

src/fides/api/ops/models/application_config.py
```diff
@@ -82,3 +112,84 @@ def get_api_set_config(cls, db: Session) -> Dict[str, Any]:
...
+    def update_api_set(
+        cls, db: Session, api_set_dict: Dict[str, Any]
+    ) -> ApplicationConfig:
+        """
+        Utility method to set the `api_set` column on the `applicationconfig`
+        db record with the provided dictionary of data.
+
+        Updates are *merged* with any existing `api_set` data in the db.
+        """
+        return cls.create_or_update(db, data={"api_set": api_set_dict})
```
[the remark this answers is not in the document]
adamsachs: updated in latest commits - please confirm!

tests/ops/api/v1/endpoints/test_privacy_request_endpoints.py
```diff
@@ -1790,10 +1791,11 @@ def url(self, db, privacy_request):
...
 
     @pytest.fixture(scope="function")
-    def privacy_request_review_notification_enabled(self):
+    def privacy_request_review_notification_enabled(self, db):
         """Enable request review notification"""
         original_value = CONFIG.notifications.send_request_review_notification
         CONFIG.notifications.send_request_review_notification = True
+        ApplicationConfig.update_config_set(db, CONFIG)
         yield
         CONFIG.notifications.send_request_review_notification = original_value
```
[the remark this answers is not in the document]
adamsachs: updated in latest commits - please confirm!

src/fides/api/ops/schemas/application_config.py
```diff
@@ -28,6 +30,46 @@ def validate_storage_type(cls, storage_type: StorageType) -> StorageType:
...
+        if value:
+            valid_values = ("MAILGUN", "TWILIO_TEXT", "TWILIO_EMAIL")
+            value = value.upper()  # force uppercase for safety
+
+            if value not in valid_values:
+                raise ValueError(
+                    f"Invalid NOTIFICATION_SERVICE_TYPE provided '{value}', must be one of: {', '.join([level for level in valid_values])}"
+                )
+
+        return value
```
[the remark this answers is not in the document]
adamsachs: updated in latest commits - please confirm!

src/fides/api/ops/schemas/application_config.py
```diff
@@ -28,6 +30,46 @@ def validate_storage_type(cls, storage_type: StorageType) -> StorageType:
         return storage_type
```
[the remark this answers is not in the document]
adamsachs: updated in latest commits - please confirm!

src/fides/api/ops/models/application_config.py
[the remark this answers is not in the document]
adamsachs: updated in latest commits - please confirm!

src/fides/api/ops/api/v1/endpoints/config_endpoints.py
adamsachs: updated in latest commits - please confirm!

adamsachs:
@pattisdr i think i've addressed all your comments in my most recent comments, if you could take another quick look to confirm things look good from your end that'd be great!

otherwise, i think this is just about ready to go, i've gotten `main` merged in.

@ThomasLaPiana for
>  but did you try this using a CONFIG object that is shared everywhere as opposed to our current methodology? 
i'm not sure i'm totally following the suggestion - maybe we can sync up quickly offline tomorrow? 

i do want to try and get this PR across the line as it's holding up the system config work. for what it's worth, this was just meant as an initial iteration on the idea of a config proxy and providing a mechanism for mutable state in our config. i am expecting we'll need to refine some of this as we go, so unless we feel this is going down a path that will be hard to work from moving forward, then i think we should get this in place as a first iteration.

adamsachs:
the only cocdecov miss here is for some exception handling within `main.py`, but it's pretty difficult for us to cover `main.py` execution itself in automated tests, and we're missing a lot of coverage there already -- so i don't think that should hold this up. curious for other thoughts.

sanders41:
> the only cocdecov miss here is for some exception handling within main.py, but it's pretty difficult for us to cover main.py execution itself in automated tests, and we're missing a lot of coverage there already -- so i don't think that should hold this up. curious for other thoughts.

Overall coverage stays the same, and that exception catches all exceptions not specific ones so I'd be ok with missing it. We can follow up later and look at main.py as a whole.

pattisdr:
Starting second pass shortly...

src/fides/api/ops/models/application_config.py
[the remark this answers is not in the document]
pattisdr: Sure, just sometimes it causes false confidence when someone sees that method, and assumes "this is how we load the `config_set` variable in our application, when we don't in fact use it anywhere. 

 For example, I saw this method and assumed you would use it in the the `get_config` endpoint where `api_set`=False and/or in `ApplicationConfig.get_resolved_config_property`.  

We just have multiple ways to get `config_set` right now, that's all.

src/fides/api/ops/models/application_config.py
```diff
@@ -82,3 +113,61 @@ def get_api_set_config(cls, db: Session) -> Dict[str, Any]:
...
+
+        An empty `dict` will be returned if no config settings have been set through config
+        """
+        config_record = db.query(cls).first()
+        if config_record:
+            return config_record.config_set
+        return {}
+
+    @classmethod
+    def update_api_set(
```
pattisdr, line 130: Is there any way to reset "api_set"? That might be a nice feature - to just remove everything set by the API and revert to the default.  I found myself wanting to do that in testing.

src/fides/core/config/__init__.py
```diff
@@ -93,18 +93,22 @@ def log_all_config_values(self) -> None:
...
     for key, value in CONFIG_KEY_ALLOWLIST.items():
-        data = as_dict[key]
-        filtered[key] = {}
-        for field in value:
-            filtered[key][field] = data[field]
+        if key in as_dict:
+            data = as_dict[key]
+            filtered[key] = {}
+            for field in value:
+                filtered[key][field] = data[field]
```
pattisdr, line 111: I think this would break if you only set one of the subsettings - for example, I had only updated notifications.notification_service_type via the API:
```json
{"notifications": {
    "notification_service_type": "twilio_email"
}}
```

You might add another check, like:
```
                if field in data:
                    filtered[key][field] = data[field]
```
or look into defaultdict, I use that too.

src/fides/api/ops/models/application_config.py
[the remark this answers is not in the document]
pattisdr: looks good. pydash is great for this use case.

src/fides/api/ops/schemas/application_config.py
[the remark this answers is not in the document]
pattisdr: looks good

src/fides/api/ops/models/application_config.py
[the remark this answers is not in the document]
pattisdr: your original design was very forward thinking with Plus!

pattisdr (commented):
Thanks for this round of changes @adamsachs, looking good, just one small bug found -

src/fides/api/ctl/migrations/versions/c9ee230fa6da_add_config_set_column_to_application_.py
```diff
@@ -0,0 +1,36 @@
...
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column(
+        "applicationconfig",
+        sa.Column(
+            "config_set",
+            sqlalchemy_utils.types.encrypted.encrypted_type.StringEncryptedType(),
+            nullable=False,
+        ),
+    )
```
pattisdr, line 28: @adamsachs Migration may need to be updated while you're here? - it looks like `api_set` is nullable=True in database, but nullable=False on the model.

adamsachs: goot catch! i'd just noticed this myself locally when looking into the weird `ctl-not-external` CI timeouts (unrelated to this i believe). but i'll make the update since yes, i'd missed this in my previous PR for the storage work.

pattisdr: ah yeah just noticed it on another branch! thank you!

src/fides/api/ops/models/application_config.py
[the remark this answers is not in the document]
adamsachs: @pattisdr let me know if i'm good to mark this as resolved, i think i'd addressed this!

src/fides/api/ops/models/application_config.py
adamsachs: @pattisdr  great idea, i also felt that need a bit - seemed simple enough to add here so i added in https://github.com/ethyca/fides/pull/2551/commits/b053049e474d08de4dfc2e72a738d6d7fabf0135, let me know what you think

src/fides/api/ctl/migrations/versions/c9ee230fa6da_add_config_set_column_to_application_.py
adamsachs: added in https://github.com/ethyca/fides/pull/2551/commits/e890bf28dbc7bc10cdb4cd1708a943edc6c57155

src/fides/core/config/__init__.py
adamsachs: fixed in https://github.com/ethyca/fides/pull/2551/commits/48f0394497cb0a6ad9717a5a128080aff775db0c

src/fides/api/ops/models/application_config.py
pattisdr: looks great thank you

pattisdr (approved):
well done 🏆

PR #2572, opened 2023-02-10
default messaging config APIs 

### LINKED ISSUE #2454: Set the values for messaging provider based on what user selects for messaging configuration: BE
### Description
In This [ticket](https://github.com/ethyca/fides/issues/2236), we planned to only work on the UI to configure default messaging provider. The default value propagation to the backend was decided to be a separate work after this [ticket](https://github.com/ethyca/fides/issues/1974) is completed.

We aim to set the default messaging value in the backend.

### Acceptance Criteria
1. When a user selects a default configuration value in the front end, the same value is used by Fides for running DSR requests.

adamsachs:
Closes #2454

`/messaging/default` endpoints for system config work.

generally the "default" endpoints added here have similar functionality to existing messaging configuration endpoints, except they do not require the caller to worry about the `name` or `key` of the messaging configuration - just the `service_type`. this provides a simplified API to update the messaging configurations of given types, and keeps it consistent with recently added `/storage/default` configuration endpoints.

since we already had a constraint that allowed just a single messaging configuration of a given type at any time on the system, the code and application logic changes here are relatively minimal - we're really just providing a slightly different API to update these artifacts.

note that existing messaging config endpoints maintain their functionality, to ensure this change keeps backward compatibility. those existing endpoints are effectively just now another mechanism for updating default messaging configurations, that allow a bit more control -- i.e. specifying the `name` and `key` of messaging configs in CRUD operations.

https://github.com/ethyca/fidesdocs/issues/56 has been created to track doc updates related to this and other related functionality updates.

### Code Changes

* `PUT /messaging/default` endpoint to upsert the messaging config details for a messaging config of the specified type (`service_type` is specified here as a field in the request payload)
* `PUT /messaging/default/{service_type}/secret` to upsert the messaging config secrets for a messaging config of the specified type
* `GET /messaging/default/{service_type}` endpoint to retrieve the messaging config details for a messaging config of the specified type
* `GET /messaging/active/default` endpoint to retrieve the messaging config details for the messaging configuration that's the "active default", i.e. the messaging configuration that will actually be used in privacy request execution
    * the default config that's returned here is based on the current `notifications.notification_service_type` app config setting, which can either be updated via traditional configuration mechanisms (env var, `fides.toml`) or via API, which will take higher precedence (as implemented in #2551)

### Steps to Confirm

* spin up a fides environment
* use swagger API UI to interact with the new endpoints
* an end-to-end test would consist of:
     * set up at least one messaging provider outside of fides, as described in our [docs](https://docs.ethyca.com/fides/configuration/messaging_services#prerequisites)
     * setting up at least one corresponding default messaging config in fides using the above endpoints
     * ensure the `notifications.notification_service_type` property is properly set to make the desired messaging config the "active default".
     * ensure one of the "send notification" config properties is set to `true`, e.g. `notifications.send_request_review_notification`, and/or the `execution.subject_identity_verification_required` setting
     * execute a privacy request and ensure that messages/notifications are being sent to the provided identity email/phone number

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes


src/fides/api/ops/api/v1/endpoints/messaging_endpoints.py
```diff
@@ -135,6 +146,63 @@ def patch_config_by_key(
...
+    """
+    logger.info(
+        "Starting upsert for default messaging config of type '{}'",
+        messaging_config.service_type,
+    )
+    incoming_data = messaging_config.dict()
+    existing_default = MessagingConfig.get_by_type(db, messaging_config.service_type)
+    if existing_default:
+        # take the key of the existing default and add that to the incoming data, to ensure we overwrite the same record
+        incoming_data["key"] = existing_default.key
```
eastandwestwind, line 170: even if we have an existing default, we generate the same key / name if it's the same service, right? So this code is just to be extra safe?

src/fides/api/ops/api/v1/endpoints/messaging_endpoints.py
```diff
@@ -231,6 +308,47 @@ def get_config_by_key(
...
+)
+def get_default_config_by_type(
+    service_type: MessagingServiceType, *, db: Session = Depends(deps.get_db)
+) -> MessagingConfig:
+    """
+    Retrieves default config for messaging service by type.
+    """
+    logger.info("Finding default messaging config of type '{}'", service_type)
+
+    messaging_config = MessagingConfig.get_by_type(db, service_type)
```
eastandwestwind, line 324: I think this is confusing because when we add a "default", we use the `Default Messaging Config` pre-fix to denote that.  This `GET` will get any message config by type, not just the "default", right?

src/fides/api/ops/models/messaging.py
```diff
@@ -144,3 +158,44 @@ def set_secrets(
 
         self.secrets = messaging_secrets
         self.save(db=db)
+
+    @classmethod
+    def get_active_default(cls, db: Session) -> Optional[MessagingConfig]:
```
eastandwestwind, line 163: this is great to have, and I think something we can re-use throughout the codebase to help validate configs prior to attempting actual messaging sends during run-time

eastandwestwind (commented):
Overall this looks good, @adamsachs , but I think I still have some confusion around "default" terminology. 

For example, when we `GET` the active default config, we're simply getting the configured messaging service. I think you've mentioned it in passing, but could you explain to me again how we're anticipating using the "default" config?

adamsachs:
> but I think I still have some confusion around "default" terminology

@eastandwestwind yeah, i need to clean up the existing/make a new confluence document to more clearly outline where we've landed on all of this. let me try to explain it a bit more here, and we can sync up offline as well if it's not clear.

essentially, this change is just to provide an interface for updating/retrieving messaging details that is similar to what was implemented for storage in #2438 - for consistency in our APIs. for storage, we actually allow for more than one storage configuration per storage type, so defining one of those configurations as a "default" was a more substantive change in application logic. for messaging though, we already only allow one messaging configuration per messaging service type, so the changes here are just providing a new interface that's styled similarly to what's there for default storage configurations. but in a certain sense, we've always had "default" messaging configurations in place, since we'd always prevented users from having more than one messaging configuration for a given service type.

the other concept here is around an "active default". the active default is the _single_ messaging configuration that's actually used in privacy request execution - and that's determined by the global application config property `notifications.notification_service_type`. you must also of course have a default messaging configuration in place for the given service type. you can also have default messaging configurations for _other_ service types, but only one of them is an "active default" at any given point in time.

this [document](https://ethyca.atlassian.net/wiki/spaces/EN/pages/2605449283/System+Config+API+DRAFT) provides the API spec across messaging and storage for the system config work, as used by the UI. it may be useful context to see what the APIs look like, holistically.

src/fides/api/ops/api/v1/endpoints/messaging_endpoints.py
adamsachs: yeah, generally that should be the case. but to keep backward compatibility, the old messaging endpoints (that allow for name/key specification) are still available, and any messaging configs created by these means will also be exposed as a "default". effectively, all messaging configs are defaults now :)

src/fides/api/ops/models/messaging.py
adamsachs: oh good point. so far i'd only updated the app code to use the `config_proxy` to determine the notification service type, e.g. https://github.com/ethyca/fides/blob/main/src/fides/api/ops/service/_verification.py#L37, but i think maybe it's a better idea to rely on this utility method in all places, as i think you're suggesting. just consolidates our code a bit...

let me know if that's what you're suggesting

eastandwestwind:
Thanks for the clarification @adamsachs ! 

To summarize and test my knowledge:

A default refers to the default config for any given service type. In the case of messaging configs, all configs will be defaults, but only one will be "active". For storage types, however, some configs may not be defaults due to the fact that we can set up multiple configs per service type. Again, only one config will be "active" for storage.

Can you link the documentation update ticket here, just so we're tracking it?

eastandwestwind:
One last thing- can we follow-up with a ticket to add these new api endpoints into Postman?

eastandwestwind (approved):

src/fides/api/ops/models/messaging.py
eastandwestwind: We can follow-up with this improvement. Can you create a new tech-debt ticket for using this utility method where we can?

src/fides/api/ops/api/v1/endpoints/messaging_endpoints.py
eastandwestwind: Yes your comments helped clarify the meaning of "default" overall. Thanks!

adamsachs:
> A default refers to the default config for any given service type. In the case of messaging configs, all configs will be defaults, but only one will be "active". For storage types, however, some configs may not be defaults due to the fact that we can set up multiple configs per service type. Again, only one config will be "active" for storage.

yup that's exactly right @eastandwestwind.

> Can you link the documentation update ticket here, just so we're tracking it?
here's the doc ticket i'd created: https://github.com/ethyca/fidesdocs/issues/56

> One last thing- can we follow-up with a ticket to add these new api endpoints into Postman?
created https://github.com/ethyca/fides/issues/2662

src/fides/api/ops/models/messaging.py
adamsachs: created https://github.com/ethyca/fides/issues/2663

PR #2586, opened 2023-02-14
ui/custom-metadata: Refactor taxonomy to use a single form submission flow

### LINKED ISSUE #2545: Custom metadata UI bug roundup
Results of PR smoke test: https://github.com/ethyca/fides/pull/2536#issuecomment-1423341630

**Main checks**
- [x] No errors seemed to be the API's fault. (No 500; validation errors were correct.)
- [x] With the feature flag (`customFields`) disabled, editing taxonomy entities still works.

**Functionality status**
- Defining custom lists
  - [x] Create
  - [x] Use for a field definition
- Defining custom fields
  - [x] Create single select
  - [x] Create multi select
  - [x] Cancel button warns about discarding changes
- [ ] "Show in forms and reports"
  - How do we verify this?
- Using single select
  - [x] Initial selection
  - [ ] Removing selection
  - [ ] Changing selection
  - [ ] Initial load when data has been saved
- Using multi select
  - [x] Initial selection
  - [ ] Removing selection
  - [ ] Changing selection
  - [ ] Initial load when data has been saved
- [ ] When there are multiple metadata fields
- [ ] Switching between taxonomy entities
- [x] Switching between resource types
  - Does seem to reset the form correctly

**Bugs**
- On initial load the spinners don't go away
- The requests 404 because no fields have been added to a category
  - The API is correct to return 404
  - The UI needs to understand the "nothing exists yet" case
  - Probably causes the spinners problem
- Clicking "edit" on another category doesn't update the custom fields form
  - Inconsistent
  - Might only happen if you made change in the form
  - Wait, unless custom fields are supposed to show up for all data categories? In which case the problem is that the form sometimes forgets which fields to show.
- Removing a item from the field select list doesn't remove the item
  - But if you "Update entity" and refresh the page it will have been saved
- Clicking "save" doesn't close the add field modal
- Clicking "save" doesn't refresh the custom field form
- Having two fields backed by the same select list causes the form fields to get mixed up
  - This should be fine. They should have the same list of choices, but the values should be independent.
  - Actually, it might not matter which custom list the fields use. Values seem to attach to the wrong fields sometimes regardless.
  - Can result in API error response because a single select was passed an array or a multi was passed a string
    - http://localhost:3000/api/v1/plus/custom-metadata/custom-field
    ```json
    {"custom_field_definition_id":"plu_a5f493f8-cbb7-435d-a94c-f28ec281f4c1","id":"plu_5ee31d1d-7ed2-44fe-9cfb-6970295f0d39","resource_id":"system.operations","value":["a"]}
    ```
    ```json
    {"detail":"The value for the specified custom field type must be a string"}
    ```
- Fields reorder after updating their values
- "Update" button stays disabled after request completed
- Form fields flicker after update.
- Can't update the values of any fields after initially setting them (couldn't tell if this was captured elsewhere here)

ssangervasi:
Closes #2545

### Code Changes

* This is a large refactor to get taxonomy and custom fields working together. See description below and my line comments.

### Steps to Confirm

* [ ] Cypress tests!
* [ ] From various taxonomy tabs
    * [ ] Assign custom fields
    * [ ] Single-select
        * [ ] Initial selection
        * [ ] Change selection
        * [ ] Clear selection
        * [ ] See value on reload
    * [ ] Multi-select
        * [ ] Initial selection
        * [ ] Change selection
        * [ ] Clear selection
        * [ ] See value on reload
* [ ] Those steps, but from the "Describe system" tab

### Pre-Merge Checklist

* Documentation:
    - Might be spun out: https://github.com/ethyca/fides/issues/2545#issuecomment-1429021292
    - https://github.com/ethyca/fides/issues/2612
* [ ] Update `CHANGELOG.md`

### Description Of Changes

* [x] I'm tired. When I finish my self-review I'll check this box.

### DIFF
--- a/clients/admin-ui/cypress/e2e/taxonomy-plus.cy.ts
+++ b/clients/admin-ui/cypress/e2e/taxonomy-plus.cy.ts
@@ -10,22 +10,47 @@ describe("Taxonomy management with Plus features", () => {
     cy.visit("/taxonomy");
   });
 
+  const RESOURCE_TYPE = {
+    label: "Data Categories",
+    key: ResourceTypes.DATA_CATEGORY,
+  };
+  const RESOURCE_PARENT = {
+    label: "User Data",
+    key: "user",
+  };
+  const RESOURCE_CHILD = {
+    label: "Biometric Data",
+    key: "user.biometric",
+  };
+
   const navigateToEditor = () => {
-    cy.getByTestId("accordion-item-User Data").click();
-    cy.getByTestId("item-Biometric Data")
+    cy.getByTestId(`accordion-item-${RESOURCE_PARENT.label}`).click();
+    cy.getByTestId(`item-${RESOURCE_CHILD.label}`)
       .click()
       .within(() => {
         cy.getByTestId("edit-btn").click();
       });
   };
 
+  // TODO: Extract these to a cypress support file.
+  const getSelectValueContainer = (selectorId: string) =>
+    cy.getByTestId(selectorId).find(`.custom-select__value-container`);
+
+  const getSelectOptionList = (selectorId: string) =>
+    cy.getByTestId(selectorId).click().find(`.custom-select__menu-list`);
+
   const selectOption = (selectorId: string, optionText: string) => {
-    cy.getByTestId(selectorId)
-      .click()
-      .within(() => {
-        cy.contains(optionText).click();
-      });
+    getSelectOptionList(selectorId).contains(optionText).click();
   };
+
+  const removeMultiValue = (selectorId: string, optionText: string) =>
+    getSelectValueContainer(selectorId)
+      .contains(optionText)
+      .siblings(".custom-select__multi-value__remove")
+      .click();
+
+  const clearSingleValue = (selectorId: string) =>
+    cy.getByTestId(selectorId).find(".custom-select__clear-indicator").click();
 
   describe("Defining custom lists", () => {
     beforeEach(() => {
@@ -137,7 +162,7 @@ describe("Taxonomy management with Plus features", () => {
         active: true,
         field_type: "string[]",
         name: "Multi-select",
-        resource_type: ResourceTypes.DATA_CATEGORY,
+        resource_type: RESOURCE_TYPE.key,
       });
     });
 
@@ -146,6 +171,108 @@ describe("Taxonomy management with Plus features", () => {
       cy.getByTestId("create-custom-fields-form").should(
         "contain",
         "Name is required"
+      );
+    });
+  });
+
+  describe("Using custom fields", () => {
+    beforeEach(() => {
+      cy.intercept(
+        {
+          method: "GET",
+          pathname: "/api/v1/plus/custom-metadata/allow-list",
+          query: {
+            show_values: "true",
+          },
+        },
+        {
+          fixture: "taxonomy/custom-metadata/allow-list/list.json",
+        }
+      ).as("getAllowLists");
+      cy.intercept(
+        "GET",
+        // Cypress route matching doesn't escape special characters (whitespace).
+        `/api/v1/plus/custom-metadata/custom-field-definition/resource-type/${encodeURIComponent(
+          RESOURCE_TYPE.key
+        )}`,
+
+        {
+          fixture: "taxonomy/custom-metadata/custom-field-definition/list.json",
+        }
+      ).as("getCustomFieldDefinitions");
+      cy.intercept(
+        "GET",
+        `/api/v1/plus/custom-metadata/custom-field/resource/${RESOURCE_CHILD.key}`,
+        {
+          fixture: "taxonomy/custom-metadata/custom-field/list.json",
+        }
+      ).as("getCustomFields");
+
+      navigateToEditor();
+
+      cy.wait([
+        "@getAllowLists",
+        "@getCustomFieldDefinitions",
+        "@getCustomFields",
+      ]);
+    });
+
+    const testIdSingle =
+      "input-definitionIdToCustomFieldValue.id-custom-field-definition-starter-pokemon";
+    const testIdMulti =
+      "input-definitionIdToCustomFieldValue.id-custom-field-definition-pokemon-party";
+
+    it("initializes form fields with values returned by the API", () => {
+      cy.getByTestId("custom-fields-list");
+      getSelectValueContainer(testIdSingle).contains("Squirtle");
+
+      ["Charmander", "Eevee", "Snorlax"].forEach((value) => {
+        getSelectValueContainer(testIdMulti).contains(value);
+      });
+    });
+
+    it("allows choosing and changing selections", () => {
+      cy.getByTestId("custom-fields-list");
+
+      clearSingleValue(testIdSingle);
+      selectOption(testIdSingle, "Snorlax");
+      getSelectValueContainer(testIdSingle).contains("Snorlax");
+      clearSingleValue(testIdSingle);
+
+      removeMultiValue(testIdMulti, "Eevee");
+      removeMultiValue(testIdMulti, "Snorlax");
+      selectOption(testIdMulti, "Eevee");
+
+      ["Charmander", "Eevee"].forEach((value) => {
+        getSelectValueContainer(testIdMulti).contains(value);
+      });
+
+      cy.intercept(
+        "DELETE",
+        `/api/v1/plus/custom-metadata/custom-field/id-custom-field-starter-pokemon`,
+        {
+          statusCode: 204,
+        }
+      ).as("deleteStarter");
+      cy.intercept("PUT", `/api/v1/plus/custom-metadata/custom-field`, {
+        fixture: "taxonomy/custom-metadata/custom-field/update-party.json",
+      }).as("updateParty");
+
+      cy.getByTestId("submit-btn").click();
+
+      cy.wait(["@updateParty", "@deleteStarter"]).then(
+        ([updatePartyInterception]) => {
+          expect(updatePartyInterception.request.body.id).to.eql(
+            "id-custom-field-pokemon-party"
+          );
+          expect(updatePartyInterception.request.body.resource_id).to.eql(
+            RESOURCE_CHILD.key
+          );
+          expect(updatePartyInterception.request.body.value).to.eql([
+            "Charmander",
+            "Eevee",
+          ]);
+        }
       );
     });
   });
--- a/clients/admin-ui/src/features/common/custom-fields/CustomFieldSelector.tsx
+++ b/clients/admin-ui/src/features/common/custom-fields/CustomFieldSelector.tsx
@@ -1,0 +1,65 @@
+import { useField } from "formik";
+import { useEffect } from "react";
+
+import { CustomSelect } from "~/features/common/form/inputs";
+import { AllowedTypes } from "~/types/api";
+
+import {
+  AllowListWithOptions,
+  CustomFieldDefinitionExisting,
+  CustomFieldExisting,
+} from "./types";
+
+type Props = {
+  allowList: AllowListWithOptions;
+  customField?: CustomFieldExisting;
+  customFieldDefinition: CustomFieldDefinitionExisting;
+};
+
+export const CustomFieldSelector = ({
+  allowList,
+  customField,
+  customFieldDefinition,
+}: Props) => {
+  const name = `definitionIdToCustomFieldValue.${customFieldDefinition.id}`;
+  const label = customFieldDefinition.name;
+  const tooltip = customFieldDefinition.description;
+  const { options } = allowList;
+
+  const [, meta, helpers] = useField({ name });
+
+  // This is a bit of a hack to set the initial value of the selector based on the CustomField
+  // returned by the API (if any). The "correct" way to do this would be to have the `initialValues`
+  // passed to `Formik` include the API values mapped by definition ID. However, because custom
+  // fields are so dynamic and depend on multiple API calls, mixing that logic into the (already
+  // complex) taxonomy form initialization isn't feasible for now. Those values are created by:
+  // src/features/taxonomy/hooks.tsx.
+  useEffect(
+    () => {
+      if (!customField?.value) {
+        return;
+      }
+
+      if (meta.touched) {
+        return;
+      }
+
+      helpers.setValue(customField.value, false);
+    },
+    // This should only ever run once, on first render.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    []
+  );
+
+  return (
+    <CustomSelect
+      isClearable
+      isMulti={customFieldDefinition.field_type !== AllowedTypes.STRING}
+      label={label}
+      name={name}
+      options={options}
+      tooltip={tooltip}
+      defaultValue={customField?.value ?? ""}
+    />
+  );
+};
--- a/clients/admin-ui/src/features/common/custom-fields/CustomFieldsList.tsx
+++ b/clients/admin-ui/src/features/common/custom-fields/CustomFieldsList.tsx
@@ -1,169 +1,41 @@
-/* eslint-disable @typescript-eslint/no-unused-vars */
 import { Center, Divider, Flex, Spinner, Text } from "@fidesui/react";
-import { FieldArray, useFormikContext } from "formik";
-import { useCallback, useEffect, useMemo, useState } from "react";
 
-import { CustomSelect } from "~/features/common/form/inputs";
-import { useAlert } from "~/features/common/hooks";
-import {
-  useGetAllAllowListQuery,
-  useGetCustomFieldDefinitionsByResourceTypeQuery,
-  useGetCustomFieldsForResourceQuery,
-  useUpsertCustomFieldMutation,
-} from "~/features/plus/plus.slice";
-import { AllowedTypes, ResourceTypes } from "~/types/api";
+import { ResourceTypes } from "~/types/api";
 
-import { useFeatures } from "../features";
+import { CustomFieldSelector } from "./CustomFieldSelector";
 import { CustomFieldsModal } from "./CustomFieldsModal";
-import { CustomFieldWithIdExtended } from "./types";
-
-const initialValues = {
-  customFields: [] as CustomFieldWithIdExtended[],
-};
-
-type FormValues = typeof initialValues;
+import { useCustomFields } from "./hooks";
 
 type CustomFieldsListProps = {
-  resourceId: string;
+  resourceFidesKey?: string;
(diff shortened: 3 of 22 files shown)


clients/admin-ui/src/features/common/custom-fields/CustomFieldSelector.tsx
```diff
@@ -0,0 +1,65 @@
...
+  const { options } = allowList;
+
+  const [, meta, helpers] = useField({ name });
+
+  // This is a bit of a hack to set the initial value of the selector based on the CustomField
+  // returned by the API (if any). The "correct" way to do this would be to have the `initialValues`
+  // passed to `Formik` include the API values mapped by definition ID. However, because custom
+  // fields are so dynamic and depend on multiple API calls, mixing that logic into the (already
+  // complex) taxonomy form initialization isn't feasible for now. Those values are created by:
+  // src/features/taxonomy/hooks.tsx.
```
ssangervasi, line 36: Now that I've refactored the taxonomy hooks anyway, it's possible the initial values could get populated in the correct way. I couldn't figure out a good way to do that yet though.

clients/admin-ui/src/features/common/custom-fields/CustomFieldsList.tsx
```diff
@@ -1,169 +1,41 @@
...
-  resourceId,
+export const CustomFieldsList = ({
+  resourceFidesKey,
   resourceType,
 }: CustomFieldsListProps) => {
-  const { errorAlert, successAlert } = useAlert();
-
-  const { flags, plus } = useFeatures();
-  const skip = !(flags.customFields && plus);
-  const [isLoading, setIsLoading] = useState(!skip);
```
ssangervasi, line 39: The loading state now comes from the queries/mutations themselves.

clients/admin-ui/src/features/common/custom-fields/CustomFieldsList.tsx
```diff
@@ -1,169 +1,41 @@
...
-  const [upsertCustomField] = useUpsertCustomFieldMutation();
-
-  // Reference parent Formik context
   const {
-    isSubmitting,
-    setSubmitting,
-    values: formValues,
-  } = useFormikContext();
-
-  const editCustomFieldList = useMemo(() => {
```
ssangervasi, line 57: The logic of these memoized bits has been extracted to a hook: `useCustomFields`. It's also been restructured, which I'll comment on that file.

clients/admin-ui/src/features/common/custom-fields/CustomFieldsList.tsx
```diff
@@ -1,169 +1,41 @@
...
+  } = useCustomFields({
+    resourceFidesKey,
+    resourceType,
+  });
 
-  useEffect(() => {
-    if (isSubmitting && (formValues as any).customFields) {
-      handleSubmit(formValues as FormValues);
-    }
-  }, [formValues, handleSubmit, isSubmitting]);
```
ssangervasi, line 161: **Key change**
A major motivation for integrating custom fields onto the taxonomy hooks is to avoid this monkey patching of Formik's submission state. Formik has good handling of async (promise) submission, so moving the logic to a single chain of API calls removes the need to do complex state-timing debugging.

clients/admin-ui/src/features/common/custom-fields/helpers.ts
```diff
@@ -6,3 +8,8 @@ export const getResourceType = (valueOrLabel: string) =>
   RESOURCE_TYPE_OPTIONS.find(
     (option) => option.value === valueOrLabel || option.label === valueOrLabel
   )?.value ?? ResourceTypes.SYSTEM;
+
```
ssangervasi, line 11: Oh, looks like `getResourceType` can be deleted now.

clients/admin-ui/src/features/common/custom-fields/hooks.ts
```diff
@@ -0,0 +1,191 @@
...
+export const useCustomFields = ({
+  resourceFidesKey,
+  resourceType,
+}: UseCustomFieldsOptions) => {
+  const { errorAlert, successAlert } = useAlert();
+  const { flags, plus } = useFeatures();
+  const isEnabled = flags.customFields && plus;
+
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const queryFidesKey = useMemo(() => resourceFidesKey ?? "", []);
```
ssangervasi, line 31: I mean to put a code comment here explaining that this prevents re-querying for a custom fields that will just 404.

clients/admin-ui/src/features/common/custom-fields/hooks.ts
```diff
@@ -0,0 +1,191 @@
...
+
+      // This will be undefined if the form never rendered a `CustomFieldList` that would assign
+      // form values.
+      if (!definitionIdToCustomFieldValue) {
+        return;
+      }
+
+      try {
+        // This would be a lot simpler (and more efficient) if the API had an endpoint for updating
+        // all the metadata associated with a field, including deleting options that weren't passed.
```
ssangervasi, line 130: Should I create an issue for this?

clients/admin-ui/src/features/common/custom-fields/hooks.ts
```diff
@@ -0,0 +1,191 @@
...
+            if (
+              value === undefined ||
+              value === "" ||
+              (Array.isArray(value) && value.length === 0)
+            ) {
+              if (!customField?.id) {
+                return undefined;
+              }
+
+              return deleteCustomFieldMutationTrigger(customField);
```
ssangervasi, line 145: The previous submission code didn't have the ability to delete custom fields. I think the API forbids setting the field to an empty string, which makes sense because that isn't an allowed value. So we have to delete the field assignment if the value is blank.

clients/admin-ui/src/features/common/custom-fields/types.ts
```diff
@@ -9,3 +14,33 @@ export interface CustomFieldWithIdExtended extends CustomFieldWithId {
 export interface Tab extends TabData {
   submitButtonText: string;
 }
+
+export interface CustomFieldDefinitionExisting
```
ssangervasi, line 18: `...Existing` is the best I could think of. The `...WithId` type allows `id` to be `undefined` which isn't useful in most of the code.

clients/admin-ui/src/features/common/form/inputs.tsx
```diff
@@ -169,6 +169,7 @@ const SelectInput = ({
       name={fieldName}
       value={selected}
       size={size}
+      classNamePrefix="custom-select"
```
ssangervasi, line 172: react-select attaches this to its various components, which allows the tests to cleanly target specific elements.

clients/admin-ui/src/features/system/DescribeSystemStep.tsx
```diff
@@ -65,9 +67,10 @@ const DescribeSystemStep = ({
         : defaultInitialValues,
     [passedInSystem]
   );
-  const [createSystem] = useCreateSystemMutation();
-  const [updateSystem] = useUpdateSystemMutation();
-  const [isLoading, setIsLoading] = useState(false);
+  const [createSystemMutationTrigger, createSystemMutationResult] =
+    useCreateSystemMutation();
+  const [updateSystemMutationTrigger, updateSystemMutationResult] =
+    useUpdateSystemMutation();
```
ssangervasi, line 73: Yes, the naming scheme of `<operation><Model>Mutation<Trigger|Result>` is verbose. But the more I encounter code where the API-calling function starts off as a simple name, then gets wrapped in handling logic, and so on, the more I think making it clear which one actually fires the request is worth the characters.

clients/admin-ui/src/features/system/form.ts
```diff
@@ -1,8 +1,10 @@
+import { CustomFieldsFormValues } from "~/features/common/custom-fields";
 import { DEFAULT_ORGANIZATION_FIDES_KEY } from "~/features/organization";
 import { DataProtectionImpactAssessment, System } from "~/types/api";
 
 export interface FormValues
-  extends Omit<System, "data_protection_impact_assessment"> {
+  extends Omit<System, "data_protection_impact_assessment">,
+    CustomFieldsFormValues {
```
ssangervasi, line 7: The benefit to putting the `CustomFieldsFormValues` into its own type is that these forms can then be explicit that they these properties as well. Previously the system and the taxonomy could end up including the custom fields in the JSON for an endpoint that doesn't expect it.

clients/admin-ui/src/features/taxonomy/TaxonomyFormBase.tsx
```diff
@@ -59,27 +55,9 @@ const TaxonomyFormBase = ({
 
   const handleSubmit = async (newValues: FormValues) => {
     setFormError(null);
-    // parent_key and fides_keys are immutable
-    // parent_key also needs to be undefined, not an empty string, if there is no parent element
-    let payload: TaxonomyEntity;
```
ssangervasi, line 64: This payload mutation that was mandatory for every resource type is now included in the hooks.

clients/admin-ui/src/features/taxonomy/TaxonomyFormBase.tsx
```diff
@@ -146,14 +124,6 @@ const TaxonomyFormBase = ({
...
                 ))}
               {renderExtraFormFields ? renderExtraFormFields(values) : null}
-              {["data category", "data use", "data subject"].includes(
-                labels.fides_key.toLowerCase()
-              ) && (
-                <CustomFieldsList
-                  resourceId={initialValues.fides_key}
-                  resourceType={getResourceType(labels.fides_key)}
-                />
-              )}
```
ssangervasi, line 156: This block now lives in `renderExtraFormFields` when appropriate for the resource type. This actually removes the need for any resource type checking.

clients/admin-ui/src/features/taxonomy/TaxonomyTabContent.tsx
```diff
@@ -69,7 +71,6 @@ const TaxonomyTabContent = ({ useTaxonomy }: Props) => {
     return null;
   }, [data]);
 
-  const [entityToEdit, setEntityToEdit] = useState<TaxonomyEntity | null>(null);
```
ssangervasi, line 72: The hooks gain a lot of info by gaining control of this state. It was necessary for looking up custom fields by fides key, but I think a later refactor could make it even ore useful.

clients/admin-ui/src/features/taxonomy/hooks.tsx
```diff
@@ -66,7 +78,33 @@ const transformTaxonomyBaseToInitialValues = (t: TaxonomyEntity) => ({
...
+  if (isCreate) {
+    const parentKey = parentKeyFromFidesKey(newValues.fides_key);
+    entity.parent_key = parentKey === "" ? undefined : parentKey;
+  } else {
+    entity.parent_key =
+      initialValues.parent_key === "" ? undefined : initialValues.parent_key;
+    entity.fides_key = initialValues.fides_key;
+  }
+
+  delete entity.definitionIdToCustomFieldValue;
```
ssangervasi, line 99: Note that the FormValues type now knows that this property exists (instead of just getting snuck in by Formik).

clients/admin-ui/src/features/taxonomy/hooks.tsx
```diff
@@ -76,22 +114,71 @@ export const useDataCategory = (): TaxonomyHookData<DataCategory> => {
...
+
+  const handleCreate = async (
+    initialValues: FormValues,
+    newValues: FormValues
+  ) => {
+    const payload = transformBaseFormValuesToEntity(initialValues, newValues);
+    const result = await createDataCategoryMutationTrigger(payload);
+
+    if (customFields.isEnabled) {
+      await customFields.upsertCustomFields(newValues);
```
ssangervasi, line 134: Voilà, the form submission now has a single flow of async calls.

clients/admin-ui/src/features/taxonomy/hooks.tsx
```diff
@@ -38,21 +41,30 @@ import {
...
-import { Labels, TaxonomyEntity } from "./types";
+import { FormValues, Labels, TaxonomyEntity } from "./types";
 
 export interface TaxonomyHookData<T extends TaxonomyEntity> {
   data?: TaxonomyEntity[];
   isLoading: boolean;
   labels: Labels;
-  handleCreate: (entity: T) => RTKResult<TaxonomyEntity>;
-  handleEdit: (entity: T) => RTKResult<TaxonomyEntity>;
+  resourceType?: ResourceTypes;
```
ssangervasi, line 57: I added this so that components that use `TaxonomyHookData` could know what resource they're working with. However, it didn't turn out to be that useful outside of the body of the hook (so far).

clients/admin-ui/src/features/common/custom-fields/CustomFieldSelector.tsx
chriscalhoun1974: Nice work encapsulating this logic using additional hooks and into its own drop down component.

chriscalhoun1974: Still don't think its possible to update the initial values in the CustomFieldsList component?

clients/admin-ui/src/features/common/custom-fields/helpers.ts
chriscalhoun1974: Yes, this method can be deleted now given there are no references.

clients/admin-ui/src/features/common/custom-fields/hooks.ts
chriscalhoun1974: You are referencing your comment on line 130?

clients/admin-ui/src/features/common/custom-fields/hooks.ts
chriscalhoun1974: The API does allow for an empty array to be submitted. However, I do understand about your point about deleting the custom field assignment.

clients/admin-ui/src/features/common/custom-fields/hooks.ts
```diff
@@ -0,0 +1,191 @@
...
+
+        successAlert(
+          `Custom field(s) successfully saved and added to this ${resourceType} form.`
+        );
+      } catch (e) {
+        errorAlert(
+          `One or more custom fields have failed to save, please try again.`
+        );
+        // eslint-disable-next-line no-console
+        console.error(e);
```
chriscalhoun1974, line 167: Can you remove this console error statement?

chriscalhoun1974 (approved):
lgtm

clients/admin-ui/cypress/e2e/taxonomy-plus.cy.ts
```diff
@@ -10,23 +10,48 @@ describe("Taxonomy management with Plus features", () => {
...
+  };
+  const RESOURCE_CHILD = {
+    label: "Biometric Data",
+    key: "user.biometric",
+  };
+
   const navigateToEditor = () => {
-    cy.getByTestId("accordion-item-User Data").click();
-    cy.getByTestId("item-Biometric Data")
+    cy.getByTestId(`accordion-item-${RESOURCE_PARENT.label}`).click();
```
allisonking, line 27: ah right, getting flashbacks of those recursive taxonomy accordion days 😭

clients/admin-ui/cypress/e2e/taxonomy-plus.cy.ts
```diff
@@ -10,23 +10,48 @@ describe("Taxonomy management with Plus features", () => {
...
-      .within(() => {
-        cy.contains(optionText).click();
-      });
+    getSelectOptionList(selectorId).contains(optionText).click();
   };
 
+  const removeMultiValue = (selectorId: string, optionText: string) =>
+    getSelectValueContainer(selectorId)
+      .contains(optionText)
+      .siblings(".custom-select__multi-value__remove")
```
allisonking, line 49: oh nice this is a good way to do it

clients/admin-ui/cypress/e2e/taxonomy-plus.cy.ts
```diff
@@ -149,4 +174,106 @@ describe("Taxonomy management with Plus features", () => {
...
+        "GET",
+        `/api/v1/plus/custom-metadata/custom-field/resource/${RESOURCE_CHILD.key}`,
+        {
+          fixture: "taxonomy/custom-metadata/custom-field/list.json",
+        }
+      ).as("getCustomFields");
+
+      navigateToEditor();
+
+      cy.wait([
```
allisonking, line 213: wow I didn't know you could do an array like this!

clients/admin-ui/src/features/common/form/inputs.tsx
allisonking: nice!

clients/admin-ui/src/features/taxonomy/hooks.tsx
allisonking: heh nice, I always felt a little too sneaky about that

allisonking (approved):
whew, nice job!!

clients/admin-ui/src/features/common/custom-fields/CustomFieldSelector.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+  return (
+    <CustomSelect
+      isClearable
+      isMulti={customFieldDefinition.field_type !== AllowedTypes.STRING}
+      label={label}
+      name={name}
+      options={options}
+      tooltip={tooltip}
+      defaultValue={customField?.value ?? ""}
```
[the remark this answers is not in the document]
ssangervasi: Yeah that's gonna be fun.

clients/admin-ui/src/features/common/custom-fields/hooks.ts
ssangervasi: https://github.com/ethyca/fides/issues/2612

clients/admin-ui/src/features/common/custom-fields/CustomFieldSelector.tsx
ssangervasi: No, I think it would have to be higher up, to where Formik actually receives the prop, which I think means it needs to fit into the taxonomy hooks' methods.

clients/admin-ui/src/features/common/custom-fields/hooks.ts
ssangervasi: I was thinking it could be helpful even after merge in case we encounter another API incompatibility.

PR #2588, opened 2023-02-14
Set the fides dev environment to the `prod` security environment

### LINKED ISSUE #2461: Make `prod` security env the default for our dev environment

With the completion of #2267 we should make the project default `prod` instead of `dev`  to make sure we're developing in the most secure environment that should also become our user's default.

ThomasLaPiana:
Closes #2461 

### Code Changes

* [x] update the `fides.toml`
* [x] update `docker compose run` -> `docker exec` where possible. This not only saves resources (one less container running!) but allows multiple commands to run in the same environment (i.e. when something gets run via `docker compose run`, it is ephemeral and can't be used by following commands
* [x] Add a "login" step that runs before any of the ctl tests get run. This only happens as part of nox, so when running locally a manual login will be needed for the `ctl` tests to pass
* [x] add the `auth_header` to the `fides db` CLI command requests

### Steps to Confirm

* [x] all tests pass

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### Description Of Changes

With the change in this PR, a CLI login will be required during development, as the new default is `prod` for the security env.


ThomasLaPiana:
I got all of the tests passing locally but then it looks like they're not passing in CI or when building the test image locally and running it that way...will trouble shoot more locally

ThomasLaPiana:
if I log into a container locally and run `fides user login` and then run the tests everything works just fine...but the login command within the fixture doesn't seem to be doing the trick despite the fact that it definitely writes a file

ThomasLaPiana:
Found the problem in the logs....

```
2023-02-15 15:19:26.362 [DEBUG] (oauth_util:verify_oauth_client:174): Unable to parse auth token.
```

So it looks like the order of execution for the fixtures might be wrong....the token gets created at the beginning, but then gets reset because its used to reset the db, so then becomes invalid. So this fixture needs to get called twice

sanders41:
Is it `autouse` fixtures you are having issues with or fixtures you are adding to the test yourself? If the latter try playing with order, fixtures run bottom to top so you may be able to change it up and get a different result.

ThomasLaPiana:
> Is it `autouse` fixtures you are having issues with or fixtures you are adding to the test yourself? If the latter try playing with order, fixtures run bottom to top so you may be able to change the order up and get a different result.

I'm trying to wrap all of the resets as needed....getting closer

ThomasLaPiana:
I've once again confirmed that it works perfectly if you have a credentials file before running the tests, but if it tries to create one during the test it doesn't work. I'm not sure why this happens. Going to try using pytest tricks to generate one as a non-fixture?

Despite all of the logs showing that one gets created just fine during the tests 🤔

SteveDMurphy (commented):
Thanks for explaining the difference between `docker run` and `docker exec` and the value of the change 👍🏽 👍🏽 when spinning up the environment in Docker I am not able to run any commands for some reason, including the `fides user login ...` - installing the branch locally in a virtual env doesn't seem to have the same issue from what I can see 🤔

sanders41:
I accidently posted my comment in the issue instead of here. I'm seeing things similar to Steve. https://github.com/ethyca/fides/issues/2461#issuecomment-1432058563

ThomasLaPiana:
ah I'm a total rookie! Sorry for wasting y'alls time, the exec for the interactive shell needs to have the `-it` flag

Added and confirmed working, please try again 🙏

SteveDMurphy (approved):
Tried out a few of these successfully! Thanks for the guidance in testing as well 🙌🏽 

The only other thing that might be helpful is if we wanted to add something around running pytest locally and needing to log in to the readme or similar? Can imagine people hitting some snags

noxfiles/ci_nox.py
```diff
@@ -125,16 +125,20 @@ def fides_db_scan(session: nox.Session) -> None:
     """Scan the fides application database to check for dataset discrepancies."""
     session.notify("teardown")
     session.run(*START_APP, external=True)
-    run_command = (
-        *RUN,
+    scan_command = (
+        "docker",
+        "container",
+        "exec",
+        "fides-fides-1",
```
TheAndrewJackson, line 132: Should this use the constant added in `constants_nox.py`?

```suggestion
        CONTAINER_NAME,
```

noxfiles/constants_nox.py
```diff
@@ -38,27 +39,41 @@
...
-    "-e",
-    "VAULT_TOKEN",
-    "--rm",
+    "exec",
+    "fides-fides-1",
+    "fides",
+    "user",
+    "login",
+    "-u",
+    "root_user",
```
TheAndrewJackson, line 57: Is it possible to pull in the root user's username and password dynamically from the `fides.toml` file? Not a huge deal but this would break if that changes in the future

tests/ctl/cli/test_cli.py
```diff
@@ -930,3 +929,17 @@ def test_user_permissions(
...
+        self, test_config_path: str, test_cli_runner: CliRunner, credentials_path: str
+    ) -> None:
+        """Test getting user permissions but the credentials file doesn't exit."""
+        print(credentials_path)
+        result = test_cli_runner.invoke(
+            cli,
+            ["-f", test_config_path, "user", "permissions"],
+            env={"FIDES_CREDENTIALS_PATH": "/root/notarealfile.credentials"},
+        )
+        print(result.output)
```
TheAndrewJackson, line 944: Do the prints still need to be here?
```suggestion

```

TheAndrewJackson (approved):
This worked great for me locally 👍 I just left a few optional nits

tests/ctl/cli/test_cli.py
ThomasLaPiana: yep! In tests, print statements are only shown if the test fails. Extremely helpful for debugging and only shows up when its needed 🙂

noxfiles/constants_nox.py
ThomasLaPiana: Currently no. Originally I had this as a feature, but realistically 99.9% of users in a "real" environment shouldn't be using the root credentials so I didn't code that feature in.

Agreed hardcoding is suboptimal, but we've standardized on these default credentials across both core and plus. Additionally it should be pretty obvious if someone decides to change them, the failed login message would be clear I think

noxfiles/ci_nox.py
ThomasLaPiana: good catch!

ThomasLaPiana:
@sanders41 seeing codecov failures here

sanders41:
Looks like they passed. Did you have to rerun for them to pass?

sanders41:
I see it now. This is the flaky issue I mentioned. It has been terrible this week. Looks like adding a token is helping, but not completely preventing the issue. I restarted the failed tests, they usually pass after a rerun.

ThomasLaPiana:
it failed the codecov upload, merging

PR #2590, opened 2023-02-14
Describe system update

allisonking:
Closes https://github.com/ethyca/fides/issues/2435

### Code Changes

* [x] Rename `DescribeSystemStep` to `SystemInformationForm` and `DescribeSystemStepExtension` to `SystemInformationFormExtension`
* [x] Move some of the form fields from `SystemInformationForm` to `SystemInformationFormExtension` since we don't need all those fields around for the initial add systems flow
* [x] Handle adding a new system success and cancel states
* [x] Handle moving to a new tab when you haven't saved yet
* [ ] Updated privacy declaration UI --> Going to split this one out into a separate PR as this one was already getting big
  * https://github.com/ethyca/fidesplus/issues/598
* [x] Cypress tests

### Steps to Confirm

* [ ] Check out the loom

### Pre-Merge Checklist

* Documentation:
  * [x] https://github.com/ethyca/fidesplus/issues/598
* [ ] Update `CHANGELOG.md`

### Description Of Changes

https://www.loom.com/share/eb0d50321faf4faa8ab737dd78d5cd4a


clients/admin-ui/src/features/system/EditSystemFlow.tsx
```diff
@@ -4,16 +4,15 @@ import { useRouter } from "next/router";
 import { useCallback, useState } from "react";
```
allisonking, line 4: Once we no longer support nav v1, this can likely be consolidated with `SystemFormTabs.tsx`

clients/admin-ui/src/features/system/SystemFormTabs.tsx
```diff
@@ -0,0 +1,208 @@
...
+  const [tabIndex, setTabIndex] = useState(0);
+  const [queuedIndex, setQueuedIndex] = useState<number | undefined>(undefined);
+  const [showSaveMessage, setShowSaveMessage] = useState(false);
+  const { systemOrDatamapRoute } = useInterzoneNav();
+  const router = useRouter();
+  const toast = useToast();
+  const dispatch = useAppDispatch();
+  const activeSystem = useAppSelector(selectActiveSystem);
+
+  const goBack = useCallback(() => {
```
allisonking, line 80: a good chunk of this was extracted from `EditSystemFlow.tsx`

clients/admin-ui/src/features/system/SystemInformationFormExtension.tsx
```diff
@@ -0,0 +1,115 @@
+import { Box, FormLabel, Stack } from "@fidesui/react";
```
allisonking, line 1: this is mostly just a rename from `DescribeSystemsFormExtension`, except some of the fields (tags and dependencies) got moved to the extension from the base

clients/admin-ui/src/features/system/SystemFormTabs.tsx
```diff
@@ -0,0 +1,208 @@
...
+            <SystemInformationForm
+              onSuccess={handleSuccess}
+              system={activeSystem}
+              abridged={isCreate}
+            >
+              <UnmountWarning
+                isUnmounting={queuedIndex !== undefined}
+                onContinue={continueTabChange}
+                onCancel={() => setQueuedIndex(undefined)}
+              />
```
allisonking, line 157: tbh I don't love this pattern, but I think it works decently well. the problem with it is I'd rather have the unmount be observed by the tab component, since now every component that is part of the tab either needs this `UnmountWarning` component as a child, or we need special logic to go around the queue/step logic (like the `if` statement in `checkTabChange`. the UnmountWarning needs to be a child in order to have access to the formik context to see if the form is dirty.

the experience team also ran into this recently and [handled it through refs](https://github.com/ethyca/fides/blob/main/clients/admin-ui/src/features/common/custom-fields/CustomFieldsModal.tsx#L49-L62) which I didn't love, but maybe I'm just averse to refs

TheAndrewJackson: This is definitely a tricky issue. I found a potential bug with the unsaved changes logic. If it would be too complex or time consuming then I don't think it's worth fixing in this PR. It seems like a good follow up bug.

The issue is if a user clicks on a link to navigate away instead of the privacy declaration tab it doesn't prompt them to save their changes.

https://user-images.githubusercontent.com/17103888/219412682-b1232477-a2a3-45fc-a61b-caa0f46b28a5.mov


Maybe in the future we could standard on a solution that we like? It might confusing if there are multiple ways that we're handling this scenario

TheAndrewJackson (approved):
Works great. I like updates. I think I found a small bug with the unmounting logic but it doesn't need to be blocking if it's a time consuming fix.

clients/admin-ui/src/features/system/SystemFormTabs.tsx
allisonking: yeah, I couldn't figure out a way to render the modal while the component is unmounting (which is what happens when you click away). it also wouldn't work if the user actually entered a different URL to navigate away (the component don't go through their unmounting phase that way).

actually maybe I'm going about this wrong and should instead try to use next router to see if I can hook into that... that wouldn't work for if the user actually navigated away from the page (going from localhost:3000 --> i.e. google.com), but hey we can't account for _everything_

I'll try it out!

allisonking: okay the good news is that we can probably intercept next router and window events to handle this

the bad news is that it's fairly complicated 😅 

https://github.com/vercel/next.js/discussions/12348

I'm going to make a separate ticket for this since otherwise I'll spend too much time on it right now

allisonking: https://github.com/ethyca/fides/issues/2631

Issue #2178, opened 2023-01-10
Track down and fix intermittent test failures


ThomasLaPiana:

ThomasLaPiana:
@sanders41 do you have any idea where to start with this one? I can find some job examples if needed

sanders41:
I don’t know specific test, but I know it happens fairly regularly with the mongodb tests. My hunch is it happens because records don’t get cleared from the database somewhere.

ThomasLaPiana:
Here's an example, looks like it is indeed mongo: https://github.com/ethyca/fides/actions/runs/3948924135/jobs/6759694017

sanders41:
Timescale seems to be another that does this. Example here: https://github.com/ethyca/fides/actions/runs/4034266232/jobs/6935416892

ThomasLaPiana:
I think the easiest short-term solution here would be to add test reruns

https://stackoverflow.com/questions/52967150/can-i-retry-for-failed-tests-in-pytest

Any arguments in favor of not doing that and instead digging into the potential race conditions?

sanders41:
This could work for timescale with a restart on `OperationalError`. For mongo it may not work depending on if the retry plug-in tears down the session fixtures first or not. That one is an assert error so if it failed because there are records there causing unexpected values, and the retry doesn’t clear the database first it will fail again.

it’s worth a try though.

PR #2591, opened 2023-02-14
Switch mongo tests from randint to uuid

sanders41:
Relates to #2178

### Code Changes

* Switch from `randint` to `uuid` in mongo tests

### Steps to Confirm

* verify mongo tests pass

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

The random int used in mongo tests is random, but not unique. The mongo tests don't clear out records between tests so by chance it is possible to have a clash in ids. This PR changes the use of random ints to using uuids. This is one of the causes mongo tests are flaky in CI, but I suspect it isn't the only reason so there is probably more to do here.


ThomasLaPiana (approved):
Great change here, I wonder if this might fix some of the pymongo update bugs I've been seeing too 🤔  (it was complaining about invalid UUID stuff)

sanders41:
In 3.8 of pymomgo there was a change to the ObjectID format and it has to be 12 bytes in a specific format https://www.mongodb.com/docs/manual/reference/method/ObjectId/. I bet this is the error.

Issue #2240, opened 2023-01-13
BE: Record all applications of a Global Privacy Control Signal


mfbrown:
This is the backed work for #2232

As a compliance person, I want a record of each consent selection which is made because of a global privacy control signal so that I can demonstrate to regulators our compliance and so I can understand the impact and use of the feature in my customer base.

AC

- Given that I have enabled global privacy control support on one or more data use cases, when a user has consent settings applied to adhere to their GPC signal, then a record is created with 1) the use case and information about the data subject, to the extent possible (existing functionality in our current `Consent` db table), and 2) both the existence of the GPC flag and whether or not the consent record conflicts with the GPC flag
- If the user has accessed the privacy center, save the identity (existing functionality in our current `Consent` db table)

eastandwestwind:
Connected with Michael and clarified a couple things:

1. Users will eventually save consent preferences via a Fides consent banner, when they won't have a true "identity" yet. Right now this is still only available on the privacy center.
2. Whether or not GPC applies to each data use case is stored in the config.json on FE. Because of this, we wouldn't be able to infer that a selection conflicted with GPC just from the database, if we only stored `hasGpcFlag`. So, I advise storing a `conflictsWithGpc` flag in the db too.
3. Currently the Consent table is NOT historical, meaning each time a user saves consent preference for a data use case, the previous one associated with that data use case gets overridden. This is OK for now, but we anticipate supporting historical reporting here- https://ethyca.atlassian.net/wiki/spaces/PM/pages/2609021029/Enhancing+Compliance+Capabilities+In+Fides+with+User+Consent+Reporting?atlOrigin=eyJpIjoiZTJmNzNkYmM1OTk1NDNlZWExZWJmNWY2ZDA2YzJiM2UiLCJwIjoiY29uZmx1ZW5jZS1jaGF0cy1pbnQifQ

PR #2599, opened 2023-02-15
Record all applications of a Global Privacy Control Signal

eastandwestwind:
Closes https://github.com/ethyca/fides/issues/2240

### Code Changes

* [ ] Updates `Consent` table to store 2 new columns pertaining to GPC: `has_gpc_flag` and `conflicts_with_gpc`
* [ ] Updates `Consent` schema to intake these 2 fields as optional
* [ ] Adds tests

### Steps to Confirm

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes


src/fides/api/ctl/migrations/versions/d65bbc647083_adds_gpc_info_to_consent_table.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('consent', sa.Column('has_gpc_flag', sa.Boolean(), nullable=True))
+    op.add_column('consent', sa.Column('conflicts_with_gpc', sa.Boolean(), nullable=True))
+    op.execute("UPDATE consent SET has_gpc_flag = false")
+    op.alter_column('consent', 'has_gpc_flag', server_default="f", default=False, nullable=False)
+    op.execute("UPDATE consent SET conflicts_with_gpc = false")
+    op.alter_column('consent', 'conflicts_with_gpc', server_default="f", default=False, nullable=False)
```
sanders41, line 26: Just double checking, is `f` correct for `server_default`? It gets set to `false` in the query.

src/fides/api/ops/models/privacy_request.py
```diff
@@ -851,6 +851,8 @@ class Consent(Base):
     data_use = Column(String, nullable=False)
     data_use_description = Column(String)
     opt_in = Column(Boolean, nullable=False)
+    has_gpc_flag = Column(Boolean, server_default="f", default=False, nullable=False)
+    conflicts_with_gpc = Column(Boolean, server_default="f", default=False, nullable=False)
```
sanders41, line 855: Same question on `server_default` of `f`. The answer in the other section will answer this on also.

src/fides/api/ops/schemas/privacy_request.py
```diff
@@ -60,6 +60,8 @@ class Consent(BaseSchema):
     data_use: str
     data_use_description: Optional[str] = None
     opt_in: bool
+    has_gpc_flag: Optional[bool] = False
+    conflicts_with_gpc: Optional[bool] = False
```
sanders41, line 64: Since these can't be `None` I don't think you need `Optional`. You can still default to `False` and they don't have to be sent. 

```suggestion
    has_gpc_flag: bool = False
    conflicts_with_gpc: bool = False
```

tests/ops/models/test_consent_request.py
```diff
@@ -64,6 +58,40 @@ def test_consent(db):
...
+    consent_2 = Consent.create(db, data=consent_data_2)
+    data_uses = [x.data_use for x in provided_identity.consent]
+
+    assert consent_data_1["data_use"] in data_uses
+    assert consent_data_2["data_use"] in data_uses
+
+    provided_identity.delete(db)
+
+    assert Consent.get(db, object_id=consent_1.id) is None
+    assert Consent.get(db, object_id=consent_2.id) is None
```
sanders41, line 92: Is this to test cascading deletes? If yes all good, if it is just to clear the records there is a fixture that will clear them.

eastandwestwind: Yes this tests the cascade delete from `ProvidedIdentity` to associated  `Consent` records

sanders41:
I just testing locally and thing started without error. Going to try a restart on the tests here. Hopefully we aren't back to a CI only issue 😕

eastandwestwind:
@sanders41 it's just the migration down rev that needed to be bumped. It's weird that this is the way it manifests in CI though.

src/fides/api/ops/models/privacy_request.py
eastandwestwind: same as above

sanders41:
> @sanders41 it's just the migration down rev that needed to be bumped. It's weird that this is the way it manifests in CI though.

And that it doesn't happen locally. Very weird.

sanders41 (approved):

PR #2602, opened 2023-02-15
Updating Firebase Auth to delete user

### LINKED ISSUE #2601: Update Firebase Auth to delete user instead of updating it

Depending on the Policy in use, the `phone_number` field might be masked with a fixed string. This would cause Firebase to throw a duplicate phone number error on subsequent updates

Switch the Firebase Auth connector to use a `delete` instead of an `update` for the user collection.

We could patch the existing Firebase Auth connector instance since the necessary request override function is already in Fides but we want to make sure the user can create additional Firebase Auth connections without having to apply the patch.

galvana:
Closes #2601 

### Code Changes

* [x] Updating config to use `firebase_auth_user_delete` instead of `firebase_auth_user_update`

### Steps to Confirm

* [x] Ran integration tests

### Pre-Merge Checklist

* [x] Update `CHANGELOG.md`

### DIFF
--- a/data/saas/config/firebase_auth_config.yml
+++ b/data/saas/config/firebase_auth_config.yml
@@ -3,7 +3,7 @@ saas_config:
   name: Firebase Auth SaaS Config
   type: firebase_auth
   description: A sample schema representing the Firebase Auth connector for Fidesops
-  version: 0.0.2
+  version: 0.0.3
 
   connector_params:
     - name: domain
@@ -37,14 +37,14 @@ saas_config:
         read:
           - request_override: firebase_auth_user_access
             param_values:
-            - name: email
-              identity: email
+              - name: email
+                identity: email
           - request_override: firebase_auth_user_access
             param_values:
-            - name: phone_number
-              identity: phone_number
-        update:
-          request_override: firebase_auth_user_update
+              - name: phone_number
+                identity: phone_number
+        delete:
+          request_override: firebase_auth_user_delete
           param_values:
             - name: email
               identity: email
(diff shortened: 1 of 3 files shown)


data/saas/config/firebase_auth_config.yml
```diff
@@ -37,14 +37,14 @@ saas_config:
         read:
           - request_override: firebase_auth_user_access
             param_values:
-            - name: email
-              identity: email
+              - name: email
```
galvana, line 40: Formatting auto-fixes

tests/ops/integration_tests/saas/request_override/test_firebase_auth_task.py
```diff
@@ -410,8 +416,8 @@ async def test_firebase_auth_delete_request(
     """
     Tests delete functionality by explicitly invoking the delete override function
 
-    We can't have an'end-to-end' privacy request test, as preferred, because
-    our delete function is not configured by default (the udpate function is).
+    We can't have an 'end-to-end' privacy request test, as preferred, because
+    our delete function is not configured by default (the update function is).
     But this at least gives us some test coverage of the delete function directly.
```
pattisdr, line 421: It looks like the delete tests aren't running erasure privacy requests because `update` wasn't the default. Now that we've disabled the update tests, can we add the erasure test here?

tests/ops/integration_tests/saas/request_override/test_firebase_auth_task.py
```diff
@@ -410,8 +563,8 @@ async def test_firebase_auth_delete_request(
...
```
pattisdr, line 568: This comment is still outdated -

galvana: Fixed!

PR #2619, opened 2023-02-15
Erasure ordering

### LINKED ISSUE #2618: Support erasure ordering

The erasure graph in Fides does not enforce a specific task execution order ([docs](https://docs.ethyca.com/fides/dsr_quickstart/dsr_support/request_execution#graph-building-1))
> The "graph" for an erasure runs on the data from the access request, which is kept in temporary storage, and can be used to locate data for each Collection individually. Because the data has already been found, each Collection could be visited in any order or run in parallel. The graph is configured so each Collection has its previous access request results passed in as inputs, and each Collection returns a count of records masked when complete.

image

Fides should allow the user to specify the order in which some tasks need to execute, similar to how we can order access requests using the `after` field available on collections and datasets.
```
endpoints:
  - name: orders
    erase_after:
      - <instance_fides_key>.orders_to_refunds
      - <instance_fides_key>.refunds_to_orders
```

galvana:
Closes #2618

### Code Changes

* [x] Add `erase_after` to collections
* [x] Updated the Dask graph generated for erasures in `graph_task.run_erasure` to take the `erase_after` field into account
* [x] Updated the parameters of `graph_task.erasure_request` to support Dask task ordering

### Steps to Confirm

* [x] Verified the functionality in the `test_integration_erasure_order` tests

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`


data/saas/config/saas_erasure_order_config.yml
```diff
@@ -0,0 +1,79 @@
...
+    protocol: https
+    host: <domain>
+
+  test_request:
+    method: GET
+    path: /test/
+
+  endpoints:
+    - name: orders
+      erase_after:
```
galvana, line 21: New `erase_after` field that maps to `collection.erase_after` in the graph generated from the SaaS config. Should we update the existing `after` field to `access_after` for consistency?

src/fides/api/ops/service/saas_request/override_implementations/no_op_request_overrides.py
```diff
@@ -0,0 +1,36 @@
...
+    policy: Policy,
+    privacy_request: PrivacyRequest,
+    input_data: Dict[str, List[Any]],
+    secrets: Dict[str, Any],
+) -> List[Row]:
+    return [{f"{node.address.collection}_id": 1}]
+
+
+@register("delete_no_op", [SaaSRequestType.DELETE])
+def delete_no_op(
```
galvana, line 29: Read and delete no-op request overrides to facilitate testing.

src/fides/api/ops/task/graph_task.py
```diff
@@ -511,7 +511,12 @@ def access_request(self, *inputs: List[Row]) -> List[Row]:
         return filtered_output
 
     @retry(action_type=ActionType.erasure, default_return=0)
-    def erasure_request(self, retrieved_data: List[Row], *inputs: List[Row]) -> int:
+    def erasure_request(
+        self,
+        retrieved_data: List[Row],
+        inputs: List[List[Row]],
+        *_: int,
```
galvana, line 518: Not sure the best name for this, this is just here so we can set the dependent tasks like we do for the access task graph

```
 saas_erasure_order_instance:labels: (<bound method GraphTask.erasure_request of <class 'fides.api.ops.task.graph_task.GraphTask'>:saas_erasure_order_instance:labels>,
                                      [{'labels_id': 1}],
                                      [[{'email': 'test@ethyca.com'}]],
                                      saas_erasure_order_instance:refunds,
                                      saas_erasure_order_instance:orders),
```

src/fides/api/ops/task/graph_task.py
```diff
@@ -746,16 +751,20 @@ def collect_tasks_fn(
...
-
-            The termination function just returns this tuple of ints."""
-            return dependent_values
+        def termination_fn(*_: int) -> Dict[str, int]:
+            """
+            The erasure order can be affected in a way that not every node is directly linked
+            to the termination node. This means that we can't just aggregate the inputs directly,
+            we must read the erasure results from the cache.
+            """
+            return resources.get_all_cached_erasures()
```
galvana, line 767: Major change: returning the erasure counts from the cache instead of the direct output of each erasure task

src/fides/api/ops/task/graph_task.py
```diff
@@ -765,34 +774,45 @@ def termination_fn(*dependent_values: int) -> Tuple[int, ...]:
...
                 ),  # Pass in the results of the access request for this collection
-                *[
+                [
                     access_request_data.get(
                         str(upstream_key), []
                     )  # Additionally pass in the original input data we used for the access request. It's helpful in
                     # cases like the EmailConnector where the access request doesn't actually retrieve data.
                     for upstream_key in t.input_keys
                 ],
+                *_evaluate_erasure_dependencies(t, end_nodes),
```
galvana, line 784: Major change: adding the dependent tasks to each task in the task graph

src/fides/api/ops/task/graph_task.py
```diff
@@ -765,34 +774,45 @@ def termination_fn(*dependent_values: int) -> Tuple[int, ...]:
...
+        return v.compute()
 
-        update_cts: Tuple[int, ...] = v.compute()
-        # we combine the output of the termination function with the input keys to provide
-        # a map of {collection_name: records_updated}:
-        erasure_update_map: Dict[str, int] = dict(
-            zip([str(x) for x in env], update_cts)
-        )
 
-        return erasure_update_map
```
galvana, line 795: The output of `v.compute` now matches the return type so no need to transform the result

src/fides/api/ops/task/graph_task.py
```diff
@@ -765,34 +774,45 @@ def termination_fn(*dependent_values: int) -> Tuple[int, ...]:
...
+    to be erased before the given task. Remove the dependent collection addresses
+    from `end_nodes` so they can be executed in the correct order.
+    """
+    erase_after = t.traversal_node.node.collection.erase_after
+    for collection in erase_after:
+        if collection in end_nodes:
+            end_nodes.remove(collection)
+    # this task will execute after the collections in `erase_after` or
+    # execute at the beginning by linking it to the root node
+    return erase_after if len(erase_after) else set([ROOT_COLLECTION_ADDRESS])
```
galvana, line 815: This function currently only looks at the collection's erase_after, not the dataset's erase_after. I wanted to make sure I was going down the right path before expanding this to include the dataset's erase_after

src/fides/api/ops/task/graph_task.py
```diff
@@ -765,34 +774,51 @@ def termination_fn(*dependent_values: int) -> Tuple[int, ...]:
...
```
galvana, line 821: For this iteration, I'm only considering the `erase_after` field defined at the collection level. I originally had it at the dataset level as well but that would have required modifying `DatasetMetadata` in the fideslang repo

src/fides/api/ops/graph/config.py
```diff
@@ -416,6 +416,7 @@ class Collection(BaseModel):
     fields: List[Field]
     # an optional list of collections that this collection must run after
     after: Set[CollectionAddress] = set()
+    erase_after: Set[CollectionAddress] = set()
```
pattisdr, line 419: To clarify, one of the reasons you added a separate "erase_after" param from "after", was not just that this wasn't getting respected in the erasure section, but that the "erase_after" ordering may conflict with "after" ordering?
So it needs to be a legitimately separate param.

src/fides/api/ops/task/graph_task.py
```diff
@@ -765,34 +774,51 @@ def termination_fn(*dependent_values: int) -> Tuple[int, ...]:
...
-        return erasure_update_map
+def _evaluate_erasure_dependencies(
+    t: GraphTask, end_nodes: List[CollectionAddress]
+) -> Set[CollectionAddress]:
+    """
+    Return a set of collection addresses corresponding to collections that need
+    to be erased before the given task. Remove the dependent collection addresses
+    from `end_nodes` so they can be executed in the correct order.
+    """
+    erase_after = t.traversal_node.node.collection.erase_after
```
pattisdr, line 815: I'd look into making sure this is internally consistent. Similar to how traversal.traverse errors if we have nodes that are unreachable, can we throw an error somewhere if you say, A should be erased after B, but B should be erased after A?

src/fides/api/ops/task/graph_task.py
pattisdr: The main thing I'd watch out for here is that we've done some check to make sure this is resolvable. I'd add a big picture check that ensures that we're not saying "erase A before B", and elsewhere saying "erase B before A".

pattisdr (commented):
I think this is a solid approach @galvana. 

My first thought would have been to build a new erasure graph outside of "run_erasure_request", similar to what we do in "traversal.traverse" where we have checks to make sure all nodes are reachable.  However, as part of digging into this,  I discovered that the "after" param is likely not working as intended either in `traversal.traverse`, so that shouldn't be used as an example. Reticketed here https://github.com/ethyca/fides/issues/2638.

So I think what you have here is fine, given that our current erasure graph didn't have any actual dependencies between nodes. You're just tacking some on right before we feed into dask.  Let's just add some validation to make sure that the graph is resolvable, and that the graph modifications for erasures that are restarted after failure still work as expected.

src/fides/api/ops/graph/config.py
galvana: Right that was the main reason, the dataset references and after field work well together but I saw instances where this order conflicted with the erasure order and I would get incomplete traversals.

src/fides/api/ops/task/graph_task.py
galvana: `erasure_prereqs` sounds good

src/fides/api/ops/task/graph_task.py
```diff
@@ -746,16 +751,20 @@ def collect_tasks_fn(
             if not tn.is_root_node():
                 data[tn.address] = GraphTask(tn, resources)
 
+        # We don't rely on the output of `traverse` for the `end_nodes` like we do for
+        # the `access` task since all nodes are considered end nodes unless
+        # they are re-prioritized by the `erase_after` collection field
```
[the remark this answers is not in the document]
galvana: Decided to keep the original traversal results as `access_end_nodes` and use a separate `erasure_end_nodes` for the erasure graph.

src/fides/api/ops/task/graph_task.py
```diff
@@ -765,34 +774,51 @@ def termination_fn(*dependent_values: int) -> Tuple[int, ...]:
...
                     )  # Additionally pass in the original input data we used for the access request. It's helpful in
                     # cases like the EmailConnector where the access request doesn't actually retrieve data.
                     for upstream_key in t.input_keys
                 ],
+                *_evaluate_erasure_dependencies(t, end_nodes),
             )
             for k, t in env.items()
         }
+
         # terminator function waits for all keys
```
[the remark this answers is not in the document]
galvana:
```
# terminator function reads and returns the cached erasure results for the entire erasure traversal
```

src/fides/api/ops/task/graph_task.py
```diff
@@ -765,34 +774,51 @@ def termination_fn(*dependent_values: int) -> Tuple[int, ...]:
...
-        dsk[TERMINATOR_ADDRESS] = (termination_fn, *env.keys())
+        dsk[ROOT_COLLECTION_ADDRESS] = 0
```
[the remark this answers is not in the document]
galvana: To be consistent with the output of each of the erasure tasks, I'll leave a comment

src/fides/api/ops/task/graph_task.py
```diff
@@ -765,34 +774,51 @@ def termination_fn(*dependent_values: int) -> Tuple[int, ...]:
...
+        dsk[TERMINATOR_ADDRESS] = (termination_fn, *end_nodes)
         update_erasure_mapping_from_cache(dsk, resources, start_function)
```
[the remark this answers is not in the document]
galvana: Nice catch, I'll write a test for this to make sure it works as expected

src/fides/api/ops/task/graph_task.py
galvana: Using `getcycle` from `dask.core` to detect cycles and raise a `TraversalError`

```
# using dask.core function to detect cycles in the generated graph
collection_cycle = getcycle(dsk, None)
if collection_cycle:
    raise TraversalError(
        f"The values for the `erase_after` fields caused a cycle in the following collections {collection_cycle}"
    )
```

galvana:
I added tests around skipping erasure tasks with cached data and cycle detection.

pattisdr:
Working on a second pass @galvana...

src/fides/api/ops/task/graph_task.py
```diff
@@ -765,36 +775,62 @@ def termination_fn(*dependent_values: int) -> Tuple[int, ...]:
...
+            raise TraversalError(
+                f"The values for the `erase_after` fields caused a cycle in the following collections {collection_cycle}"
+            )
 
+        v = delayed(get(dsk, TERMINATOR_ADDRESS, num_workers=1))
+        update_cts: Dict[str, int] = v.compute()
+        erasure_update_map = {
+            collection.value: update_cts.get(collection.value, 0) for collection in env
+        }
         return erasure_update_map
```
pattisdr, line 813: You should be able to replace L809-813 with `return v.compute()` if we cache that an erasure happened when the erasure request exits early on a node when **there is no primary key** or the **collection is write-only**:

Consider adding something like:
```python
 self.resources.cache_erasure(
                f"{self.key}", 0
            )  # Cache that the erasure was performed in case we need to restart
```

before these two locations:
https://github.com/ethyca/fides/blob/main/src/fides/api/ops/task/graph_task.py#L529
https://github.com/ethyca/fides/blob/main/src/fides/api/ops/task/graph_task.py#L543

**This will also fix a previous bug discovered in testing here**: If an erasure node is skipped because it doesn't have a primary key (your `products` collection), and that node later fails, on restart, we attempt to run that node again because there was no history in the cache.  This creates multiple sets of execution logs for that node:

```
[('products', 'in_processing'), ('products', 'complete'), 
('orders_to_refunds', 'in_processing'), ('orders_to_refunds', 'complete'), 
('refunds_to_orders', 'in_processing'), ('refunds_to_orders', 'error'), 

('products', 'in_processing'), ('products', 'complete'),
 ('refunds_to_orders', 'in_processing'), ('refunds_to_orders', 'complete'),
 ('orders', 'in_processing'), ('orders', 'complete'), 
('refunds', 'in_processing'), ('refunds', 'complete'), 
('labels', 'in_processing'), ('labels', 'complete')]
```

pattisdr: You could add tests for this in tests/ops/integration_tests/test_execution.py, which has other end-to-end tests for when we're restarting a graph from failure. It's on the mongo and postgres collections which are easier to mock specific node failures.

src/fides/api/ops/task/graph_task.py
pattisdr: Nice use of dask's existing method here

tests/ops/integration_tests/test_integration_erasure_order.py
```diff
@@ -0,0 +1,282 @@
...
+            [saas_erasure_order_connection_config],
+            identity_kwargs,
+            get_cached_data_for_erasures(privacy_request.id),
+            db,
+        )
+
+    assert (
+        f"The values for the `erase_after` fields caused a cycle in the following collections [{dataset_name}:labels, {dataset_name}:orders, {dataset_name}:labels]"
+        in str(exc.value)
+    )
```
pattisdr, line 187: Thanks for this test!!

pattisdr (commented):
OK testing complete, it's looking good @galvana. The main thing that came up here was how to return `v.compute()` directly, and in the process fix an existing bug causing the `products` collection to be re-run on restart even if that collection had already previously been run, causing multiple sets of execution logs.

data/saas/config/saas_erasure_order_config.yml
pattisdr: Do we need to prioritize this - is anyone needing the "after" functionality right now? It's something we advertise..

src/fides/api/ops/task/graph_task.py
```diff
@@ -662,7 +667,7 @@ def collect_tasks_fn(
                 data[tn.address] = GraphTask(tn, resources)
 
         def termination_fn(
-            *dependent_values: List[Row],
+            *_: List[Row],
```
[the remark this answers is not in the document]
galvana: Yes! I missed it during my last pass

data/saas/config/saas_erasure_order_config.yml
galvana: I haven't heard any reports of issues with the "after" functionality. Sentry is the only connector that uses it but it looks like that traversal is enforced by the dataset references instead of the after field.

```
- name: person
  after: [<instance_fides_key>.projects]
  requests:
    read:
      method: GET
      ignore_errors: true
      path: /api/0/projects/<organization_slug>/<project_slug>/users/
      grouped_inputs: [organization_slug, project_slug, query]
      query_params:
        - name: query
          value: email:<query>
      param_values:
        - name: organization_slug
          references:
            - dataset: <instance_fides_key>
              field: projects.organization.slug
              direction: from
        - name: project_slug
          references:
            - dataset: <instance_fides_key>
              field: projects.slug
              direction: from
```

src/fides/api/ops/task/graph_task.py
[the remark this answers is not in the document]
galvana: I'll remove the start_function param

src/fides/api/ops/task/graph_task.py
```diff
@@ -765,36 +775,62 @@ def termination_fn(*dependent_values: int) -> Tuple[int, ...]:
...
             for k, t in env.items()
         }
-        # terminator function waits for all keys
-        dsk[TERMINATOR_ADDRESS] = (termination_fn, *env.keys())
+
+        # root node returns 0 to be consistent with the output of the other erasure tasks
+        dsk[ROOT_COLLECTION_ADDRESS] = 0
+        # terminator function reads and returns the cached erasure results for the entire erasure traversal
+        dsk[TERMINATOR_ADDRESS] = (termination_fn, *erasure_end_nodes)
         update_erasure_mapping_from_cache(dsk, resources, start_function)
```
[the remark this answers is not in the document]
galvana: I'll remove the start_function param

src/fides/api/ops/task/graph_task.py
galvana: Thanks for the tip on how to test this!

src/fides/api/ops/task/graph_task.py
```diff
@@ -540,6 +549,7 @@ def erasure_request(self, retrieved_data: List[Row], *inputs: List[Row]) -> int:
                 ActionType.erasure,
                 ExecutionLogStatus.error,
             )
+            self.resources.cache_erasure(self.key.value, 0)
```
galvana, line 552: Caching erasure result for no primary_key defined and for no write access

tests/ops/integration_tests/test_integration_erasure_order.py
```diff
@@ -0,0 +1,350 @@
...
+    policy: Policy,
+    privacy_request: PrivacyRequest,
+    input_data: Dict[str, List[Any]],
+    secrets: Dict[str, Any],
+) -> List[Row]:
+    return [{f"{node.address.collection}_id": 1}]
+
+
+@register("delete_no_op", [SaaSRequestType.DELETE])
+def delete_no_op(
```
galvana, line 58: Moved these no-ops into the test to avoid having them available for general use like the other custom override functions

tests/ops/integration_tests/test_integration_erasure_order.py
```diff
@@ -0,0 +1,350 @@
...
+
+    temp_masking = CONFIG.execution.masking_strict
+    CONFIG.execution.masking_strict = False
+
+    # mock the mask_data function so we can force an exception on the "refunds_to_orders"
+    # collection to simulate resuming from error
+    def side_effect(node, policy, privacy_request, rows, input_data):
+        if node.address.collection == "refunds_to_orders":
+            raise Exception("Error executing refunds_to_orders task")
+        return 1
```
galvana, line 291: Decided to keep the "resume from error" test here and used `side_effect` to simulate an error and the subsequent fix

galvana:
Thanks @pattisdr for correcting my approach to testing for cached results, and for uncovering additional bugs along the way! Ready for one more review 🤞

data/saas/config/saas_erasure_order_config.yml
pattisdr: OK I suspect no one's running into it because no one's using it yet - i'm going to see if i can get this prioritized

tests/ops/task/test_graph_task.py
```diff
@@ -555,3 +574,88 @@ def test_multiple_rules_targeting_same_field(self, node_fixture):
...
+        assert retrieved_data == []
+        assert input_list == [[]]
+        assert erasure_prereqs == [ROOT_COLLECTION_ADDRESS]
+
+    def test_update_erasure_mapping_from_cache_with_data(self, dsk, task_resource):
+        task_resource.get_all_cached_erasures = lambda: {
+            "dr_1:ds_1": 1
+        }  # a cache with the results of the ds_1 collection erasure
+        update_erasure_mapping_from_cache(dsk, task_resource, start_function)
+        assert dsk[CollectionAddress("dr_1", "ds_1")] == 1
```
pattisdr, line 661: Just these failing tests left that still have the start_function -

pattisdr (approved):
Bravo @galvana

PR #2625, opened 2023-02-16
Custom metadata UI updates

### LINKED ISSUE #2624: Custom metadata UI Polish
### Bug Description

**Taxonomy Management**

1.  When creating a custom field, is the Resource type drop down suppose to be disabled based on the original AC ([Create Custom Field Definition#526](https://github.com/ethyca/fidesplus/issues/526))?
3. When assigning a multi select custom field list, the selected items do not persist between changing Data Categories items. Same bug exists for Data Subjects and Data Uses. **NOTE**: This will be resolved in the next sprint.

**System**

1.  When creating a custom field, is the Resource type drop down suppose to be disabled based on the original AC ([Create Custom Field Definition#526](https://github.com/ethyca/fidesplus/issues/526))?
2. When assigning a custom field value, selected item does not persist for an edited system record when a user refreshes a page or navigates between tabs.

#### Steps to Reproduce

1. _Do this..._
1. _Then This..._

### Expected behavior

_A description of what you expected to happen._

### Screenshots

### Environment

* Version:
* OS:
* Python Version:
* Docker Version:

### Additional context

chriscalhoun1974:
Closes #2624 

### Code Changes

* Added **isRequired** attribute to `CustomSelect` component. When value is true, a red asterisk is displayed to the right of the html label indicating that this input is required.
* Added **isDisabled** attribute to Chakra `Select` component nested in `CustomSelect` parent component
* Updated the `CustomFieldSelector` component to set the **Formik** field value accordingly if the custom field value is a string or an array.
* Added optional **menuPosition** attribute to `CustomSelect` component

### Steps to Confirm

**FidesPlus**
1. Execute `make server` command from Terminal prompt

**Fides Admin UI**

1. Execute `npm i` command from **clients > admin-ui** Terminal prompt
2. Execute `npm run dev` command from same Terminal prompt
3. Open browser tab and navigate to http://localhost:3000
4. Login as `root_user`
5. Click the **Management** top level menu item
6. Within **Taxonomy Management**, click either the Data Categories, Data Subjects, or Data Uses tab
7. Select a given record and click the **Edit** button
8. Click the **Add a custom field** button 
9. For Systems, you can navigate back to the **Home** landing page and click the **Add systems** or  **Manage systems** card

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### Description Of Changes

### DIFF
--- a/clients/admin-ui/src/features/common/form/inputs.tsx
+++ b/clients/admin-ui/src/features/common/form/inputs.tsx
@@ -43,6 +43,7 @@ interface CustomInputProps {
   label: string;
   tooltip?: string;
   variant?: Variant;
+  isRequired?: boolean;
 }
 
 // We allow `undefined` here and leave it up to each component that uses this field
@@ -125,6 +126,7 @@ const SelectInput = ({
   isSearchable,
   isClearable,
   isMulti = false,
+  isDisabled = false,
 }: { fieldName: string; isMulti?: boolean } & Omit<SelectProps, "label">) => {
   const [initialField] = useField(fieldName);
   const field = { ...initialField, value: initialField.value ?? [] };
@@ -193,6 +195,7 @@ const SelectInput = ({
       isClearable={isClearable}
       instanceId={`select-${field.name}`}
       isMulti={isMulti}
+      isDisabled={isDisabled}
     />
   );
 };
@@ -202,6 +205,7 @@ export const CustomTextInput = ({
   tooltip,
   disabled,
   variant = "inline",
+  isRequired = false,
   ...props
 }: CustomInputProps & StringField) => {
   const [initialField, meta] = useField(props);
@@ -213,7 +217,7 @@ export const CustomTextInput = ({
 
   if (variant === "inline") {
     return (
-      <FormControl isInvalid={isInvalid}>
+      <FormControl isInvalid={isInvalid} isRequired={isRequired}>
         <Grid templateColumns="1fr 3fr">
           <Label htmlFor={props.id || props.name}>{label}</Label>
           <Box display="flex" alignItems="center">
@@ -271,6 +275,7 @@ interface SelectProps {
   tooltip?: string;
   options: Option[];
   isDisabled?: boolean;
+  isRequired?: boolean;
   isSearchable?: boolean;
   isClearable?: boolean;
   size?: Size;
@@ -282,6 +287,7 @@ export const CustomSelect = ({
   tooltip,
   options,
   isDisabled,
+  isRequired,
   isSearchable,
   isClearable,
   size = "sm",
@@ -293,7 +299,7 @@ export const CustomSelect = ({
   const isInvalid = !!(meta.touched && meta.error);
   if (variant === "inline") {
     return (
-      <FormControl isInvalid={isInvalid}>
+      <FormControl isInvalid={isInvalid} isRequired={isRequired}>
         <Grid templateColumns="1fr 3fr">
           <Label htmlFor={props.id || props.name} {...labelProps}>
             {label}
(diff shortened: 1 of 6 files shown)


clients/admin-ui/src/features/common/form/inputs.tsx
```diff
@@ -193,6 +195,8 @@ const SelectInput = ({
       isClearable={isClearable}
       instanceId={`select-${field.name}`}
       isMulti={isMulti}
+      isDisabled={isDisabled}
+      menuPosition="fixed"
```
chriscalhoun1974, line 199: Allows the menu to be in a fixed position. Allows menu to be visible.

Screenshot 2023-02-16 at 11 53 10 AM

clients/admin-ui/src/features/plus/plus.slice.ts
```diff
@@ -162,6 +162,8 @@ export const plusApi = createApi({
         params: { show_values },
       }),
       providesTags: ["AllowList"],
+      transformResponse: (allowList: AllowList[]) =>
+        allowList.sort((a, b) => a.name!.localeCompare(b.name!)),
```
ssangervasi, line 166: The name shouldn't be undefined, but just in case I'd use `(a.name ?? "").localeCompare`

clients/admin-ui/src/features/common/form/inputs.tsx
```diff
@@ -310,6 +317,7 @@ export const CustomSelect = ({
               isSearchable={isSearchable === undefined ? isMulti : isSearchable}
               isClearable={isClearable}
               isMulti={isMulti}
+              isDisabled={isDisabled}
```
ssangervasi, line 320: I think this can also go on the FormControl^, unless our [chakra-react-select](https://github.com/csandman/chakra-react-select/tree/v3#isinvalid--default-false:~:text=You%20can%20pass%20also%20pass%20isInvalid%20or%20isDisabled%20to%20a%20wrapping%20%3CFormControl%20/%3E%20and%20it%20will%20output%20their%20corresponding%20%3CInput%20/%3E%20on%20the%20select) isn't hooked up right.

clients/admin-ui/src/features/common/form/inputs.tsx
ssangervasi: I'm giving this some testing because settings this for all of our uses of select could have unexpected layout effects. What was the problem before? Was it getting cut off by the modal edge?

ssangervasi: Yeah, so this is how the dataset editor works without fixed:
good-no-fixed

And this is what happens with the `menuPosition="fixed"`
fun-fixed
(With wider browser window the menu is completely invisible)

Fixed position is a classic monster.

ssangervasi (changes requested):
(just maker sure we don't merge with the fixed position problem)

clients/admin-ui/src/features/common/form/inputs.tsx
ssangervasi: Posible solutions:
1. Add min/max menu height as props that get [passed along to react select](https://react-select.com/props#:~:text=Set%20the%20maximum%20height%20of%20the%20menu.). Then we can avoid the menu extending beyond the modal (or causing scroll) entirely.
2. Add `menuPosition` as a prop so that it can be used here, but is not set to `fixed` by default.

clients/admin-ui/src/features/common/form/inputs.tsx
chriscalhoun1974: Not hooked up correctly. When applied to the FormControl it doesn't disable the Chakra Select component.

clients/admin-ui/src/features/common/form/inputs.tsx
chriscalhoun1974: Issue has been resolved.

PR #2641, opened 2023-02-17
Add warning to 'fides deploy' when installed outside of a virtual environment

NevilleS:
### Code Changes

* [X] Add check to `fides deploy` that logs a warning outside a virtual environment

### Steps to Confirm

* [X] Use `pip install <path/to/fides/repo/dir>` to install the local package and run `fides deploy` to confirm warning"

Example warning:
image
```
~/git/sandbox/fides-deploy-test% fides deploy up
Using default configuration values.
Using default configuration values.
in check_virtualenv(), /usr/local/opt/python@3.9/Frameworks/Python.framework/Versions/3.9, /usr/local/opt/python@3.9/Frameworks/Python.framework/Versions/3.9
WARNING: Your 'fides' CLI is installed outside of a virtual environment at: /usr/local/opt/python@3.9/Frameworks/Python.framework/Versions/3.9
Docker may not have permission to access this path, so if you encounter issues with Docker mounts, try reinstalling 'ethyca-fides' using a virtual environment (see https://docs.python.org/3/library/venv.html).
If you did install using a virtual environment, check your PATH. You may have a global version of 'ethyca-fides' installed that is taking precedence!
Docker version is compatible, starting deploy...
Pulling ethyca/fides image from DockerHub to match local fides version: 2.6.1.post0.dev70
```

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

Users regularly run into mysterious Docker permissioning issues when using `fides deploy` because, on most operating systems, the globally installed Python libraries folder isn't accessible to use for a Docker mount. We rely on this to mount some config files for the docker-compose.yml inside `fides deploy` so this results in mysterious errors like:
```
Error response from daemon: invalid mount config for type "bind": bind source path does not exist: /opt/homebrew/lib/python3.10/site-packages/fides/data/sample_project/privacy_center/config
```

This doesn't really _fix_ that issue - a real fix would likely be to not mount from the lib folder at all (and instead make a local copy of the necessary config files in the directory where `fides deploy` is invoked), but it at least triggers a warning that will let a savvy user notice the issue and recommend they use a `venv` instead.

Another common problem is  we _think_ we're calling the `fides` CLI from within a `venv`, but due to mysterious-reasons-I-cannot-justify it appears that the `venv/bin` folder is given a lower precedence in the `PATH` than globally installed Python packages... which means that when you install things in the `venv` thinking they'll be used, often you end up calling into a globally installed package instead! This warning will also trigger when that happens, so that you can fight with your local shell and do something like:
* uninstall the global `fides` CLI to avoid the footgun
* force the venv to be used like `./venv/bin/fides`

### DIFF
--- a/src/fides/core/deploy.py
+++ b/src/fides/core/deploy.py
@@ -1,3 +1,4 @@
+import sys
 import webbrowser
 from functools import partial
 from os import environ, getcwd, makedirs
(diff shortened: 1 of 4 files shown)


src/fides/core/deploy.py
```diff
@@ -97,6 +98,39 @@ def check_docker_version() -> bool:
...
+    mysterious Docker or Python errors, so warning the user helps get ahead of
+    some of these problems.
+    """
+
+    # Adapted from https://stackoverflow.com/a/58026969/
+    # Detect a virtualenv by checking sys.prefix if the  "base" prefixes match
+    real_prefix = getattr(sys, "real_prefix", None)  # set by older virtualenv
+    base_prefix = getattr(sys, "base_prefix", sys.prefix)
+    running_in_virtualenv = (base_prefix or real_prefix) != sys.prefix
+    print(f"in check_virtualenv(), {sys.base_prefix}, {sys.prefix}")
```
sanders41, line 116: This looks like a leftover print from debugging.

sanders41 (changes requested):
A debugging `print` looks to be left. Otherwise this seems to work well.

src/fides/core/deploy.py
NevilleS: Oh thank you!

sanders41 (approved):

NevilleS:
Ah, interesting - my secret, hidden reason for this PR was to also tinker with the `codecov` reporting. This exposes that the `fides deploy` CLI command itself doesn't get covered in our codecov reports, which is why my patch coverage here is only 92%.

I happen to "know" we run that command in a special check ("CLI Command Checks / Fides-Deploy (pull_request)") but it's interesting to see it flagged here as a failing patch. I like that 😄

PR #2646, opened 2023-02-17
[#2433] Log error class on privacy request failure

### LINKED ISSUE #2433: Log exception class on privacy request creation error
We recently had a case of confusing error logging on privacy request creation. Let's log the exception class [here](https://github.com/ethyca/fides/blob/main/src/fides/api/ops/api/v1/endpoints/privacy_request_endpoints.py#L1675-L1681) too.

We can do this in `dev_mode=True` if it's not already handled there.

seanpreston:
Closes #2433 

### Code Changes

* [ ] Adds the error class for extra debugging help

### DIFF
--- a/src/fides/api/ops/api/v1/endpoints/privacy_request_endpoints.py
+++ b/src/fides/api/ops/api/v1/endpoints/privacy_request_endpoints.py
@@ -1686,7 +1686,7 @@ def create_privacy_request_func(
             )
         except Exception as exc:
             as_string = Pii(str(exc))
-            error_cls = str(exc.__class__)
+            error_cls = str(exc.__class__.__name__)
             logger.error(f"Exception {error_cls}: {as_string}")
             failure = {
                 "message": "This record could not be added",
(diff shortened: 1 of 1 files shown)


src/fides/api/ops/api/v1/endpoints/privacy_request_endpoints.py
```diff
@@ -1685,7 +1685,9 @@ def create_privacy_request_func(
                 detail=exc.args[0],
             )
         except Exception as exc:
-            logger.error("Exception: {}", Pii(str(exc)))
+            as_string = Pii(str(exc))
+            error_cls = str(exc.__class__)
```
sanders41, line 1689: This will print as `<class 'ValueError'>`, we could make it print as just `ValueError` with `__name__`

```suggestion
            error_cls = str(exc.__class__.__name__)
```

sanders41 (approved):
Nice addition. Once small suggestion you can take if you like it, or leave it as is if preferred.

PR #2648, opened 2023-02-17
Updated privacy declarations

allisonking:
Closes https://github.com/ethyca/fidesplus/issues/598

### Code Changes

Okay this is a hefty one but hopefully for good reason!! On the surface, it just changes the privacy declaration form to a nifty accordion setup.

But if you go deeper, it's also trying to be modular so that it can be imported more easily into the datamap-ui.

* Update privacy declaration form to new accordion design
  * Note that this isn't really a straight up accordion design, since the "new" declaration is always outside the accordion
  * Pulled these files into a subfolder of `system` called `privacy-declarations` which has:
    * `PrivacyDeclarationStep.tsx`: the parent component which handles redux calls and the copy surrounding the more complex parts
    * `PrivacyDeclarationManager.tsx`: basically the complex component of Accordion + new declaration form + "Add a Data Use" button. No redux calls. In theory, this component can be plopped into Datamap UI
    * `PrivacyDeclarationForm.tsx`: the individual parts of the privacy declaration form, exported into modular parts. This is because the form is used both as a normal looking form _and_ as an accordion, where the title needs to be an accordion button ⬇️
    * `PrivacyDeclarationAccordion.tsx`: the parts of the form reused for Accordion form.
    * At first I thought it might make sense for datamap UI to pull its parts from `PrivacyDeclarationForm`, and you can refer to `PrivacyDeclarationAccordion` to see how that can be done. But I think it might work to pull `PrivacyDeclarationManager` directly, which would handle all the logic. provided the datamap UI designs don't look any different. If they do, then you may need to use the component parts f rom `PrivacyDeclarationForm` instead.
* No more extended privacy declaration form
* Cypress tests
  * Update existing tests
  * Added new cypress tests to cover data use collisions
  * Pulled system related fixtures into their own folder for better organization

### Steps to Confirm

* [ ] Try adding privacy declarations through the add systems flow as well as the edit system flow

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

https://www.loom.com/share/561080864cb944fc89b7f67ae18fc695


clients/admin-ui/src/features/data-use/data-use.slice.ts
```diff
@@ -24,6 +24,9 @@ export const dataUseApi = createApi({
       transformResponse: (uses: DataUse[]) =>
         uses.sort((a, b) => a.fides_key.localeCompare(b.fides_key)),
     }),
+    getDataUseByKey: build.query<DataUse, string>({
```
allisonking, line 27: turns out I didn't need this, but might still be useful to keep around

clients/admin-ui/src/features/common/form/inputs.tsx
```diff
@@ -137,6 +137,11 @@ interface SelectProps {
   isMulti?: boolean;
   variant?: Variant;
   menuPosition?: MenuPosition;
+  /**
+   * If true, when isMulti=false, the selected value will be rendered as a block,
+   * similar to how the multi values are rendered
+   */
+  singleValueBlock?: boolean;
```
allisonking, line 144: the difference between 

`singleValueBlock = true`
image

`singleValueBlock = false`
image

TheAndrewJackson (approved):
This is awesome! Very modular. It does look like I'll be able to drop `PrivacyDeclarationManager.tsx` right into the datamap ui since the redux calls are passed down from the parent.

So far everything makes sense. I think this is ready to merge into the upstream branch and I'll start testing things out in fidesplus. If I hit any snags we can merge the fixes into the `new-manual-add-design` branch.

PR #2655, opened 2023-02-21
Add Privacy Request Input Sanitization

ThomasLaPiana:
Closes https://github.com/ethyca/security-issues/issues/3
Closes https://github.com/ethyca/security-issues/issues/4

### Code Changes

* [x] create a `PhoneNumber` class with validation that can be reused across the application
* [x] update phone number fields to use the new class
* [x] update email fields to use `EmailStr` instead of `str`
* [x] create a custom `SafeStr` class that sanitizes input values
* [x] add testing for new classes
* [x] Update the privacy request schemas to use `SafeStr` in place of `str`

### Steps to Confirm

* [ ] confirm that the above-linked issues have been fixed
* [ ] poke around at other endpoints

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### Description Of Changes

This PR adds new custom types designed to improve automatic validation and sanitization of user input data. It also applies these new custom types to the privacy request endpoints to remediate the above-linked issues


ThomasLaPiana:
@RobertKeyser @daveqnet would you mind giving  me some examples here of clearly dangerous/invalid user input so I can use them for test cases? I trust y'all here as the experts 😄 thank you!

daveqnet:
> @RobertKeyser @daveqnet would you mind giving me some examples here of clearly dangerous/invalid user input so I can use them for test cases? I trust y'all here as the experts 😄 thank you!

Hmmm, it's a tricky one, as what constitutes dangerous input depends on what the input is being used for. Does the guidance here help: https://owasp.org/www-project-web-security-testing-guide/latest/4-Web_Application_Security_Testing/07-Input_Validation_Testing/ ?

I'll also try to provide you a few examples of some real injection vulnerabilities from non-fides penetration tests to help.

src/fides/api/input_validation.py
```diff
@@ -0,0 +1,56 @@
...
+        yield cls.validate
+
+    @classmethod
+    def validate(cls, value: str) -> str:
+
+        # HTML Escapes
+        value = escape(value)
+
+        # Truncate to 500 characters
+        value = value[:500]
```
NevilleS, line 27: This might surprise us a year from now - can we add an INFO log like `Truncated input string for field 'username' to safe maximum of 500 characters`?

(I'm hoping it might be possible to get the field name from the pydantic `cls` but that might be wishful thinking, if not then at least the breadcrumb trail reminding us of the existence of `InputStr`...)

NevilleS (commented):
Couple comments. But there's lots of other APIs to consider - here are a couple examples:
* `User.username`, `User.first_name`, `User.last_name`
* `System.name`, `System.description` (plus `Dataset`, `DataCategory`... all `Resource` models)
* `Organization` has many string fields (`name`, `description`, `controller.name`...)
*

src/fides/api/ops/schemas/connection_configuration/connection_secrets_email.py
```diff
@@ -10,7 +12,7 @@ class EmailSchema(ConnectionConfigSecretsSchema):
     """Schema to validate the secrets needed for the EmailConnector"""
 
     to_email: str
-    test_email: Optional[str]  # Email to send a connection test email
+    test_email: Optional[EmailStr]  # Email to send a connection test email
```
pattisdr, line 15: Pretty much all the connectors have string inputs - the sass connectors too - The Saas ones aren't as obvious because their schemas are generated dynamically https://github.com/ethyca/fides/blob/main/src/fides/api/ops/schemas/connection_configuration/connection_secrets_saas.py#L99

The connector_params and the external_references become string inputs 
https://github.com/ethyca/fides/blob/main/data/saas/config/firebase_auth_config.yml#L8

pattisdr:
If you're just looking for a list of our API endpoints that accept strings in the request body, I'd assume the majority do, too many to just list here.

Most of the ops schemas are here:
https://github.com/ethyca/fides/tree/main/src/fides/api/ops/schemas

If you're just trying to make a list to work from, you might write a script that goes through the contents of http://localhost:8080/openapi.json, for each request looks for the `requestBody` key and gets the schema from there - lots of recursion.

ThomasLaPiana:
based on the feedback here, it sounds like the safest way to go about things here is to update _every_ string to use `InputStr` and see where that lands us...

Additionally, logging around _when_ values get truncated would be helpful for debugging

src/fides/api/input_validation.py
ThomasLaPiana: Like you said, this feels like an absolute pain of a bug to hunt down in 1 year...I'm going to hard fail instead. Silently (or even logging) a truncation doesn't feel right

galvana:
@ThomasLaPiana I know the unsafe tests aren't being run in Github but let's make sure we run those to make sure we don't break any connector integrations with this change.

src/fides/api/custom_types.py
```diff
@@ -0,0 +1,56 @@
...
+        yield cls.validate
+
+    @classmethod
+    def validate(cls, value: str) -> str:
+
+        # HTML Escapes
+        value = escape(value)
+
+        if len(value) > 500:
+            raise ValueError("Value must be 500 characters or less.")
```
galvana, line 27: Some connectors require inputs longer than 500 characters as part of the setup, what is the reasoning behind limiting the input length?

ThomasLaPiana:
quick update: I can't test the specific fixes I made because I'm not sure how to use the local test env. It now requires privacy request configurations. Will message in slack and try to get this sorted

src/fides/api/custom_types.py
ThomasLaPiana: User input sanitization best practices. We can avoid using this for places that we generally trust I believe, like configuring connectors

ThomasLaPiana:
Patch coverage is lacking but overall coverage is up, we're calling this one a win 🙃

PR #2661, opened 2023-02-22
pc/config: Add tests for back-compat and build validation of config.json

### LINKED ISSUE #2381: Add a test to ensure Privacy Center always remains backwards compatible with all 2.x config.json files

We regularly need to add new features and configuration options to the `config.json` file at the heart of the Privacy Center, but we also regularly forget to ensure these are done in a backwards compatible way. A recent example of this is https://github.com/ethyca/fides/issues/2259.

Since this is easy to miss, automated tests that load in an older version of the "default" `config.json` file to check for compatibility issues will help us catch these bugs earlier, instead of waiting for a staging build to fail.

We should check in an older version of the `config.json` like this one from 2.0.0: https://github.com/ethyca/fides/blob/2.0.0/clients/privacy-center/config/config.json

There should be at least one unit or E2E test that confirms the Privacy Center correctly builds and type-checks that an older version of the `config.json` file is still considered valid. This will force our types (e.g. https://github.com/ethyca/fides/blob/main/clients/privacy-center/types/config.ts) to only introduce optional new values, new features will have to have fallbacks when configurations aren't present, etc.

We can also check for this manually in code review, QA testing, etc. We're already doing these things (and catching issues faster!) but it's more laborious.

Longer term, I also want to remove the reliance on having users manage the `config.json` themselves and replace this with an API so we can programmatically update the format. But that's a much bigger change 😄

ssangervasi:
Closes #2381

### Code Changes

* Extracted build-time config validation to a JS file.
* Added `config/examples` dir for files that we want to stay compatible with.
* Added tests for both those kinds of validations.

### Steps to Confirm

* [ ] Build the PC unchanged
* [ ] Change the `Config` type by adding something new that's required
    * Running `npm run test:ci` should fail on the typechecking step

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### Description Of Changes

There is still more [discussion](https://github.com/ethyca/fides/discussions/2392) to be had about a long term solution for PC configuration, but this should help us avoid the high-friction problem of backwards incompatible changes.


clients/privacy-center/package.json
```diff
@@ -12,7 +12,7 @@
     "format": "prettier --write types/ theme/ pages/ config/ components/ __tests__/",
     "format:ci": "prettier --check types/ theme/ pages/ config/ components/ __tests__/",
     "test": "jest --watch",
-    "test:ci": "jest",
+    "test:ci": "tsc --noEmit && jest",
```
ssangervasi, line 15: Jest just runs babel transformations without checking, so we need this to actually check them. There isn't a handy way to do this with `--watch`.

clients/privacy-center/types/config.ts
```diff
@@ -1,18 +1,24 @@
 import { ConsentValue } from "fides-consent";
 
+export type IdentityInputs = {
+  name?: string;
+  email?: string;
+  phone?: string;
+};
```
ssangervasi, line 7: Now I'm able to make this a _little_ more strict.

allisonking (approved):
hey nice this works great!! nicely done 💯

PR #2666, opened 2023-02-22
Fix support for "redis.user" setting when authenticating to the Redis cache

### LINKED ISSUE #2665: `redis.user` config setting is not used for cache connection
### Bug Description

In the Fides codebase we use the Redis server for two different purposes:

1. Fides Cache (`redis-py` client): https://github.com/ethyca/fides/blob/main/src/fides/api/ops/util/cache.py
2. Fides Worker Backend (`celery` client): https://github.com/ethyca/fides/blob/main/src/fides/api/ops/tasks/__init__.py

Because of the differences between the client libraries these two use, they are configured slightly differently: `redis-py` uses a client class, and `celery` uses a connection URL. However, they both share the same set of configuration settings: https://github.com/ethyca/fides/blob/main/src/fides/core/config/redis_settings.py

From inspection of the code, it looks like not every client supports all the config settings, which is going to lead to some confusion in the future. See below for a detailed list, but for this ticket there is one key setting that is going to surprise us if we don't normalize support across clients: `redis.user` 

We _need_ to add support for `redis.user` so the core _connection_ settings are the same between the two clients and so that we can support the use of Redis ACLs for security. This bug isn't obvious to anyone yet, since Redis doesn't _require_ usernames, so it's hidden from sight. The only way to truly reproduce this would be to add a Redis ACL to a server: https://redis.io/docs/management/security/acl

We're planning to start using Redis ACLs for our own production clusters in the near future, so this'll be important to add in advance of that.

#### Detailed "Redis" Config Settings Support

For an overall view of what Redis settings are supported by which clients, see the following table:

| Setting | Fides Cache | Fides Worker |
|---|---|---|
| host | ✅  | ✅|
| port | ✅| ✅|
| user |❌ |✅ |
| password | ✅| ✅|
| charset | ✅| ❓ |
| decode_responses |✅ | ❓|
| default_ttl_seconds | ✅ | n/a |
| identity_verification_code_ttl_seconds |✅ | n/a |
| db_index | ✅| ✅|
| enabled | ✅| ❌ |
| ssl |✅ | ✅|
| ssl_cert_reqs | ✅| ✅|
| connection_url | ❌  |✅ |

While _ideally_ it might be nice to have exact parity between the two, some of the settings are client-specific (e.g. default_ttl_seconds, identity_verification_code_ttl_seconds) and others probably shouldn't even be configurable (e.g. charset, decode_responses)... this ticket is focused just on the important `user` setting.

#### Steps to Reproduce

1. Setup a Redis server instance
2. Configure an ACL with a username & password
3. Attempt to connect to the username & password with Fides

### Expected behavior

Fides should connect to Redis server and `AUTH` with both a username and password.

### Screenshots

n/a

### Environment

* Version: 2.7.0
* OS: n/a
* Python Version: n/a
* Docker Version: n/a

NevilleS:
Closes #2665

### Code Changes

* [X] Update cache.py client with optional `username` arg
* [X] Fix unexpected default Redis password: `testpassword`
* [X] Fix support for unauthenticated Redis servers

### Steps to Confirm

I used my local Docker setup to test the following configs:
* [x] Redis server with no auth:
  * [x] Fides with no auth (neither `redis.password` nor `redis.user`) connects
  * [x] Fides with `redis.password` does not connect
  * [x] Fides with `redis.password` and `redis.user` does not connect
* [x] Redis server with just password auth (e.g. `redis-server --requirepass <password>`):
  * [x] Fides with correct `redis.password` and no `redis.user` connects
  * [X] Fides with correct `redis.password` and incorrect `redis.user` does not connect
  * [x] Fides with incorrect `redis.password` does not connect
  * [x] Fides with no auth (neither `redis.password` nor `redis.user`) does not connect
* [x] Redis server with an ACL (e.g. `redis-server <path/to/redis.conf>`):
  * [x] Fides with correct `redis.password` and `redis.user` connects
  * [x] Fides with no auth (neither `redis.password` nor `redis.user`) does not connect
  * [x] Fides with incorrect `redis.password` and `redis.user` does not connect

For testing the Redis server with an ACL, I created this file at `./docker/redis/redis.conf`:
```
user redisadmin on ~* +@all >redispassword
```

I then modified `docker-compose.yml` to configure the `redis` service as:
```
  redis:
    image: "redis:6.2.5-alpine"
    command: redis-server /usr/local/etc/redis/redis.conf
    expose:
      - 6379
    ports:
      - "0.0.0.0:6379:6379"
    volumes:
      - type: bind
        source: ./docker/redis
        target: /usr/local/etc/redis
        read_only: False
```

And updated the `.fides/fides.toml` settings like:
```
[redis]
host = "redis"
user = "redisadmin"
password = "redispassword"
port = 6379
charset = "utf8"
default_ttl_seconds = 604800
db_index = 0
ssl = false
ssl_cert_reqs = "required"
```

### Pre-Merge Checklist

* Documentation:
  * Added: https://github.com/ethyca/fides/issues/2673
* [X] Update `CHANGELOG.md`

### DIFF
--- a/.fides/fides.toml
+++ b/.fides/fides.toml
@@ -22,7 +22,7 @@ analytics_opt_out = false
 
 [redis]
 host = "redis"
-password = "testpassword"
+password = "redispassword"
 port = 6379
 charset = "utf8"
 default_ttl_seconds = 604800
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,11 +17,17 @@ The types of changes are:
 ## [Unreleased](https://github.com/ethyca/fides/compare/2.7.0...main)
 
 ### Added
+
 * Add API support for messaging config properties [#2551](https://github.com/ethyca/fides/pull/2551)
 
 ### Changed
 
 * Add warning to 'fides deploy' when installed outside of a virtual environment [#2641](https://github.com/ethyca/fides/pull/2641)
+* Removed unexpected default Redis password [#2666](https://github.com/ethyca/fides/pull/2666)
+
+### Fixed
+
+* Fix support for "redis.user" setting when authenticating to the Redis cache [#2666](https://github.com/ethyca/fides/pull/2666)
 
 ## [2.7.0](https://github.com/ethyca/fides/compare/2.6.6...2.7.0)
 
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -127,13 +127,23 @@ services:
 
   redis:
     image: "redis:6.2.5-alpine"
-    command: redis-server --requirepass testpassword
-    environment:
-      - REDIS_PASSWORD=testpassword
+    # AUTH option #1: no authentication at all
+    # command: redis-server
+    # AUTH option #2: require password
+    command: redis-server --requirepass redispassword
+    # AUTH option #3: Redis ACL defined in redis.conf
+    # command: redis-server /usr/local/etc/redis/redis.conf
     expose:
       - 6379
     ports:
       - "0.0.0.0:6379:6379"
+    volumes:
+      # Mount a redis.conf file for configuration
+      # NOTE: Only used by "AUTH option #3" above!
+      - type: bind
+        source: ./docker/redis
+        target: /usr/local/etc/redis
+        read_only: False
 
 volumes:
   postgres: null
--- a/noxfiles/dev_nox.py
+++ b/noxfiles/dev_nox.py
@@ -21,7 +21,26 @@ from utils_nox import COMPOSE_DOWN_VOLUMES
 
 @nox_session()
 def dev(session: Session) -> None:
-    """Spin up the application. Uses positional arguments for additional features."""
+    """
+    Spin up the Fides webserver in development mode alongside it's Postgres
+    database and Redis cache. Use positional arguments to run other services
+    like privacy center, shell, admin UI, etc. (see usage for examples)
+
+    Usage:
+      'nox -s dev' - runs the Fides weserver, database, and cache
+      'nox -s dev -- shell' - also open a shell on the Fides webserver
+      'nox -s dev -- ui' - also build and run the Admin UI
+      'nox -s dev -- pc' - also build and run the Privacy Center
+      'nox -s dev -- remote_debug' - run with remote debugging enabled (see docker-compose.remote-debug.yml)
+      'nox -s dev -- worker' - also run a Fides worker
+      'nox -s dev -- child' - also run a Fides child node
+      'nox -s dev -- <datastore>' - also run a test datastore (e.g. 'mssql', 'mongodb')
+
+    Note that you can combine any of the above arguments together, for example:
+      'nox -s dev -- shell ui pc'
+
+    See noxfiles/dev_nox.py for more info
+    """
 
     build(session, "dev")
     session.notify("teardown")
--- a/src/fides/api/ops/util/cache.py
+++ b/src/fides/api/ops/util/cache.py
@@ -172,6 +172,7 @@ def get_cache() -> FidesopsRedis:
             host=CONFIG.redis.host,
             port=CONFIG.redis.port,
             db=CONFIG.redis.db_index,
+            username=CONFIG.redis.user,
             password=CONFIG.redis.password,
             ssl=CONFIG.redis.ssl,
             ssl_cert_reqs=CONFIG.redis.ssl_cert_reqs,
--- a/src/fides/core/config/redis_settings.py
+++ b/src/fides/core/config/redis_settings.py
@@ -14,7 +14,7 @@ class RedisSettings(FidesSettings):
     host: str = "redis"
     port: int = 6379
     user: Optional[str] = None
-    password: str = "testpassword"
+    password: str = None
     charset: str = "utf8"
     decode_responses: bool = True
     default_ttl_seconds: int = 604800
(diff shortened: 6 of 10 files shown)


noxfiles/dev_nox.py
```diff
@@ -21,7 +21,26 @@
...
+
+    Usage:
+      'nox -s dev' - runs the Fides weserver, database, and cache
+      'nox -s dev -- shell' - also open a shell on the Fides webserver
+      'nox -s dev -- ui' - also build and run the Admin UI
+      'nox -s dev -- pc' - also build and run the Privacy Center
+      'nox -s dev -- remote_debug' - run with remote debugging enabled (see docker-compose.remote-debug.yml)
+      'nox -s dev -- worker' - also run a Fides worker
+      'nox -s dev -- child' - also run a Fides child node
+      'nox -s dev -- <datastore>' - also run a test datastore (e.g. 'mssql', 'mongodb')
```
seanpreston, line 37: 🙏 thankyou for this

NevilleS: > Leave this world a little better than you found it.
> \- Robert Baden-Powell

NevilleS:
When I configure my local Redis to enforce a username/password with an ACL, here's an example error I get:
```
fides-fides-1  | 2023-02-22 22:13:16.772 [ERROR] (main:setup_server:280): Connection to cache failed: WRONGPASS invalid username-password pair or user is disabled.
```

This tells me that I've clearly enabled the ACL correctly locally. Just doing a few more tests here but this feels good to go...

.fides/fides.toml
```diff
@@ -22,7 +22,7 @@ analytics_opt_out = false
 
 [redis]
 host = "redis"
-password = "testpassword"
+password = "redispassword"
```
NevilleS, line 25: Changed this to be extra careful - we were setting a default password in the Pydantic config (which is surprising!), so I wanted to be doubly sure that we weren't relying on that anywhere in our code.

CHANGELOG.md
```diff
@@ -17,11 +17,17 @@ The types of changes are:
 ## [Unreleased](https://github.com/ethyca/fides/compare/2.7.0...main)
 
 ### Added
+
 * Add API support for messaging config properties [#2551](https://github.com/ethyca/fides/pull/2551)
 
 ### Changed
 
 * Add warning to 'fides deploy' when installed outside of a virtual environment [#2641](https://github.com/ethyca/fides/pull/2641)
+* Removed unexpected default Redis password [#2666](https://github.com/ethyca/fides/pull/2666)
```
NevilleS, line 26: This was definitely scope creep on this change - but in keeping with the theme here, I'm trying to stay ahead of potential Redis-related gotchas.

docker-compose.yml
```diff
@@ -127,13 +127,23 @@ services:
...
     image: "redis:6.2.5-alpine"
-    command: redis-server --requirepass testpassword
-    environment:
-      - REDIS_PASSWORD=testpassword
+    # AUTH option #1: no authentication at all
+    # command: redis-server
+    # AUTH option #2: require password
+    command: redis-server --requirepass redispassword
+    # AUTH option #3: Redis ACL defined in redis.conf
+    # command: redis-server /usr/local/etc/redis/redis.conf
```
NevilleS, line 135: Just checking this in, commented out, for future reference. I used this command for my manual testing.

src/fides/api/ops/util/cache.py
```diff
@@ -172,6 +172,7 @@ def get_cache() -> FidesopsRedis:
             host=CONFIG.redis.host,
             port=CONFIG.redis.port,
             db=CONFIG.redis.db_index,
+            username=CONFIG.redis.user,
```
NevilleS, line 175: This is really the only fix truly needed...

src/fides/core/config/redis_settings.py
```diff
@@ -13,8 +13,8 @@ class RedisSettings(FidesSettings):
 
     host: str = "redis"
     port: int = 6379
-    user: Optional[str] = ""
-    password: str = "testpassword"
+    user: Optional[str] = None
+    password: str = None
```
NevilleS, line 17: Defaulting this to `testpassword` is surprising. If I boot up Fides and connect to Redis, I don't expect it to try to authenticate with the password "testpassword" and I'd take forever to figure that out in practice...

src/fides/core/config/redis_settings.py
```diff
@@ -37,17 +37,25 @@ def assemble_connection_url(
...
             connection_protocol = "rediss"
             cert_reqs = values.get("ssl_cert_reqs", "none")
             params = f"?ssl_cert_reqs={quote_plus(cert_reqs)}"
 
-        return f"{connection_protocol}://{quote_plus(values.get('user', ''))}:{quote_plus(values.get('password', ''))}@{values.get('host', '')}:{values.get('port', '')}/{db_index}{params}"
+        # Configure a basic auth prefix if either user or password is provided, e.g.
+        # redis://<user>:<password>@<host>
+        auth_prefix = ""
+        if password or user:
+            auth_prefix = f"{quote_plus(user)}:{quote_plus(password)}@"
```
NevilleS, line 56: Added this to handle the case where someone doesn't have any auth configured on their redis server at all.

NevilleS (commented):
Couple comments here - this got away from me a bit, but it should be good now.

src/fides/core/config/redis_settings.py
seanpreston: Should we also mark this as `Optional[str]` since it defaults to `None`?

NevilleS: Yes, in fact `mypy` said the same thing 👍

Fixed

seanpreston (approved):
Thanks @NevilleS

PR #2668, opened 2023-02-22
pc/cy: Use waitUntil to check for cookie consent updates

ssangervasi:
✅ Trying to resolve a build flake  ♻️

I was able to reproduce by enabling network throttling my Cypress browser. Then I was able to get `waitUntil` to resolve the issue.


chriscalhoun1974 (approved):
Nice work!

PR #2669, opened 2023-02-22
Combine conftest.py files

### LINKED ISSUE #2174: Refactor `tests/` to use a single `conftest.py`

sanders41:
Closes #2174

### Code Changes

* Where possible combined the `conftest.py` files from `ctl`, `lib`, and `ops` into a single `conftest.py` file.

### Steps to Confirm

* [ ] All tests pass

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

Because of the difference in database connections between `ctl` and `ops` I was not able to combine those fixtures. In addition some of the tests in `ctl` seem to depend on previous tests having already run and populating the database with values. Because of this the fixture to reset data between tests could not be used in ctl.

There are some fixtures here that could be combined into one. I started down that path and realized there was going to be potentially a huge amount of files changed. Because of that I wanted to get this merged as a smaller change to make sure it was working, and can follow up with the larger change of merging duplicate fixtures.


sanders41:
The ctl tests are freezing because of a database deadlock. This originally happened when I tried keeping the database connections in one file. Splitting them out fix it, but after merging main it is unhappy again. I'm looking into the new cause/solution.

sanders41:
> The ctl tests are freezing because of a database deadlock. This originally happened when I tried keeping the database connections in one file. Splitting them out fix it, but after merging main it is unhappy again. I'm looking into the new cause/solution.

Issue resolved.

tests/ctl/conftest.py
```diff
@@ -67,321 +38,26 @@ def monkeypatch_requests(test_client, monkeysession) -> None:
...
-    for manifest in test_manifests.keys():
-        nested_manifest_dir = f"{manifest_dir}/{manifest}"
-        os.mkdir(nested_manifest_dir)
-        with open(f"{nested_manifest_dir}/{manifest}.yml", "w") as manifest_file:
-            yaml.dump(test_manifests[manifest], manifest_file)
-    return manifest_dir
-
-
 @pytest.fixture
-def db() -> Generator:
```
adamsachs, line 331: not a big deal, but just curious - why the removal of the return type annotation here? does it cause problems?

tests/lib/conftest.py
```diff
@@ -1,155 +1,68 @@
...
 
+@pytest.fixture(scope="session")
+def db(api_client, config):
+    """Return a connection to the test DB"""
     # Create the test DB engine
     assert config.test_mode
+    assert requests.post != api_client.post
     engine = get_db_engine(
-        database_uri=config.database.sqlalchemy_database_uri,
+        database_uri=config.database.sqlalchemy_test_database_uri,
```
adamsachs, line 29: had this just been a miss earlier? don't _think_ it should cause any problems, especially if all tests are passing, just wanna make sure we don't accidentally break any assumptions we had :)

tests/lib/conftest.py
```diff
@@ -1,155 +1,68 @@
...
-            "first_name": "Test",
-            "last_name": "User",
-        },
-    )
-    oauth_client.user_id = user.id
-    oauth_client.save(db=db)
-    yield user
-    user.delete(db)
+@pytest.fixture(autouse=True)
+def clear_db_tables(db):
```
adamsachs, line 44: nice that you've added this here

tests/lib/conftest.py
```diff
@@ -1,155 +1,68 @@
...
+    assert config.test_mode
+    assert requests.post != api_client.post
+    yield api_client.post(url=f"{config.cli.server_url}/v1/admin/db/reset")
 
-@pytest.fixture
-def db(config):
-    """Yield a connection to the test DB."""
-    # Included so that `AccessManualWebhook` can be located when
-    # `ConnectionConfig` is instantiated.
-    from fides.api.ops.models.manual_webhook import (  # pylint: disable=unused-import
```
adamsachs, line 39: is there any worry to losing this? not sure what it's used for, and if tests are passing it should be fine - but just curious.

adamsachs (approved):
generally looks good to me! a few minor comments/questions on the code, nothing really that needs to be addressed as far as i can see.

looks like you've got a bunch of merge conflicts again with `main` that you need to resolve, sorry for not getting to this sooner :( but i'll approve this now preemptively.

a few other non-blocking notes:

- should we create a follow up ticket for this?
> can follow up with the larger change of merging duplicate fixtures



- separately, should we run the unsafe test suite, just since you've touched things used by those tests and to make sure things aren't terribly broken? we can expect some intermittent failures on the saas external tests so we can use our discretion to ignore any failures there if they seem unrelated to your changes.

tests/ctl/conftest.py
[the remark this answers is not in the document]
sanders41: It just doesn't do anything at all so I removed it. We don't type check the tests (which makes sense), and it isn't called as a regular function so it doesn't do anything for IDEs so it really has no purpose.

tests/lib/conftest.py
sanders41: I originally put everything into one conftest.py file, including the database fixtures. This caused problems because of the different kinds of connections, and when I split them back out I just copied the one from ops into here because I thought they were the same. `get_db_engine` sets `database_url` to `sqlalchemy_test_database_uri` by default when in test mode so it seems like a mistake from the start to have set it to `sqlalchemy_database_uri`

tests/lib/conftest.py
sanders41: I tried to add it everywhere but some of the ctl tests broke with it in place 😞

tests/lib/conftest.py
sanders41: It seemed to me it was imported by mistake. I never got used for anything.

sanders41:
> should we create a follow up ticket for this?

Yes, follow up created here #2688

> separately, should we run the unsafe test suite, just since you've touched things used by those tests and to make sure things aren't terribly broken? we can expect some intermittent failures on the saas external tests so we can use our discretion to ignore any failures there if they seem unrelated to your changes.

Yes, makes a lot of sense. I set them to run this time.

sanders41:
External tests are still getting skipped even after adding the label. I tried removing and re-adding it but that didn't kick them off either. Any other ideas on how to get them to run?

adamsachs:
>External tests are still getting skipped even after adding the label. I tried removing and re-adding it but that didn't kick them off either. Any other ideas on how to get them to run?

hmm not off the top of my head, pretty sure the removing and re-adding is always what i've done to trigger it when it's not cooperating... 🤔

sanders41:
I ran the external tests locally and they failed as expected. All the failures look permissions related and not fixture related. I'm not sure of a better test I can do since I can't get them to start here. Thoughts?

adamsachs:
> I ran the external tests locally and they failed as expected. All the failures look permissions related and not fixture related. I'm not sure of a better test I can do since I can't get them to start here. Thoughts?

just noting we synced up offline and i passed along the documentation about how to set up vault integration locally and @sanders41 is going to see if he can get that working. generally though we feel pretty confident this change is good to go.

PR #2671, opened 2023-02-22
[Backend] Add Role-Based Permissions [#2606]

### LINKED ISSUE #2606: Add Role-Based Permissions Foundation
This ticket would add the foundation to honor role-based permissions while still honoring scope-based ones.

- Add the new columns to fidesuserpermissions and client, and new table ~systemmanagers~
- Update verify_oauth_client to work with tokens that have scopes OR roles
    - Allow the action if the user has permission due to a direct scope OR a scope via a role
- When creating auth tokens, additionally add both current_roles and ~current_systems~ to token
- Add the Role->Scopes mapping for admin, viewer, and privacy_request_manager roles
- Adjust ClientDetail methods to work with both roles and scopes.
- Update existing permissions endpoint to optionally specify roles and make specifying scopes optional.  Allowed roles here are admin, viewer, and privacy_request_manager.

pattisdr:
Closes #2606 

### Code Changes

 - Update the permission check in `verify_oauth_client` to see if you have the scope from one of two means: either a scope you were assigned directly OR if  your role(s) are associated with that scope.
- Add `client.roles` and `fidesuserpermission.roles` column. Default values are empty lists and scopes are changed likewise.
    - The FidesUserPermission.roles column persists the user's roles, while the client.roles column is temporary, for login.
- Add a new endpoint GET `/oauth/role` to surface the possible roles and their associated scopes to which you can be assigned
- Update `POST /oauth/token` to add an admin role to the the token if the client is the root client.
- Update PATCH user permissions to be able to update both roles and/or scopes.  Current UI can just update scopes, but we're moving to a UI that updates roles.  Technically, we'll still support updating scopes in the backend.
- Update GET user permissions to return both the scope registry and admin roles if this is the root user.  Also we surface the roles the user has in addition to their direct scopes (current behavior)
- Update CLI command "fides user permissions" to return both roles and directly-assigned scopes
- Update CLI command "fides user create" which creates a user with full permissions to also have the admin role.
- Add `ExecutionSettings.root_user_roles`
- Fix the default values on `Client.scopes` and `FidesUserPermission.scopes` - this did not appear to be working as intended, but we weren't hitting the error in practice.
- Try to ensure that if a client's roles or scopes are accessed, it returns an empty list instead of null if they don't exist.
- Add first draft of role -> scope mapping.
- Remove old, unused privileges code.  We were once intending to group scopes into "privileges" but never finished that work.  This replaces that effort.
- On login, give the client both root_user_scopes and root_user_roles if it's the root user. Persist permission scopes and roles to the client if the client doesn't exist.
- Prevent logging in without any roles or scopes.

### Steps to Confirm

* [ ] Create a user via the API with a client that has permissions to do so:
POST {{host}}/user
```json
{
    "username": "test_user",
    "password": "Testpassword1!"

}
```
* [ ] Update the role of that user, one of `admin`,  `viewer_and_privacy_request_manager` and `viewer`, `privacy_request_manager`:

PUT {{host}}/user/{{user_id}}/permission
```json
{
    "id": "{user_id}",
    "roles": ["viewer"]
}

``` 
* [ ] Login as that user in the admin UI
* [ ] Navigate through various parts of the UI with different roles and observe in the network tab that some routes are restricted depending on your role.  Note that which scopes go with which role are still in flux.

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

Optionally add the ability to grant larger roles directly to users instead of just individual scopes. Roles will be associated with a list of scopes.

Currently supporting 4 roles:
- Admin role: has all scopes
- Viewer + Privacy Request Manager: Has all read permissions (minus user permissions read) + Privacy Request Scopes
- Viewer: Has read scopes (minus user permissions read)
- Privacy Request Manager: Only privacy request scopes


src/fides/api/ops/api/v1/endpoints/user_permission_endpoints.py
```diff
@@ -79,13 +85,21 @@ def update_user_permissions(
...
+
     if user.client:
-        user.client.update(db=db, data={"scopes": permissions.scopes})
+        user.client.update(
+            db=db, data={"scopes": permissions.scopes, "roles": permissions.roles}
+        )
     return user.permissions.update(  # type: ignore[attr-defined]
         db=db,
-        data={"id": user.permissions.id, "user_id": user_id, **permissions.dict()},  # type: ignore[attr-defined]
+        data={"id": user.permissions.id, "user_id": user_id, "scopes": permissions.scopes, "roles": permissions.roles},  # type: ignore[attr-defined]
```
pattisdr, line 102: This is the endpoint that allows you to update a user's permissions. On `main` you can add scopes to users, this additionally allows you to assign roles (which are associated with scopes).   We are still going to support being assigned both scopes and roles.  (For one, you can currently only assign scopes directly in the UI, but two, Sean said we should continue to allow scopes to be assigned for API-driven workflows).

src/fides/api/ops/api/v1/endpoints/user_permission_endpoints.py
```diff
@@ -109,6 +123,7 @@ async def get_user_permissions(
                 id=CONFIG.security.oauth_root_client_id,
                 user_id=CONFIG.security.oauth_root_client_id,
                 scopes=SCOPE_REGISTRY,
+                roles=[ADMIN],
```
pattisdr, line 126: For this root client, I'm adding the admin role in addition to all the scopes - either avenue gives the root user access to all the endpoints.   In case the user workflow is later deprecated to just assigning roles directly, I thought we should start assigning roles now.

src/fides/lib/models/client.py
```diff
@@ -104,6 +113,7 @@ def create_access_code_jwe(self, encryption_key: str) -> str:
             JWE_PAYLOAD_CLIENT_ID: self.id,
             JWE_PAYLOAD_SCOPES: self.scopes,
             JWE_ISSUED_AT: datetime.now().isoformat(),
+            JWE_PAYLOAD_ROLES: self.roles,
```
pattisdr, line 116: This additionally adds the roles to the access token for the client.  We wait until `verify_oauth_util` to check to see if the token has scopes directly or a scope via a role in order to access the endpoint.  This way, if we scopes associated with a role change, the tokens are still valid.

src/fides/lib/models/fides_user_permissions.py
```diff
@@ -1,34 +1,17 @@
...
-
-    @property
-    def privileges(self) -> Tuple[str, ...]:
-        """Return the big-picture privileges a user has based on their individual scopes"""
-        user_privileges: List[str] = []
-        for privilege, required_scopes in privileges.items():
-            if required_scopes.issubset(self.scopes):
-                user_privileges.append(privilege)
-
-        return tuple(user_privileges)
```
pattisdr, line 34: Last year we also briefly started down the path of grouping scopes into "privileges" -  a similar concept.  This effort was de-prioritized, so this old code is not being used.

src/fides/lib/models/fides_user_permissions.py
```diff
@@ -1,34 +1,17 @@
...
     """The DB ORM model for FidesUserPermissions"""
 
     user_id = Column(String, ForeignKey(FidesUser.id), nullable=False, unique=True)
-    # escaping curly braces requires doubling them. Not a "\". So {{{test123}}}
-    # renders as {test123}
-    scopes = Column(
-        ARRAY(String), nullable=False, default=f"{{{PRIVACY_REQUEST_READ}}}"
-    )
+    scopes = Column(ARRAY(String), nullable=True, server_default="{}", default=dict)
+    roles = Column(ARRAY(String), nullable=True, server_default="{}", default=dict)
```
pattisdr, line 13: I'm making scopes nullable here - the user can have scopes or roles.   I've also updated the default value - the formatting of `{{{PRIVACY_REQUEST_READ}}}` doesn't work as intended, but wasn't obvious because we were always supplying scopes when creating permissions - it was never left as a default.  But this default could create scopes as ["{", "P", "R", "I", "V", "A", "C", "Y"...]

src/fides/lib/oauth/privileges.py
```diff
@@ -1,61 +0,0 @@
...
-        DATASET_CREATE_OR_UPDATE,
-        DATASET_DELETE,
-        DATASET_READ,
-        SAAS_CONFIG_CREATE_OR_UPDATE,
-        SAAS_CONFIG_DELETE,
-        SAAS_CONFIG_READ,
-    },
-    VIEW_USERS: {USER_READ},
-    MANAGE_USERS: {USER_CREATE, USER_DELETE, USER_READ},
-}
```
pattisdr, line 61: Removing the unused privileges work.

src/fides/lib/oauth/api/routes/user_endpoints.py
```diff
@@ -81,7 +81,11 @@ def create_user(
     user = FidesUser.create(db=db, data=user_data.dict())
     logger.info("Created user with id: '{}'.", user.id)
     FidesUserPermissions.create(
-        db=db, data={"user_id": user.id, "scopes": [PRIVACY_REQUEST_READ]}
+        db=db,
+        data={
+            "user_id": user.id,
+            "scopes": [PRIVACY_REQUEST_READ],
+        },  # TODO - Change this to Viewer Role by default?
```
pattisdr, line 88: Currently when creating a user, they are given "privacy request read" scopes by default.  Once we have the UI for assigning roles in place, we might change the default to be a "viewer" role.

src/fides/lib/oauth/roles.py
```diff
@@ -0,0 +1,100 @@
...
+    PRIVACY_REQUEST_REVIEW,
+    PRIVACY_REQUEST_READ,
+    PRIVACY_REQUEST_DELETE,
+    PRIVACY_REQUEST_CALLBACK_RESUME,
+    PRIVACY_REQUEST_NOTIFICATIONS_CREATE_OR_UPDATE,
+    PRIVACY_REQUEST_NOTIFICATIONS_READ,
+    PRIVACY_REQUEST_TRANSFER,
+    PRIVACY_REQUEST_UPLOAD_DATA,
+    PRIVACY_REQUEST_VIEW_DATA,
+]
```
pattisdr, line 58: Which roles go with which scopes are still subject to change.  For example, I've given the Privacy Request Manager all the privacy request scopes, but there are likely requests the UI makes to load various parts of the UI that this role requires.

For example, there's a set of endpoints that aren't really related to resources created by the customer, but more system settings, like a list of all our connection types that might be loaded on various pages that make a tricky UI experience for the privacy request manager.

src/fides/lib/oauth/schemas/user_permission.py
```diff
@@ -1,22 +0,0 @@
...
-    """Data required to edit a FidesUserPermissions record."""
-
-    id: str
-
-
-class UserPermissionsResponse(UserPermissionsCreate):
-    """Response after creating, editing, or retrieving a FidesUserPermissions record."""
-
-    id: str
-    user_id: str
```
pattisdr, line 22: We had two user permissions schemas, one brought back from fideslib.  Having two sets was inadvertently causing a bug where there was no validation on scopes when editing user permissions.  I've removed one set here.

src/fides/lib/oauth/api/routes/user_endpoints.py
```diff
@@ -175,6 +179,7 @@ def user_login(
             object_id=config.security.oauth_root_client_id,
             config=config,
             scopes=config.security.root_user_scopes,
+            roles=config.security.root_user_roles,
```
pattisdr, line 182: On login, adding both the root user scopes and root user roles to the ClientDetail for the root user

src/fides/lib/oauth/api/routes/user_endpoints.py
```diff
@@ -241,6 +246,7 @@ def perform_login(
             client_id_byte_length,
             client_secret_btye_length,
             scopes=user.permissions.scopes,  # type: ignore
+            roles=user.permissions.roles,  # type: ignore
```
pattisdr, line 249: Transferring both scopes and roles from the user's Permission's object to the user's ClientDetail on login.  The token which is later created from this client then has both scopes and roles.

src/fides/api/ops/schemas/user_permission.py
```diff
@@ -31,10 +31,29 @@ def validate_scopes(cls, scopes: List[str]) -> List[str]:
...
+            )
+        return roles
+
+
+class UserPermissionsEdit(UserPermissionsCreate):
+    """Data required to edit a FidesUserPermissions record."""
+
+    id: Optional[
+        str
+    ]  # I don't think this should be in the request body, so making it optional.
```
pattisdr, line 52: The UI currently sends an id in with the edit user permission's request.  I don't think this needs to be there, so I'm making it optional here and am ensuring we ignore it.  There was no actual validation on that value, so if we weren't ignoring it, it could technically replace the id of the user permissions object.

It's fine as it's not being used, but at some point, it would be nice to remove this id from the request body in the UI too.

src/fides/core/config/security_settings.py
```diff
@@ -30,6 +31,7 @@ class SecuritySettings(FidesSettings):
     """Configuration settings for Security variables."""
 
     root_user_scopes: Optional[List[str]] = SCOPE_REGISTRY
+    root_user_roles: Optional[List[str]] = [ADMIN]
```
pattisdr, line 34: This is a new config value.  Thinking about it, there are a lot of places in the code where we assign the scope registry to a root user and now the admin role.  Instead, we might pull from this and the existing config value.

src/fides/core/user.py
```diff
@@ -116,7 +121,11 @@ def create_command(
     )
     user_id = user_response.json()["id"]
     update_user_permissions(
-        user_id=user_id, scopes=SCOPES, auth_header=auth_header, server_url=server_url
+        user_id=user_id,
+        scopes=SCOPES,
+        auth_header=auth_header,
+        server_url=server_url,
+        roles=[ADMIN],
```
pattisdr, line 128: Adjusting the CLI command to also assign all scopes and admin role here when creating a user with full permissions.

src/fides/core/user.py
```diff
@@ -152,8 +161,12 @@ def get_permissions_command(server_url: str) -> None:
...
+    print("Permissions (Direct Scopes):")
+    for scope in scopes:
+        print(f"\t{scope}")
 
-    print("Permissions:")
-    for permission in permissions:
-        print(f"\t{permission}")
+    print("Roles:")
+    for role in roles:
+        print(f"\t{role}")
```
pattisdr, line 172: Adjusting CLI command for getting permissions to show both scopes and roles.

src/fides/lib/models/client.py
```diff
@@ -36,7 +38,8 @@ def __tablename__(self) -> str:
 
     hashed_secret = Column(String, nullable=False)
     salt = Column(String, nullable=False)
-    scopes = Column(ARRAY(String), nullable=False, default="{}")
+    scopes = Column(ARRAY(String), nullable=True, server_default="{}", default=dict)
+    roles = Column(ARRAY(String), nullable=True, server_default="{}", default=dict)
```
pattisdr, line 42: Adding new roles to ClientDetail (similar to addition on UserPermissions) and changing the default value because the existing does not work as intended.

One question to clear up. I am allowing multiple "roles" to be assigned to the user for future flexibility, but does product prefer this to be limited to one role?  Originally, the requirements had an 1) Admin 2) Viewer and 3) Privacy Request Manager, but now there's 1) Admin 2) Viewer 3) Privacy Request Manager and 4) Privacy Request Manager + Viewer.   This leads me to believe the intent is now to be assigned just the one "Privacy Request Manager + Viewer" instead of two "Privacy Request Manager" AND "Viewer".

However, if there's ever the possibility that we want multiple roles in the future, I think this should remain a list.

src/fides/core/config/security_settings.py
pattisdr: Addressed in https://github.com/ethyca/fides/pull/2671/commits/ae8355b51fbcc48d014029ad80d8b40ae96e5fa6

pattisdr:
hi @TheAndrewJackson @ThomasLaPiana @sanders41 @seanpreston - I'm looking for a reviewer that's worked on permissions-related items if one of you has time to look.  

In short, this work adds the ability to assign a larger role to a user (which is associated with a list of scopes) in addition to still preserving the ability to assign scopes directly.  Permission checking still looks to see if the user has the necessary scope - whether it was assigned directly or the user has a role associated with that scope.

ThomasLaPiana:
Looking now 👁️

ThomasLaPiana:
@pattisdr is it worth updating the CLI `fides user create` command to assign roles instead of scopes? It seems like it'll work fine either way, but figured we might as well update it to reflect best practices if this is how we're going to try to do things from now on

src/fides/api/ops/schemas/user_permission.py
```diff
@@ -1,23 +1,23 @@
...
-class UserPermissionsCreate(UserPermissionsCreateLib):
-    """Data required to create a FidesUserPermissions record"""
+class UserPermissionsCreate(BaseSchema):
+    """Data required to create a FidesUserPermissions record
+
+    Users will generally be assigned role(s) directly which are associated with many scopes,
+    but we also will continue to support the ability to be assigned specific individual scopes.
+    """
+
+    scopes: List[str] = []
```
ThomasLaPiana, line 19: this is a nit, but since we know beforehand all of the defined scopes/roles this could be an Enum

src/fides/core/config/security_settings.py
```diff
@@ -10,6 +10,7 @@
 from fides.api.ops.api.v1.scope_registry import SCOPE_REGISTRY
 from fides.lib.cryptography.cryptographic_util import generate_salt, hash_with_salt
 
+from ...lib.oauth.roles import ADMIN
```
ThomasLaPiana, line 13: nit: the nested relative import

src/fides/api/ctl/migrations/versions/985d3756f908_add_role_based_foundation.py
```diff
@@ -0,0 +1,50 @@
...
+    )
+    op.add_column(
+        "fidesuserpermissions", sa.Column("roles", sa.ARRAY(sa.String()), nullable=True)
+    )
+    op.alter_column(
+        "fidesuserpermissions",
+        "scopes",
+        existing_type=postgresql.ARRAY(sa.VARCHAR()),
+        nullable=True,
+    )
```
sanders41, line 36: With all of these fields being nullable a user could have no permissions at all. Do we handle this somehow?

src/fides/lib/oauth/roles.py
```diff
@@ -0,0 +1,100 @@
...
+    RULE_READ,
+    SCOPE_READ,
+    STORAGE_READ,
+    MESSAGING_READ,
+    WEBHOOK_READ,
+    SAAS_CONFIG_READ,
+    USER_READ,
+]
+
+roles_to_scopes_mapping: Dict[str, List] = {
```
sanders41, line 82: Should this be upper case to denote a constant?

```suggestion
ROLES_TO_SCOPES_MAPPING: Dict[str, List] = {
```

src/fides/api/ops/api/v1/endpoints/user_permission_endpoints.py
sanders41: What if a user's role and scopes conflict with each other? what takes precedence and how will end users know what this precedence is?

Either way I can see this being confusing to users/administrators.  Lets say you and I both are in the `VIEWER` group and you have an added scope to also approve. You ask me to approve for you, but I get an error. It totally makes sense why I get an error, but to the end user they are going to think they are in the same group so they can do the same thing. We will need to have a way to make it clear to the end users what happened in situations like this.

src/fides/core/user.py
```diff
@@ -64,9 +64,9 @@ def create_user(
 
 def get_user_permissions(
     user_id: str, auth_header: Dict[str, str], server_url: str
-) -> List[str]:
+) -> Tuple[List[str], List[str]]:
```
sanders41, line 67: Since scopes in the database are nullable now I think the database could also return `None` for these now?

src/fides/api/ops/api/v1/endpoints/user_permission_endpoints.py
pattisdr: Thanks for mentioning this, I've been mulling over this too!  I've also been concerned that since the "scopes" UI is going away, there's going to be no easy visibility into who has these extra scopes.   From our refinement discussion on this yesterday, however, the majority opinion seemed to be that this was an edge case and we could address if it became an issue.


- We still need to support scopes generally for people just using the API, for example someone developing an app against a fides instance or someone that wants to create a very limited token to do one specific job.
- Regardless, we need to support both roles and scopes on users as a transitional measure as there is no "roles" UI yet. 
- I suggested a future ticket to put in the backlog where we **deprecated assigning scopes directly to users, but still allowed scopes to be assigned directly to clients.**  (So the user permissions table would lose the scopes column, and clients created/updated from user permissions would only have roles persisted to the client and then the token.  Folks using the API-only workflow could still create clients directly and assign scopes directly to those clients).  Would something like that help here @sanders41?

pattisdr: > What if a user's role and scopes conflict with each other? what takes precedence and how will end users know what this precedence is?

Regarding "precedence", there aren't really conflicts here.  Permission checking just looks to see "do you have this scope through some avenue?" You might have the scope because you have a specific role, or you might have the scope because you have it directly.  You might be granted the scope from both places and that's fine.

What you can see is _really the sum of all of your scopes from all your sources._  

In practice, with the roles-based UI going in, most users will have a single role assigned to them, and inherit the associated list of scopes.

sanders41: Yes, I think this would solve the issue.

src/fides/api/ops/schemas/user_permission.py
pattisdr: Yes good point, @allisonking was [also looking into making scopes an enum](https://github.com/ethyca/fides/pull/2682/files), so I'd need to coordinate there, and could also make the roles enums.  I think I was just patterning this way because scopes were already a list of strings, but an enum makes more sense.

src/fides/core/config/security_settings.py
pattisdr: ah thanks sometimes I have my IDE just import it for me and miss these

src/fides/api/ctl/migrations/versions/985d3756f908_add_role_based_foundation.py
pattisdr: Good callout. If a user has no permissions then they just can't do anything, which might make for a weird experience, but it doesn't break anything. I could add a code-side check when setting user permissions that they need either/or.

src/fides/lib/oauth/roles.py
pattisdr: Yes thank you!

src/fides/api/ctl/migrations/versions/985d3756f908_add_role_based_foundation.py
sanders41: I agree, other than a weird experience I don't see anything breaking with this. I'm open to whatever everyone thinks is best here. If we aren't worried about the chance someone has no roles I don't have an issue, it seems like a low possibility they wouldn't assign any roles.

src/fides/api/ops/schemas/user_permission.py
pattisdr: OK I've added a role enum here https://github.com/ethyca/fides/pull/2671/files#diff-2a6e5b613afe739bdfbefc63a01ce84e08aa97a31a25f5e3eaafe41affb8ff10R40 and will leave the scopes enum for Allison -

src/fides/core/config/security_settings.py
pattisdr: Fixed here https://github.com/ethyca/fides/pull/2671/files#diff-fb9c1921ed9a56a2fbcaa0df04c789950c1b5559bca82b050abe368548ab5111R12

src/fides/core/user.py
pattisdr: Thanks for pointing out this case.  We do set a default code-side, but you're right this could have returned null if it was set some other way.

Thinking more I'd really like the defaults for these to be empty lists, not null, so I've changed these back to nullable being false and added a server default https://github.com/ethyca/fides/pull/2671/files#diff-a1d112d2dcc021ea7afc1ab1d87855a1d12c47c33c42f28299cef315bb71c8b9R29-R43

src/fides/lib/oauth/roles.py
pattisdr: Turned into a constant here https://github.com/ethyca/fides/pull/2671/files#diff-2a6e5b613afe739bdfbefc63a01ce84e08aa97a31a25f5e3eaafe41affb8ff10R84

src/fides/api/ctl/migrations/versions/985d3756f908_add_role_based_foundation.py
pattisdr: OK current state is that defaults are empty lists and we still set "privacy request viewer" permissions as a default at the code level.  After the roles UI is in, we can make the default a "role" viewer code side.

In the off-chance someone does have an empty list of scopes and roles, I'm throwing a 403 on login.

pattisdr: Sorry I hadn't refreshed my page before I saw your comment Paul!

sanders41 (approved):

src/fides/lib/oauth/roles.py
ThomasLaPiana: I think this is a great starting point

ThomasLaPiana (approved):
This looks fantastic! Looks good for merge if the checklist is satisfied

src/fides/api/ctl/migrations/versions/eb1e6ec39b83_add_role_based_permissions.py
```diff
@@ -0,0 +1,63 @@
...
+revision = "eb1e6ec39b83"
+down_revision = "d65bbc647083"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column(
+        "client",
+        sa.Column("roles", sa.ARRAY(sa.String()), server_default="{}", nullable=False),
```
allisonking, line 23: thinking about the enum discussion, is there a reason to not use an enum in the db? I guess it'd require a migration when adding/deleting a role which is a pain, but arguably doing adding/deleting a role may already require a data migration of some sort for existing users. 

I ask because when I was [playing around with enums for scopes](https://github.com/ethyca/fides/discussions/2642#discussioncomment-5092052), I couldn't make the `get_user_permissions` func/endpoint typed to return a list of the Scope enums since it wasn't an enum in the db. I was able to work around it by instead having `read_scopes` return the list of Scope enums, since for my use case the frontend just needed an up to date list of the scopes. but this actually results in some type fudging that needs to be done on the frontend to convince typescript that `get_user_permissions`'s scopes are the same type of scopes that `read_scopes` returns. the type fudging is really easy to do, but is technically unsafe, and I wonder if that might mean it's a lil unsafe in other places too. 

migrating to an enum for scopes in the db seems like a big pain as well 😅 but mostly just want to get the discussion going, even if we choose not to go that route!

pattisdr: @allisonking can you show me a snippet of what you'd need the API response to look like?  I feel like there's a way to change the format of `get_user_permissions` to be what you need without making this be a database migration - I can look into this further I'm hesitant to add a db migration for this one since every time a new endpoint goes in we likely need new scopes.

allisonking: yup yup!

pattisdr:
Thanks for your reviews @sanders41 and @ThomasLaPiana!

PR #2676, opened 2023-02-23
re-enable support for lower-cased messaging service_type values

adamsachs:
As noticed by @ssangervasi in https://github.com/ethyca/fides/pull/2674, on `main` we stopped accepting lowercased `service_type` values in our basic messaging config endpoints. this was an unintended side effect of a change made in https://github.com/ethyca/fides/pull/2572.

We can fix things up to still support the lowercased `service_type` values and keep this fully backward compatible.

Also added in a fix to make sure we support lowercased `service_type` values in URL paths

### Code Changes

* add back the upper-cased (transformed) `service_type` value assignment to the `values` dict in the pydantic validator
* ensure the validator is only run on request models, since we don't need to validate this value in responses. running this validator on responses would throw an error because the response `values` object is not a true dict, and so it cannot have values assigned.
* implement a `_missing_` classmethod on our enum to allow lowercased values to be recognized. this is for use in the path URL

### Steps to Confirm

As noted by @ssangervasi  in his PR -
* [ ] Clear out your fides_env(dev) containers+volumes, then rebuild them from scratch. (ensure you've got a `.env` file with mailgun details to actually get the relevant code to run, though!)

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### DIFF
--- a/src/fides/api/ops/api/v1/endpoints/messaging_endpoints.py
+++ b/src/fides/api/ops/api/v1/endpoints/messaging_endpoints.py
@@ -46,8 +46,8 @@ from fides.api.ops.models.messaging import (
 )
 from fides.api.ops.schemas.messaging.messaging import (
     MessagingActionType,
-    MessagingConfigBase,
     MessagingConfigRequest,
+    MessagingConfigRequestBase,
     MessagingConfigResponse,
     MessagingServiceType,
     TestMessagingStatusMessage,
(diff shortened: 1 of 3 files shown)


eastandwestwind (approved):
LGTM. Thanks for adding tests for this, too.

eastandwestwind (approved):

PR #2677, opened 2023-02-23
2455/Configuration updates post-API

### LINKED ISSUE #2455: Storage and messaging configuration test against BE

_The work for storage and messaging configuration UI is hidden behind a feature flag until the BE components have been completed. This ticket is to test storage and messaging configuration against the live BE from the previous sprint.

_Add any related tests for the configuration to Cypress against the BE requests
_Additional user/functional testing and bug logging as discovered

### Additional context

I ( @adamsachs ) have started doing some testing on `main` now that the BE work is merged into `main`. i've taken note the following points that need to be resolved:

_Front end_

- local storage config requires at least an empty `details` nested object property -- @adamsachs has updated our confluence spec doc accordingly:
```
{
	"type": "local",
	"details": {
	},
	"format": "json"
  }
```
- `PATCH` calls to `/config` endpoint should not have an outer `fides` object in the payload, i.e. they should look just like:
```
{ 
    "storage" : {
      "active_default_storage_type" : "s3"
    }
}
```
- it seems like the details fields within the payload for storage `PUT` updates is not being populated with the values filled in by the form - here's an example request body i grabbed from my browser dev tools, even when i had auth_method and bucket populated in the UI:
```
{"type":"s3","details":{"auth_method":"","bucket":""},"format":"json"}
```
- the `notifications.notification_service_type` property in the payload for the `PATCH /config` calls are being populated with incorrect values: `mailgun-email` should be `mailgun`, `twilio-email` should be `twilio_email` and `twilio-text` should be `twilio_text`
- `PUT` calls for updating messaging config values should not include the messaging type in the URL path, i.e. it should just be `PUT /messaging/default` and the `service_type` is specified as a field in the request payload.
    - the confluence spec doc had `type` as the field in the request payload, @adamsachs has updated that to the correct `service_type` field now in the confluence doc.

_Back end_

- when a messaging `service_type` is referenced in a URL path, do not require it to be upper-cased. currently the backend requires the `GET` call to look like `http://localhost:8080/api/v1/messaging/default/MAILGUN`, which is ugly.

LKCSmith:
Closes #2455

Follow-up issue:
https://github.com/ethyca/fides/issues/2687

### Code Changes

### Steps to Confirm
* Go to `http://localhost:3000/privacy-requests/configure`
* Click on "Configure messaging provider'
* Go through adding both Mailgun, Twilio email, and Twilio sms as a messaging provider - you should be able to configure any as a messaging provider
* Go back to `http://localhost:3000/privacy-requests/configure`
* Click on "Configure storage"
* Clicking the 'Local' radio button sets local storage as the default storage type -- no other information needed
* Clicking the 'S3' radio button presents a form to enter Format, Auth method, and Bucket information for S3.
* If you set the option `secret_keys` under Auth method, then click save, a secondary form appears to enter your S3 secret information to save.

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes


adamsachs:
nice @LKCSmith! generally looking pretty well wired up to the BE now.

noting down the few lingering issues i've found when doing some initial testing:
- when adding secrets to storage (i.e. only in the case of s3 storage with secret keys), the `PUT` secrets call seems to not get a correctly populated URL - i see the request being sent to the following URL in my browser console: `http://0.0.0.0:8080/api/v1/storage/default/undefined/secret` --> should be: `http://0.0.0.0:8080/api/v1/storage/default/s3/secret`
    - same occurs for messaging secrets
- [minor] for twilio sms config, i believe `Messaging Service SID` and `Phone Number` fields should be optional, not required

also, it would be good to talk through API flow, as you mention. generally i think the way you have it is pretty good now- - there are just a couple of nuanced edge cases/workflows that would be good to talk through.

clients/admin-ui/src/features/common/Layout.tsx
```diff
@@ -24,14 +24,25 @@ const Layout = ({
...
-    (!activeStorage || !activeMessagingProvider) &&
-    features.flags.privacyRequestsConfiguration &&
-    (router.pathname === "/privacy-requests" ||
-      router.pathname === "/datastore-connection");
+  const showConfigurationNotificationBanner = () => {
+    if (
+      router.pathname === "/privacy-requests" ||
+      router.pathname === "/datastore-connection"
+    ) {
+      const { data: activeMessagingProvider } =
```
LKCSmith, line 32: This and the call below were being called on every page the `Layout` was being used. This allows it to be called only on the pages that need to be aware of any existing data.

LKCSmith: That being said, a little lost on how to resolve the linting errors below 🤷‍♀️

clients/admin-ui/src/features/privacy-requests/configuration/S3StorageConfiguration.tsx
```diff
@@ -24,8 +22,10 @@ interface SecretsStorageData {
 }
 
 const S3StorageConfiguration = ({
-  storageDetails: { auth_method, bucket, format },
-}: any) => {
+  auth_method,
```
LKCSmith, line 25: Not sure what this looks like deconstructed to avoid the lint error in `StorageConfiguration.tsx`

adamsachs:
@LKCSmith one thing i've noticed testing on latest version of this:

the request payload for the `PATCH /api/v1/config` call has `subject_identity_verification_required` as part of the `notifications` object; instead, it needs to be in a separate `execution` object, e.g.: 
```
```
{ 
  "notifications": {
    "notification_service_type" : "mailgun",
    "send_request_completion_notification": true,
    "send_request_receipt_notification": true,
    "send_request_review_notification": true
  }
  "execution": {
    "subject_identity_verification_required": true  
  }
}
```

additionally, i realized i'd missed out a setting when spec-ing out the API here. we should _also_ be including a `require_manual_request_approval` field under the `execution` object - but in order for this to work, i'm going to need to make a PR for updates to the backend. i'll go ahead and get that queued up, it should be a very quick change. once that's in place, we'll want the request payload here to look like:
```
{ 
  "notifications": {
    "notification_service_type" : "mailgun",
    "send_request_completion_notification": true,
    "send_request_receipt_notification": true,
    "send_request_review_notification": true
  }
  "execution": {
    "subject_identity_verification_required": true,
    "require_manual_request_approval": true
  }
}
```

(FYI @saiatethyca @NevilleS @Kelsey-Ethyca  - i'm assuming here that we want to enable manual request approval "by default" when a messaging provider is configured. initially i'd forgotten to put that in the implementation plan and only had subject identity verification automatically switched on. please let me know if we do _not_ want to enable manual request approval by default when a messaging provider is configured.)

File: clients/admin-ui/src/features/common/Layout.tsx [hunk 3]
```diff
@@ -1,3 +1,6 @@
+/* istanbul ignore file */
```
LKCSmith, line 1: I will remove these lines as soon as I can figure out the linting issues as documented below.

adamsachs:
just pulled locally and did some end-to-end smoke testing again, looking really good @LKCSmith! i was able to get things running end to end with pretty much no hurdles. there are still a few small things i noticed that i still think should be cleaned up in our initial iteration here:
- "active default" URLs seem to be a bit off, which means that we're not really retrieving the "current details" of messaging/storage config well if they've already been configured and a user returns to the screen. from what i can tell, we just need some small adjustment to the URLs being called:
    - for messaging: currently `GET http://0.0.0.0:8080/api/v1/messaging/active/default` --> should be `GET http://0.0.0.0:8080/api/v1/messaging/default/active`
    - for storage: currently `GET http://0.0.0.0:8080/api/v1/storage/default/` --> should be `http://0.0.0.0:8080/api/v1/storage/default/active`
- for `Twilio SMS`, both `Messaging Service SID` and `Phone Number` fields are required. only one of the two needs to be provided on any update. simplest thing here is to make them both "optional" and the user will have to know that _one_ is required. best thing would be to have some "save" validation that complains if neither of the two are provided, but i understand if that's out of scope of our initial iteration here. but we should not require both fields - that will necessarily lead to problems.
- `Bucket` is an optional field on s3 storage config. it should be required.
- some copy edits (i know these were on the designs, and i'm being opinionated here, would be happy to see what others think):
    - on config landing screen, text for `Configure messaging provider` card includes `You'll need to set up config variables to send out messages from Fides` -- i don't think this is true anymore, and it should be removed
    - within messaging config screen, explanatory paragraph has `Please follow the documentation to setup a messaging service that Fides supports. ` with `documentation` highlighted a link. that link isn't clickable for me, though. maybe this is pending doc updates? just want to make sure we've got a plan for this
    - nit: the radio button for `Twilio sms` should maybe read `Twilio SMS`? 
 
let me know if you've got any questions on these, happy to sync up offline to discuss further!

clients/admin-ui/src/features/common/Layout.tsx
chriscalhoun1974: I would suggest using a useEffect hook for showing the configuration notification banner along with the privacyRequestConfiguration feature flag for checking whether or not to invoke your hooks. Unfortunately, you cannot use the `useGetActiveMessagingProviderQuery` and `useGetActiveStorageQuery` hooks directly in a given function.

clients/admin-ui/src/features/privacy-requests/configuration/MailgunEmailConfiguration.tsx
```diff
@@ -19,7 +19,7 @@ const MailgunEmailConfiguration = () => {
     useState<ConnectionStep>("");
   const { handleError } = useAPIHelper();
   const { data: messagingDetails } = useGetMessagingConfigurationDetailsQuery({
-    type: "mailgun",
+    type: "MAILGUN",
```
chriscalhoun1974, line 22: Should this be a string literal constant declared in the file or an enum configuration type located in a constants.ts file?

clients/admin-ui/src/features/privacy-requests/configuration/MessagingConfiguration.tsx
```diff
@@ -118,36 +126,34 @@ const MessagingConfiguration = () => {
         >
           <Stack direction="row">
             <Radio
-              key="mailgun-email"
-              value="mailgun-email"
-              data-testid="option-mailgun-email"
+              key="MAILGUN"
```
chriscalhoun1974, line 129: I believe a string based enum suggested above would be helpful here.

clients/admin-ui/src/features/privacy-requests/configuration/StorageConfiguration.tsx
```diff
@@ -27,37 +27,48 @@ const StorageConfiguration = () => {
...
-  const [saveActiveStorage] = useCreateConfigurationSettingsMutation();
+
   const { data: storageDetails } = useGetStorageDetailsQuery({
     type: storageValue,
   });
+  const [saveStorageType, { isLoading }] = useCreateStorageMutation();
+  const [saveActiveStorage] = useCreateConfigurationSettingsMutation();
+
+  useEffect(() => {
+    if (storageDetails) {
```
chriscalhoun1974, line 38: Is there an API model response for useGetStorageDetailsQuery? If so, I would have this strongly typed.

clients/admin-ui/src/features/privacy-requests/configuration/TwilioEmailConfiguration.tsx
```diff
@@ -16,7 +16,7 @@ const TwilioEmailConfiguration = () => {
   const { successAlert } = useAlert();
   const { handleError } = useAPIHelper();
   const { data: messagingDetails } = useGetMessagingConfigurationDetailsQuery({
-    type: "twilio_email",
+    type: "TWILIO_EMAIL",
```
chriscalhoun1974, line 19: See my previous comment about a string based enum.

clients/admin-ui/src/features/privacy-requests/configuration/TwilioEmailConfiguration.tsx
```diff
@@ -45,7 +45,10 @@ const TwilioEmailConfiguration = () => {
     api_key: string;
   }) => {
     const result = await createMessagingConfigurationSecrets({
-      twilio_api_key: value.api_key,
+      details: {
```
chriscalhoun1974, line 48: I would have suggested a BaseConfiguration component given these components are identical except for the data params.

clients/admin-ui/src/features/privacy-requests/privacy-requests.slice.ts
```diff
@@ -361,7 +361,7 @@ export const privacyRequestApi = createApi({
     }),
     getActiveStorage: build.query<any, void>({
       query: () => ({
-        url: `storage/default/active`,
+        url: `storage/active/default`,
```
chriscalhoun1974, line 364: Are the API model request/response models ready to be generated? Do these API endpoints need invalidateTags such as your PUT API endpoints?

clients/admin-ui/cypress/e2e/privacy-requests.cy.ts
```diff
@@ -12,7 +12,7 @@ describe("Privacy Requests", () => {
     stubPrivacyRequestsConfigurationCrud();
   });
 
-  describe("The requests table", () => {
+  describe.skip("The requests table", () => {
```
LKCSmith, line 15: TODO: get rid of skips

clients/admin-ui/src/features/privacy-requests/types.ts
```diff
@@ -112,27 +112,27 @@ export type RetryRequests = {
...
 
 export interface MessagingConfigResponse {
-  fides: {
-    storage: {
-      active_default_storage_type: string;
-    };
+  storage: {
+    active_default_storage_type: string;
   };
 }
```
chriscalhoun1974, line 119: These interfaces can be deleted after you have generated the API models request/response.

adamsachs:
update on lingering issues after pulling latest (which includes my update to fix `active/default` --> `default/active`:
- setting s3 details is erroring out:  the call to `PUT storage/default` is not populating its details field correctly - here's an example request payload i've copied from my browser dev tools:
```{"type":"s3","details":{"auth_method":"","bucket":""},"format":"json"}```
- copied from above: `Bucket` is an optional field on s3 storage config. it should be required.
- copied from above: for Twilio SMS, both Messaging Service SID and Phone Number fields are required. only one of the two needs to be provided on any update. simplest thing here is to make them both "optional" and the user will have to know that one is required. best thing would be to have some "save" validation that complains if neither of the two are provided, but i understand if that's out of scope of our initial iteration here. but we should not require both fields - that will necessarily lead to problems.

only the first is a true blocker, but i'd say we should also aim to get the last two resolved before we release this.

my copy edits from above still hold, but i wouldn't consider those blockers.

clients/admin-ui/src/features/privacy-requests/types.ts
LKCSmith: You might need to show me what you mean by this.

Kelsey-Ethyca (approved):

PR #2682, opened 2023-02-23
Access controls UI spike

allisonking:
Closes https://github.com/ethyca/fidesplus/issues/599

### Code Changes

* [x] Small Python change to get the `read_scopes` endpoint to be typed to return an `enum` instead so we have the list of scopes in typescript
* [x] Autogenerate TS types
* [x] Update the `configureNavGroups` logic to also respect scopes (as determined by [this doc](https://ethyca.atlassian.net/wiki/spaces/EN/pages/2621734978/Access+Controls+FE+restrictions))
* [x] Refactor the code that renders the "tiles" on the home page to also respect scopes (in a similar way to what we do in nav-config)
* [x] Add a new cypress custom command `cy.assumeRole` so it's easy to test out the app as different roles
* [x] Add a `<Restrict />` component which handles hiding things certain users shouldn't see (for example: the password reset button)
* [x] Update `<ProtectedRoute />` such that it will not let you see any routes that you don't have the scope for

### Steps to Confirm

* [ ] If you log in as a root user, your experience of the app should not be any different than it was before.
* [ ] You'll have to create another user with limited scopes in order to see the app as different roles. You can find the relevant scopes needed for each path in `nav-config.ts`. For example, to view the privacy requests page, you need to be able to `privacy-request:read`, and to view the add systems page, you need to be able to `cli_objects:create` or `cli_objects:update` (note that `cli_objects:update/create` do not exist as scopes that can be assigned in the UI though)
  * [ ] If you try to navigate to any of these pages, you should be redirected to the home page
  * [ ] The home tiles should also look different depending on which user you are

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### Description Of Changes

https://www.loom.com/share/90ec381501954092b0f88f2155a6d292


clients/admin-ui/cypress/e2e/home.cy.ts
```diff
@@ -0,0 +1,68 @@
...
+    cy.getByTestId(`tile-${item}`).should("not.exist");
+  });
+};
+
+describe("Home page", () => {
+  beforeEach(() => {
+    cy.login();
+  });
+
+  describe("permissions", () => {
```
allisonking, line 28: ideally as we build out the UI respecting scopes, every cypress test will have a `permissions` block that goes through and does a `cy.assumeRole` on every relevant role

clients/admin-ui/src/features/auth/auth.slice.ts
```diff
@@ -81,8 +84,16 @@ export const authApi = createApi({
       }),
       invalidatesTags: () => ["Auth"],
     }),
+    getRolesToScopesMapping: build.query<RoleToScopes, void>({
```
allisonking, line 87: this isn't used yet, but was helpful in generating the cypress fixture. also it might get used in https://github.com/ethyca/fidesplus/issues/641

clients/admin-ui/src/features/common/Restrict.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { ScopeRegistry } from "~/types/api";
+
+export const useHasPermission = (scopes: ScopeRegistry[]) => {
+  const userScopes = useAppSelector(selectThisUsersScopes);
+  return (
+    userScopes.filter((userScope) => scopes.includes(userScope)).length > 0
+  );
+};
+
+const Restrict = ({
```
allisonking, line 12: this component is pretty basic right now in just deciding whether or not to render something based on if the user has a scope. we'll need to extend it once the data steward flow is in place

clients/admin-ui/src/features/user-management/PasswordManagement.tsx
```diff
@@ -39,14 +34,12 @@ const PasswordManagement = ({ profileId }: Props) => {
...
 
-  if (!isOwnProfile && !canForceResetPassword) {
-    return null;
-  }
-
   return (
     <HStack>
       {isOwnProfile ? <UpdatePasswordModal id={profileId} /> : null}
-      {canForceResetPassword ? <NewPasswordModal id={profileId} /> : null}
+      <Restrict scopes={[ScopeRegistry.USER_PASSWORD_RESET]}>
```
allisonking, line 40: example of `<Restrict />` in action! cleans up a good amount of boilerplate code

clients/admin-ui/src/home/tile-config.ts
```diff
@@ -0,0 +1,44 @@
+import { ScopeRegistry } from "~/types/api";
+
+import { ModuleCardConfig } from "./types";
+
+export const configureTiles = ({
```
allisonking, line 5: very much modeled after `configureNav`

clients/admin-ui/src/home/types.ts
```diff
@@ -1,13 +1,18 @@
...
-  descriptionColor: string;
-  hoverBorderColor: string;
   href: string;
   key: number;
   name: string;
-  nameColor: string;
   sortOrder: number;
   title: string;
-  titleColor: string;
+  color: string;
```
allisonking, line 10: the `color`s were all consistently derived from each other, so I changed it to just take one `color`

clients/admin-ui/cypress/e2e/home.cy.ts
TheAndrewJackson: That makes sense. Essentially verify that each role only has access to only what's expected

clients/admin-ui/cypress/support/commands.ts
```diff
@@ -26,11 +27,20 @@ Cypress.Commands.add("login", () => {
         })
       );
     });
+    cy.intercept("/api/v1/user/*/permission", {
+      fixture: "user-management/permissions.json",
+    }).as("getUserPermission");
+  });
+});
+
+Cypress.Commands.add("assumeRole", (role) => {
```
TheAndrewJackson, line 36: This is really cool! I need to read up on cypress

clients/admin-ui/src/features/auth/auth.slice.ts
```diff
@@ -53,6 +54,8 @@ export const selectToken = (state: RootState) => selectAuth(state).token;
 
 export const { login, logout } = authSlice.actions;
 
+type RoleToScopes = Record<RoleRegistry, ScopeRegistry[]>;
```
TheAndrewJackson, line 57: Great type usage!

clients/admin-ui/src/features/common/nav/v2/nav-config.ts
```diff
@@ -75,14 +124,58 @@ export type NavGroup = {
...
+/**
+ * If a group contains only routes that the user cannot access, return false.
+ * An empty list of scopes is a special case where any scope works.
+ */
+const navGroupInScope = (
+  group: NavConfigGroup,
+  userScopes: ScopeRegistry[]
+) => {
+  if (group.routes.filter((route) => route.scopes.length === 0).length === 0) {
+    const allScopesAcrossRoutes = group.routes.reduce((acc, route) => {
```
TheAndrewJackson, line 136: Great use of iterators 👀  I don't use `reduce` that often

TheAndrewJackson (approved):
I like the updates here. Pretty easy to follow along. I created a test user and tried removing and adding some permissions to see the UI update. Everything seemed to be working as expected 🚀

clients/admin-ui/src/features/common/nav/v2/nav-config.ts
allisonking: honestly I avoid it when possible because I always have to look up how to use it... but for things like this it's really useful

PR #2694, opened 2023-02-24
Auto-Generate config documentation

### LINKED ISSUE #2627: Add autogenerated configuration docs

With the merger of #2613 we'll have all relevant metadata we need within source code to produce exhaustive documentation for our configuration reference.

The inclusion of autogenerated configuration documentation included as part of the docs site, which `docs.ethyca.com` can point to.

### Additional context

Configuration docs get out of sync more than anything I would wager, so this is important for usability. Although it might feel weird to pull docs out of the new docs site and into the "legacy" site, it makes sense in this case given that we have no low-friction way to autogenerate docs for the new docs site. Note that we still use the old docs site for hosting the autogenerated API and CLI documentation.

ThomasLaPiana:
Closes #2627 

### Code Changes

* [x] autogenerate the config file as valid `toml` that includes helpful metadata
* [x] add an assertion that there are no "TODO" descriptions
* [x] add a page for it in the docs
* [x] added a new `generate_docs` Nox session, included as part of the backend checks
* [x] make this the default file that gets output as part of `fides init`
* [x] copy relevant docs from the new site back here in preparation for removing config docs from the new site (will end up cross-linking)
* [x] overhaul how analytics consent requests work
  * [x] refactor the `check_and_update_analytics_config` function so that it doesn't check for a `none` value (since the default is now `true`, this is pointless)
  * [x] do not call the function from the main code-path, only call it from `fides init` or `fides deploy up`
  * [x] combine the logic for consent/writing default config/updating config values into a reusable function for `init` and `deploy up`
* [x] add a link to the config docs page at the top of the config file

### Steps to Confirm

* [ ] spin up the docs with `nox -s docs_serve` and visit the `configuration` page. Confirm the toml snippet is rendering properly
* [ ] Verbiage/accuracy check on the docstring updates
* [ ] go into an empty dir and run `fides init`, verify that the new config file is produced
* [ ] verify that the new config file can be used to configure a fides instance (make updates to it and run `fides view config` to verify the updates)
* [ ] verify the new `generat_docs` nox session and CI check

### Pre-Merge Checklist

* Documentation:
  * [x] documentation complete, [PR opened in fidesdocs](https://github.com/ethyca/fidesdocs/pulls) - https://github.com/ethyca/fidesdocs/pull/66
* [x] Update `CHANGELOG.md`

### Description Of Changes

This PR fundamentally changes how our "default" config file is produced and therefore also fundamentally changes `fides init`.

The new config file is built from the source code directly and includes with it all of the useful metadata we have as part of Pydantic and Pydantic `Field` objects. An important piece is that the generated config file is still valid `toml` and will be what the user interacts with directly when configuring their instance! *CI Checks will fail if a config value doesn't have a description, the description is blank, or it is "TODO"*

Additionally, the new configuration file reference is built and hosted on the docs site, as a reference to users on older versions or who have been upgrading and used a past version of `fides init`. The new configuration docs will be cross-linked from the new docs site.

Note that as part of this change, how/when we collect user consent has changed. It is less intrusive now. It defaults to opt-out (no dark patterns here) but gives the option to opt-in upon `fides init` and will update the config file directly.

__Note__: **Given the lack of a `None` type in `toml`, `Optional` or empty values are no longer valid in the configuration file! If needed, you can set a default value and do a truthiness check. i.e. `"" == False`, as does `[] == False`, so we need to leverage that type of check instead.**

#### Preview of new autogenerated config

```toml
# Fides Configuration File
# Additional Documentation at : https://ethyca.github.io/fides/stable/config/

#--------------#
#-- ADMIN_UI --#
#--------------------------------------------------------------------#
[admin_ui] # Configuration settings for the Admin UI.

# Toggle whether the Admin UI is served.
enabled = true # boolean

#---------#
#-- CLI --#
#--------------------------------------------------------------------#
[cli] # Configuration settings for the command-line appliation.

# A fully anonymized unique identifier that is automatically generated
# by the application. Used for anonymous analytics when opted-in.
analytics_id = "951ddced878629858409539c4296ad8c" # string

# When set to True, disables functionality that requires making calls
# to a Fides webserver.
local_mode = false # boolean

# The protocol used by the Fides webserver.
server_protocol = "http" # string

# The hostname of the Fides webserver.
server_host = "localhost" # string

# The port of the Fides webserver
server_port = "8080" # string

#--------------#
#-- DATABASE --#
#--------------------------------------------------------------------#
[database] # Configuration settings for the application database.

# Number of concurrent database connections Fides will use for API
# requests. Note that the pool begins with no connections, but as they
# are requested the connections are maintained and reused up to this
# limit.
api_engine_pool_size = 50 # integer
...
```


src/fides/cli/commands/util.py
```diff
@@ -27,6 +30,28 @@
 from fides.core.utils import echo_green
 
 
+def create_and_update_config_file(
```
ThomasLaPiana, line 33: This function is only used for these CLI commands, but maybe it should live elsewhere?

src/fides/core/config/helpers.py
```diff
@@ -103,41 +101,6 @@ def update_config_file(  # type: ignore
             echo(f"\tSet {key}.{subkey} = {val}")
 
 
-def create_config_file(
```
ThomasLaPiana, line 106: this was moved to `docs.py` since the docs version is the only version that should be exported

src/fides/core/config/redis_settings.py
```diff
@@ -15,11 +15,14 @@ class RedisSettings(FidesSettings):
         default="utf8",
         description="Character set to use for Redis, defaults to 'utf8'. Not recommended to change.",
     )
-    db_index: Optional[int] = Field(
-        default=None,
+    db_index: int = Field(
+        default=0,
```
ThomasLaPiana, line 19: I think this should be fine as a default? `None` values cause problems in the config file (because the concept of `null` doesn't exist in TOML)

src/fides/core/config/utils.py
```diff
@@ -4,6 +4,29 @@
 DEFAULT_CONFIG_PATH_ENV_VAR = "FIDES__CONFIG_PATH"
 
 
+def replace_config_value(
```
ThomasLaPiana, line 7: This is a semi-hacky workaround. Because the new config file that we export is more than just a plain `dict` -> `toml`  string, it makes more sense to do string replacement for updating config values. This shouldn't need to be used often

src/fides/core/config/docs.py
```diff
@@ -0,0 +1,181 @@
...
+        name: settings[name] for name in sorted(set(settings.keys()))
+    }
+    nested_settings_docs: List[str] = [
+        convert_settings_to_toml_docs(settings_name, settings_schema)
+        for settings_name, settings_schema in ordered_settings.items()
+    ]
+    docs: str = "\n".join(nested_settings_docs)
+
+    # Verify it is valid TOML before writing it out
+    toml.loads(docs)
```
ThomasLaPiana, line 175: built-in test, this will fail in CI if we add something to the docs that doesn't get exported properly

src/fides/core/config/docs.py
```diff
@@ -0,0 +1,181 @@
...
+    # assert (
+    #     "# TODO" not in docs
+    # ), "All fields require documentation, no description TODOs allowed!"
+
+    with open(outfile_path, "w", encoding="utf-8") as output_file:
+        output_file.write(docs)
+        print(f"Exported configuration file to: {outfile_path}")
+
+    # Verify it is a valid Fides config file
+    get_config(outfile_path)
```
ThomasLaPiana, line 187: built-in test. If the file is confirmed good `toml`, then the config will throw an uncaught exception when it gets loaded and should warn us

ThomasLaPiana:
I'm fixing up the test coverage but this is good for people to start reviewing 🙏

ThomasLaPiana:
docstrings from sean:

* redis.decode_responses — this is a boolean value that instructs Redis to automatically decode any value fetched from the cache
* cors_origin_regex — this is a regex string that can be used to set CORS origin allowlist patterns. e.g. http://localhost/:.* to allow any localhost port

SteveDMurphy (approved):
Sorry for the delays in reviewing! Spent some time working through how the opt out functionality is changing since it has snuck up on us a couple of times in the past. Added one question and a couple nits but everything checked out great for me 🙌🏽

src/fides/cli/utils.py
```diff
@@ -144,70 +143,47 @@ def register_user(config: FidesConfig, email: str, organization: str) -> None:
...
     """
 
-    config_updates: Dict[str, Dict] = {}
-    if config.user.analytics_opt_out is None:
-        click.echo(OPT_OUT_COPY)
-        config.user.analytics_opt_out = bool(
-            input(OPT_OUT_PROMPT + "\n").lower() == "n"
+    # Check the analytics opt out env var
+    # if getenv("FIDES__USER__ANALYTICS_OPT_OUT", "true") != "false":
+    #     return config
```
TheAndrewJackson, line 157: Is this commented code supposed to be here?

```suggestion

```

src/fides/core/config/utils.py
```diff
@@ -4,6 +4,29 @@
...
+    with open(config_path, "r", encoding="utf8") as config_file:
+        previous_config = config_file.read()
+        new_config = previous_config.replace(
+            f"{key} = {old_value}", f"{key} = {new_value}"
+        )
+
+    with open(config_path, "w", encoding="utf8") as config_file:
+        config_file.write(new_config)
+
+    print(f"Config key: {key} value changed: {old_value} -> {new_value}")
```
TheAndrewJackson, line 27: Would this print sensitive values to stdout?

docs/fides/docs/config/index.md
```diff
@@ -0,0 +1,103 @@
...
+
+### CLI
+
+To view your application configuration via the CLI, run:
+
+```sh
+fides view config
+```
+
+This will show _all_ configuration variables, including sensitive ones.
```
TheAndrewJackson, line 56: This is out of scope of this PR, but would it be possible to make this not log sensitive values by default and add a opt it option to out the sensitive values. Something like `fides view config --show-sensitive-values`. I'm sure at some point someone is going to just run `fides view config` blindly and be surprised that it logged the sensitive values.

TheAndrewJackson (approved):
Awesome work here. I love this change! This will help a lot with understanding how each option works in `fides.toml`. I just left a few small not blocking nits

ThomasLaPiana:
@adamsachs I saw you made some docs updates to the config, could you help me mirror them here if they aren't already in `main`? (it doesn't look like it)

src/fides/cli/utils.py
ThomasLaPiana: yes I've fixed that....forgot to re-enable. Newer commits will respect the env var :) 

disabled it for testing since its automatically set in the docker container

src/fides/cli/utils.py
```diff
@@ -144,70 +143,47 @@ def register_user(config: FidesConfig, email: str, organization: str) -> None:
     )
 
 
-def check_and_update_analytics_config(
-    config: FidesConfig, config_path: str
-) -> FidesConfig:
+def request_analytics_consent(config: FidesConfig, opt_in: bool = False) -> FidesConfig:
```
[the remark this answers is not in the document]
ThomasLaPiana: Correct

src/fides/core/config/utils.py
ThomasLaPiana: yes, but currently this is specifically only designed to be used in this one specific situation. Currently there is no code that calls this that would have sensitive info in it

docs/fides/docs/config/index.md
ThomasLaPiana: Do you have concerns around logging sensitive info to a CLI user? This is a CLI-specific command, so any values that get logged they would have access to directly via their local config file or env vars anyway I think

TheAndrewJackson: That's true. Since they already have access locally it seems fine.

It's less about if it's secure and more about what the user expects. If CLI users are expecting this command to log everything then I'm on board

ThomasLaPiana: I don't know why they wouldn't to be honest, but maybe? I can add a note to that effect in another PR I have open to update the CLI docstrings

adamsachs:
> @adamsachs I saw you made some docs updates to the config, could you help me mirror them here if they aren't already in main? (it doesn't look like it)

@ThomasLaPiana i think we're good? this PR https://github.com/ethyca/fidesdocs/pull/52 had my changes to the docs, but i don't think what you've done here interacts with that at all? in any case, those config properties are properly annotated as pydantic `Field`s with a good `description` in `database_settings.py`, so i think we're all covered?

let me know if i'm missing something though.

ThomasLaPiana:
> > @adamsachs I saw you made some docs updates to the config, could you help me mirror them here if they aren't already in main? (it doesn't look like it)
> 
> @ThomasLaPiana i think we're good? this PR [ethyca/fidesdocs#52](https://github.com/ethyca/fidesdocs/pull/52) had my changes to the docs, but i don't think what you've done here interacts with that at all? in any case, those config properties are properly annotated as pydantic `Field`s with a good `description` in `database_settings.py`, so i think we're all covered?
> 
> let me know if i'm missing something though.

We good we good! My bad, thanks for replying 😄

PR #2700, opened 2023-02-24
Add toggle for Systems with classified data flows

SteveDMurphy:
Closes [fidesplus#400](https://github.com/ethyca/fidesplus/issues/400)

### Code Changes

* [x] Duplicate the toggle component for clarity
* [x] Add and implement the toggle component against classified systems

### Steps to Confirm

(executed while running fidesplus using the branch from [fidesplus#635](https://github.com/ethyca/fidesplus/pull/635)

* [ ] To make it easier I altered the state text for those that had labels as well
* [ ] Ran a classification and then noted the difference between the toggled states

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### Description Of Changes

Worked both PRs at the same time


SteveDMurphy:
I'm sorry @allisonking !! I thought I added this one at the same time as the backend 🤦🏽

CHANGELOG.md
```diff
@@ -1,8 +1,5 @@
-<<<<<<< HEAD
```
allisonking, line 1: uh oh we've got some merge conflict markers!

CHANGELOG.md
```diff
@@ -25,6 +22,7 @@ The types of changes are:
 * Add API support for messaging config properties [#2551](https://github.com/ethyca/fides/pull/2551)
 * Access and erasure support for Kustomer [#2520](https://github.com/ethyca/fides/pull/2520)
 * Added the `erase_after` field on collections to be able to set the order for erasures [#2619](https://github.com/ethyca/fides/pull/2619)
+* Add a toggle to filter the system classification to only return those with classification data [#2700](github.com/ethyca/fides/pull/2700)
```
allisonking, line 25:
```suggestion
* Add a toggle to filter the system classification to only return those with classification data [#2700](https://github.com/ethyca/fides/pull/2700)
```

not 100% sure it matters, but might as well have it :D

clients/admin-ui/src/features/system/ClassifySystemsTable.tsx
```diff
@@ -40,6 +59,7 @@ const ClassifySystemsTable = ({ systems }: { systems: System[] }) => {
                 _hover={{ bg: "gray.50" }}
                 cursor="pointer"
                 onClick={() => handleClick(system)}
+                hidden={hideEmpty && !classifyInstance?.has_labels}
```
allisonking, line 62: instead of using the `hidden` attribute (which I believe would still include the component in the DOM, just with a `hidden` attribute) it's more common to filter at the data level (to be data-driven!), then always have the DOM components render the same way (this way, they're not included in the DOM at all, since they weren't in the data). One way to do this would be something like, before the parent `return`, i.e. L36:

```typescript
  const filteredSystems = hideEmpty
    ? systems.filter(
        (system) => classifyInstanceMap.get(system.fides_key)?.has_labels
      )
    : systems;
```

and then you can keep the same `<Tr />` component that was already there, but replace the `systems` on L53 with `filteredSystems`

here's another example: https://github.com/ethyca/fides/blob/SteveDMurphy-400-toggle-system-classify/clients/admin-ui/src/features/system/SystemsManagement.tsx/#L19

allisonking (changes requested):
the changelog is the only really necessary change—the typescript suggestion is optional since both ways are ok!

CHANGELOG.md
SteveDMurphy: oof, thanks! There were SO MANY, I tried to use GitHub and must have gotten lost 🙇🏽

clients/admin-ui/src/features/system/ClassifySystemsTable.tsx
SteveDMurphy: thank you @allisonking ! Going down that road now, I wasn't sure if there was a cost/benefit either way and happy to know that best path is what you are suggesting 🙌🏽

CHANGELOG.md
```diff
@@ -76,15 +77,9 @@ The types of changes are:
 
 ### Added
 
-<<<<<<< HEAD
-- Added new Wunderkind Consent Saas Connector [#2600](https://github.com/ethyca/fides/pull/2600)
-- Added new Sovrn Email Consent Connector [#2543](https://github.com/ethyca/fides/pull/2543/)
-- Log Fides version at startup [#2566](https://github.com/ethyca/fides/pull/2566)
-=======
```
SteveDMurphy, line 83: @allisonking I think looking at main these might still be correct now? A little unsure if we keep overwriting things somewhere since it sounded like these might have been fixed 🤷🏽

CHANGELOG.md
allisonking: oops I also totally didn't realize you were _removing_ the markers! I think I just saw the markers and went "oh no!"

CHANGELOG.md
allisonking: yeah it sounds like it was fixed in main whew

allisonking (approved):
💯

tests/ops/integration_tests/saas/test_delighted_task.py
```diff
@@ -21,6 +21,9 @@ def test_delighted_connection_test(delighted_connection_config) -> None:
     get_connector(delighted_connection_config).test_connection()
 
 
+@pytest.mark.integration_saas
+@pytest.mark.integration_delighted
+@pytest.mark.asyncio
```
SteveDMurphy, line 26: @galvana can you confirm these should look like this? Seeing some failing tests and taking a stab at it

SteveDMurphy: Confirmed that when adding the markers that apply to other tests everything passes as expected

galvana: That looks right to me! Thanks for fixing it

PR #2703, opened 2023-02-25
Overhaul CLI documentation and aesthetics

### LINKED ISSUE #2702: Use `rich_click` instead of `click`

The CLI is a bit stale looking, and deserves to be more sleek and modern-feeling.

Using `rich_click` as a drop-in replacement for `click` would give us the visual customization options

Using `typer`, which also has `rich` support, but this would required a much broader rewrite when the real goal here is a more aesthetic CLI, not necessarily a different CLI backend

### LINKED ISSUE #2704: Add a command to view the local credentials file

Users should be able to see their credentials file without having to dig around on their computer for it

We already have a `view` subcommand so I think the following makes sense `fides view credentials`

ThomasLaPiana:
Closes #2702 
Closes #2704 

### Code Changes

* [x] add `rich-click` to `requirements`
* [x] update imports
* [x] add a file for tweaking formatting and update colors
* [x] add a new CLI command for viewing your local credentials file
* [x] take a pass at updating config documentation. The list of top-level commands to keep track of:
  * [x] annotate
  * [x] db
  * [x] delete
  * [x] deploy
  * [x] evaluate
  * [x] export
  * [x] generate
  * [x] get
  * [x] init
  * [x] ls
  * [x] parse
  * [x] pull
  * [x] push
  * [x] scan
  * [x] status
  * [x] user
  * [x] view
  * [x] webserver
  * [x] worker
  * [x] Options and arguments in `options.py`
  * [x] top-level `fides -h`
* [x] open a follow-up issue to discuss overall CLI design, there are some things here that don't feel great as a user: #2718

### Steps to Confirm

* [ ] autogenerated CLI docs still work
* [ ] no visual wonkiness is detected
* [ ] run `fides view credentials`, confirm failure (no file)
* [ ] run `fides user login`, do the thing, then `fides view credentials` and verify the output

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### Description Of Changes

The aesthetics of the UI did not spark joy, so the `rich_click` package was added as a drop-in replacement to `click` that gives us great aesthetics for free while also being highly customizable.

An important note is that as part of this change, all help text rendered by the CLI will be assumed to be Markdown


ThomasLaPiana:
little demo of what this looks like

image

ThomasLaPiana:
@SteveDMurphy do the `export` commands need to have the `dry` flag other than the datamap export?

SteveDMurphy:
> @SteveDMurphy do the `export` commands need to have the `dry` flag other than the datamap export?

They were primarily added for running tests against the commands, we can probably revamp those to either check a temp directory that the file exists if we want to get rid of them? (overlap with `ls` now I imagine)

Separately, @sanders41 and I were discussing the value of keeping all of the export commands, potentially removing them as something else to maintain and being able to simplify the export datamap + custom fields stuff. Would be curious to hear your thoughts too!

src/fides/cli/commands/view.py
```diff
@@ -27,12 +28,30 @@ def view_config(
     ctx: click.Context, section: str = "", exclude_unset: bool = False
 ) -> None:
     """
-    Prints the configuration values being used.
+    Prints the configuration values being used as a `toml` file.
     """
     config = ctx.obj["CONFIG"]
```
pattisdr, line 33: Now that Adam has some work merged that allows you to override some of the config variables via the API, this wouldn't necessarily show the config value that's being used, as the API values override.

src/fides/cli/__init__.py
```diff
@@ -69,7 +73,11 @@
...
     """
-    The parent group for the Fides CLI.
+    __Command-line tool for the Fides privacy enginering platform.__
+
+    ---
+
+    _Note: The common MANIFESTS_DIR argument _always_ defaults to ".fides/" if not specified._
     """
 
     ctx.ensure_object(dict)
```
pattisdr, line 83: Also typo on L86 Dynamically

src/fides/cli/options.py
```diff
@@ -182,7 +185,7 @@ def aws_region_option(command: Callable) -> Callable:
     command = click.option(
         "--region",
         type=str,
-        help="Use region option to connect to aws. Requires options --access_key_id, --secret_access_key and --region",
+        help="Connec to AWS using a specific `Region`. _Requires options `--access_key_id`, `--secret_access_key` & `--region`._",
```
pattisdr, line 188: Typo connect

src/fides/cli/options.py
```diff
@@ -40,25 +44,22 @@ def fides_key_option(command: Callable) -> Callable:
...
+        "manifests_dir", type=click.Path(exists=True), default=".fides/"
     )(command)
     return command
 
 
 def dry_flag(command: Callable) -> Callable:
     "Add a flag that prevents side-effects."
     command = click.option(
-        "--dry", is_flag=True, help="Prevent the persistance of any changes."
+        "--dry", is_flag=True, help="Do not upload results to the Fides webserver."
```
pattisdr, line 62: Looks like there's other places this dry flag could be used?

src/fides/cli/options.py
```diff
@@ -162,7 +165,7 @@ def aws_access_key_id_option(command: Callable) -> Callable:
     command = click.option(
         "--access_key_id",
         type=str,
-        help="Use access key id option to connect to aws. Requires options --access_key_id, --secret_access_key and --region",
+        help="Connect to AWS using an `Access Key ID`. _Requires options `--access_key_id`, `--secret_access_key` & `--region`._",
```
pattisdr, line 168: pretty!

pattisdr (commented):
CLI docs formatting is 💯 .  Just noted minor things as I experimented.

src/fides/cli/commands/generate.py
```diff
@@ -127,7 +115,7 @@ def generate_dataset_bigquery(
 @click.pass_context
 def generate_system(ctx: click.Context) -> None:
     """
-    Generate fides System resources
+    Generate Fides systems from .
```
TheAndrewJackson, line 118: Is this line missing something or is `.` supposed to be the working dir?

src/fides/cli/options.py
```diff
@@ -142,7 +145,7 @@ def okta_org_url_option(command: Callable) -> Callable:
     command = click.option(
         "--org-url",
         type=str,
-        help="Use org url option to connect to okta. Requires options --org-url and --token",
+        help="Connect to Okta using an 'Org URL'. Requires options '--org-url' and '--token'.",
```
TheAndrewJackson, line 148:
```suggestion
        help="Connect to Okta using an 'Org URL'. _Requires options `--org-url` and `--token`._",
```

TheAndrewJackson (commented):
I like these updates. Much more concise throughout. It's awesome that you're able to use markdown in the help copy. I'm not used to seeing it in the terminal.


I found one instance of missing markdown formatting. Picture below:

Screenshot 2023-03-01 at 19 41 59

src/fides/cli/commands/generate.py
TheAndrewJackson: If it is supposed to the working dir maybe it could be wrapped in backticks so the period is rendered as a code snippet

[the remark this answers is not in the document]
ThomasLaPiana: nope!

src/fides/cli/__init__.py
```diff
@@ -1,14 +1,17 @@
...
 
-import click
+import rich_click as click
 from fideslog.sdk.python.client import AnalyticsClient
 
 import fides
 from fides.cli.utils import check_server
 from fides.core.config import get_config
 
+from . import cli_formatting
```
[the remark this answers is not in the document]
ThomasLaPiana: Yep!

src/fides/cli/commands/generate.py
ThomasLaPiana: typo

src/fides/cli/options.py
ThomasLaPiana: I tried to use specific instances of this flag elsewhere when a different `help` message made more sense

src/fides/cli/commands/export.py
```diff
@@ -111,17 +111,9 @@ def export_datamap(
...
-    The default organization is used, however a custom one can be
-    passed if required.
-
-    A custom manifest directory can be provided for the output location.
-
-    The csv flag can be used to output data as csv, while the dry
-    flag can be used to return data to the console instead.
     """
     config = ctx.obj["CONFIG"]
     _export.export_datamap(
```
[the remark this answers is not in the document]
ThomasLaPiana: good catch, @SteveDMurphy for the logic here, do you think it makes sense to do an early check that the taxonomy has datasets, systems and an organization before continuing?

ThomasLaPiana:
> > @SteveDMurphy do the `export` commands need to have the `dry` flag other than the datamap export?
> 
> They were primarily added for running tests against the commands, we can probably revamp those to either check a temp directory that the file exists if we want to get rid of them? (overlap with `ls` now I imagine)
> 
> Separately, @sanders41 and I were discussing the value of keeping all of the export commands, potentially removing them as something else to maintain and being able to simplify the export datamap + custom fields stuff. Would be curious to hear your thoughts too!

I have thoughts on the entire CLI design, but export did stick out as something that confused me, which means there's a high likelihood it would confuse users too. I'll follow this up in the other issue for CLI design #2718

src/fides/cli/commands/view.py
ThomasLaPiana: Hmmmm, would a better solution here be to have a separate command that will hit the config API endpoint and show the configuration being used by the server? Sans sensitive values

It is quite possible and even expected the local configuration values for the CLI user will be different than the values that webserver is using

ThomasLaPiana: For now I'm going to note this in the help string, but will hold off on adding that command until there is a concrete need for it

ThomasLaPiana:
Patch coverage is down but overall coverage is up. Testing the remaining patch pieces is not worth it from a time perspective (and complexity of testing those specific pieces)

src/fides/cli/commands/export.py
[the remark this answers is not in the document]
ThomasLaPiana: Thinking this would be best for another small PR though 🙂

src/fides/cli/commands/view.py
pattisdr: I didn't realize the cli user config could be different than the CLI user so do what you think is best here.

By the way, Adam did add a flag to the config API endpoint that lets you surface API-set config as well.

pattisdr (approved):

src/fides/cli/commands/export.py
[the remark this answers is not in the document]
SteveDMurphy: Oof, ya that could be a lot more graceful... Datasets are technically optional here but having at least one system is a requirement. Having a fully-fledged organization is also sort of a compliance component that could (should?) be used as part of the `audit` command. Open to ideas/opinions on how strict we think it makes sense to be when exporting a datamap

src/fides/cli/commands/view.py
ThomasLaPiana: I meant as in, the configuration values of the webserver won't necessarily match the configuration values of a CLI instance. Therefore as a CLI user `fides view config` is useful for seeing what I have configured locally for my own instance. I wouldn't expect people to be running this on the webserver container, but that's an issue of documenting that expectation.

If I'm running locally and I run `fides view config`, I wouldn't expect to see values that are only configured on the webserver (especially since we don't required a webserver)

It sounds like we should document that for checking webserver configuration values specifically, one needs to use the API endpoint.

PR #2706, opened 2023-02-27
Add `execution.require_manual_request_approval` to supported config API

adamsachs:
In initial config API implementation, I forgot about the "manual request approval" setting and that we should support that through the API, in order to allow the UI-based system config to turn this on by default, when a messaging provider is configured.

This PR adds in support for setting this property via the config API.

### Code Changes

* allow/support `execution.require_manual_request_approval` setting via config API
* update our code references to this property to use the `config_proxy` rather than the basic `CONFIG` object, so that API changes are picked up by the app during privacy request execution
* some small fixes/cleanup in related areas of config API

### DIFF
--- a/src/fides/api/ops/schemas/application_config.py
+++ b/src/fides/api/ops/schemas/application_config.py
@@ -41,6 +41,9 @@ class NotificationApplicationConfig(BaseSchema):
     send_request_review_notification: Optional[bool]
     notification_service_type: Optional[str]
 
+    class Config:
+        extra = Extra.forbid
+
     @validator("notification_service_type", pre=True)
     @classmethod
     def validate_notification_service_type(cls, value: str) -> Optional[str]:
--- a/src/fides/api/ops/schemas/application_config.py
+++ b/src/fides/api/ops/schemas/application_config.py
@@ -57,7 +57,8 @@ class NotificationApplicationConfig(BaseSchema):
 
 
 class ExecutionApplicationConfig(BaseSchema):
-    subject_identity_verification_required: bool
+    subject_identity_verification_required: Optional[bool]
+    require_manual_request_approval: Optional[bool]
 
     class Config:
         extra = Extra.forbid
--- a/src/fides/api/ops/api/v1/endpoints/privacy_request_endpoints.py
+++ b/src/fides/api/ops/api/v1/endpoints/privacy_request_endpoints.py
@@ -1676,7 +1676,7 @@ def create_privacy_request_func(
                     privacy_request_data.identity,
                     config_proxy.notifications.notification_service_type,
                 )
-            if not CONFIG.execution.require_manual_request_approval:
+            if not config_proxy.execution.require_manual_request_approval:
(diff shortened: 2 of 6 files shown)


src/fides/api/ops/schemas/application_config.py
```diff
@@ -57,7 +60,8 @@ def validate_notification_service_type(cls, value: str) -> Optional[str]:
 
 
 class ExecutionApplicationConfig(BaseSchema):
-    subject_identity_verification_required: bool
+    subject_identity_verification_required: Optional[bool]
+    require_manual_request_approval: Optional[bool]
```
sanders41, line 64: Instead of `Optional` could these have a default value (`False` I assume) so you don't later have to check for both `None` and either `True` or `False`?

```suggestion
    subject_identity_verification_required: bool = False
    require_manual_request_approval: bool = False
```

adamsachs: hmmm, well here a default value actually means something different than none/null. if the value is set here via API (even by default assignment), it's going to override what's been set via non-API config (i.e. env var or .toml), and i don't think that's the behavior we want. does that make sense?

sanders41: This is referring to the value in `src/fides/api/ops/schemas/application_config.py` and you want this to determine the default instead? If yes, this make sense.

```py
        subject_identity_verification_required: bool = Field(
        default=False,
        description="Whether privacy requests require user identity verification.",
    )
```

ThomasLaPiana: Both of those are correct @sanders41 

Config values must use `Field` (or it'll trigger a CI failure, after my PR is merged) and `Optional` fields are also no longer accepted (also a CI failure). This is due to the metadata here being used to generate valid `toml`

adamsachs: right - i think this all adds up fine, but i want to make sure we're all on the same page.

here my code is effectively an API schema/model - it's _not_ application config, even though it will eventually be used alongside/reconciled with the 'proper' application config. i want to keep this field as optional in the API schema, and i do _not_ want to default to a certain value, because the config proxy will take API-set properties over the corresponding config-set properties. if we were to default here in the API model, then it'd mean that default would _override_ the default set on the application config - and that's not desirable.

but i think this is perfectly in line with @ThomasLaPiana 's changes -- essentially the `subject_identity_verification_required` field in the 'proper' application config sets the default, and here in the API-level we don't provide any sort of default, we allow it to be null/None which essentially means - "nothing has been set for this property via API, instead rely on the config setting from the 'proper' application config".

sanders41: This makes sense to me. @ThomasLaPiana will this pass in your new CI changes?

sanders41: Looked into this closer and since it isn't part of the application config files it should be fine in CI.

sanders41 (approved):

PR #2714, opened 2023-02-27
[Backend] Add System Manager (Data Steward) Foundation

### LINKED ISSUE #2607: Add New System Manager Endpoints
❗ Blocked by https://github.com/ethyca/fides/issues/2606

This ticket would add the endpoints to let us link users to systems

- Endpoint that adds a user as a system manager
- View systems attached to a user
- View a single system attached to a user
- Remove a user from a system

### LINKED ISSUE #2609: Add Object-Based Permissions Foundation
❗ Blocked by https://github.com/ethyca/fides/issues/2606

This ticket would add a new way in which users can be granted permission to a specific resource due to their system affiliation. 

- Add Role-> Scopes mapping for data_steward
- Identify endpoints where data_steward permissions can be added
- Add modified verify_oauth_client Security dependences that check if the user has_role_based_scopes OR has_resource_based_scopes. The resource ids/keys also need to be parameters of these functions.

pattisdr:
❗ CONTAINS MIGRATION Bump downrev before merging

Closes #2609 
Closes #2607 

#### Database/modeling changes
- [x] Add a `SystemManager` many-to-many table to link users to systems
- [x] Add `client.systems` field which is a list of system ids for persisting this information when the user is logged in
- [x] Add model methods `FidesUser.add_as_system_manager` and `FidesUser.remove_as_system_`manager to add and remove users from the system <-> user many-to-many table and update the related client if applicable.

#### Authentication changes 
- [x] When creating tokens, persist a list of system ids to the token for which the user is manager.
- [x] Add new `verify_oauth_client_for_system_from_request_body` and `verify_oauth_client_for_system_from_fides_key` for authorizing tokens when updating and deleting systems.  In addition to the roles and scopes checks these endpoints already have, add the ability for the user to update or delete if they are system manager of that particular system.
- [x] Split system update and system delete endpoints off from the crud endpoints defined in a loop due to the new authorization.
- [x] Split recently-added scopes: `CLI_OBJECTS_CREATE`, `CLI_OBJECTS_DELETE`, `CLI_OBJECTS_READ` and `CLI_OBJECTS_UPDATE`  into more granular scopes per object per endpoint so we can create roles with very specific scopes on the ctl-side.
- [x] On login, update the client with the latest scopes/roles/systems if one already exists.. It is possible they've been updated but the change hasn't persisted to the client.

#### Data migration
- [x] For users that have  `CLI_OBJECTS_CREATE`, `CLI_OBJECTS_DELETE`, `CLI_OBJECTS_READ` and `CLI_OBJECTS_UPDATE` in their pre-existing `fidesuserpermission.scopes` and `client.scopes` columns, add the new corresponding granular scopes to this list, as we're changing which scopes go on these endpoints.

### Steps to Confirm
* [x] Set CONFIG.security.env to `prod` so authentication is live on the ctl endpoints
* [x] run `nox -s "fides_env(dev)"`
* [x] Login as the root user and create a new user with minimal scopes with username "test"
* [x] Bring up a python shell and remove the user's scopes, give them a viewer role, and make them a system manager of a pre-existing system
```python
from fides.lib.db.base import FidesUser
from fides.api.ctl.sql_models import System
from fides.lib.db.session import get_db_session
from fides.core.config import CONFIG
SessionLocal = get_db_session(CONFIG)
db = SessionLocal()

u = db.query(FidesUser).filter(FidesUser.username=='test').first()
s = db.query(System).first()

u.set_as_system_manager(db, s)
u.permissions.scopes = []
u.permissions.roles = ["viewer"]
db.add(u)
db.commit()
```

- [x] Verify we have one system and no scopes and viewer role
```bash
>>> u.permissions.roles
["viewer"]
>>> u.permissions.scopes
[]
>>> u.system_ids
['ctl_2a29bf06-7aba-47d5-8402-3c8a4f3d122e']
```
- [x] Login as "test"
- [x] Note that you can edit the system to which you're assigned, but not the other system.  
- [x] Note you can delete the system to which you're assigned but the the other system.
- [x] You should still have broader "get" scopes because of your Viewer role.

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

Add the ability for users to be System Managers.  This is our first example of **object-level** permissions instead of model-level permissions. Rather than being granted permissions to all resources of this type, System Managers are granted a limited set of permissions (now: system update and system delete) to only the specific systems to which they are assigned.

This is not a "role" - it's a separate additional optional way in which users can be granted permissions.  The extra auth checks need to be intentionally added to each system-related endpoint where we want system managers to have access.


tests/ctl/core/test_api.py
```diff
@@ -60,19 +71,44 @@ def test_generate_resource_urls_with_id(test_config: FidesConfig) -> None:
...
+        token_scopes: List[str] = [PRIVACY_REQUEST_CREATE]
+        auth_header = generate_auth_header(scopes=token_scopes)
+        result = _api.create(
+            url=test_config.cli.server_url,
+            resource_type=endpoint,
+            json_resource=manifest.json(exclude_none=True),
+            headers=auth_header,
+        )
+        assert result.status_code == 403
+
```
pattisdr, line 111: No tests broke when I split out the cli-objects:* scopes into individual scopes, so for each crud endpoint I'm adding a test that fails if it has the wrong scope, and a test that passes with just the singular correct scope.

src/fides/lib/oauth/system_manager.py
```diff
@@ -0,0 +1,5 @@
+from fides.api.ops.api.v1.scope_registry import SYSTEM_DELETE, SYSTEM_UPDATE
+
+# System managers are separate from roles, because you are just granted these
+# permissions to specific system(s) not to all resources of a given type
+SYSTEM_MANAGER_SCOPES = [SYSTEM_UPDATE, SYSTEM_DELETE]
```
pattisdr, line 5: Starting with these scopes for system managers and we can continue to expand as extend what they can do when they've been assigned to a system.

src/fides/lib/oauth/roles.py
```diff
@@ -67,14 +77,24 @@ class RoleRegistry(Enum):
...
+    EVALUATION_READ,
+    ORGANIZATION_READ,
     POLICY_READ,
     PRIVACY_REQUEST_READ,
     PRIVACY_REQUEST_NOTIFICATIONS_READ,
+    REGISTRY_READ,
     RULE_READ,
     SCOPE_READ,
     STORAGE_READ,
+    SYSTEM_READ,
```
pattisdr, line 97: Adding newly-created "read" scopes to the Viewer role, but not that the roles->scopes mapping will be overhauled here https://github.com/ethyca/fides/issues/2720

src/fides/lib/models/client.py
```diff
@@ -114,6 +122,7 @@ def create_access_code_jwe(self, encryption_key: str) -> str:
             JWE_PAYLOAD_SCOPES: self.scopes,
             JWE_ISSUED_AT: datetime.now().isoformat(),
             JWE_PAYLOAD_ROLES: self.roles,
+            JWE_PAYLOAD_SYSTEMS: self.systems,
```
pattisdr, line 125: When creating a token on login, add the `systems` (system_ids) from the client if applicable.

src/fides/lib/models/client.py
```diff
@@ -40,6 +42,7 @@ def __tablename__(self) -> str:
     salt = Column(String, nullable=False)
     scopes = Column(ARRAY(String), nullable=False, server_default="{}", default=dict)
     roles = Column(ARRAY(String), nullable=False, server_default="{}", default=dict)
+    systems = Column(ARRAY(String), nullable=False, server_default="{}", default=dict)
```
pattisdr, line 45: When adding a user to a system or removing a user from the system, their client.systems array will be updated with that system's id so it's ready to persist to the token on login.

src/fides/api/ops/models/system_manager.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+class SystemManager(Base):
+    """Many to many table to link users to systems as system managers.
+
+    Usually access via user.systems or system.users instead of via this table directly.
+    """
+
+    user_id = Column(String, ForeignKey("fidesuser.id"), primary_key=True)
+    system_id = Column(String, ForeignKey("ctl_systems.id"), primary_key=True)
```
pattisdr, line 13: Instead of accessing this many-to-many table directly, you'd likely use the FidesUser methods instead.

src/fides/api/ops/api/v1/scope_registry.py
```diff
@@ -169,18 +219,47 @@
...
+    CTL_DATASET_CREATE: "Create a ctl dataset",
+    CTL_DATASET_READ: "Read ctl datasets",
+    CTL_DATASET_DELETE: "Delete a ctl dataset",
+    CTL_DATASET_UPDATE: "Update ctl datasets",
+    CTL_POLICY_CREATE: "Create a ctl policy",
+    CTL_POLICY_READ: "Read ctl policies",
+    CTL_POLICY_DELETE: "Delete a ctl policy",
+    CTL_POLICY_UPDATE: "Update ctl policies",
     DATABASE_RESET: "Reset the application database",
+    DATA_CATEGORY_CREATE: "Create a data category",
```
pattisdr, line 231: ..Breaking CTL-scopes into individual scopes per endpoint so we can be more targeted about which specific permissions are granted to various roles and system managers.

src/fides/api/ctl/routes/util.py
```diff
@@ -80,3 +92,20 @@ async def forbid_if_editing_any_is_default(
...
+    "data_qualifier": DATA_QUALIFIER,
+    "data_subject": DATA_SUBJECT,
+    "data_use": DATA_USE,
+    "dataset": CTL_DATASET,
+    "evaluation": EVALUATION,
+    "organization": ORGANIZATION,
+    "policy": CTL_POLICY,
+    "registry": REGISTRY,
+    "system": SYSTEM,
+}
```
pattisdr, line 111: Used for dynamically adding the new scopes to the crud endpoints

pattisdr:
One of my tests are failing and hanging when run alongside all unit tests, but not when the its class of tests is run directly.  Looking into this -

src/fides/api/main.py
```diff
@@ -131,6 +138,12 @@ def create_fides_app(
         # This removes auth requirements for CLI-related endpoints
         # and is the default
         fastapi_app.dependency_overrides[verify_oauth_client_cli] = get_root_client
+        fastapi_app.dependency_overrides[
+            verify_oauth_client_for_system_from_request_body_cli
+        ] = get_system_schema
+        fastapi_app.dependency_overrides[
+            verify_oauth_client_for_system_from_fides_key_cli
+        ] = get_system_fides_key
```
pattisdr, line 146: System update and system delete endpoints have been broken out from the crud routes and have their own oauth verification methods.  Overriding here in "dev" mode to bypass authentication.

pattisdr:
@ThomasLaPiana do you have suggestions for bringing up `nox -s "fides_env(dev)"` with a security_env of `prod`?

This fails with:
```bash
2023-03-01 20:02:57.355 | INFO     | setup.dataset:create_dataset:32 - Upserted 1 Dataset(s)
2023-03-01 20:02:57.390 | INFO     | setup.dataset:create_dataset:52 - Created fides dataset via http://fides:8080/api/v1/connection/mongodb_connector/datasetconfig
Examples loaded successfully!
nox > Pushing example resources for Data Mapping tests... (demo_resources/*)
nox > docker exec -e ANALYTICS_OPT_OUT -e FIDES__CLI__ANALYTICS_ID -t fides-fides-1 fides push demo_resources/
Loaded config from: /fides/src/fides/data/test_env/fides.test_env.toml
Loading resource manifests from: demo_resources/
Taxonomy successfully created.
----------
Processing dataset resource(s)...
{'detail': 'Not Authorized for this action'}
nox > Command docker exec -e ANALYTICS_OPT_OUT -e FIDES__CLI__ANALYTICS_ID -t fides-fides-1 fides push demo_resources/ failed with exit code 1
nox > Session fides_env(test) failed.

```

ThomasLaPiana:
> @ThomasLaPiana do you have suggestions for bringing up `nox -s "fides_env(dev)"` with a security_env of `prod`? (Doesn't need to necessarily be this command, just how to run the test_env with prod)
> 
> This fails with:
> 
> ```shell
> 2023-03-01 20:02:57.355 | INFO     | setup.dataset:create_dataset:32 - Upserted 1 Dataset(s)
> 2023-03-01 20:02:57.390 | INFO     | setup.dataset:create_dataset:52 - Created fides dataset via http://fides:8080/api/v1/connection/mongodb_connector/datasetconfig
> Examples loaded successfully!
> nox > Pushing example resources for Data Mapping tests... (demo_resources/*)
> nox > docker exec -e ANALYTICS_OPT_OUT -e FIDES__CLI__ANALYTICS_ID -t fides-fides-1 fides push demo_resources/
> Loaded config from: /fides/src/fides/data/test_env/fides.test_env.toml
> Loading resource manifests from: demo_resources/
> Taxonomy successfully created.
> ----------
> Processing dataset resource(s)...
> {'detail': 'Not Authorized for this action'}
> nox > Command docker exec -e ANALYTICS_OPT_OUT -e FIDES__CLI__ANALYTICS_ID -t fides-fides-1 fides push demo_resources/ failed with exit code 1
> nox > Session fides_env(test) failed.
> ```

looks like we'll need to add an additional command in there to do the login, sorry for not catching that

There are examples elsewhere in the code, and there is a `LOGIN` constant in `constants_nox.py` that has the command too

pattisdr:
Perfect thank you @ThomasLaPiana https://github.com/ethyca/fides/pull/2714/commits/3e23946d05d4ef988d1400f9b3b76ad5ec1b7bbf

noxfiles/dev_nox.py
```diff
@@ -170,6 +171,8 @@ def fides_env(session: Session, fides_image: Literal["test", "dev"] = "test") ->
     session.run(
         *START_TEST_ENV, "fides-ui", "fides-pc", external=True, env=test_env_vars
     )
+    session.log("Logging in...")
+    session.run(*LOGIN, external=True)
```
pattisdr, line 175: System endpoints aren't authenticated in dev mode, you need to be in prod mode.  If you're in prod mode, we need to login when running nox -s fides_env(dev)

pattisdr:
@ThomasLaPiana I'm messing with system auth here, would you mind taking a look when you have some time?

ThomasLaPiana:
@pattisdr starting to work through this, might not finish the review until Monday though!

Was this considered critical for the `2.8.0` release?

src/fides/api/ops/api/v1/scope_registry.py
ThomasLaPiana: Thank you for cleaning up my mess 🙏

src/fides/lib/models/fides_user.py
```diff
@@ -1,16 +1,26 @@
...
-from typing import Any
+from typing import TYPE_CHECKING, Any, List
 
 from citext import CIText
 from sqlalchemy import Column, DateTime, String
 from sqlalchemy.orm import Session, relationship
 
+from fides.api.ops.common_exceptions import SystemManagerException
+
+# Intentionally importing SystemManager here
```
ThomasLaPiana, line 13: Why?

src/fides/lib/oauth/api/routes/user_endpoints.py
```diff
@@ -248,11 +248,12 @@ def perform_login(
...
             scopes=user.permissions.scopes,  # type: ignore
             roles=user.permissions.roles,  # type: ignore
+            systems=user.system_ids,  # type: ignore
             user_id=user.id,
         )
 
-    if not user.permissions.scopes and not user.permissions.roles:  # type: ignore
-        logger.info("User needs scopes and/or roles to login.")
+    if not user.permissions.scopes and not user.permissions.roles and not user.systems:  # type: ignore
+        logger.info("User needs scopes, roles, or systems to login.")
```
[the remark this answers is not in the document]
ThomasLaPiana: This seems logical, if you create a user with the inability to do anything, why bother logging in 😆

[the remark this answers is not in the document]
pattisdr: makes sense, updating to `warning`!

pattisdr:
> Was this considered critical for the 2.8.0 release?

@ThomasLaPiana No it does not need to go out in this release

src/fides/lib/models/fides_user.py
```diff
@@ -10,10 +10,10 @@
...
 from fides.api.ops.common_exceptions import SystemManagerException
 
-# Intentionally importing SystemManager here
-from fides.api.ops.models.system_manager import (  # type: ignore[unused-import]
-    SystemManager,
-)
+# Intentionally importing SystemManager here to build the FidesUser.systems relationship
+# from fides.api.ops.models.system_manager import (  # type: ignore[unused-import]
+#     SystemManager,
+# )
```
pattisdr, line 16: Whoops need to revert

src/fides/lib/oauth/api/routes/user_endpoints.py
```diff
@@ -248,11 +248,18 @@ def perform_login(
...
             scopes=user.permissions.scopes,  # type: ignore
             roles=user.permissions.roles,  # type: ignore
+            systems=user.system_ids,  # type: ignore
             user_id=user.id,
         )
-
-    if not user.permissions.scopes and not user.permissions.roles:  # type: ignore
-        logger.info("User needs scopes and/or roles to login.")
+    else:
+        # Refresh the client just in case
```
ThomasLaPiana, line 255: Is there is a possibility of it not being accurate otherwise?

ThomasLaPiana (commented):
I read through the code changes and it all looks/good makes sense, but when I was testing I got some weird behavior. It seemed like I didn't get assigned systems properly, as I was able to delete systems that I shouldn't have been assigned to

ThomasLaPiana:
I just looked back through the code and was trying to add a `Systems Under Management` section to the `fides user permissions` output, but it looks like the `systems` field isn't included as part of that response. I think that should be included as  part of checking user permissions, and thought it was but I didn't see it in the response

pattisdr:
Hey @ThomasLaPiana I didn't add systems to that CLI response but it is in the API response. I'll add the CLI section.  This might not have been clear, but you don't need to be a system manager to be able to delete the system if you have permissions via another route. For example, an `owner` has full admin and can perform any action.  Being a system manager more denotes formal responsibility for the system.

src/fides/lib/oauth/api/routes/user_endpoints.py
pattisdr: I was thinking if roles or scopes were added directly through the database, or through some other location besides the one API endpoint that updates user permissions.

pattisdr:
@ThomasLaPiana I've tried to add the system managers to the CLI command here https://github.com/ethyca/fides/pull/2714/commits/739d2ebe430b3685a925b3bd2bd71394616eb109

src/fides/api/ctl/migrations/versions/68a518a3c050_system_managers.py
```diff
@@ -0,0 +1,209 @@
...
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+from sqlalchemy import text
+from sqlalchemy.engine import ResultProxy
+from sqlalchemy.sql.elements import TextClause
+
+revision = "68a518a3c050"
+down_revision = "9f38dad37628"
```
seanpreston, line 22: I think you'll need to bump this up to `2f6aa5fe797a` after today's merges into `main`

pattisdr: yep have to get this up to date again

src/fides/api/ctl/routes/crud.py
```diff
@@ -83,7 +84,8 @@ async def validate_data_categories(
         status_code=status.HTTP_201_CREATED,
         dependencies=[
             Security(
-                verify_oauth_client_cli, scopes=[scope_registry.CLI_OBJECTS_CREATE]
+                verify_oauth_client_cli,
+                scopes=[f"{CLI_SCOPE_PREFIX_MAPPING[model_type]}:{CREATE}"],
```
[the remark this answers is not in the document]
seanpreston: Thanks for this Dawn, thinking deeper into it I just want to flag that creating scopes inside endpoints is a break from the convention of pre-defining them. Although this is useful it strikes me as an anti-pattern specifically for scopes since:
- there'll be no one place where we can see at a glance exactly what scopes are supported by the application, and
- no single place we'll be able to import them all from either (which is useful when creating the root client for instance)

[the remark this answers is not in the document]
pattisdr: Yes I pondered this too, there's pros and cons about defining these crud endpoints in a loop and these scopes are one of the cons. To push back a bit though, the scope registry still has all the scopes in a single place, and the root user is also given the scopes you see here.

seanpreston:
I've gone through the functional testing steps up top and it appears to have allowed me to delete both systems:

Screenshot 2023-03-07 at 22 22 21

Here's the shell output from the verification commands:

```
In [1]: from fides.lib.db.base import FidesUser
   ...: from fides.api.ctl.sql_models import System
   ...: from fides.lib.db.session import get_db_session
   ...: from fides.core.config import CONFIG
   ...: SessionLocal = get_db_session(CONFIG)
   ...: db = SessionLocal()
   ...: 
   ...: u = db.query(FidesUser).filter(FidesUser.username=='test').first()
   ...: s = db.query(System).first()
   ...: 
   ...: u.set_as_system_manager(db, s)
   ...: u.permissions.scopes = []
   ...: u.permissions.roles = ["viewer"]
   ...: db.add(u)
   ...: db.commit()

In [2]: u.permissions.roles
Out[2]: ['viewer']

In [3]: u.permissions.scopes
Out[3]: []

In [4]: u.system_ids
Out[4]: ['ctl_045a09d3-6afd-4e37-a4d3-7ddb23596770']
```

pattisdr:
@seanpreston are you positive that the "prod" settings are in effect? If not, there's no authentication on those endpoints at all - in which case anyone could delete the systems.

src/fides/lib/oauth/api/routes/user_endpoints.py
ThomasLaPiana: I would hope users aren't doing that but I also appreciate you preempting unexpected behavior and adding a safety net, good thinking!

ThomasLaPiana:
> Hey @ThomasLaPiana I didn't add systems to that CLI response ~but it is in the API response~. I'll add the CLI section. This might not have been clear, but you don't need to be a system manager to be able to delete the system if you have permissions via another route. For example, an `owner` has full admin and can perform any action. Being a system manager more denotes formal responsibility for the system.
> 
> EDIT: Sorry I got confused here, it's not in the get permissions endpoint, it's in the new endpoints where you get system managers #2726 I'm trying to keep "roles" and "systems" separate, to distinguish between model-level and object-level permissions. I can add this to the CLI section in that PR.

That makes perfect sense, thank you for the explanation!

pattisdr:
my instructions should have been more clear @seanpreston!

pattisdr:
Getting this up-to-date again with main

src/fides/api/ctl/migrations/versions/68a518a3c050_system_managers.py
pattisdr: Updated with main - this gets out of date about every day though! imagine we'll need a few more merges before this goes out.

ThomasLaPiana:
@pattisdr thank you for that CLI update! Looks great 😄 

I think we (in this case myself) should open up a follow-up PR to make `fides user create` more robust now that we allow people to specify roles. I had avoided adding that feature previously because having a user choose from those scopes was too complicated, but with roles it becomes much more feasible

seanpreston (approved):
Thanks @pattisdr — this works great 🚀

src/fides/api/ctl/routes/util.py
seanpreston: Thanks for highlighting! I missed this first pass.

pattisdr:
Thanks very much for your review @seanpreston, much appreciated! I'll hold off on merging just yet as this PR adds new scopes.  I think we should make sure this goes out as the same release as the UI work that Allison/Catherine are doing.

PR #2728, opened 2023-03-01
Allow template for sendgrid (if exists)

### LINKED ISSUE #1866: Support Twilio Email (SendGrid) Templates
**Note:** This ticket depends on #1975 

### Description
Our current email notifications are essentially only useful for testing; we need better templates to use for production use cases. We *could* write our own HTML templates for this, but we'd quickly end up wanting to provide a lot of configuration options for those templates which is a lot of work!

Meanwhile, both of our messaging providers support their own email editors to design & build transactional email templates:
* Mailgun: https://help.mailgun.com/hc/en-us/articles/6923289387675-Mailgun-s-Email-Editor#passport-start
* Twilio SendGrid: https://docs.sendgrid.com/glossary/transactional-email-templates

For example, our current emails look like this:
image

Meanwhile, Mailgun's editor looks like this (gives users a ton of control!):
image

...with the ability to insert template variables like:
image

### Assumptions
* A single email template _should_ be sufficient for all our emails at first
* Each email notification would have template variables for "subject" and "message"

### Acceptance Criteria
* Add support for using Sendgrid templates for all email notifications (identity verification, request receipt, request review, request completion)
* Add ability to automatically populate Sendgrid with a default "Fides" template (if the user does not want to create their own)
* Add support for using Twilio templates for all email notifications (identity verification, request receipt, request review, request completion)
* Add ability to automatically populate Twilio with a default "Fides" template (if the user does not want to create their own)

### Implementation Notes/Details
For an example of how the Mailgun API works:
```
curl -s --user 'api:YOUR_API_KEY' \
https://api.mailgun.net/v3/YOUR_DOMAIN_NAME/messages \
-F from='Sender Bob <sbob@YOUR_DOMAIN_NAME>' \
-F to='alice@example.com' \
-F subject='Hello' \
-F template='template.test' \
-F t:variables='{"title": "API documentation", "body": "Sending messages with templates"}'
```

SteveDMurphy:
Closes #1866

### Code Changes

* [x] Include the option of having a template email with SendGrid
* [x] ~Handle paging of template IDs~ Use max for now and circle back (as needed)

### Steps to Confirm

* [x] Create a template to use for testing
* [x] Submit a request to use the template

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### Description Of Changes

This is following the same path that was set out for Mailgun, only minor differences exist in how the template is discovered and applied. The pagination for the SDK is not functioning, so to hit the release and not over-engineer I just ended up setting the max. Will add a comment in the files changed as well to highlight and see if we want to add a follow on issue now or wait until it becomes a problem


src/fides/api/ops/service/messaging/message_dispatch_service.py
```diff
@@ -456,18 +457,32 @@ def _twilio_email_dispatcher(
...
             api_key=messaging_config.secrets[
                 MessagingServiceSecrets.TWILIO_API_KEY.value
             ]
         )
+
+        # the pagination via the client actually doesn't work
+        # in lieu of over-engineering this we can manually call
+        # the next page if/when we hit the limit here
+        response = sg.client.templates.get(
+            query_params={"generations": "dynamic", "page_size": 200}
```
adamsachs, line 471: seems fine to me to hardcode for now to get this across the line, per your comment in the description. a couple of notes though:

1. i'd definitely create a follow up issue to track the improvement just so it doesn't go completely forgotten/untracked
2. i'd put the hardcoded limit in a docs update (see general comment for making sure we update our docs) to make the current limitation clear to anyone using it, just to reduce the amount of digging we have to do if/when we run into the limitation
3. any reason you've chosen `200` here as the hardcoded page size (and effective limit)? it feels like it could be a bit higher here and we wouldn't get overwhelmed with data, while just reducing the likelihood that we're impacted by the hardcoding. but that's just based on an uninformed gut feeling, so please ignore me if i'm spouting nonsense!

adamsachs (approved):
i like the way this looks! simple and clean implementation. the only thing i found in the code is that hardcoded limit that you've called out, due to the limitation in the SDK/client - i think your decision to hardcode is totally fine for now, but i had a couple of notes on it. nothing critical but let me know what you think. and let me know if you're running into any issues grappling with codecov, happy to lend a hand if i can.

also, is there a `fidesdocs` ticket to update our docs explaining how to use this functionality?  let's make sure we at least ticket that before merging this in - ideally we can also just quickly make those doc updates since we've got a lot of the necessary content about templating already there for mailgun it seems, and it feels like the type of thing that if we don't update now, it may just sit stale unnecessarily.

last thing - were you able to run a manual end-to-end test actually using a template in sendgrid? happy to help with that if needed - but seems like you've likely done that considering the pagination bug you've found? (i just want to be sure!)

src/fides/api/ops/service/messaging/message_dispatch_service.py
SteveDMurphy: Thanks again @adamsachs !!

1. Docs ticket still needed, will create here now and start working.
2. Great idea, I probably wouldn't have done that 🙌🏽 
3. The hard-coded limit of `200` comes from the Sendgrid API itself (I put 1000 and the response conveniently lists that max as 200)

For the end to end test, I would love some help here as I hit a snag following the docs and was about to write a bug issue. Will reach out in Slack

SteveDMurphy:
Docs issue created -> [fidesdocs#68](https://github.com/ethyca/fidesdocs/issues/68)

src/fides/api/ops/service/messaging/message_dispatch_service.py
adamsachs: 👍 nice, thanks for the follow up and clarification. sendgrid API 🙄

PR #2739, opened 2023-03-02
Role assignment UI

allisonking:
Closes https://github.com/ethyca/fidesplus/issues/641

~~This branches off of multiple branches, so the diff will be unseemly until those branches are merged~~

Branches are merged, though commit history has many of the commits from https://github.com/ethyca/fides/pull/2682 because of the squashed merge + I'm not good enough at rebasing 🙃 

### Code Changes

* [ ] See the ticket

### Steps to Confirm

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### Description Of Changes


clients/admin-ui/src/features/common/nav/v2/nav-config.test.ts
```diff
@@ -57,8 +57,8 @@ describe.skip("configureNavGroups", () => {
       config: NAV_CONFIG,
       hasSystems: true,
       userScopes: [
-        ScopeRegistryEnum.CLI_OBJECTS_CREATE,
-        ScopeRegistryEnum.CLI_OBJECTS_READ,
+        ScopeRegistryEnum.SYSTEM_CREATE,
+        ScopeRegistryEnum.SYSTEM_READ,
```
allisonking, line 61: I didn't follow too closely the decisions made last week, but I wonder if the latest in `main` would mean we wouldn't have to do this here? might be worth a merge to see if the tests "just pass"

clients/admin-ui/src/features/user-management/AssignSystemsTable.tsx
```diff
@@ -66,9 +56,14 @@ export const AssignSystemsDeleteTable = () => {
                 aria-label="Unassign system from user"
                 icon={<TrashCanSolidIcon />}
                 size="xs"
-                onClick={() => handleDelete(system)}
+                onClick={removeSystemModal.onOpen}
                 data-testid="unassign-btn"
               />
+              <RemoveSystemModal
```
allisonking, line 62: I'd prefer if we're able to use the `ConfirmationModal` in `common` just so that if we have to update our confirmation modals we can do it all in once place. but let me know if there's a good reason to keep this one as a separate component 👍

clients/admin-ui/src/features/user-management/NewUserForm.tsx
```diff
@@ -20,9 +23,11 @@ const NewUserForm = () => {
 
   const handleSubmit = (values: FormValues) => {
     const b64Password = utf8ToB64(values.password);
-    return createUser({ ...values, password: b64Password });
+    createUser({ ...values, password: b64Password });
+    router.push(`${USER_MANAGEMENT_ROUTE}`);
   };
 
+  // @ts-ignore
```
allisonking, line 30: is there a reason for this ts-ignore?

clients/admin-ui/src/features/user-management/NewUserForm.tsx
```diff
@@ -20,9 +23,11 @@ const NewUserForm = () => {
 
   const handleSubmit = (values: FormValues) => {
     const b64Password = utf8ToB64(values.password);
-    return createUser({ ...values, password: b64Password });
+    createUser({ ...values, password: b64Password });
+    router.push(`${USER_MANAGEMENT_ROUTE}`);
```
allisonking, line 27: I think we may still want the nice toasts that indicate the user was created successfully here. also I think we may want to redirect to the profile of this user? are we able to grab their created id from `createUser` and then build the URL off of that?

clients/admin-ui/src/features/user-management/PasswordManagement.tsx
```diff
@@ -1,39 +1,61 @@
...
+    <Box>
+      {activeUserId ? (
+        <HStack>
+          {isOwnProfile ? <UpdatePasswordModal id={activeUserId} /> : null}
+          <Restrict scopes={[ScopeRegistryEnum.USER_PASSWORD_RESET]}>
+            <NewPasswordModal id={activeUserId} />
+          </Restrict>
+
+          {!isNewUser && user ? (
+            <Box>
```
allisonking, line 45: I'd prefer to have this delete button not inside `PasswordManagement` since it doesn't have to do with password management. can we hoist it to be a sibling?

clients/admin-ui/src/features/user-management/RemoveSystemModal.tsx
```diff
@@ -0,0 +1,88 @@
...
+import { errorToastParams, successToastParams } from "~/features/common/toast";
+import { System } from "~/types/api";
+
+import { useRemoveUserManagedSystemMutation } from "./user-management.slice";
+
+const useRemoveSystemModal = ({ onClose }: { onClose: () => void }) => {
+  const toast = useToast();
+  const [removeUserManagedSystemTrigger] = useRemoveUserManagedSystemMutation();
+
+  const handleRemoveSystem = async (system: System, activeUserId?: string) => {
```
allisonking, line 27: if you replace this with the ConfirmationModal, you could pass this func into the modal's `onConfirm`

clients/admin-ui/src/features/user-management/UserForm.tsx
```diff
@@ -84,28 +97,50 @@ const UserForm = ({
...
+                  display="flex"
+                  alignItems="center"
+                  fontSize="sm"
+                  fontWeight="semibold"
+                >
+                  Profile
+                </Text>
+                <Box marginLeft="auto">
+                  <PasswordManagement user={user} isNewUser={isNewUser} />
+                </Box>
```
allisonking, line 112: can the delete icon go here?

clients/admin-ui/src/features/user-management/UserForm.tsx
```diff
@@ -47,6 +57,7 @@ export interface Props {
 }
 
 const UserForm = ({
+  user,
```
allisonking, line 60: hmm why do we want the `user` here?

clients/admin-ui/src/types/api/models/RolesDataMapping.ts
```diff
@@ -0,0 +1,38 @@
+import { RoleRegistryEnum } from "~/types/api";
+
+export const ROLES = [
```
allisonking, line 3: good call moving this into its own file!

clients/admin-ui/src/types/api/models/ScopeRegistryEnum.ts
```diff
@@ -60,6 +56,7 @@ export enum ScopeRegistryEnum {
   STORAGE_CREATE_OR_UPDATE = "storage:create_or_update",
   STORAGE_DELETE = "storage:delete",
   STORAGE_READ = "storage:read",
+  SYSTEM_READ = "system:read",
```
allisonking, line 59: hmm was this added recently? I remember thinking it wasn't a thing

allisonking (commented):
haven't tested it yet, but just leaving some code comments for now! nice work picking this up 💯

clients/admin-ui/src/features/common/nav/v2/nav-config.test.ts
eastandwestwind: Agree, this is not needed after pulling in latest from Dawn's branch.

clients/admin-ui/src/features/user-management/NewUserForm.tsx
eastandwestwind: nope, this can be removed!

clients/admin-ui/src/features/user-management/AssignSystemsTable.tsx
eastandwestwind: Sure thing, let me see about replacing this with our `ConfirmationModal`

clients/admin-ui/src/features/user-management/NewUserForm.tsx
eastandwestwind: Good call, I'll update to redirect to user profile, and add back the toasts for successful user creation

clients/admin-ui/src/features/user-management/UserForm.tsx
eastandwestwind: we need `user` in `PasswordManagement` -> `DeleteUserModal`. If we move the `DeleteUserModal` out of `PasswordManagement`, we'll still need a `user` passed in here, I believe

clients/admin-ui/src/types/api/models/ScopeRegistryEnum.ts
eastandwestwind: It's added in Dawn's BE work here- https://github.com/ethyca/fides/pull/2714/files#diff-64b8825251ee956a60b62e71f974e359f0b540694f46dd581d2b524c2a94617bR184

clients/admin-ui/src/features/user-management/PasswordManagement.tsx
eastandwestwind: Absolutely. This makes a lot more sense to me, thanks!

pattisdr:
There are these role descriptions hanging out on the right with Owner text duplicated in each of them - 

Screenshot 2023-03-14 at 10 20 08 AM

pattisdr:
I'm assuming this is pre-existing behavior, but the backend only allows the first name and last name of a user to be edited here after it's created, not the username.  Should we disable the username from being edited here? It gives the appearance you can edit but you can't.  This could be reticketed.


Screenshot 2023-03-14 at 11 04 45 AM

eastandwestwind:
@pattisdr I went ahead and updated the username form field to be disabled after the user is created
Screenshot 2023-03-14 at 5 18 45 PM

pattisdr:
Bravo! You're addressing changes faster than I'm finding them.

pattisdr:
I do think this should be reticketed: Saving changes on ctl datasets when you don't have permissions shows a green popup that you "Success: Successfully modified collection" although the API request returned a 403 and no data changed.


Screenshot 2023-03-14 at 11 22 46 AM

eastandwestwind:
@pattisdr ticketed here- https://github.com/ethyca/fides/issues/2827

pattisdr:
This is looking really good @allisonking and @eastandwestwind.

I've reticketed separate backend work here https://github.com/ethyca/fidesplus/issues/683

allisonking:
Great work, @eastandwestwind ! I went in and QA'd and fixed up a few things:

1. It said "Update Password" and "Reset password" which bothered me so I changed it to "Update password" to keep our casing consistent 😎 Also added a success toast on Update password since that wasn't there before (totally out of the scope of this work, but it was easy)
2. Refactored the delete modal logic a bit so that we don't have to pass the `user` object all the way down. I recently made a change (maybe it was in this PR...) where we store the `activeUser` (i.e. the user being edited) in redux, so we can take advantage of that to avoid some prop drilling. Also, before, after you deleted a user, you'd still be left on that user's page (which shouldn't exist anymore) so I added a redirect
3. There was a sneaky bug where if you went to a user who had a role, i.e. "owner", and then went to another user who had no roles, that user would show up as being an "owner" in the role selection page. This was probably my fault from earlier, but should be fixed now

eastandwestwind:
Just added a slight functionality change to User Management UI for better experience.

Before, when deleting a user on their profile page, we stayed on that page, which is misleading because you can no longer edit the user if it doesn't exist.

I'm updating such that we reroute to user management table upon user delete

eastandwestwind:
Side note: I've noticed some flakiness with the `Can edit a system` ->  `Can go through the edit flow` cypress test in `systems.cy.ts`. Sometimes cypress is able to find `cy.getByTestId("continue-btn").click();`, and sometimes this element doesn't exist, which causes the test to fail.

We'll soon be able to remove this line of code anyway, as it's a known bug, documented here- https://github.com/ethyca/fides/issues/2788

For now, though, re-running the test suite seems to work.

eastandwestwind:
Reverting to Allison's last commit, since she already covered the reroute to user management page upon user delete!

PR #2741, opened 2023-03-02
Make Scopes Easier to Use in the UI

### LINKED ISSUE #2716: [Backend] Make Scopes/Roles easier to use in the UI

I believe we need to return scopes as enum types to make it easier for the UI to stay up to date.  Also, it would be useful to have a list of all scopes the user has via any avenue.

- Add `total_scopes` key to `get_user_permissions`.  This should return a combination of scopes the user has been assigned directly along with any scopes they have via their role(s).  There shouldn't be duplicates here. 
- get_user_permissions scopes and total_scopes type should be an enum. Context here: https://github.com/ethyca/fides/pull/2671#discussion_r1118965820

### Additional context

pattisdr:
Closes #2716 

### Code Changes

- Dynamically create a ScopeRegistryEnum from the list of scopes
    - Make this the "type" when getting user permission scopes
- Delete custom validation for scopes now that we can use the enum
- If scope registry changes and user has a scope that no longer exists, filter that out instead of failing on access.
- Add a "total" scopes property that sums the scopes the user has directly along with scopes they have via roles.
- Add logic so that if we remove scopes from the scope registry, the enum doesn't break when trying to load the current user's scopes

### Steps to Confirm

* [ ] Login in as a root user
* [ ] Create another user with limited scopes
* [ ] Verify that UI loads properly
* [ ] Login as the created user
* [ ] Verify UI loads properly
* [ ] Navigate to the get user permissions endpoint and note the "total scopes" key
* [ ] Navigate to http://localhost:8080/openapi.json and note the ScopeRegistryEnum

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

Return a scopes enum so the UI can use this, as well as return "total scopes" - scopes that the user has via their roles or that they were assigned directly.


src/fides/api/ops/api/v1/scope_registry.py
```diff
@@ -228,3 +229,8 @@
 }
 
 SCOPE_REGISTRY = list(SCOPE_DOCS.keys())
+
+
+ScopeRegistryEnum = Enum(  # type: ignore[misc]
+    "ScopeRegistryEnum", {scope: scope for scope in SCOPE_REGISTRY}  # type: ignore
```
allisonking, line 235: super nit, but the `RoleRegistry` enum is just called `RoleRegistry` while this one is `ScopeRegistryEnum`. might be good to make them consistent (at least for me consuming these, would make things a bit clearer 😄 )

```suggestion
    "ScopeRegistry", {scope: scope for scope in SCOPE_REGISTRY}  # type: ignore
```

src/fides/api/ops/schemas/user_permission.py
```diff
@@ -16,21 +14,9 @@ class UserPermissionsCreate(BaseSchema):
     but we also will continue to support the ability to be assigned specific individual scopes.
     """
 
-    scopes: List[str] = []
+    scopes: List[ScopeRegistryEnum] = []
```
allisonking, line 17: ah this will also help me a lot 😸

allisonking (approved):
💯

src/fides/api/ops/api/v1/scope_registry.py
pattisdr: good call!

PR #2742, opened 2023-03-03
Mailchimp Transactional Email Messaging Service

### LINKED ISSUE #2737: Mailchimp Transactional `MessagingConfig` support
We should support `Mailchimp Transactional` as a messaging config in Fides, so that users with Mailchimp Transactional accounts can configure Fides to support transactional emails (identity verification, notifications, etc).

seanpreston:
Closes #2737 

### Code Changes

### Steps to Confirm

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

This PR adds bindings for the Mailchimp Transactional mail messaging service.

### DIFF
--- a/src/fides/api/ops/schemas/messaging/messaging.py
+++ b/src/fides/api/ops/schemas/messaging/messaging.py
@@ -181,6 +181,18 @@ class MessagingServiceDetails(Enum):
     TWILIO_EMAIL_FROM = "twilio_email_from"
 
 
+class MessagingServiceDetailsMailchimpTransactional(BaseModel):
+    """The details required to represent a Mailchimp Transactional email configuration."""
+
+    domain: str
+    email_from: str
+
+    class Config:
+        """Restrict adding other fields through this schema."""
+
+        extra = Extra.forbid
+
+
 class MessagingServiceDetailsMailgun(BaseModel):
     """The details required to represent a Mailgun email configuration."""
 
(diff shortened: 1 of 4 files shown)


src/fides/api/ops/schemas/messaging/messaging.py
```diff
@@ -167,15 +170,30 @@ class EmailForActionType(BaseModel):
...
+    DOMAIN = "domain"
+    EMAIL_FROM = "email_from"
+
     # Mailgun
     IS_EU_DOMAIN = "is_eu_domain"
     API_VERSION = "api_version"
-    DOMAIN = "domain"
 
     # Twilio Email
     TWILIO_EMAIL_FROM = "twilio_email_from"
```
seanpreston, line 182: This field should be deprecated in favour of generic `EMAIL_FROM`

adamsachs (commented):
generally all looks good to me, i tested locally that i was able to configure successfully, though i did not execute an end to end test sending emails, just because i don't have credentials.

a couple minor nits to address log message typos. then there's code cov, not sure if we want to try and address those - i'm OK with bypassing them since we're trying to squeeze this into the release. i do think we should update the CHANGELOG though since this is a substantial addition that should be noted.

besides that, seems about good to go IMO!

src/fides/api/ops/schemas/messaging/messaging.py
```diff
@@ -216,8 +236,19 @@ class MessagingServiceSecrets(Enum):
     TWILIO_API_KEY = "twilio_api_key"
 
 
+class MessagingServiceSecretsMailchimpTransactional(BaseModel):
```
adamsachs, line 239: missed this on the review - and it's not _strictly_ necessary it seems - but we should probably add this class to the `SUPPORTED_MESSAGING_SERVICE_SECRETS` constant on L370 that's used for type hints

PR #2775, opened 2023-03-08
Include custom columns in header row of datamap

### LINKED ISSUE #2768: Custom Fields Columns Not Populating on Datamap Route
### Bug Description

When exporting my datamap via the UI, custom fields are not visible.

This appears to be due to the column missing from the API response, where columns are listed as the first indexed item.

#### Steps to Reproduce

1. `nox -s dev`
1. `fides user login ...`
2. `fides push demo_resources/`
3. Add some custom field data via the database
4. `GET` datamap endpoint and see the results
5. Verify the field is returned but no column information is

### Expected behavior

Both the column and field information should be returned

### Screenshots

n/a

### Environment

* Version: 2.8.0
* OS: Darwin
* Python Version: 3.9
* Docker Version:

### Additional context

Discovered while doing extra testing with `fidesplus`

SteveDMurphy:
Closes #2768

### Code Changes

* [x] Ensure custom columns are included with the header output of `/datamap`
* [x] Resolve duplication of custom fields across all rows regardless of `fides_key`

### Steps to Confirm

* [x] Bounce this branch against `fidesplus` and test the datamap UI has the custom fields

Manual detail steps:
* [ ] Using `fides` run `nox -s "fides_env(test)"`
* [ ] Using `fidesplus` run the datamap UI locally using `npm run dev`
* [ ] Ensure the datamap looks good normally without custom fields http://localhost:4000/datamap
* [ ] Execute the following sql commands against the database to populate custom fields (I used dbeaver)
 ```
insert into public.plus_custom_field_definition (id, name, description, field_type, resource_type, field_definition, active)
select 'new_id', 'country', 'country name', 'string', 'system', 'string', true;

insert into public.plus_custom_field (id, resource_type, resource_id, custom_field_definition_id, value)
select 'updated_id', 'system', 'demo_analytics_system', 'new_id', '{portugal}';
```
* [ ] Ensure the datamap looks good with custom fields populated for `demo_analytics_system` http://localhost:4000/datamap

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### Description Of Changes

Found as part of the release for `2.8.0`

An additional finding around the application of the custom fields was also found and included in here

### DIFF
--- a/src/fides/api/ctl/routes/datamap.py
+++ b/src/fides/api/ctl/routes/datamap.py
@@ -163,8 +163,7 @@ async def export_datamap(
 
     formatted_datamap = format_datamap_values(joined_system_dataset_df, custom_columns)
 
-    # prepend column names
-    formatted_datamap = [DATAMAP_COLUMNS_API] + formatted_datamap
+    formatted_datamap = [{**DATAMAP_COLUMNS_API, **custom_columns}] + formatted_datamap
     return formatted_datamap
 
 
--- a/src/fides/core/export.py
+++ b/src/fides/core/export.py
@@ -194,7 +194,6 @@ def generate_system_records(  # pylint: disable=too-many-nested-blocks, too-many
             keys.append(key_string)
             output_list[0] = tuple(keys)
 
-    system_custom_field_data = {}
     known_fields = (
         "fides_key",
         "name",
@@ -219,6 +218,7 @@ def generate_system_records(  # pylint: disable=too-many-nested-blocks, too-many
         "system_type",
     )
     for system in server_resources["system"]:
+        system_custom_field_data = {}
         if not isinstance(system, dict):
             system = system.__dict__
 
@@ -229,12 +229,10 @@ def generate_system_records(  # pylint: disable=too-many-nested-blocks, too-many
                 keys.append(key_string)
                 output_list[0] = tuple(keys)
                 custom_columns[key_string] = key
-                child_custom_field_data = {}
                 if isinstance(value, list):
-                    child_custom_field_data[key_string] = ", ".join(value)
+                    system_custom_field_data[key_string] = ", ".join(value)
                 else:
-                    child_custom_field_data[key_string] = value
-                system_custom_field_data[system["fides_key"]] = child_custom_field_data
+                    system_custom_field_data[key_string] = value
 
         third_country_list = ", ".join(system.get("third_country_transfers") or [])
         system_dependencies = ", ".join(system.get("system_dependencies") or [])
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by SteveDMurphy** [comment]
Good news! The country is now populating in the datamap UI 🙌🏽 

image

Bad news, the records are not being applied correctly just yet. This could still be user/testing errors that will take me a bit longer to debug

**Comment by SteveDMurphy** [comment]
resolved the issue where the data was overpopulating, now just need to clean up and put into review. We should be able to move ahead with a patch release for this as early as tomorrow

**Comment by SteveDMurphy** [comment]
Updated screenshot with custom field only applied where it should be: 
image

File: src/fides/api/ctl/routes/datamap.py [hunk 1]
```diff
@@ -163,8 +163,7 @@ async def export_datamap(
 
     formatted_datamap = format_datamap_values(joined_system_dataset_df, custom_columns)
 
-    # prepend column names
-    formatted_datamap = [DATAMAP_COLUMNS_API] + formatted_datamap
+    formatted_datamap = [{**DATAMAP_COLUMNS_API, **custom_columns}] + formatted_datamap
```
SteveDMurphy, line 166: The main reason they were not populating is the custom field was not populated into the first item in the response list, which the UI uses to render columns. This fixes that by including the custom fields that exist

src/fides/core/export.py
```diff
@@ -229,10 +229,12 @@ def generate_system_records(  # pylint: disable=too-many-nested-blocks, too-many
                 keys.append(key_string)
                 output_list[0] = tuple(keys)
                 custom_columns[key_string] = key
+                child_custom_field_data = {}
                 if isinstance(value, list):
-                    system_custom_field_data[key_string] = ", ".join(value)
+                    child_custom_field_data[key_string] = ", ".join(value)
```
SteveDMurphy, line 234: some poor naming potentially, but I ended up filtering based on the system `fides_key` and adding the custom fields as children of said `fides_key`

src/fides/core/export.py
```diff
@@ -193,7 +193,6 @@ def generate_system_records(  # pylint: disable=too-many-nested-blocks, too-many
             custom_columns[key_string] = value
             keys.append(key_string)
             output_list[0] = tuple(keys)
-
```
adamsachs, line 196: wait @SteveDMurphy i think ya may have meant to remove the line below, no?

SteveDMurphy: Yep! 🤦🏽 blame it on the hour, thanks for catching

adamsachs (approved):
🎊 looks great! thanks for getting this fixed so quick

Roger-Ethyca:
Screenshot 2023-03-14 at 5 01 48 PM

PR #2780, opened 2023-03-08
update appconfig migration to handle existing record

### LINKED ISSUE #2779: DB migration fails from `2.7.x` --> `2.8.x` if `applicationconfig` record is present
### Bug Description

if you've used the `PATCH /api/v1/config` endpoint in `2.7.x`, and then you upgrade to `2.8.x`, your db migration will fail.

**NOTE**:  This only impacts cases where users have hit the _brand new_ `PATCH /api/v1/config` (added in `2.7.0`, https://github.com/ethyca/fides/pull/2551) at some point while running a `2.7.x` fides. that API is _not_ hit by the UI, so a user would need to explicitly invoke the API manually. we don't anticipate that anyone (besides some advanced internal users!) has done this.

#### Steps to Reproduce

1. `git fetch --tags`
2. `git checkout tags/2.7.1`
3. `nox -s dev`
4. go to swagger UI, make a call against the `PATCH /api/v1/config` endpoint
5. `nox -s teardown`
6. `git checkout tags/2.8.1`
7. `nox -s dev`
8. notice your db migration fail and the app will fail to start up

### Expected behavior

app should start up fine _even if_ you happened to hit the `PATCH /api/v1/config` endpoint on `2.7.x`

### Additional context

@RobertKeyser hit this on fides cloud, since he'd used the `PATCH /api/v1/config` as an advanced user when `2.7.x` was running.

Stack trace for the migration failure attached: [migration_failure.txt](https://github.com/ethyca/fides/files/10925050/migration_failure.txt)

adamsachs:
Closes #2779 

### Code Changes

* include a `server_default` of encrypted `{}` value when creating the `config_set` column in migration, so as to handle cases where an existing `appconfig` record exists while allow the new column to be nullable.

### Steps to Confirm

* see issue description for repro steps

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

~~the proposed solution is not ideal - i'm not sure of a truly clean way to handle this. the column in question here is a new column in `2.8.x` and it's an encrypted JSON column that ideally should be non-nullable. before this change, the problem with the migration was that it just added the non-nullable column to the table, but doesn't account for the fact that there may be an existing record in the db, even though that's _quite a narrow edge case_. but in that case, the new non-nullable column is added to the record, but the column is obviously null since it didn't exist, and that's the error we get.~~ 

the "obvious" fix would be to give that column a `server_default` when creating it in the migration, so that it's not null. ~~but i don't think we can do effectively that from the context of the migration, since it's encrypted JSON, and it's not like the migration has access to the app encryption key.~~ UPDATE - i was totally wrong about this, thanks to @pattisdr for showing me that we do indeed have access to the app encryption key in the migration. so this is the path i've decided on now!

it's worth noting that this is handling a pretty narrow edge case - but @RobertKeyser did run into it on a migration of an internally hosted app, so it's worth us fixing!

### DIFF
--- a/src/fides/api/ctl/migrations/versions/c9ee230fa6da_add_config_set_column_to_application_.py
+++ b/src/fides/api/ctl/migrations/versions/c9ee230fa6da_add_config_set_column_to_application_.py
@@ -8,6 +8,16 @@ Create Date: 2023-02-01 15:13:52.133075
 import sqlalchemy as sa
 import sqlalchemy_utils
 from alembic import op
+from sqlalchemy import text
+from sqlalchemy.dialects.postgresql import JSON
+from sqlalchemy.sql.elements import TextClause
+from sqlalchemy_utils.types.encrypted.encrypted_type import (
+    AesGcmEngine,
+    StringEncryptedType,
+)
+
+from fides.api.ops.db.base_class import JSONTypeOverride
+from fides.core.config import CONFIG
 
 # revision identifiers, used by Alembic.
 revision = "c9ee230fa6da"
@@ -17,13 +27,27 @@ depends_on = None
 
 
 def upgrade():
+    encryptor = sqlalchemy_utils.types.encrypted.encrypted_type.StringEncryptedType(
+        JSONTypeOverride,
+        CONFIG.security.app_encryption_key,
+        AesGcmEngine,
+        "pkcs5",
+    )
+    empty_obj = encryptor.process_bind_param({}, JSON)
+
     op.add_column(
         "applicationconfig",
         sa.Column(
             "config_set",
-            sqlalchemy_utils.types.encrypted.encrypted_type.StringEncryptedType(),
+            StringEncryptedType(),
             nullable=False,
+            server_default=empty_obj,
         ),
+    )
+
+    # reset the server_default now that it's initial value has been set
+    op.alter_column(
+        "applicationconfig", "config_set", nullable=False, server_default=None
     )
 
     # include this update here to make up for an earlier miss when creating the table


src/fides/api/ctl/migrations/versions/c9ee230fa6da_add_config_set_column_to_application_.py
```diff
@@ -22,10 +24,28 @@ def upgrade():
...
         ),
     )
 
+    # Data migration - delete the existing app config record, if it exists
+    # since we're making a new encrypted JSON column, we can't reasonably set a default
+    # so we must just delete and start over. up to this point, users could only have
+    # the storage.active_default_storage_type property set in this record. they will have
+    # to reset it after this migration
+    bind = op.get_bind()
+    existing_appconfig: TextClause = bind.execute(
```
pattisdr, line 37: eh, this type looks like it's a ResultProxy instead of a TextClause, but I've made that type mistake on earlier migrations!

src/fides/api/ctl/migrations/versions/c9ee230fa6da_add_config_set_column_to_application_.py
```diff
@@ -22,10 +24,28 @@ def upgrade():
         sa.Column(
             "config_set",
             sqlalchemy_utils.types.encrypted.encrypted_type.StringEncryptedType(),
-            nullable=False,
+            nullable=True,  # keep it nullable to start, in case we have any existing records
```
pattisdr, line 27: Alternatively, if you want to keep an existing applicationconfig record instead of deleting, you can provide a server default, and then get rid of that default later.  so here, something like:

```python
from sqlalchemy.dialects.postgresql import JSON
from fides.core.config import CONFIG

encryptor = sqlalchemy_utils.types.encrypted.encrypted_type.StringEncryptedType(
    JSONTypeOverride,
    CONFIG.security.app_encryption_key,
    AesGcmEngine,
    "pkcs5",
)
val = encryptor.process_bind_param({}, JSON)

op.add_column(
    "applicationconfig",
    sa.Column(
        "config_set",
        sqlalchemy_utils.types.encrypted.encrypted_type.StringEncryptedType(),
        nullable=False,
        server_default=val,
    ),
)

...and then at the end....
op.alter_column(
   "applicationconfig", "config_set", nullable=False, server_default=None
)

```

pattisdr (commented):
this looks fine! I've provided some other optional paths, but this is good too.

src/fides/api/ctl/migrations/versions/c9ee230fa6da_add_config_set_column_to_application_.py
adamsachs: oh wait, my whole reasoning for going the way i did (deletion) was because i'd assumed we don't have access to our fides `CONFIG` in the migration context. but maybe that assumption is wrong? i should have at least tried to verify it. i'll give it a try now - thanks for suggesting this!

adamsachs: it looks like this does indeed work! thank you again for the pointer here @pattisdr . my new commit uses this approach - i don't even think we need to get rid of the default later, it will be overwritten naturally on server bootup, which is fine.

src/fides/api/ctl/migrations/versions/c9ee230fa6da_add_config_set_column_to_application_.py
adamsachs: good catch - will no longer be relevant in my new commit though, so resolving :)

src/fides/api/ctl/migrations/versions/c9ee230fa6da_add_config_set_column_to_application_.py
```diff
@@ -22,10 +24,28 @@ def upgrade():
...
+    # to reset it after this migration
+    bind = op.get_bind()
+    existing_appconfig: TextClause = bind.execute(
+        text("SELECT id FROM applicationconfig;")
+    )
+    for row in existing_appconfig:
+        update_app_config_query: TextClause = text(
+            "DELETE from applicationconfig WHERE id= :appconfig_id"
+        )
+        bind.execute(update_app_config_query, {"appconfig_id": row["id"]})
```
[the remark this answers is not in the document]
adamsachs: good call - but should no longer be needed at all :)

src/fides/api/ctl/migrations/versions/c9ee230fa6da_add_config_set_column_to_application_.py
pattisdr: By the default, I'm not talking about the row contents, I mean this will still be value postgres considers its server default. I'd personally revert the server default right after since this was specifically for migration purposes only. 

Benefits of reverting back to null means if somewhere in the application tries to set this to applicationconfig_set to nullin the future, we'll be prevented from doing that at the database level, instead of it being set to an encrypted empty dict.

Current state here:
```
                                               Table "public.applicationconfig"
   Column   |           Type           | Collation | Nullable |                            Default                            
------------+--------------------------+-----------+----------+---------------------------------------------------------------
 id         | character varying(255)   |           | not null | 
 created_at | timestamp with time zone |           |          | now()
 updated_at | timestamp with time zone |           |          | now()
 api_set    | character varying        |           | not null | 
 single_row | boolean                  |           | not null | 
 config_set | character varying        |           | not null | 'QrPxq7t+8fWmVwJ/47xkFnYEhNOEuQyIlVPLJOou'::character varying

```

adamsachs: got it - i didn't totally grasp what a `server_default` was, so this makes sense -- thank you! i'll update accordingly.

adamsachs: updated in latest commit - let me know if it looks good to you @pattisdr! did some manual validation locally too.

pattisdr (approved):

PR #2817, opened 2023-03-13
Update miscellaneous dependencies for clients/admin-ui and clients/privacy-center

NevilleS:
### Description Of Changes

This bundles all 10 of the currently open @dependabot updates for the clients/admin-ui and clients/privacy-center together into a single PR for sanity: all of these should be pretty innocent changes, but merging all 10 of those individually would tie up our GitHub actions runner forever...!


NevilleS (commented):
LGTM! All green, and the only notable update is bumping Cypress to 12.x on the Privacy Center... but the Cypress suite passed there so we're good 👍

PR #2825, opened 2023-03-14
support >1 custom fields per system in datamap export

adamsachs:
Closes https://github.com/ethyca/fidesplus/issues/679. 

### Code Changes

* for systems with privacy declarations and with custom field data, our iteration to add the custom field data looked like this:
    * for each custom field, iterate over each product (row of output data); and then within each product (row), add the custom field value to the row, and add the row to the list of output rows (the `cartesian_product_of_declaration` itself gets appended to the `output_list`):

```
if system_custom_field_data:
  for _, v in system_custom_field_data.items():
    for product in cartesian_product_of_declaration_builder:
      product.append(v)
      cartesian_product_of_declaration.append(tuple(product))
```

* this had the effect of adding a new _row_ of output data for each custom field - which is not what we want!
* instead, what we do now is: for each product (row), iterate over each custom field; within each custom field, add it to the row; then, add the updated row to the ultimate list of output rows:

```
for product in cartesian_product_of_declaration_builder:
  if system_custom_field_data:
    for _, v in system_custom_field_data.items():
      product.append(v)
  cartesian_product_of_declaration.append(tuple(product))
```

* separately: added in a bunch of "brute force" end-to-end automated tests around datamap export

### Steps to Confirm

* it's a bit tricky to manually test this on just `fides` since we rely on both the datamap UI and the custom field UI that's available only in `fidesplus` - (even more of a reason to add the automated tests!)
    * it's worth noting, the automated tests that were added were failing before this fix was in place - specifically the `test_datamap_privacy_declarations_two_custom_fields` test case
* the steps in https://github.com/ethyca/fides/pull/2775 can be used as a reference and adjusted accordingly

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### Description Of Changes

The code change for this particular fix is small - just switching around a nested iteration that had the wrong nesting order.

More substantially, I decided to add a fair amount of "end-to-end" test coverage around the `export_datamap` endpoint and its handling of custom metadata. Previously the automated tests were not doing much besides confirming we got the appropriate response code - in fact, the response bodies in the tests weren't actually populated with any data, they just had the "column header" row. Within the 2.8.x releases/hotfixes, we've experiencing a bit of 'whack-a-mole' with the datamap generation code - as we've fixed one problem, other problems have arisen. Even if the "new" problems haven't been directly caused by the fixes to the "old" problems, trying to manually cover all test cases with each new fix has proven difficult and time consuming. Additionally, without test cases covering the export code and custom metadata, it was difficult to get a debugger onto some of the problem states, meaning we'd need to rely on static analysis of some rather complex data manipulation code. In short - automated tests are a pretty critical first step for trying to make this more robust!

The test coverage that was added is a _brute force_ approach to all (or most) of the different combinations of features/scenarios that can arise with the datamap export and custom metadata as we have it now. This probably isn't a good final state for this test coverage, but I decided to go the brute force approach for now, because:
- having _any_ automated_ test coverage of the custom metadata export, even if it's a bit unwieldy, is a lot better than not having any! the end-to-end "integration" tests here allow us to feel confident that we've got the functionality we need in place
- we need a fix for the >1 custom fields problem quickly, as it's impacting customers
- we should likely refactor some of the datamap export code anyway since the use now for the export has drifted from its original use as a CLI command; at that point, we can improve the test coverage
    - i'll create a tech debt follow-up ticket for this item so we don't lose track of it

-- UPDATE -- 

I'm noticing that we did already have _some_ test coverage around custom field and datamap export in https://github.com/ethyca/fides/blob/main/tests/ctl/core/test_export.py - but it seems minimal. going forward, we could look at updating that to provide more comprehensive _unit_ test coverage that's a bit more maintainable, rather than relying only on the brute force, end-to-end integration tests i've got in this PR. that being said, i still think the integration tests here are helpful in ensuring our end-to-end functionality is working well across a variety of use cases.

### DIFF
--- a/src/fides/core/export.py
+++ b/src/fides/core/export.py
@@ -317,15 +317,11 @@ def generate_system_records(  # pylint: disable=too-many-nested-blocks, too-many
                     for dataset_reference in dataset_references
                 ]
                 cartesian_product_of_declaration = []
-                if system_custom_field_data:
-                    for _, v in system_custom_field_data.items():
-                        for product in cartesian_product_of_declaration_builder:
+                for product in cartesian_product_of_declaration_builder:
+                    if system_custom_field_data:
+                        for _, v in system_custom_field_data.items():
                             product.append(v)
-                            cartesian_product_of_declaration.append(tuple(product))
-                else:
-                    cartesian_product_of_declaration = [
-                        tuple(x) for x in cartesian_product_of_declaration_builder
-                    ]
+                    cartesian_product_of_declaration.append(tuple(product))
 
                 output_list += cartesian_product_of_declaration
         else:
--- a/tests/ctl/api/test_datamap.py
+++ b/tests/ctl/api/test_datamap.py
@@ -1,53 +1,509 @@
 # pylint: disable=missing-docstring, redefined-outer-name
+from typing import Any
+
 import pytest
+from fideslang import models
 from starlette.testclient import TestClient
 
 from fides.api.ctl.routes.util import API_PREFIX
 from fides.api.ctl.sql_models import (  # type: ignore[attr-defined]
     CustomField,
     CustomFieldDefinition,
+    System,
 )
+from fides.api.ops.util.data_category import DataCategory
 from fides.core.config import FidesConfig
+
+HEADERS_ROW_RESPONSE_PAYLOAD = {
+    "dataset.name": "Fides Dataset",
+    "system.name": "Fides System",
+    "system.administrating_department": "Department or Business Function",
+    "system.privacy_declaration.data_use.name": "Purpose of Processing",
+    "system.joint_controller": "Joint Controller",
+    "system.privacy_declaration.data_subjects.name": "Categories of Individuals",
+    "unioned_data_categories": "Categories of Personal Data (Fides Taxonomy)",
+    "system.privacy_declaration.data_use.recipients": "Categories of Recipients",
+    "system.link_to_processor_contract": "Link to Contract with Processor",
+    "third_country_combined": "Third Country Transfers",
+    "system.third_country_safeguards": "Safeguards for Exceptional Transfers of Personal Data",
+    "dataset.retention": "Retention Schedule",
+    "organization.link_to_security_policy": "General Description of Security Measures",
+    "system.data_responsibility_title": "Role or Responsibility",
+    "system.privacy_declaration.data_use.legal_basis": "Article 6 lawful basis for processing personal data",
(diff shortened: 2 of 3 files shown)


src/fides/core/export.py
```diff
@@ -317,15 +317,11 @@ def generate_system_records(  # pylint: disable=too-many-nested-blocks, too-many
...
+                for product in cartesian_product_of_declaration_builder:
+                    if system_custom_field_data:
+                        for _, v in system_custom_field_data.items():
                             product.append(v)
-                            cartesian_product_of_declaration.append(tuple(product))
-                else:
-                    cartesian_product_of_declaration = [
-                        tuple(x) for x in cartesian_product_of_declaration_builder
-                    ]
+                    cartesian_product_of_declaration.append(tuple(product))
```
NevilleS, line 324: So the real _change_ here seems to be basically inverting the order of this, i.e.
* Before: for each custom field and each product....
* After: for each product and each custom field...

Can you provide some explanation as to why this fixes the issue?

adamsachs: right, but with one more key detail: we move the `cartesian_product_of_declaration.append(tuple(product))` operation into the level of _for each product_  - which is only really possible if we invert the nesting order as you mention.

i tried to summarize this in the PR description with this point:
> for systems with privacy declarations and with custom field data, add in all of the custom field data per output row, rather than adding new output rows for each custom field

but maybe that's keeping it too high-level? i can add in some more detail to hopefully make that clearer

adamsachs: @NevilleS i've updated the PR description to explain the fix a bit more thoroughly. let me know what you think - i hope i haven't made things more confusing!

tests/ctl/api/test_datamap.py
```diff
@@ -1,79 +1,882 @@
...
 from fides.api.ctl.routes.util import API_PREFIX
 from fides.api.ctl.sql_models import (  # type: ignore[attr-defined]
     CustomField,
     CustomFieldDefinition,
+    System,
 )
+from fides.api.ops.util.data_category import DataCategory
 from fides.core.config import FidesConfig
 
+HEADERS_ROW_RESPONSE_PAYLOAD = {
```
seanpreston, line 17: Thankyou for taking the time to add these!

seanpreston:
Thanks @adamsachs, great work!

Screenshot 2023-03-14 at 16 20 12

seanpreston (approved):

Issue #2788, opened 2023-03-10
After saving edits to a system in the "View systems" page, warning modal about unsaved changes still appears


NevilleS:
### Bug Description

When editing a system via the UI, the "Unsaved changes" modal will popup even after saving, which is unexpected. However, when you reload / refresh, the changes are definitely saved, so this modal is just triggering incorrectly.

#### Steps to Reproduce

1. Run `nox -s "fides_env(test)"
2. Navigate to "Data map" > "View systems", loads the default "System information" tab
3. Edit any system
4. Make a change to the system, hit "Save"
5. Navigate to the "Data use" tab
6. See an unexpected popup warning:
![image](https://user-images.githubusercontent.com/1834295/224346027-919e5d4d-534d-46fb-a097-70a8c9e2d041.png)

### Expected behavior

Since I've already saved my changes, the form shouldn't be considered "dirty" and the warning isn't needed.

### Environment

* Version: 2.8.1
* OS: Mac OSX
* Python Version: 3.9.6
* Docker Version: Docker version 20.10.22, build 3a2c30b

### Additional Context

Please update the `systems.cy.ts` Cypress tests to remove the unnecessary `cy.getByTestId("continue-btn").click()` call to dismiss this popup once it's fixed 👍

rsilvery:
Further testing shows it's not related to whether update are made. Any time you click on the `data use` tab this pop-up appears: https://www.loom.com/share/5ecaf5ee58114e0abf74eb7570f0064e

Kelsey-Ethyca:
QA'd and moving to done 🌵

PR #2826, opened 2023-03-14
Custom metadata refactoring so that form is not always dirty

allisonking:
Closes https://github.com/ethyca/fides/issues/2788

The issue presented itself like something was wrong with the `UnmountWarning` component, but it's actually because the form was always in a "dirty" state due to some circular logic going on in the `CustomFieldsList` component (hence the problem only showing up in plus)

Originally, `CustomFieldsList` had a `useEffect` which would read in the data from the custom fields endpoint, then populate the form on the fly via `setValue`. Using `setValue` made the form always dirty, hence the unmount warning always showing up.

I went for a pretty thorough (🤞) refactoring, where instead of populating the data on the fly, populate the data via the initial values to begin with. this cleans up a lot of the `useEffect`s and various `reload` functionality, since formik is good at handling that kind of thing. 

### Code Changes

* Refactored the `useEffect`s out of `CustomFieldsList` in favor of setting initial values 
  *  Renamed the field to `customFieldValues` since the format is a bit more streamlined for formik than the original `definitionIdToCustomFieldValue` which is still used, but just as a Map
* Updated the system and taxonomy hooks to do the initial value transformation
* Remove the original cypress workaround which was accommodating for the unmount warning popping up when it wasn't supposed to
* Added a basic test for system custom metadata since there weren't any. Luckily there were taxonomy tests already in place 😌 

### Steps to Confirm

* [ ] Run the fidesplus backend (`nox -s "dev(fidesplus)"`)
* [ ] Edit an existing system (can make one manually if you don't already have one)
* [ ] Make sure if you don't make any changes, you can still go to the "Data use" tab and the unmount warning doesn't interfere
* [ ] Make changes to custom metadata. Hopefully everything still works the same.
  * [ ] ❗ This was a pretty non-trivial refactor, so could definitely use some QA to make sure I didn't break anything since I'm not super familiar with the custom metadata flow

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### Description Of Changes

https://user-images.githubusercontent.com/24641006/225063566-2108c110-36ca-43d1-a4bb-10551f17974a.mov

### DIFF
--- a/clients/admin-ui/cypress/e2e/systems-plus.cy.ts
+++ b/clients/admin-ui/cypress/e2e/systems-plus.cy.ts
@@ -1,0 +1,69 @@
+import {
+  stubPlus,
+  stubSystemCrud,
+  stubTaxonomyEntities,
+} from "cypress/support/stubs";
+
+describe("System management with Plus features", () => {
+  beforeEach(() => {
+    cy.login();
+    stubSystemCrud();
+    stubTaxonomyEntities();
+    stubPlus(true);
+    cy.intercept("GET", "/api/v1/system", {
+      fixture: "systems/systems.json",
+    }).as("getSystems");
+    cy.visit("/system");
+  });
+
+  describe("custom metadata", () => {
+    beforeEach(() => {
+      cy.intercept(
+        {
+          method: "GET",
+          pathname: "/api/v1/plus/custom-metadata/allow-list",
+          query: {
+            show_values: "true",
+          },
+        },
+        {
+          fixture: "taxonomy/custom-metadata/allow-list/list.json",
+        }
+      ).as("getAllowLists");
+      cy.intercept(
+        "GET",
+        `/api/v1/plus/custom-metadata/custom-field-definition/resource-type/*`,
+
+        {
+          fixture: "taxonomy/custom-metadata/custom-field-definition/list.json",
+        }
+      ).as("getCustomFieldDefinitions");
+      cy.intercept(
+        "GET",
+        `/api/v1/plus/custom-metadata/custom-field/resource/*`,
+        {
+          fixture: "taxonomy/custom-metadata/custom-field/list.json",
+        }
+      ).as("getCustomFields");
+    });
+
+    it("can populate initial custom metadata", () => {
+      cy.getByTestId("system-fidesctl_system").within(() => {
+        cy.getByTestId("more-btn").click();
+        cy.getByTestId("edit-btn").click();
+      });
+
+      // Should not be able to save while form is untouched
+      cy.getByTestId("save-btn").should("be.disabled");
+      const testId =
+        "input-customFieldValues.id-custom-field-definition-pokemon-party";
+      cy.getByTestId(testId).contains("Charmander");
+      cy.getByTestId(testId).contains("Eevee");
+      cy.getByTestId(testId).contains("Snorlax");
+      cy.getByTestId(testId).type("Bulbasaur{enter}");
+
+      // Should be able to save now that form is dirty
+      cy.getByTestId("save-btn").should("be.enabled");
+    });
+  });
+});
--- a/clients/admin-ui/cypress/e2e/systems-plus.cy.ts
+++ b/clients/admin-ui/cypress/e2e/systems-plus.cy.ts
@@ -45,6 +45,9 @@ describe("System management with Plus features", () => {
           fixture: "taxonomy/custom-metadata/custom-field/list.json",
         }
       ).as("getCustomFields");
+      cy.intercept("PUT", `/api/v1/plus/custom-metadata/custom-field`, {
+        fixture: "taxonomy/custom-metadata/custom-field/update-party.json",
+      }).as("updateParty");
     });
 
     it("can populate initial custom metadata", () => {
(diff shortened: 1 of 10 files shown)


clients/admin-ui/src/features/taxonomy/hooks.tsx
```diff
@@ -164,7 +174,7 @@ export const useDataCategory = (): TaxonomyHookData<DataCategory> => {
     handleEdit,
     handleDelete,
     renderExtraFormFields,
-    transformEntityToInitialValues: transformTaxonomyBaseToInitialValues,
+    transformEntityToInitialValues: handleTransformEntityToInitialValues,
```
NevilleS, line 177: nit: this'd be less of a potential gotcha if you didn't rename it to `transformEntityToInitialValues` and kept it as `handleTransformEntityToInitialValues`.

Right now that returned function is dangerously close to another locally scoped function, `transformTaxonomyBaseToInitialValues` and I think a mis-type could happen somewhere in the future

clients/admin-ui/cypress/e2e/systems-plus.cy.ts
```diff
@@ -0,0 +1,69 @@
...
+      cy.getByTestId("save-btn").should("be.disabled");
+      const testId =
+        "input-customFieldValues.id-custom-field-definition-pokemon-party";
+      cy.getByTestId(testId).contains("Charmander");
+      cy.getByTestId(testId).contains("Eevee");
+      cy.getByTestId(testId).contains("Snorlax");
+      cy.getByTestId(testId).type("Bulbasaur{enter}");
+
+      // Should be able to save now that form is dirty
+      cy.getByTestId("save-btn").should("be.enabled");
```
NevilleS, line 66: I'd recommend going a bit further here and clicking save in the test and asserting that the right API calls get made. There's a _lot_ of magic going on in `useCustomFields` that'd benefit from some coverage to keep an eye on it

NevilleS (approved):
Couple comments from me, but I don't immediately see any issues. I'd get a second opinion on whether more/less hooks are needed for that `useCustomFields` logic...

Nice work refactoring this to streamline those hooks and rely on Formik more to handle the form inputs.

adamsachs:
confirmed with manual steps that things look good functionally in the UI, nice work cleaning this up! 👍

clients/admin-ui/src/features/common/custom-fields/hooks.ts
```diff
@@ -93,23 +93,35 @@ export const useCustomFields = ({
...
+  /**
+   * Transformed version of definitionIdToCustomField to be easy
+   * to pass into Formik
+   */
+  const customFieldValues: Record<string, string | string[]> = {};
+  if (definitionIdToCustomField) {
+    definitionIdToCustomField.forEach((value, key) => {
+      customFieldValues[key] = value.value.toString();
+    });
+  }
```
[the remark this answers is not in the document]
TheAndrewJackson: It's possible that it's not required for things to work properly. Sometimes due to the way the tree is structured it'll cause hook functions like this one to run every time a value like `customFieldValues` needs to re-initialized. `useMemo` and `useEffect` are helpful when you want control over when some code does or doesn't run. Since it's probably cheap to re-initialize `customFieldValues` a `useMemo` may not be needed. It wouldn't hurt to put it in one though.

Does that seem correct in this situation @allisonking?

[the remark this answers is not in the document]
allisonking: yeah that makes sense—I'll wrap it in a `useMemo` 👍

clients/admin-ui/src/features/system/form.ts
```diff
@@ -10,6 +10,7 @@ export interface FormValues
     progress?: DataProtectionImpactAssessment["progress"];
     link?: DataProtectionImpactAssessment["link"];
   };
+  customFieldValues?: Record<string, string | string[]>;
```
TheAndrewJackson, line 13: nit. Does it makes sense to make a type that something like

```typescript
type CustomFieldValues = Record<string, string | string[]>;
```

Then it could be reused in the few spots where this type signature is used.

TheAndrewJackson (approved):
Great work! I like the refactor to use the initial form values instead.

Just one small nit. Otherwise it looks good to go

clients/admin-ui/src/features/common/custom-fields/hooks.ts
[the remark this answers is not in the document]
allisonking: whoops sorry I didn't see @TheAndrewJackson 's comment before I commented 😅 

I do think your analysis is correct—`useMemo` in general is for expensive operations since there is a slight cost to using `useMemo` in the first place. as for what counts as "expensive", I'm not up to date on. but I don't think it hurts to put the `useMemo` in now, and if we're really coming down to perf issues, then to look into this hook more

clients/admin-ui/src/features/system/form.ts
allisonking: hehe yeah the thought crossed my mind but I got lazy 😅 will do!

clients/admin-ui/src/features/taxonomy/hooks.tsx
allisonking: the consumers of these hooks all expect `transformEntityToInitialValues` (the whole taxonomy page attempts to be very general by putting all the details in the hooks so that it can render data uses, data categories, etc. via the same interface) so I'm hesitant to change it. 

I'm not sure if this is a real pattern, or something that I just picked up somewhere, of the actual `prop` being named some action, and then whatever wrapper the component needs to be named `handle[some action]`. for instance, if a component needs an `onClose`, then for the parent component to call the function locally `handleClose` and then to pass in `onClose={handleClose}`. that's what I was trying to do here, but I understand the confusion. I'll see if there's some way to make it clearer

PR #2831, opened 2023-03-14
Add privacy declaration name to datamap response

TheAndrewJackson:
### Code Changes

* [ ] Add an api flag to conditionally add the privacy declaration name column to the datamap response

### Steps to Confirm

* [ ] call `http://0.0.0.0:8080/api/v1/datamap/default_organization?include_deprecated_columns=true` and make sure the `system.privacy_declaration.name` is in the response. You may need to create some declarations first.
* [ ] Call the same route without `include_deprecated_columns` and make sure that the names aren't included

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes
This adds the privacy declaration name to the datamap response so it can be shown in the datamap table in plus.


src/fides/core/export_helpers.py
```diff
@@ -350,7 +351,6 @@ def union_data_categories_in_joined_dataframe(joined_df: pd.DataFrame) -> pd.Dat
     # isolate the system data categories into a new dataframe and create a common column
     joined_df = joined_df.drop(
         [
-            "system.privacy_declaration.name",
```
TheAndrewJackson, line 353: I noticed that you're tagged with the git blame for this line. Do you know why this field was being dropped? @SteveDMurphy

adamsachs: i certainly don't have the necessary context, but i did notice in the datamap's system configuration UI, the `Privacy declaration name` field is listed as "deprecated" - so maybe that relates to why it was being dropped on the response here?

image

TheAndrewJackson:
@pattisdr Could you take a look at this PR? Idk if you have context for this area of the codebase but it's a relatively small change so hopefully a lot of context isn't needed 🤞

src/fides/core/export_helpers.py
SteveDMurphy: Sorry for the holdup @TheAndrewJackson ! Just looking but initial thoughts are that I believe it would have caused duplicate rows if the same data category/use/subject were used commonly across privacy declarations. 

I don't know that it was actually a requirement of what was originally on the datamap so it may have been dropped to remove duplication (which was an early problem with the datamap). 

Adding it back in will likely have a potential for further duplication in those scenarios

TheAndrewJackson: I see. Do you think it's too risky to merge this? We can always roll the change back if it causes issues. 

This column will probably be removed again once the customer has migrated away from using it.

src/fides/api/ctl/routes/datamap.py
```diff
@@ -100,6 +105,7 @@
 async def export_datamap(
     organization_fides_key: str,
     response: Response,
+    include_deprecated_columns: bool = False,
```
SteveDMurphy, line 108: I think this helps a lot, I noticed in earlier testing that the open-source export was broken with the added field and this resolves that 🙌🏽

TheAndrewJackson:
@SteveDMurphy I'm not sure why the codecov is saying those new lines aren't tested. I added a new test that makes sure that code runs and works

tests/ctl/api/test_datamap.py
```diff
@@ -544,6 +556,22 @@ def test_datamap_with_privacy_declaration(
...
+@pytest.mark.integration
+@pytest.mark.usefixtures("system_privacy_declarations")
+def test_datamap_with_privacy_declaration_with_name(
+    test_config: FidesConfig,
+    test_client: TestClient,
+) -> None:
+    response = test_client.get(
+        test_config.cli.server_url
+        + API_PREFIX
+        + "/datamap/default_organization?include_deprecated_columns=true",
```
NevilleS, line 568: nit: if I'm being really picky, you should do another query with this set to `false` and confirm the deprecated columns are excluded

NevilleS (approved):
This looks reasonable, thanks @TheAndrewJackson. I do think a tiny bit more test coverage would be worth doing though

pattisdr:
Ah sorry @TheAndrewJackson missed that you'd tagged me here, I'll try to take a look in a bit

TheAndrewJackson:
> Ah sorry @TheAndrewJackson missed that you'd tagged me here, I'll try to take a look in a bit

All good. I should be good to merge now that @NevilleS approved it

PR #2254, opened 2023-01-17
Unified Fides Resources Feature

pattisdr:
❗ ~Prerequisite: We need a fideslang release, so we can link to the proper fideslang version here~
❗ Before merge:
- Get up to date with main again and push to this branch to re-trigger CI
-  Double check if downrev needs to be bumped: this branch has several migrations. Bump the first in the series.
- See if any new saas connectors have been added - their tests will fail because they're using old DatasetConfig.dataset fields
- Other uses of now-obsolete resources should get flagged via test failures.

❗ If you primarily care about your ctl datasets, migration plan would be to delete your dataset configs prior to upgrading.

Fixes [#2262](https://github.com/ethyca/fides/issues/2262)

### Code Changes

This feature branch contains the work below from several PR's - they've all been individually reviewed, and I've been keeping this branch up-to-date with main:

* Unified Fides Resources: Remove several fidesops schemas for DSR's in favor of updated Fideslang schemas [#2009](https://github.com/ethyca/fides/pull/2009)
* Unified Fides Resources: Add new `datasetconfig.ctl_dataset_id` field to unify fides dataset resources [#2046](https://github.com/ethyca/fides/pull/2046)
* Unified Fides Resources: Remove `DatasetConfig.dataset field` [#2096](https://github.com/ethyca/fides/pull/2096)
* Unified Fides Resources: Update UI dataset config routes to use new unified routes [#2113](https://github.com/ethyca/fides/pull/2113)
* Unified Fides Resources: Validate request body on crud endpoints on upsert. Validate dataset data categories before save. [#2134](https://github.com/ethyca/fides/pull/2134/)
* Unified Fides Resources: Added a dataset dropdown selector when configuring a connector to link an existing dataset to the connector configuration. [#2162](https://github.com/ethyca/fides/pull/2162)
* Unified Fides Resources:  Update test env setup and quickstart to use new routes instead of routes we plan to deprecate at some point [#2225](https://github.com/ethyca/fides/pull/2225)

### Steps to Confirm

* [x] Run `nox -s test_env`.  Verify in your db that DatasetConfig.dataset column is gone. Instead there's a DatasetConfig.ctl_dataset_id FK to `ctl_dataset`
```
fides=# select * from datasetconfig;
                    id                    |          created_at           |          updated_at           |           connection_config_id           |           fides_key           |              ctl_dataset_id              
------------------------------------------+-------------------------------+-------------------------------+------------------------------------------+-------------------------------+------------------------------------------
[...]

PR #2841, opened 2023-03-15
fix missing fideslang Dataset attributes when overwriting ctl_datasets

adamsachs:
Closes https://github.com/ethyca/fidesplus/issues/668

Within `upsert_with_ctl_dataset()`, in the case where we have an existing `ctl.Dataset`, we were overwriting the dataset's `collections` attribute (a JSON blob) with a data object that was not generated from a `fideslang.Dataset` instantiation. That `fideslang.Dataset` instantiation provides default (generally null/empty) values for a bunch of attributes within the `collections` objects, even if they're not passed in with the incoming data dict. These attributes (even if null/empty) need to be included on the `ctl.Dataset`'s `collections` JSON blob for other downstream code (e.g. the datamap export) to work as expected. Generally, we should try to keep the `collections` data structure consistent no matter what code path in the application is creating/updating it -- unfortunately, since it's just a JSON blob on the `ctl.Dataset` (and not "strongly typed"), we miss some built-in type checking, which is how this problem came about in the first place. Which just means we need to be extra diligent...

### Code Changes

* use the well-populated `validated_data` `fideslang.Dataset` object as the basis for the `upsert_ctl_dataset` operation even in cases where it's an update
* update tests to specifically look for these additional fields on `collections` and their subobjects that the `fideslang.Dataset` instantiation initializes

### Steps to Confirm

* check the linked issue's `Steps to Reproduce (on fidesplus)` section for steps to repro the problem. the steps can be performed without `fidesplus` and instead on `fides` by going thru the same steps but just invoking the `GET /api/v1/datamap/default_organization` endpoint rather than going to the datamap UI

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

    * previously, we'd only been using the well-populated `validated_data` object for the _create_ path of the `upsert_ctl_dataset()` function, which was therefore working fine. it seems like the empty attribute instantiation was an unrecognized side effect that actually turns out to be important in some cases
    * i don't _think_ there's any problem we introduce by using this well-populated `validated_data` object for the _update_ path of the upsert function, rather than just "raw" `ctl_dataset_data` dict of update data being passed in.


src/fides/api/ops/models/datasetconfig.py
```diff
@@ -69,11 +69,10 @@ def upsert_ctl_dataset(ctl_dataset_obj: Optional[CtlDataset]) -> CtlDataset:
             """
             ctl_dataset_data = data.copy()
             validated_data = Dataset(**ctl_dataset_data.get("dataset", {}))
-
             if ctl_dataset_obj:
                 # It's possible this updates the ctl_dataset.fides_key and this causes a conflict
                 # with another ctl_dataset, if we fetched the datasetconfig.ctl_dataset.
-                for key, val in ctl_dataset_data.get("dataset", {}).items():
+                for key, val in validated_data.dict().items():
```
pattisdr, line 75: Ah geez sorry about this.  A big theme of the Unified Fides Resources feature https://github.com/ethyca/fides/pull/2254 was this very thing, where we had traditionally been saving CTL Datasets without the extra fields initialized in validation, and I'm introducing another source here 🤦 

good troubleshooting!

pattisdr (approved):
changes make sense to me @adamsachs!

src/fides/api/ops/models/datasetconfig.py
adamsachs: no need to apologize, this is very deep in the weeds and i'm sure that refactor was a very complex effort. this was a fun one to track down! 😄

PR #2747, opened 2023-03-03
Update scopes and roles in the UI

allisonking:
Closes https://github.com/ethyca/fides/issues/2743

This will impact https://github.com/ethyca/fides/pull/2739, I think only in when it lists out what roles are available

### Code Changes

* [x] Update TS autogenerated types
* [x] Update various home/tile restrictions based on the new more granular scopes
* [x] Use `total_scopes` instead of `scopes`

### Steps to Confirm

* [ ] things should work the same as before

### Pre-Merge Checklist

* [ ] All CI Pipelines Succeeded
* Documentation:
  * [ ] documentation complete, [PR opened in fidesdocs](https://github.com/ethyca/fidesdocs/pulls)
  * [ ] documentation [issue created in fidesdocs](https://github.com/ethyca/fidesdocs/issues/new/choose)
* [ ] Issue Requirements are Met
* [ ] Relevant Follow-Up Issues Created
* [ ] Update `CHANGELOG.md`

### Description Of Changes

_Write some things here about the changes and any potential caveats_

PR #2848, opened 2023-03-16
Deprecate User Scopes

### LINKED ISSUE #2696: Deprecate Assigning Scopes to Users Directly

With https://github.com/ethyca/fides/pull/2671, in addition to being able to be assigned individual scopes directly, users can also be assigned an individual role.  Roles are associated with a set of scopes. 

While we still need to support being able to create tokens with individual scopes, there is some concern over users that have both roles and scopes assigned to them.   The "scopes" UI is going away so there will be less visibility into who has these extra permissions that they've been assigned via a singular scope. 

Deprecate being able to **assign a scope to a user directly**.  The FidesUserPermission.scopes column would go away.  Clients created/updated from user permissions would only have their *roles* persisted to the client, and then to the token.

To still support users building applications on top of fides, using API-only workflows, or desiring tokens with very limited scopes, they can still use the existing client endpoints to create clients with a narrow set of scopes, and then create a token directly from that client.

### Additional context

pattisdr:
❗ Contains migration; check if downrev needs to be bumped before merging.

Closes #2696 

### Code Changes

* [ ] Removes the `fidesuserpermissions.scopes` column entirely.  Also updates any clients attached to users to remove their scopes one last time.
* [ ] Updates create user permissions and update user permissions to ignore passed-in scopes
* [ ] Also removed the "scopes" key from being returned with get user permissions
* [ ] Update having the CLI printing Associated scopes (total scopes) instead of "directly-assigned scopes" when getting user permissions
* [ ] When logging in, only create or update the ClientDetail resource to have the user's roles and systems, not the user's scopes. 

#### Verify role-only flow still works 
* [ ] Run `nox -s "fides_env(test)"`
* [ ] Login as a root user, create other users with various roles
* [ ] Login as users with various roles, verify user experience is the same.

#### Verify scopes-only flow still works
* [ ]  Get a root client token with a POST to {{host}}/oauth/token with the OAUTH_ROOT_CLIENT_ID and OAUTH_ROOT_CLIENT_SECRET
* [ ] Use the root client token to create a new client with limited scopes POST {{host}}/oauth/client:
```
[
    "system_manager:delete",
    "system_manager:update",
    "system_manager:read"
]
```
* [ ] Get a token for that client with a POST to {{host}}/oauth/token with the client_id and client_secret
* [ ] Verify that token has only the above scopes and those scopes let it read/update/delete system managers and nothing else

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

Removes assigning scopes to users directly; you can only assign roles. Clients can still be assigned scopes directly. Permission checking still looks to see if you have the correct scope either via one you were assigned directly or through a role associated with your scope.


src/fides/api/ctl/database/seed.py
```diff
@@ -89,14 +85,7 @@ def create_or_update_parent_user() -> None:
...
             db=db_session,
-            data={
-                "user_id": user.id,
-                "scopes": [
-                    PRIVACY_REQUEST_CREATE,
-                    PRIVACY_REQUEST_READ,
-                    PRIVACY_REQUEST_TRANSFER,
-                ],
-            },
+            data={"user_id": user.id, "roles": [OWNER]},
```
pattisdr, line 88: We're no longer giving users scopes directly - in creating this parent user, there is no existing role that has these scopes. Making `owner` instead - does that work when seeding this data @adamsachs?

src/fides/lib/models/fides_user_permissions.py
```diff
@@ -21,10 +20,16 @@ class FidesUserPermissions(Base):
...
             all_scopes += ROLES_TO_SCOPES_MAPPING.get(role, [])
 
         return sorted(list(set(all_scopes)))
+
+    @property
+    def scopes(self) -> List[str]:
+        """Backwards-compatible property for use while the UI is still looking at userpermission.scopes
+        Users can no longer be assigned scopes directly.
+        """
+        return self.total_scopes
```
pattisdr, line 35: For backwards compatibility if the UI is still using FidesUserPermission.scopes anywhere. I am returning all scopes they have via their roles (total scopes) but alternatively we could return an empty list.

src/fides/lib/oauth/api/routes/user_endpoints.py
```diff
@@ -82,10 +80,7 @@ def create_user(
     logger.info("Created user with id: '{}'.", user.id)
     FidesUserPermissions.create(
         db=db,
-        data={
-            "user_id": user.id,
-            "scopes": [],
-        },
+        data={"user_id": user.id, "roles": []},
```
pattisdr, line 83: Still creating a user with no roles by default but that is subject to change in https://github.com/ethyca/fides/issues/2866

src/fides/lib/oauth/api/routes/user_endpoints.py
```diff
@@ -245,20 +240,19 @@ def perform_login(
             db,
             client_id_byte_length,
             client_secret_btye_length,
-            scopes=user.permissions.scopes,  # type: ignore
+            scopes=[],  # type: ignore
```
pattisdr, line 243: When a user logs in, the created client doesn't have scopes, only roles or systems.  However, clients created directly can still have scopes.

tests/conftest.py
```diff
@@ -211,18 +208,21 @@ def user(db):
...
-        scopes=SCOPE_REGISTRY,
+        roles=[APPROVER],
+        scopes=[],
         user_id=user.id,
     )
 
-    FidesUserPermissions.create(
-        db=db, data={"user_id": user.id, "scopes": [PRIVACY_REQUEST_READ]}
-    )
+    FidesUserPermissions.create(db=db, data={"user_id": user.id, "roles": [APPROVER]})
```
pattisdr, line 216: We had a base test user whose client had more scopes than the user did which caused a couple of tests to pass that shouldn't have due to the mismatch.  Anyway, now that users can't be assigned scopes directly, making this default user an `APPROVER` instead.

tests/fixtures/application_fixtures.py
```diff
@@ -1334,37 +1334,6 @@ def succeeded_privacy_request(cache, db: Session, policy: Policy) -> PrivacyRequ
...
-    db.commit()
-    db.refresh(client)
-    yield user
-    try:
-        client.delete(db)
-    except ObjectDeletedError:
-        pass
-    user.delete(db)
-
-
```
pattisdr, line 1367: Removing in favor of the user fixture in conftest.py

tests/lib/test_oauth_util.py
```diff
@@ -197,26 +200,37 @@ async def test_verify_oauth_client_wrong_client_scope(db, config, user):
...
-    async def test_token_does_not_have_roles(self, db, config, user):
-        """Test no error even if there are not roles on the token"""
+    async def test_token_does_not_have_roles(self, db, config):
+        """Test that roles aren't required to be on the token - scopes can still be assigned directly"""
+        client, _ = ClientDetail.create_client_and_secret(
+            db,
+            CONFIG.security.oauth_client_id_length_bytes,
+            CONFIG.security.oauth_client_secret_length_bytes,
+            scopes=[PRIVACY_REQUEST_REVIEW],
+            user_id=None,
```
pattisdr, line 210: This is one of the tests that was passing because the user's client had more scopes than the user did.   This is why the test needs more refactoring just to prove that scopes can still be assigned directly.

tests/ops/api/v1/endpoints/test_privacy_request_endpoints.py
```diff
@@ -1947,7 +1948,7 @@ def test_approve_privacy_request(
         privacy_request.save(db=db)
 
         payload = {
-            JWE_PAYLOAD_SCOPES: user.client.scopes,
+            JWE_PAYLOAD_ROLES: user.client.roles,
```
pattisdr, line 1951: The `user` fixture now only has roles not scopes, so this token should likewise have roles

tests/ops/api/v1/endpoints/test_user_permission_endpoints.py
```diff
@@ -43,45 +49,20 @@ def test_create_user_permissions_wrong_scope(
...
 
-    def test_create_user_permissions_invalid_scope(
-        self,
-        db,
-        api_client,
-        generate_auth_header,
-        user,
-        url,
-    ) -> None:
-        auth_header = generate_auth_header([USER_PERMISSION_CREATE])
```
pattisdr, line 54: obsolete test

src/fides/api/ctl/database/seed.py
adamsachs: yes, that should be fine. the only risk i'd see is that we're giving the "parent user" broader scope than is strictly necessary, but no human user should actually be authenticating as this user, it's just meant for inter-server communications (i.e. the fides _child_ uses these credentials to connect to _this_ fides, which is effectively acting as a _parent_ if it has these variable set). so as long as we don't update the fides connector code to accidentally do destructive operations on the parent fides it connects to, there's no impact here.

pattisdr: thanks for weighing in Adam!

src/fides/api/ops/schemas/user_permission.py
```diff
@@ -36,8 +35,12 @@ class UserPermissionsResponse(UserPermissionsCreate):
 
     id: str
     user_id: str
-    scopes: List[ScopeRegistryEnum]
-    total_scopes: List[ScopeRegistryEnum]
+    scopes: List[
+        ScopeRegistryEnum
+    ]  # This should be removed once the UI is not using.  Currently returning "total_scopes" here.
```
allisonking, line 40: I'm pretty sure the UI stopped using `scopes` [in this PR](https://github.com/ethyca/fides/pull/2747)

src/fides/lib/models/fides_user_permissions.py
allisonking: I think [here](https://github.com/ethyca/fides/blob/main/clients/admin-ui/src/features/user-management/PermissionsForm.tsx/#L55-L57) is the only place still referring to `scopes` in the frontend, and that was also a backwards compatibility thing

allisonking (commented):
this looks great! Looking through the code, I'm pretty sure the FE doesn't rely on the `scopes` field anymore. I guess the only way to really tell would be to remove the backwards compatibility protections you put in place and see if the UI breaks (I can help test that if you'd like!)

pattisdr:
OK thanks very much @allisonking! I'll look into getting rid of scopes entirely.

allisonking (approved):
tested things out and I think everything is working well, nice job! 🎉

pattisdr:
Thanks so much for testing @allisonking 🎉

PR #2859, opened 2023-03-16
handle different orders of custom field additions in datamap export

### LINKED ISSUE #2850: Custom Fields in the map don't align with values - when adding fields at different times
1. Start Fides Plus
2. Create 3 custom fields
3. Add values to fields on 2 systems
4. Check the data map - the columns should align with the right values
5. Go back to a system
6. Add a 4th field
7. Assign a value
8. Go back to the map - the columns and values won't line up.

**Expectation**
At all times, the values of the custom fields columns should match the field name in the column header
Screenshot 2023-03-16 at 11 30 47 AM

adamsachs:
Closes #2850 

### Code Changes

* keep an ordered list of custom field headers, in the order that they appear in the header row
* leverage that ordered list when populating custom field row data
* add some test coverage for the error case and related cases

### Steps to Confirm

* we should be able to populate the DB similar to how we're doing it in the test cases and repro with the datamap export API solely on `fides`
* even better would be to cut an edge build off this release branch (not something we've normalized, yet, but maybe we can give it a try!) and test this with the fidesplus FE
    * this will allow us to create and populate custom fields a lot more easily, and therefore stress test with more scenarios

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### Description Of Changes

i started considering a more significant refactor to the code where we don't rely on an implicit ordering at all, and we're very explicit about populating the row data based on the order of the headers. i think we should be able to do that, and it'd make the code both more robust and readable. adding in a list to just keep track of the custom field header ordering feels like duct tape. but we need a fix quick, and i'm out for the rest of the weekend, so i at least wanted to get a patch pushed up that we can work with! :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@ The types of changes are:
 
 * Fixed issue where unsaved changes warning would always show up when running fidesplus [#2788](https://github.com/ethyca/fides/issues/2788)
 * Fixed problem in datamap export with datasets that had been updated via SaaS instantiation [#2841](https://github.com/ethyca/fides/pull/2841)
+* Fixed problem in datamap export with inconsistent custom field ordering [#2859](https://github.com/ethyca/fides/pull/2859)
 
 ## [2.8.3](https://github.com/ethyca/fides/compare/2.8.2...2.8.3)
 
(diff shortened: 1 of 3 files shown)


SteveDMurphy (approved):
Nice way to work through some of this @adamsachs ! Thanks again for the extra tests and breadcrumbs around what is happening

PR #2869, opened 2023-03-20
Improve standard layout for large width screens and polish misc. pages

NevilleS:
### Code Changes

* [X] Set min width on entire page to "large" 1024px (since we really don't scale down to small mobile width!)
* [X] Improve add systems layout at large width
* [X] Fix home banner layout:
  * [X] Use theme tokens for all typography
  * [X] Use background image "contain" to fit banner image to width
  * [X] Position background image to avoid overlapping with banner text
* [X] Change home screen tile grid to change from 3 -> 2 columns at large width
* [X] Fix tiles resizing on hover due to border
* [X] Add basic fluid layout to connection manager filter bar

### Steps to Confirm

* [X] Run `nox -s "fides_env(test)"` and populate with several privacy requests to allow testing scroll behaviour
* [X] View Home page at large width (1024px) and confirm banner image is legible, and tiles scroll
* [X] View Privacy Request page at large width (1024px)
* [X] Scroll Privacy Request page and ensure the full page scrolls (no persistent header)
* [X] View other pages at large width (1024px) and check layouts

| Page | Before | After |
|---|---|---|
| Home (1024px) | image | image| 
| Privacy Requests (1024px) | image| image|
| View Systems (1024px) |image| image|
| Add Systems (1024px) | image| image|
| User Management (1024px) | image |image|

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### Description Of Changes

This is a polish sweep of the Admin UI to tighten up the overall layout and improve the usability at ~large width screens and above (1024px+). Most importantly, this makes two key changes:
1. Sets a minimum width of `1024px` (medium) so that the site doesn't complete mangle itself at small widths and instead just scrolls horizontally 😄 
2. Makes the header, nav, and main layout consistently fill the viewport height (100vh) across all pages (incl. fidesplus)

⚠️ This doesn't actually make the site responsive, but it's a small step towards that ⚠️ 

Lastly this includes fixes for the home screen layout in particular, which would scroll horizontally at <1040px and squish the home tiles unreasonably. I also replaced a lot of static, `px` styles with theme tokens for consistency and improved the layout at medium width to be legible using some clever background blurring which was fun.

Note that the overall layout needs to stay consistent with `fidesplus`, which has pages that need to occupy the full viewport, so see the related PR on `fidesplus` for screenshots and tests showing why the `height: 100vh` is still needed.

### DIFF
--- a/clients/admin-ui/src/home/HomeBanner.tsx
+++ b/clients/admin-ui/src/home/HomeBanner.tsx
@@ -1,4 +1,4 @@
-import { Box, Flex, Spacer, Text } from "@fidesui/react";
+import { Flex, Text } from "@fidesui/react";
 import * as React from "react";
 
 import { useFeatures } from "~/features/common/features";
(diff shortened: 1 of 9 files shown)


NevilleS:
@TheAndrewJackson no that's not intentional - I'd like that to be consistent. It took a lot of trial & error to come up with the lowest-common-denominator CSS that worked for the three layout styles (home screen, regular pages, datamap pages). That looks like at the small screen size breakpoint (<768px) the scroll is happening for the full `body` instead of the `main` div... I'll tinker more

NevilleS:
Feedback from design:

- [ ] Set the min width to lg (998px) instead of medium (768px)
- [ ] Revert back to the full-page scrolling, no persistent/fixed header
- [ ] Remove the drop shadow on the home screen tiles
- [ ] Remove the (clever!) scrim on the home banner image; instead pin the image to the right until it collides with text box and then hide overflow

NevilleS:
OK, incorporated the design feedback and updated the screenshots. I think this is good to go now... @TheAndrewJackson while I also preferred the pinned header I checked with the design team and they prefer full page scrolling right now so I reverted that, so I didn't end up having to investigate why my implementation was bugged 😄 

Let me know if you think this is ready to go

clients/admin-ui/src/home/HomeBanner.tsx
```diff
@@ -72,12 +58,15 @@ const HomeBanner: React.FC = () => {
           </>
         )}
       </Flex>
+      {/* Render the background image, using a min-width here to ensure there is
+      enough left margin to avoid colliding with the banner text above */}
```
TheAndrewJackson, line 62: Thanks for leaving a comment about the `minWidth` below. I wasn't sure why it was there at first. Creative use of it there!

TheAndrewJackson (approved):
I think this is looking good. Solid improvement. Let's ship it and iterate more in the future if needed 🚀

Issue #2824, opened 2023-03-14
UX cleanups for Role Assignment UI


eastandwestwind:
2 remaining items for UX clean up:

### 1. Fix Role Description Drawer
- [x] Update Role Description drawer to be higher up in the page. Currently this is within the "Permissions" tab due to the fact that I'm unsure how to dynamically display the drawer if it's in a parent component. It should only display with the user is on the "Permissions" tab. 

"So on scrolling, the “help panel” for want of a better term should be fixed height, pinned to the top and bottom of the main window, below the main header and nav. Scrolling should be possible within the “help panel” so that we have the flexibility to add more items to the help panel. Otherwise, if  we have 10 help items for example, we would be scrolling the whole page to view them all rather then what is intended as a side by side panel that can be accessed while determining what role to assign. The slide-out that was built for the data map graph and table view used the chakra drawer component to achieve a similar result." -- Simon

There's a possibility we can use Chakra's drawer component so that the Role description UI sticks to the side of the page, like this- https://chakra-ui.com/docs/components/drawer

Desired:
<img width="849" alt="Screenshot 2023-03-10 at 5 09 17 PM" src="https://user-images.githubusercontent.com/7697292/224370502-532af9f2-637e-4d37-b67f-964d4831b002.png">

Current:
<img width="1403" alt="Screenshot 2023-03-10 at 5 09 06 PM" src="https://user-images.githubusercontent.com/7697292/224370770-33798826-9ab7-49c1-a5aa-36915244c11c.png">

### 2. Fix Trash Icon
- [ ] adjust trash can icon throughout to match mocks

Expected:
<img width="611" alt="Screenshot 2023-03-09 at 2 35 50 PM" src="https://user-images.githubusercontent.com/7697292/224039840-e9374879-e229-448d-b915-0a9a65ca1176.png">
Current:
<img width="513" alt="Screenshot 2023-03-09 at 12 33 46 PM" src="https://user-images.githubusercontent.com/7697292/224039847-8fc10b58-7601-45d0-954b-06b50c69f2b6.png">

rsilvery:
Can we add the garbage can icon one to this ticket?

rsilvery:
Could we add a task to show assigned systems in the user table?


![Image](https://user-images.githubusercontent.com/8682444/225638059-8af4e6cc-1c05-4aa6-9730-429f0bd38fa8.png)

rsilvery:
[...]

PR #2870, opened 2023-03-20
2824 ux cleanups role assignment

eastandwestwind:
Closes https://github.com/ethyca/fides/issues/2824

### Code Changes

* [ ] Update role description help drawer to match mocks
* [ ] Use appropriate trash can icon in Role Management UI to match mocks

### Steps to Confirm

* [ ] See https://github.com/ethyca/fides/issues/2824

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes


CHANGELOG.md
```diff
@@ -30,7 +30,8 @@ The types of changes are:
 ### Changed
 
 * Removed "progressive" navigation that would hide Admin UI tabs until Systems / Connections were configured [#2762](https://github.com/ethyca/fides/pull/2762)
-* Added `system.privacy_declaration.name` to datamap response [#2831][https://github.com/ethyca/fides/pull/2831/files]
+* Added `system.privacy_declaration.name` to datamap response [#2831](https://github.com/ethyca/fides/pull/2831/files)
+* UX cleanups in Role Management section of Admin UI [#2870](https://github.com/ethyca/fides/pull/2870)
```
allisonking, line 34: this'll have to move up to the `Unreleased` section :)

clients/admin-ui/src/features/common/custom-fields/CreateCustomLists.tsx
```diff
@@ -8,9 +8,9 @@ import { useAlert } from "~/features/common/hooks";
 import { useUpsertAllowListMutation } from "~/features/plus/plus.slice";
 import { AllowListUpdate } from "~/types/api";
 
+import { TrashCanSolidIcon } from "../Icon/TrashCanSolidIcon";
```
allisonking, line 11: nit: in general, we prefer not to do much relative imports. I'd say in the same folder is okay (i.e. `./component`) but once we're going to different folders like `../`, it starts to get confusing when you have to move components around. this could be an absolute import instead, i.e. `~/features/common/Icon/...` which means when you move the file the imports don't have to change.

but sometimes your IDE will import things for you relatively instead of absolutely and it'll sneak up on you 😅

clients/admin-ui/src/features/user-management/UserForm.tsx
```diff
@@ -41,8 +41,8 @@ export type FormValues = typeof defaultInitialValues;
 
 const ValidationSchema = Yup.object().shape({
   username: Yup.string().required().label("Username"),
-  first_name: Yup.string().label("First name"),
-  last_name: Yup.string().label("Last name"),
+  first_name: Yup.string().required().label("First name"),
+  last_name: Yup.string().required().label("Last name"),
```
allisonking, line 45: These fields aren't actually required in the backend. Rachel had filed a ticket that said they were, but I think that was actually just a bad UX thing, where if the password didn't meet requirements, it would look like the form failed (and in rachel's case, she didn't have a first or last name). I would remove the `.required()` unless we want to make the backend require it

clients/admin-ui/src/features/user-management/UserManagementTabs.tsx
```diff
@@ -36,10 +36,19 @@ const UserManagementTabs = ({
       label: "Permissions",
       content: (
         <Flex gap="97px">
-          <Box w={{ base: "100%", md: "65%", xl: "55%" }}>
+          <Box w={{ base: "100%", md: "50%", xl: "50%" }}>
             <PermissionsForm />
           </Box>
-          <Box w={{ base: "100%", md: "35%", xl: "45%" }}>
+          <Box
+            position="absolute"
```
allisonking, line 43: heh yeah that's one way to do it!

clients/admin-ui/src/features/user-management/UserForm.tsx
allisonking: just saw that @rsilvery called this out in particular in the ticket, so here's what I would push for:

* Should first and last name be required? They are not in the backend, and I don't think in general apps require it, so I would say no
* I imagine Rachel requested this at first because it _looked_ like they were required back when password validation wasn't working. But now that password validation is fixed, you should be able to input a user without a first or last name just fine, and so I'd say that we don't need to make these required fields on the frontend

clients/admin-ui/src/features/user-management/UserManagementTabs.tsx
```diff
@@ -36,10 +36,19 @@ const UserManagementTabs = ({
...
+            top="114px"
+            right="24px"
+            overflowY="scroll"
+            padding="24px"
+            w="35%"
+            borderLeftWidth="1px"
+            borderBottomWidth="1px"
```
allisonking, line 50: the designs don't look like they have a bottom border? can we try giving this a `height="100%"` so that the vertical line will go all the way down, and then we wouldn't need a bottom border?

allisonking (commented):
there's some funkiness if you shrink the width down too much, but I don't think it's a priority (we don't have designs for what a small viewport should look like anyway!)

image

CHANGELOG.md
eastandwestwind: thanks!

clients/admin-ui/src/features/common/custom-fields/CreateCustomLists.tsx
eastandwestwind: ok I'll update this

clients/admin-ui/src/features/user-management/UserForm.tsx
eastandwestwind: @rsilvery - mind confirming that first/ last name shouldn't be a requirement on FE?

clients/admin-ui/src/features/user-management/UserManagementTabs.tsx
eastandwestwind: this was a tricky one, using `height="100%"` only made the height extend to the viewable screen, not below that. I'll see if I can work around this someone.

clients/admin-ui/src/features/user-management/UserManagementTabs.tsx
```diff
@@ -36,10 +36,19 @@ const UserManagementTabs = ({
...
+          <Box w={{ base: "100%", md: "50%", xl: "50%" }}>
             <PermissionsForm />
           </Box>
-          <Box w={{ base: "100%", md: "35%", xl: "45%" }}>
+          <Box
+            position="absolute"
+            top="114px"
+            right="24px"
+            overflowY="scroll"
+            padding="24px"
```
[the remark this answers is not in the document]
eastandwestwind: gotcha, I'll test and see what these px vals correspond to for chakra vals. Thanks for pointing this out!

clients/admin-ui/src/features/user-management/UserManagementTabs.tsx
```diff
@@ -36,10 +36,19 @@ const UserManagementTabs = ({
...
+            w="35%"
+            borderLeftWidth="1px"
+            borderBottomWidth="1px"
+          >
             <RoleDescriptionDrawer />
```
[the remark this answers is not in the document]
eastandwestwind: Good call, I'll adjust

clients/admin-ui/src/features/user-management/UserManagementTabs.tsx
eastandwestwind: For now I've allowed some extra height to be added to the page such that assigning a couple systems to users does not cause the vertical line to be shorter than the page:

```
height="calc(100% + 100px)"
```

This is not dynamic based on how many systems are assigned to the user, which is one downfall of absolutely positioned elements. 

Ideally this vertical line element would live on the permissions drawer element, but since the line extends above the permissions drawer element, it would have to be absolutely positioned as well, and it would pose greater challenges for calculating exact position.

I'm open to better ideas @allisonking.

clients/admin-ui/src/features/user-management/UserForm.tsx
rsilvery: I don't think it needs to be so long as the BE isn't requiring it.

allisonking (approved):
looks great, thank you! only thing is let's make first and last name not required, then I think we are good to go!

PR #2884, opened 2023-03-21
Allow login to automatically pull from env vars

### LINKED ISSUE #2786: Allow `fides user login` to pull from env vars

Whether in CI or just as a local CLI user, it would be nice to have the `username` and `password` taken directly from environment variables as opposed to needing to use options or enter them interactively.

ThomasLaPiana:
Closes #2786 

### Code Changes

* [x] add the `user.username` and `user.password` fields to the config
* [x] update `fides user login` to pull from the config automatically if the options aren't specified, finally falling back on prompting if not provided as an option and not in the config
* [x] update `fides user create` to use username and password arguments instead of options, display assigned roles upon successful creation
* [x] add the `username` and `password` to the development config file

### Steps to Confirm

* [ ] confirm that the `user.username` and `user.password` fields are set in the `fides.toml`:
  * [ ] run `fides user login` and confirm "it just works"
  * [ ] run `fides user login -u notarealuser` and confirm it breaks 
* [ ] remove the `user.username` and `user.password` fields from the `fides.toml`
  * [ ] run `fides user login` and confirm it prompts for them
* [ ] `fides user create newuser Testpassword1!` is successful and shows the assigned roles (always `owner` currently)

### Pre-Merge Checklist

* Documentation:
  * [x] documentation complete, [PR opened in fidesdocs](https://github.com/ethyca/fidesdocs/pulls) -> [CLI Auth Documentation PR](https://github.com/ethyca/fidesdocs/pull/79)
* [x] Update `CHANGELOG.md`
* [ ] ~~For API changes, the [Postman collection](https://github.com/ethyca/fides/blob/main/docs/fides/docs/development/postman/Fides.postman_collection.json) has been updated~~

### Description Of Changes

This PR is designed to make logging in via the CLI more user-friendly. It allows users to set a `username` and `password` via the config file (and therefore via `env vars` if so inclined) that will get automatically used during the login process.

It should be noted that username/password are considered discrete values here, so one may configure their `username` via the config file and it will still prompt for the `password`

#### Login Logic Update

The new priority list is as follows:
1. If a username/password option is provided directly, use that value
1. If no flags are provided, check the config (which would automatically load env vars as well, such as `FIDES__USER__USERNAME`)
1. If no option was provided and the config is empty, prompt the user for the value

#### Example Config Section

```toml
[user]
username = "root_user"
password = "password
```


src/fides/core/user.py
```diff
@@ -139,14 +139,15 @@ def create_command(
         server_url=server_url,
     )
     user_id = user_response.json()["id"]
+    new_user_roles = CONFIG.security.root_user_roles
     update_user_permissions(
         user_id=user_id,
         scopes=CONFIG.security.root_user_scopes,
```
pattisdr, line 145: I think I missed this in my PR https://github.com/ethyca/fides/pull/2848 to remove user scopes - can you clean up here?  Users can no longer be assigned scopes directly, just roles.  The API ignores the "scopes" key if it is passed in, so this should still all be working, but it would be good to remove this scopes attribute from this function altogether.

ThomasLaPiana: Oh, thanks for the tip! Will fix :)

PR #2891, opened 2023-03-21
[Backend] Restrict Setting Approvers as System Managers [#2854]

### LINKED ISSUE #2854: Restrict ability to assign Systems to users with only the "Approver" role via the API
### Bug Description

It's currently possible to assign a user with the "Approver" role the ability to manage a System. This would create a lot of confusing patterns in the Admin UI so it's best to prevent that.

#### Steps to Reproduce

n/a

### Expected behavior

The API should throw an error if an Approver is assigned a system

pattisdr:
Closes #2854
❗ DO NOT MERGE, blocked by FE https://github.com/ethyca/fidesplus/issues/722 and https://github.com/ethyca/fides/issues/2898

### Code Changes

* [ ] Requires that a user has a "role" before systems can be added
* [ ] Along those lines, if a user has "approver" listed as one of their roles, don't let a system be added
* [ ] If a user is turned into an Approver and they have systems, remove them as system managers in the background

### Steps to Confirm

* [ ] Login as the root user. Create a Contributor. Verify you can add systems. Save systems
* [ ] Update that Contributor to an Approver. Refresh. Verify via an API call that the user's systems were removed.
* [ ] Try to add systems to the approver via the API. Verify you are prevented from doing so.

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

- Prevents systems from being added to a user with an Approver role.  (Checks to see if an Approver is in their list of roles. Technically in the backend, users can be assigned many roles. This was done for flexibility in the future. But in practice, users just get one role right now via the UI. You could alternatively check if they have any other roles besides Approver, but what if we added a role that was even more restrictive than Approver? I feel this is safer)
- If a user is downgraded from another role to an Approver and they have systems, they are removed as system manager from those systems.


src/fides/api/ops/api/v1/endpoints/user_endpoints.py
```diff
@@ -231,6 +232,18 @@ def update_managed_systems(
...
+        raise HTTPException(
+            status_code=HTTP_400_BAD_REQUEST,
+            detail=f"User {user_id} needs permissions before they can be assigned as system manager.",
+        )
+
+    if APPROVER in user.permissions.roles:  # type: ignore
+        raise HTTPException(
+            status_code=HTTP_400_BAD_REQUEST,
+            detail=f"User {user_id} is an {APPROVER} and cannot be assigned as a system manager.",
+        )
```
pattisdr, line 245: It feels like there's a mismatch between the UI and the desired logic on the backend. 

In the UI, you can currently add systems to a user _before_ their role is actually saved.  My backend changes here require a role is saved before systems are added, and would need a corresponding UI change if we went this route.
Because the latest role may not have been saved, the backend from the "Add Systems" window may not have current knowledge of the user's intended role. 


Screenshot 2023-03-21 at 4 28 08 PM


Another scenario that causes issues, say I am a root user and am trying to upgrade an Approver to "Viewer and Approver".  I can't add systems before clicking "Save" on the Viewer and Approver Role, because the user is still an Approver as far as the backend is concerned.

Screenshot 2023-03-21 at 3 02 29 PM

src/fides/api/ops/api/v1/endpoints/user_permission_endpoints.py
```diff
@@ -94,11 +95,22 @@ def update_user_permissions(
...
+    if user.systems and APPROVER in user.permissions.roles:  # type: ignore[attr-defined]
+        for system in user.systems.copy():
+            logger.info(
+                "Approvers cannot be system managers. Removing user {} as system manager of {}.",
+                user.id,
+                system.fides_key,
+            )
+            user.remove_as_system_manager(db, system)
+
+    return updated_user_perms
```
pattisdr, line 112: Not specified, but removing users as system managers if they are being made an Approver. This happens behind the scenes. Should we add a confirmation or alert that it's happening if we go this route?

pattisdr:
@rsilvery @allisonking @eastandwestwind this ticket feels like it would need more UI changes if we went this route. Can you weigh in on whether this is still the route we want to go and how we might update from a FE perspective?

src/fides/api/ops/api/v1/endpoints/user_permission_endpoints.py
rsilvery: Yes, I think that's reasonable. Can we file a ticket for this?

rsilvery:
@pattisdr  what questions remain here?

pattisdr:
This comment @rsilvery https://github.com/ethyca/fides/pull/2891#discussion_r1144014260 it seems like the flow needs to change a little here with roles being saved before systems can be added

src/fides/api/ops/api/v1/endpoints/user_endpoints.py
pattisdr: FE ticket here: https://github.com/ethyca/fidesplus/issues/722

src/fides/api/ops/api/v1/endpoints/user_endpoints.py
```diff
@@ -252,7 +265,7 @@ def update_managed_systems(
             user.set_as_system_manager(db, system)
 
     # Removing systems for which the user in no longer a manager
-    for system in user.systems:
+    for system in user.systems.copy():
```
eastandwestwind, line 268: is this just to be explicit that we're not modifying `user.systems`?

tests/fixtures/application_fixtures.py
```diff
@@ -1671,7 +1671,7 @@ def system_manager(db: Session, system) -> System:
     yield user
     try:
         user.remove_as_system_manager(db, system)
-    except SystemManagerException:
+    except (SystemManagerException, StaleDataError):
```
eastandwestwind, line 1674: when does a `StaleDataError` occur? And do we need to bypass this err in the actual [endpoint logic](https://github.com/ethyca/fides/pull/2891/files#diff-524e6648deb1d88b996157984be1e98af29b63a75af4943bdeaaac7c0431e319R270) too?

tests/ops/api/v1/endpoints/test_user_endpoints.py
```diff
@@ -1217,8 +1217,8 @@ def test_logout(self, db, url, api_client, generate_auth_header, oauth_client):
 
 class TestUpdateSystemsManagedByUser:
     @pytest.fixture(scope="function")
-    def url(self, user) -> str:
-        return V1_URL_PREFIX + f"/user/{user.id}/system-manager"
+    def url(self, viewer_user) -> str:
```
eastandwestwind, line 1220: glad these `user` vars are more explicit now :)

eastandwestwind (approved):
This looks good @pattisdr ! Especially appreciate the detailed testing around adding system manager to users as well as downgrading system managers to the `approver` role.

I have just a couple minor questions, nothing blocking.

tests/fixtures/application_fixtures.py
pattisdr: This is just for test fixtures - it doesn't need to be in the endpoint.  

The reason is that this test fixture tries to clean itself up after the test by removing the user as a system manager before the user deletes itself.  But this PR adds logic where we remove system managers behind the scenes if a user is downgraded to an approver.  So when the test fixture goes to clean itself up, that data is "stale".

pattisdr: If you look at several of the other test fixtures we've had to add logic like this.  where a test it's being used on is testing some sort of delete action that conflicts with the test fixture trying to perform cleanup as a follow up.

src/fides/api/ops/api/v1/endpoints/user_endpoints.py
pattisdr: Thanks for calling this out, I didn't explain this or add a test for it (but added one in the latest commit).  This is a bugfix.

Big picture, if you're looping through a list and potentially deleting elements in that list, you should copy the list first. This is because deleting an element in the list can change the index of subsequent elements, which can cause elements to get skipped that I need to delete.  I copy user.systems so I can modify them.

eastandwestwind: oh I see, I didn't think of that. Thanks for clarifying!

tests/fixtures/application_fixtures.py
eastandwestwind: This makes sense, thanks!

PR #2900, opened 2023-03-22
[Backend] Make New Users Viewers By Default [#2866]

### LINKED ISSUE #2866: Newly users are not assigned a default role
### Bug Description

When I create a user, if I don't explicitly select a role they will be created without any role at all. This prevents them from being able to login as they wind up with no scopes at all!

#### Steps to Reproduce

1. Login as a CONTRIBUTOR/OWNER
2. Navigate to "Management > Users"
3. Create a new user
image
4. Set a username and password, but do not select any permissions
5. Save

### Expected behavior

Either:
1. A new user should be created with a default role (VIEWER)
2. An error should be thrown for me to select a role

### Screenshots

Create `newuser`:
image

After saving:

After exiting this page without setting any permissions:
image

### Environment

* Version:
* OS:
* Python Version:
* Docker Version:

### Additional context

pattisdr:
Closes #2866 

### Code Changes

* [ ] When creating a brand new FidesUser, adjust the FidesUserPermissions object created in the background so it is given a VIEWER role by default.
* [ ] Update both POST and PUT {{host}}/user/{{user_id}}/permission endpoints to require an explicit "roles" key in the request to be intentional. An empty list value will remove the user's roles. (This doesn't have a practical effect in the UI - as it always has the roles key, and doesn't surface the option to remove roles entirely, but still technically allowing this in the backend).  Also note that the UI does not hit the POST user permissions endpoint at all since create user adds the initial permissions resource already.

### Steps to Confirm

* [ ] As the root user, login and navigate to user management http://localhost:3000/user-management
* [ ] Click "Add New User"
* [ ] Fill out username, firstname, last name, pw. Click Save
* [ ] Verify in the network tab that the followup request to the UI to fetch the user's permissions returns the viewer role that was automatically added by the b/e http://0.0.0.0:8080/api/v1/user/<user_id>/permission 
* [ ] Click on the Permissions tab. Verify the Viewer tab is already selected as they are technically a Viewer 

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

Give new users a Viewer role by default.  Also require that the get/create permissions endpoints specify a role key in the request (this is new behavior, it was previously given an empty list by default), and assert that an empty list will remove the users roles (this was existing behavior, just adding tests)


allisonking (approved):
ah, I'm relieved the UI picked up on the change without any problems 😌 thanks for the update!!

pattisdr:
Really appreciate all the review you've been doing for me the last couple days @allisonking 🏆

pattisdr:
Argh changelog 😠

PR #2901, opened 2023-03-22
use HTML property, not text, when dispatching Mailchimp Transactional emails

seanpreston:
### Description

This PR updates the Mailchimp Transactional adaptor to use the `html` attribute, and subsequently render html correctly in emails.

### DIFF
--- a/src/fides/api/ops/service/messaging/message_dispatch_service.py
+++ b/src/fides/api/ops/service/messaging/message_dispatch_service.py
@@ -408,6 +408,7 @@ def _mailchimp_transactional_dispatcher(
             "soft-bounce": "A temporary error occured with the target inbox. For example, this inbox could be full. See https://mailchimp.com/developer/transactional/docs/reputation-rejections/#bounces for more info.",
             "hard-bounce": "A permanent error occured with the target inbox. See https://mailchimp.com/developer/transactional/docs/reputation-rejections/#bounces for more info.",
             "recipient-domain-mismatch": f"You are not authorised to send email to this domain from {from_email}.",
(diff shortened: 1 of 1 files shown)


adamsachs (approved):
this looks good. just noting, per @seanpreston , this was tested manually by a customer. we don't have any automated integration tests with mailchimp transactional so it'd be difficult to have automated coverage for this code path at this point.

@seanpreston my only concern is related to your question before about whether we have any non-html templates. can you think of anything that would make this a bit more robust/fail more obviously in the case that we do get a non-HTML template into our code base? maybe some automated test around that?

not gonna hold this up for that though :)

seanpreston:
> @seanpreston my only concern is related to your question before about whether we have any non-html templates. can you think of anything that would make this a bit more robust/fail more obviously in the case that we do get a non-HTML template into our code base? maybe some automated test around that?

These are good questions. It's my understanding that if you pass text into `html` then that text still renders as text in the email that's sent, whereas if you pass HTML into the `text` attribute you get the HTML rendered as text. I'm just verifying.

seanpreston:
Hey @adamsachs, I can confirm what I expected is the behaviour:

Screenshot 2023-03-23 at 10 45 22

Both of those emails were sent using the `html` attribute, and the first contained only text.

adamsachs:
> Hey @adamsachs, I can confirm what I expected is the behaviour:
> 
> Screenshot 2023-03-23 at 10 45 22
> 
> Both of those emails were sent using the `html` attribute, and the first contained only text.

Thank you for the diligence @seanpreston!!! great to see that we're able to get testing done internally too.

0 caveats or reservations on getting this merged now 😄

PR #2902, opened 2023-03-22
Route for individual systems

allisonking:
Closes https://github.com/ethyca/fidesplus/issues/681

### Code Changes

* [x] Nest the previous `configure` page to be `configure/[id]` instead. Update the code to always query for the passed in system id on load
* [x] Update the tests accordingly

### Steps to Confirm

* [ ] Create some systems
* [ ] In the `/systems` page, click on the more icons and then "Edit"
* [ ] You should be brought to the same page as before, but instead of `/systems/configure` it'll be `/systems/configure/{fides_key}`
* [ ] Log out and log back in to clear the redux store (or use another browser or incognito window)
* [ ] Go directly to `/systems/configure/{fides_key}` where `fides_key` is one you have in your system
* [ ] You should hit the edit page for this system directly
* [ ] Go directly to `/systems/configure/{something_else}` where `something_else` is a fides key you do not have in your system
* [ ] You should get a message that we couldn't find that system

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

https://user-images.githubusercontent.com/24641006/227031387-18148fd4-4fe1-4eb8-8f28-26ff841fe666.mov


clients/admin-ui/cypress/e2e/systems.cy.ts
```diff
@@ -414,6 +434,11 @@ describe("System management page", () => {
     beforeEach(() => {
       stubSystemCrud();
       stubTaxonomyEntities();
+      cy.fixture("systems/systems.json").then((systems) => {
+        cy.intercept("GET", "/api/v1/system/*", {
+          body: systems[0],
+        }).as("getFidesctlSystem");
+      });
```
allisonking, line 441: note that we had different fixtures for 1️⃣  getting a list of systems vs 2️⃣  getting just one system. the tests were pretty dependent on the first system from the list of systems (1️⃣ ), so I made sure to keep that around and index into it, rather than relying on 2️⃣ which had data the tests didn't expect and would've required more changes

TheAndrewJackson (approved):
LGTM 💥  I'll update fidesplus to use this soon™️

PR #2903, opened 2023-03-22
Add Fides Plus Version to About Fides page

NevilleS:
### Code Changes

* [X] Update the API types in `admin-ui` (by running `npm run openapi:generate`)
* [X] Show "Fides Plus Version: <version>" in the About Fides page

### Steps to Confirm

* [X] Run `npm run dev` against a `fidesplus` backend
* [X] Open the About Fides page

image

### Pre-Merge Checklist

* Documentation:
* [X] Update `CHANGELOG.md`

### Description Of Changes

This is a quick little feature to aid in troubleshooting and aid in explaining how the Fides Plus server embeds & extends the core version 👍


clients/admin-ui/src/pages/management/about.tsx
```diff
@@ -36,6 +36,17 @@ const About: NextPage = () => {
           </Text>
         </Box>
 
+        {features.plusVersion && (
```
allisonking, line 39: we have a tendency to use `case ? <something> : null` instead of `case &&`. the danger with `case &&` is if `case` is a number, doing a `0 && <something>` will actually render the 0 which is unexpected! 

https://kentcdodds.com/blog/use-ternaries-rather-than-and-and-in-jsx

I prefer `&&` in a lot of cases, but many times it's easier to just stick with "always use `? : ` instead so we don't have to think about it. I think there may be eslint rules that can help with this sort of thing, though I haven't looked into it. anyway, the `&&` works perfectly fine here, just wanted to bring up the existing pattern in the code base

allisonking (approved):
:shipit:

clients/admin-ui/src/pages/management/about.tsx
NevilleS: Thanks - I'll change this. I actually paused to think about that footgun when putting this in place as I've been burnt by the same thing before but I couldn't remember exactly what the "safe" version was!

PR #2906, opened 2023-03-23
2213 custom connector template storage

galvana:
### Code Changes

* [ ] Adding `TestCustomConnectorTemplate`
* [ ] Adding `CustomConnectorTemplateLoader`
  * [ ] This includes the addition of `register_custom_functions`, there are some security considerations here*

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

These changes just add the `custom_connector_template` table and the `CustomConnectorTemplateLoader`. The endpoint to fully support this still needs to be added.


pyproject.toml
```diff
@@ -52,6 +53,7 @@ module = [
   "plotly.*",
   "pydash.*",
   "pymongo.*",
+  "RestrictedPython.*",
```
galvana, line 56: The RestrictedPython and AccessControl modules aren't types so they need to be added to the exception list

src/fides/api/main.py
```diff
@@ -273,7 +273,7 @@ async def setup_server() -> None:
     try:
         db = get_api_session()
         update_saas_configs(db)
-        logger.info("Finished loading saas templates")
+        logger.info("Finished loading SaaS templates")
```
galvana, line 276: Fixing typo

src/fides/api/ops/service/connectors/saas/connector_registry_service.py
```diff
@@ -85,9 +131,10 @@ class ConnectorRegistry:
     def get_instance(cls) -> "ConnectorRegistry":
         if cls._instance is None:
             cls._instance = cls()
-            cls._instance._templates = (
-                FileConnectorTemplateLoader().get_connector_templates()
-            )
+            cls._instance._templates = {
+                **FileConnectorTemplateLoader().get_connector_templates(),
+                **CustomConnectorTemplateLoader().get_connector_templates(),
```
galvana, line 136: Adding new loader

src/fides/api/ops/service/connectors/saas/connector_registry_service.py
```diff
@@ -102,6 +149,13 @@ def get_connector_template(cls, connector_type: str) -> Optional[ConnectorTempla
         """
         return cls.get_instance()._templates.get(connector_type)
 
+    @classmethod
+    def register_template(
```
galvana, line 153: To be called by the endpoint, which still needs to be implemented

src/fides/api/ops/service/saas_request/override_implementations/authentication_strategy_adobe_campaign.py
```diff
@@ -55,7 +55,7 @@ def add_authentication(
         access_token: Optional[str] = secrets.get("access_token")
         expires_at: Optional[int] = secrets.get("expires_at")
 
-        if not access_token or self._close_to_expiration(expires_at, connection_config):
+        if not access_token or self.close_to_expiration(expires_at, connection_config):
```
galvana, line 58: Underscores at the start of variable names aren't allowed by RestrictedPython. The custom functions under `override_implementations` aren't loaded with RestrictedPython right now but I just want to make sure we can at some point if needed

tests/fixtures/saas/test_data/custom/custom_functions.py
galvana, line 1: Adding a request_override and custom strategy to make sure both types of custom code are loaded correctly.

src/fides/api/ops/service/connectors/saas/connector_registry_service.py
```diff
@@ -235,3 +289,74 @@ def update_saas_instance(
     connection_config.update_saas_config(db, SaaSConfig(**config_from_template))
 
     upsert_dataset_config_from_template(db, connection_config, template, template_vals)
+
+
+def register_custom_functions(script: str) -> None:
```
galvana, line 294: Uses a combination of RestrictedPython and ZopeGuards (both from the Zope Foundation) to load custom code in a restricted environment. The restrictions work by limiting which Python functions (builtins) the custom script has access to. The out-of-the-box settings are too restrictive so we had to make a few exceptions:

- Allowing the use of type annotations (via CustomRestrictingNodeTransformer)
- Allowing all imports by default and excluding some unsafe modules (custom_guarded_import)
- Allowing `getitem`
- Allowing `staticmethod`

src/fides/api/ops/service/connectors/saas/connector_registry_service.py
galvana: Connector templates defined in the database will override connectors bundled with Fides (by design)

src/fides/api/ops/service/connectors/saas/connector_registry_service.py
adamsachs: OK, this makes sense. but i get a little bit concerned about ensuring the new template here gets persisted to the DB. i'm assuming the endpoint would couple that persistence alongside this dynamic "runtime" update of the template registry. but that still feels a bit fragile to me, i wonder if we can come up with a way to more tightly couple the dynamic registration here with the persistence to the DB, to prevent any future code updates (changes to the endpoint, new endpoints, new mechanisms of dynamic registration) from accidentally performing the dynamic registration without persisting the template to the db.

i do see why it's not natural to couple these two together in the way you've got things structured now - the db table is associated with one of potentially many loader implementations that the `ConnectorRegistry` uses, while your dynamic template registration is tied to the broader `ConnectorRegistry`.

it doesn't necessarily need to be resolved in this PR, and i generally think what you've arrived at here is nice. one idea i have is maybe to provide a hook (method) on your `ConnectorTemplateLoader` ABC that is responsible/tied to the dynamic template registration, and for the `CustomConnectorTemplateLoader`, that method would handle both updating the runtime registry while also persisting the update to the db. you could then hook into that method via some broader `register_template` here on the `ConnectorRegistry`, maybe taking the specific `ConnectorTemplateLoader` as an argument.

i don't know if that's cleaner. also, don't let it hold up this PR. just some food for thought - that it'd be nice if we could ensure that the db updates to persist the new template accompany any dynamic update to the registry.

src/fides/api/ops/service/connectors/saas/connector_registry_service.py
```diff
@@ -75,6 +83,44 @@ def get_connector_templates(self) -> Dict[str, ConnectorTemplate]:
...
+class CustomConnectorTemplateLoader(ConnectorTemplateLoader):
+    """
+    Loads custom connector templates defined in the custom_connector_template database table.
+    """
+
+    def __init__(self) -> None:
+        logger.info("Loading connectors templates from the database")
+        self.templates: Dict[str, ConnectorTemplate] = {}
+        for template in CustomConnectorTemplate.all(db=get_api_session()):
+            try:
```
adamsachs, line 95: the checks/error handling around template instantiation/registration that are done here are not being invoked in the dynamic registration. i know the already instantiated `ConnectorTemplate` is being passed into the dynamic registration method, so it may not belong in that method specifically, but i do think we need to abstract out this instantiation/initialization logic here away from the `CustomConnectorTemplateLoader` constructor to ensure that we are running through the same checks whether the template is loaded via a persisted db record or via the dynamic registration.

tests/ops/service/connectors/test_connector_template_loaders.py
```diff
@@ -30,3 +47,176 @@ def test_file_connector_template_loader_connector_not_found(self):
...
+            name="Custom",
+            config=custom_config,
+            dataset=custom_dataset,
+            icon=custom_icon,
+            functions=custom_functions,
+        )
+        custom_template.save(db=db)
+
+        # load custom connector templates from the database
+        loader = CustomConnectorTemplateLoader()
```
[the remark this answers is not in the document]
adamsachs: ok, looking a bit closer at the rest of the PR i realize that:

- we do _log_ an error here at load time
- this also isn't the code path that will be invoked for the actual dynamic registration of templates via an endpoint, this is just loading their persisted records from the db the next application bootup

so i retract most of my comment above - i think the UX is OK for this case. that being said, i do have other concerns i raised (in separate comments) about ensuring our dynamic registration path goes thru the same "validations" that we do here in the app bootstrap path. if/when we add the validation in to that path (which i think is a _must_), then in _those_ cases, i do think we need to fail fast and report that to the user immediately. i.e., if a user tries to upload a saas template with a custom function while that feature is disabled via config property, that operation should completely fail and the reason should reported clearly to the user. 

while i know your code here in this PR hasn't really gotten to that point yet, i do think what you've got here will need to be refactored a bit to support that UX, so it may be prudent to address it now.

adamsachs (commented):
great work @galvana, this is very exciting! i don't think this increment is far off from being ready, and i like the care you've taken around balancing security with usability.

there are a number of nits. the only more substantive thing i've got a bit of a concern with is making sure this will fit in smoothly with the dynamic registration code path/UX. let me know if my comments around that piece aren't clear, as i've scattered my thoughts a bit in different areas of the code changes 😄

tests/fixtures/saas_example_fixtures.py
```diff
@@ -621,3 +626,25 @@ def erasure_policy_complete_mask(
...
+
+
+@pytest.fixture
+def custom_icon() -> str:
+    return encode_file_contents("tests/fixtures/saas/test_data/custom/custom.svg")
+
+
+@pytest.fixture
+def custom_functions() -> str:
+    return load_as_string("tests/fixtures/saas/test_data/custom/custom_functions.py")
```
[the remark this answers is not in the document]
galvana: That's correct

tests/fixtures/saas/test_data/custom/custom_functions.py
```diff
@@ -0,0 +1,73 @@
...
+    secret_knock: str
+    secret_handshake: str
+
+
+class CustomAuthenticationStrategy(AuthenticationStrategy):
+    """
+    Adds the secrets to the request
+    """
+
+    name = "custom"
```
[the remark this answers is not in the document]
galvana: Better examples coming right up!

tests/fixtures/saas/test_data/planet_express/planet_express.svg
galvana, line 1: This image is in the public domain, I checked!

src/fides/api/ops/service/connectors/saas/connector_registry_service.py
```diff
@@ -232,3 +286,74 @@ def update_saas_instance(
...
+        "sys",
+        "subprocess",
+        "shutil",
+        "socket",
+        "importlib",
+        "tempfile",
+        "glob",
+    ]:
+        # raising SyntaxError to be consistent with exceptions thrown from other guarded functions
+        raise SyntaxError(f"Import of '{name}' module is not allowed.")
```
adamsachs, line 356: a bit of a nit, but i do think it would be good to get some test coverage on this path. i'm not just saying it to be pedantic - it's an important security feature you've added

adamsachs (approved):
alright, thank you for the updates! @galvana and i spoke offline and for the more substantive feedback about some refactoring of template loading, he will address that in the subsequent PR that adds the endpoint support for loading custom connectors, which seems appropriate.

i'm good where this is, so i'll approve. there is one codecov miss that i commented on, which i do think would be good to get test coverage on, if possible.

PR #2913, opened 2023-03-24
722 tie user role and system save states together

eastandwestwind:
Closes https://github.com/ethyca/fidesplus/issues/722

### Code Changes

* [ ] Tie together save states for user role and system. This change prevents changes to system from auto-saving on the BE, and instead saves any changes to systems at the same time as the overall user permissions form.

### Steps to Confirm

* [ ] Clicking this `confirm` button when assigning systems should not save to the backend. Hard refresh should show no changes to user permissions:
Screenshot 2023-03-24 at 5 17 35 PM

* [ ] Clicking the `confirm` button to assign systems, then hitting the `save` button on main form should save changes

* [ ] Clicking the trash icon should not save to the backend (until you click `save` on the main form)
Screenshot 2023-03-24 at 5 17 45 PM

* [ ] Regression test that changing user role saves on the backend, after clicking `save` button.

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

### DIFF
--- a/clients/admin-ui/cypress/e2e/user-management.cy.ts
+++ b/clients/admin-ui/cypress/e2e/user-management.cy.ts
@@ -258,14 +258,11 @@ describe("User management", () => {
           cy.getByTestId("row-fidesctl_system").within(() => {
             cy.getByTestId("unassign-btn").click();
           });
-          cy.getByTestId("remove-fidesctl_system-confirmation-modal").within(
-            () => {
-              cy.getByTestId("continue-btn").click({ force: true });
-            }
-          );
-          cy.wait("@removeUserManagedSystem").then((interception) => {
-            const { url } = interception.request;
-            expect(url).contains("fidesctl_system");
+          cy.getByTestId("save-btn").click();
+
+          cy.wait("@updateUserManagedSystems").then((interception) => {
+            const { body } = interception.request;
+            expect(body).to.eql([]);
           });
         });
       });
@@ -292,6 +289,7 @@ describe("User management", () => {
             });
           });
           cy.getByTestId("confirm-btn").click();
+          cy.getByTestId("save-btn").click();
 
           cy.wait("@updateUserManagedSystems").then((interception) => {
             const { body } = interception.request;
@@ -331,6 +329,7 @@ describe("User management", () => {
           });
 
           cy.getByTestId("confirm-btn").click();
+          cy.getByTestId("save-btn").click();
           cy.wait("@updateUserManagedSystems").then((interception) => {
             const { body } = interception.request;
             expect(body).to.eql([]);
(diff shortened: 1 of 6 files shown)


clients/admin-ui/src/features/user-management/AssignSystemsTable.tsx
```diff
@@ -5,63 +5,45 @@ import {
...
     skip: !activeUserId,
   });
-  const assignedSystems = useAppSelector(selectActiveUsersManagedSystems);
-  const toast = useToast();
-  const [removeUserManagedSystemTrigger] = useRemoveUserManagedSystemMutation();
-  const handleDelete = async (system: System) => {
-    if (!activeUserId) {
-      return;
-    }
-    const result = await removeUserManagedSystemTrigger({
```
allisonking, line 49: it's nice that this component becomes a lot simpler 😎

clients/admin-ui/src/features/user-management/PermissionsForm.tsx
```diff
@@ -126,7 +154,7 @@ const PermissionsForm = () => {
                 colorScheme="primary"
                 type="submit"
                 isLoading={isSubmitting}
-                disabled={!dirty}
+                disabled={!dirty && assignedSystems === initialManagedSystems}
```
allisonking, line 157: oh when did this case come up?

clients/admin-ui/cypress/e2e/user-management.cy.ts
```diff
@@ -258,14 +258,11 @@ describe("User management", () => {
...
-            }
-          );
-          cy.wait("@removeUserManagedSystem").then((interception) => {
-            const { url } = interception.request;
-            expect(url).contains("fidesctl_system");
+          cy.getByTestId("save-btn").click();
+
+          cy.wait("@updateUserManagedSystems").then((interception) => {
+            const { body } = interception.request;
+            expect(body).to.eql([]);
```
allisonking, line 265: nice, I think we should take this test suite a little farther too
1. Let's remove the intercept of `removeUserManagedSystem` since we're no longer calling that endpoint and don't need it
2. for the tests under `in modal`, when we do a `cy.wait("@updateUserManagedSystems")`, let's also, right before it, do a `cy.wait("@updatePermission")` so that we can verify both calls are firing. alternatively, can wait for `@updateUserManagedSystems` in the test "can assign a role to a user"—just want a test that shows both calls firing when the save button is clicked.

allisonking (approved):
this looks great, nice work! I played around and couldn't break it 😄  just a cypress test request :) 

@Roger-Ethyca will want to know about this new behavior too, so tagging him.

clients/admin-ui/cypress/e2e/user-management.cy.ts
allisonking: ah okay, so I think this test starts off with 
image

and we remove `fidesctl_system`. the update request should still have the two systems, so let's change the expected payload from `[]` to probably `["demo_analytics_system", "demo_marketing_system"]`

allisonking (changes requested):
taking back my approval just to note that we need to fix the failing test :)

allisonking (approved):
👏

Issue #2898, opened 2023-03-22
[Frontend] Alert Before System Manager Downgraded to Approver


pattisdr:
### Is your feature request related to a specific problem?

https://github.com/ethyca/fides/pull/2891#discussion_r1144015038

New backend work removes user as system manager if their role is downgrade to Approver behind the scenes.  This behavior should be made obvious to the frontend user.

### Describe the solution you'd like

Something like detecting if a user is a system manager and we're about to make them an approver on the frontend.  Perhaps add a confirmation modal that they are about to be removed as system manager. (There may be better frontend flows, this is just anidea)

### Describe alternatives you've considered, if any

_A description of any alternative solutions or features you've considered._

### Additional context

_Add any other context or screenshots about the feature request here._

rsilvery:
Let's say: "Switching to an approver role will remove all assigned systems. Do you wish to proceed?"

rsilvery:
@allisonking can you help with this?

allisonking:
oh, I think [I might have caught this too](https://github.com/ethyca/fides/pull/2953) and fixed it. let me double check

allisonking:
yup yup, this looks fixed by https://github.com/ethyca/fides/pull/2953

pattisdr:
awesome 🔥

Roger-Ethyca:
wahoo! moving to done

PR #2924, opened 2023-03-27
2898 alert system manager downgrade to approver

eastandwestwind:
Closes https://github.com/ethyca/fides/issues/2898

### Code Changes

* [ ] Adds confirmation modal when attempting to downgrade user to approver if they are assigned systems.

### Steps to Confirm

* [ ] Assign user as anything other than approver, and assign systems
* [ ] Click save
* [ ] Assign user to approver
* [ ] Click save
* [ ] See confirmation modal
* [ ] Confirm nothing happens if you click `cancel`
* [ ] Confirm user is saved with approver role if you click `yes`

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,8 @@ The types of changes are:
 * Removing payment types from Stripe connector params [#2915](https://github.com/ethyca/fides/pull/2915)
 * Change Privacy Center email placeholder text [#2935](https://github.com/ethyca/fides/pull/2935)
 * Restricted setting Approvers as System Managers [#2891](https://github.com/ethyca/fides/pull/2891)
+* Adds confirmation modal when downgrading user to "approver" role via Admin UI [#2924](https://github.com/ethyca/fides/pull/2924)
+
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


allisonking (approved):
this lgtm—the UX is a little strange right now while Dawn's PR isn't in yet, so we'll just want to test that everything feels intuitive after that PR goes in as well (or create a branch with both this change and Dawn's change just to test things out)

i.e. right now the systems are retained


https://user-images.githubusercontent.com/24641006/228259169-05d94569-92d0-43ba-9610-7f2f83811a93.mov

I think that should just be because the backend isn't getting rid of them, but there's always the chance the FE is keeping the state somewhere, so we'll want to watch for that.

eastandwestwind:
This err was being thrown when attempting to downgrade a user from viewer to approver (when the user has systems assigned):

Screenshot 2023-03-29 at 8 01 55 PM

Since the removal of systems occurs on the BE, we shouldn't call the PUT ``user/${userId}/system-manager` API endpoint when users are downgraded to approver. I've updated the code accordingly.

eastandwestwind:
> this lgtm—the UX is a little strange right now while Dawn's PR isn't in yet, so we'll just want to test that everything feels intuitive after that PR goes in as well (or create a branch with both this change and Dawn's change just to test things out)
> 
> i.e. right now the systems are retained
> 
> I think that should just be because the backend isn't getting rid of them, but there's always the chance the FE is keeping the state somewhere, so we'll want to watch for that.

I've also confirmed that this issue is now fixed after pulling in latest BE

allisonking (approved):
good catch! glad it's all working nicely 💥

PR #2928, opened 2023-03-27
initial implementation of privacy notice CRUD endpoints

### LINKED ISSUE #2832: Backend Privacy Notice Configuration

Allow a Compliance Person to create, update, retrieve, and disable Privacy Notices for their organization.  These will eventually be displayed in the Privacy Center to a user based on their determined geography instead of defining these in a config.json file.

[See Design Doc: Privacy Notice Configuration](https://ethyca.atlassian.net/wiki/spaces/EN/pages/2636382212/2023-03-10+Backend+Design+Enhancing+Consent+Management+Capabilities+and+Consent+Reporting#Privacy-Notice-Configuration)

### Additional context

adamsachs:
❗ Contains migration; check if downrev needs to be bumped before merging.

Closes #2832 

### Code Changes
* new `privacynotice` and `privacynoticehistory` DB tables (and corresponding SQL alchemy classes) to store details about privacy notices, and an "audit trail" for each privacy notice record, respectively
* new `privacy_notice:[create|update|read\` scopes added to our scope registry
    * `owner` and `contributor` roles have all these scopes
    * `viewer` (and `viewer+approver`) roles have `privacy_notice:read` scope
* `GET /api/v1/privacy-notice` endpoint to retrieve a list of privacy notices on the system
    * `show_disabled` param (optional, default =`True`) determines whether disabled privacy notice records should be returned
    * `region` param (optional, single-value) allows filtering of privacy notices to show only those associated with the given region
    * `systems_applicable` param (optional, default=`True`) determines whether to return only privacy notices that have a data use that is associated with a system
* `GET /api/v1/privacy-notice/{privacy_notice_id}` endpoint to retrieve details about a specific privacy notice
* `POST /api/v1/privacy-notice` to create one or more new privacy notice
    * ensures that only valid data uses are provided, and also ensures that the update causes no data use/region conflicts. specifically, for any region, there cannot be multiple privacy notices associated with that region that have an overlapping data use (data uses are hierarchical, so e.g. the data uses of `advertisting` and `advertising.first_party` are considered overlapping)
    * if the above validations fail for _any_ new privacy notice specified, an error is thrown and the entire operation is void
* `PATCH /api/v1/privacy-notice` to update details about one or more new privacy notices
    * an `id` must be provided on each object in the request body, to indicate which privacy notice the update applies to. the `id` must point to a valid, existing privacy notice `id`
    * similar validation behavior on the updated data as the endpoint above

### Steps to Confirm

* nothing specific, but spinning up your local fides environment in any normal way should expose these endpoints and allow for some manual testing

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### Description Of Changes

Foundational APIs for creating, reading, updating privacy notices (and their history/audit records) on Fides. I tried to stick as close to the (very thorough and clear!) [spec](https://ethyca.atlassian.net/wiki/spaces/EN/pages/2636382212/2023-03-10+Backend+Design+Enhancing+Consent+Management+Capabilities+and+Consent+Reporting) in confluence.


src/fides/api/ops/api/v1/endpoints/privacy_notice_endpoints.py
```diff
@@ -0,0 +1,218 @@
...
+    Includes some query params to help filter the list if needed
+    """
+    logger.info("Finding all PrivacyNotices with pagination params '{}'", params)
+    notice_query = db.query(PrivacyNotice)
+    if not show_disabled:
+        notice_query = notice_query.filter(PrivacyNotice.disabled.is_(False))
+    if region is not None:
+        notice_query = notice_query.filter(PrivacyNotice.regions.contains([region]))
+    if systems_applicable:
+        data_uses = System.get_system_data_uses(db)
```
adamsachs, line 61: i didn't love having to execute 2 queries here, but the join query here was becoming quite complex for me, specifically because the data uses are string fields within within a JSON blob column (that's actually an array) on the `System` table.

rather than continue down that rabbit hole, this was a lot simpler for implementation, and we're still issuing a constant (2) number of queries (not N), so i felt this was fine.

src/fides/api/ops/api/v1/endpoints/privacy_notice_endpoints.py
```diff
@@ -0,0 +1,218 @@
...
+
+    return [
+        PrivacyNotice.create(
+            db=db, data=privacy_notice.dict(exclude_unset=True), check_name=False
+        )
+        for privacy_notice in privacy_notices
+    ]
+
+
+def prepare_privacy_notice_patches(
```
adamsachs, line 143: the method implementation here isn't quite as readable as i'd like. it also iterates through the privacy notice updates more than once, so there may be some slight inefficiencies. but none of the operations on those iterations should be expensive (just in-memory CPU operations, nothing going against the db, for example), and there seemed to be different efficiency/readability tradeoffs with the other approaches i could think of. so this is what i landed on, but definitely open to suggestions!

src/fides/api/ops/models/privacy_notice.py
```diff
@@ -0,0 +1,191 @@
...
+from fides.api.ops.common_exceptions import ValidationError
+from fides.lib.db.base_class import Base, FidesBase
+
+
+class PrivacyNoticeRegion(Enum):
+    """
+    Enum is not formalized in the DB because it is subject to frequent change
+    """
+
+    us_ca = "us_ca"
```
adamsachs, line 22: probably want to double-check this before we merge?

src/fides/api/ops/models/privacy_notice.py
```diff
@@ -0,0 +1,191 @@
...
+            "displayed_in_privacy_center": self.displayed_in_privacy_center,
+            "displayed_in_banner": self.displayed_in_banner,
+            "displayed_in_privacy_modal": self.displayed_in_privacy_modal,
+            "privacy_notice_id": self.id,
+        }
+        PrivacyNoticeHistory.create(db, data=history_data, check_name=False)
+
+        # on any update to a privacy notice record, its version must be incremented
+        # version gets incremented by a full integer, i.e. 1.0 -> 2.0 -> 3.0
+        data["version"] = float(self.version) + 1.0
```
adamsachs, line 104: any other ideas for the increment logic here?

src/fides/api/ops/models/privacy_notice.py
```diff
@@ -0,0 +1,191 @@
...
+
+        # remove protected fields from the cloned dict
+        cloned_attributes.pop("_sa_instance_state")
+
+        # create a new object with the updated attribute data to keep this
+        # ORM object (i.e., `self`) pristine
+        return PrivacyNotice(**cloned_attributes)
+
+
+def check_conflicting_data_uses(
```
adamsachs, line 127: hopefully this method is at least easy to follow. there may be some optimizations to make. i'd initially tried to create a 2d/nested dict with both regions and data_uses to allow for indexing on the data_uses for a given region (constant time lookups rather than linear, i.e. iterating thru all data uses for a given region), but ran into issues given that we need to do the "hierarchical" comparison of data uses.

we could probably create some sort of hierarchical data structure that allowed for lookups, but that seemed to be too much complexity for runtime optimizations that i'm assuming are not going to have a significant impact, at this point.

but there may be other improvements here i haven't seen - my main focus was getting something in place that worked and wasn't too painful to follow!

src/fides/api/ops/schemas/privacy_notice.py
```diff
@@ -0,0 +1,105 @@
...
+        """
+        Ensures that all the provided `PrivacyNotice`s have valid data uses.
+        Raises a 422 HTTP exception if an unknown data use is found on any `PrivacyNotice`
+        """
+        valid_data_uses = [data_use.fides_key for data_use in DataUse.query(db).all()]
+        try:
+            for privacy_notice in privacy_notices:
+                privacy_notice.validate_data_uses(valid_data_uses)
+        except ValueError as e:
+            raise HTTPException(HTTP_422_UNPROCESSABLE_ENTITY, detail=str(e))
```
adamsachs, line 72: wasn't sure if it was appropriate to throw an `HTTPException` from within the pydantic schema, but this class is pretty specifically used for API modeling so i figured it may be OK? it did make the endpoint-level code cleaner/simpler to have the `HTTPException` thrown from within this class method.

src/fides/lib/db/base_class.py
```diff
@@ -151,7 +151,9 @@ def filter(
         return db.query(cls).filter(conditions)
 
     @classmethod
-    def create(cls: Type[T], db: Session, *, data: dict[str, Any]) -> T:
+    def create(
+        cls: Type[T], db: Session, *, data: dict[str, Any], check_name: bool = True
```
adamsachs, line 155: i didn't love having to change this core method, but i am hoping my changes here shouldn't be disruptive in any way, given the default value. automated testing seems to back that hope up.

i'd started on trying to work around this constraint but everything just felt very contrived. i do think adding the optionality here is ultimately an improvement, i don't think we need to be strict in requiring all of our classes/tables to have their `name` column in a unique constraint.

tests/conftest.py
```diff
@@ -966,3 +966,12 @@ def system_manager_client(db, system):
     db.refresh(client)
     yield client
     client.delete(db)
+
+
+@pytest.fixture(scope="function")
+def load_default_data_uses(db):
```
adamsachs, line 972: i didn't dig into this too much, but i needed to put this in to stop our scaffolding from holding me up in my testing. if this feels too hacky, i can dig into this further now that other pieces are in place

src/fides/api/ops/models/privacy_notice.py
```diff
@@ -0,0 +1,191 @@
...
+    enforcement_level = Column(EnumColumn(EnforcementLevel), nullable=False)
+    version = Column(Float, nullable=False, default=1.0)
+    disabled = Column(Boolean, nullable=False, default=False)
+    has_gpc_flag = Column(Boolean, nullable=False, default=False)
+    displayed_in_privacy_center = Column(Boolean, nullable=False, default=True)
+    displayed_in_banner = Column(Boolean, nullable=False, default=True)
+    displayed_in_privacy_modal = Column(Boolean, nullable=False, default=True)
+
+
+class PrivacyNotice(PrivacyNoticeBase, Base):
```
adamsachs, line 70: was thinking i should probably add some unit tests specifically around the methods in this class? my general endpoint testing does seem to provide good coverage (at least codecov isn't complaining), but those are more "end to end" and it may be better to break down some of the testing to target these components specifically...

pattisdr:
Starting review now -

src/fides/lib/db/base_class.py
pattisdr: I agree - this was more for those models like `Policy` which have a readable `key` created for it. Part of the uniqueness on the name was because if there was no key supplied in the request we could create it from the name.  And then there were more models that we wanted this same behavior on. 

But a lot of our new models don't have this additional human readable key added to it, so I like this addition.  It's now feeling like this `name`/`key` behavior should be organized into it's own mixin just used on certain models instead of it being the default behavior. Not today's concern though.

pattisdr (commented):
I think this is great work @adamsachs: your code was clear and easy to follow, and you'd taken time to add good code comments where why certain logic was added was necessary.  The `check_conflicting_data_uses` was the main place that felt a little tricky to trace through but you documented it well. I'd suggest adding some tests on that method specifically and make sure all edge cases are covered. 

 My comments are minor.

docs/fides/docs/development/postman/Fides.postman_collection.json
```diff
@@ -5114,6 +5114,172 @@
 					"response": []
 				}
 			]
+		},
+		{
+			"name": "Privacy Notices",
+			"item": [
+				{
+					"name": "Create Privacy Notices",
```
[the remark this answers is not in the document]
adamsachs: ah right, good catch, thank you!

src/fides/api/ops/models/privacy_notice.py
```diff
@@ -0,0 +1,191 @@
...
+                    # an existing DataUse
+                    if existing_use.startswith(data_use) or data_use.startswith(
+                        existing_use
+                    ):
+                        raise ValidationError(
+                            message=f"Privacy Notice '{notice_name}' has already assigned data use '{existing_use}' to region '{region}'"
+                        )
+            # add the data use to our map, to effectively include it in validation against the
+            # following incoming records
+            region_uses.append((data_use, privacy_notice.name))
```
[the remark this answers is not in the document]
adamsachs: ah yes, good catch! i'm a little bit surprised/worried things were working without this...

src/fides/api/ctl/sql_models.py
```diff
@@ -290,6 +293,18 @@ class System(Base, FidesBase):
...
+    @classmethod
+    def get_system_data_uses(cls: Type[System], db: Session):
+        """
+        Utility method to get all data uses associated with a System instance
+        """
+        data_uses = set()
+        for row in db.execute(select(cls.privacy_declarations)).all():
+            declarations: List[dict[str, Any]] = row[0]
+            for declaration in declarations:
+                data_uses.add(declaration["data_use"])
```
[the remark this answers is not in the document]
adamsachs: i don't really know the answer, so yes, let's just play it safe. thanks for catching!

src/fides/api/ctl/sql_models.py
```diff
@@ -290,6 +293,18 @@ class System(Base, FidesBase):
         "FidesUser", secondary="systemmanager", back_populates="systems"
     )
 
+    @classmethod
+    def get_system_data_uses(cls: Type[System], db: Session):
+        """
+        Utility method to get all data uses associated with a System instance
```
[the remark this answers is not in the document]
adamsachs: whoops - thanks for catching!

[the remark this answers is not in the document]
adamsachs: well, actually, i think i just worded this poorly. let me see if i can make it clearer

src/fides/api/ctl/sql_models.py
```diff
@@ -290,6 +293,18 @@ class System(Base, FidesBase):
...
+        """
+        data_uses = set()
+        for row in db.execute(select(cls.privacy_declarations)).all():
```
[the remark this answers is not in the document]
adamsachs: thanks - definitely still understanding the right patterns here...

src/fides/api/ops/api/v1/endpoints/privacy_notice_endpoints.py
```diff
@@ -0,0 +1,218 @@
...
+        Security(verify_oauth_client, scopes=[scope_registry.PRIVACY_NOTICE_READ])
+    ],
+)
+def get_privacy_notice_list(
+    *,
+    db: Session = Depends(deps.get_db),
+    params: Params = Depends(),
+    show_disabled: Optional[bool] = True,
+    region: Optional[PrivacyNoticeRegion] = None,
+    systems_applicable: Optional[bool] = True,
```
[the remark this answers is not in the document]
adamsachs: haha yeah, that was actually my instinct, i think i took your note in the design doc a bit too literally - you gave this as an example `GET /api/v1/privacy_notice?region=us_ca&show_disabled=true&systems_applicable=true` so i'd assumed that was what you were suggesting as a default...but realizing now that was a silly assumption.

anyway, agreed that a default of `false` here is more sensible!

src/fides/api/ops/models/privacy_notice.py
```diff
@@ -0,0 +1,191 @@
...
+            "enforcement_level": self.enforcement_level,
+            "version": self.version,
+            "disabled": self.disabled,
+            "has_gpc_flag": self.has_gpc_flag,
+            "displayed_in_privacy_center": self.displayed_in_privacy_center,
+            "displayed_in_banner": self.displayed_in_banner,
+            "displayed_in_privacy_modal": self.displayed_in_privacy_modal,
+            "privacy_notice_id": self.id,
+        }
+        PrivacyNoticeHistory.create(db, data=history_data, check_name=False)
```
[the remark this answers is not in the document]
adamsachs: interesting, that does seem like a reasonable thing to consider, and i don't really see any immediate downsides. do you remember why we moved away from that? i don't think i was involved at that point, but maybe i'm just forgetting 😄

src/fides/api/ops/api/v1/endpoints/privacy_notice_endpoints.py
adamsachs: yeah, i was very happy when i found overlap - it's a nice little function!

src/fides/lib/models/fides_user.py
```diff
@@ -79,6 +81,7 @@ def create(cls, db: Session, data: dict[str, Any]) -> FidesUser:
                 "first_name": data.get("first_name"),
                 "last_name": data.get("last_name"),
             },
+            check_name=check_name,
```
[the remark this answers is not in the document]
adamsachs: yeah, at least i _should_, though i think python is permissive enough where this doesn't actually have an impact at runtime. i'm surprised these weren't caught by the type checker or linter, though. it was caught for this class, but not others what you're pointing out.

let me go through a bit more thoroughly with text search rather than relying on the static analysis tools to help me out :)

src/fides/api/ops/models/privacy_notice.py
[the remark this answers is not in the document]
pattisdr: Ah, this was just how I'd laid it out in the design doc, I had planned to have the PrivacyNoticeHistory table have every version and you'd gone a different direction here, which is definitely fine, I didn't have a very strong opinion here.  I assumed you had reasons for implementing it differently.

src/fides/lib/models/fides_user.py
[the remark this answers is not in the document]
adamsachs: those seem to be the only other two files that were missed, as far as i can tell (2 classes `Rule` and `RuleTarget` within `Policy.py`). 

i'll make those updates, thanks for catching this!

src/fides/api/ops/models/privacy_notice.py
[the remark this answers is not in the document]
adamsachs: oh i see - i'd missed that, whoops! i do think it's probably more comprehensive/cleaner. now that i'm getting some more test coverage in place, i'd feel a bit better about making this update rather "last minute", so let me give it a go.

agreed it's probably not a _must have_, but i do like it better, so it feels like better to do it now, earlier rather than later.

src/fides/api/ops/models/privacy_notice.py
```diff
@@ -0,0 +1,191 @@
...
+
+        # on any update to a privacy notice record, its version must be incremented
+        # version gets incremented by a full integer, i.e. 1.0 -> 2.0 -> 3.0
+        data["version"] = float(self.version) + 1.0
+        return super().update(db=db, data=data)  # type: ignore
```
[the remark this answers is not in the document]
adamsachs: thank you for the suggestion here! with some slight adjustments, this seems to be working well 👍

adamsachs:
@pattisdr i think this is good for another look, hopefully we can get this merged today before the release train leaves the station!

let me know if there's anything else you think i need to address. since your last review, i mainly added test coverage and i also adjusted to create a history entry during privacy notice creation/with each update, not "lagging behind" as i'd had it initially. we can easily revert that if you see problems with the rather hasty adjustment there.

src/fides/api/ops/models/privacy_notice.py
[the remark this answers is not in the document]
pattisdr: ah this looks good

src/fides/api/ops/models/privacy_notice.py
```diff
@@ -0,0 +1,216 @@
...
+    accepts or rejects to indicate their consent for particular data uses
+    """
+
+    @classmethod
+    def create(
+        cls: Type[PrivacyNotice],
+        db: Session,
+        *,
+        data: dict[str, Any],
+        check_name: bool = True,
```
pattisdr, line 82: Seems like the default here should be false?

pattisdr (commented):
Thanks for this huge amount of work - this is a great foundation that really sets us up for the rest of the backend consent changes.  Really appreciate the extra care that went into the testing on this one. Just a couple of things and we're good to go.

src/fides/api/ctl/migrations/versions/6d6b0b7cbb36_add_privacynotice_tables.py
```diff
@@ -0,0 +1,122 @@
...
+Revises: 8615ac51e78b
+Create Date: 2023-03-22 12:09:47.395190
+
+"""
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "6d6b0b7cbb36"
+down_revision = "8615ac51e78b"
```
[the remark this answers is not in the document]
adamsachs: hmm no i think this is correct, i had to update this morning due to the migration added [here](https://github.com/ethyca/fides/pull/2906/files#diff-965f0fa50feac3d6510aa833a7570a9e50a5bc68ee8c81af4a2137ad086c0980) -- let me know if that still looks off to you @pattisdr  :)

(tests were indeed failing before i pushed up this change earlier)

src/fides/api/ops/models/privacy_notice.py
adamsachs: yes! thank you, good catch

src/fides/api/ctl/migrations/versions/6d6b0b7cbb36_add_privacynotice_tables.py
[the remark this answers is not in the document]
pattisdr: But that migration is not on this branch @adamsachs - you should get the branch updated with main then.  Checking this branch out and testing in isolation you can't even bring up the application because of it.

pattisdr (approved):

src/fides/api/ctl/migrations/versions/6d6b0b7cbb36_add_privacynotice_tables.py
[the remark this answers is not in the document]
adamsachs: oh, good point. i think it _would_ be OK after merging (since that's what the tests are evaluating), but you're right, definitely cleaner to get up to date with `main`. thanks for pointing out.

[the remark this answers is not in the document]
adamsachs: OK, this should be good now that i've rebased. thanks for the diligence @pattisdr , definitely can feel better about merging with this rebased off latest in `main`!

src/fides/api/ops/models/privacy_notice.py
mfbrown: @adamsachs to cover ourselves and make things easier going forward - can we add a list item for each EU country? Reason being is that per Recital 10 of the GDPR:
> Member States should be allowed to maintain or introduce national provisions to further specify the application of the rules of this Regulation. In conjunction with the general and horizontal law on data protection implementing Directive 95/46/EC, Member States have several sector-specific laws in areas that need more specific provisions. This Regulation also provides a margin of manoeuvre for Member States to specify its rules, including for the processing of special categories of personal data (‘sensitive data’). To that extent, this Regulation does not exclude Member State law that sets out the circumstances for specific processing situations, including determining more precisely the conditions under which the processing of personal data is lawful.

In fact, we may have to account for Germany individually already. This could result in list items like `eu_de` and `eu_nl` for example.

adamsachs: OK - that works for me @mfbrown . just to be clear, we will _not_ consider the EU member countries as overlapping with the broader `eu` region, though. if that's needed moving forward, we could look to add in that logic - but it feels out of scope for this initial body of work IMO.

i'm going to use [this](https://ec.europa.eu/eurostat/statistics-explained/index.php?title=Glossary:Country_codes) as a reference for country codes to support (e.g. `eu_nl`), hope that's ok? seems somewhat official :) if there's a better reference, please let me know!

src/fides/api/ops/models/privacy_notice.py
```diff
@@ -0,0 +1,242 @@
...
+class PrivacyNoticeRegion(Enum):
+    """
+    Enum is not formalized in the DB because it is subject to frequent change
+    """
+
+    us_ca = "us_ca"  # california
+    us_co = "us_co"  # colorado
+    us_va = "us_va"  # virginia
+    us_ut = "us_ut"  # utah
+    eu_be = "eu_be"  # belgium
```
adamsachs, line 26: @mfbrown does this look better? :)

src/fides/api/ops/models/privacy_notice.py
adamsachs: updated in https://github.com/ethyca/fides/pull/2928/commits/a8828f2da12b45f9b880aef91e96ffd2620a8158, new comment thread there so closing this out

src/fides/api/ops/models/privacy_notice.py
pattisdr: this looks great to me

adamsachs: thanks @pattisdr ! gonna close out and merge for now, @mfbrown if you spot any issues we can address 'em in a follow up :)

PR #2943, opened 2023-03-29
Add privacy centre support for more text

### LINKED ISSUE #2940: More advanced Privacy Centre styling
### Description

We've had requests from users to support >1 paragraph of legal text in the privacy centre descriptions.

seanpreston:
Closes #2940 

### Code Changes

* [x] Convert `description` to a list that will render multiple paragraphs
* [x] Add `addendum` to display after the privacy options
* [x] Separate `config.consent` into two distinct subsections: `config.consent.page` and `config.consent.button`

### Steps to Confirm

* [x] `npm run dev` and confirm the privacy centre looks like this:
Screenshot 2023-03-29 at 13 38 06

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`


clients/privacy-center/pages/index.tsx
```diff
@@ -130,20 +130,39 @@ const Home: NextPage = () => {
...
-              fontWeight="medium"
-              maxWidth={624}
-              textAlign="center"
-              color="gray.600"
-              data-testid="description"
-            >
-              {config.description}
-            </Text>
+
+            {config.description.map((paragraph) => {
```
NevilleS, line 134: This isn't backwards compatible though - it'll break existing configs that are a `string` and not an array

clients/privacy-center/types/config.ts
```diff
@@ -8,7 +8,8 @@ export type IdentityInputs = {
 
 export type Config = {
   title: string;
-  description: string;
+  description: string[];
```
NevilleS, line 11: If you make this `string | string[]` and Typescript is happy, I'll be happy 👍

NevilleS (changes requested):
Reasonable addition! But it'll break ;)

clients/privacy-center/config/config.json
```diff
@@ -1,6 +1,8 @@
 {
   "title": "Take control of your data",
   "description": "When you use our services, you’re trusting us with your information. We understand this is a big responsibility and work hard to protect your information and put you in control.",
+  "description_subtext": [],
+  "addendum": ["This is text that appears after the privacy options."],
```
NevilleS, line 5: Nit: this default text will appear on privacy centers by default... So you should set this to be empty 😛

clients/privacy-center/pages/consent.tsx
```diff
@@ -346,17 +346,21 @@ const Consent: NextPage = () => {
             >
               Manage your consent
```
NevilleS, line 347: Why is this not configurable too? 😅

NevilleS (approved):
Couple more asks, but once addressed don't wait for me!

clients/privacy-center/packages/fides-consent/rollup.config.js
```diff
@@ -24,7 +24,7 @@ const generateConsentConfig = () => {
    */
   const privacyCenterConfig = require("../../config/config.json");
   const privacyCenterOptions =
-    privacyCenterConfig.consent?.consentOptions ?? [];
+    privacyCenterConfig.consent?.page.consentOptions ?? [];
```
seanpreston, line 27: Hey @NevilleS, have made some changes to the overall readability of this config as prev. we ended up with too many different `title` and `description` objects.

As part of it I ended up in this file — is this the correct place to be changing this code?

clients/privacy-center/constants/index.ts
```diff
@@ -1,9 +1,34 @@
...
+ * reference config variables uniformly.
+ */
+const transformConfig = (config: Config): V2Config => {
+  if (isV1ConsentConfig(config.consent)) {
+    const v1ConsentConfig: V1Consent = config.consent;
+    const translatedConsent: V2Consent = translateV1ConfigToV2({
+      v1ConsentConfig,
+    });
+    const temp: V2Config = { ...config, consent: translatedConsent };
+    return temp;
```
allisonking, line 26: probably want to remove the `temp` var and just condense to `return { ...config, consent: translatedConsent }`

clients/privacy-center/packages/fides-consent/rollup.config.js
```diff
@@ -23,8 +23,13 @@ const generateConsentConfig = () => {
    * @type {import('../../types/config').Config}
    */
   const privacyCenterConfig = require("../../config/config.json");
-  const privacyCenterOptions =
-    privacyCenterConfig.consent?.consentOptions ?? [];
+  const isV1ConsentConifg = typeof privacyCenterConfig.consent?.button === "undefined";
+  let privacyCenterOptions = [];
+  if (isV1ConsentConifg) {
```
allisonking, line 28: small typo :)

```suggestion
  const isV1ConsentConfig = typeof privacyCenterConfig.consent?.button === "undefined";
  let privacyCenterOptions = [];
  if (isV1ConsentConfig) {
```

clients/privacy-center/packages/fides-consent/rollup.config.js
```diff
@@ -23,8 +23,13 @@ const generateConsentConfig = () => {
    * @type {import('../../types/config').Config}
    */
   const privacyCenterConfig = require("../../config/config.json");
-  const privacyCenterOptions =
-    privacyCenterConfig.consent?.consentOptions ?? [];
+  const isV1ConsentConifg = typeof privacyCenterConfig.consent?.button === "undefined";
```
allisonking, line 26: is it not possible to use the `isV1ConsentConfig` function you made earlier, here? 

I wouldn't be surprised if not, since this is a JS file and not a TS one, but I just want to make sure

clients/privacy-center/scripts/validate-config.js
```diff
@@ -9,7 +9,14 @@ const configIsValid = (
...
   // Cannot currently have more than one consent be executable
   if (config.consent) {
-    const executables = config.consent.consentOptions.filter(
+    let options = [];
+    const isV1Consent = typeof config.consent?.page === "undefined";
+    if (isV1Consent) {
+      options = config.consent?.consentOptions
+    } else {
+      options = config.consent?.page.consentOptions
+    }
```
allisonking, line 18: would typically use a ternary for something like this

```suggestion
    options = isV1Consent ? config.consent?.consentOptions : config.consent?.page.consentOptions;
```

allisonking: and then the `let options = []` might not be needed since all cases would be covered by the if/else

clients/privacy-center/pages/consent.tsx
```diff
@@ -344,19 +344,25 @@ const Consent: NextPage = () => {
...
+              (paragraph, index) => (
+                <Text
+                  fontSize={["small", "medium"]}
+                  fontWeight="medium"
+                  maxWidth={624}
+                  textAlign="center"
+                  color="gray.600"
+                  data-testid={`description-${index}`}
+                  // eslint-disable-next-line react/no-array-index-key
+                  key={`description-${index}`}
```
allisonking, line 360: react keys are generally pretty important to get right, as they result in mysterious looking errors when they aren't right. I think we are okay for now here, since there really isn't a good key for this kind of thing. errors that could come up would be things like the element still lingering in the DOM when it shouldn't be, or you make a change to the element and suddenly there are multiples of that element. since this isn't a field that dynamically changes (except for maybe if you're hot reloading config.json? not sure) we are probably okay, but just want to flag what kind of risk we might bump into

clients/privacy-center/packages/fides-consent/rollup.config.js
allisonking: this could probably be a ternary as well, i.e.

```js
const privacyCenterOptions = isV1ConsentConfig ? privacyCenterConfig.consent?.consentOptions : privacyCenterConfig.consent?.page.consentOptions;

if ( !privacyCenterOptions || privacyCenterOptions.length === 0) {
  ...
}
```

allisonking (approved):
nice improvement! just some small things that are really probably nits, do what you want with 'em :)

clients/privacy-center/constants/index.ts
seanpreston: IIRC we needed this for the typechecker?

clients/privacy-center/packages/fides-consent/rollup.config.js
seanpreston: I made the assumption here that `fides-consent.js` may not always be guaranteed to have access to the `isV1ConsentConfig` function given that this is its own package — even though it does currently.

clients/privacy-center/pages/consent.tsx
seanpreston: 100% agree. In this case I'm under the impression our hands are tied because beyond the list index there's no other locally unique value to use for the key?

clients/privacy-center/packages/fides-consent/rollup.config.js
seanpreston: Thanks!

PR #2979, opened 2023-04-03
nox support for git tagging with scheme to support feature branch builds

### LINKED ISSUE #2917: feature branch builds - nox commands and tagging scheme

Formalize a tagging scheme and develop nox commands/utilities based on that to help with commit tagging, so as to enable a sustainable and automat-able approach for feature branch building.

**Constraints**
I see the following constraints for the nox commands and tagging/versioning:

- we should support tagging as a means of test package versioning on the 3 fundamental branch types -main, an arbitrary "feature" branch, and release-... branches

- release branches should produce rc[N] suffixed tags, so as to produce proper "release candidate" packages that can be used for release testing

- main branch tags are already being versioned automatically by versioneer and published with a [latestReleaseVersion].post0.dev[N] version number. if we can continue to incorporate this behavior, that makes our implementation simpler as we don't have to do anything "special" for main commits

- feature branch versions should ideally be considered lower than any package versions built off of main commits on the same release version, so as to prevent our basic edge (and therefore nightly) builds from accidentally picking up a feature branch build that's not (by definition) ready for "prime time" yet

- our nox command on fides should "figure out" the right version tag for the developer, based on the active branch type, the current release increment, and existing versions present on testpypi. it should clearly output to the developer (and to the eventual CI job) what version tag has been used, so that the version tag can be referenced in downstream activities

- our nox command on fidesplus, which builds off an arbitrary fides package version/tag, need not have any smarts to determine which fides package version to build off; instead, it can take that version as an input param

See [proposed tagging scheme](https://ethyca.atlassian.net/wiki/spaces/EN/pages/2648408073/2023-03-21+Fidesplus+Feature+Branch+Test+Builds+fidesplus+668#Proposed-tagging-scheme) and [proposed nox commands](https://ethyca.atlassian.net/wiki/spaces/EN/pages/2648408073/2023-03-21+Fidesplus+Feature+Branch+Test+Builds+fidesplus+668) on the confluence doc.

Proposed tagging scheme is still a bit subpar. Likely want to think on that one a bit more and arrive at something a bit cleaner.

### Additional context

Part of broader feature-branch build support work

adamsachs:
Partially closes #2917 

### Code Changes

* `git_nox.py` to include git-related sessions and utilities
* a new `nox` session, `tag`, which generates a tag and applies it to the currently checked-out HEAD commit, based on the currently checked-out branch and existing tags in the repo
    *  by default (e.g. `nox -s tag `), the session is a "dry_run" - which just allows the user to check what the generated tag would be without actually applying it.
    * a `push` posarg option (e.g. `nox -s tag -- push`) generates, applies and pushes tag to the remote (on the currently checked-out HEAD commit)
    *  a `tag_only` posarg option (e.g. `nox -s tag -- tag_only`) generates a tag and applies it to the current commit locally, but doesn't actually push that tag to the remote repo
* an automated test for the utility that generates the tag, to evaluate the tagging logic/scheme. (yes, a test of our nox component is pretty meta...maybe this is a no-no? 😬 )
    * and a `nox` session to execute these tests, `nox -s "pytest(nox)"`, which is now run as part of our backend checks in CI. is this too meta???

### Steps to Confirm

* you can use `nox -s tag` locally to evaluate what some of the tags may be, switching to different branches and also applying certain tags locally to evaluate different scenarios
   * please don't push your test tags to the remote if you can avoid it, as there are GH CI actions that are triggered whenever a tag is pushed, and we don't want to create cruft/use resources unnecessarily

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### Description Of Changes

Tagging scheme [described in this doc](https://ethyca.atlassian.net/wiki/spaces/EN/pages/2648408073/2023-03-21+Fidesplus+Feature+Branch+Test+Builds+fidesplus+668#Proposed-tagging-scheme) has been updated to match what i've implemented here, and i believe it should work well for us going forward. See some of the [automated test cases in this PR](https://github.com/ethyca/fides/pull/2979/files#diff-2fbadfd853020fce2d18db143993ee0b3938dcbadffa8598c7ff4ed600245498R18-R37) to get a feel for how the tag generation works in different scenarios. There are some changes to how we've tagged and versioned up to this point, but nothing too drastic. I'm open to hearing other ideas if anyone's got concerns with what i've landed on though!

We'll need to update the release process doc a bit so that release managers:
1. ensure they make release branches with `release-` prefix
2. can leverage the new nox session to generate `rc` tags on their release branch, in a consistent and reliable mannger
3. can leverage the new nox session to generate a `b0` tag on `main` after publishing the release
Note that the final, official _release tag_ will still be done thru GH UI, as release notes are generated, etc. Changing that process would be disruptive.

Beyond those minor improvements to the release process, this functionality will more importantly allow us to publish testpypi packages based on feature branches in an easy and reliable way, which in turn will allow us to build `fidesplus` "edge" or "double-edge" environments based on feature branches. In order to do that, this change will be coupled with [a new nox option in `fidesplus`](https://github.com/ethyca/fidesplus/pull/796) that allows the feature-branch version of `fides` to be explicitly pinned into the edge/double-edge build process.


tests/nox/test_git_nox.py
```diff
@@ -0,0 +1,59 @@
+from unittest import mock
+
+import pytest
+from git import Repo
+
+from noxfiles.git_nox import generate_tag
+
+
+class TestGitNox:
```
SteveDMurphy, line 9: To keep the meta train rolling 😀  should we add a nox command to run these?

tests/nox/conftest.py
```diff
@@ -0,0 +1,68 @@
+"""This file is only for the database fixture. For all other fixtures add them to the
+tests/conftest.py file.
+"""
+
+import pytest
```
SteveDMurphy, line 5: I think we can trash this file and have a pytest command look something like `pytest tests/nox/ --noconftest`

noxfiles/git_nox.py
```diff
@@ -0,0 +1,166 @@
...
+from typing import List, Optional
+
+from packaging.version import Version
+
+import nox
+
+RELEASE_BRANCH_REGEX = r"release-(([0-9]+\.)+[0-9]+)"
+RC_TAG_REGEX = r"{release_version}rc([0-9]+)"
+RELEASE_TAG_REGEX = r"(([0-9]+\.)+[0-9]+)"
+VERSION_TAG_REGEX = "{version}{tag_type}([0-9]+)"
```
SteveDMurphy, line 14: Curious why there is no need for `VERSION_TAG_REGEX` to be treated as raw but there is for the others (this has confused me in the past too sometimes so not necessarily anything to change or do)

noxfiles/git_nox.py
```diff
@@ -0,0 +1,166 @@
...
+
+# our tag "types"
```
SteveDMurphy, line 16: Comment might be better as a doc string in `TagType` (or removed?)

noxfiles/git_nox.py
```diff
@@ -0,0 +1,166 @@
...
+    current_tag = get_current_tag(
+        existing=True,
+        repo=repo,
+        all_tags=all_tags,
+    )  # first get an existing tag if we've got one
+    generated_tag = generate_tag(
+        session, repo.active_branch.name, all_tags
+    )  # generate a tag based on the current repo state
+
+    if "dry_run" in session.posargs:
```
SteveDMurphy, line 86: I really like the `dry_run` 🙌🏽 

I noticed there was some hesitancy about pushing too many tags up etc. Does it make sense to have some sort of guardrails around not using the `dry_run` ? Some extra option like "to continue enter y" sort of thing or to have `dry_run` be the default and a separate positional argument to run it through? Also fine to leave as is, just thinking out loud

SteveDMurphy (approved):
I think this looks really good @adamsachs ! Thanks for the helpful PR description as well. Left a few comments/suggestions that shouldn't be blocking anything but let me know what you think. A lot of this is pretty difficult to see all the way through until it is in the wild but it looks great 👍🏽

noxfiles/git_nox.py
```diff
@@ -0,0 +1,166 @@
...
+
+import re
+from enum import Enum
+from typing import List, Optional
+
+from packaging.version import Version
+
+import nox
+
+RELEASE_BRANCH_REGEX = r"release-(([0-9]+\.)+[0-9]+)"
```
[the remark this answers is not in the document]
adamsachs: yeah, good call! didn't realize we've already got some docs about [branching](https://docs.ethyca.com/fides/community/development/releases#branching), i'll definitely add some notes in there

noxfiles/git_nox.py
adamsachs: hmm i may not be totally tracking what you're suggesting, but just to be clear on the behavior here:
- if no argument is passed, i.e. `nox -s tag`, then a tag is applied to the current commit locally but not pushed
- if `nox -s tag -- dry_run` is run (or if `dry_run` is included with any other additional arguments too), then no tag is applied, you just see the tag that _would_ be applied to the current commit
- if `nox -s tag -- push` is run, then a tag is generated and applied locally _and_ pushed up to the remote repo
- if `nox -s tag -- push_current` is run, then no tag is generated or applied locally, but instead we just try to push up whatever the current commit is tagged as locally

i felt like that covered the bases pretty well and defaulted to applying, but not pushing, a tag, which feels like a reasonable default because it gives you _something_, i.e. it generates and applies a tag, but it means you've got to be deliberate about actually _pushing_ that tag, which is the action with more consequences. that being said, i'm definitely open to alternatives, and it should be easy to tweak this logic if we'd like

adamsachs:
> I think this looks really good @adamsachs ! Thanks for the helpful PR description as well. Left a few comments/suggestions that shouldn't be blocking anything but let me know what you think. A lot of this is pretty difficult to see all the way through until it is in the wild but it looks great 👍🏽

Thanks for the helpful comments @SteveDMurphy! i fixed up and rebased to keep a clean commit history - i think i got them mostly addressed besides one i was a bit unclear on what you were suggesting. let me know when you get a chance to look, no huge rush!

FYI i also added in the nox tests themselves into our CI workflow, i hope this isn't getting too meta...

noxfiles/git_nox.py
SteveDMurphy: Gotcha, that makes sense - thanks for laying that out! 

I wasn't sure if having a bunch of tags created locally is something we wanted or needed to avoid but it sounds like the `nox -s tag -- push` will generate a tag of the current state regardless and push it. So although there may be a lot of tags generated locally it doesn't matter as much or add any risk around having them pushed to the repo, right?

noxfiles/git_nox.py
SteveDMurphy: I need to spend a day re-learning all of that at some point, I end up just throwing pasta at the wall most of the time 😄

SteveDMurphy:
> FYI i also added in the nox tests themselves into our CI workflow, i hope this isn't getting too meta...

I think it's nice to have! It's sort of an odd failure in CI tho after passing locally 🤔  I didn't look too closely but if it becomes a problem maybe strip it out and add a follow on issue so we can see more of this in the wild and circle back as needed

adamsachs:
> > FYI i also added in the nox tests themselves into our CI workflow, i hope this isn't getting too meta...
> 
> I think it's nice to have! It's sort of an odd failure in CI tho after passing locally 🤔 I didn't look too closely but if it becomes a problem maybe strip it out and add a follow on issue so we can see more of this in the wild and circle back as needed

aw shucks, it seems like the GH runner isn't actually able to connect to the git repo to read tags 🤔 ... i'll see if i can debug it a bit, or _maybe_ i can do some mocking in the tests so as to not rely on git access. but i'll probably timebox that effort if i can't find something relatively quick.

noxfiles/git_nox.py
adamsachs: yeah, that's generally true, although if you generate a bunch of tags locally, they'll keep on getting incremented unnecessarily. that's not super consequential, but it's probably less than ideal. maybe i'll switch the default to actually be the dry_run, since that's truly the least intrusive thing, and then there can be a `push` flag that is used to actually publish your changes. the tagging without pushing is not very useful and does have some less-than-ideal side effects if done without care, so i think you're right that it shouldn't be the default.

this is making sense to me, i'll push up a change in a bit and let me know what you think! happy to talk it over if my stream of consciousness above is hard to follow! 😄

adamsachs:
> > > FYI i also added in the nox tests themselves into our CI workflow, i hope this isn't getting too meta...
> > 
> > 
> > I think it's nice to have! It's sort of an odd failure in CI tho after passing locally 🤔 I didn't look too closely but if it becomes a problem maybe strip it out and add a follow on issue so we can see more of this in the wild and circle back as needed
> 
> aw shucks, it seems like the GH runner isn't actually able to connect to the git repo to read tags 🤔 ... i'll see if i can debug it a bit, or _maybe_ i can do some mocking in the tests so as to not rely on git access. but i'll probably timebox that effort if i can't find something relatively quick.


ok @SteveDMurphy! i think i got this figured out, just needed to explicitly fetch the tags once in the test session. tests were passing in CI on previous commits, they should be good on this re-run. i also simplified the available commands per the discussion above. i've left it as a separate commit so if we want to revert, we can - but i'll rewrite the GH description accordingly. i _think_ this new behavior is safer, easier to use, and easier to follow in the code. let me know what you think.

SteveDMurphy (approved):
Thanks for all the extra time and energy on this @adamsachs ! Great changes, excited to see it in action

noxfiles/git_nox.py
```diff
@@ -0,0 +1,154 @@
...
+        )
+        tag_increment = int(latest_tag.group(1)) + TAG_INCREMENT  # increment the tag
+        return f"{version_number}{tag_type.value}{tag_increment}"
+        # return the full {version_number}{tag_type}{increment}
+
+    # we don't have an existing tag for this version/type, so start at our initial increment
+    return f"{version_number}{tag_type.value}{INITIAL_TAG_INCREMENT}"
+
+
+def generate_tag(session: nox.Session, branch_name: str, all_tags: List) -> str:
```
NevilleS, line 139: nit: I think I'd like to see some `session.log(...)` lines in here explaining the magic. Something like:
```
Fetching current tag and branch to generate a new tag: tag = 2.10.0, branch = 'main'
Current branch 'main' matched main branch, generating a new BETA tag...
Generated new tag: 2.10.0b1
```

tests/nox/test_git_nox.py
```diff
@@ -0,0 +1,67 @@
...
+            ("release-2.9.0", ["2.9.0rc1"], "2.9.0rc2"),
+            (
+                "release-2.9.0",
+                ["2.9.1"],
+                "2.9.0rc0",
+            ),  # even with a release further ahead, the RC tag generation relies on the release branch name
+            # release branches MUST adhere to `release-n.n.n` convention
+            # if not, they will be treated as any other "feature" branch
+            ("release/2.9.0", ["2.9.0"], "2.9.1a0"),
+            ("release_2.9.0", ["2.9.0"], "2.9.1a0"),
```
NevilleS, line 44: nit: Super polish suggestion - if you detect the word `release` in the branch name, throw a warning like:
```
WARNING: Current branch name 'release/testing' does not match release branch format ('release-n.n.n'), so will be detected as a feature branch instead. Did you mean to name your branch differently?
```

NevilleS (approved):
This looks helpful, thanks! Couple minor DX nits

tests/nox/test_git_nox.py
adamsachs: nice, very defensive, i like it. i'll add. thanks!

noxfiles/git_nox.py
adamsachs: added some extra logging in https://github.com/ethyca/fides/pull/2979/commits/32bd7487b0e8b117aa2d298ce0afc5c5514cb588, resolving

PR #2995, opened 2023-04-05
Privacy notice initial page and redux

allisonking:
Closes https://github.com/ethyca/fidesplus/issues/736

### Code Changes

* [x] Add privacy notice page and flag
* [x] Add privacy notice GET and state to redux store
* [x] Add a very basic "table" component just to show the API call is working
* [x] Add the page to the nav

### Steps to Confirm

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

### DIFF
--- a/clients/admin-ui/src/features/privacy-notices/PrivacyNoticesTable.tsx
+++ b/clients/admin-ui/src/features/privacy-notices/PrivacyNoticesTable.tsx
@@ -1,0 +1,35 @@
+import { Box, List, ListItem, Text } from "@fidesui/react";
+
+import { useAppSelector } from "~/app/hooks";
+
+import {
+  selectAllPrivacyNotices,
+  selectPage,
+  selectPageSize,
+  useGetAllPrivacyNoticesQuery,
+} from "./privacy-notices.slice";
+
+const PrivacyNoticesTable = () => {
+  // Subscribe to get all privacy notices
+  const page = useAppSelector(selectPage);
+  const pageSize = useAppSelector(selectPageSize);
+  useGetAllPrivacyNoticesQuery({ page, size: pageSize });
+
+  const privacyNotices = useAppSelector(selectAllPrivacyNotices);
+
+  if (privacyNotices.length === 0) {
+    return <Text>No privacy notices found.</Text>;
+  }
+
+  return (
+    <Box>
+      <List>
+        {privacyNotices.map((notice) => (
+          <ListItem key={notice.id}>{notice.name}</ListItem>
+        ))}
+      </List>
+    </Box>
+  );
+};
+
+export default PrivacyNoticesTable;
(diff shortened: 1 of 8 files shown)


clients/admin-ui/src/features/privacy-notices/PrivacyNoticesTable.tsx
```diff
@@ -0,0 +1,37 @@
...
+  selectPage,
+  selectPageSize,
+  useGetAllPrivacyNoticesQuery,
+} from "./privacy-notices.slice";
+
+const PrivacyNoticesTable = () => {
+  // Subscribe to get all privacy notices
+  const page = useAppSelector(selectPage);
+  const pageSize = useAppSelector(selectPageSize);
+  useGetAllPrivacyNoticesQuery({ page, size: pageSize });
```
TheAndrewJackson, line 16: Is there a specific reason that you didn't do something like this?

```typescript
const { data } = useGetAllPrivacyNoticesQuery({ page, size: pageSize });
```

I see the `selectAllPrivacyNotices` selector in the slice. Is it to try and memoize the selection? I thought RTK handles that for us with RTK query

allisonking: yeah, so `useGetAllPrivacyNoticesQuery` returns the paged response of privacy notices `{items: [list of privacy notices], page: number, size: number, total: number}`. 

using the selector is nice since it'll always just return what's in `items` instead of the component having to index into it itself. the other nice thing is that it'll return an empty array instead of undefined if the query isn't done yet which makes some things easier in terms of types and handling undefined's. 

I definitely appreciate the inquiry since this is my first time really working with an endpoint that pages, so I am a little unsure about if the way I'm passing the page parameters in is the right way in terms of memoization. I expect these redux calls may change as I continue building the feature out (and as we have more privacy notices to play with!)

TheAndrewJackson: That makes sense. Handling `undefined` can be annoying. I just did this yesterday for the dataflow worked to resolve a list of systems into one.

https://github.com/ethyca/fides/blob/3a78ea56cb5f44adcfda61bdb67d5ae72a94cf11/clients/admin-ui/src/features/common/DataFlowForm.tsx#L27-L31


`selectFromResult ` is a helpful callback transform a response from an individual hook level instead of the slice level. You could potentially do something like that and have it return an empty list while data is undefined.

TheAndrewJackson: I actually looked for something like `privacyNoticesApi.endpoints.getAllPrivacyNotices.select` and found `selectFromResult` first 😅

allisonking: oh that's neat, I haven't used `selectFromResult` before but I do remember seeing it in some docs before! I suppose the selector is good for if you want to reuse it elsewhere, but other than that it seems like your solution with `selectFromResult` should work just as well!

PR #2998, opened 2023-04-05
DynamoDB Connector

### LINKED ISSUE #2938: DynamoDB Connector - Consent Policy Spike
### Description

Fides needs to be able to integrate with DynamoDB for the purposes of consent management (DSR support will come later). 
 Datastore consent management most commonly involves storing a list of consent records that the owner of the datastore can then query later on to ascertain what's permissible for user data. This ticket is to suggest a path to the storing of consent preferences dynamically within DynamoDB, and to build tickets + start implementing the feature once design is signed off.

### Acceptance Criteria

- [ ] Fides can build queries to insert consent records into a DynamoDB collection
- [ ] A user can customise which DynamoDB collection the consent preferences are stored in

### Questions

- [ ] How can we best extend the Fides dataset configurations to facilitate storing of consent records within connected datastores? (e.g. if somebody has connected Postgres, what do they need to add to their Postgres dataset to support consent records)

SteveDMurphy:
Closes #2938

### Code Changes

* [x] Add aws connector commands for `DynamoDB`
* [x] Include a CLI command to generate a dataset
* [x] Update the API to generate a `DynamoDB` dataset
* [x] ~Store the primary key and secondary sort (optional) as dataset meta~ inferred from `get_item`
* [x] Add tests for the `request_runner_service` (access & erasure)
* [x] Add tests for the CLI command
* [x] Add tests for the query param generation
* [x] Create and add an SVG template for the frontend for DynamoDB 

### Steps to Confirm

* [x] Set credentials as stored in 1Password
* [x] Create a connector for Dynamo (the config options are sneakily flipped, not sure how here)
* [x] Test the connection

### Pre-Merge Checklist

* Documentation:
  * [x] documentation [issue created in fidesdocs](https://github.com/ethyca/fidesdocs/issues/new/choose) [fidesdocs#91](https://github.com/ethyca/fidesdocs/issues/91)
* [x] Update `CHANGELOG.md`

### Description Of Changes

This is the connector for DynamoDB, the consent component will be worked separately in both documentation and fidesplus. 

It appears that some of the other unsafe checks that are not currently passing are being ignored for now, although I have validated the ones added here for DynamoDB appear to be in the clear :)

There have been more than a few migration conflicts so there may be some further need to keep up with more of those changes until this merges


SteveDMurphy:
I'm still resolving a couple of spooky issues that I think were related to migration conflicts but mostly just planning to add some DynamoDB request specific unit tests and would love any early feedback @galvana !

src/fides/api/ctl/routes/generate.py
```diff
@@ -69,6 +73,7 @@ def target_matches_type(cls, values: Dict) -> Dict:
         """
         target_type = (values.get("target"), values.get("type"))
         valid_target_types = [
+            ("aws", "datasets"),
```
SteveDMurphy, line 76: I'm thinking this might be better off as `dynamodb` instead of `aws` to reduce some of the confusion and keep in line with `bigquery` (it doesn't say `gcp` for instance)

tests/fixtures/application_fixtures.py
```diff
@@ -144,6 +144,15 @@
             integration_config, "fides_example.polling_timeout"
         ),
     },
+    "dynamodb_example": {
```
SteveDMurphy, line 147: I think these can be removed until we look at shift to use https://hub.docker.com/r/amazon/dynamodb-local/

data/dataset/dynamodb_example_test_dataset.yml
```diff
@@ -0,0 +1,62 @@
+dataset:
+  - fides_key: dynamodb_example_test_dataset
+    name: DynamoDB Example Test Dataset
+    description: Example of a DynamoDB dataset containing a single customer table
+    collections:
+      - name: customer_identifier
```
galvana, line 6: Hey! It's that linking table we talked about

src/fides/api/ctl/routes/generate.py
galvana: I agree, I'd rather have it be in line with the name of the connector.

src/fides/api/ops/schemas/connection_configuration/connection_secrets_dynamodb.py
```diff
@@ -0,0 +1,24 @@
...
+from fides.api.ops.schemas.base_class import NoValidationSchema
+from fides.api.ops.schemas.connection_configuration.connection_secrets import (
+    ConnectionConfigSecretsSchema,
+)
+
+
+class DynamoDBSchema(ConnectionConfigSecretsSchema):
+    """Schema to validate the secrets needed to connect to an Amazon DynamoDB cluster"""
+
+    region_name: Optional[str]
```
galvana, line 12: Would it be more accurate for these to be required string type instead of optional given the `_required_components` definition?

src/fides/cli/commands/generate.py
```diff
@@ -111,6 +111,52 @@ def generate_dataset_bigquery(
     )
 
 
+@generate_dataset.group(name="aws")
+@click.pass_context
+def generate_dataset_aws(ctx: click.Context) -> None:
```
galvana, line 116: Is this a placeholder for something else?

tests/fixtures/application_fixtures.py
galvana: I think it's fine to keep it like this for now, we can come back to it in another ticket

src/fides/cli/commands/generate.py
SteveDMurphy: These are the CLI commands to generate a dataset based off of what is discovered in DynamoDB - somewhat opposite to the above comment actually where we were going to make the API grouping be specific to DynamoDB this one might make sense to keep as AWS for now 🤔 could probably use some wider thinking tbh

data/dataset/dynamodb_example_test_dataset.yml
SteveDMurphy: I really wanted to try and that in as a test case like we discussed 🤗

src/fides/api/ctl/routes/generate.py
SteveDMurphy: Cool, will clean this up 👍🏽 comment further down has me in a bit of a pickle with the inconsistency between cli and api but they are different patterns to maybe it makes sense 🤷🏽 more opinions needed I think in the future

galvana: We should also update the example in the docstring `('aws', 'datasets')` is a valid example now and that is too close to the `('aws', 'datasets')` in the invalid example.

src/fides/api/ctl/routes/generate.py
```diff
@@ -192,11 +200,15 @@ def generate_aws(
             status_code=status.HTTP_401_UNAUTHORIZED,
             detail=str(error),
         )
+    if generate_type == "systems":
+        log.info("Generating systems from AWS")
+        aws_resources = generate_aws_systems(
+            organization=organization, aws_config=aws_config
+        )
+    else:
+        aws_resources = [generate_dynamo_db_datasets(aws_config=aws_config)]
```
galvana, line 209: It's a bit confusing for the `generate_aws` function to return lists of `Datasets` and `Systems`. I see the type hint on the function allows for this but should we tighten up the types here and move the dataset generation to a new function that returns datasets?

src/fides/api/ctl/routes/generate.py
SteveDMurphy: Updated all in 5bc94373add7218f9f2aeac5afd2ceb7ab85351d

src/fides/api/ctl/routes/generate.py
SteveDMurphy: Totally agree 👍🏽 changed this up as part of 5bc94373add7218f9f2aeac5afd2ceb7ab85351d as well

src/fides/api/ops/service/connectors/dynamodb_connector.py
```diff
@@ -0,0 +1,135 @@
...
+            ]
+            query_param = query_config.generate_query(input_data, policy)
+            item = client.get_item(
+                TableName=collection_name,
+                Key=query_param,  # type: ignore
+            )
+            result = {}
+            if "Item" in item:
+                for key in item["Item"]:
+                    result[key] = [*item["Item"][key].values()][0]
```
galvana, line 99: We should check for instances where the value is a list or a complex type. For example:
```
{
  "book_id": 1,
  "title": "Sample Book",
  "author": "John Doe",
  "genres": {"Mystery", "Thriller"},
  "publish_date": "2022-01-01",
  "ratings": [
    {
      "user_id": 101,
      "rating": 4
    },
    {
      "user_id": 102,
      "rating": 5
    }
  ]
}
```
We might miss some data if we only take the first element

src/fides/api/ops/service/connectors/dynamodb_connector.py
```diff
@@ -0,0 +1,135 @@
...
+        """Retrieve DynamoDB data"""
+        query_config = self.query_config(node)
+        collection_name = node.address.collection
+        client = self.client()
+        try:
+            describe_table = client.describe_table(TableName=collection_name)
+            input_data["attribute_definitions"] = describe_table["Table"][
+                "AttributeDefinitions"
+            ]
+            query_param = query_config.generate_query(input_data, policy)
```
galvana, line 91: We need to support the use case where we receive multiple IDs in the `input_data` map. Using the book example from my other comment, it's possible that the input data might contain
```
{"review_ids": [1,2,3]}
```
and in this case we would need to generate three `query_params` or if possible, have `generate_query` generate the DynamoDB equivalent of a where clause, where the ID can be one of [1,2,3]

src/fides/api/ops/service/connectors/dynamodb_connector.py
SteveDMurphy: That makes sense 👍🏽 thanks @galvana ! [Data types listed here](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/HowItWorks.NamingRulesDataTypes.html#HowItWorks.DataTypes) so it is definitely a thing.

Do you know if there are any bounds to the scenarios we are trying to account for? i.e. I'm trying to think of if we currently support referencing a list of IDs instead of a single item or if it is primarily attributes

src/fides/api/ops/service/connectors/dynamodb_connector.py
SteveDMurphy: Perfect, will use this - thanks! `get_item()` will force us to do one key at a time, so I think the multiple query params makes sense here

SteveDMurphy: I think I have this updated but not able to confirm if it is functioning how I want just yet. May look for an assist later if you are available

SteveDMurphy:
@galvana - This should be ready to review again as of now with tests passing for the query params and access/erasure! I am likely going to give it another review for general cleanup as well and maybe add some further query param tests but it shouldn't be functionally different. Thanks again for your help

src/fides/cli/commands/generate.py
seanpreston: Of course! Now the codebases are unified it makes sense that we'd add this too — nice job!

tests/ops/service/privacy_request/test_request_runner_service.py
```diff
@@ -2254,3 +2258,148 @@ def test_needs_batch_email_send_new_workflow(
             {"email": "customer-1@example.com", "ljt_readerID": "12345"},
             privacy_request_with_consent_policy,
         )
+
+
+@pytest.fixture(scope="function")
+def dynamodb_resources(
```
seanpreston, line 2264: Thanks for using this pattern 🙌

tests/fixtures/dynamodb_fixtures.py
```diff
@@ -0,0 +1,94 @@
...
+) -> Generator:
+    """
+    Returns a DynamoDB ConectionConfig with secrets attached if secrets are present
+    in the configuration.
+    """
+    dynamodb_connection_config = dynamodb_connection_config_without_secrets
+    region = integration_config.get("aws", {}).get("region") or os.environ.get(
+        "DYNAMODB_REGION"
+    )
+    aws_access_key_id = integration_config.get("aws", {}).get(
```
seanpreston, line 56: These fixtures fail to load locally, since the config dict is referenced as `aws` here and `dynamodb_example` in the [`integration_test_config.toml` file](https://github.com/ethyca/fides/pull/2998/files#diff-de8140bfa2cfbae36f4b1d426dce530a9a1b3aaa1f5bc9f06a3e3daab6e5b665R62)

tests/ops/service/privacy_request/test_request_runner_service.py
```diff
@@ -2254,3 +2258,148 @@ def test_needs_batch_email_send_new_workflow(
...
+    )
+    results = pr.get_results()
+    customer_table_key = (
+        f"EN_{pr.id}__access_request__dynamodb_example_test_dataset:customer"
+    )
+    address_table_key = (
+        f"EN_{pr.id}__access_request__dynamodb_example_test_dataset:address"
+    )
+    assert len(results[customer_table_key]) == 1
+    assert len(results[address_table_key]) == 2
```
seanpreston, line 2360: Would be good to also assert what's in the address table

tests/fixtures/dynamodb_fixtures.py
SteveDMurphy: Great catch, thank you!

SteveDMurphy: Fixed in https://github.com/ethyca/fides/pull/2998/commits/6cbb1f664fa12be058b1ad1bfdf0c6db5f3b0f17

tests/ops/service/privacy_request/test_request_runner_service.py
SteveDMurphy: Added assertion for the only value there in https://github.com/ethyca/fides/pull/2998/commits/2be77e4c8135ff543f5f5bb389157dbf52f8e656

src/fides/api/ops/schemas/connection_configuration/connection_secrets_dynamodb.py
```diff
@@ -0,0 +1,24 @@
...
+
+    region_name: str
+    aws_secret_access_key: str
+    aws_access_key_id: str
+
+    _required_components: List[str] = [
+        "region_name",
+        "aws_secret_access_key",
+        "aws_access_key_id",
+    ]
```
seanpreston, line 20: A more succinct version of this would be something in the base class to autogenerate this property, so in cases like this where all components are required we don't need to redefine this. Probably not worth changing here though.

src/fides/api/ops/service/connectors/dynamodb_connector.py
```diff
@@ -0,0 +1,166 @@
...
+        In the case of complex objects, returns multiple rows
+        as the product of options to query against.
+        """
+        deserializer = TypeDeserializer()
+        collection_name = node.address.collection
+        client = self.client()
+        try:
+            results = []
+            query_config = self.query_config(node)
+            # TODO check this out with a primary key and a sort key
```
seanpreston, line 102: What's the plan for this action?

src/fides/core/dataset.py
```diff
@@ -356,3 +361,16 @@ def generate_bigquery_datasets(bigquery_config: BigQueryConfig) -> List[Dataset]
         for dataset in bigquery_datasets
     ]
     return unique_bigquery_datasets
+
+
+def generate_dynamo_db_datasets(aws_config: Optional[AWSConfig]) -> Dataset:
+    """
+    Given an AWS config, extract all DynamoDB tables/fields and generate corresponding datasets.
+    """
+    import fides.connectors.aws as aws_connector
```
seanpreston, line 370: Do we need this here? Looks like we're already importing from `fides.connectors.aws` at the top of the file

tests/ops/service/privacy_request/test_request_runner_service.py
```diff
@@ -62,7 +66,7 @@
 
 PRIVACY_REQUEST_TASK_TIMEOUT = 5
 # External services take much longer to return
-PRIVACY_REQUEST_TASK_TIMEOUT_EXTERNAL = 30
+PRIVACY_REQUEST_TASK_TIMEOUT_EXTERNAL = 300
```
seanpreston, line 69: This will have ramifications for our Github actions runtime, so please let's make this as small as it can be

src/fides/api/ops/schemas/connection_configuration/connection_secrets_dynamodb.py
SteveDMurphy: Agreed, it felt like a better follow on issue here to look at both the way `ctl` and `ops` worked and unify them, will add a follow on issue 👍🏽

tests/ops/service/privacy_request/test_request_runner_service.py
SteveDMurphy: yes, debug issue 🤦🏽

src/fides/core/dataset.py
SteveDMurphy: I think this is a pattern holdover from when we had `boto` as an optional extra on the install (we do this same thing with other `aws` connections) and needed to control import errors.

Will import at the top and create a follow on issue to fix the rest 👍🏽

src/fides/api/ops/service/connectors/dynamodb_connector.py
SteveDMurphy: Tested locally but I don't think moved it into the datasets we are actually testing repeatedly and I didn't remove the comment. Will dig in here

src/fides/api/ops/service/connectors/dynamodb_connector.py
```diff
@@ -0,0 +1,175 @@
...
+        """
+        deserializer = TypeDeserializer()
+        collection_name = node.address.collection
+        client = self.client()
+        try:
+            results = []
+            query_config = self.query_config(node)
+            for attribute_definition in query_config.attribute_definitions:  # type: ignore
+                attribute_name = attribute_definition["AttributeName"]
+                for identifier in input_data[attribute_name]:
```
galvana, line 111: There it is, this was the main thing I was looking for from the last iteration.

src/fides/api/ops/service/connectors/dynamodb_connector.py
```diff
@@ -0,0 +1,175 @@
...
+                    selected_input_data[attribute_name] = [identifier]
+                    query_param = query_config.generate_query(
+                        selected_input_data, policy
+                    )
+                    items = client.query(
+                        TableName=collection_name,
+                        ExpressionAttributeValues=query_param[
+                            "ExpressionAttributeValues"
+                        ],
+                        KeyConditionExpression=query_param["KeyConditionExpression"],
```
galvana, line 122: I'm glad you got this working!

src/fides/api/ops/schemas/connection_configuration/connection_secrets_dynamodb.py
SteveDMurphy: Added in #3257

src/fides/api/ops/service/connectors/dynamodb_connector.py
SteveDMurphy: Yes! This one didn't have as many drawbacks from what I could tell and allowed us to account for SortKey usage as well 💙

SteveDMurphy:
@seanpreston expected checks are good and opened up https://github.com/ethyca/fides/issues/3278 for the two that are currently failing

seanpreston (approved):
Thanks for those extra changes @SteveDMurphy

PR #2999, opened 2023-04-06
remove datamap export API

adamsachs:
Partially closes https://github.com/ethyca/fidesplus/issues/801

### Code Changes

- remove datamap export API (redefined in `fidesplus`)
- remove `datamap:read` scope from `fides` (will be redefined in `fidesplus`)
- rewire datamap UI to look at the plus endpoint instead

### Steps to Confirm

* confirm that datamap export API is not exposed/available if just running fides from this repo (not `fidesplus`)

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

We're moving the datamap export API over to `fidesplus`


clients/admin-ui/src/features/datamap/datamap.slice.ts
```diff
@@ -68,7 +69,7 @@ const DEPRECATED_COLUMNS = [
 ];
 
 // API endpoints
-const datamapApi = baseApi.injectEndpoints({
+const datamapApi = plusApi.injectEndpoints({
   endpoints: (build) => ({
```
adamsachs, line 73: this is me totally hacking on the FE code, but it _seemed_ to be sending a `GET` to the plus endpoint (`http://localhost:3000/api/v1/plus/datamap/default_organization?include_deprecated_columns=true`) when i went to `localhost:3000/datamap` so...maybe not too far off?

adamsachs: ok - i think this now does what you suggested! let me know if it looks alright

adamsachs:
@allisonking i _think_ this is just about good to go, if you could take another look - specifically at the FE pieces - that would be much appreciated! 

@SteveDMurphy going to tag you as a reviewer here too, this is just the cleanup on `fides` corresponding to the addition to `fidesplus`

@TheAndrewJackson just an FYI since i know you're going to be working on datamap API code soon

allisonking (approved):
FE looks good to me! thanks @adamsachs 🙏

src/fides/api/ctl/routes/datamap.py
SteveDMurphy, line 1: Strictly for conversation at this point, is there any reason we would prefer some other response than a 404 and `Item not found`?

SteveDMurphy (approved):
Everything looks and works as expected! Great work @adamsachs - left one comment just for potential discussion

src/fides/api/ctl/routes/datamap.py
adamsachs: oh hmmm interesting, are you thinking something like a [`301`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/301)?

adamsachs: oh man, `Gone` has some sass. i like it.

adamsachs:
i was just doing a final pass through myself, and realized that one thing i missed in the migration was moving the scope over to `fidesplus` - it's still defined here in core `fides`: https://github.com/ethyca/fides/blob/main/src/fides/api/ops/api/v1/scope_registry.py#L125 

the new route in `fidesplus` references this scope from core `fides`, and that all seems to work fine. it's certainly a bit against the paradigm that's been set up, i think - where `fidesplus` endpoints have their scopes (and roles) defined in that code base. i'm not really sure what the end-user implications are, if any.

i was thinking of creating a follow up ticket for this migration rather than taking care of it immediately, since i _think_ it's more of a tech-debt issue rather than one that impacts users significantly. but i could be wrong about that.

@pattisdr @allisonking - tagging you both since i think you are closest to some of the scopes/roles refactoring that was done, specifically with moving some of that to `fidesplus`? @SteveDMurphy also curious if you've got any thoughts off the top of your head.

PR #1083, opened 2022-09-15
ui/state: Deduplicate query results that were being stored in state slices

ssangervasi:
Related to #1075 

### Code Changes

- Unused state:
    - Systems
    - Datasets
    - Organization hydrate
- Queries with duplicate state:
    - Taxonomy DataCategories query
    - DataSubjects query
    - DataQualifiers query
    - DataUses query

### Steps to Confirm

* [ ] Config wizard should work the same

### Pre-Merge Checklist

* [ ] All CI Pipelines Succeeded
* Documentation Updated:
  - [ ] documentation complete, or draft/outline provided (tag docs-team to complete/review on this branch)
  - [ ] documentation issue created (tag docs-team to complete issue separately)
* [ ] Issue Requirements are Met
* [ ] Relevant Follow-Up Issues Created
* [ ] Update `CHANGELOG.md`

### Description Of Changes

This does away with some state duplication that has sprung up due to some confusion about RTK Query vs slices.

**TLDR**: Use `useFooQuery` and/or `createSelector(fooApi.endpoints.fooQuery.select(), ...)`. Do not re-store query results.

With this PR I think the pattern should be easier to follow going forward, and this will live as a bit of documentation. I'll comment explaining on individual lines changed. First, an overview:

- `createApi` already stores query results in the Redux state.
    - This is referred to as the "cache".
    - [RTK provides a helper](https://redux-toolkit.js.org/rtk-query/usage/usage-without-react-hooks#accessing-cached-data--request-status) for creating a selector: `api.endpoints.fooQuery.select()`
    - This will only be populated if the query has run. The hooks are still useful!
    - The cache also takes into account the parameters of the request. This PR only has examples where the request takes no params.
- `useQuery` [subscribes](https://redux-toolkit.js.org/rtk-query/api/created-api/hooks#usequery) to the API _and_ provides access to the cache.
    - For any query, you **need at least one hook call** to actually populate the cache.
    - The response `data` is the same as the value in the cache.
- `selectFromResult` is a [shortcut](https://redux-toolkit.js.org/rtk-query/usage/queries#selecting-data-from-a-query-result) for subscribing _and_ deriving a value from the cache.
    - This is handy for one-off query hook uses, but if the derived value is shared then the logic should be in its own selector.
    - [Example use](https://github.com/ethyca/fides/blob/dc8b6f3ee9889aac4039ad180e9e5601706c4bdf/clients/ctl/admin-ui/src/features/common/plus.slice.ts#L89)
[...]

PR #3008, opened 2023-04-07
Data Flow Modal

### LINKED ISSUE #2941: Add data flow accordion and modal
https://www.figma.com/file/dmEwdK3xZwjKfGVQ9t66xe/Fides-v.2-Master-Working-Files?node-id=3892-123818&t=deCDCLuWBOGAFzeL-0

TheAndrewJackson:
Closes #2941 

### Code Changes

### Steps to Confirm

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes


clients/admin-ui/src/features/common/system-data-flow/DataFlowAccordionForm.tsx
```diff
@@ -0,0 +1,189 @@
...
+  const [updateSystemMutationTrigger] = useUpdateSystemMutation();
+
+  const { data } = useGetAllSystemsQuery();
+  const systems = useMemo(() => data || [], [data]);
+
+  const initialDataFlows = useMemo(() => {
+    const dataFlows = isIngress ? system.ingress! : system.egress!;
+    const systemFidesKeys = systems ? systems.map((s) => s.fides_key) : [];
+
+    return dataFlows.filter((df) => systemFidesKeys.includes(df.fides_key));
```
TheAndrewJackson, line 61: This filters out returned data flows that doesn't have a fides key that's tied to an active system. Right now we're only configuring data flows for systems so it doesn't make sense to factor them in.

The pixie scan was returning and `external` flow that had no system tied to it.

clients/admin-ui/src/features/common/system-data-flow/DataFlowAccordionForm.tsx
```diff
@@ -0,0 +1,189 @@
...
+    };
+    await updateSystemMutationTrigger(updatedSystem);
+    toast(successToastParams(`${pluralFlowType} updated`));
+  };
+
+  return (
+    <AccordionItem>
+      <AccordionButton
+        height="68px"
+        paddingLeft={isSystemTab ? 6 : 2}
```
TheAndrewJackson, line 84: The System configuration page the drawer require slightly different padding values

clients/admin-ui/src/features/datamap/SpatialDatamap.tsx
```diff
@@ -34,26 +37,30 @@ const useSpatialDatamap = (rows: Row<DatamapRow>[]) => {
         }, {} as Record<string, SystemNode>),
     [rows]
   );
-
   const data = useMemo(() => {
     let nodes: SystemNode[] = [];
-    let links: Link[] = [];
+    const links: Set<Link> = new Set([]);
```
TheAndrewJackson, line 42: I redid the link creation to use a set because it's possible that two systems could create identical links that would end up rendering on top of each other

example:

```
system_1.ingress = [system_2]
system_2.egress = [system_1]
```

Both of those would create an identical link object.

allisonking:
Still testing things out, but going to report this one first. I'm having trouble navigating away from the Data Flow tab. I haven't looked into it, but if I had to guess, I'd point at the dirty form slice?


https://user-images.githubusercontent.com/24641006/230980048-57f49884-3e83-47d1-9889-a09d32bd8ce8.mov

clients/admin-ui/src/features/system/SystemFormTabs.tsx
```diff
@@ -182,6 +183,27 @@ const SystemFormTabs = ({
...
       isDisabled: !activeSystem,
     },
+    {
+      label: "Data flow",
+      content: activeSystem ? (
+        <Box width={{ base: "100%", lg: "70%" }}>
+          <Box px={6} paddingBottom={2}>
+            <Text fontSize="md" lineHeight={6} fontWeight="bold">
+              Data flow
+            </Text>
```
allisonking, line 193: I think this may need some margin bottom?

hmm this doesn't look quite like the design to me.

| design | my instance |
|--------|------------|
|  image |  image |

clients/admin-ui/src/features/common/system-data-flow/DataFlowAccordionForm.tsx
```diff
@@ -0,0 +1,193 @@
...
+  isSystemTab?: boolean;
+};
+
+export const DataFlowAccordionForm = ({
+  system,
+  isIngress,
+  isSystemTab,
+}: DataFlowAccordionItemProps) => {
+  const toast = useToast();
+  const flowType = isIngress ? "Source" : "Destination";
```
allisonking, line 49: this should be lowercase. or the part that renders "Configure {source | destination} " (at least that's what I'm seeing in the design. in general, I've been trying to sentence case everything)

clients/admin-ui/src/features/common/system-data-flow/DataFlowAccordionForm.tsx
```diff
@@ -0,0 +1,193 @@
...
+  const pluralFlowType = `${flowType}s`;
+  const dataFlowSystemsModal = useDisclosure();
+  const [updateSystemMutationTrigger] = useUpdateSystemMutation();
+
+  const { data } = useGetAllSystemsQuery();
+  const systems = useMemo(() => data || [], [data]);
```
allisonking, line 55: oh this is actually a great case to use the selector `selectAllSystems`

```
// set up the subscription
useGetAllSystemsQuery();

// access the data
const systems = useAppSelector(selectAllSystems);
```

I'd actually say `selectAllSystems` should probably return an empty array if there's no data. that would simplify some code elsewhere, too.

clients/admin-ui/src/features/common/system-data-flow/DataFlowAccordionForm.tsx
```diff
@@ -0,0 +1,193 @@
...
+        paddingLeft={isSystemTab ? 6 : 2}
+        data-testid={`data-flow-button-${flowType}`}
+      >
+        <Flex
+          alignItems="center"
+          justifyContent="start"
+          flex={1}
+          textAlign="left"
+        >
+          <Text fontSize="sm" lineHeight={5} fontWeight="semibold" mr={4}>
```
allisonking, line 96: I think there might be too much `mr` here

allisonking:
Another thing—I think this flow should let me see a line on the spatial map for pixie operator, right?


https://user-images.githubusercontent.com/24641006/230983875-17eb2b45-38b1-4b67-8c32-8a6329fc30b6.mov

allisonking:
Oh never mind, I think the line _is_ there, it just doesn't show in search since the destination node doesn't have `pixie-operator` as a source. tricky problem!

clients/admin-ui/src/features/common/system-data-flow/DataFlowAccordionForm.tsx
```diff
@@ -0,0 +1,193 @@
...
+    setAssignedDataFlows(initialDataFlows);
+  }, [initialDataFlows]);
+  const handleSubmit = async ({ dataFlowSystems }: FormValues) => {
+    const updatedSystem = {
+      ...system,
+      ingress: isIngress ? dataFlowSystems : system.ingress,
+      egress: !isIngress ? dataFlowSystems : system.egress,
+    };
+    await updateSystemMutationTrigger(updatedSystem);
+    toast(successToastParams(`${pluralFlowType} updated`));
```
allisonking, line 80: let's handle the error state too :)

clients/admin-ui/src/features/datamap/SpatialDatamap.tsx
allisonking: woof this is tricky. I suspect the FE may need to dedupe these further to make the map usable. right now we get a lot of bidirectional arrows which aren't useful. for example, why do we have double arrows going back and forth from `api-service` to `profile-service`?

image


as you mentioned we can do it in a follow up, though we should check with product if it's okay to merge this one for the release without that follow up. maybe it's something the backend can help with, I'm not sure without digging into the endpoint data more

allisonking (commented):
this is a really great improvement! I peppered some thoughts throughout. the main thing is the bug about not being able to tab away from the Data flow tab. I think we should work through what's going on with all the "new" links on the map, too (at least, I haven't wrapped my head around it!). Let me know if you want to pair! 

and yes, I didn't realize this modal/assigning thing would be a pattern throughout the app. we should standardize it at some point, as you suggested.

clients/admin-ui/src/features/common/system-data-flow/DataFlowAccordionForm.tsx
```diff
@@ -0,0 +1,193 @@
...
+        <Flex
+          alignItems="center"
+          justifyContent="start"
+          flex={1}
+          textAlign="left"
+        >
+          <Text fontSize="sm" lineHeight={5} fontWeight="semibold" mr={4}>
+            {pluralFlowType}
+          </Text>
+          <QuestionTooltip label="helpful tip" />
```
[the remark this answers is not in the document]
TheAndrewJackson: Good point 😅 

@rsilvery do you have any copy for the source/destination tool tips?

clients/admin-ui/src/features/datamap/SpatialDatamap.tsx
TheAndrewJackson: I'll look the deduping code once more so see if there is anything subtle. 

I confirmed with Rachel that it's okay if we add bi-directional links in the future that contain 1 line and 2 arrows as well as fixing the zoomed out graph. Currently no one has a really complex graph like this with a ton of links so it won't show up anywhere


https://github.com/ethyca/fides/issues/3030

clients/admin-ui/src/features/common/system-data-flow/DataFlowAccordionForm.tsx
TheAndrewJackson: great idea! I went ahead and updated the selection to return an empty array along with using the selector instead

TheAndrewJackson: That seems like a good pattern to follow going forward. It makes component logic easier to follow when `undefined` never shows up

clients/admin-ui/src/features/common/system-data-flow/DataFlowAccordionForm.tsx
```diff
@@ -0,0 +1,221 @@
...
+  useEffect(() => {
+    if (mutationResult.isLoading || mutationResult.isUninitialized) {
+      return;
+    }
+
+    if (mutationResult.isError) {
+      toast(errorToastParams("Failed to update data flows"));
+    }
+    if (mutationResult.isSuccess) {
+      toast(successToastParams(`${pluralFlowType} updated`));
```
TheAndrewJackson, line 66: I had to add this janky `useEffect` to properly show the error toast. The reason was the  `handleSubmit` call has an older closure with the previous `mutationResult` values. updating to use a `useCallback` didn't worth. I looked at the docks for a better pattern and I didn't see anything that covers this scenario. I did see a mention of a global middleware that will automatically handle toasts. 

https://redux-toolkit.js.org/rtk-query/usage/error-handling#handling-errors-at-a-macro-level

It might be worth adding this in because it will remove a lot of boilerplate. The only issues I see if we need special copy for different api slices.

clients/admin-ui/src/features/common/system-data-flow/DataFlowAccordionForm.tsx
TheAndrewJackson: I only lowercased it in that one spot since the uppercase version is needed in a couple other spots

clients/admin-ui/src/features/common/system-data-flow/DataFlowAccordionForm.tsx
TheAndrewJackson: Good catch. I switched it to 2

clients/admin-ui/src/features/datamap/SpatialDatamap.tsx
```diff
@@ -34,26 +37,34 @@ const useSpatialDatamap = (rows: Row<DatamapRow>[]) => {
...
-              target: dependency,
-            })) ?? ([] as Link[])),
-        ],
-        []
-      );
+              target: egress_system,
+            })),
+        ])
+        .flatMap((link) => link)
+        .forEach((link) => links.add(JSON.stringify(link)));
```
TheAndrewJackson, line 61: @allisonking The reason why you were seeing so many duplicate links is because the set was never working... 😅 It turns out that Sets in typescript don't work on objects (unlike Rust 🙄) so I need to convert each link into a string so it does that deduping then convert them back to an object.

TheAndrewJackson:
@allisonking Everything should be fixed now 👍

clients/admin-ui/src/features/system/SystemFormTabs.tsx
```diff
@@ -120,7 +124,11 @@ const SystemFormTabs = ({
   const checkTabChange = (index: number) => {
     // While privacy declarations aren't updated yet, only apply the "unsaved changes" modal logic
     // to the system information tab
-    if (index === 0) {
+    if (
+      index === 0 ||
+      (index === 1 && tabIndex === 2) ||
```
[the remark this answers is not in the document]
allisonking: ohh okay, it's the fault of my shoddy initial attempt at an unsaved data warning 😅 thanks for finding an easy fix for it!

clients/admin-ui/src/features/common/system-data-flow/DataFlowAccordionForm.tsx
```diff
@@ -0,0 +1,223 @@
...
+  Spacer,
+  Stack,
+  Tag,
+  Text,
+  useDisclosure,
+  useToast,
+} from "@fidesui/react";
+import { FormGuard } from "common/hooks/useIsAnyFormDirty";
+import { GearLightIcon } from "common/Icon";
+// import QuestionTooltip from "common/QuestionTooltip";
```
allisonking, line 18: can remove this

clients/admin-ui/src/features/common/system-data-flow/DataFlowAccordionForm.tsx
allisonking: oh weird. sorry I didn't realize it'd be a big lift to add the error toast, otherwise we could've punted!

clients/admin-ui/src/features/common/system-data-flow/DataFlowAccordionForm.tsx
```diff
@@ -0,0 +1,223 @@
...
+  const handleSubmit = async (
+    { dataFlowSystems }: FormValues,
+    { resetForm }: FormikHelpers<FormValues>
+  ) => {
+    const updatedSystem = {
+      ...system,
+      ingress: isIngress ? dataFlowSystems : system.ingress,
+      egress: !isIngress ? dataFlowSystems : system.egress,
+    };
+    await updateSystemMutationTrigger(updatedSystem);
```
allisonking, line 107: I didn't totally understand your comment about the older closure—why can't we grab the result from the await here and then handle any errors right afterwards? 

like this pattern: https://github.com/ethyca/fides/blob/78448d9ff4d04fd279087e3889c4c728a86b1e89/clients/admin-ui/src/features/config-wizard/AuthenticateAwsForm.tsx#L102-L115

clients/admin-ui/src/features/datamap/SpatialDatamap.tsx
allisonking: ah tricky!!! nice job catching that, the map is looking cleaner 👍

clients/admin-ui/src/features/system/system.slice.ts
```diff
@@ -129,7 +129,7 @@ export const selectActiveClassifySystemFidesKey = createSelector(
 
 export const selectAllSystems = createSelector(
   systemApi.endpoints.getAllSystems.select(),
-  ({ data }) => data
+  ({ data }) => data || []
```
allisonking, line 132: we use a pattern of declaring the empty array as a const, see https://github.com/ethyca/fides/pull/1083/files#r972486065

allisonking (approved):
things are working well! left a few small things, but it's not blocking

clients/admin-ui/src/features/system/system.slice.ts
TheAndrewJackson: Oh, that's helpful and makes sense. I'll add that now

clients/admin-ui/src/features/common/system-data-flow/DataFlowAccordionForm.tsx
TheAndrewJackson: ahh,  thanks for sharing this. It worked great.

I checked what was returned by the mutation and saw it only return an object like this 

```json
{ "data": "..."}
```

so I assumed that it only returned that. I thought it would the normal result object that as `isError` and `isSuccess`.


The actually return type is something like `{"data": any} | {"error": any}`. I didn't realize that.

PR #3054, opened 2023-04-12
Update TS types

allisonking:
Just some typescript autogeneration! Backend has been moving fast, so our types are a lil out of date now (particularly in privacy notices)

### Code Changes

* [x] Ran `npm run openapi:generate` followed by `npm run format`
* [x] Moved some code that was manually put into the autogenerated folder into a safer spot
* [x] Updated the `UserPermissionResponse` to `UserPermissionPlusResponse` (see comment for more context)

### Steps to Confirm

* [ ] Everything should still build ok! Types should now properly reflect the backend.

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes


clients/admin-ui/src/types/api/models/ConnectionSystemTypeMap.ts
```diff
@@ -15,13 +14,3 @@ export type ConnectionSystemTypeMap = {
   human_readable: string;
   encoded_icon?: string;
 };
-
-export const isConnectionSystemTypeMap = (
```
allisonking, line 19: fyi @galvana I moved this from here to `datastore-connections/types.ts` instead, since this file is autogenerated so gets overwritten fairly often

clients/admin-ui/src/types/api/models/UserPermissionsPlusResponse.ts
```diff
@@ -7,8 +7,10 @@ import type { ScopeRegistryEnum } from "./ScopeRegistryEnum";
 
 /**
  * Response after creating, editing, or retrieving a FidesUserPermissions record.
+ *
+ * Overrides in Plus to use the Plus ScopeRegistryEnum so Plus scopes are returned.
  */
-export type UserPermissionsResponse = {
+export type UserPermissionsPlusResponse = {
```
allisonking, line 13: I think this should be safe, but here is what is going on:

* the backend is overwriting the permission response to a plus specific response with plus specific scopes.
* the actual type of the result is the same (still an array of `roles` and `total_scopes`), which is why the only change in this file is the name of the type. The contents of those arrays are different between plus and OSS though (where plus is a superset of OSS)
* we always run our types off of plus since it has all of the types, and this has worked well for us in the past. I think this should still be okay since there isn't really a need for the FE to know which scopes belong to plus vs OSS, at least not right now

TheAndrewJackson: This seems safe to me for now. If we run into any issues we can update plus to return a type that's only the OSS scopes

TheAndrewJackson (approved):

allisonking:
Updated the types [one more time](https://github.com/ethyca/fides/pull/3054/commits/7a1880a705b19d86cc5ce2c37c2915dbb2306569) based off of updates since last Thursday

PR #3059, opened 2023-04-12
Merge RTK APIs

allisonking:
Closes https://github.com/ethyca/fides/issues/1386

### Code Changes

* [x] Merges all of our `createApi`s into one!
  * [x] ...almost. the exception is the healthApi since that one isn't prefixed by `/api/v1`. we could write a wrapper around our baseApi if we really wanted them all to use the same one, but the healthApi is a bit of an outlier for not being under `/api/v1` imo, so I left it as-is for now
* [x] Fix some backend endpoints to use the `APIRouter` that handles trailing slashes
  * [x] This became apparent because many of the traditional "ops" query calls were not using nextjs rewrites, and instead were swapping out the query call to `localhost:8080`. this is not ideal, since then you can run into cors problems when developing locally (I think our backend might allow cors to work around this though, I haven't looked into this recently). Once I updated those endpoints to use nextjs rewrites, the trailing slash problem cropped up since next always adds a trailing slash
* [x] Consolidate data category API calls. The datamap UI had a separate API for this, now it uses the same one
* [x] Refactor `store.ts` to be more consistent and make more sense (imo 😄 )
* [x] Use consistent tag names (we had i.e. "CustomFieldDefinition" and "Data Category". I made them all space separated since RTK can handle it and I think it's more readable that way)

### Steps to Confirm

* Everything still works 🤞 Would be good to run the QA tests here since this touches nearly all of the API calls the FE makes

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes


clients/admin-ui/src/app/store.ts
```diff
@@ -5,15 +5,6 @@ import {
   StateFromReducersMapObject,
 } from "@reduxjs/toolkit";
 import { setupListeners } from "@reduxjs/toolkit/query/react";
-import {
```
allisonking, line 8: this file has a lot of changes, but they can mostly be broken down into:

* import the slice instead of the reducer or API (and use absolute imports off of `~/feature` so that the imports are grouped together properly)
  * only baseApi needs to be imported since the rest extend off of this one
* remove all the other API calls from the redux-persist blocklist and middleware (again, only baseApi needed here)
* rework the `reducers` dictionary so that they are always indexed by the slice name. this makes things more consistent, since otherwise we have to make sure the string we use is the same as the slice name 
  * i.e. `{ auth: authSlice.reducer }` vs `{[authSlice.name]: authSlice.reducer}`. in this case `authSlice.name` _is_ `auth`, but we had to do just a bit more work and write `auth` as the key. so if we ever changed the name of the slice, we'd have to update in two places. this change makes it so only the one update is needed

clients/admin-ui/src/features/common/api.slice.ts
```diff
@@ -12,19 +11,37 @@ export const baseApi = createApi({
   baseQuery: fetchBaseQuery({
     baseUrl: process.env.NEXT_PUBLIC_FIDESCTL_API,
     prepareHeaders: (headers, { getState }) => {
-      const token: string | null = selectToken(getState() as RootState);
+      const { token } = (getState() as RootState).auth;
```
allisonking, line 14: had to pull in some of the `auth` logic to avoid a circular dependency

clients/admin-ui/src/features/connection-type/connection-type.slice.ts
```diff
@@ -147,8 +135,11 @@ export const {
...
-    search: "",
-  }),
-  ({ data }) => data?.items ?? []
+  [
+    (RootState) => RootState,
+    connectionTypeApi.endpoints.getAllConnectionTypes.select({
+      search: "",
+    }),
+  ],
+  (RootState, { data }) => data?.items ?? []
```
allisonking, line 144: I had to do this change a few times which is a workaround for https://github.com/reduxjs/redux-toolkit/issues/2484

from what I understand, when code splitting, typescript doesn't understand that doing a `.endpoints.{something}.select()` is okay since it isn't part of the `RootState`. doing an initial `(RootState) => RootState` seems to make it okay though (though I think technically it's casting to `any` and I'm not sure what that is). other than that line, the following lines are all typed properly, so I think it should be okay, but this might be something we want to come back to in the future

clients/admin-ui/src/features/datamap/datamap.slice.ts
```diff
@@ -121,29 +120,6 @@ const datamapApi = baseApi.injectEndpoints({
 
 export const { useGetDatamapQuery, useLazyGetDatamapQuery } = datamapApi;
 
-const dataCategoriesApi = baseApi.injectEndpoints({
```
allisonking, line 124: replaced this chunk of code with the existing version in `taxonomy.slice`

TheAndrewJackson:
@allisonking This is awesome! It's great to have everything be unified together. 

I'm running into some issues with logging in. I'm using `nox -s "dev(slim)"` in plus. When I try to login I get a 405.

Did you experience this at all while developing?

allisonking:
> @allisonking This is awesome! It's great to have everything be unified together.
> 
> I'm running into some issues with logging in. I'm using `nox -s "dev(slim)"` in plus. When I try to login I get a 405.
> 
> Did you experience this at all while developing?

Ah yeah auth was one of the endpoints that needed the trailing slash fix, which is only in this branch. So when you do a `nox -s "dev(slim)"` in plus, it doesn't get that fix yet (at least that's my theory!)

See if things work for you using `nox -s dev` in `fides` first, then maybe we can use some of Adam's fancy new release branch tagging to test things out with plus! (I think we'd have to create new tags, and I don't know if that would interfere with the ongoing release right now)

TheAndrewJackson (approved):
@allisonking The changes not being propagated to fidesplus was the issue. Everything is working great!

PR #3081, opened 2023-04-17
Obfuscate Sensitive Database Information In Logs

ThomasLaPiana:
Closes #https://github.com/ethyca/security-issues/issues/12

After poking around and thinking about potential solutions, the following page provides the answers:

https://docs.sqlalchemy.org/en/14/core/engines.html#configuring-logging

Given that some of what is being shown is actually parameters, these logging configuration options should prevent this.

### Code Changes

* [x] add the `hide_parameters=True` flag to the SQL Alchemy Engines
* [x] add an obfuscation filter to the logs that replaces the username/password if found
* [x] toggle the obfuscation based on `dev` mode
* [x] limit the stack trace depth when running outside of `dev` mode

### Steps to Confirm

* [ ] delete a migration file and run `nox -s dev`
* [ ] see that the full stack trace, including database credentials, are visible
* [ ] turn off dev mode
* [ ] run it again and confirm that the stack trace is limited to the final error, thus not leaking the username/password in the stack trace

### Pre-Merge Checklist

* [x] Update `CHANGELOG.md`

### Description Of Changes

Small change that obfuscates logs and hides parameters. This prevents things from getting leaked into logs if there is a particularly nasty application crash. It is toggled based on `dev_mode` to keep development convenient


ThomasLaPiana:
@RobertKeyser can you try to recreate the previous issue with this branch to help me double-check my work? This _should_ solve both issues :)

src/fides/api/ctl/database/session.py
```diff
@@ -6,13 +6,21 @@
...
+    hide_parameters=True,
+    logging_name="AsyncEngine",
 )
 async_session = sessionmaker(engine, class_=AsyncSession, expire_on_commit=False)
 
-sync_engine = create_engine(CONFIG.database.sync_database_uri, echo=False)
+sync_engine = create_engine(
+    CONFIG.database.sync_database_uri,
+    echo=False,
+    hide_parameters=True,
```
ThomasLaPiana, line 21: I'm thinking if this should actually depend on the `dev` mode? Might be useful to still see these during development

src/fides/api/ctl/utils/logger.py
```diff
@@ -105,3 +106,27 @@ def setup(level: str, serialize: str = "", desination: str = "") -> None:
...
+    ]
+
+    for string in strings_to_replace:
+        result = message.replace(string, "*****")
+
+    return result
+
+
+# Loguru doesn't export the Record type so this can't be properly typed
+def formatter(record) -> str:
```
ThomasLaPiana, line 130: Maybe this should be renamed to something more obvious/clear 🤔

src/fides/api/ctl/database/database.py
```diff
@@ -35,6 +35,7 @@ def get_alembic_config(database_url: str) -> Config:
     return config
 
 
+# Prevent logging out sensitive values during failures
```
SteveDMurphy, line 38: Unfortunately I think I am still seeing the secrets, even when setting dev_mode to False. I think it might have something to do with Alembic raising the exception, maybe we can `try/except` here to either sanitize messaging or obfuscate with the loguru function you made

SteveDMurphy (commented):
I was able to reproduce the error by dropping the `ctl_organizations` table, tearing it all down, and then running `nox -s dev` again but am still seeing the error from Alembic (I flipped the env var in the compose file as well just in case). Added an idea but curious to see what you think might be happening

src/fides/api/ctl/database/database.py
ThomasLaPiana: oh interesting, thank you for figuring out how to reproduce!

I think the solution here then is to get `alembic` added to the logger then

ThomasLaPiana: for reference later: https://stackoverflow.com/questions/61415220/change-alembic-logger

ThomasLaPiana:
once #3220 is merged (it contains some server startup refactoring) I'm going to come back and add a broad try/except on the server startup (there this vuln is showing up) so prevent full traces

ThomasLaPiana:
I figured out an easy way to reproduce (delete a migration file and spin up the server). Confirmed the credential showed up in the stack trace.

Easy solution here is to limit the stack trace depth. Will make it conditional so that we can still see full traces in dev mode

SteveDMurphy (approved):
Tested this in both dev mode and not and got the expected results! Clever fix @ThomasLaPiana 🙌🏽

ThomasLaPiana:
`mypy` failure is a known issue

PR #3102, opened 2023-04-19
Update `fides deploy` to use a new `database.load_samples` setting to initialize sample Systems, Datasets, and Connections for testing

NevilleS:
### Code Changes

* [X] Move `cookie_house` sample app to `clients/cookie_house` instead of buried in the `sample_project/` dir
* [X] Fix Cypress import errors when running locally via `nox -s cypress_tests`
* [X] Add new `FIDES__DATABASE__LOAD_SAMPLES` config setting to automatically load sample data on startup
* [X] Add `load_sample_resources_from_project` that parses the resource YMLs (systems, datasets, etc), instead of relying on `fides push`
* [X] Add `load_sample_resources_from_project` that parses a (new) connections YML to configure various database & SaaS connectors , instead of relying on `load_examples.py`
* [X] Remove the `seed_example_data` step from `fides deploy`, and use the "load samples" config setting instead 👍 
* [X] Tests for all the above

### Steps to Confirm

* [X] Run `nox - s "fides_env(test)"` and confirm that sample systems, datasets, and Postgres & Mongo connectors work
* [X] Edit `.env` to include Mailchimp, Stripe, or Hubspot secrets and confirm those connectors are created
* [X] Run `nox -s e2e_test` to confirm the Cypress smoke tests still work as expected
* [X] Edit `sample_project/sample_connectors.yml` to add/remove connectors and confirm they're included in the test env

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

While continuing to overhaul the reusability of our test environment, I've found that it's generally unreliable to use Docker commands to run scripts to populate sample resources, connections, etc. It's a slightly long story, but overall the `seed_example_data` function in the `fides deploy` command is tightly coupled to the `fides` image having all the source code files available, and that's not reliable when we try to reuse this setup with different images (e.g. `fidesplus`, or custom `fides` images, etc.)

This is a fairly large refactor of the "sample data" part of the `fides deploy` script that removes the need to run scripts to make API requests against the server to setup sample data, and instead adds a new config setting, `FIDES__DATABASE__LOAD_SAMPLES` which will automatically load sample data into the database at startup time. In doing this, I also converted the "sample connectors" part of the `load_examples.py` script into a `sample_connections.yml` file that defines the connectors that should be supported by the `fides deploy` command. This makes it very easy to add/remove connectors by editing that YML file and providing the necessary secrets as ENV vars.

The messiest part here is the `seed.py` logic that imports & reuses existing API endpoints, but calls them directly instead of as part of HTTP request handling...


NevilleS:
Hmm, the `fides` container isn't coming up healthy in some of the test suites here. I'll have to troubleshoot that a bit; that shouldn't have been affected, but... 🤔

example.env
```diff
@@ -6,29 +6,19 @@
 #   2) Rename it to ".env" (which will always be ignored by git)
 #   3) Uncomment the various ENV vars below and provide secrets for your test accounts
 #
-# See the "nox -s "fides_env(test)"" command or scripts/setup/load_examples.py for
-# details on how these are used.
+# See the "nox -s "fides_env(test)"" command or browse the src/fides/data/sample_project/
+# folder (especially sample_connections/sample_connections.yml) for how these are used.
 #
-# Storage options
```
NevilleS, line 12: nb. this change will remove the ability to programmatically configure storage & messaging providers via ENV vars in the test env, which is why I removed those example variables from this file. These features were (effectively) easter eggs in the test env that I think only Roger and I knew about, so I'm pretty sure it'll be OK to do. I'll double-check with him!

noxfiles/docker_nox.py
```diff
@@ -105,7 +105,7 @@ def build(session: nox.Session, image: str, machine_type: str = "") -> None:
         session.run(
             "docker",
             "build",
-            "src/fides/data/sample_project/cookie_house",
+            "clients/cookie_house",
```
NevilleS, line 108: This is why you see a ton of moved files above 😄

requirements.txt
```diff
@@ -8,6 +8,7 @@ colorama>=0.4.3
 cryptography==38.0.3
 dask==2022.9.2
 deepdiff==5.8.1
+expandvars==0.9.0
```
NevilleS, line 11: nb. I originally used https://docs.python.org/3/library/os.path.html#os.path.expandvars, but it didn't do quite what I wanted, so I've added this library: https://pypi.org/project/expandvars/

See the pytest changes for a good picture of what this is used for.

src/fides/api/ctl/database/samples.py
```diff
@@ -0,0 +1,151 @@
...
+            elif yaml_dict.get("system", []):
+                systems = [System.parse_obj(e) for e in yaml_dict.get("system")]
+                sample_resources_dict["system"].extend(systems)
+            else:
+                if strict:
+                    unknown_resource_keys = list(yaml_dict.keys())
+                    raise NotImplementedError(
+                        f"Unexpected sample resource type, keys={unknown_resource_keys}"
+                    )
+                # quietly ignore failures when strict=False, so we only fail during tests
```
NevilleS, line 75: Quick explanation of this: I anticipate that months from now we'll add some other resource YAMLs to the project (like a custom `data_category` or something) and this code doesn't have the logic to automatically infer the Pydantic model for that and create it. To at least warn us about that, I added this `strict=True` flag and use it during the `pytest` run to cause test failures if that does happen.

src/fides/api/ctl/database/samples.py
```diff
@@ -0,0 +1,151 @@
...
+            # NOTE: in load_sample_connections, we use `expandvars` to expand
+            # ENV vars into string to allow injecting secrets, etc. We don't do
+            # that for the sample_resources/*, however, to stay consistent with
+            # the behaviour of `fides push`
+            yaml_dict = yaml.safe_load(file)
+
+            # Parse the contents of the YAML into the respective Pydantic models
+            if yaml_dict.get("dataset", []):
+                datasets = [Dataset.parse_obj(e) for e in yaml_dict.get("dataset")]
+                sample_resources_dict["dataset"].extend(datasets)
```
NevilleS, line 57: Note: this is effectively re-implementing _some_ of the guts of the logic that powers `fides push` to read resource files and deserialize them into Pydantic models. I think it's reasonable to do like this but does create some duplication that could be tech debt in the future.

src/fides/api/ctl/database/samples.py
```diff
@@ -0,0 +1,151 @@
...
+
+    While parsing the YAML file, this also performs ENV variable expansion,
+    similar to what Docker/Bash/etc. support. This allows secrets to be provided
+    as ENV var names (e.g. $FIDES_DEPLOY__CONNECTORS__POSTGRES__PASSWORD) and
+    replaced with "real" values at runtime.
+
+    After parsing the models, this function then *excludes* any connection missing
+    any of it's secrets. This prevents gotchas with failing connections at runtime
+    and also makes it easy to programmatically control which connections should be
+    enabled by simply adding/removing the secrets to the config file.
```
NevilleS, line 109: Both of these notes are pretty key - this helper will use `expandvars()` to insert secrets at runtime, and if any connection has undefined secrets they'll be ignored.

This means that if you run `fides deploy` with `FIDES_DEPLOY__CONNECTORS__MAILCHIMP_API_KEY` unset, the Mailchimp connection isn't created.

src/fides/api/ctl/database/samples.py
```diff
@@ -0,0 +1,151 @@
...
+                connection.disabled = True
+
+    # Exclude any disabled connections from the final results
+    return [e for e in sample_connections if not e.disabled]
+
+
+def load_sample_yaml_file(file: TextIO, expand_vars: bool = True) -> Dict:
+    yaml_str = file.read()
+    if expand_vars:
+        return yaml.safe_load(expandvars(yaml_str))
```
NevilleS, line 150: Note the usage of `expandvars` here! It's pretty handy, it just zooms through the whole string and replaces `$FOO` with the value of `os.environs["FOO"]` 👍

src/fides/api/ctl/database/seed.py
```diff
@@ -385,3 +404,128 @@ async def load_default_resources(async_session: AsyncSession) -> None:
     await load_default_organization(async_session)
     await load_default_taxonomy(async_session)
     load_default_dsr_policies()
+
+
+async def load_samples(async_session: AsyncSession) -> None:
```
NevilleS, line 409: This function is probably the grossest part. It effectively just repurposes other functions to take all the Pydantic models and then convert those into database objects. Some of that requires reformatting the data into formats that the various internal functions expect, etc. etc.

I tried to avoid calling any ORM creation methods directly, since I figured that'd be too easy to break.

src/fides/api/main.py
```diff
@@ -296,6 +298,7 @@ async def setup_server() -> None:
         )
     )
 
+    # TODO: move this higher up
```
NevilleS, line 301: Note to self: since we do the logging last, all the logging config isn't used for the rest of server startup... probably should be a separate ticket

src/fides/cli/commands/util.py
```diff
@@ -166,8 +165,6 @@ def up(
         check_fides_uploads_dir()
         print("> Starting application...")
         start_application()
-        print("> Setting up sample data...")
-        seed_example_data()
```
NevilleS, line 170: RIP

src/fides/cli/options.py
```diff
@@ -198,7 +198,7 @@ def prompt_username(ctx: click.Context, param: str, value: str) -> str:
         return value
 
     if config_value:
-        click.echo("> Username found in configuration file.")
+        click.echo("> Username found in configuration.")
```
NevilleS, line 201: Sometimes the configuration is in the ENV 😅

src/fides/core/deploy.py
```diff
@@ -130,17 +130,6 @@ def check_virtualenv() -> bool:
     return True
 
 
-def seed_example_data() -> None:
-    run_shell(
-        DOCKER_COMPOSE_COMMAND
-        + """exec fides /bin/bash -c "fides user login && fides push src/fides/data/sample_project/sample_resources/" """
```
NevilleS, line 136: Amusingly, this one line is basically why this PR exists.

src/fides/core/deploy.py
```diff
@@ -130,17 +130,6 @@ def check_virtualenv() -> bool:
...
-    )
-    run_shell(
-        DOCKER_COMPOSE_COMMAND
-        + """exec fides /bin/bash -c "python scripts/load_examples.py" """
```
NevilleS, line 140: Note to self: I'm inclined to delete these scripts, I think they are most likely just going to surprise us in the future as they are tightly coupled to APIs that evolve over time...

src/fides/data/sample_project/docker-compose.yml
```diff
@@ -31,6 +31,17 @@ services:
...
+      FIDES_DEPLOY__CONNECTORS__POSTGRES__HOST: postgres-test
+      FIDES_DEPLOY__CONNECTORS__POSTGRES__PORT: 5432
+      FIDES_DEPLOY__CONNECTORS__POSTGRES__DBNAME: postgres_example
+      FIDES_DEPLOY__CONNECTORS__POSTGRES__USERNAME: postgres
+      FIDES_DEPLOY__CONNECTORS__POSTGRES__PASSWORD: postgres
+      FIDES_DEPLOY__CONNECTORS__MONGO_HOST: mongodb-test
+      FIDES_DEPLOY__CONNECTORS__MONGO_PORT: 27017
+      FIDES_DEPLOY__CONNECTORS__MONGO_DEFAULTAUTHDB: mongo_test
+      FIDES_DEPLOY__CONNECTORS__MONGO_USERNAME: mongo_user
+      FIDES_DEPLOY__CONNECTORS__MONGO_PASSWORD: mongo_pass
```
NevilleS, line 44: These were previously defined in `scripts/setup/constants.py`, which was miles away from the sample project configuration. This should make it really clear how the secrets for these test databases are kept in sync with the service definitions themselves.

src/fides/data/sample_project/sample_connections/sample_connections.yml
```diff
@@ -0,0 +1,48 @@
+connection:
```
NevilleS, line 1: FWIW, this file is a very useful side-effect of this refactor. It makes it possible to trivially add a connector to the test environment for testing and inject all the necessary secrets for it as ENV vars, and do so in a pretty declarative fashion.

The format here is designed to mimic the `PATCH /api/v1/connections/` endpoint directly, but with the notable addition of the `dataset` field which isn't supported by that API.

src/fides/data/sample_project/sample_resources/sample_systems.yml
```diff
@@ -13,8 +13,7 @@ system:
       progress: Complete
       link: https://example.org/cookie_house_dpia
     privacy_declarations:
-      - name: Analyze customer behaviour for improvements.
```
NevilleS, line 16: Removed the `name` here as we're deprecating that on privacy declarations, and instead the Admin UI shows the descriptive name of the `data_use` itself.

tests/ctl/api/test_seed.py
```diff
@@ -436,3 +444,193 @@ async def test_load_orginizations(loguru_caplog, async_session, monkeypatch):
 
     assert "INSERTED 1" in loguru_caplog.text
     assert f"SKIPPED {current_orgs}" in loguru_caplog.text
+
+
+@pytest.mark.integration
+class TestLoadSampleResources:
+    """Tests related to load_sample_resources"""
```
NevilleS, line 451: Should rename this to `TestLoadSamples` as it's both resources + connections.

tests/ctl/api/test_seed.py
```diff
@@ -436,3 +444,193 @@ async def test_load_orginizations(loguru_caplog, async_session, monkeypatch):
...
+                (await async_session.execute(select(ConnectionConfig))).scalars().all()
+            )
+            dataset_configs = (
+                (await async_session.execute(select(DatasetConfig))).scalars().all()
+            )
+            assert len(systems) == 4
+            assert len(datasets) == 3
+            assert len(policies) == 1
+            assert len(connections) == 2
+            assert len(dataset_configs) == 2
```
NevilleS, line 494: For those keeping track at home:
* 4 systems (defined in sample_resources/sample_systems.yml)
* 3 datasets (2 defined in sample_resources/ for Postgres & Mongo, 1 defined in the Stripe SaaS connector template)
* 1 policy (define in sample_resources/sample_policy.yml)
* 2 connections (for this test, only Postgres & Stripe have their secrets defined)
* 2 dataset configs (Postgres & Stripe)

tests/ctl/api/test_seed.py
```diff
@@ -436,3 +444,193 @@ async def test_load_orginizations(loguru_caplog, async_session, monkeypatch):
...
+              - key: test_connection
+                name: Test Connector $TEST_VAR_1
+                connection_type: postgres
+                access: write
+                secrets:
+                  host: test-host
+                  port: 9001
+                  dbname: $TEST_VAR_2
+                  username: user-${TEST_VAR_2}
+                  password: ${TEST_VAR_1}-${TEST_VAR_2}
```
NevilleS, line 620: This is a good example of what `expandvars()` supports. It's actually more fully featured and allows for bash-isms like default values (`${FOO:-defaultvaluehere}`)...

tests/ops/api/v1/endpoints/test_connection_config_endpoints.py
```diff
@@ -225,7 +225,7 @@ def test_patch_connection_saas_with_secrets_new(
                     },
                     "name": "My Mailchimp Test",
                     "description": "Mailchimp ConnectionConfig description",
-                    "saas_connection_type": "mailchimp",
+                    "saas_connector_type": "mailchimp",
```
NevilleS, line 228: I found this while doing some searching through the codebase. Technically this API expects `saas_connector_type` so I think it was getting a bad request error, but since this test also *expects* an error, it passes. It was testing the wrong thing though 😉

NevilleS (commented):
Notes for reviewers!

src/fides/data/sample_project/sample_connections/sample_connections.yml
pattisdr: This is the best part! We didn't have a good way to set up a consent connector automatically before, could add this if interested https://github.com/ethyca/fides/pull/3104/files

src/fides/api/ctl/database/seed.py
```diff
@@ -385,3 +404,128 @@ async def load_default_resources(async_session: AsyncSession) -> None:
...
+                        saas_connector_type=connection.saas_connector_type,
+                        template_values=SaasConnectionTemplateValues.parse_obj(
+                            saas_template_data
+                        ),
+                    )
+
+                    # Check that it succeeded!
+                    connection_config = ConnectionConfig.get_by(
+                        db=db_session, field="key", value=connection.key
+                    )
```
pattisdr, line 466: Wanted to make sure this is intentional - ignoring failures instead of trying to exit when something doesn't happen as expected.

I experimented with adding a consent connector, something I've wanted for awhile, and I had the wrong `secrets` type in `sample_connections.yml`, so even though all the secrets were populated, it was still missing a required secret and failed secrets validation here.  The logs were all very clear and it was easy to track down what I was missing, but I guess I was surprised everything came up properly, just missing the bad connector.

https://github.com/ethyca/fides/pull/3104/files

src/fides/api/ctl/database/samples.py
```diff
@@ -0,0 +1,151 @@
...
+    Schema for creating a sample connection to initialize the database, usable
+    for both database & SaaS connections. Extends the existing models with an
+    optional 'dataset' field.
+    """
+
+    dataset: Optional[FidesKey] = None
+    """
+    (Optional) Valid fides_key for an existing dataset that should be used
+    for this ConnectionConfig. If a dataset does not exist for this key,
+    creation will fail.
```
pattisdr, line 90: Even with this docstring, it was still a little confusing why this was here until I traced through your creation of the connection configs -

pattisdr (commented):
I like this @NevilleS good code comments and great tests to protect your code.  It's a little duplicative in places but you've motivated it well. Thanks for the self-review.

example.env
```diff
@@ -6,29 +6,19 @@
...
-# NOTE: When Mailchimp or Stripe secrets are set, the setup scripts will configure connectors
-# MAILCHIMP_DOMAIN=
-# MAILCHIMP_USERNAME=
-# MAILCHIMP_API_KEY=
-# STRIPE_DOMAIN=
-# STRIPE_API_KEY=
\ No newline at end of file
+# Sample SaaS connections
+# NOTE: Whenever *all* the secrets for a give connection are set, 'fides deploy' will
+# instantitate the connection when it loads the sample data at startup.
```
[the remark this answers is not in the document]
NevilleS: Fixed 👍

src/fides/api/main.py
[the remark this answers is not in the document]
NevilleS: I'm going to try it out and see if anything explodes

"If not now, then when?" 😄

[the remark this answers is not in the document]
NevilleS: Yup, that worked. I went ahead and cleaned up some of the startup logging for the seed.py functions as well so it was less verbose in INFO and above.

src/fides/api/ctl/database/seed.py
```diff
@@ -21,7 +36,11 @@
...
 from fides.lib.utils.text import to_snake_case
 
-from .crud import create_resource, list_resource
+from .crud import create_resource, list_resource, upsert_resources
+from .samples import (
+    load_sample_connections_from_project,
+    load_sample_resources_from_project,
+)
 
 FIDESOPS_AUTOGENERATED_CLIENT_KEY = "default_oauth_client"
```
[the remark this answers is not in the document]
NevilleS: Done

src/fides/api/ctl/database/seed.py
NevilleS: Yes, I thought about this intentionally - the logs are there to provide the breadcrumbs for us, but I'd like that command to be pretty resilient against small misconfigurations. After all it'll be the _main_ command to use for all automated testing, so I'd prefer the whole environment still comes up but the failure is just that the connector doesn't appear. That's the difference between 1/100 failing tests and 100/100 failing tests when a small change like that is made 👍

pattisdr: > After all it'll be the main command to use for all automated testing
makes sense

src/fides/api/ctl/database/samples.py
pattisdr: this looks good!

src/fides/api/ctl/database/crud.py
```diff
@@ -38,7 +38,7 @@ async def create_resource(
             already_exists_error = errors.AlreadyExistsError(
                 sql_model.__name__, resource_dict["fides_key"]
             )
-            log.bind(error=already_exists_error.detail["error"]).error(  # type: ignore[index]
+            log.bind(error=already_exists_error.detail["error"]).info(  # type: ignore[index]
```
ThomasLaPiana, line 41: nice stealth change! I think we had an issue open for this

src/fides/api/ctl/database/samples.py
ThomasLaPiana: I want to make sure I'm not missing something here, is there a reason we can't reuse that code?

for instance, in `parse.py` there is a simple example of how this is done. It does return a proper Taxonomy object as well instead of a dict of objects

tests/ctl/api/test_seed.py
```diff
@@ -436,3 +444,193 @@ async def test_load_orginizations(loguru_caplog, async_session, monkeypatch):
...
+        "FIDES_DEPLOY__CONNECTORS__POSTGRES__PORT": "9090",
+        "FIDES_DEPLOY__CONNECTORS__POSTGRES__DBNAME": "test-var-db",
+        "FIDES_DEPLOY__CONNECTORS__POSTGRES__USERNAME": "test-var-user",
+        "FIDES_DEPLOY__CONNECTORS__POSTGRES__PASSWORD": "test-var-password",
+        "FIDES_DEPLOY__CONNECTORS__STRIPE__DOMAIN": "test-stripe-domain",
+        "FIDES_DEPLOY__CONNECTORS__STRIPE__API_KEY": "test-stripe-api-key",
+    }
+
+    @patch.dict(os.environ, SAMPLE_ENV_VARS, clear=True)
+    async def test_load_samples(
```
ThomasLaPiana, line 465: In my own personal opinion I'm not a huge fan of mega-asserts in tests. I generally find it easier to debug if tests can only fail for one specific reason. I would prefer a pattern where there is an individual test for each resource type, even if it is done via parametrization

tests/ctl/api/test_seed.py
```diff
@@ -436,3 +444,193 @@ async def test_load_orginizations(loguru_caplog, async_session, monkeypatch):
...
+        See src/fides/api/ctl/database/samples.py for details.
+
+        Sorry for the trouble, but we want to ensure there isn't a subtle bug
+        sneaking into our sample project code!
+        """
+        error_message = (
+            "Unexpected error loading sample resources; did you make changes to the sample project? "
+            "See tests/ctl/api/test_seed.py for details."
+        )
+        try:
```
ThomasLaPiana, line 546: The `try` and the `assertion` here aren't really logically connected to me. As it is currently written, it seems as if it is actually checking the validity of the `resources_dict`

I recommend something like this:
```python
try:
    samples.load_sample_resource_from_project(strict=true)
    assert True
except Exception as exc:
     print(exc)
    assert False
```

To me this makes it more clear that if the command doesn't fail, it's considered passing

tests/ctl/api/test_seed.py
NevilleS: I'd like that, but a consequence of the test setup right now is that the database is persisted (at least for ctl tests) so they aren't isolated from each other in a meaningful way... so it'd just add a lot of test time but all pass or fail as a group?

tests/ctl/api/test_seed.py
NevilleS: That makes sense, will tweak!

src/fides/api/ctl/database/crud.py
NevilleS: Yes, after some comments I felt the need to tidy up the startup logs a bit. I'm surprised you didn't comment on the _other_ stealth logging change and massive QOL improvement in this PR... 😉

tests/ctl/api/test_seed.py
NevilleS: Done

src/fides/api/ctl/database/samples.py
```diff
@@ -0,0 +1,163 @@
...
+from typing import Dict, List, Optional, TextIO, Union
+
+import yaml
+from expandvars import expandvars  # type: ignore
+from fideslang.models import Dataset, Organization, Policy, System
+from fideslang.validation import FidesKey
+
+# TODO: is there a way to use the stdlib in Python 3.9+?
+# from importlib.resources import files
+from importlib_resources import files
```
NevilleS, line 16: @ThomasLaPiana is there a better way to do this? ie. use a backported lib in Python 3.8, but the stdlib in newer versions... or do we just need to "remember" to remove this later

src/fides/api/ctl/database/samples.py
NevilleS: You're not missing anything - I think we probably could. I'm a little hesitant to trust some of the file-loading logic (there's a reason I pursued `importlib` here, it's much more reliable for loading resources from within the package itself) but I think the core YAML reading & parsing could be reused...

I'll give it a quick try now but might need an assist

src/fides/api/ctl/database/samples.py
pattisdr: this seems fine to me. Are you talking about instead something like a try/except where you try importing files from importlib.resources first and the except gets it from importlib_resources? Or conditionally importing from one or the other based on the version?

NevilleS: Yeah, something like that feels slightly better... I dunno, I think I'll just accept it. I'm overengineering aren't I

src/fides/api/ctl/database/samples.py
NevilleS: That worked! Pushed a new commit. The only side-effect of `parse.py` that I don't like are the `print()` statements in there, which don't mix & match well with the logger... but it's OK, small price to pay.

NevilleS:
OK, I fixed the problem that was causing failures in Python 3.8 (thanks, CI!) and finished tidying things up.

At this point, it's ready to merge, but it has a major caveat- it removes the ability to automatically configure the S3 & Email features. I don't want to break that for testers, so I need to just go back and rethink this.

pattisdr:
Looking..

pattisdr (approved):
great @NevilleS 🏆

example.env
NevilleS: Found a solution for this that I don't mind - I'm going to remove all these scripts from this repo (where they _imply_ some degree of maintenance, but we're not maintaining them well...) and re-implement a subset of this in our separate test tooling.

NevilleS:
Added one more tweak to check the installed Python version is one we support, but otherwise this is good to go.

I'm going to wait to merge this until I get some more folks to test it out though, especially in it's `fidesplus` variant.

Issue #3042, opened 2023-04-12
Infrastructure to share code between all our clients- create shared components-ui


eastandwestwind:
Include refactoring fides-consent to fides-js

PR #3105, opened 2023-04-19
3042: Refactor client applications into a monorepo with `turborepo`, extract `fides-js` into a standalone package, and improve `privacy-center` to load configuration at runtime

eastandwestwind:
Closes https://github.com/ethyca/fides/issues/3042

❗ Potential area of risk in deployment- https://github.com/ethyca/fides/pull/3105/files/faab9e7d25dc92c4463b2d3facde7a1858cc64ac#diff-0f3c0f74b23a0b3e4798f66a9b7ffd7f5d850165a10793f9a101235e75f1b61bR36, let's be sure to watch for this at time of deployment to ensure admin-ui is deployed appropriately to pypi

### Code Changes

- [X] Integrate [Turborepo](https://turbo.build/repo) in `/clients` so that we can share code between all clients.
- [X] Extracted out `fides-consent` from `privacy-center` into a first-class package in `clients` so that 1) we can share code from `fides-consent` between all clients, and 2) we're no longer reliant on having to build/run the `privacy-center` to use `fides-consent`.
- [X] Renamed `fides-consent` to `fides-js` to clarify that it can support other, non-consent functionality in the future (like proxying API calls to Fides, being used for DSR requests, etc.). See https://github.com/ethyca/fides/pull/3135
- [X] Refactored `privacy-center` and `fides-js` to load configuration at run-time, instead of build-time. This allows editing the `config.json` file provided to the Privacy Center and no longer requiring a `npm run build` to update the changes. See https://github.com/ethyca/fides/pull/3135
- [X] Restructured Docker builds across all environments (dev, test, prod) to accommodate the new infra to install/build/run clients together and separately.
- [X] Updated nox commands, gitlab ci commands, and front-end dev docs to call necessary turbo commands.
- [x] Change bundle output of `fides-js` to `fides.js` instead of `fides.global.js`
- [x] Remove duplicate fides-demo.html from `fides-js`
- [X] Address TODOs for `fides-js`: document `src/fides.ts`, cleanup `Fides.init`, etc.
- [x] Add (empty?) Jest test suite to `fides-js`
- [x] Fix Cypress & Jest tests for all clients
- [x] Rollback to Rollup for bundling the `fides-js` package

Follow-up issues:
- [ ] https://github.com/ethyca/fides/issues/3170
- [ ] https://github.com/ethyca/fides/issues/3171
- [ ] https://github.com/ethyca/fides/issues/3172
- [ ] https://github.com/ethyca/fides/issues/3173
- [ ] https://github.com/ethyca/fides/issues/3174
- [ ] https://github.com/ethyca/fides/issues/3175
- [ ] https://github.com/ethyca/fides/issues/3209
- [ ] https://github.com/ethyca/fides/issues/3212
- [ ] https://github.com/ethyca/fides/issues/3216

### Dev Changes
If you only use `nox` commands, continue as before, nothing needed.

If you wish to run any/all clients outside of the `nox` environment, with Turborepo as part of the codebase, you'll need to do 3 things:

1. Install Turborepo globally so that you can use it on the command line `npm install turbo --global`
2. In `clients` folder, run `npm install`. This will install the new dev dependency `turbo`, which is basically a wrapper around npm that helps manage shared dependencies. This will also install the appropriate dependencies in `clients` and in each client within `clients` folder. You may need to first remove all `node_modules` in each of the clients (admin-ui, pc, etc)
3. In short, use `turbo` instead of `npm`. E.g. instead of running `npm run dev`, use `turbo run dev`, etc. All commands that existed before should still exist, given you're using `turbo`. Let me know if something is broken for you!

Also note that to install packages in any package in `clients`:
```
npm install <package> --workspace=<workspace>
```
Example:
```
npm install react --workspace=admin-ui
```
See https://turbo.build/repo/docs/handbook/package-installation#addingremovingupgrading-packages for more details

### Steps to Confirm
* [ ] All commands used to develop against the project should still work as before

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`


.github/workflows/frontend_checks.yml
```diff
@@ -34,17 +34,16 @@ jobs:
         run: npm install
 
       - name: Lint
-        run: npm run lint
+        run: turbo run lint
```
eastandwestwind, line 37: probably have to change this back to `npm` since gitlab runner doesn't have turbo

Dockerfile
```diff
@@ -93,18 +93,28 @@ RUN pip install -e . --no-deps
...
+COPY clients/admin-ui/package.json ./admin-ui/package.json
+COPY clients/privacy-center/package.json ./privacy-center/package.json
 
-# Build the admin-ui frontend
-WORKDIR /fides/clients/admin-ui
-COPY clients/admin-ui/package.json clients/admin-ui/package-lock.json ./
 RUN npm install
-COPY clients/admin-ui/ .
+
+COPY $CLIENT_FOLDER .
```
eastandwestwind, line 111: currently, when `CLIENT_FOLDER` is `clients/privacy-center`, everything from `clients` is copied over, so need to debug this.

clients/admin-ui/next.config.js
```diff
@@ -8,16 +8,6 @@ const withBundleAnalyzer = require("@next/bundle-analyzer")({
 const nextConfig = {
     reactStrictMode: true,
     webpack(config) {
-        Object.assign(config.resolve.alias, {
```
allisonking, line 11: out of curiosity, what was the motivation for removing this? I don't actually know what it does, so just wondering if you figured it out

clients/admin-ui/package.json
```diff
@@ -10,7 +10,7 @@
     "copy-export": "mkdir -p ../../src/fides/ui-build/static/admin/ && rsync -a --delete out/ ../../src/fides/ui-build/static/admin/",
     "cy:open": "cypress open",
     "cy:run": "cypress run",
-    "cy:start": "NODE_ENV=test next build && NODE_ENV=test next start",
+    "cy:start": "NODE_ENV=test next start",
```
allisonking, line 13: hmm I think we do still want the `build` here, otherwise next might not serve up the latest thing?

clients/cypress.config.ts
```diff
@@ -0,0 +1,22 @@
+/* eslint-disable import/no-extraneous-dependencies */
```
allisonking, line 1: hm it's kind of weird this file showed up! do you know why that happened?

noxfiles/docker_nox.py
```diff
@@ -78,6 +79,7 @@ def build(session: nox.Session, image: str, machine_type: str = "") -> None:
         "test": {"tag": lambda: IMAGE_LOCAL, "target": "prod"},
         "dev": {"tag": lambda: IMAGE_LOCAL, "target": "dev"},
         "admin_ui": {"tag": lambda: IMAGE_LOCAL_UI, "target": "frontend"},
+        "privacy_center": {"tag": lambda: IMAGE_LOCAL_PC, "target": "frontend"}
```
allisonking, line 82: it might make sense to have a separate stage for privacy center, since often we're running only the admin ui

Dockerfile
```diff
@@ -93,19 +93,63 @@ RUN pip install -e . --no-deps
...
-RUN npm run export
+
+# Builds and exports admin-ui
+RUN npm run export-ui
+# Builds privacy-center
+RUN npm run build-pc
+
+#############################
+## Production PC ##
+#############################
```
allisonking, line 123: are we moving away from PC having its own docker file?

Dockerfile
```diff
@@ -93,19 +93,63 @@ RUN pip install -e . --no-deps
 ###################
 ## Frontend Base ##
 ###################
-FROM node:16-slim as frontend
+FROM node:16-alpine as frontend
+
+RUN apk add --no-cache libc6-compat
+# Build the frontend clients
```
allisonking, line 99: I'm not convinced that we want to be building PC in this dockerfile since pc is a separate app and already has its own dockerfile. to me this Dockerfile is just for `fides` and should be pretty similar to what the python package can serve up, but with a few helpful stages. maybe I'm missing something and can still be convinced though :)

allisonking (commented):
woo turbo!! nice job working through setting this up!

Dockerfile
eastandwestwind: yup! No more dockerfile for just privacy-center

Dockerfile
eastandwestwind: So PC is still deployed as a separate app, which is what we do in our `prod_pc` stage of this Dockerfile. 

With my changes, both PC and admin-ui require the same `frontend` base target now, which includes everything in `clients`. This is because using Turborepo, we need to install and build everything from the base `clients` folder. The benefit of using the same Dockerfile for both PC and admin-ui is that we get to re-use this cached `frontend` image whenever we run or deploy both PC and admin-ui, saving build time.

Could you maybe explain more about the downfalls of having this Dockerfile build PC too?

.github/workflows/publish_package.yaml
```diff
@@ -27,13 +27,13 @@ jobs:
...
-          cd clients/admin-ui
+          cd clients
           npm install
 
       - name: Build and export frontend files
         run: |
-          cd clients/admin-ui
-          npm run prod-export
+          cd clients
+          npm run prod-export-ui
```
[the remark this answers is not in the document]
eastandwestwind: yup, this `npm run prod-export-ui` only exports admin-ui. Maybe I should rename to `prod-export-admin-ui` to be clearer?

clients/admin-ui/next.config.js
eastandwestwind: Basically, we can't mess with how turbo is resolving the `node_modules`, as it has a broader view from the `clients` directory. Without removing these lines, we'll run into the following err upon build (this is an example err from PC, but the same will apply to admin-ui): 

```
privacy-center:build: Failed to compile.
privacy-center:build: 
privacy-center:build: ../node_modules/@chakra-ui/accordion/dist/chakra-ui-accordion.esm.js
privacy-center:build: Module not found: Can't resolve 'react' in '/Users/neville/git/fides/clients/node_modules/@chakra-ui/accordion/dist'
```

clients/admin-ui/package.json
eastandwestwind: Oh so I've moved this dependency on `build` to the root `turbo.json` file - https://github.com/ethyca/fides/pull/3105/files#diff-c8d74f2ac008193a3c9c6bfe765cccd0b4cf6413b857ff5ca4d8b27b6bee1ef0R41

clients/cypress.config.ts
eastandwestwind: Ah I don't think I need this file anymore, thanks!

noxfiles/docker_nox.py
eastandwestwind: Just so I understand- you mean have a new `target` for privacy-center instead of using the `frontend` target?

Right now I'm using the same stage for both, because they both require an install at the root `clients` folder. Even if we only run the admin-ui, we'll still need to do the same steps anyway. See https://github.com/ethyca/fides/pull/3105/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R96

Dockerfile
ThomasLaPiana: While generally I do like separation of concerns, I think there is enough overlap here to justify combining the Dockerfiles if it means faster build/dev times and is well documented!

Dockerfile
```diff
@@ -93,19 +93,63 @@ RUN pip install -e . --no-deps
...
 ####################
 FROM frontend as built_frontend
-RUN npm run export
+
+# Builds and exports admin-ui
+RUN npm run export-ui
+# Builds privacy-center
+RUN npm run build-pc
+
+#############################
```
ThomasLaPiana, line 121: The nittiest of all nits:

The `#` should match the length of the title line

noxfiles/docker_nox.py
```diff
@@ -88,20 +90,21 @@ def build(session: nox.Session, image: str, machine_type: str = "") -> None:
             tag_name = get_current_tag()
         if image == "test":
             tag_name = "local"
-        privacy_center_image_tag = f"{PRIVACY_CENTER_IMAGE}:{tag_name}"
-        sample_app_image_tag = f"{SAMPLE_APP_IMAGE}:{tag_name}"
-
         session.log("Building extra images:")
+        # prod/test PC is pushed to ethyca/fides-privacy-center image
```
ThomasLaPiana, line 94: This is already documented in `push`, I don't think it is useful documentation to have in `dev`

noxfiles/docker_nox.py
```diff
@@ -111,35 +114,24 @@ def build(session: nox.Session, image: str, machine_type: str = "") -> None:
...
-            "--tag",
-            tag(),
-            ".",
-        )
-        if "nocache" in session.posargs:
-            build_command = (*build_command, "--no-cache")
-
-        session.run(*build_command, external=True)
+    # Build the main ethyca/fides image
+    # E.g. of PC command: docker build --target=frontend --platform linux/arm64 --tag ethyca/fides:local-pc .
```
ThomasLaPiana, line 118: This feels redundant since `nox` will log out the command it is running anyway

ThomasLaPiana:
@eastandwestwind I mostly just looked over the Docker/Noxfile updates, great work!

noxfiles/docker_nox.py
eastandwestwind: good point, I'll remove

noxfiles/docker_nox.py
eastandwestwind: this was leftover from my own notes while testing, will remove!

Dockerfile
eastandwestwind: hahaha will fix thanks!

.github/workflows/publish_package.yaml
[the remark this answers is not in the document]
allisonking: oh! yeah I guess I got tripped up by the name 😄

clients/admin-ui/package.json
allisonking: thanks for pointing that out! I guess this is something that wouldn't work well if you tried to use npm commands straight inside the admin-ui package (which I know you said not to do! just thinking about if I were a React dev looking at our code I'd probably hop right into admin-ui and try to run some of these commands). this one would be particularly tricky to figure out since you might still serve up a page, just not the latest version of the page.

perhaps you're already working on this, but a README in the `clients` folder really hammering in to use `turbo` could address this. also I know there's a `ui.md` doc that needs an all around update (not just for turbo, but it still references the fidesops repo, so it's very outdated!). I saw you replaced the npm commands with turbo commands, but we probably want a doc with a general overview of turbo and how we use it

Dockerfile
allisonking: ah yeah, I see your point. I keep forgetting that they will all share `node_modules` now. 

although I imagine you could do probably install only the admin ui dependencies, then run just the admin ui here? admin-ui and privacy-center share some dependencies but admin-ui has a lot more than privacy-center does, so if you're just trying to run the privacy center via nox, you'll end up needing to install a bunch of packages you don't need. so it's kind of like yes, they can share the same `frontend` stage, but only because the `frontend` stage is installing more things

that said, I rarely run the frontends through nox/docker since the hot reloading is slower. so we likely won't notice this kind of thing, but just want to call it out so we are aware!

all that to say, let's use what you have here and we can keep iterating if needed :)

Dockerfile
```diff
@@ -93,19 +93,63 @@ RUN pip install -e . --no-deps
...
 ####################
 ## Built frontend ##
 ####################
 FROM frontend as built_frontend
-RUN npm run export
+
+# Builds and exports admin-ui
+RUN npm run export-ui
+# Builds privacy-center
+RUN npm run build-pc
```
allisonking, line 119: another docker optimization thing which we might not need right now, but it might be worth splitting this one into two stages, one for admin ui and one for privacy center. building UI is relatively slow, and we never use _both_ the built admin ui and the built privacy center in the same subsequent stage 

so maybe something like

```Dockerfile
####################
##  Built admin UI ##
####################
FROM frontend as built_admin_ui

# Builds and exports admin-ui
RUN npm run export-ui

####################
## Built privacy center ##
####################
FROM frontend as built_privacy_center
# Builds privacy-center
RUN npm run build-pc

#############################
## Production PC ##
#############################
FROM node:16-alpine as prod_pc
...
COPY --from=built_privacy_center /fides/clients /fides/clients

...

#############################
## Production Application ##
#############################
FROM backend as prod

# Copy frontend build over
COPY --from=built_admin_ui /fides/clients/admin-ui/out/ /fides/src/fides/ui-build/static/admin
```

eastandwestwind: > we never use both the built admin ui and the built privacy center in the same subsequent stage

We don't actually build from dockerfile as part of dev (e.g. `nox -s dev -- pc` stops at the `frontend` target in `Dockerfile`). For this command, the build step happens as a Turbo dependency of when the `docker.compose.yml` calls `npm run dev-pc`.

The only time we run the `built_frontend` stage is for the prod and test (`nox -s "fides_env(test)"`) environments, in which we _do_ subsequently use both the built admin ui and build privacy center for the prod builds (even though those builds are technically separate images).

NevilleS: I think that separation makes good sense - and if we keep moving in this direction, we can overall refactor the `build` command to use distinct application names like `build(privacy-center)`, `build(admin-ui)`, `build(fides`), etc.

We did this for the `fidesplus` nox setup and doing the same thing here would be make sense. That can be a follow-up though, no need to expand scope too much here 😄

Dockerfile
```diff
@@ -93,19 +93,63 @@ RUN pip install -e . --no-deps
...
+
+#############################
+## Production PC ##
```
NevilleS, line 122:
```suggestion
## Production Privacy Center ##
```

I'd like us to avoid the "pc" short form, it's not very intuitive 😄

Dockerfile
```diff
@@ -93,19 +93,63 @@ RUN pip install -e . --no-deps
...
+ENV NODE_ENV production
+ENV NEXT_TELEMETRY_DISABLED 1
+
+RUN addgroup --system --gid 1001 nodejs
+RUN adduser --system --uid 1001 nextjs
+
+RUN chown nextjs:nodejs /fides/clients
+
+# We need to copy everything so we can rebuild with the new configs if needed
+COPY --from=built_frontend /fides/clients /fides/clients
```
NevilleS, line 137: I really want to get rid of this rebuild logic! Going to poke at it this ASAP... 😄

Dockerfile
eastandwestwind: Ah I see what you're saying @allisonking , and that would be ideal if we could just install the dependencies for 1 workspace e.g. `privacy-center`. I'm digging in docs a bit, and everything seems to recommend installing everything at root https://turbo.build/repo/docs/handbook/workspaces#managing-workspaces, and I'm unsure if there's a way to filter out specific workspaces here. This is because install itself uses `npm`, not turborepo. Turbo itself is capable of filtering workspaces.

clients/fides-consent/package.json
```diff
@@ -18,7 +18,8 @@
     "url": "https://github.com/ethyca/fides.git",
     "directory": "clients/privacy-center/packages/fides-consent"
   },
-  "module": "./dist/fides-consent.mjs",
+  "main": "dist/fides-consent.mjs",
+  "module": "dist/fides-consent.mjs",
```
NevilleS, line 22: Warning: I'm going to get really annoying and ask to do a lot of renaming here... (e.g. fides-consent -> fides-components) but let's stabilize first 😄

clients/fides-consent/rollup.config.js
```diff
@@ -20,9 +20,9 @@ const generateConsentConfig = () => {
   );
 
   /**
-   * @type {import('../../types/config').Config}
+   * @type {import('privacy-center/types/config').Config}
    */
-  const privacyCenterConfig = require("../../config/config.json");
+  const privacyCenterConfig = require("privacy-center/config/config.json");
```
NevilleS, line 25: Oof, this is some circular stuff... again, not strictly related to this change but evidence that we need to decouple this from the *build*.

clients/package.json
```diff
@@ -0,0 +1,33 @@
+{
+  "workspaces": [
+    "admin-ui/",
+    "privacy-center/",
+    "fides-consent/"
+  ],
+  "scripts": {
+    "build": "turbo run build",
+    "build-admin-ui": "turbo run build --filter=admin-ui",
+    "build-pc": "turbo run build --filter=privacy-center",
```
NevilleS, line 10:
```suggestion
    "build-privacy-center": "turbo run build --filter=privacy-center",
```

Sorry but I think it'll be important to be clear about this 😄 

Also, shouldn't you have some options for building the fides-consent package?

clients/privacy-center/package.json
```diff
@@ -1,11 +1,9 @@
 {
-  "name": "fidesops-privacy-center",
+  "name": "privacy-center",
```
NevilleS, line 2: Thanks! 😄

clients/privacy-center/packages/fides-consent/src/consent-config.json
```diff
@@ -0,0 +1,31 @@
+{
+  "options": [
```
NevilleS, line 2: This is auto-generated, isn't it? Did a .gitignore get missed somewhere maybe?

src/fides/data/sample_project/docker-compose.yml
```diff
@@ -61,7 +61,7 @@ services:
     volumes:
       - type: bind
         source: ./privacy_center/config
-        target: /app/config
+        target: /fides/clients/privacy-center/config
```
NevilleS, line 64: Note to self: we're pretty sure this doesn't work right now.

NevilleS: ...and I'd like to preserve this option to mount to `/app/config` for backwards-compat as per our docs: https://docs.ethyca.com/fides/dsr_quickstart/privacy_center

src/fides/data/sample_project/privacy_center/config/examples/basic.json
```diff
@@ -0,0 +1,36 @@
+{
+  "title": "Privacy center",
```
NevilleS, line 2: This looks wrong - no need to include all these in the sample privacy center config. Can you revert all the changes to the `sample_project` folder, basically?

src/fides/data/sample_project/docker-compose.yml
NevilleS: Let's revert this change for now and keep it with `/app/config`.

NevilleS (changes requested):
Couple comments to get this to a stable place 👍

clients/fides-consent/package.json
eastandwestwind: yeahhh I was unsure whether to tackle that renaming here or in another PR- but I suppose we could do here after we're at a stable place

clients/admin-ui/package.json
eastandwestwind: Good idea! I'll get that added

.github/workflows/publish_package.yaml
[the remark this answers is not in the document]
eastandwestwind: updated!

clients/fides-consent/rollup.config.js
eastandwestwind: yes! This is what I was talking about on our call - not great to have this circular dep

.github/workflows/frontend_checks.yml
```diff
@@ -61,23 +60,32 @@ jobs:
...
+            start: npm run cy-pc:start
+            wait-on: "http://localhost:3000"
+            wait-on-timeout: 180
+
       - uses: actions/upload-artifact@v3
         if: failure()
         with:
           name: cypress-videos
-          path: /home/runner/work/fides/fides/clients/admin-ui/cypress/videos/*.mp4
+          path: /home/runner/work/fides/fides/clients/cypress-e2e/videos/*.mp4
```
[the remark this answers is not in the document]
eastandwestwind: oh.. I see, let me look into how we're adding the vids for the other clients, and get these actions added.

clients/package.json
eastandwestwind: I can add some. Right now, because `fides-consent` is a dependency of privacy center, it gets automatically built as part of `build-privacy-center`. Right now, you can `cd` into `fides-consent` and run `turbo run build` and that will work. The npm stuff in base `package.json` is more for convenience, particularly when needing to run a particular command across all workspaces

src/fides/data/sample_project/privacy_center/config/examples/basic.json
eastandwestwind: yup, not sure what happened here

Dockerfile
```diff
@@ -93,19 +93,63 @@ RUN pip install -e . --no-deps
...
+FROM node:16-alpine as frontend
+
+RUN apk add --no-cache libc6-compat
+# Build the frontend clients
+WORKDIR /fides/clients
+COPY clients/package.json ./
+COPY clients/package-lock.json ./
+COPY clients/fides-consent/package.json ./fides-consent/package.json
+COPY clients/admin-ui/package.json ./admin-ui/package.json
+COPY clients/privacy-center/package.json ./privacy-center/package.json
```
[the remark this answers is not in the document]
eastandwestwind: I'm struggling to find a `COPY` command that handles multiple origin files AND multiple destinations. From what I can see, it's only possible to copy multiple files into a common destination.

E.g. I can update lines 101-102 to `COPY clients/package.json clients/package-lock.json ./`

Maybe I'm missing it in the [docs?](https://docs.docker.com/engine/reference/builder/#copy)

src/fides/data/sample_project/docker-compose.yml
eastandwestwind: Reverting this causes the following error upon `"fides_env(test)"`:

```
2023-04-21 17:20:35 > start
2023-04-21 17:20:35 > next start
2023-04-21 17:20:35 
2023-04-21 17:20:36 ready - started server on 0.0.0.0:3000, url: http://localhost:3000
2023-04-21 17:20:37 error - Failed to load next.config.js, see more info here https://nextjs.org/docs/messages/next-config-error
2023-04-21 17:20:37 Error: Cannot find module '../config/config.json'
2023-04-21 17:20:37 Require stack:
2023-04-21 17:20:37 - /fides/clients/privacy-center/scripts/validate-config.js
2023-04-21 17:20:37 - /fides/clients/privacy-center/next.config.js
2023-04-21 17:20:37     at Function.Module._resolveFilename (node:internal/modules/cjs/loader:1026:15)
2023-04-21 17:20:37     at Function.mod._resolveFilename (/fides/clients/privacy-center/node_modules/next/dist/build/webpack/require-hook.js:27:32)
2023-04-21 17:20:37     at Function.mod._resolveFilename (/fides/clients/privacy-center/node_modules/next/dist/build/webpack/require-hook.js:27:32)
2023-04-21 17:20:37     at Function.Module._load (node:internal/modules/cjs/loader:871:27)
2023-04-21 17:20:37     at Module.require (node:internal/modules/cjs/loader:1098:19)
2023-04-21 17:20:37     at require (node:internal/modules/cjs/helpers:108:18)
2023-04-21 17:20:37     at validateConfig (/fides/clients/privacy-center/scripts/validate-config.js:31:31)
2023-04-21 17:20:37     at Object.<anonymous> (/fides/clients/privacy-center/next.config.js:4:1)
2023-04-21 17:20:37     at Module._compile (node:internal/modules/cjs/loader:1196:14)
2023-04-21 17:20:37     at Object.Module._extensions..js (node:internal/modules/cjs/loader:1250:10) {
2023-04-21 17:20:37   code: 'MODULE_NOT_FOUND',
2023-04-21 17:20:37   requireStack: [
2023-04-21 17:20:37     '/fides/clients/privacy-center/scripts/validate-config.js',
2023-04-21 17:20:37     '/fides/clients/privacy-center/next.config.js'
2023-04-21 17:20:37   ]
2023-04-21 17:20:37 }
```

Can we update this once we fix the ability to load in config at runtime?

clients/privacy-center/packages/fides-consent/src/consent-config.json
eastandwestwind: We actually need to remove the entire `clients/privacy-center/packages/` directory, as that's basically empty now that we've extracted out `fides-consent` from privacy center

clients/fides-consent/rollup.config.js
```diff
@@ -65,12 +65,6 @@ export default [
       esbuild({
         minify: !isDev,
       }),
-      copy({
-        // Automatically add the built script to the privacy center's static files for testing:
```
eastandwestwind, line 69: note to self- pretty sure we still need this. This was previously added to `privacy-center/public`

.github/workflows/frontend_checks.yml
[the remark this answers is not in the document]
allisonking: I don't think we upload from the other clients. I think this was put in originally because we were debugging something specific to `admin-ui`. it's rare to need to debug something that is only failing in CI, so we never bothered to put it in for the other clients.

ideally, instead of using this `upload-artifact` workaround, we should configure cypress dashboard for all of the clients, since dashboard automatically records videos. but I think we'd need more than the free plan to do that.

for now, maybe we can leave it with the old path and expect only admin-ui uploads? or was there a reason for the change to `cypress-e2e`?

src/fides/data/sample_project/docker-compose.yml
NevilleS: Since I merged in https://github.com/ethyca/fides/pull/3135, this is addressed 👍

Dockerfile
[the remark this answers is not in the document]
NevilleS: Yup, you're right! Only so much we can do there. In general let's just defer "docker optimizations" to another time, it's not critical-path stuff. I have a mega-comment in the Dockerfile now with some thoughts

Dockerfile
NevilleS: There's some really cool & interesting tools for optimizing the Docker builds with Turborepo. The docs for this are well written too: https://turbo.build/repo/docs/handbook/deploying-with-docker

The TL;DR is that it's entirely possible to use turborepo to:
* "prune" the node_modules / dependencies / etc for a single app in the monorepo (using `turbo prune`)
* copy over the build dependencies to a Docker layer (optimally) using `turbo prune --docker`
* trace the exact built files needed for the production server
* etc.

I made a "note to self" about these in a comment here: https://github.com/ethyca/fides/blob/80ec6695f1f636a7e8a32935a411b8f9b69769b8/Dockerfile#L134

My plan is to just farm this out into a separate follow-up ticket though. Let's get the monorepo working 👍

Dockerfile
NevilleS: To resolve this, I'll create a ticket for Docker build optimization

noxfiles/docker_nox.py
NevilleS: Same comment as the other threads - a new ticket for Docker optimization makes sense, and I agree that having some app-specific build stages will work best.

NevilleS:
Hey @eastandwestwind: I cleaned up most of my nonsense, fixed lint/formatting/etc errors, and got all the CI checks running again. As of now there's just one more failing, which is the Cypress tests for privacy center that I broke when I refactored how config is loaded!

I kept a few more TODOs open on the PR but I think this is pretty stable now. I'm still nervous to merge as I worry there is a breaking change in here that we're missing so far, but this is getting close to landing I think.

eastandwestwind:
Oh excellent! @NevilleS - Should we plan to further clean up / test this branch to merge into `main` ahead of any consent banner infra changes?

clients/sample-app/src/components/PurchaseModal/index.tsx
```diff
@@ -2,19 +2,17 @@ import React, {
   memo, useEffect, useRef, useState,
 } from 'react';
 import { useForm } from 'react-hook-form';
-import { Product, UserData } from '../../types';
```
NevilleS, line 5: Unrelated lint fixes ;)

clients/sample-app/src/lib/db.ts
```diff
@@ -1,18 +1,13 @@
 import { Pool } from 'pg';
 
 const { DATABASE_HOST, DATABASE_PORT, DATABASE_USER, DATABASE_PASSWORD, DATABASE_DB } = process.env;
-if (!DATABASE_HOST) throw new Error('No DATABASE_HOST');
-if (!DATABASE_PORT) throw new Error('No DATABASE_PORT');
-if (!DATABASE_USER) throw new Error('No DATABASE_USER');
-if (!DATABASE_PASSWORD) throw new Error('No DATABASE_PASSWORD');
-if (!DATABASE_DB) throw new Error('No DATABASE_DB');
```
NevilleS, line 8: I moved defaults for all of these just straight into the app. This removes the need to use the Dockerfile to spin this up and use it, though it'll basically just not work if you don't have a DB running!

...but it'll build & lint which is what I wanted!

clients/fides-js/tsconfig.json
```diff
@@ -0,0 +1,24 @@
...
+    "isolatedModules": true,
+    "jsx": "preserve",
+    "lib": ["dom", "dom.iterable", "esnext"],
+    "module": "esnext",
+    "moduleResolution": "node",
+    "noEmit": true,
+    "resolveJsonModule": true,
+    "skipLibCheck": true,
+    "strict": true,
+    "target": "es6",
```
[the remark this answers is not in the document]
allisonking: > Note: JSON doesn't allow comments

this is true, but tsconfig actually does https://github.com/microsoft/TypeScript/issues/4987 (I haven't tried it, hopefully there aren't any sneaky compiling issues)

clients/privacy-center/cypress/e2e/consent.cy.ts
```diff
@@ -333,18 +293,12 @@ describe("Consent settings", () => {
       });
 
       it("lets the user consent to override GPC", () => {
-        cy.visit("/consent?globalPrivacyControl=true");
```
allisonking, line 336: is this flag `globalPrivacyControl=true` still somehow captured?

clients/privacy-center/cypress/support/commands.ts
```diff
@@ -79,7 +79,18 @@ declare global {
             Cypress.Withinable &
             Cypress.Shadow
         >
-      ): Chainable<JQuery<HTMLElement>>;
+      ): Chainable<boolean>;
+      /**
+       * Custom command to load a Privacy Center configuration JSON file from a fixture.
+       *
```
allisonking, line 85: might be helpful to add a comment about the footgun here. fixtures normally persist across `.visit()`s, I guess in this case it isn't persisting because it's actually going into redux?

clients/privacy-center/package.json
```diff
@@ -19,20 +17,16 @@
     "analyze": "cross-env ANALYZE=true next build",
     "analyze:server": "cross-env BUNDLE_ANALYZE=server next build",
     "analyze:browser": "cross-env BUNDLE_ANALYZE=browser next build",
-    "build-privacy-portal": "npm run build",
     "openapi:generate": "openapi --input http://localhost:8080/openapi.json --output ./types/api --exportCore false --exportServices false --indent 2"
   },
-  "workspaces": [
-    "packages/fides-consent"
-  ],
```
allisonking, line 27: 🎉

clients/turbo.json
```diff
@@ -0,0 +1,56 @@
+{
+  "$schema": "https://turbo.build/schema.json",
+  "pipeline": {
+    "build": {
+      "dependsOn": ["^build"],
+      "outputs": [".next/*", "!.next/cache/*"]
+    },
+    "dev": {
+      "dependsOn": ["build"],
```
allisonking, line 9: I don't think you need to `npm run build` before doing an `npm run dev`, you should only need `npm run install`—or am I misinterpreting how this config works?

clients/turbo.json
```diff
@@ -0,0 +1,56 @@
...
+    },
+    "test:ci": {
+      "dependsOn": ["build"]
+    },
+    "cy:open": {
+      "dependsOn": ["build"]
+    },
+    "cy:run": {
+      "dependsOn": ["build"]
+    },
```
allisonking, line 39: same thing here, I don't think either of these depend on `npm run build`

clients/turbo.json
```diff
@@ -0,0 +1,56 @@
...
+    "analyze": {
+      "dependsOn": ["build"]
+    },
+    "analyze:browser": {
+      "dependsOn": ["build"]
+    },
+    "analyze:server": {
+      "dependsOn": ["build"]
+    },
```
allisonking, line 51: these might not either, but I'm less sure about that

allisonking (commented):
whew this is a lot 😅 but a great improvement!! haven't started playing around with it yet, going to leave some comments first and then test it out 👍

allisonking:
played around with the test env yesterday and things seemed to be working well! I'd vote to :shipit: and then iron things out as we really start using it this week

clients/fides-js/__tests__/lib/cookie.test.ts
```diff
@@ -0,0 +1,146 @@
...
+        saveFidesCookie(cookie);
+        // NOTE: signature of the setCookie fn is: setCookie(name, value, attributes, encoding)
+        expect(mockSetCookie.mock.calls).toHaveLength(1);
+        expect(mockSetCookie.mock.calls[0][0]).toEqual("fides_consent"); // name
+        expect(mockSetCookie.mock.calls[0][1]).toEqual(expectedCookieString); // value
+        expect(mockSetCookie.mock.calls[0][2]).toHaveProperty("domain", "localhost"); // attributes
+        expect(mockSetCookie.mock.calls[0][2]).toHaveProperty("expires", 365); // attributes
+    });
+
+    it.each([
```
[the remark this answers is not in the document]
TheAndrewJackson: Me either. That's pretty cool!

clients/privacy-center/app/server-environment.ts
```diff
@@ -0,0 +1,210 @@
...
+export const loadConfigFromFile = async (configJsonUrl?: string): Promise<Config | undefined> => {
+  const urls = [
+    configJsonUrl,
+    "file:///app/config/config.json",
+    "file:./config/config.json",
+  ];
+  const file = await loadConfigFile(urls);
+  if (file) {
+    const config = JSON.parse(file) as Config;
+    // TODO: validate config?
```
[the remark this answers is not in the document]
TheAndrewJackson: [Zod](https://github.com/colinhacks/zod) would be helpful to validate this. It's essentially pydantic but for typescript. It's super powerful! 


This might helpful too. There is a package called [t3-env](https://github.com/t3-oss/t3-env) that will uses a zod schema to validate env vars on startup. If any are missing or invalid it will fail startup and let the user know what's missing.

TheAndrewJackson (approved):
Things seemed to working well! I agree with what @allisonking said. It's probably worth shipping it and ironing out the kinks this week.

I'm looking forward to using this

clients/admin-ui/cypress/.eslintrc.json
```diff
@@ -1,11 +1,12 @@
...
-  "plugins": ["no-only-tests"],
+  "plugins": ["no-only-tests", "cypress"],
   "parserOptions": {
     "project": "./tsconfig.json",
     "tsconfigRootDir": "cypress"
   },
   "rules": {
-    "no-only-tests/no-only-tests": "error"
+    "no-only-tests/no-only-tests": "error",
+    "cypress/unsafe-to-chain-command": "off"
```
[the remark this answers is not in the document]
eastandwestwind: eslint was complaining about some pre-existing code due to some chaining we were doing in cypress, so I turned it off

clients/admin-ui/turbo.json
```diff
@@ -0,0 +1,26 @@
+{
+  "extends": ["//"],
+  "pipeline": {
```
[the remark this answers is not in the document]
eastandwestwind: This file defines turbo commands unique to the `admin-ui` workspace, so we should keep them scoped here

clients/turbo.json
eastandwestwind: I think you're right, I'll remove

eastandwestwind: this config basically says the `build` pipeline step needs to run for all applicable workspaces before any `dev` commands run

clients/fides-js/rollup.config.mjs
```diff
@@ -0,0 +1,90 @@
...
+              );
+              if (isDev) {
+                process.exit(1);
+              }
+            } else {
+              console.log(
+                `✅ ${fileName} gzipped size passed maximum size checks (${gzipSize} < ${GZIP_SIZE_ERROR_KB} KB)`
+              );
+            }
+          },
```
NevilleS, line 55: FYI: I just wired this up to keep our eye on bundle size. I set a max target of 20 KB (compressed), which is still pretty big, but didn't want to overdo it yet.

.github/workflows/publish_package.yaml
```diff
@@ -27,13 +27,13 @@ jobs:
...
-          cd clients/admin-ui
+          cd clients
           npm install
 
       - name: Build and export frontend files
         run: |
-          cd clients/admin-ui
-          npm run prod-export
+          cd clients
+          npm run prod-export-admin-ui
```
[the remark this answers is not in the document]
eastandwestwind: Good call. I'll add to top of ticket as an area of potential risk.

clients/fides-js/rollup.config.mjs
eastandwestwind: This is great! I feel way better having bundle size checks automated  🚀

clients/turbo.json
eastandwestwind: you're right, I've removed!

clients/privacy-center/next.config.js
```diff
@@ -10,19 +10,29 @@ const withBundleAnalyzer = require("@next/bundle-analyzer")({
 
 const nextConfig = {
   reactStrictMode: true,
+  // TODO: I think we want this
+  // transpilePackages: ["fides-js"],
```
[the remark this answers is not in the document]
eastandwestwind: updated todo with a more descriptive comment

.github/workflows/publish_package.yaml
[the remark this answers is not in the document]
NevilleS: I double-posted here too for paranoia: https://github.com/ethyca/fides/issues/3042#issuecomment-1533435778

Thanks!

clients/privacy-center/cypress/support/commands.ts
NevilleS: I think it may actually be a side-effect of this, honestly: https://github.com/ethyca/fides/issues/3212...

clients/fides-js/tsconfig.json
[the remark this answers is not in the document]
NevilleS: Hah- that's amazing, what an incredible hack! TIL @allisonking, I love that kind of random trivia

NevilleS (approved):
OK, I think this is ready to go. Lots of follow-up issues to tackle though, so before we merge we'll just need to figure out a safe release plan!

PR #3117, opened 2023-04-19
[#3103] update `RuleTarget.data_category` API validation

### LINKED ISSUE #3103: Rule Targets should allow for custom data categories
Today in Fides, you cannot create or modify rule targets to use custom data categories, which will limit the ability of customers to configure DSR rules based on their own customizations to the Fideslang taxonomy. The rule targets currently only support data categories from the default taxonomy.

**Steps to reproduce**
1. Create a custom data category - this can be done either via a YAML file extension of the taxonomy or via the UI at /taxonomy.
2. Try a PATCH request to: "/api/v1/dsr/policy/{policy_key}/rule/{rule_key}/target"
3. You can use the policy key `default_access_policy` and the rule key `default_access_policy_rule`
4. You should get an error that lists the valid enumeration members as the default data categories only. 

**AC**
- Update the rule target data category validation to allow any data category from the instance of Fides. I should be able to extend the taxonomy and create custom data categories, then create a rule target for a policy that uses my newly created data category.
- The validation should not allow any data category, just those defined in that Fides' instances taxonomy - findable at /api/v1/data_category

seanpreston:
Closes #3103 

### Description

Updates `RuleTarget.data_category` API validation to also check custom categories defined in the app DB. There are some limitations:
- The `DataCategory` class from the taxonomy cannot be hashed in its current form, so I'm unable to use a `Set`
- As a result the `get_fides_data_category_superset()` function returns a `List` of `DataCategory` schemas
- and to be sure the return type from `get_fides_data_category_superset()` is uniform `get_fides_data_categories_from_db()` casts the DB models into `DataCategory` schemas.

### Code Changes

* [x] adds new util functions to `util/data_category.py` to get all categories from both the DB and taxonomy, and create a superset of both
* [x] updates data category validation to use this new superset

### Steps to Confirm

* [ ] Create a custom data category in the UI at `/taxonomy`
* [ ] Send a `PATCH` request to `/api/v1/dsr/policy/{policy_key}/rule/{rule_key}/target` with the new data category as the value:
```
[
  {
    "data_category": "<custom_data_category>"
  }
]
```
* [ ] Assert the update succeeds
* [ ] Send a `PATCH` request to `/api/v1/dsr/policy/{policy_key}/rule/{rule_key}/target` with an invalid data category as the value:
```
[
  {
    "data_category": "something_else"
  }
]
```
* [ ] Assert the update fails

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### DIFF
--- a/src/fides/api/ops/schemas/policy.py
+++ b/src/fides/api/ops/schemas/policy.py
@@ -1,12 +1,14 @@
 from typing import Any, Dict, List, Optional
 
 from fideslang.validation import FidesKey
+from pydantic import validator
 
+from fides.api.ops import common_exceptions
 from fides.api.ops.models.policy import ActionType, DrpAction
 from fides.api.ops.schemas.api import BulkResponse, BulkUpdateFailed
 from fides.api.ops.schemas.base_class import BaseSchema
 from fides.api.ops.schemas.storage.storage import StorageDestinationResponse
-from fides.api.ops.util.data_category import DataCategory
+from fides.api.ops.util.data_category import _validate_data_category
 
 
 class PolicyMaskingSpec(BaseSchema):
@@ -31,12 +33,20 @@ class RuleTarget(BaseSchema):
 
     name: Optional[str]
     key: Optional[FidesKey]
-    data_category: DataCategory  # type: ignore
+    data_category: str
 
     class Config:
         """Populate models with the raw value of enum fields, rather than the enum itself"""
 
         use_enum_values = True
+
+    @validator("data_category")
+    def validate_data_category(cls, data_category: str) -> str:
+        """Ensures the value passed in is either in the default Fides taxonomy or in the database"""
+        try:
+            return _validate_data_category(data_category)
+        except common_exceptions.DataCategoryNotSupported:
+            raise ValueError(f"The data category {data_category} is not supported.")
 
 
 class RuleBase(BaseSchema):
--- a/src/fides/api/ops/util/data_category.py
+++ b/src/fides/api/ops/util/data_category.py
@@ -17,7 +17,9 @@ def get_fides_data_categories_from_db() -> List[FideslangDataCategory]:
     default taxonomy.
     """
     db: Session = get_db_session(get_config())()
-    return [cat.to_fideslang_obj() for cat in DataCategoryDbModel.all(db)]
+    all_categories = DataCategoryDbModel.all(db)
+    db.close()
+    return [cat.to_fideslang_obj() for cat in all_categories]
 
 
 def get_fides_data_categories_from_taxonomy() -> List[FideslangDataCategory]:
--- a/src/fides/api/ops/util/data_category.py
+++ b/src/fides/api/ops/util/data_category.py
@@ -1,9 +1,36 @@
 from enum import Enum as EnumType
-from typing import Type
+from typing import List, Set, Type, Union
 
 from fideslang import DEFAULT_TAXONOMY
+from fideslang.models import DataCategory as FideslangDataCategory
+from sqlalchemy.orm import Session
 
+from fides.api.ctl.sql_models import DataCategory as DataCategoryDbModel
 from fides.api.ops import common_exceptions
+from fides.core.config import get_config
+from fides.lib.db.session import get_db_session
+
+
+def get_fides_data_categories_from_db() -> List[FideslangDataCategory]:
+    """
+    Returns each DataCategory from the database as it would have been defined in the
+    default taxonomy.
+    """
+    db: Session = get_db_session(get_config())()
+    return [cat.to_fideslang_obj() for cat in DataCategoryDbModel.all(db)]
+
+
+def get_fides_data_categories_from_taxonomy() -> List[FideslangDataCategory]:
+    """Returns the list of DataCategories from the default Fides taxonomy."""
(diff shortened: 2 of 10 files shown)


src/fides/api/ops/util/data_category.py
```diff
@@ -1,9 +1,36 @@
...
+
+
+def get_fides_data_categories_from_taxonomy() -> List[FideslangDataCategory]:
+    """Returns the list of DataCategories from the default Fides taxonomy."""
+    return DEFAULT_TAXONOMY.data_category
+
+
+def get_fides_data_category_superset() -> List[FideslangDataCategory]:
+    """Returns the union of DataCategories from the database and the default Fides taxonomy."""
+    # TODO: De-duplicate this list
```
seanpreston, line 30: Unfortunately `FideslangDataCategory` doesn't hash straight out the box, which rules out using a `Set` without some tweaks, so I've used a `List` for now.

pattisdr:
Starting review @seanpreston ..

src/fides/api/ops/util/data_category.py
```diff
@@ -1,9 +1,38 @@
...
+
+
+def get_fides_data_categories_from_db() -> List[FideslangDataCategory]:
+    """
+    Returns each DataCategory from the database as it would have been defined in the
+    default taxonomy.
+    """
+    db: Session = get_db_session(get_config())()
+    all_categories = DataCategoryDbModel.all(db)
+    db.close()
```
pattisdr, line 21: I don't think we should be creating a new session here, this also creates a new engine just to get the data categories.  The rule target endpoint has the session injected - I think we should pass that session to a method that validates the data category instead.

src/fides/api/ops/schemas/policy.py
```diff
@@ -31,13 +33,21 @@ class RuleTarget(BaseSchema):
...
         use_enum_values = True
 
+    @validator("data_category")
+    def validate_data_category(cls, data_category: str) -> str:
+        """Ensures the value passed in is either in the default Fides taxonomy or in the database"""
+        try:
+            return _validate_data_category(data_category)
+        except common_exceptions.DataCategoryNotSupported:
+            raise ValueError(f"The data category {data_category} is not supported.")
+
```
pattisdr, line 50: Typically validators should be more lightweight and not rely on a database session - which you're having to create buried within _validate_data_category.  

Instead, I'd recommend sticking to other patterns in our codebase by adding follow-up validation requiring the database to the API endpoint itself, in this case the Rule Target endpoint.

For example: see the validation on our Dataset endpoints that make sure the data categories match ones in the database. 
 https://github.com/ethyca/fides/blob/main/src/fides/api/ops/api/v1/endpoints/dataset_endpoints.py#L128

src/fides/api/ops/util/data_category.py
```diff
@@ -20,8 +49,14 @@ def generate_fides_data_categories() -> Type[EnumType]:
...
+    from_taxonomy = [cat.fides_key for cat in get_fides_data_categories_from_taxonomy()]
+    if all(
+        [
+            data_category not in from_db,
+            data_category not in from_taxonomy,
+        ]
+    ):
         raise common_exceptions.DataCategoryNotSupported(
             f"The data category {data_category} is not supported."
         )
```
pattisdr, line 62: We added similar logic in the Unified Fides Resources release - can we consolidate please?   We have too many "validate data category" methods.

https://github.com/ethyca/fides/blob/main/src/fides/api/ops/api/v1/endpoints/dataset_endpoints.py#L87-L99

src/fides/api/ops/util/data_category.py
```diff
@@ -20,8 +49,14 @@ def generate_fides_data_categories() -> Type[EnumType]:
 
 def _validate_data_category(data_category: str) -> str:
     """Checks that the data category passed in is currently supported."""
-    valid_categories = DataCategory.__members__.keys()
-    if data_category not in valid_categories:
+    from_db = [cat.fides_key for cat in get_fides_data_categories_from_db()]
+    from_taxonomy = [cat.fides_key for cat in get_fides_data_categories_from_taxonomy()]
```
pattisdr, line 53: Is the main reason you're having to get data categories from the taxonomy our fides ops testing issue?  If what really matters is what is in the customer's database, should we instead try to work towards that?

pattisdr (changes requested):

src/fides/api/ops/util/data_category.py
seanpreston: > Is the main reason you're having to get data categories from the taxonomy our fides ops testing issue?

It's because (AFAIK) we can't be certain Fides is running with the data categories from the taxonomy populated in the application DB.

pattisdr: To rephrase, I just meant, can't we check what's in the database? I thought the custom taxonomy is loaded on startup.  (Our ops tests delete but that's separate).

seanpreston: Hey @pattisdr, I've had another pass at this [here](https://github.com/ethyca/fides/pull/3147/). The main difference is that it attempts to address the issues in the tests where we can't guarantee the taxonomy data categories are in the DB.

seanpreston:
Closing this in favour of https://github.com/ethyca/fides/pull/3147/

PR #3120, opened 2023-04-21
[Backend] Privacy Notice Templates 

### LINKED ISSUE #2833: Backend Out of the Box Privacy Notices
❗ Blocked by product supplying the default privacy notices.

Allows Ethyca to define default privacy notices that can be returned to the Compliance person as a starting point from which to tailor their Privacy Notices. 

See Design Doc: [Out of the Box Privacy Notices](https://ethyca.atlassian.net/wiki/spaces/EN/pages/2636382212/2023-03-10+Backend+Design+Enhancing+Consent+Management+Capabilities+and+Consent+Reporting#Out-of-the-Box-Privacy-Notices)

### Additional context

pattisdr:
Closes #2833 

### Code Changes

* [ ] Add a new PrivacyNoticeTemplate table to store our out of the box templates. These are not the notices surfaced to the users.
* [ ] Add an optional FK from PrivacyNoticeHistory and PrivacyNotice back to PrivacyNoticeTemplate, if the notice was, in fact, created from a template
* [ ] Add a yaml file data/privacy_notices/privacy_notice_templates.yml for defining the out of the box notices. Give each template an id ahead of time so we can upsert.
* [ ] Add some new regions
* [ ] Extract the code in the existing endpoint to create privacy notices/update privacy notices to use for upserting templates creating privacy notices on startup.

### Steps to Confirm

* [ ] Bring up the server with `nox -s dev`
* [ ] Issue an API call to get privacy notices GET {{host}}/privacy-notice
* [ ] Make invalid changes to `privacy_notice_templates.yml` (such as a bad data use, or duplicate data uses, or an invalid consent mechanism).  Stop the server and then restart. Verify you get validation errors in the logs but server still comes up.
* [ ] Revert the bad updates. Update a yaml. Restart the server.  Verify the template is updated but it's corresponding privacy notice is not.

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### Description Of Changes

Start fides with an initial set of privacy notices.  Create a yaml file to define them and a separate table to store the templates.  On startup, upsert the templates into the PrivacyNoticeTemplate table (the templates have ids already in the yaml).  Then if the templates are new, create corresponding PrivacyNotice and PrivacyNoticeHistory records. Otherwise don't touch those tables.

Note that templates are not what's shown to users, it's the enabled notices in the PrivacyNotice table.  But this action lets us give customers a starting set of notices out of the box.

Adds what templates I can from: https://ethyca.atlassian.net/wiki/spaces/PM/pages/2656108545/Privacy+Notice+Templates


data/privacy_notices/privacy_notice_templates.yml
```diff
@@ -0,0 +1,227 @@
...
+    - eu_mt
+    - eu_nl
+    - eu_at
+    - eu_pl
+    - eu_pt
+    - eu_ro
+    - eu_si
+    - eu_sk
+    - eu_fi
+    - eu_se
```
pattisdr, line 32: Currently listing all the EU regions out individually, let's double check this is how we want to define this

data/privacy_notices/privacy_notice_templates.yml
```diff
@@ -0,0 +1,227 @@
...
+  consent_mechanism: notice_only
+  data_uses:
+    - provide
```
pattisdr, line 35: Our current validation doesn't allow data uses in the same hierarchy to be used on the same region, including if they are on the same notice.

This is defined in JIRA as being both `provide` and `provide.service`.  I've added just `provide` here. Alternatively, we could change the validation, but I think `provide.service` is irrelevant if `provide` is also added here.

data/privacy_notices/privacy_notice_templates.yml
```diff
@@ -0,0 +1,227 @@
...
+    - eu_se
+  consent_mechanism: opt_in
+  data_uses:
+    - improve.system
+  enforcement_level: frontend
+  disabled: False
+  has_gpc_flag: False
+  displayed_in_privacy_center: False
+  displayed_in_overlay: True
+  displayed_in_api: False
```
pattisdr, line 81: What kinds of notices get displayed in the API? Should this actually be true for everything?

src/fides/api/ctl/migrations/versions/d8a50048dfed_add_privacy_notice_templates.py
```diff
@@ -0,0 +1,94 @@
...
+        ["origin"],
+        ["id"],
+    )
+    op.create_foreign_key(
+        "privacynoticehistory_origin_fkey",
+        "privacynoticehistory",
+        "privacynoticetemplate",
+        ["origin"],
+        ["id"],
+    )
```
pattisdr, line 77: I don't know if this really needs to be a FK to the template. The PrivacyNoticeTemplate table might be a temporary solution, but I think it's easier to add it up front and remove it later, then to handle the reverse.

src/fides/api/ctl/migrations/versions/d8a50048dfed_add_privacy_notice_templates.py
```diff
@@ -0,0 +1,94 @@
...
+
+    conn = op.get_bind()
+    conn.execute("COMMIT")
```
pattisdr, line 80: I'm committing at the end of this schema migration so I can write to the PrivacyNoticeTemplate table in the following migration.  I generally try not to do this, so let's make sure this is okay.

src/fides/api/ops/api/v1/endpoints/privacy_notice_endpoints.py
```diff
@@ -203,27 +204,11 @@ def create_privacy_notices(
...
-    except ValidationError as e:
-        raise HTTPException(HTTP_422_UNPROCESSABLE_ENTITY, detail=e.message)
-
-    return [
-        PrivacyNotice.create(
-            db=db, data=privacy_notice.dict(exclude_unset=True), check_name=False
-        )
-        for privacy_notice in privacy_notices
-    ]
+        return create_privacy_notices_util(db, privacy_notices, PrivacyNotice)  # type: ignore[return-value]
```
pattisdr, line 209: I move all this into `create_privacy_notices_util` so I can share all this nice validation when populating the templates and then the notices from those templates.

src/fides/api/ctl/migrations/versions/8173a96aad1a_populate_default_notice_templates.py
```diff
@@ -0,0 +1,33 @@
...
+    "/fides/data/privacy_notices/privacy_notice_templates.yml"
+)
+
+
+def upgrade():
+    """Add default privacy notices to the database, and then create
+    PrivacyNotices and PrivacyNoticeHistories from there"""
+    sessionlocal = get_db_session(CONFIG)
+    with sessionlocal() as session:
+        load_default_notices(session, DEFAULT_PRIVACY_NOTICES_PATH)
```
pattisdr, line 29: I generally think it is not a great practice to use SQLAlchemy models within a migration as I"m doing here, as the models themselves can change, making the data migration at this point in time invalid.  But I think this would be caught in development if this happened in the future, and I think the benefits are worth all the extra validation we get here.  

Alternatively, this could be something that runs on startup, perhaps if there are no Privacy Notice Templates in the database at all?

data/privacy_notices/privacy_notice_templates.yml
```diff
@@ -0,0 +1,227 @@
...
+    - eu_pl
+    - eu_pt
+    - eu_ro
+    - eu_si
+    - eu_sk
+    - eu_fi
+    - eu_se
+  consent_mechanism: opt_in
+  data_uses:
+    - advertising.third_party.personalized
```
pattisdr, line 115: Data uses are a required item for creating a privacy notice so I've had to omit all privacy notices without them.

src/fides/api/ctl/migrations/versions/8173a96aad1a_populate_default_notice_templates.py
pattisdr, line 29: This is another example of something that is problematic in ops unit testing due to https://github.com/ethyca/fides/issues/2016

pattisdr: Ah, silly, I can't create this in a data migration, it has to go in on startup anyway, which is fine, this works better.  Wrapped in a try/except, if it fails, we just continue to bring everything up.

src/fides/api/ctl/migrations/versions/d8a50048dfed_add_privacy_notice_templates.py
pattisdr: This is no longer needed, phew

pattisdr:
@adamsachs I've mostly repurposed your logic from creating and updating privacy notices to be able to upsert privacy notice templates and then create privacy notices from those templates on startup.  Let me know what you think!

data/privacy_notices/privacy_notice_templates.yml
pattisdr: @mfbrown

data/privacy_notices/privacy_notice_templates.yml
pattisdr: @mfbrown

data/privacy_notices/privacy_notice_templates.yml
pattisdr: @mfbrown

data/privacy_notices/privacy_notice_templates.yml
```diff
@@ -0,0 +1,227 @@
...
+    - eu_ro
+    - eu_si
+    - eu_sk
+    - eu_fi
+    - eu_se
+  consent_mechanism: opt_in
+  data_uses:
+    - improve.system
+  enforcement_level: frontend
+  disabled: False
```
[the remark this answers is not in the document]
pattisdr: @mfbrown

data/privacy_notices/privacy_notice_templates.yml
mfbrown: I'd say make this false for all the templates now. The API flag is meant to be for custom notices that customers create and expose/collect consent for in their own products and services. The expectation would be that all notices can be retrieved via an API and the API flag here is to support some future validation logic that allows a customer to create a notice for presentation in a different medium instead of the privacy center or overlay.

data/privacy_notices/privacy_notice_templates.yml
pattisdr: @mfbrown not currently no, an `advertising.first_party.personalized` for CA and an `advertising.third_party_personalized` for CA would be disallowed.It's pretty strict at the moment.  I can look into allowing this case.

data/privacy_notices/privacy_notice_templates.yml
pattisdr: Thanks for clarifying! Great, they're all False for now.

data/privacy_notices/privacy_notice_templates.yml
pattisdr: @mfbrown are you sure we shouldn't wait to release this until Fideslang 1.4?

data/privacy_notices/privacy_notice_templates.yml
pattisdr: Thanks @mfbrown I've reticketed this here: advertising.third_party.personalized

pattisdr: Ah I was wrong, it does work currently, so I'm just adding more tests for this.

pattisdr:
I'm going to close this PR for now as it's going to get out of date quickly, but I'll borrow heavily from this when we're ready to add the privacy notices post- fideslang 1.4.

PR #3131, opened 2023-04-21
[#807] adds fides_user_device_id to ProvidedIdentityType Enum

seanpreston:
Closes https://github.com/ethyca/fidesplus/issues/807

### Code Changes

* [ ] Adds `fides_user_device_id` to `ProvidedIdentityType` enum and migration

### Steps to Confirm

* [ ] `nox -s dev -- shell`
* [ ] `pip install ipython`
* [ ] `ipython` then enter the following:
```
from fides.api.ops.models.privacy_request import ProvidedIdentityType, ProvidedIdentity
from fides.core.config import get_config
from fides.lib.db.session import get_db_session
db = get_db_session(get_config())()
pi = ProvidedIdentity.create(db=db, data={"field_name": "fides_user_device_id"})
assert pi.field_name.value == "fides_user_device_id"
```

### Pre-Merge Checklist

* [ ] All CI Pipelines Succeeded
* Documentation:
  * [ ] documentation complete, [PR opened in fidesdocs](https://github.com/ethyca/fidesdocs/pulls)
  * [ ] documentation [issue created in fidesdocs](https://github.com/ethyca/fidesdocs/issues/new/choose)
* [ ] Issue Requirements are Met
* [ ] Relevant Follow-Up Issues Created
* [ ] Update `CHANGELOG.md`
* [ ] For API changes, the [Postman collection](https://github.com/ethyca/fides/blob/main/docs/fides/docs/development/postman/Fides.postman_collection.json) has been updated

PR #3132, opened 2023-04-23
[Backend] Save and Retrieve Privacy Preferences by Fides User Device Id 

### LINKED ISSUE #2838: Backend Saving/Retrieving Preferences on User Device Id
- New columns on PrivacyPreference and PrivacyPreference history, FK to provided identity created from "fides user device id", and then the id itself for consent reporting.
- Endpoint to save privacy preferences for a fides user device id
- Endpoint to retrieve privacy preferences on a fides user device id

pattisdr:
Closes #2838 
❗ Dependent on https://github.com/ethyca/fides/pull/3131

### Description Of Changes

- Adds two new endpoints to Save and Retrieve Privacy Preferences with respect to a `fides_user_device_id` that was in the browser instead of just a verified identity like an email or a phone number
- Start saving current and historical privacy preferences with respect to an optional fides user device id
- Allow current and historical preferences to be saved with respect to one or both of a verified provided identity and a fides user device provided identity id simultaneously
- If preferences are saved with respect to a verified provided identity and a fides user device id simultaneously, consolidate the current preferences into single records
- Expose the new fides user device id in the historical report
- Enforce on the backend that a privacy preference must either be saved with respect to a verified provided identity or a fides user device id provided identity (or both!) but we require at least one.

### Code Changes

* [x] Add new indexed FK column on the current record: `CurrentPrivacyPreference.fides_user_device_provided_identity_id`.  A unique constraint added here so we maintain one "current" record for each "fides_user_provided_identity_id" and each "privacy_notice"
* [x] Add new indexed FK column on the historical record: `PrivacyPreferenceHistory.fides_user_device_provided_identity_id`
* [x] Add hashed field `PrivacyPreferenceHistory.hashed_fides_user_device` field so we can search historical records by this device id more easily
* [x] Cache the encrypted value `PrivacyPreferenceHistory.fides_user_device` directly on this field so we can surface for reporting purposes without having to join in the ProvidedIdentity table.
* [x] Add new method `get_or_create_fides_user_device_id_provided_identity` to get and create provided identity records for device ides and keeping ProvidedIdentities for a fides user device id unique when saved for the purposes of consent. We separately create these records for privacy requests, and those are not expected to be unique.
* [x] Update existing endpoint for saving privacy preferences in the verified identity flow to allow a fides user device id to optionally be saved alongside these preferences
* [x] Pull out a new method `_save_privacy_preferences_for_identities` to be shared between the existing endpoint that saves privacy preferences with respect to a verified identity and the new endpoint that only saves preferences with respect to a fides user device id.
* [x] Add the two new endpoints: `PATCH /privacy-preferences` and `GET /privacy-preferences?fides_user_device_id=XXXXXX`

### Steps to Confirm

* [ ] In postman, create a client with permission to save privacy notices.
* [ ] Use the Privacy Notices > Create Privacy Notices endpoint to create a starting set of notices.  Note one of the "privacy_notice_history_id"s that are created
* [ ] Under the Privacy Preferences > Save Privacy Preferences for Device Id collection, save a preference with respect to a `privacy_notice_history_id` you created above. Verify response matches what you'd expect
* [ ]  Get saved Privacy Preferences with respect to device id under the Privacy Preferences > Privacy Preferences by Fides User Device Id collection
* [ ] Access the historical report via the PrivacyPReferences > Get Historical Preferences collection and verify that fides user device id shows up under "fides_user_device_id"

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`


src/fides/api/ops/api/v1/endpoints/privacy_preference_endpoints.py
```diff
@@ -241,21 +295,117 @@ def save_privacy_preferences(
...
+@router.patch(
+    PRIVACY_PREFERENCES,
+    status_code=HTTP_200_OK,
+    response_model=List[CurrentPrivacyPreferenceSchema],
+)
+def save_privacy_preferences(
+    *,
+    request: Request,
+    db: Session = Depends(get_db),
+    data: PrivacyPreferencesCreateWithCode,
```
pattisdr, line 374: For now, using a very similar request body to the existing endpoints where we save preferences with respect to a verified identity, but plan to refactor this request body to pull more items from request headers in https://github.com/ethyca/fides/issues/3133

pattisdr:
@seanpreston requesting your review since I'm saving privacy preferences here with respect to the new style PrivacyPreferenceHistory and CurrentPrivacyPreference records and you are working on saving this with respect to the old-style Consent records.   This is built off of your ProvidedIdentity migration, and I also add a method or two that might be useful for your work. 

@eastandwestwind I was hoping to get your review on this as well as it's the start of the backend for being able to save and retrieve privacy preferences from things like overlays.

src/fides/api/ops/api/v1/endpoints/privacy_preference_endpoints.py
```diff
@@ -304,6 +454,7 @@ def get_historical_consent_reporting(
             PrivacyPreferenceHistory.id,
             PrivacyRequest.id.label("privacy_request_id"),
             PrivacyPreferenceHistory.email.label("user_id"),
```
eastandwestwind, line 456: I know this code already existed, but can you tell me more about why email is labeled as the `user_id`? For example, if a privacy center is configured to only accept phone number, would this field be `null`, and would the phone show up under `secondary_user_ids`?

eastandwestwind (approved):
Great work on this @pattisdr ! I've gone through your testing steps and confirmed that everything worked as expected for me. 

I can send everything expected in `PATCH` `{{host}}/privacy-preferences` with no issues, from FE for now. We can definitely refactor some into BE later.

src/fides/api/ops/api/v1/endpoints/privacy_preference_endpoints.py
```diff
@@ -241,21 +295,117 @@ def save_privacy_preferences(
...
+    response_model=Page[CurrentPrivacyPreferenceSchema],
+)
+def get_privacy_preferences_by_device_id(
+    *,
+    request: Request,
+    fides_user_device_id: str,
+    db: Session = Depends(get_db),
+    params: Params = Depends(),
+) -> AbstractPage[CurrentPrivacyPreference]:
+    """Saves privacy preferences with respect to a fides user device id.
```
eastandwestwind, line 333: I think this docstr needs to be updated to match this GET endpoint

src/fides/api/ops/models/privacy_preference.py
```diff
@@ -69,18 +93,24 @@ class PrivacyPreferenceHistory(Base):
...
+    # Optional FK to a verified provided identity (like email or phone), if applicable
     provided_identity_id = Column(String, ForeignKey(ProvidedIdentity.id), index=True)
-    relevant_systems = Column(
-        MutableList.as_mutable(ARRAY(String))
-    )  # Systems whose data use match.  This doesn't necessarily mean we propagate. Some may be intentionally skipped later.
+
+    # Systems whose data use match.  This doesn't necessarily mean we propagate. Some may be intentionally skipped later.
+    relevant_systems = Column(MutableList.as_mutable(ARRAY(String)))
+    # Location where we received the request
     request_origin = Column(EnumColumn(RequestOrigin))
```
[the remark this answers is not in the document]
pattisdr: > one question I have about this RequestOrigin enum- overlay corresponds to banner or modal components?

The way I understand it, both.  The Overlay is where the user makes their consent choices.  The Privacy Experience determines whether that overlay was triggered by a banner or a link.  Separately, when I get the Privacy Experiences backend, we can start tracking here specifically which privacy experience the user saw too.

> And in which scenarios (from a customer-facing perspective) will api be used?

I don't think it will be used for anything yet, I've been asked to say that none of our privacy notice templates are "api".
https://github.com/ethyca/fides/pull/3120#discussion_r1175586491. From Michael: " API flag here is to support some future validation logic that allows a customer to create a notice for presentation in a different medium instead of the privacy center or overlay."

> Also, particularly with GPC (and other scenarios too?), we'll automatically send consent preference downstream to fides api. In this case, what does the request origin map to?

RequestOrigin is nullable right now, so in this case, I'd put nothing.  If we need to add more fields to this RequestOrigin enum to support these use cases we can do that.

src/fides/api/ops/api/v1/endpoints/privacy_preference_endpoints.py
```diff
@@ -2,21 +2,29 @@
...
+    HTTP_403_FORBIDDEN,
+    HTTP_404_NOT_FOUND,
+)
 
 from fides.api.ctl.database.seed import DEFAULT_CONSENT_POLICY
 from fides.api.ops.api.deps import get_db
 from fides.api.ops.api.v1.endpoints.consent_request_endpoints import (
     _get_consent_request_and_provided_identity,
+    _get_fides_user_device_id_provided_identity,
+    _get_or_create_fides_user_device_id_provided_identity,
```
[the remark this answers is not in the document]
pattisdr: I can look into extracting this into a util, appreciate you flagging.

src/fides/api/ops/api/v1/endpoints/privacy_preference_endpoints.py
pattisdr: ah thanks copy paste

src/fides/api/ops/api/v1/endpoints/privacy_preference_endpoints.py
pattisdr: Good highlight.  I'd like to update this so email and phone number have their own columns.

 It was like this because `user_id` was a field requested in the consent reporting requirements and it was intended to be the primary identifier, so the email, the phone number, or the fides user device id. But then I started deviating from that because for filtering purposes it's going to be a lot simpler to keep each "type" of data in its own column.  I forgot to go back and update this. 

I intend for secondary_user_ids to be the user ids beyond the "primary id" (mostly ids we get from the browser)

pattisdr:
> One more question as I'm looking through docs- will we need an endpoint to retrieve privacy preference by user id (email / phone)? This can be a follow-up of course depending on need, specifically in privacy center, when an identity is known, but it was a first time visitor.

@eastandwestwind I am not sure I understand your question.

I do have this separate endpoint for retrieving privacy preferences by an email or phone number that's very similar to the request we make in the privacy center now except this returns the new-style privacy preferences.  This endpoint requires you follow that flow to verify your identity.

Screenshot 2023-04-24 at 7 28 53 PM

eastandwestwind:
@pattisdr this was what I was looking for, thanks!

src/fides/api/ops/api/v1/endpoints/privacy_preference_endpoints.py
pattisdr: Updated!

PR #3142, opened 2023-04-24
Updating dataset reference type name

### LINKED ISSUE #3141: Unable to save connectors with external references
### Bug Description

Fides throws an exception when trying to save the connection config for a connector with external data references

#### Steps to Reproduce

1. Navigate to **Privacy requests > Connection manager > Create new connection**
2. Click on the **Yotpo Reviews** connector
3. Fill out the values and click **Save**

### Expected behavior

The connection should save with no errors

### Screenshots
invalid_values

galvana:
Closes #3141

### Code Changes

* [ ] Updating the type name for external dataset references in the front-end

### Steps to Confirm

* [ ] Verified by successfully saving a connector config with external references (Yotpo Reviews)

### Pre-Merge Checklist

* [ ] Update `CHANGELOG.md`

### Description Of Changes
We key off of the type names to determine if we should convert a dataset reference string into a dataset reference object. The type in the front-end no longer matched the type in the backend.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@ The types of changes are:
 ### Fixed
 
 - The `--local` flag is now respected for the `scan dataset db` command [#3096](https://github.com/ethyca/fides/pull/3096)
+- Fixing issue where connectors with external dataset references would fail to save [#3142](https://github.com/ethyca/fides/pull/3142)
 
 ### Developer Experience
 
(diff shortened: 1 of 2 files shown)


allisonking (approved):
good catch! unfortunately the connectors UI might be the least tested part of our UI, and code that relies on constants being the same between the FE and the BE are the hardest to test 😬 

if we did want to cover something like this, we'd have a few options:
* Cypress test which triggers this flow and verifies the payload sent to the backend
  * We do this the most often because it doesn't require a running backend—we just verify that the backend receives what we think it should receive. this works well most of the time, but might not have caught this error since the FE would happily keep thinking it should reference `FidesopsDatasetReference` and the cypress test would probably say so too
* automated cypress smoke test. this tests fully end to end. we only have a few tests here right now, and they're meant for paths that are critical to usability (since they are slow and run on every PR)
* QA cypress test. this would fall in a suite that Roger has made and is run on release, sometimes more often

RobertKeyser (approved):

Issue #3118, opened 2023-04-20
`SafeStr` inputs can cause unwanted behavior in frontend inputs


allisonking:
### Bug Description

* If a user types something with certain characters i.e. `&` or `'` to a field that is a `SafeStr` in the backend, the backend sanitizes the string before saving it to the database
* this results in the user receiving the sanitized version back from the backend, with escaped characters, i.e. `&amp;#x27;` which is unpleasant UX
* but worse than the unpleasant UX, is that if the user saves again, because there are special characters within the escaped content, the problem amplifies and the user input actually changes, i.e. from `&amp;#x27;` to `&amp;amp;amp;#x27;t`. which is not just bad UX, but just bad!

#### Steps to Reproduce

1. Edit a privacy notice (steps described in [this PR](https://github.com/ethyca/fides/pull/3058))
1. Edit the description to have an apostrophe
2. Save the form
3. You'll be returned the sanitized version
4. Make any other edit to the form
5. Save again
6. You'll get even more ampersands in your string

### Expected behavior

The string the user saved to be the one they get back

### Screenshots

![image](https://user-images.githubusercontent.com/24641006/233402022-dd13c91a-f838-41f4-bb89-ee119ca03a87.png)

Saving again
![image](https://user-images.githubusercontent.com/24641006/233402151-72c1ac0b-c225-4a21-8487-f7984f286dd8.png)

### Environment

* Version:
* OS:
* Python Version:
* Docker Version:

### Additional context

`SafeStr` was introduced to prevent XSS. @daveqnet wrote up a 💯 [doc on Fides and validation/sanitization/encoding](https://ethyca.atlassian.net/wiki/spaces/~80431056/pages/2670886919/Fides+AppSec+Validation+Sanitization+and+Encoding)

ThomasLaPiana:
I think it makes sense to be able to "unsafe" strings for certain fields/certain models, maybe via a Header (like @allisonking mentioned in Slack) or a param

This would make it easy to explicitly define where we want unsafe-d strings vs safe ones, and would still default to safe

PR #3144, opened 2023-04-24
Add option to unescape SafeStr

allisonking:
Closes https://github.com/ethyca/fides/issues/3118

### Code Changes

* [x] add a function that takes an object and a list of field names and escapes them
* [x] escape the `name`, `description`, `internal_description` and `origin` fields for the PrivacyNotice
* [x] add escape/unescape to the `privacy_notice` endpoints

### Steps to Confirm

* [x] new test passes

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### Description Of Changes

Instead of using the more heavy-handed `SafeStr`, we've decided to replace it with a little utility function that can escape/unescape things in a more explicit way. See code for implementation details.

For now, this is only used with the `PrivacyNotice` endpoints


src/fides/api/ops/api/v1/endpoints/privacy_notice_endpoints.py
```diff
@@ -161,10 +161,12 @@ def get_privacy_notice(
     *,
     privacy_notice_id: str,
     db: Session = Depends(deps.get_db),
+    request: Request,
 ) -> PrivacyNotice:
     """
     Return a single PrivacyNotice
     """
+    unescape = request.headers.get("unescape-safestr")
```
allisonking, line 169: I'd like to do this at a higher level so that not every endpoint that wants to do this has to do it themselves. but I'm not familiar enough with fastapi to know where that kind of logic should go

src/fides/api/ops/api/v1/endpoints/privacy_notice_endpoints.py
```diff
@@ -161,10 +161,12 @@ def get_privacy_notice(
...
     return get_privacy_notice_or_error(db, privacy_notice_id)  # type: ignore[return-value]
```
allisonking, line 170: At some point we need to call the `unescape` method, but I'm a bit muddied on where that should be. These appear to all operate on the pydantic model, while the `escape`/`unescape` logic is on the sqlalchemy model

tests/ops/api/v1/endpoints/test_privacy_notice_endpoints.py
```diff
@@ -646,6 +646,61 @@ def test_get_privacy_notice(
...
+                        PrivacyNoticeRegion.eu_be.value,
+                        PrivacyNoticeRegion.us_ca.value,
+                    ],
+                    "consent_mechanism": ConsentMechanism.opt_in.value,
+                    "data_uses": ["advertising"],
+                    "enforcement_level": EnforcementLevel.system_wide.value,
+                }
+            ],
+        )
+        assert resp.status_code == 200
```
allisonking, line 678: this test fails here, which is the wrong place where it should be failing! I haven't figured out why it's failing yet—it's complaining about the scope, but from what I can tell it has the right scope...

allisonking: because the test is failing here, I don't actually know if the rest of the test actually does what I want it to, but it should be a reasonable attempt

allisonking:
okay @ThomasLaPiana here's my feeble initial attempt! I'm having trouble wrapping my mind around a few things:
1. where to do this unescaping. it'd be great if the endpoints could all know how to interpret a special header, and then operate on their payloads accordingly
2. how to call the unescaping. I'm getting lost between the pydantic and sqlalchemy models 

left a few comments throughout which go into more detail about what I'm thinking about. baton passing this to you!

ThomasLaPiana:
@allisonking I've caught the baton! Thank you so much for the first pass here 😄 I'll take a look and we'll work through it! Everything is possible through the power of friendship! Gundam taught me that

ThomasLaPiana:
@allisonking I'm having a deep think about this...here are some of my thoughts while I'm twirling the proverbial baton that was metaphorically passed to me

1. Now that I think about it, I'm not so sure I like the original implementation. By doing it at the `pydantic` model level, we make it kind of opaque and hard to "undo". It just "happens"
2. Given the above, maybe we change it to be an explicit step that happens on certain endpoints and for certain fields?

For instance...
1. Privacy notice gets created/sent
2. In the endpoint logic, we make the string safe
3. String gets stored in the database
4. Another endpoint gets called, it explicitly unsafes the string before sending it

I see what you mean about leveraging some kind of Middleware in FastAPI to do this automatically based on the header. So quick question here:
1. Are there endpoints that would _sometimes_ want to send safe strings, and _sometimes_ want to send unsafe strings?

I like the idea of explicitly making the string safe/unsafe, but if we need to make it configurable via header I can see that getting a bit trickier.

Maybe we can do it explicitly for now in each endpoint and work up a middleware solution later?

ThomasLaPiana:
> > I see what you mean about leveraging some kind of Middleware in FastAPI to do this automatically based on the header. So quick question here:
> > 
> > 1. Are there endpoints that would _sometimes_ want to send safe strings, and _sometimes_ want to send unsafe strings?
> > 
> > I like the idea of explicitly making the string safe/unsafe in the endpoint logic, but if we need to make it configurable via header I can see that getting a bit trickier.
> > Maybe we can do it explicitly for now in each endpoint and work up a middleware solution later?
> 
> I don't think there are any endpoints that the frontend would want the safe string for right now. React handles strings safely as long as we aren't doing anything like trying to render user defined HTML.
> 
> That said, I'm perfectly fine with not using a header for now, and using something more flexible. Then if we are happy with the implementation and find ourselves needing to use it more widely, we can iterate and upgrade to some middleware implementation 👍

Interesting, so basically none of the endpoints used by the front-end should be getting safe strings? I

In that case, it seems like it might be easier to unsafe things on the front-end instead of the back-end? Would it be possible to do that programmatically? Where we store everything as safe on the back-end (so that API users are also protected), but when our front-end ingests the API it can unescape them automagically?

ThomasLaPiana:
@allisonking Understood, I think the best way forward then is as follows:

1. We create a util function called `unescape_str` or something similar that does the `unescape` logic we want on the backend
2. We add a conditional to every endpoint that needs it to check for the header set by the frontend. If the header is there, we unsafe all strings in the response (I'm assuming we would want to unsafe all strings to make sure? Can't think of a situation where we'd want one str safe but another not safe?)
3. Profit

Does that make sense to you? Did I leave anything out?

I think this method gives us the benefits of:
1. Easily testable for the `unsafe_str` function
2. From a security standpoint, easy to see explicitly where we're unsafing things/where they could be unsafed

allisonking:
Yes, that makes sense to me! And agree that if the header is passed we should unsafe all strings 👍 If it's easier to iterate, we can just start with the privacy notice endpoint and extend as needed to other endpoints?

ThomasLaPiana:
> Yes, that makes sense to me! And agree that if the header is passed we should unsafe all strings 👍 If it's easier to iterate, we can just start with the privacy notice endpoint and extend as needed to other endpoints?

for sure, a localized change as a first iteration sounds perfect 😄 

I'm deep into the dictionary service spike, do you have the bandwidth to tackle this this sprint? I might be able to get to it but it'd be later since that is my top priority. If both of us are booked we can ask for help

allisonking:
I should have time @ThomasLaPiana , though I may need some guidance on how to go about this part: https://github.com/ethyca/fides/pull/3144#discussion_r1175768900

I suppose an easier thing I could do is get the privacy notice back, then unescape the individual fields I know have been escaped? but ideally, I'd like to be able to do a 

```python
privacy_notice = PrivacyNotice(name="some safe str", ...)

# some func on PrivacyNotice which knows which fields to unescape and handles it
unescaped_privacy_notice = privacy_notice.unescape()
return unescaped_privacy_notice
```

I tried to add an `unescape` func to `SafeStr` itself, but maybe that's not the way to go since that's on the sqlalchemy obj, not the pydantic obj...

ThomasLaPiana:
> I should have time @ThomasLaPiana , though I may need some guidance on how to go about this part: [#3144 (comment)](https://github.com/ethyca/fides/pull/3144#discussion_r1175768900)
> 
> I suppose an easier thing I could do is get the privacy notice back, then unescape the individual fields I know have been escaped? but ideally, I'd like to be able to do a
> 
> ```python
> privacy_notice = PrivacyNotice(name="some safe str", ...)
> 
> # some func on PrivacyNotice which knows which fields to unescape and handles it
> unescaped_privacy_notice = privacy_notice.unescape()
> return unescaped_privacy_notice
> ```
> 
> I tried to add an `unescape` func to `SafeStr` itself, but maybe that's not the way to go since that's on the sqlalchemy obj, not the pydantic obj...

Sorry I should've clarified!

I think we should actually back out the `SafeStr` object entirely, and explicitly call a function to make those fields safe when receiving the object. We can then explicitly call that function to `unsafe` them when we return them. Does that make sense? I know that's a very "functional" approach, but my brain subjectively prefers that to a more object-oriented route

allisonking:
> Sorry I should've clarified!
>
> I think we should actually back out the SafeStr object entirely, and explicitly call a function to make those fields safe when receiving the object. We can then explicitly call that function to unsafe them when we return them. Does that make sense? I know that's a very "functional" approach, but my brain subjectively prefers that to a more object-oriented route. I'm open to either way though! Also it was meant to be on the Pydantic model and not the SQLAlchemy model, maybe I messed that up 🤔

oh you're right, I got 'em mixed up! the endpoint returns the SQLAlchemy model, but I can't just unescape there since the unescape method is on the pydantic model. but thanks for clearing things up, let me take another stab at it and then seek your guidance again later 😄

src/fides/api/ops/api/v1/endpoints/privacy_notice_endpoints.py
```diff
@@ -253,7 +262,9 @@ def prepare_privacy_notice_patches(
                 detail=f"No PrivacyNotice found for id {update_data.id}.",
             )
 
-        updates_and_existing.append((update_data, existing_notices[update_data.id]))
+        updates_and_existing.append(
+            (update_data.escaped(), existing_notices[update_data.id])
```
allisonking, line 266: this isn't properly typed—I put the `escaped()` method on the base method so it returns the base type, not the extended type, not sure how to work around that

allisonking:
Okay @ThomasLaPiana attempt number two! https://github.com/ethyca/fides/pull/3144/commits/caa8f0ebdef50ade14843194d2599d4e142c5ed4 implements a naive poc, which only tries to operate on `name` (I couldn't find a good way to iterate dynamically through the fields I wanted to escape... 😬 ). This implementation still relies on SafeStr, but basically rejiggers the response models a bit such that the GETs don't marshal back into `SafeStr` but we add a function to unescape within the GETs

https://github.com/ethyca/fides/pull/3144/commits/1632560de53c16907cad44b37d8a4d66f5ea33f7 tries to do away with `SafeStr` but then we don't get the <500 char validator either 😕 

both of them would need to touch all the endpoints in not very obvious ways. for instance, I haven't applied this to the GET list endpoint yet, only the GET details endpoint

so overall I don't love this implementation, but figured I'd push some commits and iterate based on your feedback 🙏

ThomasLaPiana:
> Okay @ThomasLaPiana attempt number two! [caa8f0e](https://github.com/ethyca/fides/commit/caa8f0ebdef50ade14843194d2599d4e142c5ed4) implements a naive poc, which only tries to operate on `name` (I couldn't find a good way to iterate dynamically through the fields I wanted to escape... 😬 ). This implementation still relies on SafeStr, but basically rejiggers the response models a bit such that the GETs don't marshal back into `SafeStr` but we add a function to unescape within the GETs
> 
> [1632560](https://github.com/ethyca/fides/commit/1632560de53c16907cad44b37d8a4d66f5ea33f7) tries to do away with `SafeStr` but then we don't get the <500 char validator either 😕
> 
> both of them would need to touch all the endpoints in not very obvious ways. for instance, I haven't applied this to the GET list endpoint yet, only the GET details endpoint
> 
> so overall I don't love this implementation, but figured I'd push some commits and iterate based on your feedback 🙏

Thank you! I'll take a look 🙂

src/fides/api/ops/api/v1/endpoints/privacy_notice_endpoints.py
```diff
@@ -161,11 +162,17 @@ def get_privacy_notice(
...
+    request: Request,
 ) -> PrivacyNotice:
     """
     Return a single PrivacyNotice
     """
-    return get_privacy_notice_or_error(db, privacy_notice_id)  # type: ignore[return-value]
+    should_unescape = request.headers.get("unescape-safestr")
+    notice = get_privacy_notice_or_error(db, privacy_notice_id)
+    if should_unescape:
+        print("orig", notice.name)
```
ThomasLaPiana, line 173: I can't get onboard with this abbreviation 😂 

```suggestion
        print(f"Original Value: {notice.name}")
```

ThomasLaPiana:
yay it looks like tests passed with the last run! This implementation is pretty much what I envisioned, so I'm glad we came to the same place 🙂 taking a closer look now as well but I think this is close to ready

src/fides/api/ops/api/v1/endpoints/privacy_notice_endpoints.py
allisonking: haha that was just for debugging! I'll remove it 👍

tests/ops/api/v1/endpoints/test_privacy_notice_endpoints.py
ThomasLaPiana: im working on a fix here!

ThomasLaPiana:
@allisonking digging into this now and cooking up a fix 🙂 

https://media.giphy.com/media/U25N5SEqHDgqY/giphy.gif

src/fides/api/ops/api/v1/endpoints/utils.py
```diff
@@ -28,3 +29,18 @@ def validate_start_and_end_filters(
                 status_code=HTTP_400_BAD_REQUEST,
                 detail=f"Value specified for {field_name}_lt: {end} must be after {field_name}_gt: {start}.",
             )
+
+
+def transform_fields(
```
ThomasLaPiana, line 34: @allisonking this is my "magic" function, but now it's causing more bugs 🤷 should be what we need here though, figuring out why it's breaking things

allisonking: that's slick!

ThomasLaPiana: Oh in the actual func I put a check but messed it up! I'll fix and try again. I think something a bit weird is going on so I'm digging in

ThomasLaPiana:
@allisonking this should be about good to go on the functionality side! Are you available to get this over the line today? Not sure what else might be needed 🤔

allisonking:
> @allisonking this should be about good to go on the functionality side! Are you available to get this over the line today? Not sure what else might be needed 🤔

yes I can do that! although I can't approve this PR since I made it

ThomasLaPiana:
@allisonking The Unit tests take so long that codecov doesn't even count it...lol

allisonking:
needed one more applied to the GET for the list of privacy notices. I just added that, though it involved fiddling with pagination a bit, so hope that's okay

FE is looking good though
image

image

ThomasLaPiana:
@allisonking Nice catch!

It's taking all of my willpower not to rename `maybe_dangerous_description` to `sus_description`

tests/ops/api/v1/endpoints/test_privacy_notice_endpoints.py
```diff
@@ -579,6 +579,60 @@ def test_pagination(
...
+                "data_uses": ["advertising"],
+                "enforcement_level": EnforcementLevel.system_wide,
+                "displayed_in_overlay": True,
+            },
+        )
+
+        unescape_header = {"Unescape-Safestr": "yes"}
+        auth_and_unescape_header = {**auth_header, **unescape_header}
+
+        print("here")
```
TheAndrewJackson, line 614: Should this be removed or have more context added?

```suggestion
```

tests/ops/api/v1/endpoints/test_privacy_notice_endpoints.py
```diff
@@ -579,6 +579,60 @@ def test_pagination(
...
+                "regions": [PrivacyNoticeRegion.us_ca],
+                "consent_mechanism": ConsentMechanism.opt_in,
+                "data_uses": ["advertising"],
+                "enforcement_level": EnforcementLevel.system_wide,
+                "displayed_in_overlay": True,
+            },
+        )
+
+        unescape_header = {"Unescape-Safestr": "yes"}
+        auth_and_unescape_header = {**auth_header, **unescape_header}
```
TheAndrewJackson, line 612: These two lines may be duplicated. The same code is on lines 626 and 627
```suggestion
```

tests/ops/api/v1/endpoints/test_privacy_notice_endpoints.py
```diff
@@ -649,6 +703,67 @@ def test_get_privacy_notice(
...
+        self,
+        api_client: TestClient,
+        generate_auth_header,
+    ):
+        auth_header = generate_auth_header(
+            scopes=[scopes.PRIVACY_NOTICE_READ, scopes.PRIVACY_NOTICE_CREATE]
+        )
+
+        # post a new privacy notice that has some sneaky characters
+        maybe_dangerous_description = "user's description <script />"
```
TheAndrewJackson, line 716: 😉 

```suggestion
        sus_description = "user's description <script />"
```

TheAndrewJackson (approved):
This looks great! I like the solution. I just left a couple small nits in the unit tests

ThomasLaPiana:
looks like some codecov hick-ups, should be all good

PR #3147, opened 2023-04-24
[#3103] `RuleTarget` validation take #2

seanpreston:
An alternative approach to https://github.com/ethyca/fides/pull/3117 where we can take for granted the presence of all Fides taxonomy DataCategory objects in the app DB.

### Steps to Confirm

* [ ] Create a custom data category in the UI at `/taxonomy`
* [ ] Send a `PATCH` request to `/api/v1/dsr/policy/{policy_key}/rule/{rule_key}/target` with the new data category as the value:
```
[
  {
    "data_category": "<custom_data_category>"
  }
]
```
* [ ] Assert the update succeeds
* [ ] Send a `PATCH` request to `/api/v1/dsr/policy/{policy_key}/rule/{rule_key}/target` with an invalid data category as the value:
```
[
  {
    "data_category": "something_else"
  }
]
```
* [ ] Assert the update fails

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

### DIFF
--- a/pyrightconfig.json
+++ b/pyrightconfig.json
@@ -1,0 +1,4 @@
+{
+    "venvPath": ".",
+    "venv": ".venv"
+}
--- a/pyrightconfig.json
+++ b/pyrightconfig.json
@@ -1,4 +1,0 @@
-{
-    "venvPath": ".",
-    "venv": ".venv"
-}
--- a/src/fides/api/ctl/database/session.py
+++ b/src/fides/api/ctl/database/session.py
@@ -2,9 +2,10 @@ from typing import AsyncGenerator
 
 from sqlalchemy import create_engine
 from sqlalchemy.ext.asyncio import AsyncSession, create_async_engine
-from sqlalchemy.orm import Session, sessionmaker
+from sqlalchemy.orm import sessionmaker
 
 from fides.core.config import CONFIG
+from fides.lib.db.session import ExtendedSession
 
 engine = create_async_engine(
     CONFIG.database.async_database_uri,
@@ -15,7 +16,7 @@ async_session = sessionmaker(engine, class_=AsyncSession, expire_on_commit=False
 sync_engine = create_engine(CONFIG.database.sync_database_uri, echo=False)
 sync_session = sessionmaker(
     sync_engine,
-    class_=Session,
+    class_=ExtendedSession,
     expire_on_commit=False,
     autocommit=False,
 )
--- a/src/fides/api/ctl/database/session.py
+++ b/src/fides/api/ctl/database/session.py
@@ -2,10 +2,9 @@ from typing import AsyncGenerator
 
 from sqlalchemy import create_engine
 from sqlalchemy.ext.asyncio import AsyncSession, create_async_engine
-from sqlalchemy.orm import sessionmaker
+from sqlalchemy.orm import Session, sessionmaker
 
 from fides.core.config import CONFIG
-from fides.lib.db.session import ExtendedSession
 
 engine = create_async_engine(
     CONFIG.database.async_database_uri,
@@ -16,7 +15,7 @@ async_session = sessionmaker(engine, class_=AsyncSession, expire_on_commit=False
 sync_engine = create_engine(CONFIG.database.sync_database_uri, echo=False)
 sync_session = sessionmaker(
     sync_engine,
-    class_=ExtendedSession,
+    class_=Session,
     expire_on_commit=False,
     autocommit=False,
 )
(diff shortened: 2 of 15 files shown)


tests/fixtures/application_fixtures.py
```diff
@@ -155,6 +160,39 @@ def mock_upload_logic() -> Generator:
         yield _fixture
 
 
+@pytest.fixture(scope="function", autouse=True)
```
seanpreston, line 163: I'm going to watch the CI time here, since this effectively tears down and re-adds the data categories before and after each test is run.

src/fides/api/ctl/database/session.py
```diff
@@ -15,7 +16,7 @@
 sync_engine = create_engine(CONFIG.database.sync_database_uri, echo=False)
 sync_session = sessionmaker(
     sync_engine,
-    class_=Session,
+    class_=ExtendedSession,
```
seanpreston, line 19: This in will run `session.rollback()` automatically on every failed transaction.

src/fides/api/ops/api/v1/endpoints/policy_endpoints.py
```diff
@@ -474,6 +503,11 @@ def create_or_update_rule_targets(
             detail=f"No Rule found for key {rule_key} on Policy {policy_key}.",
         )
 
+    _validate_data_categories(
+        db=db,
+        data_categories=[schema.data_category for schema in input_data],
+    )
```
seanpreston, line 509: This now uses the DB session available to the API endpoint.

src/fides/api/ops/models/policy.py
```diff
@@ -190,31 +198,39 @@ def _is_ancestor_of_contained_categories(
...
+    db: Session,
+    target_categories: List[str],
+) -> None:
     """
     Validates that no erasure rules within the Policy have conflicting data category targets:
         - We cannot mask the same data categories multiple times
         - We cannot perform separate masking upon a data category's sub-categories
     """
-
+    all_categories = DataCategory.all(db=db)
```
seanpreston, line 227: I've made a point of fetching these here because:
- a DB session is available from where we're calling this method
- this is the last point before entering a recursive flow

src/fides/api/ops/api/v1/endpoints/policy_endpoints.py
pattisdr: It looks like this is also being checked in the `_validate_data_category` method at the model level too here? I don't know if we need both since it looks like this endpoint allows a subset to succeed and fail.

src/fides/api/ops/api/v1/endpoints/policy_endpoints.py
```diff
@@ -443,6 +449,29 @@ def get_rule_target(
...
+            db=db,
+            conditions=DataCategory.fides_key.in_(data_categories),
+        ).all()
+    ]
+    invalid_categories = list(set(data_categories) - set(from_db))
+    if invalid_categories:
+        raise HTTPException(
+            status_code=HTTP_422_UNPROCESSABLE_ENTITY,
+            detail=f"Invalid data categories: {invalid_categories}",
+        )
```
pattisdr, line 472: I think model-level validation is sufficient here?

seanpreston: I added this to best-replicate the previous experience where data is validated before we attempt to store it in the DB, rather than by the DB.

pyrightconfig.json
```diff
@@ -0,0 +1,4 @@
+{
+    "venvPath": ".",
+    "venv": ".venv"
+}
```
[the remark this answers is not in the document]
seanpreston: It's not, will remove!

src/fides/api/ops/api/v1/endpoints/policy_endpoints.py
pattisdr: Well it isn't being validated by the db specifically, it's being validated before it gets there by the create/update methods.  And the rule target endpoint has a try except specifically to catch the DataCategoryNotSupported error that is now superfluous.

So we are running identical checks back-to-back here, but you can leave this if you prefer.

pattisdr (approved):
Approved with the removal of that extra committed file. Nice work Sean.

src/fides/api/ops/api/v1/endpoints/policy_endpoints.py
pattisdr: Ah I see thanks for clarifying that you're matching existing behavior

PR #3157, opened 2023-04-25
include 'id' on embedded privacy_declaration objects in system API responses

adamsachs:
Closes https://github.com/ethyca/fideslang/issues/102

Adds an `id` field on the `privacy_declaration`s embedded objects returned in `system` APIs.

### Code Changes

* extend our base pydantic models defined in `fideslang` with response-specific models that include `PrivacyDeclaration.id` fields in the response objects
* update tests to work and also confirm the `id` field is present in the response

### Steps to Confirm

* made some manual API calls against the `GET /api/v1/system` endpoint and confirmed the field was present. sample response from my local:
```
{
  "fides_key": "test_system_1",
  "organization_fides_key": "default_organization",
  "tags": [],
  "name": "test system 1",
  "description": "",
  "registry_id": null,
  "meta": null,
  "fidesctl_meta": null,
  "system_type": "",
  "data_responsibility_title": "Controller",
  "egress": null,
  "ingress": null,
  "privacy_declarations": [
    {
      "id": "pri_eb5ad4c6-164a-47c1-a803-a84078b24dde",
      "name": "test",
      "data_categories": [
        "user"
      ],
      "data_use": "advertising",
      "data_qualifier": "aggregated.anonymized.unlinked_pseudonymized.pseudonymized.identified",
      "data_subjects": [
        "commuter"
      ],
      "dataset_references": [],
      "egress": null,
      "ingress": null
    },
    {
      "id": "pri_ce89b310-3015-4373-bc94-2180b63d270e",
      "name": "test 2",
      "data_categories": [
        "user.browsing_history"
      ],
      "data_use": "advertising",
      "data_qualifier": "aggregated.anonymized.unlinked_pseudonymized.pseudonymized.identified",
      "data_subjects": [
        "next_of_kin"
      ],
      "dataset_references": [],
      "egress": null,
      "ingress": null
    }
  ],
  "system_dependencies": [],
  "joint_controller": null,
  "third_country_transfers": [],
  "administrating_department": "Not defined",
  "data_protection_impact_assessment": {
    "is_required": false,
    "progress": null,
    "link": null
  }
}
```

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

We decided to go with this implementation rather than an update to the base model in `fideslang` as was proposed in https://github.com/ethyca/fideslang/pull/103. The two main reasons motivating that decision:
1. we don't want to allow users to specify `PriavcyDeclaration.id` fields in their request bodies - it's meant to be response/read-only
2. we don't want to muddy `fideslang` with db-specific internals like `id`. no other `fideslang` models have this `id` field exposed.

See issue for more context on why this change is needed

### DIFF
--- a/src/fides/api/ctl/routes/system.py
+++ b/src/fides/api/ctl/routes/system.py
@@ -372,7 +372,6 @@ async def create(
     name="List",
 )
 async def ls(  # pylint: disable=invalid-name
-    resource_type: str = "system",
     db: AsyncSession = Depends(get_async_db),
 ) -> List:
     """Get a list of all of the resources of this type."""
@@ -391,7 +390,6 @@ async def ls(  # pylint: disable=invalid-name
 )
 async def get(
     fides_key: str,
-    resource_type: str = "system",
     db: AsyncSession = Depends(get_async_db),
 ) -> Dict:
     """Get a resource by its fides_key."""
--- a/src/fides/api/ctl/routes/system.py
+++ b/src/fides/api/ctl/routes/system.py
@@ -12,8 +12,15 @@ from sqlalchemy.orm import Session
 from starlette import status
 from starlette.status import HTTP_200_OK, HTTP_400_BAD_REQUEST, HTTP_404_NOT_FOUND
 
-from fides.api.ctl.database.crud import create_resource, get_resource, update_resource
+from fides.api.ctl.database.crud import (
+    create_resource,
+    get_resource,
+    get_resource_with_custom_fields,
+    list_resource,
+    update_resource,
+)
 from fides.api.ctl.database.session import get_async_db
+from fides.api.ctl.schemas.system import SystemResponse
 from fides.api.ctl.sql_models import (  # type: ignore[attr-defined]
     DataUse,
     PrivacyDeclaration,
@@ -27,6 +34,7 @@ from fides.api.ops.api.v1.scope_registry import (
     CONNECTION_READ,
     SYSTEM_CREATE,
     SYSTEM_DELETE,
+    SYSTEM_READ,
     SYSTEM_UPDATE,
 )
 from fides.api.ops.api.v1.urn_registry import SYSTEM_CONNECTIONS, V1_URL_PREFIX
@@ -125,7 +133,7 @@ def patch_connections(
 
 @system_router.put(
     "/",
-    response_model=SystemSchema,
+    response_model=SystemResponse,
     responses={
         status.HTTP_403_FORBIDDEN: {
             "content": {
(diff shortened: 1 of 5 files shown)


NevilleS (approved):
This makes sense to me!

src/fides/api/ctl/routes/system.py
```diff
@@ -350,3 +358,41 @@ async def create(
...
+        Security(
+            verify_oauth_client_prod,
+            scopes=[SYSTEM_READ],
+        )
+    ],
+    response_model=SystemResponse,
+)
+async def get(
+    fides_key: str,
+    resource_type: str = "system",
```
pattisdr, line 394: Since we're not using this resource_type query param (i see that it's in the base crud get endpoint), should we remove or add an underscore instead for resource_type?

src/fides/api/ctl/routes/system.py
```diff
@@ -350,3 +358,41 @@ async def create(
...
+        Security(
+            verify_oauth_client_prod,
+            scopes=[SYSTEM_READ],
+        )
+    ],
+    response_model=List[SystemResponse],
+    name="List",
+)
+async def ls(  # pylint: disable=invalid-name
+    resource_type: str = "system",
```
pattisdr, line 375: Same thing here

src/fides/api/ctl/routes/system.py
adamsachs: yeah, i was just a bit confused as to why we have it there at all in the base crud endpoints, and started getting worried that we rely on it (i.e. maybe some API callers, either the UI or tests, specify the query param?)

adamsachs: on some quick testing though things still seem to work fine in the UI, and i'm not seeing any references in the tests, so let me try removing it!

pattisdr: My guess was that it was there originally because of how the endpoints are defined in that loop.
They chose to inject the default resource type as a query param, so it wasn't actually intended to be used as a query param by the caller.

pattisdr: Oh this is interesting, I am pretty sure the query param is not actually supposed to be used, see this bug:

GET {{host}}/data_category/?resource_type=system

```
[
    {
        "fides_key": "my_test_system",
        "organization_fides_key": "default_organization",
        "tags": null,
        "name": "Test system",
        "description": "Collect data about our users for marketing.",
        "parent_key": null,
        "is_default": false
    }
]

```json

adamsachs: good find!

pattisdr (approved):

PR #3158, opened 2023-04-25
Support pseudonymous consent requests with `fides_user_device_id`, update `fides_consent` cookie format, and add smoke tests for consent management

allisonking:
Closes https://github.com/ethyca/fidesplus/issues/807

### Description

This PR adds a `fides_user_device_id` to every consent request creation, which will only be registered and stored by the backend as a last resort, if no other identity field is available to store the consent preferences against.

### Code Changes

* [x] Adds getters + settings for `fides_user_device_id` cookie
* [x] Omit `email` and `phone_number` from consent request payload if neither are available
* [x] Adds `fides_user_device_id` to consent request payload _always_

### Steps to Confirm

* [x] `nox -s dev`
* [x] `cd clients/privacy-center` + `npm run dev`
* [x] Submit a consent request with no identity data present -> observe that preferences are stored based on the _cookie_
* [x] Submit a consent request _with_ identity data present -> observe that preferences are stored based on the _identity_ despite the cookie's value remaining the same

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`


clients/privacy-center/components/modals/consent-request-modal/ConsentRequestForm.tsx
```diff
@@ -254,7 +257,7 @@ const ConsentRequestForm: React.FC<ConsentRequestFormProps> = ({
             _active={{ bg: "primary.500" }}
             colorScheme="primary"
             isLoading={isSubmitting}
-            isDisabled={isSubmitting || !(isValid && dirty)}
+            // isDisabled={isSubmitting || !(isValid && dirty)}
```
seanpreston, line 260: We'll need to switch this to selectively disable the button only when no field inside `config.consent.button.identity_inputs` is `"required"`

seanpreston:
Hey @allisonking — I can't add you as a reviewer since you're the author, but this is worth another peek in the morning :)

clients/privacy-center/common/hooks/useCookie.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  const rootDomain = window.location.hostname.split(".").slice(-2).join(".");
+  const uuid = crypto.randomUUID();
+
+  setCookie(
+    FIDES_USER_DEVICE_ID_COOKIE_NAME,
+    uuid,
+    {
+      domain: rootDomain,
+      expires: MAX_AGE_DAYS,
```
seanpreston, line 27: `1` seems like a reasonable value here, but even better would probably be ~an hour

clients/privacy-center/components/modals/consent-request-modal/ConsentRequestForm.tsx
```diff
@@ -42,16 +43,18 @@ const useConsentRequestForm = ({
...
   const identityInputs =
     config.consent?.button.identity_inputs ?? defaultIdentityInput;
+  const fidesUserDeviceId = useFidesUserDeviceIdCookie();
   const toast = useToast();
   const formik = useFormik({
     initialValues: {
-      email: "",
-      phone: "",
+      email: undefined,
+      phone: undefined,
```
seanpreston, line 51: This was done to avoid the fields being sent in the eventual payload (as they were before when the defaults were `""`). React doesn't like it, because it throws a warning when the inputs are used now:

```
xt-dev.js:20 Warning: A component is changing an uncontrolled input to be controlled. This is likely caused by the value changing from undefined to a defined value, which should not happen. Decide between using a controlled or uncontrolled input element for the lifetime of the component. More info: https://reactjs.org/link/controlled-components
```

src/fides/api/ops/api/v1/endpoints/consent_request_endpoints.py
```diff
@@ -518,6 +520,11 @@ def _get_or_create_provided_identity(
                     "encrypted_value": {"value": identity_data.phone_number},
                 },
             )
+    elif target_identity_type == ProvidedIdentityType.fides_user_device_id.value:
+        identity = get_or_create_fides_user_device_id_provided_identity(
+            db, identity_data
+        )
```
seanpreston, line 526: Thanks @pattisdr, this was super handy

src/fides/api/ops/api/v1/endpoints/consent_request_endpoints.py
```diff
@@ -518,6 +520,11 @@ def _get_or_create_provided_identity(
                     "encrypted_value": {"value": identity_data.phone_number},
                 },
             )
+    elif target_identity_type == ProvidedIdentityType.fides_user_device_id.value:
+        identity = get_or_create_fides_user_device_id_provided_identity(
```
pattisdr, line 524: Our workflows overlap more than I thought.  My workflow also uses this endpoint but will need to be able to save preferences with a  fides user device id _and_ an email/phone_number.

I think I can handle the fact that ConsentRequests are still restricted to being against one type of ProvidedIdentity downstream in my new endpoints, reticketing here:https://github.com/ethyca/fidesplus/issues/845

src/fides/api/ops/api/v1/endpoints/consent_request_endpoints.py
```diff
@@ -167,10 +170,10 @@ def create_consent_request(
             "Application redis cache required, but it is currently disabled! Please update your application configuration to enable integration with a redis cache."
         )
 
-    if not data.email and not data.phone_number:
+    if not data.email and not data.phone_number and not data.fides_user_device_id:
         raise HTTPException(
             HTTP_400_BAD_REQUEST,
-            detail="An email address or phone number identity is required",
+            detail="An email address, phone number, or fides_user_device_id is required",
```
pattisdr, line 176: I haven't tested this yet but I assume if identity verification is on, there would be more we would need to do to bypass this.

clients/privacy-center/common/hooks/useCookie.ts
```diff
@@ -0,0 +1,49 @@
...
+const CODEC: Types.CookieCodecConfig<string, string> = {
+  decodeName: decodeURIComponent,
+  decodeValue: decodeURIComponent,
+  encodeName: encodeURIComponent,
+  encodeValue: encodeURIComponent,
+};
+
+export const setFidesUserDeviceIdCookie = () => {
+  // Generates a new `fides_user_device_id` and stores it to the cookie
+  if (typeof document === "undefined") {
```
allisonking, line 15: a docstring in this format helps IDEs tell you what a func is about 

```suggestion
/**
 * Generates a new `fides_user_device_id` and stores it to the cookie
 */
export const setFidesUserDeviceIdCookie = () => {
  if (typeof document === "undefined") {
```

clients/privacy-center/common/hooks/useCookie.ts
```diff
@@ -0,0 +1,49 @@
+import { getCookie, setCookie, Types } from "typescript-cookie";
```
allisonking, line 1: neat, this lib is probably a lot better for setting cookies! I remember not using it before since we were only reading, but it's a good call to bring it in now that we are setting as well

clients/privacy-center/components/modals/consent-request-modal/ConsentRequestForm.tsx
allisonking: yeah, this is an annoying thing about forms and react. we work around it by transforming `undefined` to `str`s before marshalling it into a form, and then un-marshalling it before submission. 

something like these two:

https://github.com/ethyca/fides/blob/1550545d8732e88b343f2a5b10958fdf2ac6e638/clients/admin-ui/src/features/common/form/inputs.tsx#L50-L56

https://github.com/ethyca/fides/blob/1550545d8732e88b343f2a5b10958fdf2ac6e638/clients/admin-ui/src/features/privacy-notices/form.ts#L22-L43

I can take a look later if you'd like, to see what makes sense here, just let me know!

clients/privacy-center/components/modals/consent-request-modal/ConsentRequestForm.tsx
allisonking: 👍 we can do that through the `validationSchema`, i.e. pass a different schema depending on how the config looks. let me know if that makes sense or if you'd like to pair on it!

clients/privacy-center/pages/_app.tsx
```diff
@@ -17,23 +17,30 @@ import "@fontsource/inter/700.css";
...
-          <Component {...pageProps} />
-        </ChakraProvider>
-      </PersistGate>
-    </Provider>
-  </SafeHydrate>
-);
+const MyApp = ({ Component, pageProps }: AppProps) => {
+  // Ensure each privacy centre has a unique fides_user_device_id
+  // for recording anonymised consent request preferences
+  useFidesUserDeviceIdCookie();
```
allisonking, line 31: since `useFidesUserDeviceIdCookie` automatically sets if it doesn't exist already, I'm thinking maybe we don't actually need to call it here—as long as everyone who needs the cookie uses this func, then we will always have a uuid, right? i.e. we could only call this in `ConsentRequestForm.tsx` instead, and we'd still have the same functionality?

(I know I was the one who suggested putting it in `_app.tsx` in the first place, but now I'm second guessing 😄 )

allisonking (commented):
haven't tested yet, but leaving a few comments! nice job figuring this out @seanpreston !

clients/privacy-center/common/hooks/useCookie.ts
seanpreston: This is used already in `fides-consent.js` so seemed like a safe bet to use here too

src/fides/api/ops/api/v1/endpoints/consent_request_endpoints.py
seanpreston: This isn't intended to be used at all when identity verification is switched on by the server, since a `fides_user_device_id` can't really be used to verify anything. Looks like I'll need to update this block to account for that, thanks!

clients/privacy-center/components/modals/consent-request-modal/ConsentRequestForm.tsx
allisonking: I looked into this a bit and there's already logic in `validationSchema` which only makes the field required if the config says to, so I think we are good there.

it looks like the problem is actually that the form isn't dirty—we make it so you can't submit the form if it isn't dirty. so you can type something into the email form and then erase it and then that actually lets you submit, which likely isn't what we want.

I think we could change the `dirty` check such that it's only required to be dirty if there is a required field

```typescript
  // figure out all inputs that are required
  const requiredInputs = Object.entries(identityInputs).filter(
    ([, required]) => required === "required"
  );
  // it's ok to bypass the dirty check if there are no required inputs
  const dirtyCheck = requiredInputs.length === 0 ? true : dirty;

  ...
  // in <Button />
  isDisabled={isSubmitting || !(isValid && dirtyCheck)}
```

that said, that UX feels a little weird, to just bypass the email field

clients/privacy-center/common/hooks/useCookie.ts
NevilleS: FWIW: since the `fides-consent` package already exports `getConsentCookie` and `setConsentCookie` methods, I find this pretty duplicative. That means that we'll be saving two cookies to the browser (`fides_consent` and `fides_user_device_id`) and we'll need to support that going forward too...

The way we've built `fides-consent` right now, though, makes it pretty hard to work on it. This is *dramatically* better on the monorepo branch so within a few weeks this'll be a non-issue, but the timing of this is really unfortunate...

I'd say this: let's get it working, tested, and ready to go... and I _might_ try to refactor this so that we just save this data into the existing cookie. It's *much* more maintainable for us to just extend that format (we're going to need to do that, anyways) instead of trying to maintain two cookie formats

NevilleS (changes requested):
Hesitant to ship this right now - I think we're being a bit fast-and-loose with the cookie! Will do some testing and make some actual concrete suggestions though.

clients/privacy-center/components/modals/consent-request-modal/ConsentRequestForm.tsx
allisonking: should be fixed by https://github.com/ethyca/fides/pull/3158/commits/8134418705dcaec8b2a3b3a35b2377dbdfe15f0c

I checked and the payload doesn't send the email/phone fields unless something is filled in 👍

clients/privacy-center/pages/_app.tsx
allisonking: moved it into the `onSubmit` of `ConsentRequestForm`

clients/privacy-center/common/fides-user-cookie.ts
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  return getCookie(FIDES_USER_DEVICE_ID_COOKIE_NAME, CODEC);
+};
+
+/**
+ * Retrieves the fides user device UUID as stored in the `fides_user_device_id` cookie
+ * If no cookie is found, sets a new UUID on the cookie and returns the value
+ */
+export const getFidesUserDeviceUuid = () => {
```
allisonking, line 50: fyi I renamed this—again, I know I was the one who suggested to make this a hook with `use...`, but it's not actually a hook since it doesn't use react state or anything. it's just a plain ol' function, so best not to prepend it with `use` which implies a hook (and implies you can only use it at the beginning of a component, which isn't true and would be very limiting!) 

that said, I don't love this name since it gets but also sets if it can't get. there must be a word for that, right... it's escaping me at the moment 😄

NevilleS:
So this now works great, and I refactored the UUID cookie storage bit extensively here: https://github.com/ethyca/fides/pull/3189

However I just finally tried out the UX for the email input, and here are the three configurations:

| Required Email (default) | No Email | Optional Email |
|---|---|---|
|`"identity_inputs": { "email": "required" }`|`"identity_inputs": {}`|`"identity_inputs": { "email": "optional" }`|
| image|image|image|

The desired flow for this was to have it skip the modal entirely, and we didn't _need_ to support optional email. Both the "No Email" and "Optional Email" flow are pretty confusing as-is, though.

So to get this release-ready we should:
- [ ] Update the "No Email" flow to skip the consent request modal entirely
- [ ] Add configurable text to the "Required/Optional Email" modal to explain why the user should provide their email

NevilleS (approved):
OK, added some new configurable text to the consent modal. It's still janky when email is removed (just becomes a... notice?) but revising that UX will require requiring the consent flow, because today it _assumes_ a modal is required to submit the initial consent request form (to identify the user and fetch existing preferences).

| Required Email (default) | No Email | Optional Email |
|---|---|---|
|`"identity_inputs": { "email": "required" }`|`"identity_inputs": {}`|`"identity_inputs": { "email": "optional" }`|
| image| image| image |

Issue #3123, opened 2023-04-21
Table for privacy experiences


allisonking:
### Is your feature request related to a specific problem?

On the privacy experiences page, we need a table to display the experiences

### Describe the solution you'd like

Use the new `FidesTable` component and populate it with privacy experience data.

![image](https://user-images.githubusercontent.com/24641006/233683175-3c2fdebc-c643-45c2-8bdd-4797387b116b.png)

We won't render any flags yet since our region breakdown is still up in the air.

Enabling should issue `PATCH api/v1/privacy_experience/<privacy_experience_id>/`

### Describe alternatives you've considered, if any

_A description of any alternative solutions or features you've considered._

### Additional context

Blocked by https://github.com/ethyca/fides/issues/3122

allisonking:
plopping some location thoughts here (thanks @pattisdr for the background, maybe we can do some location coordination here!)

[per Neville](https://ethyca.slack.com/archives/C04S7K0T2V7/p1682029136225489?thread_ts=1682027538.778159&cid=C04S7K0T2V7), we'll want to use the [ISO_3166-1_alpha-2](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2) standard for regions which is what Ethyca Pro uses, though we are not currently using those for privacy notices/experiences (we are using a small list coded into the backend).

interestingly, dataset `third_country_transfers` [appear to use alpha-3](https://github.com/ethyca/fideslang/blob/main/src/fideslang/models.py/#L545-L547) 🤔 

in any case, we'll need a way to go from `ISO_3166-1_alpha-2` with a region name to a human friendly name... on the frontend we've used [i18n-iso-countries](https://www.npmjs.com/package/i18n-iso-countries) for `third_country_transfers` which lets us get a friendly name for a _country_ but it does not handle "subdivisions" (such as a state within the US)

then there's the matter of grouping by larger region, i.e. EU or US. with the current format, we could just check if every region in the list starts with i.e. `eu_` or `us_`, but with ISO_3166 we'd need to do something else to tell (if I set my VPN to Madrid, ethyca pro tells me I'm in `{"country":"ES","ip":"...","location":"ES-MD","region":"MD"}` which does not give any indication of `EU`). even to go from `US-IA` to "Iowa (IA)" like in the design, we'd probably need another library to handle subdivision names... a quick google shows there's at least one, though [it hasn't been updated since 2017](https://github.com/olahol/iso-3166-2.js/).

[...]

Issue #3122, opened 2023-04-21
Initial page for privacy experiences


allisonking:
### Is your feature request related to a specific problem?

Create a page for privacy experience work. 

### Describe the solution you'd like

* Standard FE page setup
* Potentially do not render the page in the nav if there are no privacy notices yet?
* Set up redux to query `GET api/v1/privacy_experience/?show_disabled=true&has_notices=true`
* Nested under `Consent` side nav

We might also need to query privacy notices and do a join here, depending on if the experience API will return consent mechanisms

### Describe alternatives you've considered, if any

_A description of any alternative solutions or features you've considered._

### Additional context

Won't be able to use "real" data until https://github.com/ethyca/fides/issues/3019 but we can still mock things out in cypress if needed

Some questions for @mfbrown :
* Should we put this behind the same privacy notice feature flag? Or should it be a separate flag?
* Empty state discussion
  * When there are no notices, there will be no experiences. We could mark the page as disabled or not render it at all if there are no notices, but I'm a little wary of this because of our regrets with the progressive nav
    * for example, if a user creates a privacy notice linked to a data use `advertising` which has a corresponding experience, and that is their only notice/experience, but then they remove the system that had that data use, the experience page would totally disappear, and it wouldn't be obvious why since the last edit was on a system page, not anything consent related
  * I'd push for some simple text that says you need privacy notices first since that would also keep the nav config logic simpler. thoughts?

mfbrown:
@allisonking I think keeping this a separate feature flag for now makes sense (Dev settable only) and your instincts sound right regarding the simple option for an empty state.

Roger-Ethyca:
moving to done

Roger-Ethyca:
moving to done

PR #3186, opened 2023-04-27
Privacy experiences initial pages

allisonking:
Closes https://github.com/ethyca/fides/issues/3122, closes https://github.com/ethyca/fides/issues/3123

### Code Changes

* [x] Update TS autogenerated types
* [x] Initial page for privacy experiences (table view, edit view)
* [x] Table for privacy experiences (easy to do thanks to @TheAndrewJackson 's table refactor!)
* [x] Tweaks to side nav that came up now that we have multiple child side navs
  * [x] Consent page now has its own "page" which just redirects to notices. This makes the "tree" view of the side nav work properly (before I just had the Consent page linking to "/consent/privacy-notices" but that makes "/consent/" not highlight when under "/consent/privacy-experiences" since it doesn't look like part of the same tree)
* [x] Cypress tests

### Steps to Confirm

This is a little tricky to test right now since the backend isn't in `main` yet, so we can't just use fidesplus double edge. I've found the easiest thing to do is to run the OSS fides backend, then in `features.slice.ts`, set `plus=true` instead of the plus health check.

You'll also need a privacy experiences json file to load in, similar to how we load in privacy notices while there aren't any out of the box ones yet.

<details>
<summary>privacyExperiences.json</summary>

```json
[
  {
    "component": "overlay",
    "regions": ["eu_fr", "eu_ie"],
    "delivery_mechanism": "banner"
  },
  {
    "component": "privacy_center",
    "regions": ["us_ca"],
    "delivery_mechanism": "link"
  },
  {
    "component": "overlay",
    "regions": ["eu_de"],
    "delivery_mechanism": "link"
  }
]

```

</details>

* Click the "Privacy requests" nav link, then in the side nav, click "Privacy experiences"
* If you've loaded the experiences above, you should see them in the table
* You should be able to click into a row and go to a work in progress page

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

https://user-images.githubusercontent.com/24641006/235742903-870975e0-c3f9-4907-86ff-6ce28ecc5a8c.mov

A few things about this table:
image

* We're punting on how the locations should show up for now. Still need to think through how we will group them and display them
* The privacy experience table says "Last update" while the notice table says "Last modified". Should we make these consistent @mfbrown ? Simon [deferred to you](https://www.figma.com/file/dmEwdK3xZwjKfGVQ9t66xe?node-id=5292:164434#431038877) :) 
* ~~The enable/disable button doesn't currently work with the backend yet due to the [backend doing more validation than necessary](https://ethyca.slack.com/archives/C04S7K0T2V7/p1683047701741599?thread_ts=1683047447.874469&cid=C04S7K0T2V7) right now. will be fixed soon!~~ Fixed!

image

* I copied the empty state from notices. I feel like the copy still applies since data uses drive everything, but it might read weird, not sure. Let me know if you want to change this copy @mfbrown


clients/admin-ui/src/features/common/table/FidesTable.tsx
```diff
@@ -133,7 +134,7 @@ export const FidesTable = <T extends FidesObject>({
                     ? { backgroundColor: "gray.50", cursor: "pointer" }
                     : undefined
                 }
-                data-testid={`row-${rowName}`}
+                data-testid={`row-${rowName ?? rowId}`}
```
allisonking, line 137: privacy experiences don't have a `name` field, so rely on `id` for the test id instead

clients/admin-ui/src/features/privacy-experience/privacy-experience.slice.ts
```diff
@@ -0,0 +1,109 @@
...
+} from "~/types/api";
+
+export interface State {
+  page?: number;
+  pageSize?: number;
+}
+
+const initialState: State = {
+  page: 1,
+  pageSize: 10,
```
TheAndrewJackson, line 19: Should the pagesize be a small number like 10 when there is no way to switch between pages?

TheAndrewJackson (approved):
Looks great! Just one small nit about the pageSize

clients/admin-ui/src/features/privacy-experience/privacy-experience.slice.ts
allisonking: good idea, I'll increase to 50 😎

PR #3188, opened 2023-04-28
refactor privacy declaration upsert to maintain constant IDs

### LINKED ISSUE #3187: System API to maintain constant `PrivacyDeclaration.id`s when possible

As it stands now, when updating `PrivacyDeclaration` records by means of the `/system` API (which is the only API currently available to update `PrivacyDeclaration` records), the `id`s are effectively unstable, since the `PrivacyDeclaration` records are deleted and replaced "behind the scenes" on every update API call, no matter what. While this approach was taken because it was a simpler backend implementation, it has side effects that make it problematic, specifically for the "custom fields" use case we are trying to support, which require a constant identifier (i.e. the `id` attribute) to point to.

Update our `system` APIs to update `PrivacyDeclaration` records _in place_ when applicable, rather than relying on the simpler but more naive kill-and-fill approach taken by the update APIs currently.

In order to update the records in place, the BE will use the `data_use` and `name` (if provided) attributes as lookup/uniqueness criteria for a `PrivacyDeclaration` on a given `System`. For any `PrivacyDeclaration`s specified on a `PUT` call, if a "matching" `PrivacyDeclaration` is not found, then a new one will be created; if an existing `PrivacyDeclaration` tied to the given system is _not_ specified on a `PUT` call on that system, then the existing `PrivacyDeclaration` record will be deleted. This will keep the API semantics basically the same as it has always been, with the added benefit of providing some stability to the `id`s now associated with `PrivacyDeclaration` records.

We considered, and still want to, create an API (more imperative in nature) that's dedicated specifically to managing `PrivacyDeclaration` records. It would allow us to avoid some of the assumptions and complexities that we're arriving at when trying to manage them _through_ the `System`.

That shouldn't be too tricky in and of itself, now that the `PrivacyDeclaration`s are formalized in their own table but it will require some care to also reconcile with the system-based endpoints that are leveraged by the CLI and related workflows. More than anything, it's just a more substantive update than we're ready to take on in the tight timeline that this enhancement is needed.

### Additional context

To make this work smoothly with the UI - namely to avoid a nasty edge case brought up by @pattisdr - we will need the UI to "lock down" the `Data Use` and `Processing Activity` (which translates to `name` on the BE) fields as read-only once a `PrivacyDeclaration` has been created in the UI. The edge case that this avoids (with some paraphrasing of a message from @pattisdr):
> say I have PD 1 advertising and PD 2 improve, and then FE makes a request to make PD 1 improve, and PD 2 advertising.  Swaps data uses.
BE looks up PD on "improve" which actually returns PD2 and replaces it with contents of PD1's data and vice versa.
So the contents end up completely switching on the two records. This would be fine if it weren't for custom fields, which are going to be tied to the records based on their IDs, not their contents, meaning that they'd be "switched" from underneath the user.

adamsachs:
Closes #3187 

### Code Changes

* change privacy declaration upsert logic used by `/system` APIs to update `PrivacyDeclaration` records in place when "matched" using `data_use` and `name` as matching criteria
* add in BE validation to prevent a request from specifying "duplicate" `PrivacyDeclaration`s based on the above matching criteria
    * this validation is now important on the BE since our upsert logic now relies on this assumption being true
* add automated test coverage around this use cases
    * there may still be some edge cases we need to cover here...

### Steps to Confirm

* i did some testing in the UI with these changes, while checking the DB records, and things generally looked good. but would appreciate others hitting this hard with the API and with the UI
    * NOTE - as of now, if you change the "data use" or the "processing activity" of an existing declaration in the UI, it's going to create a new DB record. that's expected, and we will look to prevent this edge case by making those fields "read only" in the UI after creation.

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### Description Of Changes

See issue for more context. This is needed to get custom fields working with privacy declarations, i.e. "data use declarations".

### DIFF
--- a/src/fides/api/ctl/routes/system.py
+++ b/src/fides/api/ctl/routes/system.py
@@ -224,16 +224,42 @@ async def upsert_privacy_declarations(
 ) -> None:
     """Helper to handle the specific upsert logic for privacy declarations"""
 
-    async with db.begin():
-        # clear out any existing privacy declarations on the System
-        for existing_declaration in system.privacy_declarations:
-            await db.delete(existing_declaration)
+    def privacy_declaration_logical_id(
+        privacy_declaration: PrivacyDeclaration,
+    ) -> str:
+        """
+        Internal helper to standardize a logical 'id' for privacy declarations.
+        As of now, this is based on the `data_use` and the `name` of the declaration, if provided.
+        """
+        return f"{privacy_declaration.data_use}:{privacy_declaration.name or ''}"
 
-        # iterate through declarations specified on the request and create them
+    async with db.begin():
+        # map existing declarations by their logical identifier
+        existing_declarations: Dict[str, PrivacyDeclaration] = {
+            privacy_declaration_logical_id(existing_declaration): existing_declaration
+            for existing_declaration in system.privacy_declarations
+        }
+
+        # iterate through declarations specified on the request and upsert
+        # looking for "matching" existing declarations based on data_use and name
         for privacy_declaration in resource.privacy_declarations:
+            # prepare our 'payload' for either create or update
             data = privacy_declaration.dict()
-            data["system_id"] = system.id  # add FK back to system
-            PrivacyDeclaration.create(db, data=data)
+            data["system_id"] = system.id  # include FK back to system
+
+            # if we find matching declaration, remove it from our map
+            if existing_declaration := existing_declarations.pop(
+                privacy_declaration_logical_id(privacy_declaration), None
+            ):
+                # and update existing declaration *in place*
+                existing_declaration.update(db, data=data)
+            else:
+                # otherwise, create a new declaration record
+                PrivacyDeclaration.create(db, data=data)
+
+        # delete any existing privacy declarations that have not been "matched" in the request
+        for existing_declarations in existing_declarations.values():
+            await db.delete(existing_declarations)
 
 
 async def update_system(resource: SystemSchema, db: AsyncSession) -> Dict:
--- a/src/fides/api/ctl/routes/system.py
+++ b/src/fides/api/ctl/routes/system.py
@@ -67,15 +67,15 @@ def get_system(db: Session, fides_key: str) -> System:
     return system
 
 
-async def validate_privacy_declarations_data_uses(
-    db: AsyncSession, system: SystemSchema
-) -> None:
+async def validate_privacy_declarations(db: AsyncSession, system: SystemSchema) -> None:
     """
-    Ensure that the `PrivacyDeclaration`s on the provided `System` resource reference
-    valid `DataUse` records.
+    Ensure that the `PrivacyDeclaration`s on the provided `System` resource are valid:
+     - that they reference valid `DataUse` records
+     - that there are not "duplicate" `PrivacyDeclaration`s as defined by their "logical ID"
 
     If not, a `400` is raised
     """
+    logical_ids = set()
     for privacy_declaration in system.privacy_declarations:
         try:
             await get_resource(
@@ -88,6 +88,14 @@ async def validate_privacy_declarations_data_uses(
                 status_code=HTTP_400_BAD_REQUEST,
                 detail=f"Invalid privacy declaration referencing unknown DataUse {privacy_declaration.data_use}",
             )
+        logical_id = privacy_declaration_logical_id(privacy_declaration)
+        if logical_id in logical_ids:
+            raise HTTPException(
+                status_code=HTTP_400_BAD_REQUEST,
+                detail=f"Duplicate privacy declarations specified with data use {privacy_declaration.data_use}",
+            )
+
+        logical_ids.add(logical_id)
 
 
 @system_connections_router.get(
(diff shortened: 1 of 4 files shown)


src/fides/api/ctl/routes/system.py
```diff
@@ -219,21 +227,48 @@ async def upsert(
...
+            ):
+                # and update existing declaration *in place*
+                existing_declaration.update(db, data=data)
+            else:
+                # otherwise, create a new declaration record
+                PrivacyDeclaration.create(db, data=data)
+
+        # delete any existing privacy declarations that have not been "matched" in the request
+        for existing_declarations in existing_declarations.values():
+            await db.delete(existing_declarations)
```
pattisdr, line 271: What happens if we delete a privacy declaration attached to a custom field here?

src/fides/api/ctl/routes/system.py
```diff
@@ -219,21 +227,48 @@ async def upsert(
...
+
+            # if we find matching declaration, remove it from our map
+            if existing_declaration := existing_declarations.pop(
+                privacy_declaration_logical_id(privacy_declaration), None
+            ):
+                # and update existing declaration *in place*
+                existing_declaration.update(db, data=data)
+            else:
+                # otherwise, create a new declaration record
+                PrivacyDeclaration.create(db, data=data)
```
pattisdr, line 267: oh interesting in testing with when adjusting name or data use, I'm usually seeing a new one get created and the old one get deleted which makes sense in hindsight. understood though that we're blocking these edits from the UI

src/fides/api/ctl/routes/system.py
```diff
@@ -88,6 +88,14 @@ async def validate_privacy_declarations_data_uses(
                 status_code=HTTP_400_BAD_REQUEST,
                 detail=f"Invalid privacy declaration referencing unknown DataUse {privacy_declaration.data_use}",
             )
+        logical_id = privacy_declaration_logical_id(privacy_declaration)
+        if logical_id in logical_ids:
```
pattisdr, line 92: If we're looking at `logical_ids` to determine data use uniqueness then it's possible at least from the backend that two privacy declarations could be added with different names but the same data uses, causing the same data use to be added twice.

src/fides/api/ctl/routes/system.py
adamsachs: great question. at this point, none of our deletes of primary artifacts referenced by custom fields will cascade to the custom field, because the custom field has a "soft" reference. my understanding is that this is a known and accepted limitation at this point. the custom field will essentially be orphaned (and not really accessible in any way, at least not through the UI).

just cc'ing @mfbrown here for confirmation. pretty sure this is the same behavior we've got for custom fields associated with `System`s generally.

src/fides/api/ctl/routes/system.py
adamsachs: that's right, and my understanding is that is actually desired behavior. we want the `name` (which shows up as `processing activity` in the UI), to be a mechanism that user's can leverage to create >1 privacy declarations on the same system with the same data use.  my understanding is that it's an edge case, but it's one we want/need to continue supporting at this time.

src/fides/api/ctl/routes/system.py
adamsachs: yup, good observation. but that's exactly why we want to block those edits from the UI!

agreed that it's generally a bit fraught. goes back to the broader discussion about how we probably want to move more toward an imperative API squared directly on `PrivacyDeclaration`s (or maybe they will be renamed) in the medium term future.

pattisdr (approved):
Thanks for clarifying some of my comments here, looks good to me

adamsachs:
i'm pretty clearly hitting the line that codecov says i'm missing in my `test_api.py` tests i've added, so not sure why it's complaining. anyway, going to go ahead and merge now - thanks for the speedy review on this @pattisdr!

adamsachs:
> i'm pretty clearly hitting the line that codecov says i'm missing in my test_api.py tests i've added, so not sure why it's complaining. anyway, going to go ahead and merge now - thanks for the speedy review on this @pattisdr!


actually, a quick thought on the codecov "miss" -- maybe because the test code asserts the errored HTTP status codes without actually doing any explicit _catching_ of the `HTTPException`, it's showing up as a false negative? perhaps that's just some oddity in the way our test API client works and interacts with codecov?

it's a total shot in the dark, and i'm not very concerned (especially with this PR specifically), but @ThomasLaPiana maybe something you have thoughts on, since i've noticed some false negative codecov misses recently. anyway, still merging here!

ThomasLaPiana:
> > i'm pretty clearly hitting the line that codecov says i'm missing in my test_api.py tests i've added, so not sure why it's complaining. anyway, going to go ahead and merge now - thanks for the speedy review on this @pattisdr!
> 
> actually, a quick thought on the codecov "miss" -- maybe because the test code asserts the errored HTTP status codes without actually doing any explicit _catching_ of the `HTTPException`, it's showing up as a false negative? perhaps that's just some oddity in the way our test API client works and interacts with codecov?
> 
> it's a total shot in the dark, and i'm not very concerned (especially with this PR specifically), but @ThomasLaPiana maybe something you have thoughts on, since i've noticed some false negative codecov misses recently. anyway, still merging here!

I think you're right on the money there! It is probably wanting to test the code after the catch? But anyway, seems fine to me :) 

These are merely guidelines, so we're still capable of making the decision that it's "good enough"!

Issue #3043, opened 2023-04-12
Add consent project skeleton in clients folder


eastandwestwind:
add Preact, and ensure all infra is set up for consent development

Roger-Ethyca:
moving to done

![Image](https://github.com/ethyca/fides/assets/115798183/32a6267a-6335-4e7f-8184-931a2f41ebdd)

Issue #3051, opened 2023-04-12
Render banner and link components


eastandwestwind:
Not yet config/notice-driven, but includes default styles

Default link exists on website (with id `fides--consent--link`), our responsibility is to hide it under certain circumstances

eastandwestwind:
Closed as part of https://github.com/ethyca/fides/issues/3043

Roger-Ethyca:
moving to done

PR #3191, opened 2023-04-29
Add infrastructure for consent banner and link

eastandwestwind:
Closes https://github.com/ethyca/fides/issues/3043 and https://github.com/ethyca/fides/issues/3051

### Code Changes

* [ ] Add consent infrastructure that renders dummy banner and link
* [ ] Uses Preact + TSX
* [ ] Link generation requires an HTML element with the id = `'fides-consent-link'` somewhere on the page. We should either surface this info within the experience / notice flow OR make this configurable in privacy experiences

### Steps to Confirm

Pre-requisites to test:
* [ ] If you haven't already, install Turborepo globally so that you can use it on the command line `npm install turbo --global`
* [ ] In `clients` folder, run `npm install`. 

To test:
* [ ] In `clients/privacy-center` folder, run `turbo run dev`, then navigate to http://localhost:3000/fides-js-components-demo.html
* [ ] To test other configuration, you can edit the `fidesConfig` object passed into `Fides.init()` in `privacy-center/public/fides-js-components-demo.html`

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Follow-up Tickets

- Move `consent-banner.cy.ts` to `fides-js` instead of `privacy-center`- https://github.com/ethyca/fides/issues/3274
- Use constants instead of strings in class / id names in fides-js - https://github.com/ethyca/fides/issues/3287
- Bundle CSS into fides.js script, and add styles to DOM from the package - https://github.com/ethyca/fides/issues/3288
- Add support for consent banner position - https://github.com/ethyca/fides/issues/3294
- Allow mounting fides-js components into a specified HTML container - https://github.com/ethyca/fides/issues/3293
- Privacy experience list API should support filtering for country code - https://github.com/ethyca/fides/issues/3300


clients/fides-js/.eslintrc.json
```diff
@@ -2,12 +2,15 @@
   "extends": [
     "airbnb",
     "airbnb-typescript/base",
-    "prettier"
+    "prettier",
+    "standard",
+    "standard-preact"
```
NevilleS, line 7: FYI: I started tinkering with this branch and the eslint configs you added (`preact`, `standard`, `standard-preact`) caused my linter to crash and not obey the Typescript tsconfig settings... so I removed them for now 🙃 

I think you were probably just experimenting with these anyways

clients/fides-js/package.json
```diff
@@ -23,13 +23,26 @@
     "directory": "clients/fides-js"
   },
   "dependencies": {
+    "eslint-config-preact": "^1.3.0",
+    "npm": "^9.6.5",
+    "preact": "^10.13.2",
```
NevilleS, line 28: Note that only `preact` should be here - the others are dev dependencies. I fixed that 👍

clients/fides-js/package.json
```diff
@@ -23,13 +23,26 @@
...
+    "eslint-plugin-node": "^11.1.0",
+    "eslint-plugin-promise": "^6.1.1",
+    "eslint-plugin-react": "^7.32.2",
+    "eslint-plugin-standard": "^5.0.0",
     "jest": "^29.5.0",
     "tsup": "^6.7.0",
     "typescript": "^4.9.5"
+  },
+  "eslintConfig": {
+    "extends": "preact"
```
NevilleS, line 46: This competes with the extensions in `.eslintrc.json` so I removed it from here

clients/fides-js/src/fides-demo.html
```diff
@@ -35,6 +35,9 @@ <h2>Was the fides global defined?</h2>
 
       <h2>Consent JSON</h2>
       <pre id="consent-json"></pre>
+
+      <h2>Banner Options</h2>
+      <pre id="consent-banner-options"></pre>
```
NevilleS, line 40: Note that I deleted this duplicate demo site on the latest updates to turborepo branch, so this file got deleted when I merged those changes in

clients/fides-js/src/fides.ts
```diff
@@ -14,22 +14,27 @@
...
-import { shopify } from "./integrations/shopify";
-import { ConsentConfig } from "./lib/consent-config";
-import { getConsentContext } from "./lib/consent-context";
-import { CookieKeyConsent, getConsentCookie, makeDefaults } from "./lib/cookie";
+import {gtm} from "./integrations/gtm";
+import {meta} from "./integrations/meta";
+import {shopify} from "./integrations/shopify";
+import {ConsentConfig} from "./lib/consent-config";
+import {getConsentContext} from "./lib/consent-context";
+import {InitFidesConsent} from "./lib/consent";
```
NevilleS, line 22: nit: shouldn't capitalize a function

clients/fides-js/src/fides.ts
```diff
@@ -14,22 +14,27 @@
...
  * Configuration options for the fides.js script
  */
 export interface FidesConfig {
   consent: ConsentConfig;
+  bannerOptions?: ConsentBannerOptions;
 }
 
 type Fides = {
   consent: CookieKeyConsent;
+  getBannerOptions: ConsentBannerOptions;
```
NevilleS, line 37: Shouldn't this be a function, like `getBannerOptions: () => ConsentBannerOptions`

clients/fides-js/src/lib/consent-banner.tsx
```diff
@@ -0,0 +1,122 @@
...
+class ConsentBanner extends Component<BannerProps, BannerState> {
+
+    /**
+     * Builds a button DOM element with the given id, class name, and text label
+     */
+    static buildButton = (id: string, className: string, label?: string, onClick?: () => void) => {
+        return(<button id={id} class={`fides-consent-banner-button ${className}`} onClick={onClick ? onClick : undefined}>
+            {label || ""}
+        </button>);
+    }
```
NevilleS, line 27: Cool - I like seeing some jsx in here 👍 

I'd like us to refactor this further though- it doesn't make much sense to have `buildXYZ`-style functions when you have JSX, we should just have components that we nest.

clients/fides-js/src/lib/consent-banner.tsx
```diff
@@ -0,0 +1,122 @@
...
+    /**
+     * Builds the DOM elements for the consent banner (container, buttons, etc.) and
+     * return a single div that can be added to the body.
+     */
+    buildBanner = (defaults: CookieKeyConsent) => {
+        const [isShown, setisShown] = useState(false);
+        // TODO: support option to specify top/bottom
+
+        // Show the banner after a small delay, to allow animation to occur
+        setTimeout(() => setisShown(true), 100);
```
NevilleS, line 59: Heh, something tells me this is an anti-pattern in preact

clients/fides-js/src/lib/consent-banner.tsx
```diff
@@ -0,0 +1,122 @@
...
+                <div id="fides-consent-banner-description" class="fides-consent-banner-description">
+                    {this.options.labels?.bannerDescription || ""}
+                </div>
+                <div id="fides-consent-banner-buttons" class="fides-consent-banner-buttons">
+                    {ConsentBanner.buildButton(
+                        "fides-consent-banner-button-tertiary",
+                        "fides-consent-banner-button-tertiary",
+                        this.options.labels?.tertiaryButton,
+                        this.navigateToPrivacyCenter,
+                    )}
```
NevilleS, line 71: This should be something like `<ConsentBannerButton type="tertiary" onClick={navigateToPrivacyCenter} />`

clients/fides-js/src/lib/consent-banner.tsx
```diff
@@ -0,0 +1,122 @@
...
+                </div>
+            </div>);
+    };
+
+    render(): VNode | null {
+        let bannerBuild = null
+        try {
+            if (hasSavedConsentCookie()) {
+                // todo- check for device id too
```
NevilleS, line 104: This is implemented now once I push the merged commits in 👍

clients/fides-js/src/lib/consent-link.tsx
```diff
@@ -0,0 +1,54 @@
...
+    render(): VNode | null {
+        let linkBuild = null
+        // If the user provides any extra options, override the defaults
+        try {
+            if (hasSavedConsentCookie()) {
+                // todo- check for device id too
+                // also does the existence of a device id prevent links from rendering too, or just banners?
+                debugLog("Fides consent cookie already exists, skipping banner initialization!");
+                return null;
+            }
```
NevilleS, line 43: Why would the link rendering depend on the cookie?

clients/fides-js/src/lib/consent-utils.ts
```diff
@@ -0,0 +1,77 @@
+import {ConsentBannerOptions} from "~/lib/consent-types";
+
+
+// Adapted from https://gist.github.com/henrik/1688572?permalink_comment_id=4317520#gistcomment-4317520o
+// (NOTE: Surprisingly, there's not really a list of these anywhere easily...?)
+const EU_COUNTRY_CODES = [
```
NevilleS, line 6: TBH I really don't know where this should live. See this discussion too... https://github.com/ethyca/fides/issues/3123#issuecomment-1528146093

clients/fides-js/src/lib/consent.tsx
```diff
@@ -0,0 +1,305 @@
...
+
+    return true;
+}
+
+/**
+ * Fetch the user's geolocation from an external API
+ */
+const getLocation = async (): Promise<UserGeolocation> => {
+    debugLog("Running getLocation...");
+    const geolocationApiUrl = "http://dev-cdn-api.ethyca.com/location"; // TODO: replace with prod
```
NevilleS, line 56: Note that this should be a config option (with a default)

clients/fides-js/src/lib/consent.tsx
```diff
@@ -0,0 +1,305 @@
...
+        }
+        debugLog(`Country ${location.country} not included in isEnabledCountries, banner must be hidden. isBannerEnabledForLocation = false`);
+        return false;
+    }
+
+    debugLog("No location-specific rules matched, assume banner must be shown. isBannerEnabledForLocation = true");
+    return true;
+}
+
+const defaultStyle = `:root {
```
NevilleS, line 108: Duplicated from banner.css, can this one get removed?

clients/fides-js/src/lib/consent.tsx
```diff
@@ -0,0 +1,305 @@
...
+        {defaultStyle}
+    </style>)
+}
+
+/**
+ * Initialize the Fides Consent Banner or Link, with optional extraOptions to override defaults.
+ *
+ * (see the type definition of ConsentBannerOptions for what options are available)
+ */
+export const InitFidesConsent = async(defaults: CookieKeyConsent, extraOptions?: ConsentBannerOptions): Promise<void> => {
```
NevilleS, line 255: I *really* dislike that the consent `defaults` are forced to be passed into this. This was a messy part of my POC so I do want to untangle that- it doesn't provide good separation of concerns for the banner code itself to be aware of the consent preferences configs, etc.

clients/privacy-center/cypress/e2e/consent-banner.cy.ts
```diff
@@ -0,0 +1,233 @@
...
+declare global {
+    namespace Cypress {
+        interface Chainable {
+            visitConsentDemo(bannerOptions?: ConsentBannerOptions): Chainable<any>;
+        }
+    }
+}
+
+
+describe("Consent banner", () => {
```
NevilleS, line 21: Hah, awesome! I'm glad these tests survived the wash.

What do you think about moving this to fides-js/cypress instead? I feel like it might make more sense to have the tests that are very focused on fides-js over there.

clients/privacy-center/cypress/e2e/consent-banner.cy.ts
```diff
@@ -0,0 +1,233 @@
...
+                //     </button>
+                //     <button id='fides-consent-banner-secondary-button' class='fides-consent-banner-button fides-consent-banner-secondary-button'>
+                //       {labels.secondaryButton}
+                //     </button>
+                //     <button id='fides-consent-banner-primary-button' class='fides-consent-banner-button fides-consent-banner-primary-button'>
+                //       {labels.primaryButton}
+                //     </button>
+                //   </div>
+                // </div>
+                cy.get("div#fides-consent-banner.fides-consent-banner").within(() => {
```
NevilleS, line 112: Hopefully this was useful as you rewrote in Preact 😄

clients/privacy-center/cypress/e2e/consent-banner.cy.ts
```diff
@@ -0,0 +1,233 @@
...
+                    expect(cookieKeyConsent).property("data_sales").is.eql(true);
+                    expect(cookieKeyConsent).property("tracking").is.eql(true);
+                })
+                cy.contains("button", "Accept Test").should("not.be.visible");
+            });
+
+            it("should support rejecting all consent options", () => {
+                cy.contains("button", "Reject Test").should("be.visible").click();
+                // TODO: consent.cy.ts uses cy.waitUntil to wait longer for the cookie to save in CI - will this be needed?
+                cy.getCookie(CONSENT_COOKIE_NAME).should("exist").then((cookie) => {
```
NevilleS, line 152: It wil be very much needed! I lost like an hour of sanity fighting with cookies in CI. There's a helper now called `cy.waitUntilCookieExists(...)` that you should use, see examples in consent.cy.ts

NevilleS (changes requested):
Looking hawt! Couple comments on this WIP

clients/fides-js/src/lib/consent-banner.tsx
eastandwestwind: yeahhh maybe something more like https://stackoverflow.com/a/61390352

clients/fides-js/src/lib/consent-link.tsx
eastandwestwind: I just didn't see this acceptance criteria explicitly addressed. I'll clarify with @mfbrown

clients/fides-js/src/lib/consent.tsx
eastandwestwind: Oh right, I still haven't gotten standalone css loading properly using tsup. I'll aim for this today/tomorrow

clients/fides-js/src/lib/consent-link.tsx
mfbrown: Apologies for the confusion here. The link should always be present in the locations for which it is configured.

clients/privacy-center/public/fides-consent-demo.html
```diff
@@ -19,19 +19,23 @@
 
   <body>
     <main>
-      <h1>fides.js demo page</h1>
+      <h1>fides-consent.js demo page</h1>
```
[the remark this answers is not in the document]
eastandwestwind: oops 👍

clients/privacy-center/cypress/e2e/consent-banner.cy.ts
eastandwestwind: ah yeah that makes more sense to me

clients/fides-js/src/fides.ts
```diff
@@ -0,0 +1,130 @@
...
+  getOrMakeFidesCookie,
+  makeConsentDefaults,
+} from "./lib/cookie";
+import { ConsentBannerOptions } from "./lib/consent-types";
+import { getBannerOptions } from "./lib/consent-utils";
+
+export interface FidesConfig {
+  consent: ConsentConfig;
+  bannerOptions?: ConsentBannerOptions;
+}
```
NevilleS, line 55: New strong opinion here!

I'd suggest we be super clear and opinionated on this and re-think the config options for this package into:
```
export interface FidesConfig {
  consent: ConsentConfig; // (optional) set the consent defaults from a "legacy" Privacy Center config.json
  experience: ExperienceConfig; // (optional) set the "experience" to be used for this Fides.js instance -- overrides the "legacy" config
  geolocation: UserGeolocation; // (optional) set the geolocation for this Fides.js instance
  options: FidesOptions; // (optional) global options for this Fides.js instance
}
```

I think this'll make it much easier to test, configure, and migrate to if we tinker it this way...

NevilleS (commented):
Couple comments that I was thinking about recently - I think we can architect this in a super clean and functional way to make this easy to incrementally build, but want your $0.02 on it too

NevilleS (commented):
Couple comments that I was thinking about recently - I think we can architect this in a super clean and functional way to make this easy to incrementally build, but want your $0.02 on it too

clients/fides-js/src/lib/consent.tsx
```diff
@@ -0,0 +1,173 @@
...
+      console.error(
+        "Invalid 'extraOptions' arg for Fides.banner(), ignoring",
+        extraOptions
+      );
+    } else {
+      // If the user provides any extra options, override the defaults
+      setBannerOptions({ ...getBannerOptions(), ...extraOptions });
+    }
+  }
+  const options: ConsentBannerOptions = getBannerOptions();
```
NevilleS, line 124: TBH this is pretty sloppy initialization logic. I recognize it as my own so I can say that 😄 

This version of the banner is looking _way_ more maintainable with Preact in place, so we can definitely remove some of the unnecessary footguns here with state management. You'd be able to delete all of this logic if the `ConsentBanner` component itself would just manage merging options with defaults - that's standard React stuff that won't require this back-breaking API 👍

clients/fides-js/src/lib/cookie.ts
```diff
@@ -217,3 +217,31 @@ export const saveFidesCookie = (cookie: FidesCookie) => {
     CODEC
   );
 };
+
+export const setConsentCookieAcceptAll = (defaults: CookieKeyConsent): void => {
```
NevilleS, line 221: nit: I still don't know what a good name for this function is and I think it's a sign 😓 

Maybe `acceptAllAndSaveFidesCookie`?

clients/fides-js/tsconfig.json
```diff
@@ -13,10 +15,11 @@
     "skipLibCheck": true,
     "strict": true,
     "target": "es2015",
-
+    "plugins": [{ "name": "typescript-plugin-css-modules" }],
     "baseUrl": "./",
     "paths": {
-      "~/*": ["src/*"]
+      "react": ["./node_modules/preact/compat/"],
+      "react-dom": ["./node_modules/preact/compat/"]
```
NevilleS, line 22: Curious, what are these paths used for?

NevilleS: After all I think the `preact` module would be up in the `clients/` folder, so `../node_modules/...`?

clients/privacy-center/cypress/e2e/consent-banner.cy.ts
NevilleS: Let's do this as a follow-up ticket though, get this all nice and stable first.

clients/privacy-center/cypress/e2e/consent-banner.cy.ts
```diff
@@ -0,0 +1,181 @@
...
+            labels: {
+                bannerDescription: "This test website is overriding the banner description label.",
+                primaryButton: "Accept Test",
+                secondaryButton: "Reject Test",
+                tertiaryButton: "Customize Test",
+            }
+        }
+    };
+
+    Cypress.Commands.add("visitConsentDemo", (consent?: ConsentConfig, bannerOptions?: ConsentBannerOptions) => {
```
NevilleS, line 52: You should move this to the commands.ts file with all it's friends 👍

clients/privacy-center/cypress/e2e/consent-banner.cy.ts
```diff
@@ -0,0 +1,181 @@
...
+                //       {labels.tertiaryButton}
+                //     </button>
+                //     <button id='fides-consent-banner-secondary-button' class='fides-consent-banner-button fides-consent-banner-secondary-button'>
+                //       {labels.secondaryButton}
+                //     </button>
+                //     <button id='fides-consent-banner-primary-button' class='fides-consent-banner-button fides-consent-banner-primary-button'>
+                //       {labels.primaryButton}
+                //     </button>
+                //   </div>
+                // </div>
```
NevilleS, line 114: This comment feels unnecessary now - it's no longer as fragile (thank you!) so what this test should do is just help us double-check the render behaviour stays consistent as we continue to tinker with the internals.

clients/privacy-center/pages/api/fides-js.ts
```diff
@@ -43,11 +43,18 @@ export default async function handler(
...
+  const fidesCSS: string = JSON.stringify(fidesCSSBuffer.toString());
   if (!fidesJS || fidesJS === "") {
     throw new Error("Unable to load latest fides.js script from server!");
   }
   const script = `
   (function () {
+    // Include default CSS
+    const style = document.createElement('style');
+    style.innerHTML = ${fidesCSS};
+    document.head.append(style);
```
NevilleS, line 56: I think this would be a lot better if it was handled by `fides.js` itself - see my other comments on the topic above in the rollup config 😄

clients/privacy-center/public/lib/fides.css
```diff
@@ -0,0 +1,146 @@
+/**
+* Default CSS styles for the banner
+*/
+:root {
+  --fides-consent-banner-font-family: inherit;
```
NevilleS, line 5: This is a build-time artifact so you should `.gitignore` it

NevilleS (changes requested):
OK, looking good so far!

There are some large structural asks in this review- hopefully we're at a point now where we have a solid test harness for all the logic so we can retool this to fit our liking

clients/fides-js/rollup.config.mjs
```diff
@@ -28,6 +30,14 @@ export default [
         verbose: true,
         hook: "writeBundle",
       }),
+      copy({
+        // Automatically add the built css to the privacy center's static files for bundling:
+        targets: [
+          { src: `dist/${name}.css`, dest: "../privacy-center/public/lib/" },
+        ],
+        verbose: true,
```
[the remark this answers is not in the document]
eastandwestwind: I agree that it should work like this, but I've been wrangling with this for a while, and doesn't seem to work in the way we expect. There must be some sort of tsconfig setting I'm missing. I'll track in a follow-up

clients/fides-js/src/fides.ts
```diff
@@ -83,17 +91,21 @@ const init = (config: FidesConfig) => {
   // Load any existing user preferences from the browser cookie
   const cookie = getOrMakeFidesCookie(defaults);
 
+  await initFidesConsent(defaults, config.bannerOptions);
```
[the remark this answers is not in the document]
eastandwestwind: this initializes any and all components, including overlay or link. It also could init setting GPC flag downstream without ever rendering a visual component. So with that in mind, I wanted to keep this more generic. Any other ideas?

clients/fides-js/src/lib/banner.module.css
```diff
@@ -0,0 +1,146 @@
+/**
+* Default CSS styles for the banner
+*/
+:root {
+  --fides-consent-banner-font-family: inherit;
+  --fides-consent-banner-font-size: 16px;
```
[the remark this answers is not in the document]
eastandwestwind: absolutely, I'll update

clients/fides-js/src/lib/consent-types.ts
```diff
@@ -0,0 +1,40 @@
...
+    primaryButton?: string;
+    secondaryButton?: string;
+    tertiaryButton?: string;
+  };
+
+  // URL for the Privacy Center, used to customize consent preferences. Required.
+  privacyCenterUrl?: string;
+
+  // Text for consent link
+  consentLinkText?: string;
```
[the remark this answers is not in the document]
eastandwestwind: We won't need this text config option

clients/fides-js/src/lib/consent-banner.tsx
```diff
@@ -0,0 +1,122 @@
...
+import {CookieKeyConsent, setConsentCookieAcceptAll, setConsentCookieRejectAll} from "./cookie";
+import {ButtonType, ConsentBannerOptions} from "./consent-types";
+import debugLog from "./consent-utils";
+import ConsentBannerButton from "../components/ConsentBannerButton";
+
+interface BannerProps {
+  defaults: CookieKeyConsent;
+  options: ConsentBannerOptions;
+  waitBeforeShow: number;
+}
```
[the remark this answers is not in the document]
eastandwestwind: this makes a lot more sense to me, thanks!

clients/fides-js/rollup.config.mjs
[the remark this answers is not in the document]
NevilleS: Totally fine. I'd be OK with either:
1. Bundling the CSS as a separate `fides.css` output, and updating our docs so customers import it directly
2. OR, bundling the CSS into the `fides.js` itself with rollup

I don't love this current setup though, where we're inserting the CSS at the last minute in the bundle on the privacy center. That feels very error-prone, so let's create a follow-up ticket and tackle that.

clients/fides-js/src/lib/consent-utils.ts
```diff
@@ -0,0 +1,56 @@
+import { ConsentBannerOptions } from "./consent-types";
+
+/**
+ * Configuration options used for the consent banner. The default values (below)
+ * will be mutated by the banner() function to override with any user-provided
+ * options at runtime.
+ *
+ * This is effectively a global variable, but we provide getter/setter functions
+ * to make it seem safer and only export the getBannerOptions() one outside this
+ * module.
```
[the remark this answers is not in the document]
eastandwestwind: The naming of this `globalBannerOptions` var is misleading because some of these options are more generic- e.g. geolocation, privacy center URL, and are needed at a parent level. Also we will still need a way to inject banner-specific options at runtime, so we can't delete it yet from being a global var. Unless there's a better pattern you see to handle this?

clients/fides-js/src/lib/consent.tsx
eastandwestwind: See https://github.com/ethyca/fides/pull/3191#discussion_r1188869210. My main question is how to manage configs for more global vars like those related to geolocation (happens outside the banner component)

clients/fides-js/src/lib/consent-utils.ts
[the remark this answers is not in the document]
NevilleS: > The naming of this globalBannerOptions var is misleading because some of these options are more generic- e.g. geolocation, privacy center URL, and are needed at a parent level. 

Yeah, that makes sense - if it's not banner-specific, let's just call them... `options`? We could design the options object to be something like:
```
options = {
  overlay: {
    ...
  },
  fidesApiUrl: "...",
  isGeolocationEnabled: true,
  ...
}
```

Having a shared `options` object is definitely reasonable for the overall `Fides.js` package, and instead of being in some clever state in this file it could just be a field on the `window.Fides` global

clients/fides-js/src/lib/consent.tsx
NevilleS: Replied over there, but to double-down: I think we need a `Fides.options` object we can put all these things in 👍

clients/fides-js/src/lib/consent-utils.ts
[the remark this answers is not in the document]
NevilleS: See this comment too: https://github.com/ethyca/fides/pull/3191#discussion_r1187857126

clients/fides-js/rollup.config.mjs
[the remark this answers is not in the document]
eastandwestwind: to be clear, rollup already bundles the CSS into a separate `fides.css` output in `fides-js/dist/fides.css`. I've added the above lines to copy it over into `privacy-center/public/lib/fides.css` so that `privacy-center/public/fides-js-components-demo.html` can import it directly in the HTML. This file is only used by cypress tests.

For our original `privacy-center/public/fides-js-demo.html` file, we dynamically insert the css from `fides-js/dist/fides.css`,  just like we do the js file from `fides-js/dist/fides.js`

clients/fides-js/src/lib/consent.tsx
```diff
@@ -0,0 +1,182 @@
...
+        debugLog("Fides consent link el found");
+        consentLinkEl.onclick = () => {
+          debugLog("Navigate to Privacy Center URL:", options.privacyCenterUrl);
+          // todo- depending on notices / experience config, we update onclick of this link to nav to PC or open modal,
+          //  or hide link entirely
+          if (options.privacyCenterUrl) {
+            window.location.assign(options.privacyCenterUrl);
+          }
+        };
+      } else {
```
allisonking, line 172: something like

```suggestion
      if (
        consentLinkEl &&
        consentLinkEl instanceof HTMLAnchorElement &&
        options.privacyCenterUrl
      ) {
        debugLog("Fides consent link el found");
        consentLinkEl.href = options.privacyCenterUrl;
      } else {
```

allisonking (commented):
this is a really solid foundation, @eastandwestwind ! I'm going to keep playing with things today, but here are some comments for now. looking forward to building off of this!

clients/fides-js/rollup.config.mjs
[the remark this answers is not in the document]
NevilleS: Yup!

I did a bit of research and found that several other rollup plugins exist that support the ability to automatically "inject" styles into the `<head>`. npm tends to be a bit of a popularity contest so finding a good module that is popular and well-maintained tends to be the safest bet.

After poking around a bit I found these three:
* `rollup-plugin-postcss` (700k weekly downloads, but unmaintained - last release Nov 2021)
* `rollup-plugin-styles` (53k weekly downloads, also unmaintained - last release Jan 2022)
* `rollup-plugin-import-css` (21k weekly downloads, maintained - last release Mar 2023) <-- this is the one you're using

While *most* of the articles online push towards the postcss plugin, I think we'd be OK to stick with this `import-css` plugin and just use the `document.adoptedStylesheets` API: https://developer.mozilla.org/en-US/docs/Web/API/Document/adoptedStyleSheets

That should be as simple as changing this to be:
```
import styles from "./styles.css" assert { type: "css" };

document.adoptedStyleSheets = [styles];
```

clients/fides-js/tsconfig.json
```diff
@@ -13,10 +15,11 @@
...
     }
   },
   "include": ["**/*.ts", "**/*.d.ts"],
```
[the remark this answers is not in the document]
NevilleS: I'd recommend pushing commits where you can - there's lots of little cleanups to do and I think it's safe enough to push & pull to @eastandwestwind's branch

clients/fides-js/src/components/ConsentBannerButton.tsx
```diff
@@ -0,0 +1,20 @@
...
+interface ButtonProps {
+    buttonType: ButtonType;
+    label?: string,
+    onClick?: () => void
+}
+
+const ConsentBannerButton: FunctionComponent<ButtonProps> = ({ buttonType, label, onClick }) => (
+    <button
+        id={`fides-consent-banner-button-${buttonType.valueOf()}`}
+        className={`fides-consent-banner-button fides-consent-banner-button-${buttonType.valueOf()}`}
```
allisonking, line 13: at some point we will probably want to make constants out of these

clients/fides-js/src/lib/consent.tsx
allisonking: although I suppose if they made it a button and not a link, then there wouldn't be an `href`. in which case we could still use an `onClick`, but we'd only want that after we made sure it wasn't an `HTMLAnchorElement` first.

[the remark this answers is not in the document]
allisonking: I'll update it to do an `href` swap, although I'm just now thinking through the TODO Catherine left of changing the onClick to open the modal instead. in that case, it might still make sense to use an `onClick`, I'm not sure. ideally it'd be a button in that case though

clients/fides-js/src/lib/cookie.ts
```diff
@@ -217,3 +217,31 @@ export const saveFidesCookie = (cookie: FidesCookie) => {
...
+  if (defaults === undefined) {
+    // eslint-disable-next-line no-console
+    console.error("Unable to set consent cookie to accept all: invalid defaults.");
+    return;
+  }
+
+  // Override all consent values to true and save the cookie
+  const entries: [string, boolean][] = Object.keys(defaults).map((key) => [key, true]);
+  const cookieKeyConsent = Object.fromEntries(entries);
+  const cookie: FidesCookie = getOrMakeFidesCookie(cookieKeyConsent);
```
allisonking, line 231: actually, do we want to use this function? I think this one queries the existing cookie, and then merges it with the passed in `defaults`, but would prioritize the existing cookie. for example, when I open up http://localhost:3000/fides-js-demo.html, I already have a cookie with 

```
{
  "data_sales": true,
  "tracking": false,
  "analytics": false
}
```

and when I click "Accept all", that cookie is still the same since the `cookieKeyConsent` where every value is `true` doesn't actually overwrite anything

eastandwestwind: I haven't thought much about cookie updating, and you're right- this definitely doesn't work right now. We could modify `getOrMakeFidesCookie` so that it works for this use case, I'm open to ideas here

clients/fides-js/rollup.config.mjs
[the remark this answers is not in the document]
eastandwestwind: I did see those other plugins, but opted for mine as it was the most lightweight AND also fairly well maintained.

I didn't know about `document.adoptedStyleSheets` at all, thanks! I'll try this out

clients/fides-js/src/lib/cookie.ts
allisonking: I think we might want to `getOrMakeFidesCookie` and then overwrite its `consent`. I'm not 100% sure about this, but let me try it out and I'll push a commit and you can tell me if it makes sense

allisonking: https://github.com/ethyca/fides/pull/3191/commits/5f649968a3498d3adf52d24fac4a27f636179553

allisonking: you can see this is working since you can "Accept all", then refresh the page and see the `Consent JSON` values update to all be true, then "Reject all" and see them all change to false

allisonking:
oh exciting, the E2E smoke test failed! though we can't tell why from CI because we are over our free limit 😅

eastandwestwind:
I've restructured quite a bit such that we can now inject all dependencies into `Fides.init()`.

A couple things to note:

1. cypress is broken, and I'm unsure why. I'll try to debug more tomorrow. This is the error message I get when I try to `cd clients/privacy-center`, `turbo run dev`, then in another terminal `turbo run cy:run`
```
Error: Webpack Compilation Error
../fides-js/src/lib/banner.module.css 4:0
Module parse failed: Unexpected token (4:0)
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
| * Default CSS styles for the banner
| */
> :root {
|   --fides-consent-overlay-font-family: inherit;
|   --fides-consent-overlay-font-size: 16px;
 @ ../fides-js/src/components/ConsentBanner.tsx 20:0-34
 @ ../fides-js/src/fides.ts
 @ ./cypress/e2e/consent-banner.cy.ts
```

2. Since we're dependency-injecting into `Fides.init()`, the easiest way to test/dev this branch is to navigate to http://localhost:3000/fides-js-components-demo.html. This page calls `Fides.init()` with a default config object already packaged and ready to go. But if you wish to test with other config options, simply modify the config object in `fides-js-components-demo.html`.

@NevilleS / @allisonking let me know your thoughts, possibly how to improve dev workflow, too, with this new config injection pattern.

clients/privacy-center/cypress/e2e/consent-banner.cy.ts
eastandwestwind: ticketed and documented in description of this PR- https://github.com/ethyca/fides/pull/3191#issue-1689319530

clients/fides-js/rollup.config.mjs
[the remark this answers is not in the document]
eastandwestwind: Hm, so this isn't working for me. Here's what I did: 

1. Created a `declaration.d.ts` file with `declare module "*.module.css";` in `fides-js/src`
2. Added the below to `ConsentBanner.tsx`:

```
import styles from "../lib/banner.module.css" assert { type: "css" };

if (typeof window !== 'undefined') {
  // @ts-ignore
  // Create an empty "constructed" stylesheet
    const sheet = new CSSStyleSheet();
  // Apply a rule to the sheet
  sheet.replaceSync(JSON.stringify(styles));
  document.adoptedStyleSheets = [sheet];
}
```

3. Removed the extra css imports from `privacy-center/public/fides-js-components-demo.html` and `privacy-center/pages/api/fides-js.ts`

This did bundle css into `dist/fides.js`, but the actual css styles were not reflected on the demo page
Screenshot 2023-05-10 at 12 18 02 PM

clients/fides-js/src/components/ConsentBannerButton.tsx
eastandwestwind: agree! Let's follow-up with these

clients/fides-js/src/components/ConsentBanner.tsx
```diff
@@ -1,19 +1,32 @@
...
+}
+
 const ConsentBanner: FunctionComponent<BannerProps> = ({
-  options,
+  options = {
+    bannerTitle: "Manage your consent",
+    bannerDescription: "This website processes your data respectfully, so we require your consent to use cookies.",
+    confirmationButtonLabel: "Accept All",
+    rejectButtonLabel: "Reject All"
+  },
```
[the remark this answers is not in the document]
eastandwestwind: ah yes, I'll adjust so that they're optional

clients/fides-js/package.json
```diff
@@ -37,17 +38,28 @@
...
+    "eslint-config-preact": "^1.3.0",
     "eslint-config-prettier": "^8.8.0",
+    "eslint-plugin-import": "^2.27.5",
+    "eslint-plugin-node": "^11.1.0",
+    "eslint-plugin-promise": "^6.1.1",
+    "eslint-plugin-react": "^7.32.2",
     "jest": "^29.5.0",
     "jest-environment-jsdom": "^29.5.0",
     "js-cookie": "^3.0.5",
+    "npm": "^9.6.5",
```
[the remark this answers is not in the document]
eastandwestwind: honestly don't remember at this point but it's not needed.

allisonking:
> I've restructured quite a bit such that we can now inject all dependencies into Fides.init().
>
> A couple things to note:
> 
> cypress is broken, and I'm unsure why. I'll try to debug more tomorrow. This is the error message I get when I try to cd clients/privacy-center, turbo run dev, then in another terminal turbo run cy:run

hmm that's a little mysterious, I guess it's because cypress doesn't know how bundle css modules, but it's also odd that it's trying to bundle the module. I'll take a look and see if there's a workaround.

clients/fides-js/tsconfig.json
eastandwestwind: this was probably leftover from testing, and can be removed.

clients/fides-js/rollup.config.mjs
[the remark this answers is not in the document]
eastandwestwind: tracking follow-up here- https://github.com/ethyca/fides/issues/3288

clients/fides-js/src/lib/consent.tsx
```diff
@@ -0,0 +1,188 @@
...
+          bannerTitle={config.experience?.banner_title}
+          bannerDescription={config.experience?.banner_description}
+          confirmationButtonLabel={config.experience?.confirmation_button_label}
+          rejectButtonLabel={config.experience?.reject_button_label}
+          privacyCenterUrl={config.options.privacyCenterUrl}
+          onAcceptAll={onAcceptAll}
+          onRejectAll={onRejectAll}
+          waitBeforeShow={100}
+        />,
+        document.body
```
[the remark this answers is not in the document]
allisonking: I experimented with doing something like that earlier to see if it would help with the hydration issue. it didn't help there, but I agree it's probably safer to do. I'll try to cook something up!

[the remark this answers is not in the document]
allisonking: so from what I can tell, doing something like that may conflict with [the hydration issue](https://www.joshwcomeau.com/react/the-perils-of-rehydration/). the init script would look something like

```typescript
// Add root div for Fides components to live in
render(<div id="fides-js-root" />, document.body);

// Append to Fides root
const fidesRoot = document.getElementById(FIDES_ROOT_ID);
if (fidesRoot) {
  render(<ConsentBanner ... />, fidesRoot);
}
```

However, that triggers the hydration warning in cookie house since the script injects `fides-js-root` which the server does not expect. so then we add our workaround and make a `<FidesRoot />` component which only renders the component after the initial mount

```typescript
const FidesRoot = () => {
    if (!hasMounted) {
      return null;
    }
  return <div id="fides-js-root"></div>;
};

// Add root div for Fides components to live in
render(<FidesRoot />, document.body);

// Append to Fides root
const fidesRoot = document.getElementById(FIDES_ROOT_ID);
if (fidesRoot) {
  // fidesRoot isn't ready yet because it hasn't been mounted yet, so the banner never loads!! 🙀 
  render(<ConsentBanner ... />, fidesRoot);
}
```

I think our choices would be
* ask the customer put an empty `fides-js-root` div somewhere in their app which we append to
* do something clever to check if `FidesRoot` is ready yet before trying to append the `ConsentBanner` to it
* ask the customer to implement the hydration workaround if they are using SSR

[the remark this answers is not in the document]
allisonking: another option would be to do something like

```typescript
render(<FidesRoot><ConsentBanner ... /></FidesRoot>, document.body);
```

but that's kind of similar to the original implementation

clients/fides-js/rollup.config.mjs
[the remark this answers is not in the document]
NevilleS: 👍 FWIW @eastandwestwind this'll be a good amount of trial-and-error, as most build system jiggery pokery tends to be, so I do think it's best to do in a singular ticket & PR.

clients/fides-js/src/fides.ts
```diff
@@ -71,28 +85,45 @@ declare global {
...
+  const consentDefaults = makeConsentDefaults({
     config: config.consent,
     context,
   });
 
   // Load any existing user preferences from the browser cookie
-  const cookie = getOrMakeFidesCookie(defaults);
+  const cookie = getOrMakeFidesCookie(consentDefaults);
+
+  await initFidesConsent(config);
```
NevilleS, line 102: nit: How about `initOverlay`?

clients/fides-js/src/lib/consent-utils.ts
```diff
@@ -0,0 +1,14 @@
...
+ * option is truthy
+ */
+type ConsoleLogParameters = Parameters<typeof console.log>;
+export const debugLog = (
+  enabled: boolean,
+  ...args: ConsoleLogParameters
+): void => {
+  if (enabled) {
+    // eslint-disable-next-line no-console
+    console.log(...args); // TODO: use console.debug instead?
```
NevilleS, line 12: TODO here- what do y'all think?

NevilleS (changes requested):
Couple more nits and a bug or two but this is super close!

NevilleS (changes requested):
Couple more nits and a bug or two but this is super close!

clients/fides-js/src/fides.ts
allisonking: wow, I'm only just realizing that we may have been interpreting `overlay` differently this whole time! I think your definition is what `fides-js` overlays onto a customer's website, while I had always been thinking of it as specifically the modal (because there are ModalOverlays). a lot of things are suddenly making more sense!

clients/fides-js/src/lib/consent.tsx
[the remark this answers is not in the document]
allisonking: I'll make a follow-up issue!

clients/fides-js/package.json
```diff
@@ -37,7 +38,12 @@
     "eslint": "^8.36.0",
     "eslint-config-airbnb": "^19.0.4",
     "eslint-config-airbnb-typescript": "^17.0.0",
+    "eslint-config-preact": "^1.3.0",
     "eslint-config-prettier": "^8.8.0",
+    "eslint-plugin-import": "^2.27.5",
+    "eslint-plugin-node": "^11.1.0",
+    "eslint-plugin-promise": "^6.1.1",
```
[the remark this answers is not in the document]
eastandwestwind: ah nope, good catch!

clients/fides-js/src/lib/consent.tsx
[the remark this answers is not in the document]
allisonking: https://github.com/ethyca/fides/issues/3293

clients/fides-js/src/components/ConsentBanner.tsx
```diff
@@ -0,0 +1,112 @@
...
+    if (privacyCenterUrl) {
+      window.location.assign(privacyCenterUrl);
+    }
+  };
+
+  if (!hasMounted) {
+    return null;
+  }
+
+  // TODO: support option to specify top/bottom
```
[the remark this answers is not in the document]
eastandwestwind: We currently don't have an experience config option for this (that I know of), so I've created this ticket to account for the config + banner component support- https://github.com/ethyca/fides/issues/3294

clients/fides-js/src/lib/consent.tsx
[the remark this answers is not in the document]
eastandwestwind: ty! Also added to our follow-up ticket list in this PR diescription

clients/privacy-center/README.md
```diff
@@ -99,6 +99,17 @@ To run the interactive test interface, run:
 turbo run test
 ```
 
+For a fully-loaded development & test setup of both the Privacy Center and the
+Fides Consent library, run the following commands in three separate terminals:
+
+[//]: # "fixme- turbo doesn't work right now in privacy center"
```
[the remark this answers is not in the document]
eastandwestwind: ah this was leftover, turbo works fully now. I'm also updating this README.md with more details on how to test `fides-js`

clients/fides-js/src/fides.ts
```diff
@@ -19,7 +19,16 @@
  *         default: true,
  *         fidesDataUseKey: "advertising"
  *       }]
- *     }
+ *     },
+ *     experience: {},
+ *     geolocation: {},
+ *     options: {
+ *           debug: true,
+ *           isDisabled: false,
```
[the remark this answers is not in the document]
eastandwestwind: yes, the experience API returns `isDisabled`, but I'll just map that to `isOverlayDisabled` in `fides-js`

clients/fides-js/src/lib/consent-utils.ts
eastandwestwind: I vote stick with `console.log`. We already don't log anything unless `options.debug: true` is passed in to `Fides.init()`. 

In addition, `console.debug` is not enabled by default in browsers, which raises barriers (OK, slightly 😉 ) for ease of dev/testing

clients/fides-js/src/lib/consent.tsx
```diff
@@ -0,0 +1,188 @@
...
+  }
+};
+
+/**
+ * Initialize the Fides Consent Banner or Link, with optional extraOptions to override defaults.
+ *
+ * (see the type definition of ConsentBannerOptions for what options are available)
+ */
+export const initFidesConsent = async (config: FidesConfig): Promise<void> => {
+  debugLog(config.options.debug, "Initializing Fides consent...");
```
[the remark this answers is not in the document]
eastandwestwind: ah awesome, I've cherry-picked this commit into this PR. Thanks!

clients/fides-js/src/lib/consent.tsx
```diff
@@ -0,0 +1,188 @@
...
+
+  if (config.options.isDisabled) {
+    debugLog(
+      config.options.debug,
+      "Fides consent banner is disabled, skipping banner initialization!"
+    );
+    return Promise.resolve();
+  }
+
+  document.addEventListener("DOMContentLoaded", () => {
```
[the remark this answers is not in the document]
eastandwestwind: We have to ensure the DOM is loaded, otherwise we'll see a preact error when we try to `render(<App config={config} />, document.body);`. I'll instead add a condition to test if DOM is loaded

clients/fides-js/src/components/ConsentBanner.tsx
```diff
@@ -26,19 +27,25 @@ const ConsentBanner: FunctionComponent<BannerProps> = ({
   waitBeforeShow,
 }) => {
   const [isShown, setIsShown] = useState(false);
+  const hasMounted = useHasMounted();
```
allisonking, line 30: ah right I wanted to move this `hasMounted` logic to `App.tsx` instead and then we should only need to do it once in `App.tsx`. I'll do that in https://github.com/ethyca/fides/pull/3291

clients/fides-js/src/lib/consent.tsx
```diff
@@ -0,0 +1,209 @@
...
+  debug: boolean,
+  geoLocation: UserGeolocation
+): string | null => {
+  debugLog(debug, "validating getLocation...");
+  if (geoLocation.location) {
+    return geoLocation.location;
+  }
+  if (geoLocation.country && geoLocation.region) {
+    return `${geoLocation.country}-${geoLocation.region}`;
+  }
```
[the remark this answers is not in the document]
NevilleS: Fixed

clients/fides-js/src/lib/consent.tsx
```diff
@@ -0,0 +1,209 @@
...
+
+/**
+ * Initialize the Fides Consent overlay components.
+ *
+ * Includes fetching location and experience, if not provided, setting up banner and modal,
+ * showing/hiding links, and showing/hiding the banner.
+ *
+ * (see the type definition of ConsentBannerOptions for what options are available)
+ */
+export const initOverlay = async (config: FidesConfig): Promise<void> => {
```
NevilleS: Similar to the comment on the `App` component, passing through the whole config object feels like a code smell - we're basically just passing around a huge global variable when we do that, instead of defining sharp boundaries. It also shows up in this function: you end up doing a hundred little `config.options`, `config.experience`, `config.geolocation` calls.

I'm pushing a lil' refactor of this now to show you what I mean 👍

clients/privacy-center/.gitignore
```diff
@@ -1,5 +1,5 @@
 # A copy of fides.js is vendored into public/lib/ at build-time
-public/lib/fides.js
+public/
```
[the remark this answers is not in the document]
NevilleS: Fixed

NevilleS (approved):
OK, ship it!

I still have a bunch of nitpicks but I pushed a few tidy up commits. Let's see if CI still passes!

clients/fides-js/src/components/App.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { FidesConfig } from "../lib/consent-types";
+import ConsentBanner from "./ConsentBanner";
+import { getConsentContext } from "../lib/consent-context";
+import {
+  makeConsentDefaults,
+  setConsentCookieAcceptAll,
+  setConsentCookieRejectAll,
+} from "../lib/cookie";
+
+const App = ({ config }: { config: FidesConfig }) => {
```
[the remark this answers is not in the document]
NevilleS: Done - check it out

clients/fides-js/src/components/ConsentBanner.tsx
```diff
@@ -0,0 +1,100 @@
...
+interface BannerProps {
+  bannerTitle?: string;
+  bannerDescription?: string;
+  confirmationButtonLabel?: string;
+  rejectButtonLabel?: string;
+  privacyCenterUrl: string;
+  onAcceptAll: () => void;
+  onRejectAll: () => void;
+  waitBeforeShow: number;
+}
```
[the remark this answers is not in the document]
NevilleS: Done!

clients/fides-js/src/lib/consent.tsx
NevilleS: Done, take a look...

clients/fides-js/src/lib/cookie.ts
```diff
@@ -217,3 +217,36 @@ export const saveFidesCookie = (cookie: FidesCookie) => {
     CODEC
   );
 };
+
+/**
+ * Given a CookieKeyConsent and a `value`, update and set the fides cookie
+ * such that every key is set to the passed in `value`
+ */
+const setConsentCookieValuesTo = (
```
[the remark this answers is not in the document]
NevilleS: Fixed 👍

clients/fides-js/src/lib/consent.tsx
```diff
@@ -55,6 +52,9 @@ const constructLocation = (
   if (geoLocation.country && geoLocation.region) {
     return `${geoLocation.country}-${geoLocation.region}`;
   }
+  if (geoLocation.country) {
```
eastandwestwind, line 55: I don't think the experience API currently supports filtering when only country is provided. Is this something we need to add on the backend? If so, this would lead to a case where multiple experiences could be returned from the API

eastandwestwind: Unless we plan on adding just country to this enum? https://github.com/ethyca/fides/blob/main/src/fides/api/ops/models/privacy_notice.py#L20

clients/fides-js/src/lib/consent.tsx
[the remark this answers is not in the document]
eastandwestwind: ah, question about this one here- https://github.com/ethyca/fides/pull/3191#discussion_r1193020874

clients/privacy-center/pages/api/fides-js.ts
```diff
@@ -51,11 +58,20 @@ export default async function handler(
   );
   const fidesJSBuffer = await fsPromises.readFile("public/lib/fides.js");
   const fidesJS: string = fidesJSBuffer.toString();
+  const fidesCSSBuffer = await fsPromises.readFile(
+    "../fides-js/dist/fides.css"
+  );
```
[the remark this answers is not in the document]
eastandwestwind: yes! good catch

clients/fides-js/src/lib/consent.tsx
NevilleS: Ah- we'll need to handle that case then. In reality we can't geolocate *that* accurately every time... can you open a ticket and let's check in with @pattisdr and team?

clients/fides-js/src/lib/consent.tsx
eastandwestwind: thank you! This is a lot cleaner

clients/fides-js/src/lib/consent.tsx
pattisdr: It's just currently filtering on the regions in that list - but stepping back for a second, what do we do if the united states is what is filtered on. How do we decide which experience to show then?

NevilleS: We'll have to have some fallbacks & defaults defined for when we don't get exact matches... I did have a good chat about some edge cases like this with @mfbrown yesterday. We need a way to guarantee we get a defined experience no matter what location is determined (including "unknown"), which'll require some updates 👍

clients/fides-js/src/lib/consent.tsx
NevilleS: ✨

clients/fides-js/src/lib/consent.tsx
eastandwestwind: Ticket here- https://github.com/ethyca/fides/issues/3300

clients/fides-js/src/components/ConsentBanner.tsx
```diff
@@ -0,0 +1,100 @@
...
+      setIsShown(true);
+    }, waitBeforeShow);
+    return () => clearTimeout(delayBanner);
+  }, [setIsShown, waitBeforeShow]);
+
+  const navigateToPrivacyCenter = (): void => {
+    if (privacyCenterUrl) {
+      window.location.assign(privacyCenterUrl);
+    }
+  };
```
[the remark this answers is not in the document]
eastandwestwind: it's an `onClick` because lower down, we're using a `<button>` element

Issue #2976, opened 2023-04-03
Merge "forked" PrivacyDeclaration components


allisonking:
### Is your feature request related to a specific problem?

We have copied/pasted a good number of components between the datamap-ui and admin-ui. We should merge these back together now that they are in the same app.

### Describe the solution you'd like

Specifically, we'll need to look at:

* cypress configs
* dirty form hook (perhaps separate ticket to actually use it in admin-ui)
* ConfirmationModal
* Header
* Image
* Layout
* ProtectedRoute
* SearchBar
* toast.tsx
* PrivacyDeclarations

### Describe alternatives you've considered, if any

_A description of any alternative solutions or features you've considered._

### Additional context

Depends on https://github.com/ethyca/fides/issues/2975

allisonking:
Most of this except PrivacyDeclarations is done. That'll take more time to comb through the differences and resolve them

Roger-Ethyca:
moving to done

PR #3197, opened 2023-05-02
Custom fields in data use declaration form

### LINKED ISSUE #3177: Add custom fields to privacy declaration form

TheAndrewJackson:
Closes #3177 

### Code Changes

* [ ] Make the custom field location readonly when editing a custom field
* [ ] Make name and data use readonly when editing a privacy declaration
  * [ ] This was done in the datamap PD form as well but in a very hardcoded way. This is intentional because the separate forms are about to be merged together in #2976 
* [ ] Add custom fields to the privacy declaration form (in the portal only)

### Steps to Confirm

* [ ] Run fidesplus (`nox -s "dev(slim)"`) backend and admin ui
* [ ] Create some custom fields that have the `system:data use` location and create a system
* [ ] Create a a few new privacy declarations with data in the custom fields. Make sure that the new PD maintains it's custom field data after the save
* [ ] Edit all of the PDs and make sure the edited custom field data persists as expected
* [ ] Make sure the name and data use fields should be read only when editing

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### Description Of Changes

https://user-images.githubusercontent.com/17103888/235666532-e4a76a2c-9a04-44c2-9402-2bdd86d6410a.mov


clients/admin-ui/src/features/system/privacy-declarations/PrivacyDeclarationForm.tsx
```diff
@@ -264,6 +329,7 @@ export const PrivacyDeclarationForm = ({
       onSubmit,
       initialValues: passedInInitialValues,
       allDataUses: dataProps.allDataUses,
+      privacyDeclarationId: passedInInitialValues?.id,
```
allisonking, line 332: it seems like the hook should be able to grab the id based on `passedInInitialValues` since it's already getting that as a child (rather than having to pass in another prop with the id separately

clients/admin-ui/src/types/api/models/SystemResponse.ts
```diff
@@ -0,0 +1,112 @@
...
+   */
+  egress?: Array<DataFlow>;
+  /**
+   * The resources from which the System receives data.
+   */
+  ingress?: Array<DataFlow>;
+  /**
+   * Extension of base pydantic model to include DB `id` field in the response
+   */
+  privacy_declarations: Array<PrivacyDeclarationResponse>;
```
allisonking, line 74: oh okay, the typing problem I [mentioned in the ts-ignore](https://github.com/ethyca/fides/blob/3054d348f7ddad93d4a953cbf83da77c12c768f7/clients/admin-ui/src/features/system/privacy-declarations/PrivacyDeclarationForm.tsx#L64-L66) might actually be that we should be using `SystemResponse`, not `System`. `System` has `privacy_declarations` listed without the id, while `SystemResponse`  has it with the `id`. so then the question is... what is `System`?? 😅 

in any case, that's too large of a change for this PR 👍

clients/admin-ui/src/features/common/custom-fields/hooks.ts
```diff
@@ -136,7 +136,11 @@ export const useCustomFields = ({
 
       // When creating an resource, the fides key may have initially been blank. But by the time the
       // form is submitted it must not be blank (not undefined, not an empty string).
-      const fidesKey = formValues.fides_key || resourceFidesKey;
+      const fidesKey =
+        "fides_key" in formValues && formValues.fides_key !== ""
+          ? formValues.fides_key
+          : resourceFidesKey;
+
```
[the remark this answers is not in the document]
TheAndrewJackson: My understanding is custom fields can be added to multiple fides resources (ex: systems, PDs, etc). In the past fides resources were a 1:1 with the fideslang models which used `fides_key` as the id. The new privacy declaration table is a departure from that. It's not a 1:1 match with fideslang so it uses an id instead. 

I added a bit of code that placed the new PD id as a `fides_key` so it would be backwards compatible with the `useCustomField` hook. The reason is the custom fields attached to systems use the `fides_key` as the id.

PR #3211, opened 2023-05-03
Hotfix - merge cached datamap settings with new data

### LINKED ISSUE #3212: Fix Privacy Center "config" Redux state to not fallback to unsafe defaults, reset on error, etc.
### Bug Description

In https://github.com/ethyca/fides/pull/3105 we refactored the Privacy Center Redux state to load the contents of the `config.json` file into the global store, instead of importing it from disk (e.g. `import { config } from "config.json"` becomes `import { useConfig } from "~/features/common/config.slice"`)

There are a few rough edges here that I worry about, but can't 100% confirm will be true - so this is a somewhat vague bug ticket, but I think these bugs exist.

Here's what we need to investigate and/or fix:
- Remove the reliance on a "default" config state that is known, statically, at build time. This creates a footgun where it becomes technically possible for an unexpected error to cause the site to revert to confusing defaults (see `dangerouslyStaticDefaultConfig defined in clients/privacy-center/constants/index.ts)
- Add a global error state to handle unexpected issues. This'll provide a straightforward fallback for when the environment fails to load successfully
- Store all the "server-defined" state into the Redux store: environment settings, config (config.json), and styles (config.css)
- Load all the "server-defined" state in a Redux store in the App component, prior to even providing it via the Redux provider. This means that all normal components can assume the environment is loaded into the store and use it, without replicating a ton of "if not loaded, render some error state" logic everywhere
- Extract a basic Layout component from the current _app.tsx, just so that we can use a hook to fetch the styles from the state (right now it reads the server-side prop directly)
- Add some tests for failing config loading to ensure things are handled

#### Steps to Reproduce

Force the server-side config loading to fail by providing a bad config file.

### Expected behavior

Overall, it should be impossible to *accidentally* show the default configuration at runtime, unless you specifically choose *not* to change the defaults.

TheAndrewJackson:
Closes #3212 

### Code Changes

* [ ] Create new `mergeColumns` function that handles combining the cached columns with the new columns

### Steps to Confirm

* [ ] Visit the datamap page so the column settings get cached
* [ ] Go create a new custom field
* [ ] create a system and populate the custom field with data
* [ ] visit the datamap table and open the settings modal. The new custom field should be in the listed columns. NOTE: it won't show up if no system is using the column
* [ ] Go back and edit the name of the custom field
* [ ] Visit the datmamp table again and check the settings modal. The column should be in the modal with the new name. It will most likely be at the bottom. That is expected.

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

### DIFF
--- a/clients/admin-ui/src/features/datamap/Datamap.tsx
+++ b/clients/admin-ui/src/features/datamap/Datamap.tsx
@@ -1,6 +1,8 @@
 import { Box, Center, Flex, Spinner } from "@fidesui/react";
+import { baseApi } from "common/api.slice";
 import dynamic from "next/dynamic";
-import { useCallback, useContext, useState } from "react";
+import { useCallback, useContext, useEffect, useState } from "react";
+import { useDispatch } from "react-redux";
 
 import { useAppSelector } from "~/app/hooks";
 import { useIsAnyFormDirty } from "~/features/common/hooks/useIsAnyFormDirty";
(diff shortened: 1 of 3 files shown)


adamsachs:
this is looking really good to me in some manual testing locally! things get updated seamlessly as far as i can tell.

i happened to notice a BE regression in doing testing for this, i'll file/follow up with that separately.

adamsachs (approved):

Issue #3119, opened 2023-04-20
Placeholder - better handle cascading for deletes of taxonomy elements


adamsachs:
### Is your feature request related to a specific problem?

A placeholder follow up -- we agreed that we need better general handling of how we handle cascading the effects of deleting/updating taxonomy elements to their associated artifacts in the application.

This is per conversations in https://github.com/ethyca/fides/pull/3098 and other features.

### Describe the solution you'd like

Per @mfbrown in slack:
> IMO, ideally you would choose to delete from the taxonomy, get a UI warning telling you which systems (and other resources) would be affected, then able to confirm or cancel. If you confirm then the cascading delete would occur

very broadly needs:
- likely some combination of DB-level and API-level checks/handling on the BE
- FE support!

### Additional context

I'm sure this has come up in relation to other feature work, please others fill this in!

PR #3214, opened 2023-05-03
handle deleted taxonomy elements in datamap export

adamsachs:
‼️ This should be hotfixed for `2.12.1`

Closes https://github.com/ethyca/fidesplus/issues/858

### Code Changes

* treat defunct `data_use` or `data_subject` references from a privacy (data use) declaration as empty rows in the datamap export
* ensure at least one "placeholder" system row is returned if no rows are produced from declarations

### Steps to Confirm

- requires integration with fidesplus to properly test (will be a corresponding PR there with increased test coverage)

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### Description Of Changes

A broader improvement around taxonomy element deletion is covered in https://github.com/ethyca/fides/issues/3119.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,10 @@ The types of changes are:
 
 - Merge instances of RTK `createApi` into one instance for better cache invalidation [#3059](https://github.com/ethyca/fides/pull/3059)
 
+### Fixed
+
+- Datamap export mitigation for deleted taxonomy elements referenced by declarations [#3214](https://github.com/ethyca/fides/pull/3214)
+
 ## [2.12.0](https://github.com/ethyca/fides/compare/2.11.0...2.12.0)
 
 ### Added
--- a/src/fides/core/export.py
+++ b/src/fides/core/export.py
@@ -310,6 +310,11 @@ def generate_system_records(  # pylint: disable=too-many-nested-blocks, too-many
                 system.get("data_protection_impact_assessment", {})
             )
         )
+
+        # variable to keep track of rows added based on system declarations
+        # used to determine whether we need to add a "placeholder" row for the system generally
+        declaration_rows = []
+
         if system.get("privacy_declarations"):
             for declaration in system["privacy_declarations"]:
                 if isinstance(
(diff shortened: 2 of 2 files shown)


adamsachs:
@SteveDMurphy or @ThomasLaPiana, my datamap friends...would you mind taking a look at this quick bugfix to handle an edge case that we can hit since our handling of taxonomy element deletion is still a bit immature throughout the application?

@pattisdr FYI, since you found the bug :) if you want to dig into the datamap export code, please be my guest and i would be very happy to get your review! @Kelsey-Ethyca also want to loop you in for release logistics

CHANGELOG.md
```diff
@@ -25,6 +25,12 @@ The types of changes are:
 
 - Merge instances of RTK `createApi` into one instance for better cache invalidation [#3059](https://github.com/ethyca/fides/pull/3059)
 
+## [2.12.1](https://github.com/ethyca/fides/compare/2.12.0...2.12.1)
```
adamsachs, line 28: heads up on the changelog here -- i _think_ this should be right, but let's confirm this is how we want to do this before we merge

src/fides/core/export.py
```diff
@@ -363,6 +370,7 @@ def generate_system_records(  # pylint: disable=too-many-nested-blocks, too-many
                     ]
                     for category in data_categories
```
SteveDMurphy, line 371: @adamsachs do we know if this is limited to these taxonomy types? Deletes have never been something we handled fully so am curious if the same should apply for data category, etc.

adamsachs: a very good question! i thought about this, but the reason we don't need to run any checks here is because there's no actual "lookup" of the data category object, the string value(s) referenced by the declaration are just taken at face value

i suppose we _could_ just run a query against the db to see if it exists and skip the row if it doesn't. but it's not like we're actually using anything in the data category, and we're sort of punting on more "proper" handling of taxonomy deletion generally, so my inclination here is just to leave it be, since we won't error.

SteveDMurphy: fair point, thanks! Ya let's leave it and see if we can prioritize some of the work we have been punting on to stop this cycle we keep finding ourselves in 🌀

src/fides/core/export.py
```diff
@@ -392,8 +400,12 @@ def generate_system_records(  # pylint: disable=too-many-nested-blocks, too-many
                             )
                     cartesian_product_of_declaration.append(tuple(product))
 
-                output_list += cartesian_product_of_declaration
-        else:
+                declaration_rows += cartesian_product_of_declaration
+            output_list += declaration_rows
+
+        # if we haven't added any output rows for the system based on declarations,
+        # then add a "placholder" generla system row to ensure the system is captured in the output
```
SteveDMurphy, line 407:
```suggestion
        # then add a "placeholder" general system row to ensure the system is captured in the output
```

SteveDMurphy (approved):
Let's do this thing 👍🏽 

It might be helpful to alter the comments for the (hopefully impending) refactor here that these are only for deleted items instead of invalid (the resources would fail for just a typo or something so we might as well be prescriptive as to the why)

thanks for tackling more of these 🙌🏽

CHANGELOG.md
pattisdr: Do we need to adjust the unreleased line too then?

adamsachs: i decided to just put this under `Unreleased` and we can make the changelog updates on the the hotfix branch. i'm not sure what i was thinking originally, trying to be clever but it felt misguided on a second look. thanks for the 👀 !

src/fides/core/export.py
adamsachs: 👍 fixed (rebased to keep commit history nice and simple)

PR #3215, opened 2023-05-03
update custom field definition uniqueness criteria

adamsachs:
Partially closes https://github.com/ethyca/fidesplus/issues/826

### Code Changes

* create the proper unique (functional) index for the uniqueness criteria
    * include a migration that "handles" (i.e. reports with a cleaner error message) potentially invalid data on migration
    * migration also has some small unrelated cleanup on `plus_custom_field_definition` table since we're touching it
* put in some more graceful handling of the unique constraint in the sqlalchemy model class
* add in some test coverage directly on the sqlalchemy model

### Steps to Confirm

* any manual testing here would require hitting the DB directly
* manual testing at the API level will come in corresponding `fidesplus` [PR](https://github.com/ethyca/fidesplus/pull/862) that has the API which exposes this model

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### Description Of Changes

We'll now ensure custom field definition uniqueness with a case insensitive name check, per `resource_type` (location). This implements that constraint on the DB level and also has some handling of the error to give a "cleaner" exception for consumption higher up in the stack.

### DIFF
--- a/src/fides/api/ctl/database/session.py
+++ b/src/fides/api/ctl/database/session.py
@@ -2,9 +2,10 @@ from typing import AsyncGenerator
 
 from sqlalchemy import create_engine
 from sqlalchemy.ext.asyncio import AsyncSession, create_async_engine
-from sqlalchemy.orm import Session, sessionmaker
+from sqlalchemy.orm import sessionmaker
 
 from fides.core.config import CONFIG
+from fides.lib.db.session import ExtendedSession
 
 engine = create_async_engine(
     CONFIG.database.async_database_uri,
@@ -15,7 +16,7 @@ async_session = sessionmaker(engine, class_=AsyncSession, expire_on_commit=False
 sync_engine = create_engine(CONFIG.database.sync_database_uri, echo=False)
 sync_session = sessionmaker(
     sync_engine,
-    class_=Session,
+    class_=ExtendedSession,
     expire_on_commit=False,
     autocommit=False,
 )
--- a/src/fides/api/ctl/migrations/versions/15a3e7483249_make_customfielddefinition_uniqueness_.py
+++ b/src/fides/api/ctl/migrations/versions/15a3e7483249_make_customfielddefinition_uniqueness_.py
@@ -1,0 +1,53 @@
+"""make customfielddefinition uniqueness case insensitive
+
+Revision ID: 15a3e7483249
+Revises: e92da354691e
+Create Date: 2023-05-02 15:03:56.256982
+
+"""
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.exc import IntegrityError
+
+# revision identifiers, used by Alembic.
+revision = "15a3e7483249"
+down_revision = "e92da354691e"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+
+    try:
+        op.execute(
+            """ CREATE UNIQUE INDEX ix_plus_custom_field_definition_unique_lowername_resourcetype
+                ON plus_custom_field_definition
+                (lower(name), resource_type)
+            """
+        )
+    except IntegrityError as exc:
+        raise Exception(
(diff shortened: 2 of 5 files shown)


src/fides/api/ctl/database/session.py
```diff
@@ -15,7 +16,7 @@
 sync_engine = create_engine(CONFIG.database.sync_database_uri, echo=False)
 sync_session = sessionmaker(
     sync_engine,
-    class_=Session,
+    class_=ExtendedSession,
```
adamsachs, line 19: this helped my tests run, and it _seemed_ like a good idea regardless since our `ExtendedSession` just provides some nice error handling around the base ORM session - so why not use it with our `ctl` session? 

i think we have plans to consolidate our `session` objects anyway across `ops` and `ctl` code paths, so this is just a step in that direction. all tests pass so hopefully we are good?

src/fides/api/ctl/migrations/versions/15a3e7483249_make_customfielddefinition_uniqueness_.py
```diff
@@ -0,0 +1,53 @@
...
+            """
+        )
+    except IntegrityError as exc:
+        raise Exception(
+            f"Fides attempted to create new custom field definition unique index but got error: {exc}. "
+            f"Adjust custom field names to avoid case-insensitive name overlaps for the same resource type."
+        )
+
+    # remove unnecessary index of unused field
+    op.drop_index(
```
adamsachs, line 35: this is an unused field/column altogether. removing the field/column/attribute wholesale felt out of scope for this PR, but i figured we could at least get rid of the index here as we were updating indexes on this table.

src/fides/api/ctl/sql_models.py
```diff
@@ -514,7 +517,50 @@ class CustomFieldDefinition(Base):
...
+                raise KeyOrNameAlreadyExists(
+                    "Custom field definitions must have unique names for a given resource type"
+                )
+            raise e
+
+    # unique index on the lowername/resource type for case-insensitive name checking per resource type
+    name_resource_index = (
+        "ix_plus_custom_field_definition_unique_lowername_resourcetype"
+    )
+    __table_args__ = (
```
adamsachs, line 556: is there a cleaner way of doing this? it's my first time using "functional" indexes, and i didn't see many (any?) other examples in our codebase (certainly could have missed them though!). this reads relatively clearly to me, but i get a little bit nervous any time i reference a dunder variable like this.

src/fides/api/ctl/sql_models.py
```diff
@@ -514,7 +517,50 @@ class CustomFieldDefinition(Base):
...
+            return super().update(db=db, data=data)
+        except IntegrityError as e:
+            if CustomFieldDefinition.name_resource_index in str(e):
+                raise KeyOrNameAlreadyExists(
+                    "Custom field definitions must have unique names for a given resource type"
+                )
+            raise e
+
+    # unique index on the lowername/resource type for case-insensitive name checking per resource type
+    name_resource_index = (
```
adamsachs, line 553: just storing the string as a constant variable here to allow it to be referenced when looking for the index name in error messages above.

adamsachs:
@pattisdr this is primarily a migration/db model update, so i'd appreciate a quick review when you have a chance! 

(it's not a major change, and also not blocking any of my work currently).

src/fides/api/ctl/sql_models.py
pattisdr: something similar here: https://github.com/ethyca/fides/blob/main/src/fides/api/ops/models/storage.py#L53

looks okay to me

pattisdr (approved):
looks good

adamsachs:
thanks for the quick review @pattisdr! just merged in for changelog conflict resolution, will let CI run to completion and then merge into main

PR #3218, opened 2023-05-04
Fix Deploy Version Check

### LINKED ISSUE #3217: Python/Docker version checks are breaking
### Bug Description

The `fides deploy` version checks are breaking which is preventing `fides deploy up` from running

#### Steps to Reproduce

1. _Do this..._
1. _Then This..._

### Expected behavior

The command to still be attempted even if the version check fails

### Additional context

[Link to slack thread](https://fidescommunity.slack.com/archives/C03L22ZPYTV/p1683130896524299)

ThomasLaPiana:
Closes #3217 

[docker version docs](https://docs.docker.com/engine/reference/commandline/version/)

### Code Changes

* [x] give the user the option to attempt a `deploy` even if the Docker check fails (hedges against any potential version check logic misses)
* [x] update the version check to use a specific format for the `docker version` command

### Steps to Confirm

* [ ] `pip install` locally (testing done via CLI)
* [ ] bump the `REQUIRED_DOCKER_VERSION` in `fides.deploy.core` to something very high (force failure)
* [ ] run `fides deploy up`
  * [ ] confirm that you can see the `ERROR:` message in red from the Docker exception
  * [ ] confirm it gives you a prompt
  * [ ] Test both the `y` and `n` answers to the prompt 

### Pre-Merge Checklist

* [x] Update `CHANGELOG.md`

### Description Of Changes

This is a fix for an issue a user was having. The existing logic was very fragile and made lots of assumptions about the format of the docker version output. This method should be highly deterministic, and in the case that there is still an issue, there is now an escape-hatch to attempt the deploy anyway.

#### Screenshots

Error when Docker version is too low:
image

Error when Docker check fails hard:
image


ThomasLaPiana:
@Kelsey-Ethyca FYI we'll want this in a patch/hotfix release if possible, a user is hitting this issue

ThomasLaPiana:
setting ready for review but I'm going to add to the test coverage here

src/fides/core/exceptions.py
```diff
@@ -0,0 +1,10 @@
+from rich_click import secho
+
+
+class DockerCheckException(Exception):
+    """Raise an error when `local_mode` is enabled with an incompatible command."""
```
adamsachs, line 5: nit - looks like a copy pasta miss

src/fides/cli/commands/util.py
```diff
@@ -146,9 +146,18 @@ def up(
     """
 
     check_virtualenv()
-    check_docker_version()
+    try:
+        check_docker_version()
+    except:  # pylint: disable=bare-except
+        response = click.confirm(
```
adamsachs, line 152: so this will actually now catch the exception thrown in the case where the version is found, but determined to be too low, right? i'm not against allowing the escape hatch in that scenario, feels like why not - but is this message going to be a bit confusing to make it seem like there was an error rather than just "correctly" determining the version was too low?

granted, i haven't tested this myself so i'm not exactly sure what the output/DX is, it may be perfectly easy to understand! i just wanted to pose the question

adamsachs (approved):
this looks good to me! just a minor nit and question. thanks for getting a fix so quick :)

src/fides/core/exceptions.py
ThomasLaPiana: I've been found out 😂  good catch, thank you!

src/fides/cli/commands/util.py
ThomasLaPiana: I posted an example of the differences up above. This will preserve showing the proper Error message if there is a version mismatch, but if _any_ exception is thrown will still give the ability to forge ahead (YOLO)

ThomasLaPiana:
@adamsachs thanks for the quick review!

src/fides/cli/commands/util.py
adamsachs: that looks good to me 👍 thank you for the screenshots/confirmation!

Issue #3126, opened 2023-04-21
Individual experience form page


allisonking:
### Is your feature request related to a specific problem?

We need a form for editing a privacy experience

### Describe the solution you'd like

![image](https://user-images.githubusercontent.com/24641006/233688273-2a8af6d7-2677-4bb9-b1b8-1a6c41b6ac70.png)

Notices are not linked to experiences in the backend, so we may need to separately query for the notice based on the experience. For example:

Clicking on a privacy experience row of "California" / "Overlay" means we should query for notices with `regions=[us_ca]` and `displayed_in_overlay=true`

* For all forms
  * Location section (no editing in v1)

* If the experience has any "opt in" notices
  * Banner text section: title + description
  * Banner actions section: link label, confirmation button, reject button
  * Copy for overlay

@allisonking to continue filling this out as questions are answered. currently some confusion around what happens when there are multiple privacy notices "linked" to an experience but they have different consent mechanisms

### Describe alternatives you've considered, if any

_A description of any alternative solutions or features you've considered._

Roger-Ethyca:
moving to done

PR #3219, opened 2023-05-04
Privacy experience form

allisonking:
Closes https://github.com/ethyca/fides/issues/3126

### Code Changes

* [x] Add new `PrivacyExperienceForm.tsx`
  * [x] This one has a few subcomponents which control the various form rendering logic from privacy experiences
  * [x] Other than that, it's pretty similar to `PrivacyNoticeForm.tsx`
* [x] Cypress tests, including a handy function to easily test different experiences backed by varying notices

### Steps to Confirm

* Same setup as described in https://github.com/ethyca/fides/pull/3186
* Click into one of the experiences
* You should be able to fill out the form and persist the data 

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

https://www.loom.com/share/556bc39d3f6b4f7ca50a4c2cef14929a

https://user-images.githubusercontent.com/24641006/236256514-8be4a0c7-e771-4719-bb93-9ce28ef5571b.mov

### DIFF
--- a/clients/admin-ui/cypress/e2e/privacy-experiences.cy.ts
+++ b/clients/admin-ui/cypress/e2e/privacy-experiences.cy.ts
@@ -144,4 +144,160 @@ describe("Privacy experiences", () => {
       });
     });
   });
+
+  describe("form", () => {
+    beforeEach(() => {
+      cy.intercept("PATCH", "/api/v1/privacy-experience*", {
+        fixture: "privacy-experiences/list.json",
+      }).as("patchExperiences");
+    });
+    interface Props {
+      mechanisms: ("opt_in" | "opt_out" | "notice_only")[];
+      component?: "overlay" | "privacy_center";
+    }
+    /**
+     * Helper function to swap out notices and components in a stubbed experience
+     * @example stubExperience({mechanisms: ["notice_only", "opt_in"], component: "overlay"})
+     */
+    const stubExperience = ({ mechanisms, component }: Props) => {
+      const notices = [];
+      mechanisms.forEach((mechanism) => {
+        cy.fixture(`privacy-notices/${mechanism}.json`).then((notice) =>
+          notices.push(notice)
+        );
+      });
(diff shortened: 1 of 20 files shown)


TheAndrewJackson (approved):
I played around with all the example experiences the original PR had and was able to edit everything as expected. I like the implementation for dynamically showing different parts of the form.


I think it was a good choice to not attempt a dynamic validation schema like I did with the custom fields 😅 At least not for the first pass. I wish that formik supported that pattern better.


:shipit:

allisonking:
> I played around with all the example experiences the original PR had and was able to edit everything as expected. I like the implementation for dynamically showing different parts of the form.
> 
> I think it was a good choice to not attempt a dynamic validation schema like I did with the custom fields 😅 At least not for the first pass. I wish that formik supported that pattern better.
> 
> :shipit:

haha yeah, I thought I could get away with no ValidationSchema, but it turns out we probably do want some validation https://github.com/ethyca/fides/issues/3226 so I might have to restructure how this works

allisonking:
I refactored the form logic a bit in order to accommodate required fields (somewhat dynamic validation schemas—not to the same extent you had to do @TheAndrewJackson in custom fields, since they don't change on user input, just depending on what data is received). I added a hook `useExperienceForm` to encapsulate the various rule logic, so that all rules can hang out in one place, which hopefully makes adding/editing/removing rules easier

TheAndrewJackson (approved):
Solid refactor. The new hook cleans the logic up a lot. It's helpful that the validation logic doesn't need to change based on a drop down option! It's still dynamic but not as dynamic as the custom fields form which is nice.

allisonking:
I had a suspicion my hook was being called more than I wanted it to be and it turned out to be true! so I refactored one more time 😅 this time rules are passed around as props, but I think it works pretty well since the same set of rules get passed everywhere

PR #3220, opened 2023-05-04
Add Manual Migration Feature

### LINKED ISSUE #3093: Move migrations to a CLI command/API endpoint instead of happening automatically

As the title states, while sometimes convenient it is also potentially very dangerous or inconvenient to _always_ run migrations on server start.

We should make this functionality configurable, and add a `fides db migrate` CLI command + admin endpoint

#### New `fides.toml` option

```toml
[database]
automigration = false # The server will not run initialization steps and will instead require the user to run 'fides db migrate'
```

ThomasLaPiana:
Closes #3093 

### Code Changes

* [x] add a `database` setting for `automigrate`
* [x] rename `init` to `migrate` when referring to the database migration steps, since that feels more accurate and less surprising to me
* [x] add a `deprecation` label on `fides db init` and point people to `fides db migrate`
* [x] add a `migrate` CLI command
* [x] check for the `automigrate` config value when doing migrations
* [x] do some code reorganization to cleanup `main.py` and move some of the verbose setup to `app_setup.py`

### Steps to Confirm

* [x] run `nox -s teardown -- volumes` to clear the database
* [x] add `automigrate = false` to the `fides.toml` in the `[database]` section
* [x] run `nox -s dev` and see that it doesn't run the database migrations and subsequently fails startup
* [x] remove the line in `fides.toml` that disables auto-migrations
* [x] run `nox -s dev` again and see that everything works (it will also log an `INFO` statement that migrations are being run)
* [ ] spin everything down and re-add `automigrate = false`
* [ ] run `nox -s dev` again and see that it doesn't run migrations, as evidenced by a log `INFO` stating that migrations are being skipped and lack of migration logging

### Pre-Merge Checklist

* Documentation:
  * [x] Handled by autogeneration
* [x] Update `CHANGELOG.md`

### Description Of Changes

Add the ability for users to turn off auto-migrations. Also rename `init` to `migrate` everywhere since it makes more sense (in my opinion) and is used for more than just initializing.

### Breaking Changes

The `CLI` won't see any breaks here, as the `init` command has been deprecated instead of removed.

However, the `db/<action>` endpoint has been broken, as it no longer accepts `init`. This is reflected in the autogenerated documentation though since it uses a proper `Enum` for the actions.


src/fides/api/app_setup.py
```diff
@@ -0,0 +1,201 @@
...
+async def run_database_startup() -> None:
+    """
+    Perform all relevant database startup activities/configuration for the
+    application webserver.
+    """
+
+    if not CONFIG.database.sync_database_uri:
+        raise FidesError("No database uri provided")
+
+    if CONFIG.database.automigrate:
```
ThomasLaPiana, line 150: this is where the actual new check is

ThomasLaPiana:
@pattisdr This is ready for your review/opinions! Sorry it took so long for me to circle back and put the finishing touches on it

Honestly I'm still iffy on the implementation/UX, given the caveat that it _needs_ to be `true` on the first startup of a webserver instance. I think given that this will primarily be for development purposes it should be ok though 🙂

src/fides/api/main.py
```diff
@@ -380,6 +204,28 @@ def read_other_paths(request: Request) -> Response:
...
+    await send_analytics_event(
+        AnalyticsEvent(
+            docker=in_docker_container(),
+            event=Event.server_start.value,
+            event_created_at=datetime.now(tz=timezone.utc),
+        )
+    )
+
+    logger.info(FIDES_ASCII_ART)
+    logger.info(f"Fides startup complete! v{VERSION}")
```
pattisdr, line 226: Thanks for splitting some of this out -

pattisdr (commented):
this is a nice addition, I think it's fine if it needs to be true on startup.  The main thing I thought would be helpful here is for when you're trying to figure out a custom migration and to keep it from auto-migrating on you all the time.

src/fides/api/app_setup.py
```diff
@@ -0,0 +1,203 @@
...
+    try:
+        db = get_api_session()
+        ApplicationConfig.update_config_set(db, CONFIG)
+    except Exception as e:
+        logger.error("Error occurred writing config settings to database: {}", str(e))
+        raise FidesError(
+            f"Error occurred writing config settings to database: {str(e)}"
+        )
+    finally:
+        db.close()
```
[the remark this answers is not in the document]
ThomasLaPiana: yeah I saw that too...I guess it's safer to use a context manager here

src/fides/api/app_setup.py
```diff
@@ -0,0 +1,203 @@
...
+    try:
+        get_cache()
+    except (RedisConnectionError, RedisError, ResponseError) as e:
+        logger.error("Connection to cache failed: {}", str(e))
+        return
+    else:
+        logger.debug("Connection to cache succeeded")
+
+    if not scheduler.running:
+        scheduler.start()
```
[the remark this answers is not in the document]
ThomasLaPiana: Thank you for the tip! Moved

ThomasLaPiana:
@pattisdr Thank you for the great catches here!

Addressed the comments, ready for another review 🙂

src/fides/api/app_setup.py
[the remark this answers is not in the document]
ThomasLaPiana: So actually that other function expects a session, not a session transaction, so the context manager can't work as expected. Instead I've gotten the db session directly (if that is broken, the whole app is) and close it as needed in `finally` statements or at the end if everything succeeds

src/fides/api/main.py
```diff
@@ -380,6 +192,31 @@ def read_other_paths(request: Request) -> Response:
...
+    await send_analytics_event(
+        AnalyticsEvent(
+            docker=in_docker_container(),
+            event=Event.server_start.value,
+            event_created_at=datetime.now(tz=timezone.utc),
+        )
+    )
+
+    if not scheduler.running:
+        scheduler.start()
```
pattisdr, line 214: I think this still needs to run in the same order as before, at least before the `initiate_scheduled_batch_email_send()`

ThomasLaPiana: this should be documented then, at least in code comments 🙂 

Would you mind adding code comments detailing what needs to run in what order/what depends on what? Or if you describe it here I can translate it to code comments! If it was easy for me to trip over this not realizing there is a specific order here, we should write it down for future reference

pattisdr: I wouldn't recommend reordering anything casually  🙂   

Almost everything we write has a general intentional sequence. Here, I assume the scheduler needs to be started before a task (the email batch send) can be added to it.  I haven't verified that this breaks, but you should.

 Feel free to add code comments where you'd like, but in my opinion, order matters here no more than anywhere else.

ThomasLaPiana: everything ran for me just fine, but i guess because I kept the order despite rearranging? (tests all passed too) so I think I'll do some further clean up and document clearly that startup order matters and then ship it, if that sounds good to you?

pattisdr: It doesn't look like you "kept the order" - unless you have unpushed changes?  The `initiate_scheduled_batch_email_send` is happening before the scheduler is started here. 

 Even if it doesn't error, my unverified concern would be that that batch email send task wasn't actually registered.  The email send is a task that gets scheduled but then only fires once a week -  I don't know if we have a unit test that would specifically break if the scheduling was broken. 

```python
@app.on_event("startup")
async def setup_server() -> None:
    "Run all of the required setup steps for the webserver."

    log_startup()
    await run_database_startup()
    check_redis()
    initiate_scheduled_batch_email_send() ***

    logger.debug("Sending startup analytics events...")
    await send_analytics_event(
        AnalyticsEvent(
            docker=in_docker_container(),
            event=Event.server_start.value,
            event_created_at=datetime.now(tz=timezone.utc),
        )
    )

    if not scheduler.running: ***
        scheduler.start()
```

ThomasLaPiana: Ah, I see your point. I will update properly to match the current order in `main`

src/fides/api/main.py
```diff
@@ -380,6 +192,38 @@ def read_other_paths(request: Request) -> Response:
...
+    log_startup()
+
+    await run_database_startup()
+
+    check_redis()
+
+    if not scheduler.running:
+        scheduler.start()
+
+    initiate_scheduled_batch_email_send()
```
ThomasLaPiana, line 212: @pattisdr I fixed the order here, but I don't remember if we allow the server to run without a scheduler? If so, I can add another conditional check here and raise an error if the server isn't running when the email batch gets added.

Likewise we can just leave it as-is 🙂

src/fides/api/ops/service/privacy_request/email_batch_service.py
```diff
@@ -130,6 +130,8 @@ def initiate_scheduled_batch_email_send() -> None:
     if CONFIG.test_mode:
         return
 
+    assert scheduler.running, "Scheduler is not running! Cannot add Batch Email job."
```
ThomasLaPiana, line 133: @pattisdr added a simple check here, this feels like the right amount of protection to me

ThomasLaPiana:
@pattisdr tossing back to you! 

Lmk if there are any big issues here, otherwise I'd prefer to get this merged and iterate

pattisdr (approved):
thanks @ThomasLaPiana

PR #3224, opened 2023-05-04
use regex in docker version check to specifically target semantic ver…

adamsachs:
Closes n/a

Hopefully should patch up e2e failures on `main`, while also keeping the desired functionality in the version check...

Drafted this in a pinch, very open to suggestions!

### Code Changes

* handle some uglier docker version outputs by targeting a semver number directly with a regex

### Steps to Confirm

* works locally with at "clean" docker version (very basic regression test)
* fixes e2e test in CI runner which seemed to have an _uglier_ docker version

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### DIFF
--- a/src/fides/core/deploy.py
+++ b/src/fides/core/deploy.py
@@ -1,3 +1,4 @@
+import re
 import sys
 import webbrowser
 from functools import partial
@@ -89,9 +90,19 @@ def check_docker_version(
             raise DockerCheckException(
                 "Could not determine Docker version from 'docker' commands. Please ensure that Docker is installed and running and try again."
             )
-        parsed_version = (
-            raw_version.stdout.decode("utf-8").rstrip("\n").replace("'", "")
+        stripped_version = (
+            raw_version.stdout.decode("utf-8").rstrip("\n").replace("'", "").strip()
         )
+        version_match = re.match(r"((\d+\.)+\d+)", stripped_version)
+        if (
+            not version_match
+        ):  # if this is empty, we could not find a version # matching the regex
+            raise DockerCheckException(
+                f"Could not parse your Docker version (v{stripped_version})!"
+            )
+        parsed_version = version_match.group(
+            0
+        )  # extract the version # from the regex match
 
         split_docker_version = convert_semver_to_list(parsed_version)
         split_required_docker_version = convert_semver_to_list(required_docker_version)


src/fides/core/deploy.py
```diff
@@ -89,9 +90,17 @@ def check_docker_version(
             raise DockerCheckException(
                 "Could not determine Docker version from 'docker' commands. Please ensure that Docker is installed and running and try again."
             )
-        parsed_version = (
+        stripped_version = (
             raw_version.stdout.decode("utf-8").rstrip("\n").replace("'", "")
         )
+        parsed_version = re.match(r"((\d+\.)+\d+)", stripped_version)
```
adamsachs, line 96: used `match` instead of `search` because i think we can assume the version number is the beginning of the string, especially after stripping it above???

NevilleS: Eh... if there's just a single preceding space ` ` you'll be left with nothing!

The `rstrip("\n")` seems to just strip trailing newlines, not touch whitespace at all.

TL;DR: I would chain one more `.strip()` to the parsing

NevilleS (commented):
Feels a bit brittle still...

src/fides/core/deploy.py
adamsachs: ooh, right didn't pay close enough attention to what we were stripping. i'll do as you suggest, that feels reasonable.

NevilleS (approved):
LGTM!

ThomasLaPiana:
@adamsachs for my own curiosity here, what was the output that broke this? I was thinking the formatting I used would've been quite clear/robust, but apparently not 🤔

adamsachs:
@ThomasLaPiana what @NevilleS spotted this in the [nox output from that CI job](https://github.com/ethyca/fides/actions/runs/4885622167/jobs/8719901123#step:5:9):

```
nox > Running session fides_env(test)
nox > Creating virtual environment (virtualenv) using python3 in .nox/fides_env-test
Docker version 20.10.24+azure-1, build 297e1284d3bd092e9bc96076c3ddc4bb33f8c7ab
```
so we assumed something along the likes of `20.10.24+azure-1` was still getting returned to your docker version command in the CI runtime. i suppose we could have (or still could) confirm more precisely by adding in some additional logging 🤔

ThomasLaPiana:
Oh that makes sense 🤔 I feel like this could be cleaned up further still but honestly the value isn't there right now, we can revisit later 🙂

ThomasLaPiana:
Thanks for the rescue there!

PR #3245, opened 2023-05-08
Account for Snowflake Case Sensitivity

### LINKED ISSUE #1838: fides generate dataset should retain object name case
### Bug Description

`fides generate dataset` generates dataset for snowflake db in all lowercase while all object names in our snowflake database are upper case. On the other hand, `fidesops` generates SQL statement in case sensitive manner. (i.e. Object names are double-quoted like `SELECT "accountId", "email" from "Your_table_NAME"`. Therefore, when we try to use dataset generated by fides CLI, it fails to find objects and throws error like:

```bash
[2022-11-21 21:03:02,382: ERROR/ForkPoolWorker-2] (snowflake.connector.errors.ProgrammingError) 002003 (42S02): 01a87709-0502-33b6-0003-778331f012ba: SQL compilation error:
Object '"sm_accountsnew__dbo__sm_useraccount__stage"' does not exist or not authorized.
```

#### Steps to Reproduce

1. perform `fides generate dataset` against the snowflake with table / column names are defined in upper case.
1. check the generated dataset file and all object names are lower cased regardless of the original case in snowflake.

### Expected behavior

It should keep the original case.

### Screenshots

n.a.

### Environment

* Version: fidesctl-plus 1.1.6

(We use the docker image given by Ethyca.)

### Additional context

n.a.

### LINKED ISSUE #2091: Consistent handling of Snowflake object quoting

Currently, the handling of quoted object names in Snowflake-related code is inconsistent. `fides generate` will produce lower-cased table names, while DSR processing double quotes all table names and therefore will end up throwing `relation not found` errors. This will become a bug once we fully unify the dataset resources.

Refactor how we handling quoting objects in all snowflake-related code (or maybe all database connectors?) to be consistent across the application

SteveDMurphy:
Closes #2091
Closes #1838

### Code Changes

* [x] Create a dialect specific function to get Snowflake schemas and maintain casing

### Steps to Confirm

* [ ] Using the POCDB (found in 1Password, search by PR#) generate a dataset
* [ ] Set up a default policy using the postman collection
* [ ] Set up the snowflake connector to process a DSR (ensure the database and schema name are part of the connection config)
* [ ] Generate an access request
* [ ] Validate it succeeds

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### Description Of Changes

Some further issues still exist here but this should be enough to allow us some forward motion. A design doc will discuss how to handle the need for a fully qualified table name for cases where the default database and schema are not applicable(or provided).

I'm not entirely sure if we want to update our docs to specify some of the connection as we will be planning to make some rather immediate changes there as well. Will open an issue if required 👍🏽


SteveDMurphy:
I think the patch coverage might be wrong here but just a bit hidden behind the parameterized tests?

src/fides/core/dataset.py
```diff
@@ -30,7 +31,7 @@
     "postgresql": ["information_schema"],
     "mysql": ["mysql", "performance_schema", "sys", "information_schema"],
     "mssql": ["INFORMATION_SCHEMA", "guest", "sys"],
-    "snowflake": ["information_schema"],
+    "snowflake": ["INFORMATION_SCHEMA"],
```
seanpreston, line 34: Is there any method behind the case inconsistency of these different datastores you're aware of?

src/fides/core/dataset.py
```diff
@@ -373,3 +377,33 @@ def generate_dynamo_db_datasets(aws_config: Optional[AWSConfig]) -> Dataset:
...
     dynamo_dataset = create_dynamodb_dataset(described_dynamo_tables)
     return dynamo_dataset
+
+
+def get_snowflake_schemas(
+    engine: Engine,
+) -> Dict[str, Dict[str, List[str]]]:
+    """
+    Returns Datasets that match the case-sensitivity that may be
+    required by Snowflake.
```
adamsachs, line 387: nit - may be helpful to provide a bit more explanation here? specifically - what about this differs from the "standard"  `inspector.get_schema_names()`, etc. calls being made on the other codepath? or is there any documentation/reference that explains why we need to do this? 

granted i'm looking at this code with not a lot of context/experience with `sqlalchemy.inspect` so maybe that's why it's just a bit hard for me to understand what exactly we're doing differently here to ensure we get case-sensitive metadata. i'm just imagining coming back to this code in some months/years and looking for a bit more context 😄

src/fides/core/dataset.py
```diff
@@ -373,3 +377,33 @@ def generate_dynamo_db_datasets(aws_config: Optional[AWSConfig]) -> Dataset:
...
+    """
+    schema_cursor = engine.execute(
+        text("SHOW /* sqlalchemy:get_schema_names */ SCHEMAS")
```
adamsachs, line 390: sort of piling on the comment above -- the comment block (i assume that the `/*` indicates a comment!) just confuses me a bit. is there a reason it comes _within_ the `SHOW SCHEMAS` command?

again, this could very well just be my lack of experience speaking here 😳

adamsachs (approved):
i think this looks good! i'm not sure i've got all the context on the broader issue we were facing, but these changes look like a straightforward and targeted improvement.

my only comments are just around whether we could increase some of the code comments to make this a _little_ bit more self explanatory for someone coming back to the code without context -- but feel free to ignore that nit, as it may be more a result of my lack of experience in this area.

src/fides/core/dataset.py
SteveDMurphy: Snowflake just happens to default to all caps, it's a common gotcha when people move to Snowflake and are used to using something like `CamelCase`. Some users will initially try and double-quote columns to maintain casing across data stores because it feels a bit funky that any non-quoted item defaults to all caps. MSSQL is case sensitive by default and just happens to implement the `INFORMATION_SCHEMA` in that case. It is sort of a window into the wild west that can be different databases and their implementation of SQL :)

src/fides/core/dataset.py
SteveDMurphy: No this makes total sense! A common first impression of people using Snowflake is some general surprise to their defaulting of casing to all caps. What makes this extra interesting is the officially supported sqlalchemy dialect then has the `inspect()` method default everything to lower case and the issues to address this are more than three years stale at this point. 

I'll add this context in the docs block of the code so that we can have it closer to the issue, let me know if that helps cover that a bit more and thanks for the suggestion 👍🏽

src/fides/core/dataset.py
SteveDMurphy: No I thought about removing the comments but ended up liking the paper trail. This is primarily my wanting to keep it as common as possible with the existing `sqlalchemy` connector. Ideally, I'd like to open a PR against `snowflake-sqlalchemy` to resolve it and I liked the idea of the breadcrumbs there but it does feel a bit out of place looking at it again

src/fides/core/dataset.py
adamsachs: yup that sounds great!

src/fides/core/dataset.py
adamsachs: oh i think the paper trail makes sense, i was _thinking_ it was something like that. maybe just an explanatory (python-level 😄 ) code comment to clarify that? or does that feel too verbose? you could throw it in wherever you end up adding the additional context discussed above maybe...

anyway, i get the motivation with the paper trail, so not suggesting you remove it! just wanted to double check on this and see if we can make it a little bit more explicit for someone re-looking at this down the road.

SteveDMurphy: I just ended up removing the code comments inline and adding a lot more into the docs block. It felt like there is a chance those extra comments could get stale and not be as meaningful anyways so hopefully that should be enough. Also, best next move is likely sorting out the actual connector with a PR so we can step away from this change here 🤞🏽

PR #3255, opened 2023-05-09
Update nox -s tag docstring

ThomasLaPiana:
Closes #<issue>

### Code Changes

### Steps to Confirm

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

### DIFF
--- a/noxfiles/git_nox.py
+++ b/noxfiles/git_nox.py
@@ -77,6 +77,8 @@ def tag(session: nox.Session) -> str:
     - (default) = don't actually apply the tag or push, just show the tag that will be generated
     - only_tag = generate and apply the a tag locally to the current commit, but don't push the tag
     - push = generate, apply and push a tag
+    
+    Add examples
     """
     from git.repo import Repo
 


ThomasLaPiana:
@adamsachs I opened this to simply update the docstring and ended up taking things a bit further...

To give some background, I made the switch from `posargs` to `params` because in my opinion, the uses generally match the following patterns:

- `posargs` = Good for situations where you might want _multiple_ to be applied at the same time, or want to pass in some kind of arbitrary value
- `params` = Good for situations where you have a clear pre-defined notion of what each option would do, and you wouldn't want/need them to overlap in functionality

Let me know if that makes sense to you or you have strong feelings about keeping them as posargs!

adamsachs (approved):
this makes a lot of sense, thank you for taking the time and thought to update! the choice to go with posargs was definitely a product of me just getting my bearings with nox. a param makes a lot more sense to me here!

things looked good when i ran locally, i also liked the improved usage docs and the additional cleanup. thanks again for the nice improvement!

because we don't share nox code across fides/fidesplus (not saying we should), we should also adjust the same command on the `fidesplus` side. i can open up a quick PR to do that. it's pretty much just c+p.

ThomasLaPiana:
> this makes a lot of sense, thank you for taking the time and thought to update! the choice to go with posargs was definitely a product of me just getting my bearings with nox. a param makes a lot more sense to me here!
> 
> things looked good when i ran locally, i also liked the improved usage docs and the additional cleanup. thanks again for the nice improvement!
> 
> because we don't share nox code across fides/fidesplus (not saying we should), we should also adjust the same command on the `fidesplus` side. i can open up a quick PR to do that. it's pretty much just c+p.

Huge thanks for the "mirror" update on plus! I so wish we had a way to share `nox` updates across but alas it truly isn't feasible or probably worth it hah

PR #3272, opened 2023-05-10
Add the ability to "inject" location into `/fides.js` bundles and cache responses for one hour

### LINKED ISSUE #3170: Optimize Privacy Center `/fides.js` download with cache headers and a local `public/assets/fides.js` file
### Description
In https://github.com/ethyca/fides/pull/3105 we refactored the Privacy Center to serve a custom bundle of the `fides.js` Javascript package, initialized with the tenant-specific configuration sourced from the Privacy Center backend. This works swimmingly, but needs some optimization before it's ready for production use.

### Acceptance Criteria
- [ ] Add reasonable cache headers to the `/api/fides-js` route to ensure both browsers and CDNs will cache the result
- [ ] During the `privacy-center` build, copy the generic `fides.js` package to `/public/assets/fides.js`
- [ ] Test that this "virtual" file is properly cached by browsers and CDNs

NevilleS:
Closes #3170, plus unticketed work to support injecting location 👍 

### Code Changes

* [X] Add `cache-control` headers to the `/fides.js` bundle to instruct browsers/CDNs to cache for ~1 hour
* [X] Add `/fides.js?location=` query param to inject a given location to the bundle
* [X] Add support for CloudFront location headers (e.g. `CloudFront-Viewer-Country`) to inject a given location to the bundle
* [X] Include the injected location in the `ETag` for caching `/fides.js` bundle
* [X] Add lots of tests

### Steps to Confirm

* [X] Run all Cypress tests
* [x] Run `nox -s "fides_env(test)"`, load the cookie house sample app, and confirm that `/fides.js` bundle is cached across page refreshes

Loom of some manual cache testing: https://www.loom.com/share/e840916e3c964b00b8d428a07492c9b4

### Pre-Merge Checklist

* Documentation:
* [x] Relevant Follow-Up Issues Created:
  * [X] Create issue to add API documentation: https://github.com/ethyca/fides/issues/3275
* [x] Update `CHANGELOG.md`

### Description Of Changes

This adds an (important!) optimization to the `/fides.js` route that will enable HTTP clients/CDNs/etc. to cache the results, so we don't unnecessarily hammer the Privacy Center with requests. This part was fairly easy to do, but I realized while doing this that we should bake in support for generating *different* bundles for different locations as early as possible- since those cache semantics are kinda hard to reason about...

So this PR *also* proactively adds the ability to request a `fides.js` bundle with an explicit location, e.g.: https://localhost:3000/fides.js?location=US-CA

☝️ This will parse the query param out, and add a `location` property to the config object passed into the bundled initialization of Fides.js, ie. `window.Fides.init({ consent, location })` 

This'll should enable us to set up test harnesses to exercise different locations without resorting to a lot of VPN chicanery 😄


NevilleS:
I definitely didn't realize I had some mixing of fides.js builds from local branches! I think I fixed that now, but CI is still mysteriously cancelling/failing... hmm.

clients/privacy-center/__tests__/common/location.test.ts
```diff
@@ -0,0 +1,108 @@
...
+
+import { getLocation } from "~/common/location";
+
+describe("getLocation", () => {
+  describe("when using location headers", () => {
+    it("returns location data from country & region headers", () => {
+      const req = createRequest({
+        url: "https://privacy.example.com/fides.js",
+        headers: {
+          "CloudFront-Viewer-Country": "US",
```
allisonking, line 11: just so I understand, would this only happen if the privacy center were running through cloudfront?

clients/privacy-center/common/location.ts
```diff
@@ -0,0 +1,75 @@
...
+];
+
+/**
+ * Lookup the "location" (ie country and region) for the given request by looking for either:
+ * 1) Supported geolocation headers (e.g. "Cloudfront-Viewer-Country: US")
+ * 2) An explicit "location" query param (e.g. https://privacy.example.com/some/path?location=US-CA)
+ *
+ * If neither of these are found, return an undefined location.
+ *
+ * NOTE: This specifically *does not* include performing a geo-IP lookup... yet!
```
allisonking, line 35: when we do query for geo IP, what priority should that get? i.e. would we only look for that if there isn't a query param?

clients/privacy-center/common/location.ts
```diff
@@ -0,0 +1,75 @@
...
+const CLOUDFRONT_HEADER_REGION = "cloudfront-viewer-country-region";
+export const LOCATION_HEADERS = [
+  CLOUDFRONT_HEADER_COUNTRY,
+  CLOUDFRONT_HEADER_REGION,
+];
+
+/**
+ * Lookup the "location" (ie country and region) for the given request by looking for either:
+ * 1) Supported geolocation headers (e.g. "Cloudfront-Viewer-Country: US")
+ * 2) An explicit "location" query param (e.g. https://privacy.example.com/some/path?location=US-CA)
```
allisonking, line 31: nit: I feel like these should be swapped in order since we prioritize the location param over the geolocation header

allisonking (approved):
this is great, and thanks for all of the tests! 🎉 

I left a few questions, but those are just to make sure I'm thinking about this the right way 👍

clients/privacy-center/__tests__/common/location.test.ts
NevilleS: Exactly, yup! I think you'd also need to configure a cache policy that forwards those headers, etc.

But what that means is that, by routing through the CloudFront CDN, we'd instantly have an accurate picture of the user's location before the bundle is even built. That allows us to server location-specific bundles via this API 👍 

We use CloudFront for Fides Cloud so we'd be able to try this out ourselves

clients/privacy-center/common/location.ts
NevilleS: Yeah, I'd ladder that in as the lowest priority. Anything explicit (query param / headers) should take precedence.

clients/privacy-center/common/location.ts
NevilleS: Good call, fixed!

PR #3290, opened 2023-05-11
Prevent sort reset on data change

### LINKED ISSUE #3246: Cache `FidesTable` column sorts
# Description
Currently, when a page using `FidesTable` has the table data get reloaded it will remove any sorts the user has applied. This is potentially due to the table instance being recreated. It would be ideal to cache the users column sorts so if the data gets reloaded (ex: disabling a custom field) the sorts remain.

TheAndrewJackson:
Closes #3246 

### Code Changes

* [ ] Update react table config to not automatically reset sorts when data updates
* [ ] Add test for this update

### Steps to Confirm

* [ ] Run fidesplus and the admin ui
* [ ] Add a few custom fields
* [ ] Apply a column sort
* [ ] enable/disable a field
* [ ] ensure the sort is saved

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### Description Of Changes
I thought I was going to need to cache the sort objects but it turns out the desired behavior is a config option 🤷‍♂️ 

https://react-table-v7.tanstack.com/docs/faq#how-do-i-stop-my-table-state-from-automatically-resetting-when-my-data-changes

https://github.com/ethyca/fides/assets/17103888/36d160c3-058a-43f5-aae3-39a17fdf0213


clients/admin-ui/src/features/common/table/FidesTable.tsx
```diff
@@ -53,7 +53,10 @@ export const FidesTable = <T extends FidesObject>({
     return [useGlobalFilter, useSortBy];
   }, [customHooks]);
 
-  const tableInstance = useTable({ columns, data }, ...plugins);
+  const tableInstance = useTable(
+    { columns, data, autoResetSortBy: false },
```
allisonking, line 57: wow, they make it so easy!

clients/admin-ui/cypress/e2e/custom-fields.cy.ts
```diff
@@ -117,19 +117,61 @@ describe("Custom Fields", () => {
...
+      beforeEach(() => {
+        cy.intercept(
+          "PUT",
+          "/api/v1/plus/custom-metadata/custom-field-definition*",
+          {
+            fixture: "custom-fields/list.json",
+          }
+        ).as("patchCustomFields");
+      });
+      it("should be table to sort", () => {
```
allisonking, line 130: did you mean

```suggestion
      it("should be able to sort", () => {
```

clients/admin-ui/cypress/e2e/custom-fields.cy.ts
```diff
@@ -117,19 +117,61 @@ describe("Custom Fields", () => {
...
+          .first()
+          .should("contain", "Taxonomy - Single select");
+      });
 
-      // sort reverse
-      cy.getByTestId("column-Label").click();
-      cy.get("tbody > tr")
-        .first()
-        .should("contain", "Taxonomy - Single select");
+      it("should maintain sort after custom field is enabled/disabled", () => {
```
allisonking, line 145: thank you for adding this test!! I was playing around with it and it seems like this test passes even without the change in config in `FidesTable`. the only reason I suspected it might is because it's kind of difficult to get cypress to respond with _updated_ data.

I think what you may have to do is intercept `@getCustomFields` again, after the initial visit, and change its data such that "Taxonomy - Single select" will return as enabled. it's always kind of annoying to do, but something like

```typescript
cy.fixture("custom-fields/list.json").then(customFieldsList => {
  // iterate through customFieldsList and change the value of enabled/disabled for the entry you care about
  const updatedFieldsList = ...
  cy.intercept(
    "GET",
    "/api/v1/plus/custom-metadata/custom-field-definition*",
    {
      body: updatedFieldsList,
    }
  ).as("getCustomFieldsSingleSelectEnabled");
});

// sort, then enable the custom field, and wait for `@patchCustomFields`
cy.wait("@getCustomFieldsSingleSelectEnabled")

```

I haven't tested any of that though, but hopefully we can get the test into a state of failing without your change, then succeeding after the change!

TheAndrewJackson: @allisonking better late than never :sweat_smile: I believe this is fixed now. I tried turning the config option on and off to see the test pass and fail

clients/admin-ui/cypress/e2e/custom-fields.cy.ts
```diff
@@ -150,17 +150,43 @@ describe("Custom Fields", () => {
...
         });
 
         cy.wait("@patchCustomFields");
         // redux should requery after invalidation
-        cy.wait("@getCustomFields");
+        cy.wait("@getCustomFieldSingleSelectEnabled");
 
         cy.get("tbody > tr").first().should("contain", "Multiple select list");
 
+        // the original mock needs to be brock back
```
allisonking, line 181:
```suggestion
        // the original mock needs to be brought back
```

allisonking (approved):
nice!

Issue #3258, opened 2023-05-09
Refactor consent components to use CSS variables that match the brandable colour palette from designs


NevilleS:
### Description
Our latest beautiful consent components designs [Figma link](https://www.figma.com/file/dmEwdK3xZwjKfGVQ9t66xe/Fides-v.2-Master-Working-Files?type=design&node-id=6464-134350&t=8104bV4t4W5TBKPv-0) use 12 brandable colours to set the styles:
<img width="849" alt="image" src="https://github.com/ethyca/fides/assets/1834295/814aa679-5b9e-47b8-ae44-4130aec34f3f">

I'd recommend a nice global colour palette of CSS variables to match these like:
```
--fides-consent-overlay-primary-color (used for "Modal & Banner Outline")
--fides-consent-overlay-primary-button-color (used for "Primary button resting")
--fides-consent-overlay-primary-hover-color (used for "Primary button hover")
--fides-consent-overlay-primary-active-color (used for "Switch active")
--fides-consent-overlay-primary-active-disabled-color (used for "Switch active disabled")
--fides-consent-overlay-inactive-color (used for "Switch inactive")
--fides-consent-overlay-disabled-color (used for "Switch disabled")
--fides-consent-overlay-background-color (used for "Background")
--fides-consent-overlay-row-hover-color (used for "Row hover")
--fides-consent-overlay-row-divider-color (used for "Row divider")
--fides-consent-overlay-title-font-color (used for "Title font")
--fides-consent-overlay-body-font-color (used for "Body font")
```

### Assumptions
Main assumption here is that these should form the *core* 12 brandable colours for all the components. It's also safe to assume we'll want to make it easy to configure these with a styling editor in the future!

### Acceptance Criteria
* Banner styles can be customized using the 12-colour palette above
* Modal styles can be customized using the 12-colour palette above

### Mockups
See Figma links:
* Stylesheet: https://www.figma.com/file/dmEwdK3xZwjKfGVQ9t66xe/Fides-v.2-Master-Working-Files?type=design&node-id=6464-134350&t=8104bV4t4W5TBKPv-0
* Two alternative mockups using different branding: https://www.figma.com/file/dmEwdK3xZwjKfGVQ9t66xe/Fides-v.2-Master-Working-Files?type=design&node-id=6464-134350&t=ls6XpE1B1Ume7uLk-0

Roger-Ethyca:
moving to done

Issue #3161, opened 2023-04-25
Implement Modal Component that renders notices according to config, and saves preferences


eastandwestwind:
- The overlay needs to be triggered by banner or link click
- We can probably use the same Preact component from https://github.com/ethyca/fides/issues/3162

allisonking:
Mostly done here https://github.com/ethyca/fides/pull/3291

But isn't integrated with any endpoints yet, blocked by https://github.com/ethyca/fides/issues/3159

Roger-Ethyca:
moving to done

Issue #3159, opened 2023-04-25
Replace mock GET experience and PATCH privacy-preferences endpoints with real calls


eastandwestwind:
- dependent on https://github.com/ethyca/fides/issues/3228

Replace the following with real API calls:

* GET {{host}}/privacy-experience- this will be a public endpoint in https://github.com/ethyca/fides/issues/3229
* PATCH {{host}}/privacy-preferences- this is currently a public endpoint

eastandwestwind:
Ah good call! https://www.npmjs.com/package/whatwg-fetch looks like a good solution

pattisdr:
Wanted to highlight some intersections: https://github.com/ethyca/fides/issues/3133 and https://github.com/ethyca/fides/issues/3196

I'll be pulling more of these items off of request headers where I can too, but I'll also be adding more fields I need to be in the request body including the experience/experience config history that surfaced the notices to the user to track for consent reporting.

Roger-Ethyca:
moving to done

PR #3291, opened 2023-05-11
Notice toggle table for fides-js

allisonking:
Most of https://github.com/ethyca/fides/issues/3161

Still need to:
* Integrate with the APIs (blocked by https://github.com/ethyca/fides/issues/3159)
* Figure out importing React so that we can maybe use a 3rd party modal for a11y (my limited understanding of rollup is still causing trouble!)
  * Alternatively, don't use the 3rd party modal and try to build it all ourselves (the modal as-is right now is not a11y friendly) 

### Code Changes

* [x] Update rollup to support React (kind of, doesn't work 100% yet 🥲 )
* [x] Add a new component `ConsentModal.tsx` 
  * [x] ❗ doesn't follow a11y rules yet. I think it may be worth using a 3rd party for this one since modals are tricky, but I haven't gotten React importing to work quite right yet
* [x] New component `NoticeToggle` which renders the bulk of the modal contents
* [x] New `Toggle` component which adheres to a11y rules
* [x] Accordion behavior for the notices which also adheres to a11y rules
* [x] Some progress towards using our new CSS vars (which is actually part of https://github.com/ethyca/fides/issues/3258)
* [x] Cypress tests (we should be able to add more once the endpoints are linked up)

### Steps to Confirm

* Run `npm run dev-pc` in the `clients` folder
* Navigate to http://localhost:3000/fides-js-components-demo.html
* In the banner that comes up, click "Manage preferences" which will open the modal
* Toggle the modal and click "Save". Open the console log to see the checked/unchecked values

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### Description Of Changes

image


eastandwestwind:
ah, that's unfortunate to hear. Have you tried adding the rollup typescript plugin?- https://www.npmjs.com/package/rollup-plugin-typescript2

I'm reading https://stackoverflow.com/a/70163537 which seems to be a similar issue, using Rollup. 

If this takes too much time, though, I'm OK with not sharing the toggle component between fides-js and privacy center for now. This can be a follow-up item.

allisonking:
turns out I just needed to alias in two places! 🤦‍♀️ https://github.com/ethyca/fides/pull/3291/commits/0ddc5c4770a21372058c6375a9a476e9e93d9755 thanks for the assist @eastandwestwind 🙌

allisonking:
status so far:
image

and the toggles can update/change their data accordingly.

I'm having trouble importing the `NoticeToggleTable` into the privacy-center, and I'm thinking there's more rollup investigation needed there. I'll probably punt on that for now, and continue trying to get `fides-js` mostly working. What's left:

* divider lines
* accordion to show more info about the notice
* close the modal on save
* mock out the API calls we need with sample payloads. this will be very helpful for testing!
* Cypress tests
* experiment with using https://github.com/KittyGiraudel/react-a11y-dialog to take care of our modal a11y needs

clients/fides-js/src/components/Button.tsx
```diff
@@ -7,7 +7,7 @@ interface ButtonProps {
   onClick?: () => void;
 }
 
-const ConsentBannerButton: FunctionComponent<ButtonProps> = ({
+const Button: FunctionComponent<ButtonProps> = ({
```
allisonking, line 10: I used these in the modal too, so thought I'd make the name more general :)

clients/fides-js/src/components/ConsentBanner.tsx
```diff
@@ -1,35 +1,42 @@
...
 import "../lib/banner.module.css";
 import { useHasMounted } from "../lib/hooks";
 
 interface BannerProps {
+  experience: ExperienceConfig;
+  privacyCenterUrl: string;
   onAcceptAll: () => void;
   onRejectAll: () => void;
-  privacyCenterUrl: string;
-  bannerDescription?: string;
```
allisonking, line 12: instead of taking these individually, I thought it made sense to just pass in the whole `experience` obj

clients/fides-js/src/components/ConsentBanner.tsx
```diff
@@ -38,9 +45,11 @@ const ConsentBanner: FunctionComponent<BannerProps> = ({
...
 
-  const navigateToPrivacyCenter = (): void => {
-    if (privacyCenterUrl) {
+  const handleManagePreferencesClick = (): void => {
+    if (experience.component === ExperienceComponent.PRIVACY_CENTER) {
       window.location.assign(privacyCenterUrl);
+    } else {
+      onOpenModal();
     }
   };
```
allisonking, line 54: ❗ I'm really not sure if this is the right logic. I thought we sometimes wanted to go to PC and sometimes wanted to open the modal here, but I think we might _only_ want to open the modal here? and leave PC opening to links, is that right?

eastandwestwind (commented):
Just tested manually and it's looking good so far! Nice to see you even got the background scrolling working behind the modal :)

https://www.loom.com/share/a9038b903dbd4177b346a82a861570a5

On thing I want to point out is I see that the modal does not currently: 

1. close the banner on save / accept all / reject all- could we add this in easily? Since testing data is passed in on every page load, the banner should still re-open on subsequent page loads, so testing won't be inhibited by this.
2. write the cookie on save, like we do for accept all / reject all- do we have a follow-up ticket for this, or should I handle it on my end?

> Update rollup to support React (kind of, doesn't work 100% yet 🥲 )

Could you describe how to replicate the part that doesn't work with React?

allisonking:
> 1. close the banner on save / accept all / reject all- could we add this in easily? Since testing data is passed in on every page load, the banner should still re-open on subsequent page loads, so testing won't be inhibited by this.
> 2. write the cookie on save, like we do for accept all / reject all- do we have a follow-up ticket for this, or should I handle it on my end?
> 
> > Update rollup to support React (kind of, doesn't work 100% yet 🥲 )
> 
> Could you describe how to replicate the part that doesn't work with React?

Thanks @eastandwestwind , all great points!

1. I just happened to chat with Michael about this behavior, and it turns out we actually want to close the banner when the modal opens. The modal can't be exited until one of the save/accept/reject buttons is clicked. So I'll update this PR to do that 
2. Yes, I forgot about this part 😅 will do that!
3. For React, I didn't commit the parts that weren't working, but basically if we try to add a React package, then import it and use it, the build will fail. I was trying with [react-a11y-dialog](https://github.com/KittyGiraudel/react-a11y-dialog)

eastandwestwind:
Testing functionality now and all is looking good! The banner closes when the modal is opened, and I see the modal preserves state of consented notices when I reload the page to re-trigger the banner. Also the cookie is appropriately updated, along with the window.Fides obj upon saving consent preferences.

I want to confirm- if we have legacy consent preferences, we currently don't override them, just ensure that they remain in the cookie. The only use case where we touch legacy consent preferences is if the overlay is disabled, and we must use read from the legacy `config.json` in the Privacy Center. Does that align with your understanding as well?

eastandwestwind (approved):

allisonking:
> I want to confirm- if we have legacy consent preferences, we currently don't override them, just ensure that they remain in the cookie. The only use case where we touch legacy consent preferences is if the overlay is disabled, and we must use read from the legacy `config.json` in the Privacy Center. Does that align with your understanding as well?

Oh, no I don't think I have that behavior. Am I understanding correctly that let's say...

1. A user has a cookie with a consent entry of i.e. `{ data_sales: true }`
2. This new code now wants to build the cookies based on privacy notices. for now, we are using privacy notice IDs but we will move to keys based on names soon. but let's say it's still id for now, so we add `{ pri_123: true }`
3. The cookie should have `{ data_sales: true, pri_123: true}` ?

That is not true right now, but I can probably merge the cookie values instead of overwriting them?

eastandwestwind:
Ok I think we should keep the functionality already in place, where we overwrite the cookie with the most recent notices. 2 reasons behind this:

1. When we migrate customers to the new privacy experiences, we'll map the old keys to new notices entirely. So we won't have to worry about supporting legacy keys in this scenario.

2. Separately, there's a case where a customer updates their notices so that the browser cookie contains the older invalid cookie key. In this scenario, we should try to honor the new notices as much as possible.

eastandwestwind (approved):

eastandwestwind (approved):

Issue #3241, opened 2023-05-05
Add all US states to location enum


mfbrown:
Support all US states on the enum for privacy notice and experience locations. New state regulations are coming fast so let's have the product ready to support each state, rather than adding new options as laws are passed.

**AC**
* I can specify all 50 US states (and Puerto Rico!) as locations on a privacy notice and privacy experience.

pattisdr:
It would be handy for me to have this in some of the work that I'm doing so I'm going to fold this into another ticket

pattisdr:
Included in https://github.com/ethyca/fides/pull/3292, still hardcoded for now

Roger-Ethyca:
moving to done

Issue #3226, opened 2023-05-04
[Backend] Add Validation on Creating and Updating Experiences


pattisdr:
### Is your feature request related to a specific problem?

- When creating or updating an experience of a certain type, it should have certain fields displayed. For example, a banner needs to have the button texts (among other things)
- When creating or updating an experience of a certain type, check the relevant notices, and make sure the experience updates are valid for the existing notices.

### Describe the solution you'd like

_A description of what you want to happen._

### Describe alternatives you've considered, if any

_A description of any alternative solutions or features you've considered._

### Additional context

_Add any other context or screenshots about the feature request here._

pattisdr:
Included in https://github.com/ethyca/fides/pull/3292

Roger-Ethyca:
moving to done

PR #3292, opened 2023-05-11
[Backend] Syncing Privacy Notices and Privacy Experiences

### LINKED ISSUE #3252: [Backend] Split out Experience Config from Privacy Experiences

Medium-sized backend change to Privacy Experiences to separate storing Privacy Experiences for a single region from common Experience Language that multiple regions can share to aid in keeping privacy notices and experiences in sync for a given region as well as allowing more targeted region-specific changes without affecting other regions.

Simultaneously add the logic to keep Privacy Experiences updated and valid as Privacy Notices and Experience Language change.

See [design doc here](https://ethyca.atlassian.net/wiki/spaces/EN/pages/2683830495/2023-05-06+Backend+Spike+Keeping+Notices+and+Experiences+in+Sync)

This is the result of the spike completed here: https://github.com/ethyca/fides/issues/3193.  A POC is being done for that work, so this should be largely getting that POC ready for review.

It will require corresponding FE changes.

### Additional context

pattisdr:
Closes #3252, #3226, #3241

❗ Contains data migration
❗ Breaks existing privacy experiences
❗ ~Expected to break Admin UI around retrieving experiences. Admin UI for experiences should call new experience config endpoints instead.~
❗ Components should still call existing experience endpoints although the response body has changed.

### Description Of Changes

[See backend design here](https://ethyca.atlassian.net/wiki/spaces/EN/pages/2683830495/2023-05-06+Backend+Spike+Keeping+Notices+and+Experiences+in+Sync)

Medium-sized data model changes to ensure there is always a valid Privacy Experience for each region at all times and keep edits to an experience for one region from conflicting with another. In short, adjusts PrivacyExperiences so instead of representing multiple regions, they represent a _single region_. Create a new PrivacyExperienceConfig table to share common language like banner text and link labels between experiences. PrivacyExperiences are linked to a PrivacyExperienceConfig via FK.  There are also historical tables for both experiences and experienceconfig to track how both of these changed over time for consent reporting.

Further, each time Privacy Notices are created or updated, adjust Privacy Experiences behind the scenes so we have an experience that covers all the notices.  Additionally, when PrivacyExperienceConfigs are updated, also adjust the PrivacyExperiences to match, provided none of the rules with notices are violated. In other words, PrivacyExperiences are not edited directly, but indirectly by PrivacyNotice or PrivacyExperienceConfig updates.

Removes the PrivacyExperienceTemplate table (this wasn't yet being used).  My goal instead is to have a few "PrivacyExperienceConfig" records that are labeled as "default" that we can link to when we need to create or update a PrivacyExperience on the user's behalf.
Screenshot 2023-05-13 at 4 56 58 PM

##### Model changes
* [ ] Splits the `PrivacyExperience` table into two halves: the `PrivacyExperience` and `PrivacyExperienceConfig` table. Each `PrivacyExperience` now supports only a single region. Create historical tables for both.

##### Experience Config 
* [ ] Add new endpoints for the admin UI: `POST {{host}}/experience-config` and `PATCH {{host}}/experience-config/{{experience-config-id}}` for creating and editing experience configs. Have these endpoints create or update a single resource instead of multiple to reduce complexity due to the fact that several PrivacyExperience updates happen behind the scenes.
    * [ ]  When an ExperienceConfig is updated, any regions specified in the request have their PrivacyExperience records upserted to match and linked to the ExperienceConfig via FK provided the rules aren't violated.
* [ ] Adds new endpoints to GET Experience Config List and Detail endpoints for the Admin UI: {{host}}/experience-config?component=overlay&show_disabled=False, and {{host}}/experience-config/{{experience-config-id}}
* [ ] Like notices and experiences, if an ExperienceConfig is updated, the version is bumped and a historical record is created
 

##### Privacy Experience
* [ ] Update PrivacyExperienceList response body  `{{host}}/privacy-experience/?region=eu_de`- since things like component title and banner title are now stored on a different record, these details are now nested under the experience_config_id field.
* [ ] Removes the endpoints to create and update privacy experiences as these records are no longer updated directly
* [ ] Add query params `has_config` to PrivacyExperienceList endpoint to surface which experiences have and don't have the "experience config" linked. If there's no experience config linked, there's no text for the banners, overlays, or links
* [ ] Remove region filter on the PrivacyExperienceDetail endpoint. This was previously used to filter an Experience's associated regions, but now that an Experience is associated with a single region, we just use that one.
* [ ] Note that as before, changes to the PrivacyExperience trigger a version to be bumped and a historical record saved.  I am storing both the `experience_config_id` and the `experience_config_history_id` so if the experience config is updated, this triggers a version bump on the experience.

##### Privacy Notice
- [ ] Add logic to upsert Privacy Experiences behind the scenes when Privacy Notices are created or updated. For example, if a California Opt In Privacy Notice is created that has to be displayed in an overlay, we need to create a California Overlay Privacy Experience that is delivered by link

### Worth noting
- Privacy Notices drive Privacy Experiences not the reverse.  I always check to make sure I have an experience for the given notices not the reverse.  If PrivacyNotice changes are made, they should generally succeed, and you adjust the Experiences to match.  However, if ExperienceConfig changes are made that would conflict with the notices, throw an error message, so the user knows to delete the offending regions from their request.
- If I have to create an overlay experience automatically where I have a choice between being delivered by link or banner I choose banner by default to be consistent.
- Provided privacy notices are valid, I create or update privacy experiences to match.  If one already exists, I leave it.  If a link overlay needs to be updated to a banner overlay, I do so, and unlink its experience config.  If a region is removed from a notice, I leave the related experience untouched. I never delete an experience. It is okay if an experience doesn't have notices.  Just leave the experience there for future use if notices for that region are added again. If a PrivacyExperienceConfig adds a region for which there are no notices, also go ahead and create the Experience behind the scenes.  Again, it's fine if there are no notices.
- If a user updates the PrivacyExperienceConfig to be a delivered via link instead of banner or vice versa, I don't remove the other fields behind the scenes that are traditionally associated with link/banner.  I just leave them there and they won't be used.
- When a user creates or updates ExperienceConfig, the regions included in the request body are not stored on that table, they are each separate records in the PrivacyExperience table.  I make the ExperienceConfig updates first, and then upsert the PrivacyExperiences.  Therefore, require regions to be present in each request to create or update an ExperienceConfig if you want the ExperienceConfig to have those regions. They need to be supplied every time because absence of regions will result in any existing linked PrivacyExperiences to be unlinked
- Right now if I have to create or update a PrivacyExperience automatically when notices change, they don't have an ExperienceConfig attached, so they're not yet valid for display. I'd like to add the concept of default ExperienceConfigs in the future, so I can link these instead.
- Further consider adding a flag in the future to PrivacyExperience for when we make a change on the customer's behalf so we can surface this for approval

 
### Steps to Confirm

See postman collection
* [ ] Create Privacy Notices with `POST {{host}}/privacy-notice`
* [ ] Verify Privacy Experiences were created in the background with `GET {{host}}/privacy-experience/` for each region.
   - Verify privacy center experiences created where notice needs to be displayed in privacy center, same for overlay
   - Note that the relevant notices are still nested under each experience in the response
   - Verify no `experience_config` exists yet
* [ ] Create the language and configuration options for those an overlay experience with `POST {{host}}/experience-config`.  
* [ ] Do a `GET {{host}}/privacy-experience/` for the relevant regions to see that ExperienceConfig is linked to the experience
* [ ] Experiment with updating that ExperienceConfig `{{host}}/experience-config/{{experience-config-id}}` with various types of edits including component, overlay and regions to see that certain regions are blocked if they'll interfere with notices for that region.

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`


src/fides/api/ops/api/v1/endpoints/privacy_experience_config_endpoints.py
```diff
@@ -0,0 +1,327 @@
...
+    linked, unlinked, skipped = upsert_privacy_experiences_after_config_update(
+        db, experience_config, new_regions
+    )
+
+    return ExperienceConfigCreateOrUpdateResponse(
+        experience_config=experience_config,
+        linked_regions=linked,
+        unlinked_regions=unlinked,
+        skipped_regions=skipped,
+    )
```
pattisdr, line 213: This endpoint attempts to _create_ a PrivacyExperienceConfig (with the button labels, component descriptions, etc.) and saves that in one table, and then attempts to _upsert_ PrivacyExperiences for each "region" in the request separately.  

For example, if I am creating a banner overlay PrivacyExperienceConfig and specify the regions are CA, CO, and VA, I make sure banner overlay PrivacyExperiences exist (or create them if they don't) for CA, CO, and VA. I then link each Experience back to the Config via FK.

This is what I'm attempting to surface in the response here with "linked" regions - these are the Experiences that were linked to the Config via FK.  In practice, this endpoint should only have "linked" regions.

src/fides/api/ops/api/v1/endpoints/privacy_experience_config_endpoints.py
```diff
@@ -0,0 +1,327 @@
...
+
+    All regions that should be linked to this ExperienceConfig (or remain linked) need to be
+    included in this request.
+    """
+    experience_config: PrivacyExperienceConfig = get_experience_config_or_error(
+        db, experience_config_id
+    )
+    experience_config_data_dict: Dict = experience_config_data.dict(exclude_unset=True)
+    # Pop the regions off the request
+    regions: List[PrivacyNoticeRegion] = experience_config_data_dict.pop("regions")
```
pattisdr, line 266: You can update most fields individually for a PrivacyExperienceConfig, but one field that I have that needs to be included each time are the regions.  This is because I want to make it possible to unlink all regions too.  If no regions are supplied in the request, all PrivacyExperiences that were previously linked will be unlinked.

Another reason I think regions should be included here is this field in the request body actually pertains to resources in a different table. There are two operations here.  Update the ExperienceConfig, and then Upsert the Privacy Experiences.

src/fides/api/ops/api/v1/endpoints/privacy_experience_config_endpoints.py
```diff
@@ -0,0 +1,327 @@
...
+        data=experience_config_data_dict
+    )
+    try:
+        ExperienceConfigCreate.from_orm(dry_update)
+    except ValueError as exc:
+        raise HTTPException(
+            status_code=HTTP_422_UNPROCESSABLE_ENTITY,
+            # pylint: disable=no-member
+            detail=exc.errors(),  # type: ignore
+        )
```
pattisdr, line 280: This makes sure that we have a link label if this is a Privacy Center Experience Config, for example

src/fides/api/ops/api/v1/endpoints/privacy_experience_config_endpoints.py
```diff
@@ -0,0 +1,327 @@
...
+    ] = remove_config_from_matched_experiences(
+        db,
+        experience_config,
+        [
+            PrivacyNoticeRegion(reg)
+            for reg in {reg.value for reg in current_regions}.difference(
+                {reg.value for reg in regions}
+            )
+        ],
+    )
```
pattisdr, line 306: Here I remove any regions from the ExperienceConfig that weren't included in the request.  This just means I remove the FK between the this ExperienceConfig and any linked PrivacyExperiences: I never delete a PrivacyExperience.

src/fides/api/ops/api/v1/endpoints/privacy_experience_config_endpoints.py
```diff
@@ -0,0 +1,327 @@
...
+        skipped,
+    ) = upsert_privacy_experiences_after_config_update(db, experience_config, regions)
+
+    return ExperienceConfigCreateOrUpdateResponse(
+        experience_config=experience_config,
+        linked_regions=linked,
+        unlinked_regions=unlinked_for_conflict
+        + [omitted.region for omitted in not_included_in_request],
+        skipped_regions=skipped,
+    )
```
pattisdr, line 327: In the response, I distinguish between which new regions were linked to this ExperienceConfig and which regions may have been unlinked from the experience config because they weren't included in the request in case this is useful information to surface in the UI.

Because I'm actually failing early if there are any incompatible regions above in `validate_experience_config_and_region_compatibility`   there shouldn't be any regions "unlinked for conflict" but I still include for completeness.

src/fides/api/ops/api/v1/endpoints/privacy_experience_endpoints.py
```diff
@@ -81,56 +70,34 @@ def privacy_experience_list(
...
             PrivacyExperience.component == component
         )
+    if has_config is True:
+        experience_query = experience_query.filter(
+            PrivacyExperience.experience_config_id.isnot(None)
+        )
+    if has_config is False:
+        experience_query = experience_query.filter(
+            PrivacyExperience.experience_config_id.is_(None)
+        )
```
pattisdr, line 85: I think it would be a really great addition if we had "default" ExperienceConfigs - one for a banner overlay, link overlay, and privacy center that we could link to automatically on the user's behalf, but right now we don't have these in the system. So it's possible we created an Experience that doesn't have any language associated (the config).  

These filters let us surface with PrivacyExperiences have a config (and can be surfaced to an end user) versus don't have a config (and need editing by the compliance person)

src/fides/api/ops/api/v1/endpoints/privacy_experience_endpoints.py
```diff
@@ -81,56 +70,34 @@ def privacy_experience_list(
...
 
     results: List[PrivacyExperience] = []
     for privacy_experience in experience_query.order_by(
         PrivacyExperience.created_at.desc()
     ):
         privacy_notices: List[
             PrivacyNotice
-        ] = privacy_experience.get_related_privacy_notices(db, region, show_disabled)
+        ] = privacy_experience.get_related_privacy_notices(db, show_disabled)
```
pattisdr, line 93: A PrivacyExperience's notices aren't filtered on region anymore, since a PrivacyExperience is only associated with one region.  We just filter then any notices that have the same region as the experience, plus need to be displayed in the same way.

src/fides/api/ops/api/v1/endpoints/privacy_experience_endpoints.py
```diff
@@ -81,56 +70,34 @@ def privacy_experience_list(
...
-    for experience_data in bulk_experience_data:
-        experience = PrivacyExperience.create(
-            db, data=experience_data.dict(exclude_unset=True), check_name=False
-        )
-        # Temporarily stash the privacy notices on the experience for display
-        experience.privacy_notices = experience.get_related_privacy_notices(db)
-        experiences.append(experience)
-    return experiences
-
-
```
pattisdr, line 133: You can no longer edit a PrivacyExperience directly.  Instead, add or update Privacy Notices and I make sure a PrivacyExperience exists in the background that matches those notices.  Alternatively, update the ExperienceConfigs to format the display, and I persist those same changes back to the PrivacyExperiences.

For example, customer creates an Advertising CA Privacy Notice displayed in an overlay.  I then create a CA banner overlay Experience for them.   The customer then creates a banner overlay PrivacyExperienceConfig and specifies CA as a region.  I then link that CA overlay Experience to this PrivacyExperience config via FK.   (another alternative is that I have a default banner overlay experience that I could link to the experience as soon as it was created, that the customer could then updated).

src/fides/api/ops/api/v1/endpoints/privacy_experience_endpoints.py
```diff
@@ -169,51 +130,6 @@ def privacy_experience_detail(
...
-    ]
-
-    return [
-        existing_experience.update(
-            db, data=experience_update_data.dict(exclude_unset=True)
-        )
-        for (existing_experience, experience_update_data) in zip(
-            loaded_privacy_experiences, privacy_experience_updates
-        )
-    ]
```
pattisdr, line 219: You no longer update privacy experiences directly.  They are indirectly updated in the backend as Privacy Notices and ExperienceConfigs change.

src/fides/api/ops/api/v1/endpoints/privacy_notice_endpoints.py
```diff
@@ -252,6 +256,12 @@ def create_privacy_notices(
             check_name=False,
         )
         created_privacy_notices.append(created_privacy_notice)
+        affected_regions.update(created_privacy_notice.regions)
+
+    # After created all notices, make sure experiences exist to back all notices.
+    upsert_privacy_experiences_after_notice_update(
+        db, affected_regions=list(affected_regions)
+    )
```
pattisdr, line 264: Main change to privacy notice endpoints - any time regions are adjusted on privacy notices - whether they're created or updated, for every region potentially touched, I double check that we have a valid PrivacyExperience. If not, I create or update it.   For example, if a CA notice is created that needs to be displayed in the privacy center, I need to make sure I have a CA Privacy Center Experience.  If a CA notice was opt out but updated to be opt in, I need to make sure I have a CA Overlay Experience that is delivered by a banner.

src/fides/api/ops/models/privacy_experience.py
```diff
@@ -32,88 +35,214 @@ class DeliveryMechanism(Enum):
...
+        index=True,
+    )
+
+    experience_config_history_id = Column(
+        String,
+        ForeignKey(PrivacyExperienceConfigHistory.id_field_path),
+        nullable=True,
+        index=True,
+    )  # Also links to the historical record, so if the version of config gets updated, that
+    # triggers a new version of the experience.
```
pattisdr, line 197: I thought it made sense to link both the ExperienceConfig and the ExperienceConfigHistoryId to the PrivacyExperience.    

- experience_config_id makes it easier to fetch `PrivacyExperienceConfig.experiences` - all experiences that currently use this PrivacyExperienceConfig 
- `experience_config_history_id` makes it easier to bump versions.  If the ExperienceConfig itself is updated, its version is bumped, which I think should trigger that the PrivacyExperience's version is likewise incremented.  The PrivacyExperience is now showing different language or features potentially.

src/fides/api/ops/models/privacy_experience.py
```diff
@@ -32,88 +35,214 @@ class DeliveryMechanism(Enum):
...
+
+    # There can only be one component per region
+    __table_args__ = (UniqueConstraint("region", "component", name="region_component"),)
```
pattisdr, line 200: This important piece now enforces at the database level that we can only have one component per region.  One CA overlay. One CA Privacy Center Experience.

src/fides/api/ops/models/privacy_experience.py
```diff
@@ -165,26 +297,302 @@ def update(self, db: Session, *, data: dict[str, Any]) -> PrivacyExperience:
...
+                    component == ComponentType.overlay,
+                    PrivacyNotice.displayed_in_overlay,
+                ),
+                and_(
+                    component == ComponentType.privacy_center,
+                    PrivacyNotice.displayed_in_privacy_center,
+                ),
+            )
+        )
+    )
```
pattisdr, line 394: This existing method extract to be reusued in multiple places.  Noting that PrivacyNotices and PrivacyExperiences still are not linked at all in the database.   We just filter an Experience at runtime to see which Notices share a region and a component at that time.

src/fides/api/ops/models/privacy_experience.py
```diff
@@ -165,26 +297,302 @@ def update(self, db: Session, *, data: dict[str, Any]) -> PrivacyExperience:
...
+                and existing_experience.experience_config_id
+                and existing_experience.experience_config_id == experience_config.id
+            ):
+                # If existing experience is already linked to a config that is now invalid, unlink it.
+                existing_experience.unlink_privacy_experience_config(db)
+                unlinked_regions.append(region)
+            else:
+                # Skip linking experience altogether to config for which it would be incompatible
+                skipped_regions.append(region)
+            continue
```
pattisdr, line 574: In practice, I added validation to the endpoints that create and update an experience config to fail early if we're attempting to make a change that would put Experiences in a bad state, so I wouldn't expect this to get hit.

But I think it's important to prioritize that Experiences always match the Notices, so I'm leaving this second level here as a fallback.

src/fides/api/ops/models/privacy_experience.py
```diff
@@ -165,26 +297,302 @@ def update(self, db: Session, *, data: dict[str, Any]) -> PrivacyExperience:
...
+) -> Tuple[List[PrivacyExperience], List[PrivacyExperience]]:
+    """
+    Keeps Privacy Experiences in sync with *PrivacyNotices* changes.
+    Create or update PrivacyExperiences based on the PrivacyNotices in the "affected_regions".
+    To be called whenever PrivacyNotices are created or updated (pass in any regions that were potentially affected)
+
+    PrivacyExperiences should not be deleted.  It's okay if no notices are associated with an Experience.
+    """
+    added_experiences: List[PrivacyExperience] = []
+    updated_experiences: List[PrivacyExperience] = []
```
pattisdr, line 408: This is one of the primary methods added, to keep PrivacyExperiences updated if PrivacyNotices are changed.

src/fides/api/ops/models/privacy_experience.py
```diff
@@ -165,26 +297,302 @@ def update(self, db: Session, *, data: dict[str, Any]) -> PrivacyExperience:
...
+    regions: List[PrivacyNoticeRegion],
+) -> Tuple[
+    List[PrivacyNoticeRegion], List[PrivacyNoticeRegion], List[PrivacyNoticeRegion]
+]:
+    """
+    Keeps Privacy Experiences in sync with ExperienceConfig changes.
+    Create or update PrivacyExperiences for the regions we're attempting to link to the
+    ExperienceConfig.  If the region cannot be linked, or if it is currently attached
+    and shouldn't be, skip or remove that region.
+    """
```
pattisdr, line 538: This is another of the primary methods added, to keep PrivacyExperiences updated if PrivacyExperienceConfig is changed.

src/fides/api/ops/models/privacy_experience.py
```diff
@@ -165,26 +297,302 @@ def update(self, db: Session, *, data: dict[str, Any]) -> PrivacyExperience:
...
+
+        # See if we need to update an existing Link Overlay to be delivered by Banner instead
+        if banner_delivery_update_needed(overlay_experience, overlay_notices):
+            assert overlay_experience  # For mypy.  Overlay_experience guaranteed to exist here.
+            overlay_experience.update(
+                db=db,
+                data={
+                    "delivery_mechanism": DeliveryMechanism.banner,
+                    "experience_config_id": None,  # We have to unlink the invalid experience config here if it exists
+                    "experience_config_history_id": None,
```
pattisdr, line 476: It would be great if we could have a default banner overlay ExperienceConfig here in the system that we could link to on the user's behalf, same for any other experiences we're automatically creating/updating here.

src/fides/api/ops/schemas/privacy_experience.py
```diff
@@ -12,86 +12,172 @@
...
+                "Link label required when the delivery mechanism is of type link."
+            )
+
+        if component == ComponentType.overlay:
+            if delivery_mechanism == DeliveryMechanism.banner:
+                required_banner_fields = [
+                    "acknowledgement_button_label",
+                    "banner_title",
+                    "confirmation_button_label",
+                    "reject_button_label",
```
pattisdr, line 78: I included all three of these button labels here as "required" for setting up a banner overlay.  There may not be notice only notices, but this prevents us from having to check that. I'm trying to minimize the number of things we need to check on a notice when we create an ExperienceConfig and link a region to it.

pattisdr:
@allisonking, @eastandwestwind, @NevilleS could you take a look at the changes made here?  Playing with the postman collection here is a good way to get up to speed with what's happening.

src/fides/api/ops/models/privacy_experience.py
```diff
@@ -165,26 +297,302 @@ def update(self, db: Session, *, data: dict[str, Any]) -> PrivacyExperience:
...
+        db: Session, region: PrivacyNoticeRegion
+    ) -> Tuple[Optional[PrivacyExperience], Optional[PrivacyExperience]]:
+        """Load both the overlay and privacy center experience for a given region"""
+        overlay_experience: Optional[PrivacyExperience] = (
+            db.query(PrivacyExperience)
+            .filter(
+                PrivacyExperience.region == region,
+                PrivacyExperience.component == ComponentType.overlay,
+            )
+            .first()
```
eastandwestwind, line 325: this .first() is just a formality, right? As in the db should validate uniqueness on region and overlay

src/fides/api/ops/models/privacy_experience.py
```diff
@@ -165,26 +297,302 @@ def update(self, db: Session, *, data: dict[str, Any]) -> PrivacyExperience:
...
+    )
+
+
+def upsert_privacy_experiences_after_notice_update(
+    db: Session, affected_regions: List[PrivacyNoticeRegion]
+) -> Tuple[List[PrivacyExperience], List[PrivacyExperience]]:
+    """
+    Keeps Privacy Experiences in sync with *PrivacyNotices* changes.
+    Create or update PrivacyExperiences based on the PrivacyNotices in the "affected_regions".
+    To be called whenever PrivacyNotices are created or updated (pass in any regions that were potentially affected)
```
eastandwestwind, line 403: really descriptive comment here!

src/fides/api/ops/models/privacy_experience.py
eastandwestwind: agreed. It will be super useful to have some default ExperienceConfigs on the BE in these scenarios. But also, I assume we should still surface to the user when a default has been applied so that they can choose to edit it?

eastandwestwind (commented):
This looks great overall, just a couple Qs for you @pattisdr. Also I haven't manually tested this PR yet, just looked over the code portion.

src/fides/api/ops/models/privacy_experience.py
pattisdr: yes exactly! I've been mulling over how best to do this (think it should be in the follow-up "defaults" ticket) https://github.com/ethyca/fides/issues/3305.  I was wondering about just a flag on PrivacyExperience, "automatically applied" that I would use if this happened for privacy notices (but not experience configs).  I'm not sure how they "confirm" this though?

src/fides/api/ops/api/v1/endpoints/privacy_experience_config_endpoints.py
pattisdr: sure can!

src/fides/api/ops/api/v1/endpoints/privacy_experience_config_endpoints.py
pattisdr: I personally don't think you need to distinguish between that in the UI, but you could surface which regions were "unlinked" as a whole to the user, if that might be useful to them.  Mainly, if they have *removed* a region from this ExperienceConfig that currently means that region isn't linked to an ExperienceConfig at all (or we added the default for them) and it likely needs more attention.

src/fides/api/ops/models/privacy_experience.py
pattisdr: yes and further, I do need to return the first PrivacyExperience/None, not the "Query" object. What is nice about `first()` is that it returns the first result of the query, or None if it doesn't exist!

src/fides/api/ops/models/privacy_experience.py
pattisdr: @eastandwestwind are you okay with us continuing to discuss how to best do this in the followup default ticket?

allisonking (approved):
👏

PR #3299, opened 2023-05-12
Handle edge case where no attributes are defined

seanpreston:
### Description Of Changes

On further testing the DynamoDB connector this afternoon I was finding the following errors in the logs:
```
2023-05-12 22:08:37.130 [INFO] (graph_task:pre_process_input_data:327): Consolidating incoming data into dynamodb_example_test_dataset:login from __ROOT__:__ROOT__.
2023-05-12 22:08:37.130 [INFO] (graph_task:pre_process_input_data:327): Consolidating incoming data into dynamodb_example_test_dataset:login from dynamodb_example_test_dataset:customer_identifier.
  File "/usr/local/lib/python3.10/site-packages/fides/api/ops/task/graph_task.py", line 92, in result
    return func(*args, **kwargs)
  File "/usr/local/lib/python3.10/site-packages/fides/api/ops/task/graph_task.py", line 529, in access_request
    output: List[Row] = self.connector.retrieve_data(
  File "/usr/local/lib/python3.10/site-packages/fides/api/ops/service/connectors/dynamodb_connector.py", line 111, in retrieve_data
    for identifier in input_data[attribute_name]:
KeyError: 'customer_id'
```
where the DynamoDB connector was failing due to Fides expecting attribute definitions to always exist.

This PR also adds effective `traeback.print_exc()` to the exceptions raised when graph nodes fail. This will print the exception trace each time it fails (for every retry) instead of the code's original intention, which was to print it only at the end once every try has failed. There's a trade off here:
- this could spam the logs in the event that Fides prints the same error `n` times
- this could be useful if the task fails `n` times and then succeeds before the retry limit is reached — anybody reading those logs would see why the node failed

My instinct is that the first case there will prove the most common, to maintain that functionality we'd have to use something other than `traceback.print_exc()`, because it only works within the context of raised exceptions.

### Code Changes

* [ ] Only iterate through attribute defs if they exist
* [ ] Move `traceback.print_exc()` calls to inside the `except` block, where they can function
* [ ] Set `log_pii` always to `true` in the sample environment

### Steps to Confirm

* [ ] Add a DynamoDB connector and hook it up to the same account our tests use
* [ ] Configure the [same dataset as our tests use](https://github.com/ethyca/fides/blob/main/data/dataset/dynamodb_example_test_dataset.yml)
* [ ] Run a DSR through
* [ ] Assert that it doesn't fail on the DynamoDB dataset

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### DIFF
--- a/src/fides/api/ops/service/connectors/dynamodb_connector.py
+++ b/src/fides/api/ops/service/connectors/dynamodb_connector.py
@@ -108,7 +108,7 @@ class DynamoDBConnector(BaseConnector[Any]):  # type: ignore
             query_config = self.query_config(node)
             for attribute_definition in query_config.attribute_definitions:  # type: ignore
                 attribute_name = attribute_definition["AttributeName"]
-                for identifier in input_data[attribute_name]:
+                for identifier in input_data.get(attribute_name, []):
                     selected_input_data = input_data
                     selected_input_data[attribute_name] = [identifier]
                     query_param = query_config.generate_query(
--- a/src/fides/api/ops/task/graph_task.py
+++ b/src/fides/api/ops/task/graph_task.py
@@ -91,6 +91,7 @@ def retry(
                     # Run access or erasure request
                     return func(*args, **kwargs)
                 except PrivacyRequestPaused as ex:
+                    traceback.print_exc()
                     logger.warning(
                         "Privacy request {} paused {}",
                         method_name,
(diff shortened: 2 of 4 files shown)


SteveDMurphy (approved):
Nice catch 👍🏽 thanks for digging into this and the steps to test @seanpreston ! More learning for me in all of the moving pieces here

PR #3312, opened 2023-05-16
Harmonize fideskey regex with fideslang

daveqnet:
The fideslang regex in the fidesKey class at https://github.com/ethyca/fideslang/blob/main/src/fideslang/validation.py is less permissive than the one being replaced here

### Code Changes

* [x] removed the custom `ALLOWED_CHARACTERS` regex validation in favor of the standard `FidesKey` validation
* [x] reorganize tests for `save`
* [x] refactor/clean up validation flow for `save`

### Steps to Confirm

* [ ] tests pass

### Pre-Merge Checklist

* [x] Update `CHANGELOG.md`

### Description Of Changes

See summary


ThomasLaPiana:
@daveqnet I'm assigning this to myself because I might end up just replacing this validation with what exists in `fideslang`

Thanks for the catch here!

adamsachs (commented):
generally seems like a nice clean update. i don't think there is any good reason for why the `key`s on the DB classes/tables with that column _shouldn't_ also have the same standards as `fides_key`s. 

a couple of questions:
- is it worth us creating a follow-up ticket to migrate over to using `fides_key` as the attribute everywhere, rather than `key`? i think that discrepancy is a vestige of the ops/ctl divide and it feels odd to me that we're continuing to use both in different places, especially if we're now saying the values have to meet the same validation criteria. updating `key` --> `fides_key` everywhere is a _significantly_ more involved update, though, so definitely not suggesting we tackle that as part of this PR. also would like @pattisdr to weigh-in here as i think she probably has the most historical context around the `key` attribute in the "ops" world.
- what about existing records of classes that inherit from `OrmWrappedFidesBase`? should we not include a migration (or something else) to flag (if not attempt to remediate) any non-compliant `key` values on existing records? or are we just OK leaving those as-is? i suppose there wouldn't be any real _problems_ if we did leave those records as-is, it just doesn't feel very clean.

fwiw, the `assert` in application code caught my attention, but it feels appropriate in this context since it indicates a condition we should never be able to hit without a significant programming error (that would need to be addressed immediately!)

tests/ops/models/test_base.py
```diff
@@ -75,28 +75,37 @@ def test_update_key(db: Session, storage_config, privacy_request):
...
+class TestSaveKey:
+    def test_save_key_invalid(self, db: Session, storage_config):
+        "Test prevent saving bad keys."
+        storage_config.key = "bad key"
+        with pytest.raises(FidesValidationError) as exc:
+            storage_config.save(db)
+
+        error_output = str(exc.value)
+        expected_output = "FidesKeys must only contain alphanumeric characters, '.', '_', '<', '>' or '-'. Value provided: bad key"
+        print(error_output)
```
pattisdr, line 87: I'd remove these extra print statements out of these tests -

src/fides/api/ops/db/base_class.py
```diff
@@ -296,15 +294,8 @@ def save(self, db: Session) -> FidesBase:
...
-
-            is_valid = False
-            if key is not None:
-                is_valid = len(ALLOWED_CHARS.sub("", key)) == 0
-
-            if not is_valid:
-                raise KeyValidationError(
-                    f"Key '{key}' on {self.__class__.__name__} is invalid."
-                )
+            assert key, f"Key on class '{self.__class__.__name__}' cannot be empty!"
```
pattisdr, line 297: can we remove the assert and throw a proper error?

src/fides/api/ops/db/base_class.py
```diff
@@ -296,15 +294,8 @@ def save(self, db: Session) -> FidesBase:
...
             key = getattr(self, "key")
-
-            is_valid = False
-            if key is not None:
-                is_valid = len(ALLOWED_CHARS.sub("", key)) == 0
-
-            if not is_valid:
-                raise KeyValidationError(
-                    f"Key '{key}' on {self.__class__.__name__} is invalid."
-                )
```
pattisdr, line 307: This seems like it has some unintentional side effects.  This validation was preventing "<" and ">" characters from getting saved as fides keys, and that's no longer the case.


 The Fideslang fides key validation allows "<" ">" characters for the purpose of being able to serialize the Saas Configs in memory which often have a placeholder key until the user gives it an actual key:

```yaml
dataset:
  - fides_key: <instance_fides_key>
    name: Amplitude
    description: A sample dataset representing the Amplitude connector for Fides
    collections:
   ...
``` 

But it still seems useful to prevent those characters when we actually save the resource.

src/fides/api/ops/db/base_class.py
```diff
@@ -20,7 +19,6 @@
 from fides.api.ops.util.text import to_snake_case
 
 T = TypeVar("T", bound="OrmWrappedFidesBase")
-ALLOWED_CHARS = re.compile(r"[A-z0-9\-_]")
```
pattisdr, line 23: As @adamsachs said, remnants of the fidesops/fidesctl merge.  An earlier version of this was in solon:

https://github.com/ethyca/solon/commit/1b2aa1bc509338045350185619e78d36975cd52a#diff-e6e83ed3d3f8c6a40f4532ab5b6233b24ff79303fcaa6ae12bb08689e546ed18R220

src/fides/api/ops/db/base_class.py
adamsachs: i too thought this looked funny but figured it may be OK in this sort of context? but if you're also mentioning it, then seems like it's worth addressing!

tests/ops/models/test_base.py
ThomasLaPiana: Why? I find them extremely useful when debugging tests, and they only show up in the `stdout` if the test fails, so I consider them not only harmless but actually free debugging help. Pytest captures `print` statements on a per-test basis

src/fides/api/ops/db/base_class.py
ThomasLaPiana: I haven't heard that word in a long time, nostalgic haha

src/fides/api/ops/db/base_class.py
ThomasLaPiana: When to use `assert` vs a general `exception` is very much a subjective matter, but for me the guidelines are generally:

- Exceptions: Use an `exception` when the error would be caused by input from a 3rd-party, such as user input or API call
- Assertion: Use when the "call is coming from inside the house", i.e. your code itself created the input that would allow the program to get into a bad state and you want to catch that early and fail fast.

To confirm, there is no validation being done before this step? Such as when Pydantic/FastAPI validate the payload of the incoming request?

I'm by no means going to staunchly defending using an `assert` here, but based on my understanding of the code and generally how I employ asserts it made sense to me here. I also could be making an incorrect assumption about the code up to this point though, and that this really is the only point of validation

src/fides/api/ops/db/base_class.py
ThomasLaPiana: I see, so you're saying it makes more sense to have a _stricter_ character check here because it could prevent a bug where a user doesn't assign a proper value and a placeholder key gets saved?

ThomasLaPiana:
I'll clarify here that this is an _extremely_ low-priority change, it was something that was automatically flagged by CodeQL I think. So if this has _any_ potential to cause bugs or headaches, it's not worth it in my opinion.

Do you concur with my assessment @daveqnet or is there other value here?

src/fides/api/ops/db/base_class.py
pattisdr: Pydantic is validating the request but this step was previously getting hit for various characters discussed below https://github.com/ethyca/fides/pull/3312#discussion_r1196965142 (which also seems like something we should have been catching and throwing a proper HTTP Exception in the endpoint.)  However, I don't expect this particular assert to be hit.

src/fides/api/ops/db/base_class.py
pattisdr: Yes, this wasn't intentional but it seems useful in hindsight.

pattisdr:
> To be very specific, it's the A-z part that's overly permissive.

Thanks for explaining @daveqnet!  Your suggestion to restrict to `[A-Za-z0-9\-_]` makes sense to me.  Luckily we're also usually validating against `FidesKey` as a first round on the endpoint before we get here too, so hopefully that means these characters haven't squeaked through.

ThomasLaPiana:
> > To be very specific, it's the A-z part that's overly permissive.
> 
> Thanks for explaining @daveqnet! Your suggestion to restrict to `[A-Za-z0-9\-_]` makes sense to me. Luckily we're also usually validating against `FidesKey` as a first round on the endpoint before we get here too, so hopefully that means these characters haven't squeaked through.

Thank you all for the discussion, I'll implement as discussed 🙂

ThomasLaPiana:
@adamsachs I know you're busy with releasing but can you give this a quick review? thank you!

adamsachs (changes requested):
@ThomasLaPiana  unless i'm missing something, it looks like we're not actually making the discussed change :).

@daveqnet suggested `[A-Za-z0-9\-_]` as the new regex, but i think you accidentally reverted back to `[A-z0-9\-_]`, which is what had been in place originally. looking at the full diff makes this obvious as it shows no change to the `ALLOWED_CHARS` constant!

relatedly, i also wonder if we could include a test case that covers the special chars that we're blocking with this proposed change? i.e. trying to specify a fides key with one of the ``` [ \ ] ^ _ ` ``` characters and ensuring it gets rejected.

ThomasLaPiana:
> @ThomasLaPiana unless i'm missing something, it looks like we're not actually making the discussed change :).
> 
> @daveqnet suggested `[A-Za-z0-9\-_]` as the new regex, but i think you accidentally reverted back to `[A-z0-9\-_]`, which is what had been in place originally. looking at the full diff makes this obvious as it shows no change to the `ALLOWED_CHARS` constant!
> 
> relatedly, i also wonder if we could include a test case that covers the special chars that we're blocking with this proposed change? i.e. trying to specify a fides key with one of the `` [ \ ] ^ _ ` `` characters and ensuring it gets rejected.

You're totally right, no idea how I messed up a copy/pasta 🤦 I fixed the regex and added some tests for invalid characters, good call there 🙂

adamsachs (approved):
looks great @ThomasLaPiana, thanks for the quick updates! think you just need to merge in/rebase `main`, will let you take care of that!

PR #3331, opened 2023-05-18
Enhance Audit Log Data Beyond Privacy Requests

SteveDMurphy:
Closes [fde#8](https://github.com/ethyca/fde/issues/8)

### Code Changes

* [x] Create a new model and migration to house the data
* [x] Add middleware to write the non-`GET` data to the server
* [x] Create a new `middleware.py` to house functionality
* [x] Set a config variable to allow the logging to be enabled or disabled (disabled by default)

### Steps to Confirm

* [ ] To enable the new audit logging, add `FIDES__SECURITY__ENABLE_AUDIT_LOG_RESOURCE_MIDDLEWARE: "True"` to the compose file then running `nox -s dev`
* [ ] Create a system (or do anything the requires a `POST` or `PUT` 
* [ ] Validate the record is written to the audit log resource table (`select * from public.audit_log_resource;`)
* [ ] Click around and do other things to ensure that a record is properly written
* [ ] Cypress tests pass for all routes when middleware enabled

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### Description Of Changes

This PR will address the who and when with a high-level what more than the what (detail) has changed. A design doc will follow as part of this issue closure.

Interestingly, the new middleware severely impacted pytest (i.e. it didn't run). Adding the config variable to disallow it was an option I was already interested in but became a necessity for that reason.


src/fides/api/ctl/sql_models.py
```diff
@@ -579,3 +579,15 @@ class CustomField(Base):
...
     UniqueConstraint("resource_type", "resource_id", "custom_field_definition_id")
+
+
+class AuditLogResource(Base):
+    """The log of user actions against fides resources."""
+
+    __tablename__ = "audit_log_resource"
+
+    user_id = Column(String, nullable=True, index=True)
+    request_path = Column(String, nullable=True)
```
seanpreston, line 590: This could give out-of-date information if we ever changed the URLs around in Fides — I wonder if we can resolve this field to some constant representing the endpoint name? Probably something to log for a future update.

SteveDMurphy: Ah, ya I can see that making sense for sure. I do still sort of like the traceability as to exactly what endpoint was hit at what point in time. If the data is still planned to be life-limited it might be moot as well

SteveDMurphy:
@seanpreston still adding some more tests but wanted to at least get this back on the radar so it can go with the release - thanks again for the help!!

.fides/db_dataset.yml
adamsachs, line 1: whoa, the diff here is huge. a bit hard for me to track but wanted to flag this just to ensure the massive diff is deliberate?

src/fides/api/middleware.py
```diff
@@ -0,0 +1,113 @@
...
+    db: Session = deps.get_api_session()
+
+    # get the user id associated with the request
+    token = request.headers.get("authorization")
+    if token:
+        client = await extract_client_id(token.replace("Bearer ", ""), db)
+        audit_log_resource_data["user_id"] = client.user_id or "root"
+
+    # Access request body to check for fides_keys
+    await set_body(request, await request.body())
```
adamsachs, line 45: do you mean to be calling this here and then again within `get_body()` (i.e. on L112). i could definitely be missing the point here, but this strikes me as a bit redundant - it feels like `get_body()` may have all that's needed, and this was just left in from an earlier iteration?

if i'm wrong about that, then maybe you could just update the code comments a bit to clarify specifically what's going on here, since this does feel a bit circular/redundant. or maybe i'm just being dense and not following something that's obvious 😄

src/fides/api/ctl/sql_models.py
```diff
@@ -579,3 +579,15 @@ class CustomField(Base):
...
 
     UniqueConstraint("resource_type", "resource_id", "custom_field_definition_id")
+
+
+class AuditLogResource(Base):
+    """The log of user actions against fides resources."""
+
+    __tablename__ = "audit_log_resource"
+
+    user_id = Column(String, nullable=True, index=True)
```
adamsachs, line 589: can this really be nullable? looking at the code, it's null in the case where we don't find an authorization header - but is that ever really the case? i guess it doesn't hurt too much to keep it flexible, so i'm not opposed to it. just wanted to flag it in case this wasn't a deliberate choice.

src/fides/api/middleware.py
```diff
@@ -0,0 +1,113 @@
...
+        "request_type": request.method,
+        "fides_keys": None,
+        "extra_data": None,
+    }
+    db: Session = deps.get_api_session()
+
+    # get the user id associated with the request
+    token = request.headers.get("authorization")
+    if token:
+        client = await extract_client_id(token.replace("Bearer ", ""), db)
```
adamsachs, line 41: looks like you'd extracted (pun intended 😉 ) this into a utility function in `get_client_user_id` and just forgot to invoke that? i like the refactor to put that in a utility function i think, but also good with not using it. but feels redundant/confusing to me as it is now!

src/fides/api/oauth/utils.py
```diff
@@ -187,6 +187,23 @@ async def verify_oauth_client(
...
+async def extract_client_id(
+    authorization: str,
+    db: Session = Depends(get_db),
+) -> ClientDetail:
+    """
+    Verifies that the access token provided in the authorization header contains
+    the necessary scopes or roles specified by the caller. Yields a 403 forbidden error
+    if not.
+
+    NOTE: This function may be overwritten in `main.py` when changing
```
adamsachs, line 199: nit - looks like this may have been copy pasta'd from the `verify_oauth_client` doc string, and it doesn't apply to this function. is that right?

adamsachs (commented):
overall i think this looks like a really solid first iteration @SteveDMurphy. the code is pretty easy to follow, i like how you've organized things and you've set up a great foundation! gating this behind a config var (that defaults to false) i think is a great and necessary choice to get this out in the wild while mitigating any risk.

i've got some minor cleanup comments, some of which would be nice to address before merging, i think. beyond that, there are a couple of more "meaty" questions about how we can ensure this all runs properly _in the background_ and stays decoupled from actual API/application tasks - as i note in the comments, i don't think these need to block this initial iteration, but i believe they should be seriously considered before we're ready to deploy this in a production (or any client) environment.

let me know what you think and if you'd like to link up at all to discuss!

src/fides/api/ctl/sql_models.py
SteveDMurphy: Ya so it is null on login and null for the root user (I think I handled those so far) so after finding it in a couple of places I felt like keeping it as flexible as possible. If it fails I also want it to not really impact the user either (although I don't really know of a great way to be alerted to issues)

src/fides/api/middleware.py
SteveDMurphy: No not obvious at all! I accidentally stashed the comments instead of committing them, which linked to [this GitHub issue](https://github.com/tiangolo/fastapi/issues/394#issuecomment-883524819) that highlights this as a problem with FastAPI/Starlette and provides the workaround. Will fix this and make it clear as to why this is happening (lost half a day on this)

src/fides/core/config/security_settings.py
```diff
@@ -122,6 +122,10 @@ class SecuritySettings(FidesSettings):
         default=False,
         description="Enables or disables the ability to import connector templates with custom functions. When enabled, custom functions which will be loaded in a restricted environment to minimize security risks.",
     )
+    enable_audit_log_resource_middleware: Optional[bool] = Field(
```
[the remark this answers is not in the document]
SteveDMurphy: Sounds good, thank you for pointing me to it!

src/fides/api/middleware.py
SteveDMurphy: Yep ha, did some reorganization in test writing and missed - nice catch!

src/fides/api/oauth/utils.py
SteveDMurphy: Correct 👍🏽

src/fides/api/middleware.py
```diff
@@ -0,0 +1,113 @@
...
+
+    # details to be stored as a row on the server
+    audit_log_resource_data = {
+        "user_id": None,
+        "request_path": request.scope["path"],
+        "request_type": request.method,
+        "fides_keys": None,
+        "extra_data": None,
+    }
+    db: Session = deps.get_api_session()
```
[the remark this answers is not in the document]
SteveDMurphy: Ya I honestly didn't dive in deeply there - I thought we had made a change where everything was async. Will add a follow up for that as well unless we want it now. I don't think it should be too much work to make the change but will defer to you

src/fides/api/oauth/utils.py
SteveDMurphy: I _think_ I am able to just remove it completely, tested well for me just so now so am going to just go straight to `extract_token_and_load_client` instead

src/fides/api/middleware.py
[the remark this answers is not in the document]
adamsachs: addressing in a follow up sounds great 👍

src/fides/api/ctl/sql_models.py
adamsachs: that makes sense, let's keep it flexible for now!

src/fides/api/ctl/sql_models.py
```diff
@@ -579,3 +579,15 @@ class CustomField(Base):
     )
 
     UniqueConstraint("resource_type", "resource_id", "custom_field_definition_id")
+
+
+class AuditLogResource(Base):
```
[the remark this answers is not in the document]
adamsachs: no strong opinion, your reasoning makes sense. i'm good with this for now, we can pretty easily update this later so no need to split hairs on it now!

src/fides/api/ctl/sql_models.py
SteveDMurphy: Maybe it makes sense to at least include the field in the migration for now and then we can add it in later as needed - would that work?

src/fides/api/main.py
```diff
@@ -239,3 +242,22 @@ def start_webserver(port: int = 8080) -> None:
...
+) -> Response:
+    """Log basic information about every non-GET request handled by the server."""
+
+    if (
+        request.method != "GET"
+        and request.scope["path"] not in IGNORED_AUDIT_LOG_RESOURCE_PATHS
+        and CONFIG.security.enable_audit_log_resource_middleware
+    ):
+        try:
+            await handle_audit_log_resource(request)
```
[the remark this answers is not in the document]
SteveDMurphy: I completely agree with this @adamsachs , the big issue I have seen related however is that `BackgroundTask` isn't actually running in the background (i.e. it blocks subsequent calls)

We ran into this same scenario with running `Classify` assuming a `BackgroundTask` ran in the background and instead had to opt for some threading workarounds. It made for some interesting learning at the time, (I still need to validate this is the case today) but it was viewed more as a feature than a bug by FastAPI/Starlette from what I can recall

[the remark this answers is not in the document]
adamsachs: yup, i had that suspicion too (which is why i was hedging and not suggesting we try to address right this moment! 😄 ). completely agree with your take on it. let's just make sure we dig into this more deeply (and hopefully resolve!) in a follow-up.

tests/ops/api/v1/test_middleware.py
```diff
@@ -0,0 +1,128 @@
...
+        "user_id": None,
+        "request_path": "some/path",
+        "request_type": "POST",
+        "fides_keys": ["new_fides_key"],
+    }
+    yield audit_log_resource_data
+
+
+def test_record_written_to_db(db, test_audit_log_resource_data: Dict[str, Any]) -> None:
+    assert _middleware.write_audit_log_resource_record(db, test_audit_log_resource_data)
```
adamsachs, line 74: what are we actually asserting here? i also think this would need to be awaited since it's an async function right?

i may be wrong about this, just stuck out to me on first glance. anyway, we don't need to hold up the PR, just noting here so we don't lose track of this.

adamsachs: resolving the conversation to unblock merging but let's come back to this unless i'm missing the point!

.fides/db_dataset.yml
adamsachs: updated in https://github.com/ethyca/fides/pull/3331/commits/0a2f66ac2d83d2befc59ad1d336219e11105b9fd, we can unindent in a follow up PR dedicated to this cleanup if that's a good route to go

src/fides/api/ctl/sql_models.py
adamsachs: seems like a reasonable idea, but given some time sensitivity with the release, let's tackle as a follow up to get this merged 😄

adamsachs:
merging in `main` then merging the PR - CI had passed previously, just wanted to make sure the `fides_db_scan` still passed after my `db_dataset.yml` tweak!

adamsachs (approved):
looks great as a first iteration, follow ups have been noted and issues will be created 👍

PR #3337, opened 2023-05-19
[Backend] Add Required Notice Key to Privacy Notices 

### LINKED ISSUE #3281: Add a notice_key to the privacy notices to enable FE integrations

pattisdr:
Closes #3281 
❗ Contains migration

### Description Of Changes

We need a key for each notice which will be exposed via the Fides consent cookie and the consent object for third parties to use when deciding to load scripts and pixes.

We want a notice key in the database.  Out of the box we will pre-populate this but then we can edit it as needed.  We will likely add some frontend UX to discourage changing this key. An important criteria right now is that the keys are not unique.

[Internal context here](https://ethyca.slack.com/archives/C04S7K0T2V7/p1684440314641139)

### Code Changes

* [ ]  Add new required PrivacyNotice.notice_key field that needs to be a fides key
* [ ] Migrate any existing notices in the db to have this field from a lower cased underscored name
* [ ] Now required in the request to create a PrivacyNotice

### Steps to Confirm

* [ ] In Postman the Post privacy notices should be supplied a notice key.  If none supplied, I create one.

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`


src/fides/api/models/privacy_notice.py
```diff
@@ -160,6 +162,15 @@ def applies_to_system(self, system: System) -> bool:
...
         return False
 
+    @classmethod
+    def generate_notice_key(cls, name: Optional[str]) -> FidesKey:
+        """Generate a notice key from a notice name"""
+        if not isinstance(name, str):
+            raise Exception("Privacy notice keys must be generated from a string.")
+        notice_key: str = re.sub(r"\s+", "_", name.lower().strip())
+        FidesKey.validate(notice_key)
+        return notice_key
```
pattisdr, line 172: Not yet being used, but I imagine I can do something like this when we start generating notices out of the box - or! we can already have the notice key in the yaml as well.

src/fides/api/schemas/privacy_notice.py
```diff
@@ -105,6 +107,7 @@ class PrivacyNoticeCreation(PrivacyNotice):
     """
 
     name: str
+    notice_key: FidesKey
```
pattisdr, line 110: This makes this field required on create.  Do we have the ability to create privacy notices in the frontend yet (I don't think so). If not, I wouldn't consider this being blocked by any frontend change.

allisonking (approved):
this looks great! let's coordinate on FE changes...

src/fides/api/models/privacy_notice.py
pattisdr: We can go either way here! In the current form, I make `notice_key` required, but I can make it optional and use this instead.  That would also help separate the frontend from the backend.

pattisdr:
OK @allisonking I'm creating the notice key if not supplied here as a temporary measure to move this forward.

allisonking (approved):
perfect, thanks @pattisdr !!

PR #3338, opened 2023-05-19
Fix `mypy src/`

seanpreston:
This PR fixes the `mypy` issues caused by removing `lib`. Some general rules I've followed:
- Often during request execution we'll add temporary fields onto the `PrivacyRequest` object ([e.g `PrivacyRequest.action_required_details`](https://github.com/ethyca/fides/blob/main/src/fides/api/api/v1/endpoints/privacy_request_endpoints.py#L519)) — in these instances I've chosen to add those fields as `field: Optional[...] = None` to the `PrivacyRequest` model such that they won't be recorded at the DB level and always default to None
- There are lots of instances where our API specifies a `PrivacyRequestResponse` return type, and a `PrivacyRequest` model is returned — I've opted to ignore those as FastAPI joins the dots for us
- Anywhere `sqlalchemy` foreign keys are used from the linked model, `mypy` does not recognize them, and throws an `attr-defined` error — those are also safely ignored in many other places throughout the codebase


src/fides/api/api/v1/endpoints/privacy_request_endpoints.py
```diff
@@ -840,7 +853,7 @@ def resume_privacy_request(
         privacy_request_id=privacy_request.id,
         from_webhook_id=webhook.id,
     )
-    return privacy_request
+    return privacy_request  # type: ignore[return-value]
```
seanpreston, line 856: I've chosen to ignore these as it's a common pattern where by returning a `PrivacyRequest` it coerces to a `PrivacyRequestResponse`

src/fides/api/api/v1/endpoints/privacy_request_endpoints.py
```diff
@@ -1749,4 +1759,4 @@ def _process_privacy_request_restart(
         from_step=failed_step.value,
     )
 
-    return privacy_request
+    return privacy_request  # type: ignore
```
ThomasLaPiana, line 1762: instead of blanket ignores, can you add specific ignore rules here like you did elsewhere? Makes it a lot easier to understand _why_ it is ignored and potentially fix later

src/fides/api/models/privacy_request.py
```diff
@@ -957,7 +964,7 @@ def get_cached_identity_data(self) -> Dict[str, Any]:
     def verify_identity(
         self,
         db: Session,
-        provided_code: str,
+        provided_code: str = None,
```
ThomasLaPiana, line 967:
```suggestion
        provided_code: Optional[str] = None,
```

I don't think the current type makes sense technically? Not sure why `mypy` isn't catching it. An alternative is also

`provided_code: str = "",`

seanpreston: I left it as `None` explicitly to handle the case where the method is called like this:

```
verify_identity(db, provided_code=None)

# or
code = None
verify_identity(db, provided_code=code)
```

ThomasLaPiana: are there cases where the method is called like that? That still doesn't seem right...

If it is invoked as `provided_code=None` that's not correct from a type standpoint. An empty string also evaluates to `False` in a truthy check so `provided_code: str = ""` makes sense if we want to let it default to a falsey value?

seanpreston: This is the consent request verification method — there are several existing places in the codebase this is called with a `None` value:
- https://github.com/ethyca/fides/blob/main/src/fides/api/api/v1/endpoints/consent_request_endpoints.py#L283
- https://github.com/ethyca/fides/blob/main/src/fides/api/api/v1/endpoints/consent_request_endpoints.py#L414

ThomasLaPiana (approved):
Thank you for digging into these fixes <3

PR #3342, opened 2023-05-22
Rename migrate -> upgrade

ThomasLaPiana:
### Code Changes

* [x] update `db.py` in the CLI commands
* [x] add another `CLI` test for the `upgrade` command
* [x] update `admin.py`

### Steps to Confirm

* [ ] `nox -s dev` & `fides db upgrade`

### Pre-Merge Checklist

* [x] Update `CHANGELOG.md`

### Description Of Changes

In a recent change I made to make automigration optional, I did some renaming of the `init` CLI command and API endpoints. Upon further reflection, I realized there may be a point where we want to start supporting downgrades, as this is a great feature for users in case of a bad upgrade.

Given this fact, I want to preemptively fix this since the feature is still new and users should still be using `init`. Eventually we can now add a `fides db downgrade`


src/fides/cli/commands/db.py
```diff
@@ -21,7 +21,9 @@ def database(ctx: click.Context) -> None:
 @with_analytics
 def db_init(ctx: click.Context) -> None:
     """
-    Initialize the Fides database.
+    Runs all upgrade migrations for the Fides database.
+
+    Will also automatically initialize a fresh database.
 
     **WARNING**: Deprecated, use `migrate` instead.
```
SteveDMurphy, line 28:
```suggestion
    **WARNING**: Deprecated, use `upgrade` instead.
```

SteveDMurphy (approved):
Looks good! Nice future planning 👍🏽 
One nit in the `db_init` docs block is all

ThomasLaPiana:
> Looks good! Nice future planning 👍🏽 One nit in the `db_init` docs block is all

nice catch!

PR #3343, opened 2023-05-22
Bump Fideslang Version

### LINKED ISSUE #3334: Bump Fideslang to v1.4
This should happen once [this issue](https://github.com/ethyca/fideslang/issues/104) is closed

ThomasLaPiana:
Closes #3334 

### Code Changes

* [x] add upgrade migration
  * [ ] ~~update `ViolationAttributes` (part of `Evaluations`)~~ - There is no fundamental relationship here, and since this is a historical artifact I don't think we need to update it. We should leave it as-is since it is a record of how the system _was_
  * [x] Update `PolicyRules` (part of `Policy`, stored as JSON in the database)
  * [x] Update `PrivacyDeclaration` (part of `System` but has its own database table)
  * [ ] ~~`PrivacyNotice`~~
  * [ ] ~~`PrivacyNoticeHistory`~~
  * [ ] ~~`Consent`~~
* [x] add downgrade migration
  * [ ] ~~update `ViolationAttributes` (part of `Evaluations`)~~ - There is no fundamental relationship here, and since this is a historical artifact I don't think we need to update it. We should leave it as-is since it is a record of how the system _was_
  * [x] Update `PolicyRules` (part of `Policy`, stored as JSON in the database)
  * [x] Update `PrivacyDeclaration` (part of `System` but has its own database table)
  * [ ] ~~`PrivacyNotice`~~
  * [ ] ~~`PrivacyNoticeHistory`~~
  * [ ] ~~`Consent`~~
* [x] update `requirements.txt`
* [x] fix tests

### Steps to Confirm

* [ ] `git checkout main; nox -s teardown -- volumes; nox -s dev` - let the server get into a running state before the next commands
* [ ] `fides user login; fides push; fides get system privacy_request_fulfillment` - confirm that the expected data use says `provide.service`
* [ ] spin down the docker containers (but _do not teardown the volumes!_)
* [ ] `git checkout ThomasLaPiana-update-fideslang-1-4; nox -s dev`
* [ ] `fides user login;  fides get system privacy_request_fulfillment` - confirm that the new data use (`essential.service`) has replaced the old (`provide.service)`

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`
* [ ] make sure migrations are in proper order

### Description Of Changes

This PR is designed to run in-place data migrations after the upgrade to `fideslang` version `1.4`, which solely affects data uses. This isn't intended to be a solution for _every_ use case, as it won't help people using only YAML files for instance, but the database migrations should take care of most people who can then use `fides pull` to get their local files up-to-date.


ThomasLaPiana:
I'm thinking through the implementation here and I'm not really sure if we have an elegant way of doing this...

We still use JSON in the database, and I don't know of a way to recurse through the database JSON fields to catch all of the deeply nested instances of these data uses. Flat objects should be fine, but particularly `datasets` is where I foresee us having problems.

A potential solution here would be to pull everything out as YAML, do the updates there, and then load it back into the database as an upsert.

The things we'll need to update are:
* data uses (actually these will get updated automatically when the new taxonomy is seeded, we'll just need to make sure to remove the old ones)
* datasets
* systems (privacy declarations)
* policies (ctl version)
* I might be missing other's but these should be the major ones...all of these support YAML

src/fides/api/ctl/migrations/versions/63b482f5b49b_upgrade_to_fideslang_1_4.py
```diff
@@ -0,0 +1,93 @@
...
+        assert not all(
+            [
+                False if rule.data_use in data_use_updates.keys() else True
+                for resource in existing_resources
+                for rule in resource
+            ]
+        ), "ERROR: Data Use Migration for Fideslang v1.4 failed for Policy Rules!"
+
+
+def upgrade():
```
ThomasLaPiana, line 84: @adamsachs @SteveDMurphy looking for an early review here to double-check my thinking for this data use migration

Does this method make sense? Can you think of any other pattern that might work better?

I haven't finalized the change over in `fideslang` but I was planning on using this design to do the in-place database upgrade. Luckily data uses aren't nearly as widely used as categories so there aren't many places that need updating here

src/fides/api/ctl/migrations/versions/63b482f5b49b_upgrade_to_fideslang_1_4.py
```diff
@@ -0,0 +1,93 @@
...
+            current_data_use = resource.data_use
+
+            if current_data_use in data_use_updates.keys():
+                resource.data_use = data_use_updates[resource.data_use]
+
+        await upsert_resources(PrivacyDeclaration, resources, session)
+
+        # Since we can't write tests for data migrations, we do it live
+        existing_resources = await list_resource(PrivacyDeclaration, session)
+        assert not all(
```
adamsachs, line 55: definitely could be reading this wrong, but i am thinking we'd want to remove the `not` here? i.e. you want `all` elements to be `True`, i.e. no more resources have a `data_use` that exists in `data_use_updates.keys()`.

src/fides/api/ctl/migrations/versions/63b482f5b49b_upgrade_to_fideslang_1_4.py
adamsachs: i like this! it's pretty easy and clean to follow. would you not want to do a reverse mapping for the downgrade?

also, i had a question on the `assert`, i feel like it may be slightly off? but i may be reading it wrong. generally i do like the extra safety in asserting what we expect here, that's a nice touch.

src/fides/api/ctl/migrations/versions/730a4f771dfa_.py
adamsachs, line 1: is this needed? i realize still a draft PR, so forgive me for commenting on more than what you asked 😄

src/fides/api/ctl/migrations/versions/63b482f5b49b_upgrade_to_fideslang_1_4.py
ThomasLaPiana: You're totally right here, and the way I wrote it was confusing to the point I confused myself...going to refactor this slightly

src/fides/api/ctl/migrations/versions/63b482f5b49b_upgrade_to_fideslang_1_4.py
ThomasLaPiana: Definitely going to add the reverse downgrade, but didn't want to get too deep since I'd use the same paradigm there!

Will add now and clean up the bug you found, thanks for the early and quick review!

src/fides/api/ctl/migrations/versions/730a4f771dfa_.py
ThomasLaPiana: this is a merge migration, notice the two `down_revisions` 🙂

ThomasLaPiana:
confirmed with @pattisdr that no changes are needed for `Consent`, `PrivacyNotice` and `PrivacyNoticeHistory`

[slack thread](https://ethyca.slack.com/archives/C04S7K0T2V7/p1685528163656159)

ThomasLaPiana:
> These won't be broken, since if the user is coming from an older version those old taxonomy items will still be there (we aren't purging them completely, just updating where they're used where it makes sense)
 are we going to run into issues when we try to serialize privacy requests, for example, with the old data uses?

@pattisdr Can you explain specifically what part of privacy requests would break?

These are the test cases I can think of:
1. 🔴 Someone is using YAML files exclusively, they upgrade `fides` and now they can't run local evaluations because the old data uses they're referencing don't exist in the default taxonomy
2. ✅ Someone is using the webserver exclusively, they upgrade and start the webserver, migrations run, all works as expected
3. ✅ Someone is using the YAML files _as well as_ the webserver. They upgrade and start the webserver, and they'll need to do a `fides pull` after the migration to make sure the YAML files match the webserver

I'm sure there are more though...I'd like to really think through these and figure out all possible cases. I'm sure we're going to break something for someone that we didn't think through properly so the more brains I have pondering this the better, I know I'll miss one (or more)

ThomasLaPiana:
very confused why the evaluation error message test is failing...these look identical to me:

```
'message':'','status':<StatusEnum.FAIL:'FAIL'>,'violations':[{'detail':'Declaration(SharePoliticalOpinions)of''system(customer_data_sharing_system)failed''rule(reject_targeted_marketing)frompolicy''(primary_privacy_policy).Violatedusageof''datacategories(user.political_opinion)with''qualifier''(aggregated.anonymized.unlinked_pseudonymized.pseudonymized.identified)''fordatauses''(marketing.advertising.third_party)and''subjects(customer)','violating_attributes':{'data_categories':['user.political_opinion'],'data_qualifier':'aggregated.anonymized.unlinked_pseudonymized.pseudonymized.identified','data_subjects':['customer'],'data_uses':['marketing.advertising.third_party']}}]}
'message':'','status':<StatusEnum.FAIL:'FAIL'>,'violations':[{'detail':'Declaration(SharePoliticalOpinions)of''system(customer_data_sharing_system)failed''rule(reject_targeted_marketing)frompolicy''(primary_privacy_policy).Violatedusageof''datacategories(user.political_opinion)with''qualifier''(aggregated.anonymized.unlinked_pseudonymized.pseudonymized.identified)''fordatauses''(marketing.adversiting.third_party)and''subjects(customer)','violating_attributes':{'data_categories':['user.political_opinion'],'data_qualifier':'aggregated.anonymized.unlinked_pseudonymized.pseudonymized.identified','data_subjects':['customer'],'data_uses':['marketing.advertising.third_party']}}]}
```

pattisdr:
> very confused why the evaluation error message test is failing...these look identical to me:

That one's just a typo looks like

ThomasLaPiana:
> > very confused why the evaluation error message test is failing...these look identical to me:
> 
> That one's just a typo looks like

Thank you so much...I completely missed it. Gotta be a better way for me to write that test, it was a pain to fix (past me is to blame)

src/fides/api/ctl/migrations/versions/730a4f771dfa_.py
pattisdr: Running into issues here though with the downgrade: I'm just testing your migration now, trying to think through edge cases. I'm going to get rid of the merge migration and just update the fides lang down_revision to point to the correct previous version instead if that's alright:

```
root@b19b85180f5f:/fides/src/fides/api/ctl# alembic downgrade -1
Alembic - ERROR [alembic.util.messaging] Ambiguous walk
  FAILED: Ambiguous walk
```


Expected order of migrations, most recent first:

- 730a4f771dfa merge migration.
- 63b482f5b49b fides lang migration.     * MERGE MIGRATION POINTS HERE
- 587c53fe3e99 add audit log resource table
- 8a71872089e4 add notice key to privacy notices
- 2661f31daffb tracking experiences on privacy preferences    * MERGE MIGRATION ALSO POINTS HERE
- b546ce845e6c  privacy experience config table 
- fc04e3e637c0 add dynamo db connnector
- ...

pattisdr:
> These won't be broken, since if the user is coming from an older version those old taxonomy items will still be there (we aren't purging them completely, just updating where they're used where it makes sense)

> are we going to run into issues when we try to serialize privacy requests, for example, with the old data uses?

> @pattisdr Can you explain specifically what part of privacy requests would break?

@ThomasLaPiana I was just concerned that Pydantic would fail validation on the response when we tried to return old data uses in old resources.  However, looking further it doesn't seem like we surface the consent preferences on the PrivacyRequest (which is good, this is really just for consent propagation).  Similarly, I'm not worried about the `Consent` resource (soon to be obsolete) because it doesn't look like we validate the data uses there in the request or the response, they're just required to be string types.

pattisdr:
I've also updated the default data uses in the `config.json` for the privacy center but that is causing some cypress tests to fail ^

ThomasLaPiana:
@pattisdr excellent, I was pretty sure we only validated data uses in specific cases but glad to hear it confirmed...

I wrote up test steps and confirmed your finding that the current migration method using asyncio doesn't work 🙁 so I'll merge in your branch and add downgrade steps as well

src/fides/api/ctl/migrations/versions/63b482f5b49b_upgrade_to_fideslang_1_4.py
```diff
@@ -0,0 +1,104 @@
...
+    "third_party_sharing.personalized_advertising": "marketing.advertising.third_party.targeted",
+    "provide": "essential",
+    "provide.service": "essential.service",
+}
+data_use_downgrades: Dict[str, str] = {
+    value: key for key, value in data_use_upgrades.items()
+}
+
+
+def update_privacy_declaration_data_uses(bind, data_use_map: Dict[str, str]):
```
ThomasLaPiana, line 42: @pattisdr these are super elegant ❤️ 🚀

SteveDMurphy:
Starting to review now @ThomasLaPiana and reminded me that we combined the `system_dependencies` work with `fideslang` version `1.4` - not sure if any coordination is required but figured I would mention here, there is a draft PR still (I need to look at again and catch up to `main` in #3285)

SteveDMurphy (approved):
So many test related updates needed! 😅 
Thanks for the helpful testing instructions, validated everything looked good and admired the way @pattisdr can provide clarity to a db migration 🔥 🔥

Dockerfile
```diff
@@ -12,6 +12,7 @@ ARG TARGETPLATFORM
 RUN apt-get update && \
     apt-get install -y --no-install-recommends \
     g++ \
+    git \
```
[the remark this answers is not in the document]
ThomasLaPiana: this is actually required, see the later down `git reset` command. I mistakenly removed this in a previous commit and realized the hard way!

docs/fides/docs/development/development_tips.md
```diff
@@ -100,6 +100,7 @@ The commands will need to be run inside a shell on your Docker containers, which
 In the `/src/fides/api/ctl` directory:
 
 - Migrate your database to the latest state: `alembic upgrade head`
+- Merge heads (for when you have conflicting heads from a merge/rebase): `alembic merge heads`
```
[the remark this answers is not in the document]
ThomasLaPiana: for sure. It can get weird when it comes to downgrades but we don't actually support those anyway technically so 🤷

pattisdr:
I can get this updated with main again ^ more migrations to reconcile

adamsachs (approved):
everything looks fine to me!

must have been pretty tedious, props to getting thru all of this while making some additional improvements along the way. and thanks for taking extra care around the migration of existing data!

PR #3355, opened 2023-05-22
respect notification_service_type config property in email connectors

### LINKED ISSUE #3354: Email connector should respect `notifications.notification_service_type` for messaging service

Currently, our email connectors [use a hardcoded preference hierarchy](https://github.com/ethyca/fides/blob/2.13.1b1/src/fides/api/service/connectors/base_email_connector.py#L66) to determine which messaging service provider to use, if more than one happens to be configured.

This hardcoded logic is inflexible and leaves no room for configuring a preferred messaging provider on a per-deployment basis. It also creates inconsistencies between how we determine the messaging provider for system-wide "notifications" (e.g. identity verification, request approval, error notifications, etc.), which contributed to the need to hide the UI that handled a system-wide messaging configuration.

We should make a simple update to have the email connector code path respect the `notifications.notification_service_type` config property (accessed thru a `ConfigProxy` to account for API updates), in the same way that it's done for our other code paths in the application. We could fall back to the current hardcoded "preference" logic if the property has not been set, mainly for backward compatibility/reducing unnecessary change.

As a longer-term goal, we'd probably like to configure a messaging provider on a per-email-connector basis, to allow for the most optionality/configuration. That'll be a slightly more complex implementation task, but should be pretty achievable. In the meantime, I think this should be a helpful increment in getting us toward a more flexible and consistent user experience.

### Additional context

@galvana and i were revisiting some of the email connector code (for a different purpose), which got me thinking about this improvement. improvement of this hardcoded logic had been discussed by me (@adamsachs), @rsilvery and @seanpreston when talking about the drawbacks of the messaging config UI we'd developed a couple of months back.

There's a really narrow edge case where this change could cause a problem. If an organization has >1 messaging provider defined and they have the `notifications.notification_service_type` property set to a provider that they do not want to use for email connector messaging, and instead are relying on the "hardcoded" preference logic for the email connector, then their email connector would now start using the messaging provider specified by the `notifications.notification_service_type`. this seems unlikely, but it's worth calling out.

Another consequence that's probably more likely/relevant: if/when we do re-enable a system-wide messaging config UI, assuming it updates the `notifications.notification_service_type` setting based on user selections, then we'll need to be clear that the change in the UI will also apply to email connectors.

adamsachs:
Closes #3354 

### Code Changes

* in code path used by email connectors to dispatch emails, first check whether a `notification_service_type` property has been set to an email provider, and if it has, use that type of messaging provider
    * if it has not (property is either unset or is set to a non-email messaging provider), then we fall back to the hardcoded "preference" logic that existed previously

### Steps to Confirm

* setup an invalid twilio messaging config and a valid mailgun messaging config
* set the `notifications.notification_service_type` config variable to `mailgun` (in e.g. `fides.toml`)
* set up an email connector (either consent or erasure)
* confirm email gets sent successfully - if so, then we've used the valid mailgun messaging config, which wouldn't have happened previously
* *note*: this may be hard to test since we currently only have harcoded email connectors for sovrn and attentive. soon we should have generic email connectors that allow us to send emails to arbitrary addresses, and that would make this easier to confirm

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

Check #3354 for more context/motivation/caveats...


tests/ops/service/connectors/test_base_email_connector.py
```diff
@@ -0,0 +1,55 @@
...
+)
+
+
+class TestBaseEmailConnector:
+    @pytest.mark.parametrize(
+        "fixtures,expected_value",
+        [
+            (
+                [],
+                MessagingServiceType.mailgun.value,  # mailgun is set to notification service type provider by autouse fixture
```
seanpreston, line 15: Thanks for the comments here!

seanpreston:
Thanks @adamsachs — this is a really helpful change

Issue #3352, opened 2023-05-22
Use `notice_key`s to drive key values in `fides-js`


allisonking:
### Is your feature request related to a specific problem?

After https://github.com/ethyca/fides/pull/3337 goes in, we will have access to a `notice_key` field for PrivacyNotices. Currently, [we use the privacy notice ID instead](https://github.com/ethyca/fides/blob/main/clients/fides-js/src/lib/preferences.ts#L36) 

### Describe the solution you'd like

Swap this to use `notice_key` instead. Will also need to rework the logic which figures out which notice IDs are enabled

### Describe alternatives you've considered, if any

_A description of any alternative solutions or features you've considered._

### Additional context

Note that these keys are purposefully not unique, so that we can stay flexible across regions. This means it is possible that we could have multiple `notice_key`s that are the same, though in theory, this shouldn't happen for one experience.

eastandwestwind:
Will be covered as part of https://github.com/ethyca/fides/pull/3361

Roger-Ethyca:
moving to done

Issue #3261, opened 2023-05-09
Generate a fides device id when Fides.js is initialized


eastandwestwind:
* Generate a fides device id when Fides.js is initialized, if it doesn't already exist
* Save device id to fides cookie

eastandwestwind:
Will be complete as part of https://github.com/ethyca/fides/pull/3361

Roger-Ethyca:
moving to done

PR #3361, opened 2023-05-24
3159 Add calls to Fides API from fides-js

eastandwestwind:
Closes https://github.com/ethyca/fides/issues/3159 and https://github.com/ethyca/fides/issues/3377 and https://github.com/ethyca/fides/issues/3261 and https://github.com/ethyca/fides/issues/3352

### Remaining TODOs:
* [x] Fix broken e2e tests after recent refactor
* [x] Write e2e test for generating a fides device id when Fides.js is initialized

### Follow-up tickets created:
- Refactor: give the detailed UserGeolocation object to overlay component- https://github.com/ethyca/fides/issues/3398

### Code Changes

* [ ] Use real API calls to Fides for GET privacy experiences and PATCH user preference
* [ ] Adds detailed e2e tests to mock different scenarios with API calls
* [ ] Refactor initialization of fides-js to handle privacy notice defaults
* [ ] Fixes [broken privacy-center env logic](https://github.com/ethyca/fides/pull/3361/commits/51fee9b75e74e3439c064ebbc1a75727f2a089b4#diff-be59d16bce2fb406235ed12d63487dd30c01524f6988e233bbc71de845e25a93R246-R255)

### Steps to Confirm

* [ ] Get fides-plus server up and running
* [ ] Add experiences / notices for your specific region (.e.g `us_ny`), ensuring you use an Overlay config
* [ ] Add the following env file to `clients/privacy-center`:
```
FIDES_PRIVACY_CENTER__GEOLOCATION_API_URL=[ask me for a geolocation url]
FIDES_PRIVACY_CENTER__IS_OVERLAY_DISABLED=false
```
* [ ] `cd clients/privacy-center`
* [ ] `turbo run dev`
* [ ] Navigate to `http://localhost:3000/fides-js-demo.html`
* [ ] Consent banner should match your configs
* [ ] Save your preferences by either using the banner or the modal
* [ ] Reload page
* [ ] Banner should not pop up, since no net new notices were added
* [ ] Using Postman to GET `{{host}}/historical-privacy-preferences` should surface your preferences with the notices
* [ ] Clear cookies to replicate a new user (no device id)

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes


clients/privacy-center/cypress/e2e/consent-banner.cy.ts
```diff
@@ -188,11 +215,44 @@ describe("Consent banner", () => {
...
+            expect(body.privacy_experience_history_id).to.eql(
+              expected.privacy_experience_history_id
+            );
+            expect(body.user_geography).to.eql(expected.user_geography);
+            expect(body.method).to.eql(expected.method);
+          });
+
           // check that the cookie updated
           cy.waitUntilCookieExists(CONSENT_COOKIE_NAME).then(() => {
             cy.getCookie(CONSENT_COOKIE_NAME).then((cookie) => {
```
allisonking, line 258: might be worth adding a check that the `fides_user_device_id` saved to the cookie is the one that was in the request

allisonking, line 258: might be worth adding a check that the `fides_user_device_id` saved to the cookie is the one that was in the request

allisonking (commented):
this is awesome, great work @eastandwestwind ! I won't approve yet since it hasn't been manually tested fully, but things are looking good 😎

clients/fides-js/src/fides.ts
```diff
@@ -120,15 +180,58 @@ const init = async ({
   geolocation,
   options,
 }: FidesConfig) => {
+  const { fides_user_device_id: fidesUserDeviceId } =
+    getOrMakeFidesCookie().identity;
```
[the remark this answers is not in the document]
NevilleS: I think this means we need to rethink a bit how we initialize that cookie here. The `getOrMakeFidesCookie()` function is a two-parter (the hint is in the name!) that will either:
* get the cookie, if it exists
* make a new cookie based on the consent defaults

Given that we can't rely on having the consent defaults at startup anymore, we likely need to split this up into something more like:
```typescript
// Read existing cookie, if one exists
let fidesCookie = getExistingFidesCookie();

// Generate a device ID (or use the one from the Fides cookie)
const fidesUserDeviceID = // ...

// Load the existing user preferences
const existingPreferences = fidesCookie?.consent;

// Geolocate the local device (or use the provided geolocation)
// ...

// Fetch the experience (or use the provided experience)
// ...

// Initialize the overlay (or not)
// ...

// Apply GPC preferences
// ...

// Initialize _Fides object
// ...

// Save the Fides cookie
saveFidesCookie({ ... });

// Emit an "onInit" event (or similar)
// ...

// Done!
```

clients/fides-js/src/lib/consent-types.ts
```diff
@@ -138,16 +140,20 @@ export enum ButtonType {
   TERTIARY = "tertiary",
 }
 
-export type PrivacyPreferencesCreateWithCode = {
-  // TODO: update this schema
+export enum ConsentMethod {
+  button = "button",
+  gpc = "gpc",
+  individual_notice = "api",
```
NevilleS, line 146: Why the rename?

clients/fides-js/src/lib/preferences.ts
```diff
@@ -59,13 +67,21 @@ export const updateConsentPreferences = ({
...
-  const privacyPreferenceCreate: PrivacyPreferencesCreateWithCode = {
+  const privacyPreferenceCreate: PrivacyPreferencesRequest = {
     browser_identity: cookie.identity,
     preferences: fidesUserPreferences,
+    privacy_experience_history_id: experienceHistoryId,
+    user_geography: userLocationString,
+    method: consentMethod,
   };
-  patchUserPreferenceToFidesServer(privacyPreferenceCreate, debug);
+  patchUserPreferenceToFidesServer(
```
NevilleS, line 79: What happens if this fails - like a 500 error?

We should still continue and save the cookie and update the object locally in that case so that everything works for the user.

clients/fides-js/src/services/external/geolocation.ts
```diff
@@ -7,15 +7,15 @@ import { debugLog } from "../../lib/consent-utils";
 export const getGeolocation = async (
   geolocationApiUrl?: string,
   debug: boolean = false
-): Promise<UserGeolocation> => {
+): Promise<UserGeolocation | null> => {
```
NevilleS, line 10: Yeah, that's more explicit 👌

clients/privacy-center/cypress/support/commands.ts
```diff
@@ -27,6 +27,18 @@ Cypress.Commands.add("waitUntilCookieExists", (cookieName: string, ...args) => {
...
 });
 
+Cypress.Commands.add("waitUntilFidesInitialized", (...args) => {
+  cy.waitUntil(
+    () =>
+      cy
+        .window()
+        .its("Fides")
+        .its("initialized")
+        .then(() => true),
```
NevilleS, line 37: Good example of why we need a `onInit` event 😄

clients/privacy-center/public/fides-js-components-demo.html
```diff
@@ -170,34 +176,40 @@ <h2>Consent Options</h2>
...
-      );
-
-      // Test behavior of integrations that can be configured without/before their platform scripts.
-      Fides.gtm();
-      Fides.meta({
-        consent: Fides.consent.tracking,
-        dataUse: Fides.consent.data_sales,
-      });
+      // Ensure we only render the consent config to the DOM once Fides is fully initialized
+      let timerId = setInterval(() => {
```
NevilleS, line 180: Can you add a `DEFER: ...` for the ticket related to the JS events API? We should tidy up this timeout when that is ready

clients/privacy-center/cypress/e2e/consent-banner.cy.ts
```diff
@@ -53,39 +55,99 @@ const stubConfig = ({
           ? undefined
           : Object.assign(config.options, options),
     };
-    if (typeof options !== "string" && options?.geolocationApiUrl) {
-      cy.intercept("GET", options.geolocationApiUrl, {
+    if (
+      typeof updatedConfig.options !== "string" &&
+      updatedConfig.options?.geolocationApiUrl
+    ) {
```
NevilleS, line 61: Why is this so conditional and we don't just always stub this API?

Likely deserves a explanatory comment at least

clients/privacy-center/cypress/e2e/consent-banner.cy.ts
```diff
@@ -53,39 +55,99 @@ const stubConfig = ({
...
+      cy.intercept(
+        "GET",
+        updatedConfig.options.geolocationApiUrl,
+        geoLocationResp
+      ).as("getGeolocation");
+    }
+    if (
+      typeof updatedConfig.options !== "string" &&
+      updatedConfig.options?.fidesApiUrl
+    ) {
```
NevilleS, line 79: Same here

clients/privacy-center/cypress/e2e/consent-banner.cy.ts
```diff
@@ -53,39 +55,99 @@ const stubConfig = ({
...
+      ).as("patchPrivacyPreference");
     }
     cy.visitConsentDemo(updatedConfig);
   });
 };
 
-const PRIVACY_NOTICE_ID_1 = "pri_4bed96d0-b9e3-4596-a807-26b783836374";
-const PRIVACY_NOTICE_ID_2 = "pri_4bed96d0-b9e3-4596-a807-26b783836375";
+const PRIVACY_NOTICE_KEY_1 = "advertising";
+const PRIVACY_NOTICE_KEY_2 = "essential";
```
NevilleS, line 100: ✨

NevilleS (changes requested):
Looking really good here. Test coverage is excellent - that should make my nitpicky recommendations on some refactoring and cleanup easier to do with confidence :)

Note that I think there's at least one subtle bug here related to the device ID that might be generated twice which could desync some stuff

NevilleS (changes requested):
Looking really good here. Test coverage is excellent - that should make my nitpicky recommendations on some refactoring and cleanup easier to do with confidence :)

Note that I think there's at least one subtle bug here related to the device ID that might be generated twice which could desync some stuff

clients/fides-js/src/fides.ts
```diff
@@ -98,19 +99,78 @@ let _Fides: Fides;
...
+    debugLog(
+      options.debug,
+      "No experience config found with for experience. Skipping overlay initialization."
+    );
+    return false;
+  }
+  // Check if there are any notices within the experience that do not have a user preference
+  return effectiveExperience.privacy_notices.some(
+    (notice) => notice.current_preference == null
+  );
```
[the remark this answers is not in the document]
eastandwestwind: The backend surfaces all notices applicable for the location, even if the user has previously consented to it. We need to ensure we don't re-render the banner if we have a `current_preference` for all notices

clients/fides-js/src/fides.ts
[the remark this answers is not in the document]
eastandwestwind: I've updated `getOrMakeFidesCookie` to take an explicit `device_id` param, which will use that instead of generating a new one. So I do this the 2nd time I call the method on L:229:

```
// Load any existing user preferences from the browser cookie
  const cookie = getOrMakeFidesCookie(
    consentDefaults,
    fidesUserDeviceId,
    options.debug
  );
```

Will that work, or you still think I should refactor as you mentioned above?

clients/fides-js/src/lib/consent-types.ts
eastandwestwind: I was using the incorrect type from the BE before, so I copied over the correct one here, using the same naming.

clients/fides-js/src/lib/preferences.ts
eastandwestwind: good point, I'll update

clients/privacy-center/cypress/support/commands.ts
eastandwestwind: yes this will help so much!

clients/fides-js/src/fides.ts
```diff
@@ -98,19 +99,78 @@ let _Fides: Fides;
...
+  if (!effectiveExperience) {
+    debugLog(
+      options.debug,
+      `No relevant experience found. Skipping overlay initialization.`
+    );
+    return false;
+  }
+  if (
+    !effectiveExperience.privacy_notices ||
+    effectiveExperience.privacy_notices.length === 0
```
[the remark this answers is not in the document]
eastandwestwind: Hmm, so `effectiveExperience.privacy_notices?.length` might be undefined, so I'm getting a TS error using that. 

I can instead reverse my logic to make it slightly more readable(?):
```
!(effectiveExperience.privacy_notices &&
    effectiveExperience.privacy_notices.length >= 0)
```

clients/fides-js/src/components/Overlay.tsx
```diff
@@ -1,25 +1,27 @@
...
   experience: PrivacyExperience;
-  geolocation?: UserGeolocation;
+  fidesRegionString: string;
 }
 
-const Overlay: FunctionComponent<OverlayProps> = ({ experience }) => {
+const Overlay: FunctionComponent<OverlayProps> = ({
+  experience,
+  options,
+  fidesRegionString,
```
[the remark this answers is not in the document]
eastandwestwind: ah, you're totally right. I'll make a follow-up to refactor- https://github.com/ethyca/fides/issues/3398

clients/privacy-center/cypress/e2e/consent-banner.cy.ts
eastandwestwind: you're right, we can always stub these APIs

eastandwestwind: Added a comment - `we conditionally stub these APIs because we need the exact API urls, which can change or not even exist in the specific test case.`

clients/fides-js/src/lib/preferences.ts
eastandwestwind: This method will always resolve, even if we don't have a 200 resp:

```
if (!response.ok) {
    debugLog(
      debug,
      "Error patching user preference Fides API. Response:",
      response
    );
  }
  return Promise.resolve();
```

clients/fides-js/src/fides.ts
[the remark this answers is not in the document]
NevilleS: Ah, ok... that definitely feels separate from `experienceIsValid` then, and would instead factor into a decision of whether or not to show the banner. I'd move this check into that part of the code instead 👍

clients/fides-js/src/fides.ts
```diff
@@ -120,15 +182,58 @@ const init = async ({
...
+      experience: effectiveExperience,
+      fidesRegionString,
+      options,
+    }).catch(() => {});
+  }
+
   // Configure the default consent values
   const context = getConsentContext();
   const consentDefaults = makeConsentDefaults({
+    experience,
```
NevilleS, line 226: I understand why you've done this, to mirror the existing flow, but to me this smells feels really "late" to be setting defaults on the cookie - after all, those defaults should also be what's driving the look & feel for the banner.

I'd recommend taking another pass at this without editing the `makeConsentDefaults` function at all, and instead just modifying `cookie.consent` with the defaults provided from the experience data - it just feels less likely to result in an unexpected issue here...

clients/fides-js/src/lib/preferences.ts
```diff
@@ -20,52 +19,55 @@ import { patchUserPreferenceToFidesServer } from "../services/fides/api";
...
-  privacyNotices: PrivacyNotice[];
-  enabledPrivacyNoticeIds: Array<PrivacyNotice["id"]>;
+  consentPreferencesToSave: Array<SaveConsentPreference>;
+  experienceHistoryId: string;
+  fidesApiUrl: string;
+  consentMethod: ConsentMethod;
+  userLocationString: string;
   debug?: boolean;
 }) => {
   const cookie = getOrMakeFidesCookie();
```
NevilleS, line 36: aside: this method reading from the cookie makes me mildly nervous as well. We're basically creating a shared global variable but not being careful with it's lifecycle - definitely think there's probably a bug here too.

NevilleS (changes requested):
Another round of comments here - largely this is just my instinct kicking in and detecting that there is likely some logic bugs here and I'd like to try to structure the code to keep us safe - let's pair on it a bit

clients/fides-js/src/lib/preferences.ts
eastandwestwind: fixed by passing in the `cookie` obj to this method

clients/fides-js/src/fides.ts
eastandwestwind: Updated to only `getOrMakeFidesCookie` once with the legacy values, then overwriting `cookie.consent` if we have new experiences.

NevilleS (approved):
Overall I think this looks good!

I think the _main_ risk I'm worried about is unexpected bugs for existing usage of the legacy-style `consentOptions` configuration with `config.json`, etc. That'll come from some testing.

clients/fides-js/src/fides.ts
```diff
@@ -121,112 +181,86 @@ const init = async ({
   geolocation,
   options,
 }: FidesConfig) => {
-  // Configure the default consent values
+  if (!validateOptions(options)) {
+    debugLog(options.debug, "Invalid overlay options", options);
+    return;
```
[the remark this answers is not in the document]
eastandwestwind: ah you're totally right, good catch!

clients/fides-js/src/fides.ts
```diff
@@ -121,112 +181,86 @@ const init = async ({
...
+  } else if (!effectiveExperience) {
     effectiveExperience = await fetchExperience(
-      userLocationString,
+      fidesRegionString,
+      options.fidesApiUrl,
+      cookie.identity.fides_user_device_id,
+      options.debug
+    );
+  }
+  if (effectiveExperience) {
```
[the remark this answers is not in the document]
eastandwestwind: yes that would make sense to me

Issue #3320, opened 2023-05-17
Update privacy experience forms with new requirements


allisonking:
### Is your feature request related to a specific problem?

As we iterate on privacy experience forms, there are some new requirements to properly drive the designs for the `fides-js` side.

### Describe the solution you'd like

Here's a sketch of some of the updates needed, where the new updates are in purple
![image](https://github.com/ethyca/fides/assets/24641006/13a96e20-0699-4a9f-b792-4f93232b73a3)

* [ ] New "Privacy policy" block to each form type
  * [ ] Need backend columns for this new field
* [ ] Add "Actions" block to privacy center
* [ ] Add "Messaging" block to all overlay forms
* [ ] Add "Manage preferences link label" to overlay banner form
* [ ] Move the location fields to the bottom of the form

### Describe alternatives you've considered, if any

_A description of any alternative solutions or features you've considered._

### Additional context

https://ethyca.slack.com/archives/C04S7K0T2V7/p1684181017234309

We may also get updated Figma designs for this

allisonking:
Blocked while experience model is being discussed: https://ethyca.atlassian.net/wiki/spaces/EN/pages/2697166893/Handling+Error+Conditions+for+Experience+Configuration+May+2023

mfbrown:
@allisonking Designs: https://www.figma.com/file/dmEwdK3xZwjKfGVQ9t66xe/Fides-v.2-Master-Working-Files?type=design&node-id=6980-351576&t=Q8vzrc7cfau41H2J-4

allisonking:
Sure, that sounds good!

Roger-Ethyca:
moving to done

Issue #3305, opened 2023-05-15
[Backend] Create Default Experience Configs 


pattisdr:
### Is your feature request related to a specific problem?

With https://github.com/ethyca/fides/pull/3292,  we start creating PrivacyExperiences for the customer when they edit Privacy Notices or ExperienceConfigs.  Sometimes this means I need to create or update a Privacy Experience that doesn't have an Experience Config, meaning there isn't any language for that experience (no banner text, button labels, etc.) and invalid to be shown to the user. 

I've added filters to surface to endpoints so we can surface to the customer which regions need to be attached to a config, but in the interim, users won't have valid experiences.

### Describe the solution you'd like

It would be nice to have **three default experience configs** a 1) Banner Overlay a 2) Link Overlay and 3) Privacy Center Overlay that we can link an experience to by default.

### Describe alternatives you've considered, if any

_A description of any alternative solutions or features you've considered._

### Additional context

_Add any other context or screenshots about the feature request here._

mfbrown:
@pattisdr Updated UI for reference: https://www.figma.com/file/dmEwdK3xZwjKfGVQ9t66xe/Fides-v.2-Master-Working-Files?type=design&node-id=6980-351576&t=Q8vzrc7cfau41H2J-4

Roger-Ethyca:
moving to done

PR #3387, opened 2023-05-26
Remove Experience Config Delivery Mechanism [#3322]

### LINKED ISSUE #3322: [Backend] Update PrivacyExperienceConfig fields to support new requirements

[See design doc here](https://ethyca.atlassian.net/wiki/spaces/EN/pages/2697166893/Handling+Error+Conditions+for+Experience+Configuration+May+2023)

- Removing concept of delivery mechanism, replacing with banner_enabled field
- Removing banner_title and banner_description in favor of component_title and component_description
- Rename link_label to open_modal_link
- Add privacy_policy_link
- Add privacy_policy_url
- Remove validation on how notices must be delivered in a banner

Related to https://github.com/ethyca/fides/issues/3320

### Additional context

pattisdr:
Closes #3322 

❗ Contains migration; check downrev before merging
❗ Blocked by frontend ticket https://github.com/ethyca/fides/issues/3320

### Vocabulary

_Privacy Notice:_   The entity that a user accepts or rejects in terms of consent.  A notice encapsulates a name, consent mechanism, data uses, and potentially many regions.  It is not linked to Privacy Experiences via FK.  Privacy Experiences are potentially created behind the scenes when Privacy Notices are created/updated to make sure we always have experiences to present the given notices.

_Privacy Experience:_  Governs how privacy notices are presented to the end user for a given region. Represents the Experience for a single region and component (overlay or privacy center).  Has a FK to Experience Config.  At runtime, determines what Privacy Notices are relevant for a particular experience and returns.  Components like the privacy center, banners, and overlays should call the Privacy Experience API.

_Experience Config:_  Holds the copy for an experience, such as the overlay's title or button labels.  This ExperienceConfig can be shared between multiple Experiences.  The Admin UI queries the Experience Configs so the customer can set these up.  Experiences are updated behind the scenes when Experience Config updates are made.

### Description Of Changes

Building on the [previous data model changes](https://github.com/ethyca/fides/pull/3292) to [keep Privacy Notices in sync with Privacy Experiences](https://ethyca.atlassian.net/wiki/spaces/EN/pages/2683830495/2023-05-06+Backend+Spike+Keeping+Notices+and+Experiences+in+Sync), this PR simplifies the constraints.

Follows what @NevilleS [proposes here](https://ethyca.atlassian.net/wiki/spaces/EN/pages/2697166893/Handling+Error+Conditions+for+Experience+Configuration+May+2023), primarily by removing `ExperienceConfig.delivery_mechanism` and all associated validation.  Previously, PrivacyNotices that had `opt_in` or `notice_only` consent mechanisms had to be presented in an overlay that was delivered by a **banner** (field stored on the PrivacyExperienceConfig).  There was a lot of behind the scenes activity to ensure that we always kept our PrivacyExperiences in the correct state, when both PrivacyNotices and PrivacyExperienceConfigs were updated. We remove this here and instead surface a property calculated at runtime for the frontend to determine if they need to show a banner.

Similarly makes the related changes to update the ExperienceConfig with the [new fields per updated requirements](https://www.figma.com/file/dmEwdK3xZwjKfGVQ9t66xe/Fides-v.2-Master-Working-Files?type=design&node-id=5115-139765&t=GsLD1WwrrU5ZmeGS-0)

### Code Changes

* [ ] In the GET Privacy Experience List endpoint, surface a new field `PrivacyExperience.show_banner` which is calculated at runtime as to whether or not the given experience needs to be triggered via a banner
* [ ] Remove checking associated privacy notice consent mechanisms when updating or creating ExperienceConfigs to make sure there's not a delivery mechanism mismatch.  Reduces the links between the two.
    * [ ] Remove automatically updating Experiences behind the scenes when PrivacyNotices were adjusted in a way that forced the delivery_mechanism to change
* [ ] Add/rename many fields on the `ExperienceConfig` table to reflect the new mockups.
* [ ] Allow ExperienceConfigs to be updated _without_ passing in a regions key, to support the first release, where the copy can be updated, but the regions can't.  If no regions key supplied, Experiences that are linked to the ExperienceConfig won't be touched.
    * [ ] Add new validation that prevents `ExperienceConfig.component` from being updated once it's created to avoid a bunch of extra backend messiness.
* [ ] Bug fix that changes when we validate `PrivacyNotice` component and `displayed_in_*` fields - there were edge cases where we would be validating too early.  We really need the existing PrivacyNotice object to already be loaded before we validate.

### Steps to Confirm

* [ ] POST `{{host}}/privacy-notice` to create some notices
* [ ] GET  `{{host}}/privacy-experience/` to verify Privacy Experiences still created behind the scenes when the Privacy Experiences were created.  For now `experience_config` should be null.  Verify that `show_banner` is True if there is a notice under `privacy_notices` that has a `consent_mechanism` of `opt_in` or `notice_only`.  `privacy_center` notices however, always have `show_banner` = False
* [ ] POST `{{host}}/experience-config` where your request body has `component`=overlay, and use a region that matches one of your regions in experiences
* [ ] `GET {{host}}/privacy-experience/` and verify that `experience_config` is no longer null for the matching overlay type and region.
* [ ] Do a PATCH {{host}}/experience-config/{{experience-config-id}} to an experience config that was associated with an experience whose "show_banner" was True.  In the request body, set `banner_enabled`: `always_disabled`.  Verify that this should flip `show_banner` to False, for Experiences it is attached to (`GET {{host}}/privacy-experience/` to verify 

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`


src/fides/api/models/privacy_experience.py
```diff
@@ -28,32 +28,32 @@ class ComponentType(Enum):
...
     """
-    The delivery mechanism - not formalized in the db
+    Whether the banner should display
     """
 
-    banner = "banner"
-    link = "link"
+    always_enabled = "always_enabled"
+    enabled_where_required = "enabled_where_required"  # If the user's region has at least one opt-in or notice-only notice
+    always_disabled = "always_disabled"
```
pattisdr, line 38: A primary change of this PR; `delivery_mechanism` has been replaced in favor of `banner_enabled` which is similar in concept, however this construct further reduces the connections between a Notice and an Experience.    Previously, we did a lot of behind the scenes validation to make sure an experience's delivery_mechanism always matched its notices.  This new field doesn't have that validation, we don't check this at all, so it's possible `ExperienceConfig.banner_enabled` = `always_disabled` regardless of how the underlying notices should be displayed.

src/fides/api/models/privacy_experience.py
```diff
@@ -28,32 +28,32 @@ class ComponentType(Enum):
...
 
 
 class ExperienceConfigBase:
     """Base schema to share common experience config."""
 
-    acknowledgement_button_label = Column(String)
```
pattisdr, line 43: Changing to `acknowledge_button_label`

src/fides/api/models/privacy_experience.py
```diff
@@ -28,32 +28,32 @@ class ComponentType(Enum):
...
-    banner_title = Column(String)
-    banner_description = Column(String)
```
pattisdr, line 45: Seems like there was a simplification in the UI that we can just define a title and a description, and they can both be used for a banner title and an overlay title, so I'm removing the duplicate set of fields.

src/fides/api/models/privacy_experience.py
```diff
@@ -28,32 +28,32 @@ class ComponentType(Enum):
...
+    accept_button_label = Column(String)
+    acknowledge_button_label = Column(String)
+    banner_enabled = Column(EnumColumn(BannerEnabled), index=True)
     component = Column(EnumColumn(ComponentType), nullable=False, index=True)
-    component_title = Column(String)
-    component_description = Column(String)
```
pattisdr, line 48: Simplifying to title/description instead.

src/fides/api/models/privacy_experience.py
```diff
@@ -483,24 +493,10 @@ def new_experience_needed(
                 data={
                     "region": region,
                     "component": ComponentType.privacy_center,
-                    "delivery_mechanism": DeliveryMechanism.link,
                 },
             )
             added_experiences.append(privacy_center_experience)
 
-        # See if we need to update an existing Link Overlay to be delivered by Banner instead
-        if banner_delivery_update_needed(overlay_experience, overlay_notices):
```
pattisdr, line 492: Removing some of this validation and behind-the-scenes efforts now that this requirement has been removed, as I'm updating PrivacyExperiences to make sure we always have the appropriate rows to back PrivacyNotices.

src/fides/api/schemas/privacy_experience.py
```diff
@@ -16,20 +17,46 @@ class ExperienceConfigSchema(FidesSchema):
...
-    disabled: Optional[bool] = False
-    is_default: Optional[bool] = False
-    link_label: Optional[SafeStr]
-    reject_button_label: Optional[SafeStr]
+    Note component is intentionally not included here.  When creating component, you can add it, but when updating,
+    the component cannot be updated.
+    """
+
+    accept_button_label: Optional[SafeStr] = Field(
+        description="Overlay 'Accept button displayed on the Banner and Privacy Preferences' or Privacy Center 'Confirmation button label'"
```
pattisdr, line 26: I've tried to mark here how the fields are shared between the Overlay Experience Config and the Privacy Center Experience Config.

src/fides/api/schemas/privacy_experience.py
```diff
@@ -56,56 +86,45 @@ def validate_regions(
 
     @root_validator
     def validate_attributes(cls, values: Dict[str, Any]) -> Dict[str, Any]:
-        """Validate minimum set of required fields exist given the type of component and/or delivery_mechanism"""
+        """Validate minimum set of required fields exist given the type of component"""
         component: Optional[ComponentType] = values.get("component")
```
pattisdr, line 90: Updating the logic here with my understanding of what fields are required when saving an overlay.  Currently overlays and privacy center experience configs are stored in the same table, as I feel there are a lot of overlapping fields. The Privacy Center needs a subset of the Overlay fields.

src/fides/api/schemas/privacy_experience.py
```diff
@@ -56,56 +86,45 @@ def validate_regions(
...
-    regions: List[PrivacyNoticeRegion]
+    regions: Optional[List[PrivacyNoticeRegion]] = Field(
+        default=None,
+        description="If None, no edits will be made to regions.  If an empty list, all regions will be removed.",
+    )
+
+    class Config:
+        """Forbid extra values - specifically we don't want component to be updated here."""
+
+        extra = Extra.forbid
```
pattisdr, line 119: New piece of validation here but I think it makes sense to not allow an ExperienceConfig's component to be updated once it's created.  We either do this, or add more behind the scenes logic to unlink more experiences as the ExperienceConfig.component is updated.

src/fides/api/schemas/privacy_experience.py
```diff
@@ -56,56 +86,45 @@ def validate_regions(
...
     """
-    Updating ExperienceConfig - requires regions to be patched specifically
+    Updating ExperienceConfig. Note that component cannot be updated once its created
     """
 
-    regions: List[PrivacyNoticeRegion]
+    regions: Optional[List[PrivacyNoticeRegion]] = Field(
+        default=None,
+        description="If None, no edits will be made to regions.  If an empty list, all regions will be removed.",
+    )
```
pattisdr, line 114: As mentioned elsewhere, regions are no longer required in the request to update the ExperienceConfig. If you just want to update the copy and not the regions, don't include the regions key in the request at all.

src/fides/api/schemas/privacy_notice.py
```diff
@@ -61,12 +61,38 @@ def validate_data_uses(self, valid_data_uses: List[str]) -> None:
...
+        """
+        Generate the notice_key from the name if not supplied
+        """
+        if not values.get("notice_key"):
+            values["notice_key"] = PrivacyNoticeModel.generate_notice_key(
+                values.get("name")
+            )
+
+        return values
+
```
pattisdr, line 89: The diff here looks strange but I mostly moved the `validate_consent_mechanisms_and_display` under PrivacyNoticeCreation instead of the base `PrivacyNotice`

src/fides/api/api/v1/endpoints/privacy_experience_config_endpoints.py
```diff
@@ -256,14 +197,17 @@ def experience_config_update(
...
     """
     experience_config: PrivacyExperienceConfig = get_experience_config_or_error(
         db, experience_config_id
     )
     experience_config_data_dict: Dict = experience_config_data.dict(exclude_unset=True)
     # Pop the regions off the request
-    regions: List[PrivacyNoticeRegion] = experience_config_data_dict.pop("regions")
+    regions: Optional[List[PrivacyNoticeRegion]] = experience_config_data_dict.pop(
+        "regions", None
+    )
```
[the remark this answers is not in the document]
allisonking: ah cool, that means I can get rid of this bit in the FE 😄 https://github.com/ethyca/fides/blob/main/clients/admin-ui/src/features/privacy-experience/privacy-experience.slice.ts#L45-L47

src/fides/api/models/privacy_experience.py
allisonking: nice, that lets us not have to suffer from the question `acknowledgment` or `acknowledgement` 😄

src/fides/api/schemas/privacy_experience.py
allisonking: this was very helpful, thank you!

allisonking (commented):
nice job! was straightforward to integrate with 👍

src/fides/api/schemas/privacy_experience.py
```diff
@@ -39,73 +66,67 @@ class ExperienceConfigCreate(ExperienceConfigSchema):
...
 
 
 class ExperienceConfigUpdate(ExperienceConfigSchema):
     """
-    Updating ExperienceConfig - requires regions to be patched specifically
+    Updating ExperienceConfig. Note that component cannot be updated once its created
     """
 
-    regions: List[PrivacyNoticeRegion]
+    regions: Optional[List[PrivacyNoticeRegion]] = Field(
```
adamsachs, line 113: this now seems to be identical to what you've got in the `ExperienceConfigCreate` model, including the validator method. is there still a reason to have it split out here in the subclasses and not the base class? i see you've got a field `default` and `description` specified here that may not apply to the "create". still feels like there's some room to DRY this up a bit, perhaps?

src/fides/api/models/privacy_experience.py
```diff
@@ -215,6 +217,9 @@ class PrivacyExperience(PrivacyExperienceBase, Base):
     # related to experiences.
     privacy_notices: List[PrivacyNotice] = []
 
+    # Attribute that is cached on the PrivacyExperience object by "get_should_show_banner", calculated at runtime
+    show_banner: bool
```
adamsachs, line 221: somewhat related to my comment below about considering whether we still need this db table/model - it just struck me as a bit odd that this property is on the db model. i can see why you've done it, and i'm not saying we should change it now...just some food for thought.

src/fides/api/models/privacy_experience.py
```diff
@@ -510,58 +509,57 @@ def new_experience_needed(
...
+    experiences_to_unlink: List[PrivacyExperience] = experience_config.experiences.filter(  # type: ignore[call-arg]
+        PrivacyExperience.region.in_(regions_to_unlink)
+    ).all()
+
+    for experience in experiences_to_unlink:
+        experience.unlink_experience_config(db)
+    return [experience.region for experience in experiences_to_unlink]  # type: ignore[misc]
 
 
 def upsert_privacy_experiences_after_config_update(
```
adamsachs, line 537: yet another comment that i'm not suggesting we address in this PR 😄

as i wrap my head around all of this, part of me is starting to feel like the `PrivacyExperience` db model may be a bit superfluous? again, it makes total sense to have this as an API-level model - i like how simple it is as an API consumer to get this simple representation of experiences, all of which have been calculated by some relatively complex logic/maintenance on the BE. but do we get a lot of benefit from keeping this maintained in a db table?

i get the need to audit things and maintain a detailed history, so that could be one motivation -- but i still feel like the audit history could be pieced together by just the `PrivacyNoticeHistory` and `PrivacyExperienceConfigHistory` tables? we'd maybe need to provide some "utility" endpoints to make it easier on the consumer to piece that all together....

i guess what i get worried about here is the maintaining duplicate state across the `PrivacyExperience` and `PrivacyExperienceConfig` _and_ their associated history tables. i went down this line of thinking because it struck me as a bit funny that we're keeping a list of regions on the `PrivacyExperienceConfig` while _also_ maintaining links from `PrivacyExperience` records, whose main purpose is just to record the experience type for a given region (admittedly, that's a bit of an oversimplification). the method you've got here is pretty easy to follow, but we could potentially remove the need for it (and the `upsert_privacy_experiences_after_notice_update` function) and just rely on runtime lookups across the `PrivacyExperienceConfig` and `PrivacyNotice` tables to populate a `PrivacyExperience` API object that looks identical to what you have now - without maintaining a DB table (and associated history table).

i haven't totally thought this through! i can definitely see many good reasons to keep this in place. but it's a line of thinking i wanted to at least entertain as i am making my way through this. let me know what you think!

adamsachs (approved):
-- note: my review is scoped to the BE changes, i didn't take a look at any FE code changes (though i will try to do some manual testing!) -- 

this looks great @pattisdr! as far as i can tell, this iteration greatly simplifies things. after getting a bit of context to get back up to speed on the use case, i was able to follow just about everything that's going on here - which is a testament to how well you've structured the code, given the complex requirements.

i've really got only a few minor, nonblocking comments that may be worth acting upon in the scope of this PR - so i will preemptively approve this. i do have a few more meaty questions on the data structures/db models that i do not expect to be addressed in this iteration, but that i'd like to get your thinking on.

i haven't done any manual testing, but i'll look to spin that up a bit locally, if only to immerse myself a bit more in all the great work you and the rest of the team have put in place 😄 i'm not letting that hold up an approval here, since i'm assuming you and others have done enough of that manual testing to feel comfortable merging. if that's not the case, please let me know and i can be sure to do that testing before we decide to merge!

pattisdr:
thanks very much for taking the time to review @adamsachs ⭐

src/fides/api/models/privacy_experience.py
pattisdr: hahaha

src/fides/api/models/privacy_experience.py
```diff
@@ -28,39 +28,40 @@ class ComponentType(Enum):
...
     """
 
-    banner = "banner"
-    link = "link"
+    always_enabled = "always_enabled"
+    enabled_where_required = "enabled_where_required"  # If the user's region has at least one opt-in or notice-only notice
+    always_disabled = "always_disabled"
 
 
 class ExperienceConfigBase:
```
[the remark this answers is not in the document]
pattisdr: Thanks for your comments I think @NevilleS also would have stored in a JSON blob, so maybe it's two against one 😄 

My reasoning for storing fields separately:

- Originally experience fields had more "rules" on them, much more upfront validation needed
- Model has evolved a lot, was much more tightly linked to Notices previously. Right now it is more or less contains the "copy", but I'm not convinced that will hold for future iterations as we continue to debate this feature and roll out more functionality
- Easier to validate
- Unknown future filtering and search requirements.  Not a big deal to search json blobs for filtering in the admin ui but what about Consent Reporting search? If we end up needing to search privacypreferencehistory or currentprivacypreference for related experience information I'd rather have these columns separate so I can index.
- I likely would want to keep the jsonb fields clean regardless, so might end up writing migrations to update jsonb still.
- Easier to store with more structure initially and collapse into a json field later than the reverse

Anyway, if it settles out that the ExperienceConfig is truly just copy and we don't have performance concerns with search, that might be a reason to make this a jsonb field instead.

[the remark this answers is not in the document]
adamsachs: that's very good reasoning! thanks for spelling it out. totally good with leaving it how it is for now based on those considerations.

src/fides/api/models/privacy_experience.py
```diff
@@ -161,7 +162,7 @@ def experience_config_history_id(self) -> Optional[str]:
 
 
 class PrivacyExperienceConfigHistory(ExperienceConfigBase, Base):
```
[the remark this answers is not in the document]
pattisdr: Good callout, I've been copying and pasting your original method all over the place with only minor modifications. I'll dry that up now.

src/fides/api/models/privacy_experience.py
```diff
@@ -227,6 +232,39 @@ def privacy_experience_history_id(self) -> Optional[str]:
...
+        self, db: Session, show_disabled: Optional[bool] = True
+    ) -> bool:
+        """Returns True if this Experience should be delivered by a banner
+
+        Relevant privacy notices are queried at runtime.
+        """
+        if self.component != ComponentType.overlay:
+            return False
+
+        if self.experience_config:
```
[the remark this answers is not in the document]
pattisdr: This is a current expected state!  Another [ticket this sprint](https://github.com/ethyca/fides/issues/3305) will improve this by adding two default `ExperienceConfigs` to the db on startup, one overlay, and one privacy center, to which I can automatically link a `PrivacyExperience` on the user's behalf where needed.

I do think we'll need further product discussion around "what I should do automatically" versus where we should surface required actions needed to the customer:

Let me give you some examples:
- A customer creates a CA Privacy Notice that should be displayed in an "overlay", I create a CA overlay Privacy Experience in the background.  Right now there's no copy to link this to automatically.
- Another example from the other end, say I have an overlay Experience Config with VA on it.  Say the user removes VA. Now the VA Privacy Experience is not linked to any copy.

src/fides/api/models/privacy_experience.py
```diff
@@ -227,6 +232,39 @@ def privacy_experience_history_id(self) -> Optional[str]:
...
+        )
+        if show_disabled is False:
+            privacy_notice_query = privacy_notice_query.filter(
+                PrivacyNotice.disabled.is_(False)
+            )
+
+        return bool(
+            privacy_notice_query.filter(
+                PrivacyNotice.consent_mechanism.in_(
+                    [ConsentMechanism.notice_only, ConsentMechanism.opt_in]
```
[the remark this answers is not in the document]
pattisdr: What do you know, a `Set` does work with `in_`! Change made.

src/fides/api/models/privacy_experience.py
```diff
@@ -227,6 +232,39 @@ def privacy_experience_history_id(self) -> Optional[str]:
         ).first()
         return history.id if history else None
 
+    def get_should_show_banner(
```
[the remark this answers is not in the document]
pattisdr: Thank you! Readability is what I was going for!

src/fides/api/ctl/migrations/versions/deb97d9393f3_remove_delivery_mechanism.py
```diff
@@ -0,0 +1,284 @@
...
+        table_name="privacyexperienceconfig",
+    )
+    op.create_index(
+        "ix_privacyexperienceconfig_delivery_mechanism",
+        "privacyexperienceconfig",
+        ["delivery_mechanism"],
+        unique=False,
+    )
+
+    op.drop_column("privacyexperienceconfig", "title")
```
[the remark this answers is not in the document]
pattisdr: Unless you want to migrate your JSON blob too and keep things clean!

src/fides/api/schemas/privacy_experience.py
pattisdr: yes room for cleanup there! I'd originally treated regions differently for create and update in an earlier draft of this PR and later consolidated the behavior to reduce confusion.

src/fides/api/models/privacy_experience.py
pattisdr: To clarify one item up front, regions are _not_ duplicated across the `PrivacyExperienceConfig` and the `PrivacyExperience` models.  PrivacyExperience has a FK to PrivacyExperienceConfig but there is no corresponding list of regions on `PrivacyExperienceConfig`.   There is a `PrivacyExperienceConfig.regions` property but it's calculated from the PrivacyExperience FK relationship.

src/fides/api/models/privacy_experience.py
pattisdr: I do like how the Experience has the all the details for a single region.  `show_banner` really is a property of a single region.  It's looking at notices for that region, and then its ExperienceConfig via the FK to see if this specific region should surface a banner!

src/fides/api/ctl/migrations/versions/deb97d9393f3_remove_delivery_mechanism.py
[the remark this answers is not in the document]
adamsachs: good point, i was getting ahead of myself 😬

src/fides/api/schemas/privacy_experience.py
adamsachs: looks good, thanks for addressing!

src/fides/api/models/privacy_experience.py
adamsachs: > To clarify one item up front, regions are _not_ duplicated across the `PrivacyExperienceConfig` and the `PrivacyExperience` models. PrivacyExperience has a FK to PrivacyExperienceConfig but there is no corresponding list of regions on `PrivacyExperienceConfig`. There is a `PrivacyExperienceConfig.regions` property but it's calculated from the PrivacyExperience FK relationship.

aha, i think i missed this detail since that particular model attribute wasn't being changed in this PR - but it is an important detail that does settle the main source of my qualms, so thank you for recognizing my confusion and pointing this out!

> One of the main purposes of the Privacy Experience table at this point, in my mind, is to enforce that we only have one experience type per region at all times down to the database level.

this makes sense, and especially with the context you've provided on the logical nightmare that existed before (some of which is outside of this PR/discussion). i do think that at least some of the problems you were running into before are probably no longer as pertinent given the delivery mechanism removal you've got in this PR -- but i can certainly better see the motivation to have this sort of join/constraint table, given that changes from "both sides" (i.e. the `PrivacyNotice` and the `PrivacyExperienceConfig`) could prompt an update to what experience is shown for a given reason.

i do think that moving forward we need to be careful to _not_ duplicate data across the tables (i.e. `PrivacyNotice`, `PrivacyExperience` and `PrivacyExperienceConfig`). you likely have a much clearer picture on all of this than i do (still getting up to speed!), but given how interrelated these concepts are, i could imagine there being some confusion about what attributes belong where, as the data model evolves - so we'll need to take care when making updates (as you've done here!).

all that being said, i've been convinced that splitting the tables is a sensible and well-motivated choice. thank you for continuing to provide the context behind all of this!

adamsachs (approved):
looking great, really awesome work here @pattisdr. 

thank you for the time and care in addressing all my comments - both the adjustments as well as the reasoning and context behind some of the more _philosophical_ questions! 😄

src/fides/api/models/privacy_experience.py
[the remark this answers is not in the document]
pattisdr: Naming made sense to me!

pattisdr:
Thanks again for getting up to speed to get this reviewed Adam! Especially appreciate the comments about keeping an eye on what fields are stored in which model and will keep a watch on this one.

PR #3397, opened 2023-05-29
Fix overlay rendering issue by finding/creating a dedicated parent element for Preact

### LINKED ISSUE #3293: Allow mounting `fides-js` components into a specified HTML container

https://github.com/ethyca/fides/pull/3191 introduces mounting a Preact component into a site's `document.body`. This may or may not be ideal in terms of manipulating a site's DOM.

As we develop more, we should see how this plays out, and also give the option for a user to specify a "container node selector" which fides-js loads all of its components into. That selector can default to `document.body` for now, but should be able to be overwritten

### Additional context

Original discussion: https://github.com/ethyca/fides/pull/3191#discussion_r1191465687

NevilleS:
Closes #3293 

### Code Changes

* [X] Add a `overlayParentId` option that defaults to `fides-overlay` to specify the parent element ID
* [X] Either find _or_ create that parent element when rendering the overlay via `Preact.render()`
* [X] Use `readystatechange` event listening to ensure DOM is loaded

### Steps to Confirm

* [X] Manually edit your local privacy center setting to enable the overlay, populate with notices, etc. (easier said than done!)
* [X] Run `nox -s "fides_env(test)"` to run everything locally
* [X] Open the sample app in a location where the banner would be displayed
* [X] Ensure the banner renders correctly and the rest of the app is maintained

#### Before (Banner Disabled)
image

Notice how `document.body` contains a `div` with `id="__next"` and a few other things...

#### Before (Banner Enabled)
image

Notice how `document.body` no longer has the Next root and instead we've replaced it with our own `div` with `id="fides-js-root"`

### Pre-Merge Checklist

* Documentation:
* [X] Update `CHANGELOG.md`

### Description Of Changes

This fixes a rendering issue I ran into when doing some E2E testing locally. I can't pinpoint exactly why this is fundamentally required because in our testing of "simpler" pages like the static HTML demo page we use, this doesn't seem to happen - I think it must be an interaction between the NextJS VDOM and our own.

This just defensively adds a parent element to the `document.body` where we insert our root node to keep it away from thrashing with other running React apps, etc.

### DIFF
--- a/clients/fides-js/src/lib/consent.tsx
+++ b/clients/fides-js/src/lib/consent.tsx
@@ -28,8 +28,22 @@ export const initOverlay = async ({
     try {
       debugLog(
         options.debug,
-        "Rendering Fides overlay CSS & HTML into the DOM..."
+        "Rendering Fides overlay CSS & HTML into the DOM... (updated!)"
       );
+
+      // TODO: clean this up
+      const rootElementId = "fides-overlay";
+      let root = document.getElementById(rootElementId);
+      if (!root) {
+        debugLog(
+          options.debug,
+          "Root #fides-overlay not found, inserting before rendering..."
+        );
+        // Create our own root element and append to body first
+        root = document.createElement("div");
+        root.id = rootElementId;
+        document.body.appendChild(root);
+      }
 
       if (experience.component === ComponentType.OVERLAY) {
         if (experience.delivery_mechanism === DeliveryMechanism.BANNER) {
@@ -41,7 +55,7 @@ export const initOverlay = async ({
               experience={experience}
               geolocation={geolocation}
             />,
-            document.body
+            root
           );
           debugLog(options.debug, "Fides overlay is now showing!");
         } else if (experience.delivery_mechanism === DeliveryMechanism.LINK) {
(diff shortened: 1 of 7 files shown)


allisonking (approved):
this makes sense to me! a few things I'm thinking about...

* did this trigger the [react hydration error](https://nextjs.org/docs/messages/react-hydration-error) at all?
* accessibility wise, I think we'll want our banner to be one of the first things a user can tab to. this worked before since, iirc, the banner rendered at the top of the DOM. but we'll have to think of an actual solution if users could put their `<div id="fides-overlay"/> wherever they want
* should we add a `fides-overlay` div in the sample app? or we could add that as a use case we make sure to QA down the line

clients/fides-js/src/lib/consent.tsx
```diff
@@ -58,9 +72,11 @@ export const initOverlay = async ({
   // Ensure we only render the overlay to the DOM once it's loaded
   if (document?.readyState !== "complete") {
     debugLog(options.debug, "DOM not loaded, adding event listener");
-    document.addEventListener("DOMContentLoaded", async () => {
-      debugLog(options.debug, "DOM fully loaded and parsed");
-      await renderFidesOverlay();
+    document.addEventListener("readystatechange", async () => {
```
eastandwestwind, line 75: this makes more sense, thanks for updating!

eastandwestwind (approved):
Looks great!

PR #3401, opened 2023-05-30
[Backend] Out of the Box Privacy Notices

### LINKED ISSUE #2833: Backend Out of the Box Privacy Notices

pattisdr:
Closes #2833 

### Description Of Changes

Loads some out of the box privacy notices on startup.  Also creates the "experiences" for these notices behind the scenes 

[Out of the box copy here](https://ethyca.atlassian.net/wiki/spaces/PM/pages/2656108545/Privacy+Notice+Templates)

Relates to out of the box [Experience Configs](https://github.com/ethyca/fides/pull/3449)

### Code Changes

* [ ] Adds a yaml file with five out-of-the-box notices.
* [ ] Adds a new PrivacyNoticeTemplate table 
* [ ] On startup, add any new templates or update existing templates from the `data/privacy_notices/privacy_notice_templates.yml` file into the PrivacyNoticeTemplates file
* [ ] If there is not a PrivacyNotice for the given PrivacyNoticeTemplate, create one.  If a Notice already exists, don't update it.
* [ ] Note that if automatically-created notices would create a conflict with notices already created, we skip creating notices

screencapture-localhost-3000-consent-privacy-notices-2023-05-30-15_20_53 (1)

### Steps to Confirm
* [ ] Run `nox -s dev -- shell`
* [ ] Verify in fides logs that you see a message like this: `2023-05-30 22:52:43.877 [INFO] (consent_util:load_default_notices_on_startup:462): Loading default privacy notice templates from /fides/data/privacy_notices/privacy_notice_templates.yml`
* [ ] Visit `{{host}}/privacy-notice?systems_applicable=false&show_disabled=true` to see the automatically created privacy_notices.  If "origin" is not null, then that notice was created from a template.
* [ ] Visit GET {{host}}/privacy-experience/? to see the automatically created experiences

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`


src/fides/api/api/v1/endpoints/privacy_notice_endpoints.py
```diff
@@ -227,140 +214,14 @@ def create_privacy_notices(
...
-            status_code=HTTP_422_UNPROCESSABLE_ENTITY,
-            # pylint: disable=no-member
-            detail=exc.errors(),  # type: ignore
-        )
-
-
-def prepare_privacy_notice_patches(
-    privacy_notice_updates: List[schemas.PrivacyNoticeWithId],
-    db: Session,
-) -> List[Tuple[schemas.PrivacyNoticeWithId, PrivacyNotice]]:
```
pattisdr, line 291: Moved to src/fides/api/util/consent_util.py and heavily adapted in several places to share between updating privacy notices and new updating privacy templates on startup.

src/fides/api/api/v1/endpoints/privacy_notice_endpoints.py
```diff
@@ -227,140 +214,14 @@ def create_privacy_notices(
...
-    """
-    ids = set()
-    for privacy_notice in privacy_notices:
-        if privacy_notice.id not in ids:
-            ids.add(privacy_notice.id)
-        else:
-            raise HTTPException(
-                HTTP_422_UNPROCESSABLE_ENTITY,
-                detail=f"More than one provided PrivacyNotice with ID {privacy_notice.id}.",
-            )
```
pattisdr, line 361: Moved to src/fides/api/util/consent_util.py - unchanged

src/fides/api/app_setup.py
```diff
@@ -162,6 +166,11 @@ async def run_database_startup() -> None:
         return
     finally:
         db.close()
+
+    if not CONFIG.test_mode:
+        # Default notices subject to change, so preventing these from
+        # loading in test mode to avoid interfering with unit tests.
+        load_default_privacy_notices()
```
pattisdr, line 173: We have hundreds of existing tests across privacy notices, experiences, and experience configs whose notices would conflict with those that we're loading on startup.  Especially because we're still figuring out "regions" - rather than loading these in tests by default and changing all of our tests to accommodate this, I'm just not loading them in test mode.

I do call this method separately in a unit test that will fail if any validation fails as a sanity check.

src/fides/api/models/privacy_notice.py
```diff
@@ -176,13 +204,43 @@ def generate_notice_key(cls, name: Optional[str]) -> FidesKey:
...
+        cloned_attributes = self.__dict__.copy()
+        for key, val in data.items():
+            cloned_attributes[key] = val
+
+        # remove protected fields from the cloned dict
+        cloned_attributes.pop("_sa_instance_state")
+
+        # create a new object with the updated attribute data to keep this
+        # ORM object (i.e., `self`) pristine
+        return PrivacyNotice(**cloned_attributes)
```
pattisdr, line 224: Moving this method so I can share with templates and notices

src/fides/api/util/consent_util.py
```diff
@@ -228,3 +247,278 @@ def get_or_create_fides_user_device_id_provided_identity(
         )
 
     return identity  # type: ignore[return-value]
+
+
+def validate_notice_data_uses(
+    privacy_notices: List[Union[PrivacyNoticeWithId, PrivacyNoticeCreation]],
```
pattisdr, line 253: This method was just moved from src/fides/api/api/v1/endpoints/privacy_notice_endpoints.py.  The only change is the type signature so it can be shared to validate data uses before creating a Notice or a Template

src/fides/api/util/consent_util.py
```diff
@@ -228,3 +247,278 @@ def get_or_create_fides_user_device_id_provided_identity(
...
+    """
+    ids = set()
+    for privacy_notice in privacy_notices:
+        if privacy_notice.id not in ids:
+            ids.add(privacy_notice.id)
+        else:
+            raise HTTPException(
+                HTTP_422_UNPROCESSABLE_ENTITY,
+                detail=f"More than one provided PrivacyNotice with ID {privacy_notice.id}.",
+            )
```
pattisdr, line 283: This method was moved from src/fides/api/api/v1/endpoints/privacy_notice_endpoints.py and is unchanged.

src/fides/api/util/consent_util.py
```diff
@@ -228,3 +247,281 @@ def get_or_create_fides_user_device_id_provided_identity(
...
+    check_conflicting_data_uses(new_notices, existing_notices)
+
+    created_privacy_notices: List[PrivacyNotice] = []
+    affected_regions: Set = set()
+
+    for privacy_notice in privacy_notice_schemas:
+        if should_escape:
+            # should_escape flag is for when we're creating a notice
+            # from a template. The content was already escaped in the
+            # template - we don't want to escape twice.
```
pattisdr, line 311: This method was largely relocated from the original API `create_privacy_notices` endpoint.  

It is shared between the endpoint  that creates privacy notices in the API and the method that loads out of the box privacy notices on startup.  The main change is that we can toggle whether we should escape the notices or not.  If we're creating the notices from a template, the notice was already escaped in the template so we don't want to escape it twice.

Otherwise this method should be nearly identical to the body of the original create_privacy_notices function.

src/fides/api/util/consent_util.py
```diff
@@ -228,3 +247,281 @@ def get_or_create_fides_user_device_id_provided_identity(
...
+    of the privacy notice in the database combined with the patch updates from the request
+    """
+    try:
+        PrivacyNoticeCreation.from_orm(dry_update)
+    except ValueError as exc:
+        raise HTTPException(
+            status_code=HTTP_422_UNPROCESSABLE_ENTITY,
+            # pylint: disable=no-member
+            detail=exc.errors(),  # type: ignore
+        )
```
pattisdr, line 347: Moved from src/fides/api/api/v1/endpoints/privacy_notice_endpoints.py and unchanged

src/fides/api/util/consent_util.py
```diff
@@ -228,3 +247,281 @@ def get_or_create_fides_user_device_id_provided_identity(
...
+
+
+def prepare_privacy_notice_patches(
```
pattisdr, line 350: This method is copied over from src/fides/api/api/v1/endpoints/privacy_notice_endpoints.py but has the most modifications so it can be shared between updating a privacy notice via the API or _upserting_ a privacy notice _template_ on startup.

It's a little extra logic that can make this confusing to follow but I do like being able to share this code between both usages. There are so many things we want to double check and I want to make sure this happens for both resources.

src/fides/api/util/consent_util.py
```diff
@@ -228,3 +247,281 @@ def get_or_create_fides_user_device_id_provided_identity(
...
+    privacy_notice_updates: List[PrivacyNoticeWithId],
+    db: Session,
+    model: Union[Type[PrivacyNotice], Type[PrivacyNoticeTemplate]],
```
pattisdr, line 353: New argument - specifying whether this is a notice or template

src/fides/api/util/consent_util.py
```diff
@@ -228,3 +247,281 @@ def get_or_create_fides_user_device_id_provided_identity(
...
+            model=update_data,
+            fields=PRIVACY_NOTICE_ESCAPE_FIELDS,
+        )  # type: ignore
+        if update_data.id not in existing_notices:
+            if not allow_create:
+                raise HTTPException(
+                    status_code=HTTP_404_NOT_FOUND,
+                    detail=f"No {model.__class__.__name__} found for id {update_data.id}.",
+                )
+            updates_and_existing.append((update_data, None))
```
pattisdr, line 400: New capability that is for templates only - we can stage a template for creation here.

src/fides/api/util/consent_util.py
```diff
@@ -228,3 +247,281 @@ def get_or_create_fides_user_device_id_provided_identity(
...
+                dry_update
+            )  # Checks consent mechanism + delivery location
+            validation_updates.append(
+                existing_notice.dry_update(data=update_data.dict(exclude_unset=True))
+            )
+            # and don't include it anymore in the existing notices used for validation
+            existing_notices.pop(existing_notice.id, None)
+        else:
+            # Add the pending created record to the the temporary updates for validation
+            validation_updates.append(model(**update_data.dict(exclude_unset=True)))
```
pattisdr, line 422: New - for templates, we're adding the pending created record here if it didn't already exist

src/fides/api/util/consent_util.py
```diff
@@ -228,3 +247,281 @@ def get_or_create_fides_user_device_id_provided_identity(
...
+def load_default_notices_on_startup(
+    db: Session, notice_yaml_file_path: str
+) -> Tuple[List[PrivacyNoticeTemplate], List[PrivacyNotice]]:
+    """
+    On startup, loads any new PrivacyNoticeTemplates into the database, and updates
+    existing templates where applicable.
+
+    Creates Privacy Notices from these templates, if no notices are linked to the given template
+    already.  Otherwise, we skip making updates to the Notices themselves automatically.
+    """
```
pattisdr, line 488: This is my opinion of how this function should behave.  I upsert the templates from the yaml file we ship with. And then, in the other table, if the templates were new, I create privacy notices from them, exposing them to the end user.  But if they aren't new, I don't update existing notices.  Customers may have updated the auto-generated notices since their creation.

src/fides/api/util/consent_util.py
```diff
@@ -228,3 +247,281 @@ def get_or_create_fides_user_device_id_provided_identity(
...
+    the existing PrivacyNoticeTemplate if it exists or None otherwise.
+
+    :param privacy_notice_updates: List of privacy notice schemas with ids: appropriate
+    for editing PrivacyNotices or upserting PrivacyNoticeTemplates
+    :param db: Session
+    :param model: one of PrivacyNotice or PrivacyNoticeTemplate
+    :return:
+    """
+    allow_create: bool = model == PrivacyNoticeTemplate
+    ignore_disabled: bool = model != PrivacyNoticeTemplate
```
pattisdr, line 374: Some new flags to flip some behavior in this method depending on whether the resource is a template or a notice

src/fides/api/util/consent_util.py
```diff
@@ -228,3 +247,281 @@ def get_or_create_fides_user_device_id_provided_identity(
...
+                    check_name=False,
+                )
+            )
+    return upserted_templates  # type: ignore[return-value]
+
+
+def load_default_notices_on_startup(
+    db: Session, notice_yaml_file_path: str
+) -> Tuple[List[PrivacyNoticeTemplate], List[PrivacyNotice]]:
+    """
```
pattisdr, line 482: New function and one of the primary changes on this PR.

tests/fixtures/application_fixtures.py
```diff
@@ -1446,7 +1446,6 @@ def privacy_notice(db: Session) -> Generator:
             "name": "example privacy notice",
             "notice_key": "example_privacy_notice",
             "description": "a sample privacy notice configuration",
-            "origin": "privacy_notice_template_1",
```
pattisdr, line 1449: Notices now have proper "origin" FK's to the template they were created from where applicable so we can't put random strings here any longer.

tests/fixtures/test_bad_privacy_notices_update.yml
```diff
@@ -0,0 +1,34 @@
+privacy_notices:
```
pattisdr, line 1: These and similar files are added here to test various scenarios when templates are added on startup

pattisdr:
Out of the box experience configs were just merged which conflicts in several locations. Getting this up-to-date now.

src/fides/api/models/privacy_notice.py
```diff
@@ -108,6 +108,36 @@ class PrivacyNoticeRegion(Enum):
...
+    no_11 = "no_11"  # norway, rogaland
+    no_54 = "no_54"  # norway, troms og finnmark
+    no_50 = "no_50"  # norway, trøndelag
+    no_38 = "no_38"  # norway, vestfold og telemark
+    no_46 = "no_46"  # norway, vestland
+    no_30 = "no_30"  # norway, viken
+    no_22 = "no_22"  # norway, jan mayen
+    no_21 = "no_21"  # norway, svalbard
+
+    li = "li"  # liechtenstein
```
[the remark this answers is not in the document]
adamsachs: interesting, but couldn't they have `"is"` or `"no"` as the enum _values_ while the enum _name_ would be something different, so as to avoid the conflict with the python keywords?

i'm guessing you thought through this a fair amount and it's not an easy adjustment. and i know we've got some time pressure, so not trying to make more work than needed. but if there's no real requirement to have regions within iceland or norway, it just feels a bit odd to do this, especially because i think as a consequence we'll have `PrivacyExperience` records created for each region?

i may not be understanding this correctly. just struck me as a bit surprising to have all of these regions if the only purpose is to avoid a python conflict, i feel like there has to be some other way around it... 🤷

src/fides/api/models/privacy_notice.py
```diff
@@ -176,13 +204,43 @@ def generate_notice_key(cls, name: Optional[str]) -> FidesKey:
...
+    """
+
 
 class PrivacyNotice(PrivacyNoticeBase, Base):
     """
     A notice set up by a system administrator that an end user (i.e., data subject)
     accepts or rejects to indicate their consent for particular data uses
     """
 
+    origin = Column(
```
adamsachs, line 239: nice refactors here - small adjustments to make sure we're reusing code 👍

tests/ops/util/test_consent_util.py
```diff
@@ -461,6 +479,505 @@ def test_get_fides_user_device_id_provided_identity(
...
+        new_history.delete(db)
+        history.delete(db)
+
+        new_privacy_notice.delete(db)
+        notice.delete(db)
+
+        new_template.delete(db)
+        template.delete(db)
+
+    def test_load_actual_default_notices(self, db):
```
adamsachs, line 768: i was walking thru the tests, and was wondering whether you'd include one to load the actual defaults. of course you did! 😄

tests/ops/util/test_consent_util.py
```diff
@@ -461,6 +479,505 @@ def test_get_fides_user_device_id_provided_identity(
...
+        """Sanity check, makings sure that default privacy notices don't load with errors"""
+        new_templates, new_privacy_notices = load_default_notices_on_startup(
+            db, DEFAULT_PRIVACY_NOTICES_PATH
+        )
+        assert len(new_templates) == 5
```
adamsachs, line 773: nit: i feel like whenever we do this sort of len comparison in our tests, e.g. with saas templates, it ends up being really cumbersome and annoying to keep updated as that default list changes. this one's a bit simpler, but still - may just be worth `assert`ing that `len(new_templates) >= 1` and `len(new_privacy_notices) >= 1`?

src/fides/api/util/consent_util.py
```diff
@@ -238,6 +255,284 @@ def get_or_create_fides_user_device_id_provided_identity(
...
+            detail=exc.errors(),  # type: ignore
+        )
+
+
+def prepare_privacy_notice_patches(
+    privacy_notice_updates: List[PrivacyNoticeWithId],
+    db: Session,
+    model: Union[Type[PrivacyNotice], Type[PrivacyNoticeTemplate]],
+) -> List[
+    Tuple[PrivacyNoticeWithId, Optional[Union[PrivacyNotice, PrivacyNoticeTemplate]]]
```
adamsachs, line 361: is there a reason we can't use `PrivacyNoticeBase` as an abstract type here in the type annotation? it'd read a lot cleaner IMO. if mypy isn't liking that...then i'm a bit angry at mypy

src/fides/api/util/consent_util.py
adamsachs: this is nice, pretty straightforward branching. i remember all of this validation logic is so complex, so i agree with you that slotting this in here pays dividends, even if there's some branching. the core validation flow stays in tact, and that's better to have consolidated in one place rather than being duplicated.

nice job thinking thru this, i'm sure it took a bit of mental grit to get this new use case integrated!

src/fides/api/util/consent_util.py
adamsachs: that makes sense, i think it's also how i'd expect this to behave (though i'm a lot less up to speed on the use case than you are!)

have you clarified with product that this is the desired behavior? just because this feels like business logic rather than technical constraints driving this (not that it's a bad thing) - so i'd expect us to be doing whatever product thinks is best.

src/fides/api/util/consent_util.py
```diff
@@ -238,6 +255,284 @@ def get_or_create_fides_user_device_id_provided_identity(
...
+        for privacy_notice_data in notices:
+            template_schemas.append(PrivacyNoticeWithId(**privacy_notice_data))
+
+        # Upsert Privacy Notice Templates
+        privacy_notice_templates: List[
+            PrivacyNoticeTemplate
+        ] = upsert_privacy_notice_templates_util(db, template_schemas)
+
+        # Determine which templates don't have corresponding records in PrivacyNotice
+        new_templates: List[PrivacyNoticeTemplate] = [
```
adamsachs, line 513: list comprehension at its finest!

src/fides/api/util/consent_util.py
```diff
@@ -238,6 +255,284 @@ def get_or_create_fides_user_device_id_provided_identity(
...
+    for editing PrivacyNotices or upserting PrivacyNoticeTemplates
+    :param db: Session
+    :param model: one of PrivacyNotice or PrivacyNoticeTemplate
+    :return:
+    """
+    allow_create: bool = model == PrivacyNoticeTemplate
+    ignore_disabled: bool = model != PrivacyNoticeTemplate
+
+    # first we populate a map of privacy notices or templates in the db, indexed by ID
+    existing_notices: Dict[str, Union[PrivacyNotice, PrivacyNoticeTemplate]] = {}
```
adamsachs, line 383: same question re: `PrivacyNoticeBase` in the type annotation

tests/ops/api/v1/endpoints/test_privacy_notice_endpoints.py
```diff
@@ -37,6 +30,7 @@
     PrivacyNoticeCreation,
     PrivacyNoticeResponse,
 )
+from fides.api.util.consent_util import validate_notice_data_uses
```
adamsachs, line 33: nit: may make sense to move `TestValidateDataUses` test class over to `test_consent_util.py` now that the underlying utility function has been moved to `consent_util.py`?

adamsachs (approved):
nice work @pattisdr , this is pretty straightforward! special thanks for all your annotations with comments, it definitely helped reading through things.

just a few minor comments that you could look into, i don't think any of them are blockers. i think you're right that likely the riskiest/trickiest thing here is integrating your new use case into all that pretty nasty update logic -- as far as i can tell, you've done a nice job with that, and have kept things about as simple as they can be!

src/fides/api/models/privacy_notice.py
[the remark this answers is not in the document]
pattisdr: You're right I can do "iceland" or similar for now. 

I learned today too the frontend location service is actually by region, and we need to do a larger discussion around how we're going to sync up around regions but that's likely next sprint.

Your point about creating excessive PrivacyExperience records when we still don't know how this is going to settle out is definitely valid.

src/fides/api/util/consent_util.py
pattisdr: good callout - will update when I hear back from Michael

data/privacy_notices/privacy_notice_templates.yml
```diff
@@ -0,0 +1,296 @@
+privacy_notices:
+- name: Essential
+  notice_key: essential
+  id: pri_51c85583-40ec-4257-95ef-db3eea860ess
```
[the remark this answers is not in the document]
pattisdr: ah it was a modification of the FidesBase.generate_uuid, except I replaced the last three characters with the first three characters of the notice_key.  It would be nice to just rely on "notice_key" but these are not at all guaranteed to be unique!

tests/ops/util/test_consent_util.py
pattisdr: that's fair. That's also a big reason why I'm not loading these out of the box notices for tests too.

tests/ops/api/v1/endpoints/test_privacy_notice_endpoints.py
pattisdr: that makes sense, thanks

src/fides/api/util/consent_util.py
adamsachs: ugh i figured that might be the case, that's so frustrating. i think this may be my least favorite thing about mypy, the support for base/abstract classes seems very lacking. i wonder if then it may be worth an alias here though? could even redefine `PrivacyNoticeBase` here in `consent_util.py`, e.g. `PrivacyNoticeBase = Type[Union[PrivacyNotice, PrivacyNoticeTemplate]]`? i could have that all wrong, just a stab in the dark.

anyway, definitely not a blocker, just trying to think of how to make it clearer to the next reader that these are really two very similar subtypes!

src/fides/api/models/privacy_notice.py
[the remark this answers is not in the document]
pattisdr: ah right norway "no" also was problematic for some reason? Anyway, to not spend too much time here, I'm using three letter country codes for both of those for now, but it is likely all these regions will get an overhaul in a future sprint.

src/fides/api/util/consent_util.py
pattisdr: I can add some sort of alias!

src/fides/api/util/consent_util.py
pattisdr: OK I am preemptively merging this work ahead of the follow-up changes I need to make tonight in a separate ticket, but will reticket if the default behavior needs to change -

pattisdr:
Preemptively merging @adamsachs so I can make the changes to https://github.com/ethyca/fides/issues/3465 tonight - happy to address any more followups there.

src/fides/api/util/consent_util.py
pattisdr: OK confirmed with Michael that this logic is ok for now -

adamsachs:
> Preemptively merging @adamsachs so I can make the changes to #3465 tonight - happy to address any more followups there.

yeah of course! didn't mean to suggest this should hold things up. thanks for confirming 👍

PR #3405, opened 2023-05-31
Handle tags returned in an unexpected order

### LINKED ISSUE #3404: Unsorted tags casuing conflicts in `nox -s "tag(push)"`
### Bug Description

When trying to push a new alpha tag I hit an error due to the tags being found being unsorted. The attempted tag to be generated already existed.

#### Steps to Reproduce

_will stage a test using `["2.14.1a0", "2.14.1a1"],`_

### Expected behavior

The tag would have been generated as `2.14.1a2`

### Screenshots

n/a

### Environment

* Version:
* OS:
* Python Version:
* Docker Version:

### Additional context

Found while trying to push an updated alpha tag for testing the bastion host feature

SteveDMurphy:
Closes #3404

### Code Changes

* [x] Add a failing test for tags that are returned in an unexpected order
* [x] Sort matched tags in descending order to ensure the increment is as expected

### Steps to Confirm

* [ ] Inspecting the matched tags (`2.14.1a*`) will show the two in tests as out of sequence. Reverting the last commit will show the test failing in that scenario 

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### Description Of Changes

I did this manually last night to force `2.14.1a2` to push, then created an issue and PR to see if this was the right path forward.


tests/nox/test_git_nox.py
adamsachs, line 1: wait a sec, did you move this over to this dir deliberately? it used to be here, but then @ThomasLaPiana moved it over to `noxfiles`, i.e. right next to the "source" code: https://github.com/ethyca/fides/blob/main/noxfiles/test_git_nox.py

let's keep it there.

tests/nox/test_git_nox.py
```diff
@@ -0,0 +1,72 @@
...
+            ("some-test-feature-branch", ["2.9.1"], "2.9.2a0"),
+            ("some-test-feature-branch", ["2.9.1", "2.9.0"], "2.9.2a0"),
+            (
+                "some-test-feature-branch",
+                ["2.9.3a0", "2.9.2"],
+                "2.9.3a1",
+            ),  # our repo happens to already have a 2.9.3a0 tag, let's use it for testing here
+            (
+                "some-other-feature",
+                ["2.14.1a0", "2.14.1a1", "2.14.0"],
```
adamsachs, line 36: this is what you've added, right? looks good to me!

noxfiles/git_nox.py
```diff
@@ -145,12 +146,16 @@ def increment_tag(
...
         (
-            re.fullmatch(version_branch_tag_pattern, tag.name)
+            tag.name
             for tag in all_tags
             if re.fullmatch(version_branch_tag_pattern, tag.name)
         ),
+        reverse=True,
+    )
+    latest_tag = next(
+        (re.fullmatch(version_branch_tag_pattern, tag) for tag in sorted_tag_matches),
```
adamsachs, line 158: this `re.fullmatch` shouldn't be needed again since we've still got that conditional in the initial sort on L152/3, right? shouldn't be a problem, i think it's just redundant if i'm reading correctly.

adamsachs (commented):
this looks great, thank you @SteveDMurphy for making this more robust!

i'm pretty surprised that you were getting the tags returned out of order - it does indeed look like the [2.14.1a0](https://github.com/ethyca/fides/releases/tag/2.14.1a0) tag was created _after_ the [2.14.1a1](https://github.com/ethyca/fides/releases/tag/2.14.1a1) tag - i'm guessing someone must have re-published the tag.

i assumed tags would be published in the right order and that they were basically immutable, but it's certainly possible to delete and re-add a tag. i guess i didn't expect anyone to do that 😄. regardless, there's no benefit in assuming this, and doing the more defensive, explicit sorting you've put in here is a good safety measure (that we clearly need!). so thank you again for the catch/fix!

just some minor comments, the gist of this looks good to me 👍

noxfiles/git_nox.py
SteveDMurphy: I don't think it is, but I figured it was safest to keep the output of `latest_tag` consistent. Will take a closer look 👍🏽

adamsachs: cool, really not a big deal either way - just think if we can remove it, it'd make things a bit more readable!

adamsachs:
oh also, i forgot to mention - we should make the same changes over on the `fidesplus` side! we've got the same nox commands over there, unfortunately we don't have a good way of sharing nox code between the two repos so we've gotta just c+p for now. hopefully the tagging code stabilizes after this change! 🤞

noxfiles/git_nox.py
SteveDMurphy: Agreed! I tried a few things and it still felt a bit wonky unfortunately with how we increment. The closest possible improvement that I could think of without some risky/fuzzy string matching is in https://github.com/ethyca/fides/pull/3405/commits/76fc550af2451900f6a22d85dcf8a3b7879fb755 if you want to let me know what you think. Happy to go any route you see fit!

SteveDMurphy:
> oh also, i forgot to mention - we should make the same changes over on the `fidesplus` side! we've got the same nox commands over there, unfortunately we don't have a good way of sharing nox code between the two repos so we've gotta just c+p for now. hopefully the tagging code stabilizes after this change! 🤞

Once we are good here I can pop a PR open over there as well, thanks for calling that out!

tests/nox/test_git_nox.py
SteveDMurphy: Yep! Thanks 🙌🏽

tests/nox/test_git_nox.py
SteveDMurphy: Resolved in https://github.com/ethyca/fides/pull/3405/commits/0ee846711c88eae0acea6ca3f22f3a54d4a4a6bd

noxfiles/git_nox.py
adamsachs: ohh right i get it now, sorry this was an unnecessary nit from me. i'm good with where you landed!

adamsachs (approved):
looks good on my end, thanks again for being proactive and making this fix!

PR #3410, opened 2023-05-31
Don't divulge user's existence

seanpreston:


src/fides/api/api/v1/endpoints/user_endpoints.py
```diff
@@ -554,13 +554,17 @@ def user_login(
...
+
         if not user_check:
-            raise HTTPException(status_code=HTTP_404_NOT_FOUND, detail="No user found.")
+            raise HTTPException(
+                status_code=HTTP_404_NOT_FOUND, detail=invalid_user_error_msg
+            )
 
         if not user_check.credentials_valid(user_data.password):
             raise HTTPException(
                 status_code=HTTP_403_FORBIDDEN,
```
daveqnet, line 566: @seanpreston, I'd suggest a 404 here instead of a 403 to keep it consistent with the previous exception.

pattisdr (approved):
this looks good -

Issue #3423, opened 2023-06-01
Fix styling issues with consent modal


NevilleS:
### Bug Description

The current styling of the modal is very functional/WIP:
<img width="1001" alt="image" src="https://github.com/ethyca/fides/assets/1834295/a7fa4d85-a3eb-4846-a610-86567cd6e191">

This needs a solid polish pass to bring it up to our goal:
<img width="649" alt="image" src="https://github.com/ethyca/fides/assets/1834295/148d2a3c-246b-4190-9344-13691f9db25b">

### Expected behavior

Lots of individual details, but the most significant are:
- [ ] Font
- [ ] Padding
- [ ] Borders
- [ ] etc.

Check out the stylesheet here: https://www.figma.com/file/dmEwdK3xZwjKfGVQ9t66xe/Fides-v.2-Master-Working-Files?type=design&node-id=6967-209288&t=Y5Krqi13KhRxafU9-0

allisonking:
Hmm there must be some browser compatibility / other sneaky stuff going on here.

When I visit `fides-js-demo.html` in chrome:


![Image](https://github.com/ethyca/fides/assets/24641006/5bb2cf35-d5b4-4b63-96c1-328ad75a4388)

But when I visit in firefox 🙀 



![Image](https://github.com/ethyca/fides/assets/24641006/cda6bd3a-66d1-4234-8fc0-0987a136626c)

And Safari 🙀 



![Image](https://github.com/ethyca/fides/assets/24641006/17da7366-16df-4d7a-91d9-433662a1c36b)

allisonking:
Okay for firefox and safari, I think this is because I was using [css nesting](https://developer.chrome.com/articles/css-nesting/) which turns out [isn't widely supported](https://caniuse.com/css-nesting)

in order to use css nesting, I think we'd need to add a css processor to rollup. it might be easier to unnest everything instead. nesting is nice because it keeps the css a bit cleaner/logical, but we could live without it.

still investigating the diff between the cookie house modal and the demo page modal though... stay tuned!

allisonking:
Closed by https://github.com/ethyca/fides/pull/3429

Roger-Ethyca:
moving to done

PR #3429, opened 2023-06-01
Modal css fixes

allisonking:
Closes https://github.com/ethyca/fides/issues/3423

### Code Changes

* [x] Fix padding issues. This was sneaky because browsers include their own margins on certain elements by default. This didn't become apparent until using it in something like a real next app which probably has a css reset to undo all browser defaults. 
* [x] Unnest CSS selectors. It turns out this isn't widely supported, resulting in much of the modal CSS not rendering at all in firefox and safari!

### Steps to Confirm

* Follow the steps in https://github.com/ethyca/fidesplus/pull/913 and take a look at the modal in the cookie house app
* Take a look in Firefox and Safari too

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

A note on [CSS resets](https://www.joshwcomeau.com/css/custom-css-reset/):

Apps can normally do something like 
```css
* {
  margin: 0
}
```

We can't do that, or else we might mess up the host app's CSS! we could do

```css
#fides-overlay {
  * {
    margin: 0
  }
}
```
but that's nested CSS so isn't widely supported. the usual conversion of
```css
#fides-overlay * {
  margin: 0
}
```
doesn't quite work because of CSS specificity rules I haven't quite wrapped my head around. It does apply that `margin:0`, but then our subsequent css doesn't overwrite it properly. I opted not to look into this too much, but we may want to consider adding a css reset to our fides demo page so that styling issues like this don't come up again (assuming we develop against the demo page)

### DIFF
--- a/clients/fides-js/src/lib/overlay.module.css
+++ b/clients/fides-js/src/lib/overlay.module.css
@@ -61,11 +61,6 @@ div#fides-overlay {
   /* CSS reset values, adapted from https://www.joshwcomeau.com/css/custom-css-reset/ */
   line-height: calc(1em + 0.5rem);
   -webkit-font-smoothing: antialiased;
-
-  /* Reset browser defaults for margins so we can consistently apply our own */
-  * {
-    margin: 0;
-  }
 }
 
 #fides-consent-modal-link {
@@ -205,21 +200,21 @@ div#fides-consent-modal {
   top: 50%;
   left: 50%;
   transform: translate(-50%, -50%);
+}
 
-  .modal-header {
-    text-align: center;
-    font-weight: 600;
-    font-size: var(--fides-consent-overlay-font-size-title);
-    color: var(--fides-consent-overlay-title-font-color);
-  }
+div#fides-consent-modal .modal-header {
+  text-align: center;
+  font-weight: 600;
+  font-size: var(--fides-consent-overlay-font-size-title);
+  color: var(--fides-consent-overlay-title-font-color);
+}
 
-  .modal-description {
-    margin: 1em 0 1em 0;
-  }
+div#fides-consent-modal .modal-description {
+  margin: 1em 0 1em 0;
+}
 
-  .modal-button-group {
-    display: flex;
-  }
+div#fides-consent-modal .modal-button-group {
+  display: flex;
 }
 
 .modal-overlay {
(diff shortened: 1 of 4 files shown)


clients/fides-js/src/lib/overlay.module.css
```diff
@@ -45,22 +45,29 @@
   );
 
   /* Everything else */
-  --fides-consent-overlay-font-family: "inherit";
-  --fides-consent-overlay-font-size: 16px;
-  --fides-consent-overlay-font-size-title: 18px;
-  --fides-consent-overlay-padding: 0.75em 2em 1em;
+  --fides-consent-overlay-font-family: Inter, sans-serif;
```
NevilleS, line 48: Annoying-but-true-fact... we can't rely on `Inter` being available... I'm not 100% sure what to do here yet.

clients/fides-js/src/lib/overlay.module.css
```diff
@@ -193,16 +200,21 @@ div#fides-consent-modal {
   top: 50%;
   left: 50%;
   transform: translate(-50%, -50%);
+}
 
-  .modal-header {
```
NevilleS, line 197: Technically I'm adding `postcss` here: https://github.com/ethyca/fides/pull/3431

...which'll let us enable lots of CSS preprocessing features like nesting and other super-reasonable things like that. But let's do vanilla CSS for now!

NevilleS (approved):
This looks great to me, thanks. The only issue here is the use of `Inter` which isn't a standard browser font...

Assuming y'all figure out something for that, let's merge this in

clients/fides-js/src/lib/overlay.module.css
allisonking: yeah, that's why I added `sans-serif` as a [backup option](https://css-tricks.com/css-basics-fallback-font-stacks-robust-web-typography/), which is a browser default. I think it looks reasonable (at least better than Times!). `sans-serif` kicks in in the demo page since that one doesn't have `Inter`

image

I was hoping `inherit` would just use whatever font the base website was already using, but that doesn't seem to be the case. could look into it more though!

NevilleS: OK, I think that's probably fine to inherit the page default... in theory that works? It's a bit fragile though, so we should consider shipping a font-face with the JS itself. Can you add a follow-up issue for that in the next sprint and we'll just make a plan?

NevilleS: It'd also be easy for a customer to override that with a tiny stylesheet like:
```css
:root {
  --fides-consent-overlay-font-family: "Comic Sans"
}
```

...I think

NevilleS (approved):
Thanks 👍

clients/fides-js/src/lib/overlay.module.css
allisonking: https://github.com/ethyca/fides/issues/3440

PR #3452, opened 2023-06-04
Dockerfile: Do not pip install --user, venv instead

ewdurbin:
This allows for the resulting container to be run by non-root users.

My only reasoning for this is that it was an attempt to keep from polluting/conflicting-with the system Python.

A `venv` is just as good at this, but can also be used by non-root users!

### Code Changes

* [x] Refactor Dockerfile to use a `venv` rather than `pip install --user`

### Steps to Confirm

* [ ] `docker run --rm -it --user nobody ethyca/fides:2.14.1a7` (Errors out)
* [ ] build this image and run it with `--user nobody`

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### DIFF
--- a/.github/workflows/backend_checks.yml
+++ b/.github/workflows/backend_checks.yml
@@ -57,7 +57,7 @@ jobs:
           builder: ${{ steps.buildx.outputs.name }}
           context: .
           build-args: PYTHON_VERSION=${{ matrix.python_version }}
-          target: dev
+          target: prod
           outputs: type=docker,dest=/tmp/python-${{ matrix.python_version }}.tar
           push: false
           tags: ${{ env.IMAGE }}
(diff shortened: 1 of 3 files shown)


ThomasLaPiana:
@ewdurbin we specifically use `prod` in CI when testing to make sure that we're testing the application in the same way that our users would be using it. This especially helps catch packaging issues that otherwise would go unnoticed due to the `pip install -e`

I'm going to re-add the dev-requirements to the base build step, I don't think this is a significant issue and having those packages available in the prod image can come in handy for debugging.

ThomasLaPiana:
steps to confirm work for me, just waiting on these final checks to pass before merging

ThomasLaPiana (commented):
LGTM! Nice usability and security change

Issue #3263, opened 2023-05-09
BE: Add `system_connection` table and endpoints


galvana:
### Is your feature request related to a specific problem?

Systems (ctl_system) in Fides need to be associated to Connections (connectionconfig).

### Describe the solution you'd like

Create a new linking table `system_connection` with the following columns

- system_id
- connection_config_id
- privacy_action (enum for access, erasure, consent)

### Acceptance criteria
- There must be one row per privacy action, so if a connection will be used for access and erasure for a given system, then
- Expose an API to associate or remove connections from a system

PR #3461, opened 2023-06-05
3263 system connections

### LINKED ISSUE #3253: Disabled Connectors can't be successfully re-enabled without re-providing credentials
1. Run a Fides instance
2. Add a connector (Hubspot works well)
3. Test the connector (Should be successful)
4. Disable the connector from the menu Screenshot 2023-05-08 at 1 50 30 PM
5. Re-enable the connector
6. Test again 

This will expose two problems:
1. The UI makes it look like it was successfully tested
2. It doesn't successfully test. If you inspect the request, you'll see a 500 error for the test request.

**AC**
- I should be able to disable a connector, re-enable it, and re-test without providing the credentials again
- If the test request from the connector list returns an error, the UI should indicate that the error occurred by saying the last test failed. 

### Dev Notes
When we disable a connection, we PATCH http://localhost:3000/api/v1/connection with this payload
```
[
    {
        "key": "friendbuy_nextgen",
        "name": "Friendbuy Nextgen",
        "disabled": true,
        "connection_type": "saas",
        "access": "write"
    }
]
```
The unintended side-effect is that we null out the `secrets` field on the `connectionconfig` row. We need to look into the patch connection endpoint and see why the patch is acting like a post.

galvana:
Closes #3253 #3263

### Code Changes

* [ ] Added `enabled_actions` to `connectionconfig` to track which actions are enabled for a given connection
* [ ] Replicated the `instantiate_connection` endpoint under system
* [ ] Fixed issue where disabling a connection caused the secrets to get wiped out. I fixed this as part of this ticket since its involved in the connection instantiation logic.
* [ ] Moved `ActionType` and `DrpAction` under `fides.api.schemas.policy` to avoid cyclic imports, this is the reason for the high number of file diffs

### Pre-Merge Checklist

* [ ] Update `CHANGELOG.md`


src/fides/api/util/connection_util.py
```diff
@@ -197,6 +198,16 @@ def patch_connection_configs(
         if system:
             config_dict["system_id"] = system.id
 
+        if existing_connection_config:
```
adamsachs, line 201: this is fixing the bug where the secrets got wiped out, right?

tests/ops/api/v1/endpoints/test_system.py
```diff
@@ -33,7 +38,37 @@ def url_invalid_system() -> str:
...
+            "secrets": {
+                "url": None,
+                "host": "http://localhost",
+                "port": 5432,
+                "dbname": "test",
+                "db_schema": "test",
+                "username": "test",
+                "password": "test",
+            },
+            "enabled_actions": ["access", "erasure"],
```
adamsachs, line 58: do we actually ever validate that this is getting saved to the connection config?

it doesn't need to be in this test module specifically, but it'd be nice to get some validation that this property (as well as some of the other properties we're passing in the payload here) are actually correctly populated. essentially, there's a fair amount of logic in `connection_util.patch_connection_configs` that i think it deserves some more test coverage than we have currently (as far as i can tell).

it's not that you've changed that function so much in this PR, it's more just that it feels like that's potentially a blind spot in some of our test coverage, and now that's becoming more evident as i look to see whether we're validating that your updates here to include `enabled_actions` actually get persisted. i wouldn't consider this a blocker for this PR, but i'm a bit worried about that lack of test coverage on an important piece of code that's not so straightforward.

maybe i've missed some of the test coverage though and i'm making noise about nothing! 😄

tests/ops/api/v1/endpoints/test_system.py
```diff
@@ -33,7 +38,37 @@ def url_invalid_system() -> str:
     return V1_URL_PREFIX + f"/system/not-a-real-system/connection"
 
 
-class TestPatchSystemConnections(TestPatchConnections):
```
adamsachs, line 36: nit: this struck me as a bit odd. after a bit of digging, i realize it makes sense - this test file/module is dedicated to tests on the system _connections_ endpoints, and so having these fixtures at the module level makes sense. but it made me wonder whether we should just tweak the file/module name to something like `test_system_connections.py`, so it's a bit clearer? the tests of the "base" system endpoints comes in `test_api.py`, so with as things are now, it's a bit confusing to track down which tests are where!

(i realize this is nothing you've changed in this PR, but just noticed as i was going through this! so maybe it makes sense to adjust here.)

adamsachs (approved):
nice changes @galvana! looks good to me, just some minor comments, i don't think any of em are blockers.

just to be sure, there are no corresponding UI changes in `main` yet, right? from what i can tell, creating systems and/or connections both worked fine (i.e. no regressions that i could see), but nothing changed substantively.

Issue #3349, opened 2023-05-22
Trigger modal from link on customers' site


eastandwestwind:
AC

Given a static link on the DOM `<a id="fides-consent-modal-link">Manage Consent</a>`, we need to ensure we can reliably trigger the modal overlay upon click.

Roger-Ethyca:
moving to done

PR #3467, opened 2023-06-05
3349 trigger modal from link

eastandwestwind:
Closes https://github.com/ethyca/fides/issues/3349

### Code Changes

* [ ] Updates link to open modal, given element exists and valid experience exists for the overlay component
* [ ] Updates e2e tests

### Steps to Confirm

* [ ] Nav to http://localhost:3000/fides-js-components-demo.html
* [ ] Confirm `Manage Consent` link click triggers the modal to open

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes


clients/fides-js/src/components/fides.css
```diff
@@ -60,9 +60,7 @@ div#fides-overlay {
 }
 
 #fides-modal-link {
-  text-decoration: underline;
```
eastandwestwind: we should allow our customers to style their own links as desired.

clients/fides-js/src/components/Overlay.tsx
```diff
@@ -21,22 +24,43 @@ export interface OverlayProps {
...
+      "Modal link element found, updating it to show and trigger modal on click."
+    );
+    // Update modal link to trigger modal on click
+    const modalLink = modalLinkEl;
+    modalLink.onclick = () => {
+      setModalIsOpen(true);
+      setBannerIsOpen(false);
+    };
+    // Update to show the pre-existing modal link in the DOM
+    modalLink.style.display = "inline";
```
NevilleS, line 59: nit: the CSS for the banner show/hide uses the classname `.fides-banner-hidden` instead of mutating styles directly.

It might be better to add/remove a class to the link element similarly:
```css
#fides-modal-link {
  display: none;
}

#fides-modal-link.fides-modal-link-shown {
  display: inline;
}
```

clients/fides-js/src/components/Overlay.tsx
```diff
@@ -107,13 +131,17 @@ const Overlay: FunctionComponent<OverlayProps> = ({
...
   return (
     <div id="fides-js-root">
-      <ConsentBanner
-        experience={experience.experience_config}
-        onAcceptAll={onAcceptAll}
-        onRejectAll={onRejectAll}
-        waitBeforeShow={100}
-        onOpenModal={() => setModalIsOpen(true)}
-      />
+      {experience.show_banner ? (
```
NevilleS, line 134: This is an unrelated change, right?

clients/fides-js/src/components/Overlay.tsx
```diff
@@ -21,22 +24,43 @@ export interface OverlayProps {
...
+  const [bannerIsOpen, setBannerIsOpen] = useState(false);
 
   if (!experience.experience_config) {
+    debugLog(options.debug, "No experience config found");
     return null;
   }
 
   const privacyNotices = experience.privacy_notices ?? [];
 
+  if (modalLinkEl) {
```
NevilleS, line 47: Are you sure it's safe to call these kind of DOM methods here in a render function? This code will run many times (every render)

clients/fides-js/src/fides.ts
```diff
@@ -224,20 +224,12 @@ const init = async ({
     );
 
     if (shouldInitOverlay) {
-      // Check if there are any notices within the experience that do not have a user preference
-      const noticesWithNoUserPreferenceExist: boolean = Boolean(
-        effectiveExperience?.privacy_notices?.some(
-          (notice) => notice.current_preference == null
-        )
-      );
```
NevilleS, line 232: This check also happens inside the `experienceIsValid` function, right...? That feels duplicative and probably shouldn't exist in the experienceIsValid function at all anymore?

clients/fides-js/src/lib/consent.tsx
```diff
@@ -1,10 +1,9 @@
 import { h, render } from "preact";
 
-import { ComponentType } from "./consent-types";
+import { ComponentType, FIDES_MODAL_LINK } from "./consent-types";
```
NevilleS, line 3: Feels like this should mirror the `overlayParentElementId` option, no? Instead of a constant?

NevilleS (changes requested):
Couple nits here - would prefer to stick to a consistent pattern of providing options (with defaults) for everything instead of having some things hardcoded, etc.

clients/fides-js/src/components/Overlay.tsx
allisonking: ^ good point, probably safest to put in a `useEffect`

clients/fides-js/src/lib/consent.tsx
```diff
@@ -20,6 +19,16 @@ export const initOverlay = async ({
...
 
   async function renderFidesOverlay(): Promise<void> {
+    // Check if there are any notices within the experience that do not have a user preference
+    const noticesWithNoUserPreferenceExist: boolean = Boolean(
+      experience?.privacy_notices?.some(
+        (notice) => notice.current_preference == null
+      )
+    );
+    const modalLinkEl = document.getElementById(FIDES_MODAL_LINK);
+    if (!(noticesWithNoUserPreferenceExist || modalLinkEl)) {
```
allisonking, line 29: this is a little confusing to me—why do we skip rendering the overlay if there's no modal link?

clients/fides-js/src/lib/consent.tsx
```diff
@@ -20,6 +19,16 @@ export const initOverlay = async ({
   debugLog(options.debug, "Initializing Fides consent overlays...");
 
   async function renderFidesOverlay(): Promise<void> {
+    // Check if there are any notices within the experience that do not have a user preference
+    const noticesWithNoUserPreferenceExist: boolean = Boolean(
+      experience?.privacy_notices?.some(
+        (notice) => notice.current_preference == null
+      )
+    );
+    const modalLinkEl = document.getElementById(FIDES_MODAL_LINK);
```
allisonking, line 28: to keep the logic together, would it make sense to have the Overlay component do the `document.getElementById`?

clients/fides-js/src/components/Overlay.tsx
eastandwestwind: Its related. Since we now need to initialize the Overlay for the following use case:

* experience config `show_banner` is `false`
* experience is otherwise valid and has notices
* link el exists in DOM

I had to move this check for `show_banner` within the Overlay component itself

clients/fides-js/src/lib/consent.tsx
eastandwestwind: yes good idea

clients/fides-js/src/lib/consent.tsx
eastandwestwind: fixed- _only_ if we don't have new notices and modal link does not exist, we shouldn't render overlay

clients/fides-js/src/lib/consent.tsx
eastandwestwind: I have this logic here because I want to be sure we're not rendering the overlay to the DOM in the following case:

- no new notices exist
- link el does not exist in DOM

So that's why I'm trying to keep this check for link element as early as makes sense

clients/fides-js/src/components/ConsentBanner.tsx
```diff
@@ -19,8 +21,9 @@ const ConsentBanner: FunctionComponent<BannerProps> = ({
   onRejectAll,
   waitBeforeShow,
   onOpenModal,
+  bannerIsOpen,
+  setBannerIsOpen,
```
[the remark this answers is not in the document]
eastandwestwind: updated!

eastandwestwind:
Ready for you again @allisonking / @NevilleS

allisonking (approved):
nice job, this seems to work well!

clients/fides-js/src/lib/consent.tsx
```diff
@@ -20,6 +19,18 @@ export const initOverlay = async ({
...
+      experience?.privacy_notices?.some(
+        (notice) => notice.current_preference == null
+      )
+    );
+    const modalLinkId = options.modalLinkId || "fides-modal-link";
+    const modalLinkEl = document.getElementById(modalLinkId);
+    // If we don't have new notices and modal link does not exist, don't render overlay
+    if (!noticesWithNoUserPreferenceExist && !modalLinkEl) {
+      return Promise.resolve();
+    }
```
NevilleS, line 33: I have a feeling there's a bug here... it's a race condition. What if the link renders to the page after our render method fires?

I think the safest thing to do is not gate _any_ overlay rendering logic on the existence of that link, and throw it into a delayed interval - wait like 200ms and then check the DOM if it exists and add the click handler then, etc.

Sorry!

You can carve that out into a follow-up issue though, as long as it's not going to cause any other problems?

clients/sample-app/src/components/Home/index.tsx
```diff
@@ -69,9 +69,9 @@ const Home = ({ privacyCenterUrl, products }: Props) => {
           </a>
         </div>
         <div>
-          <a className={css.modalLink} href="#" id="fides-modal-link">
-            Manage preferences
-          </a>
+          <button type="button" className={css.modalLink} id="fides-modal-link">
```
NevilleS, line 72: ❤️ thanks for making it a button. You can remove the eslint-disable rule on line 1 in this file now, I added that to silence the linter!

NevilleS (approved):
LGTM. I'm pretty sure there's a bug here but let's carve that out into a known issue follow-up

clients/fides-js/src/lib/consent.tsx
eastandwestwind: We already wait until the DOM is completely loaded before calling our `renderFidesOverlay()` method within this file, so I think we should be safe?

eastandwestwind: I went ahead and create a ticket to track, but we can close out if not needed- https://github.com/ethyca/fides/issues/3483

PR #3470, opened 2023-06-06
[Backend] Further Deduplicate PrivacyExperience and PrivacyExperienceConfig 

### LINKED ISSUE #3465: Remove PrivacyExperienceHistory and Duplication b/t Experiences and Configs

A continuation of the removal of the "delivery mechanism" constraint -

Now that several of the constraints have been lifted between notices and experiences, we are duplicating too much information on PrivacyExperience that is also on PrivacyExperienceConfig.   The introduction of loading in ExperienceConfigs by default is highlighting how there are too many things to keep in sync between PrivacyExperiences and PrivacyExperienceConfigs.  We need to have one source of truth.

Our data model has PrivacyNotices, PrivacyExperiences (for a specific region) and have a FK to Privacy Experience Configs, which contain the copy.  There are a few duplicate fields on PrivacyExperiences and PrivacyExperienceConfigs which are hard to keep in sync, and we are unnecessarily versioning PrivacyExperience.

- Remove disabled, privacy_experience_config_history_id and version fields from PrivacyExperience. 
   - Look to the associated PrivacyExperienceConfig to determine if the resource is disabled.
- Remove the PrivacyExperienceHistory table.  Since we've made some changes that you can't change the type of config once it's created, there's no real need to version this experience any longer.
- When saving privacy preferences, save the privacy experience id rather than the privacyexperiencehistory id (which I'm removing)

### Additional context

pattisdr:
Closes #3465 

❗ Contains data migration; check downrev before merge

### Code Changes

Continue to deduplicate info between PrivacyExperience and PrivacyExperienceConfig.

- Remove common fields between the models like PrivacyExperience.disabled and PrivacyExperience.experience_config_history_id - these are at risk of getting out of sync
- Now that we've removed most of the fields from PrivacyExperience and it is largely just a linking table, I don't think we need to version this. Remove PrivacyExperience.version, and the entire PrivacyExperienceHistory table.
- PrivacyPreferenceHistory was referencing the PrivacyExperienceHistory id, but now that this doesn't exist, have it reference the PrivacyExperience id instead.
- When filtering PrivacyExperiences on disabled, actually filter the corresponding field on the PrivacyExperienceConfig resource.

### Steps to Confirm

* [ ] `nox -s dev -- shell`
* [ ] Get a privacy experience for a specific region, ex. {{host}}/privacy-experience/?region=eu_ee
* [ ] Note that version, disabled, and experience_config_history_id have been removed from the response, (but `version` and `disabled` are still present in the embedded `experience_config` object)
* [ ] Note both the "id" of that experience, and the historical `id` of one of the notices embedded in that experience
* [ ] Save privacy preferences with {{host}}/privacy-preferences. Note that fides_user_device_id must be a valid uuid4.  Note that a new field in this request is `privacy_experience_id`
```
{
   "browser_identity": {
       "ga_client_id": "UA-XXXXXXXXX",
       "ljt_readerID": "test_sovrn_id",
       "fides_user_device_id": "{{fides_user_device_id}}"
   },
   "preferences": [{
       "privacy_notice_history_id": "{{privacy_notice_history_id}}",
       "preference": "opt_out"
   }],
   "user_geography": "us_ca",
   "privacy_experience_id": "{{privacy_experience_id}}",
   "method": "button"
}
```
* [ ] Get the historical preferences {{host}}/historical-privacy-preferences and verify both the `privacy_notice_history_id` and `privacy_experience_id` as well as the `experience_config_history_id` which was lifted from the experience and added here for record keeping

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### Description Of Changes

Further side effects of https://github.com/ethyca/fides/pull/3387

Now that we've relaxed a lot of the constraints around privacy experiences we no longer need as many fields on the PrivacyExperience table.  There used to be necessary duplication between PrivacyExperience and PrivacyExperienceConfig but now this isn't required and we're just at risk of the tables getting out of sync.  This surfaced when testing "creating experiences out of the box - as an area that had gotten out of sync).  h/t to @adamsachs for also raising the importance of what tables were storing which information here: https://github.com/ethyca/fides/pull/3387#discussion_r1214456787

Primarily I deduplicate all the info between the two tables except for "component" and stop tracking history of privacy experiences now that there's barely anything on this table.

Saving privacy preferences should now pass on the "experience id" instead of the "experience history id" (because it doesn't exist)

### DIFF
--- a/.fides/db_dataset.yml
+++ b/.fides/db_dataset.yml
@@ -1474,15 +1474,6 @@ dataset:
     - name: created_at
       data_categories: [system.operations]
       data_qualifier: aggregated.anonymized.unlinked_pseudonymized.pseudonymized.identified
-    - name: delivery_mechanism
-      data_categories: [system.operations]
-      data_qualifier: aggregated.anonymized.unlinked_pseudonymized.pseudonymized.identified
-    - name: disabled
-      data_categories: [system.operations]
-      data_qualifier: aggregated.anonymized.unlinked_pseudonymized.pseudonymized.identified
-    - name: experience_config_history_id
-      data_categories: [ system.operations ]
-      data_qualifier: aggregated.anonymized.unlinked_pseudonymized.pseudonymized.identified
     - name: experience_config_id
       data_categories: [system.operations]
       data_qualifier: aggregated.anonymized.unlinked_pseudonymized.pseudonymized.identified
@@ -1493,9 +1484,6 @@ dataset:
       data_categories: [system.operations]
       data_qualifier: aggregated.anonymized.unlinked_pseudonymized.pseudonymized.identified
     - name: updated_at
-      data_categories: [system.operations]
-      data_qualifier: aggregated.anonymized.unlinked_pseudonymized.pseudonymized.identified
-    - name: version
       data_categories: [system.operations]
       data_qualifier: aggregated.anonymized.unlinked_pseudonymized.pseudonymized.identified
   - name: privacyexperienceconfig
@@ -1573,33 +1561,6 @@ dataset:
     - name: save_button_label
       data_categories: [ system.operations ]
     - name: title
-      data_categories: [system.operations]
-    - name: updated_at
-      data_categories: [system.operations]
-    - name: version
-      data_categories: [system.operations]
-  - name: privacyexperiencehistory
-    description: 'Stores the history of an experience for a region for consent reporting'
-    data_categories: []
-    data_qualifier: aggregated.anonymized.unlinked_pseudonymized.pseudonymized.identified
-    fields:
-    - name: component
-      data_categories: [system.operations]
-    - name: created_at
-      data_categories: [system.operations]
-    - name: delivery_mechanism
-      data_categories: [system.operations]
-    - name: disabled
-      data_categories: [system.operations]
-    - name: experience_config_history_id
-      data_categories: [system.operations]
-    - name: experience_config_id
-      data_categories: [system.operations]
-    - name: id
-      data_categories: [system.operations]
-    - name: privacy_experience_id
-      data_categories: [system.operations]
-    - name: region
       data_categories: [system.operations]
     - name: updated_at
       data_categories: [system.operations]
@@ -2199,7 +2160,7 @@ dataset:
         data_categories:
         - system.operations
         data_qualifier: aggregated.anonymized.unlinked_pseudonymized.pseudonymized.identified
-      - name: privacy_experience_history_id
+      - name: privacy_experience_id
         data_categories:
         - system.operations
         data_qualifier: aggregated.anonymized.unlinked_pseudonymized.pseudonymized.identified
--- a/src/fides/api/api/v1/endpoints/privacy_preference_endpoints.py
+++ b/src/fides/api/api/v1/endpoints/privacy_preference_endpoints.py
@@ -37,7 +37,7 @@ from fides.api.api.v1.urn_registry import (
 )
 from fides.api.ctl.database.seed import DEFAULT_CONSENT_POLICY
 from fides.api.models.fides_user import FidesUser
-from fides.api.models.privacy_experience import PrivacyExperienceHistory
+from fides.api.models.privacy_experience import PrivacyExperience
 from fides.api.models.privacy_notice import PrivacyNotice, PrivacyNoticeHistory
 from fides.api.models.privacy_preference import (
     CurrentPrivacyPreference,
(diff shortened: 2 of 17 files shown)


.fides/db_dataset.yml
```diff
@@ -1474,15 +1474,6 @@ dataset:
     - name: created_at
       data_categories: [system.operations]
       data_qualifier: aggregated.anonymized.unlinked_pseudonymized.pseudonymized.identified
-    - name: delivery_mechanism
-      data_categories: [system.operations]
-      data_qualifier: aggregated.anonymized.unlinked_pseudonymized.pseudonymized.identified
```
pattisdr, line 1479: Ah this should have been removed earlier, note to self that the fides scan doesn't detect extra fields

src/fides/api/api/v1/endpoints/privacy_preference_endpoints.py
```diff
@@ -205,25 +205,29 @@ def _get_request_origin_and_config(
...
+        if not privacy_experience:
             raise HTTPException(
                 status_code=HTTP_404_NOT_FOUND,
-                detail=f"Privacy Experience History '{data.privacy_experience_history_id}' not found.",
+                detail=f"Privacy Experience '{data.privacy_experience_id}' not found.",
+            )
+        experience_config_id = privacy_experience.experience_config_id
+        if experience_config_id:
+            experience_config_history_id = (
+                privacy_experience.experience_config.experience_config_history_id
```
pattisdr, line 222: When saving privacy preferences, I now need the frontend to pass in the "privacy_experience_id" instead of the "privacy_experience_history_id" (which no longer exists).  We can't derive the experience from the notice alone, so this extra field lets us look up the experience config id that would have been used to surface the notices to the customer and save them for record keeping.

src/fides/api/api/v1/endpoints/privacy_preference_endpoints.py
```diff
@@ -529,8 +533,8 @@ def get_historical_consent_report(
             ),
             PrivacyPreferenceHistory.url_recorded.label("url_recorded"),
             PrivacyPreferenceHistory.user_agent.label("user_agent"),
-            PrivacyPreferenceHistory.privacy_experience_history_id.label(
-                "privacy_experience_history_id"
+            PrivacyPreferenceHistory.privacy_experience_id.label(
+                "privacy_experience_id"
```
pattisdr, line 537: Similarly the historical report now surfaces the experience id instead of the experience history id.  This isn't that informative of a property, as it only stores the region and component, but I want to record it in case we need to filter on it.  This was, after all, the primary id of the resource that housed everything we displayed to the end user.

src/fides/api/ctl/migrations/versions/317e6197c76a_reduce_experience_and_config_duplication.py
```diff
@@ -0,0 +1,250 @@
...
+down_revision = "b8248ce6b8a1"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.drop_index(
+        "ix_privacyexperiencehistory_experience_config_history_id",
+        table_name="privacyexperiencehistory",
+    )
```
pattisdr, line 23: Note, lots of rearranging had to be done in this migration to make sure we were dropping indices/constraints before tables.

src/fides/api/models/privacy_experience.py
```diff
@@ -175,68 +175,38 @@ class PrivacyExperienceConfigHistory(ExperienceConfigBase, Base):
...
-    )  # Also links to the historical record, so if the version of config gets updated, that
-    # triggers a new version of the experience.
-
     # There can only be one component per region
     __table_args__ = (UniqueConstraint("region", "component", name="region_component"),)
 
-    histories = relationship(
-        "PrivacyExperienceHistory", backref="privacy_experience", lazy="dynamic"
-    )
-
```
pattisdr, line 222: I don't think there's any reason to track PrivacyExperienceHistory any longer, so I'm getting rid of this table and associated `relationship` properties.  Similarly there is no "version".  

"experience_config_history_id" was only here originally for versioning purposes, but the primary purpose of this PR, is that I'd already seen an example of where experience_config_id and experience_config_history_id were mismatched as I hadn't called the code to sync experiences to configs when the config was updated on startup.

src/fides/api/models/privacy_experience.py
```diff
@@ -314,27 +284,6 @@ def get_related_privacy_notices(
...
-        # create the history after the initial object creation succeeds, to avoid
-        # writing history if the creation fails and so that we can get the generated ID
-        data.pop("id", None)
-        history_data = {
-            **data,
-            "privacy_experience_id": privacy_experience.id,
-        }
-        PrivacyExperienceHistory.create(db, data=history_data, check_name=False)
-        return privacy_experience
-
```
pattisdr, line 337: I don't think we need any special logic here for create any longer, same for update below.

src/fides/api/models/privacy_experience.py
```diff
@@ -175,68 +175,38 @@ class PrivacyExperienceConfigHistory(ExperienceConfigBase, Base):
...
-class PrivacyExperience(PrivacyExperienceBase, Base):
+class PrivacyExperience(Base):
     """Stores Privacy Experiences for a given just a single region.  The Experience describes how to surface
     multiple Privacy Notices to the end user in a given region.
 
     There can only be one component per region.
     """
 
+    component = Column(EnumColumn(ComponentType), nullable=False)
```
pattisdr, line 185: Component is the last remaining duplicated field between PrivacyExperience and PrivacyExperienceConfig.  There's a lot of logic centered around this field as well as a FK constraint so I propose leaving this one.  We just need to make sure they stay in sync. I do have a prevention from keeping ExperienceConfig component types from changing after they're created which helps a lot.

src/fides/api/ctl/migrations/versions/317e6197c76a_reduce_experience_and_config_duplication.py
adamsachs: better now than when these tables are actually in use by customers 😄

src/fides/api/models/privacy_experience.py
adamsachs: yup, this seems fine. it feels like the `component` is fundamental to both the `PrivacyExperience` and the `ExperienceConfig`. And like you said - it's effectively _fixed_  (immutable) for records of both models.

src/fides/api/models/privacy_experience.py
adamsachs: i like these simplifications a lot! they felt a bit overkill to me 😛

adamsachs (approved):
this looks great @pattisdr, nothing stood out to me as as a concern or in need of adjustment. i'm very happy to see things going in this direction, i think this has simplified the db models as well as the application logic significantly, and i don't think we lose any meaningful functionality as far as i can tell. nice job getting this in place before we roll it out 👍

pattisdr:
appreciate the quick review @adamsachs 🙏

Issue #3485, opened 2023-06-07
Fix false positives in consent-banner.cy.ts cypress tests


eastandwestwind:
@allisonking reported that the following is running on every test, even though it should be scoped to one test:

```
Cypress.on("fail", (error) => {
          if (error.message.indexOf("Timed out retrying") !== 0) {
            throw error;
          }
        });
```

allisonking:
Closed by https://github.com/ethyca/fides/pull/3487

Roger-Ethyca:
moving to done

PR #3487, opened 2023-06-07
Fix consent-banner cypress tests

allisonking:
Closes https://github.com/ethyca/fides/issues/3485

### Code Changes

### Steps to Confirm

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

### DIFF
--- a/clients/privacy-center/cypress/e2e/consent-banner.cy.ts
+++ b/clients/privacy-center/cypress/e2e/consent-banner.cy.ts
@@ -543,7 +543,7 @@ describe("Consent banner", () => {
       });
       it("does not set user consent preference automatically", () => {
         // timeout means API call not made, which is expected
-        Cypress.on("fail", (error) => {
+        cy.on("fail", (error) => {
           if (error.message.indexOf("Timed out retrying") !== 0) {
             throw error;
           }
@@ -604,7 +604,7 @@ describe("Consent banner", () => {
 
       it("does not set user consent preference automatically", () => {
         // timeout means API call not made, which is expected
-        Cypress.on("fail", (error) => {
+        cy.on("fail", (error) => {
           if (error.message.indexOf("Timed out retrying") !== 0) {
             throw error;
           }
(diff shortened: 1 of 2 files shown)


clients/privacy-center/cypress/e2e/consent-banner.cy.ts
```diff
@@ -946,12 +946,11 @@ describe("Consent banner", () => {
       });
 
       it("closes banner and opens modal when modal link is clicked", () => {
-        cy.get("div#fides-banner").should("exist");
+        cy.get("div#fides-banner").should("be.visible");
```
allisonking, line 949: this is a little subtle—the banner always exists, we just move it out of the screen so that we can animate it. the odd thing is that without this change, the env would end up as

image

showing _both_ the modal and overlay. I think what is happening is

1. the page loads, the banner does not appear yet because it is still animating in
2. cypress clicks on the modal link, so the modal opens (also sets that the banner should not be open, but I think at this point the banner isn't technically open yet because of [`delayBanner`](https://github.com/ethyca/fides/blob/aking/3485/fix-cy-tests/clients/fides-js/src/components/Overlay.tsx#L40-L45))
3. now the banner has finished animating so both the modal and banner are open!

by waiting for the banner to be visible, that delays cypress enough so that the banner is set to open, then the logic all works accordingly.

this is probably a bug, so maybe it should be skipped instead of the fix I put in 🤔 let me know what you think @eastandwestwind . I think if a user were very speedy, they could trigger this by clicking on the modal link before the banner has appeared

eastandwestwind: I think this cypress adjustment is fine for now, as the odds that a user immediately clicks on consent (usually in the footer of a site) before the banner slides in is slim. 

I'll create a follow-up for this bug fix.

eastandwestwind: https://github.com/ethyca/fides/issues/3488

eastandwestwind (approved):

PR #3492, opened 2023-06-07
Parallelize Docker Publishing

### LINKED ISSUE #3422: Parallelize docker publishing

The Docker publish job takes almost 2 hours in some cases, or an hour at best

Parallelize the commands so that we can build the 3 different images separately

Example Workflow: https://github.com/ethyca/fides/actions/runs/5203045330/jobs/9385379072

ThomasLaPiana:
Closes #3422 

### Code Changes

* [x] update the `nox -s push` session to use multiprocessing

### Steps to Confirm

* [ ] run `nox -s "push(dev)"` locally and confirm the output looks wild (like 6 images being built at once) and that the job finishes (might fail on `push` step of the command)

### Pre-Merge Checklist

* [ ] ~~Update `CHANGELOG.md`~~

### Description Of Changes

I thought about adding an additional parameter to build each one separately, i.e. `nox -s "push(git-tag, fides)" "push(git-tag, sample)"` etc. but felt this would also add additional complication to the CI workflows. I also couldn't think of a case where we _wouldn't_ want to build everything at once, so this seemed like the lowest-friction solution.

### DIFF
--- a/noxfiles/docker_nox.py
+++ b/noxfiles/docker_nox.py
@@ -1,6 +1,6 @@
 """Contains the nox sessions for docker-related tasks."""
 import platform
-from typing import List, Tuple
+from typing import List, Tuple, Dict, Callable
 from multiprocessing import Pool
 from subprocess import run
 from functools import partial
@@ -193,11 +193,9 @@ def session_runner(command: Tuple[str, ...], session: nox.Session) -> None:
     session.run(*command, external=True)
 
 
-def push_fides_images(session: nox.Session, tag_suffixes: List[str]) -> None:
+def get_buildx_commands(tag_suffixes: List[str]) -> List[Tuple[str, ...]]:
     """
     Build and publish each image to Dockerhub
-
-    Builds for multiple platforms and does so in parallel.
     """
     fides_tags = [f"{IMAGE}:{tag_suffix}" for tag_suffix in tag_suffixes]
     fides_buildx_command = generate_multiplatform_buildx_command(
@@ -226,9 +224,7 @@ def push_fides_images(session: nox.Session, tag_suffixes: List[str]) -> None:
         privacy_center_buildx_command,
         sample_app_buildx_command,
     ]
-    number_of_processes = len(buildx_commands)
-    with Pool(number_of_processes) as process_pool:
-        process_pool.map(partial_run, buildx_commands)
+    return buildx_commands
 
 
 @nox.session()
@@ -274,7 +270,7 @@ def push(session: nox.Session, tag: str) -> None:
     )
 
     # Use lambdas to force lazy evaluation
-    param_tag_map = {
+    param_tag_map: Dict[str, Callable] = {
         "dev": lambda: [DEV_TAG_SUFFIX],
         "prerelease": lambda: [PRERELEASE_TAG_SUFFIX],
         "rc": lambda: [RC_TAG_SUFFIX],
@@ -282,4 +278,10 @@ def push(session: nox.Session, tag: str) -> None:
         "prod": lambda: [get_current_tag(), "latest"],
     }
 
-    push_fides_images(session=session, tag_suffixes=param_tag_map[tag]())
+    # Get the list of Tupled commands to run
+    buildx_commands = get_buildx_commands(tag_suffixes=param_tag_map[tag]())
+
+    # Parallel build the various images
+    number_of_processes = len(buildx_commands)
+    with Pool(number_of_processes) as process_pool:
+        process_pool.map(partial_run, buildx_commands)
--- a/noxfiles/docker_nox.py
+++ b/noxfiles/docker_nox.py
@@ -1,6 +1,9 @@
 """Contains the nox sessions for docker-related tasks."""
 import platform
 from typing import List, Tuple
+from multiprocessing import Pool
+from subprocess import run
+from functools import partial
 
 import nox
 from constants_nox import (
@@ -22,6 +25,7 @@ DOCKER_PLATFORM_MAP = {
     "x86_64": "linux/amd64",
 }
 DOCKER_PLATFORMS = "linux/amd64,linux/arm64"
+partial_run = partial(run, shell=True, check=True)
 
 
 def verify_git_tag(session: nox.Session) -> str:
@@ -184,15 +188,22 @@ def build(session: nox.Session, image: str, machine_type: str = "") -> None:
     session.run(*build_command, external=True)
 
 
+def session_runner(command: Tuple[str, ...], session: nox.Session) -> None:
+    """Utility function for use with multiprocessing."""
+    session.run(*command, external=True)
+
+
 def push_fides_images(session: nox.Session, tag_suffixes: List[str]) -> None:
-    # Publish Fides
+    """
+    Build and publish each image to Dockerhub
+
+    Builds for multiple platforms and does so in parallel.
+    """
     fides_tags = [f"{IMAGE}:{tag_suffix}" for tag_suffix in tag_suffixes]
     fides_buildx_command = generate_multiplatform_buildx_command(
         image_tags=fides_tags, docker_build_target="prod"
     )
-    session.run(*fides_buildx_command, external=True)
 
-    # Publish the Privacy Center
     privacy_center_tags = [
         f"{PRIVACY_CENTER_IMAGE}:{tag_suffix}" for tag_suffix in tag_suffixes
     ]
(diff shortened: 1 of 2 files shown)


NevilleS:
No way to do this with a `matrix` in GitHub actions? That way you could monitor the jobs separately in the actions UI?

noxfiles/test_docker_nox.py
```diff
@@ -60,13 +68,15 @@ def test_different_path(self) -> None:
         )
         assert actual_result == expected_result
 
+
+class TestGetBuildxCommands:
```
ThomasLaPiana, line 72: @adamsachs I changed up the tests here and would like your thoughts!

I made a major change away from the mock-based tests for a few reasons:
1. I am honestly biased against mocking for the most part. I think mocking is a bit of a crutch to avoid rewriting code in a more testable manner (weakly held strong belief). Not saying that's why you wrote mock-based tests, just saying why I shy away from it
2. It tested more of the execution in my opinion instead of focus testing the command/generation output
3. Perhaps most importantly, mocking gets complicated, and the things I would've had to mock to get it to "fake" execute given the new complexity of multiprocessing didn't feel right. Simpler and more concise tests felt _better_ to me, subjectively in this case

All that being said, I also don't want to inadvertently leave any testing gaps open. Please check me here if you think I've missed out on testing functionality and I've regressed our test coverage

ThomasLaPiana:
> No way to do this with a `matrix` in GitHub actions? That way you could monitor the jobs separately in the actions UI?

There is definitely a way to do that, but I had some reservations:
1. The CI file would get _even more complex_ since it wouldn't be straightforward matrix processing. Everything in there is conditional based on the tag matching steps, and it would make everything 3x more gnarly
2. This method also lets users benefit when running these commands locally
3. This method is more scalable. What if we add a 4th image down the line? No worries, update the `docker_nox` file and we're good to go. Otherwise, see #1 about having to do _even more_ CI workflow wizardry
4. I think this is actually a feature not a bug. If one fails, at some point, we probably want to stop them all? I know it won't always work perfectly since some images might finish before others, and the final one might fail at the final step when the others have already been pushed, but it's better than nothing? I can't think of a time we'd want the PC and Sample app image to push even though the Fides image failed early on.

I understand there is definitely a clarity benefit to doing it via a matrix and letting GH Actions handle the parallelism instead of Python, but after weighing the pros and cons I went with the Python solution. Absolutely still open for discussion though 🙂

SteveDMurphy:
I've been hitting some failures around the builder this morning, even having run `docker buildx rm` & `nox -s clean` then it recreates and exits without building - very well could be a me problem however 🤔 

I like the idea of having the commands run locally as well, and agree with nearly everything you said above! _but_ am curious if there is not some benefit in using the [existing build action that seemed to run quickly and keep the CI file fairly clean](https://github.com/ethyca/fidesplus/pull/630/files#diff-b9be73d696410d1d77d08986ae35f3c62dea59f7592790d90559f5cb1f0681d2R39-R47). We could potentially rely on a local command to push individual images as needed for errors etc. but majority of this work should happen in CI so it would be the exception vs the rule I think. Curious to hear more on what you all think !

adamsachs:
> I've been hitting some failures around the builder this morning, even having run docker buildx rm & nox -s clean then it recreates and exits without building - very well could be a me problem however 🤔

i'm also hitting this i think - at least very similar symptoms. i've dug on it, and i've narrowed it down to something pretty odd:
- the output from each of my 3 parallel processes is just the output that one gets when simply running the `docker` command, with no args -- i.e, i get the usage docs
- with that, i inferred that for some reason, each subprocess is only running `docker` - and not the remaining list of args that we're intending to pass in (as tuples). i.e. `subprocess.run(args, shell=True, external=True)` where `args` is a tuple of CLI args was _not_ working as expected.
- if i update the parallelized process to `join` the tuple into a space-delimited string, then _that_ seems to correctly invoke the 3 parallelized `buildx` commands. here's what i replaced the `partial_run` function with, just for testing -- and it got things working:
```
def runner(args):
    args_str = " ".join(args)
    run(args_str, shell=True, check=True)
```

@ThomasLaPiana was the code as you've got it here in the PR working well for you? if so, perhaps this is a mac (or ARM?) specific thing with `subprocess.run`?  it seems odd though, passing a tuple of args to `subprocess.run` seems like a supported way of doing things. i even tried converting those tuples to lists (which is exactly what the [docs](https://docs.python.org/3/library/subprocess.html#subprocess.run) say is supported) and still no dice. i'm a bit stumped. we could of course adjust use the single string as i have above, but it feels like a bit of a hack and i'm a bit unsettled...

ThomasLaPiana:
> > I've been hitting some failures around the builder this morning, even having run docker buildx rm & nox -s clean then it recreates and exits without building - very well could be a me problem however 🤔
> 
> i'm also hitting this i think - at least very similar symptoms. i've dug on it, and i've narrowed it down to something pretty odd:
> 
> * the output from each of my 3 parallel processes is just the output that one gets when simply running the `docker` command, with no args -- i.e, i get the usage docs
> * with that, i inferred that for some reason, each subprocess is only running `docker` - and not the remaining list of args that we're intending to pass in (as tuples). i.e. `subprocess.run(args, shell=True, external=True)` where `args` is a tuple of CLI args is _not_ working as expected.
> * if i update the parallelized process to `join` the tuple into a space-delimited string, then _that_ seems to correctly invoke the 3 parallelized `buildx` commands. here's what i replaced the `partial_run` function with, just for testing -- and it got things working:
> 
> ```
> def runner(args):
>     args_str = " ".join(args)
>     run(args_str, shell=True, check=True)
> ```
> 
> @ThomasLaPiana was the code as you've got it here in the PR working well for you? if so, perhaps this is a mac (or ARM?) specific thing with `subprocess.run`? it seems odd though, passing a tuple of args to `subprocess.run` seems like a supported way of doing things. i even tried converting those tuples to lists (which is exactly what the [docs](https://docs.python.org/3/library/subprocess.html#subprocess.run) say is supported) and still no dice. i'm a bit stumped. we could of course adjust use the single string as i have above, but it feels like a bit of a hack and i'm a bit unsettled...

You're right on the money. Yes this is a specific mac issue, I thought I had properly copy/pastad from elsewhere to avoid this issue but it looks like I walked right into it again hah. I forgot the `join` part!

The `.join` solution is exactly correct 🙂 see what we do in `deploy.py` for instance

noxfiles/docker_nox.py
```diff
@@ -184,25 +188,28 @@ def build(session: nox.Session, image: str, machine_type: str = "") -> None:
     session.run(*build_command, external=True)
 
 
-def push_fides_images(session: nox.Session, tag_suffixes: List[str]) -> None:
-    # Publish Fides
+def session_runner(command: Tuple[str, ...], session: nox.Session) -> None:
```
adamsachs, line 191: i don't _think_ this is being used? perhaps this was part of a previous iteration and was left in by accident? i think the `partial_run` takes its place if i'm not mistaken

adamsachs (commented):
i really like this! leaving a few inline code comments that are self explanatory. besides that, some broader points:

- per the above note, looks like we need an adjustment to get this working in mac/unix shells (i think this is a blocker assuming the GH runner is unix and has similar problems to what @SteveDMurphy and i are seeing locally). i can push up a proposed fix.
- i appreciated the pros/cons discussion above re: @NevilleS 's GH matrix suggestion. fwiw, i think i'd actually lean toward the GH matrix myself, but i don't have a strong enough opinion to change what we've got here -- so i think let's go with what you've got here for now, and we can adjust to a GH matrix in a future iteration if we're finding a strong need to
- i'm curious to understand @SteveDMurphy's suggestion a bit better. i don't know if i understand the consequences of that approach well enough.

all in all, i'm going to push up a commit or two that i think should get this just about ready to go, and i think it's a huge improvement. there do seem to be some alternatives to explore, but IMO i'd like to get this in as an incremental improvement that should hopefully cut down our build times immediately, and then evaluate the alternatives after the fact.

noxfiles/docker_nox.py
adamsachs: removed in https://github.com/ethyca/fides/pull/3492/commits/efe2d0f8b163f34ad237d170d79e26248f8d3eb0 - let me know if i was mistaken, and we can easily add back

noxfiles/docker_nox.py
```diff
@@ -257,12 +270,18 @@ def push(session: nox.Session, tag: str) -> None:
...
         "prod": lambda: [get_current_tag(), "latest"],
     }
 
-    push_fides_images(session=session, tag_suffixes=param_tag_map[tag]())
+    # Get the list of Tupled commands to run
+    buildx_commands = get_buildx_commands(tag_suffixes=param_tag_map[tag]())
+
+    # Parallel build the various images
+    number_of_processes = len(buildx_commands)
+    with Pool(number_of_processes) as process_pool:
```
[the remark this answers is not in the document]
ThomasLaPiana: It looks like the default setup is 2 cores and 7gb of memory 🤔 we'll have to see I guess how much of a difference this makes

noxfiles/docker_nox.py
ThomasLaPiana: correct! That was before I added the `partial` solution!

ThomasLaPiana:
job failures were unrelated, the same intermittent image bug

PR #3496, opened 2023-06-07
Updating down_revision for conflicting migration

galvana:
### Code Changes

* [ ] Updating down revision for conflicting migration script


SteveDMurphy (approved):
assuming checks will pass, thanks for fixing this! Thought I had checked right before merging that it was up to date as well 😞

Issue #3309, opened 2023-05-15
Accessibility requirements for `fides-js`


allisonking:
### Is your feature request related to a specific problem?

We need to make sure our `fides-js` components are as accessible as possible!

### Describe the solution you'd like

Here's a running checklist of things we need to do. I'm not 100% sure if these are strictly requirements are suggestions, so we should do our due diligence with reading the standards, but for now I'm going to drop everything I see.

These don't include standard a11y things, like high contrast text and heading logic, as those things can typically be caught by automated tools (https://github.com/ethyca/fides/issues/3306)

#### Banner
* [x] Banner should be the first focusable element in the page, before skip link and website's nav (so that visitors interact with the banner before anything else)
* [x] Banner buttons must be of type `<button>` 
* [x] When banner closes, we should not be able to tab through it

#### Modals
* [x] Trap focus inside the modal (cannot tab outside the modal)
* [x] Modal size should allow text to be readable at all screen sizes
* [x] Initial focus should be set within the modal (often to the first focusable element)
* [x] Closing a modal (either through confirmation or closing) returns focus to the last focused item on the main screen
* [x] ESC closes the dialog
* [x] Can Tab + Shift Tab through the modal
* [x] Strongly recommended to include a close icon or button
* If the modal is meant to operate such that users cannot interact with anything outside of the modal and dims the background:
  * [x] Element that serves as the dialog container has a role of `dialog`
  * [x] All elements required to operate the dialog are descendants of the element that has role `dialog`
  * [x] Dialog container element has `aria-modal` set to true
* [x] The dialog has either 
  * A value set for the `aria-labelledby property that refers to a visible dialog title
  * A label specified by `aria-label`

#### Toggles
* [x] SPACE changes the state of the switch
* [x] Switch has the role `switch`
* [x] Switch has an accessible label by one of the following:
  * Visible text content contained within the element with role switch.
  * A visible label referenced by the value of [aria-labelledby](https://w3c.github.io/aria/#aria-labelledby) set on the element with role switch.
  * [aria-label](https://w3c.github.io/aria/#aria-label) set on the element with role switch.
* [x] When on, the switch element has state [aria-checked](https://w3c.github.io/aria/#aria-checked) set to true.
* [x] When off, the switch element has state [aria-checked](https://w3c.github.io/aria/#aria-checked) set to false.
* [x] If the switch element is an HTML input[type="checkbox"], it uses the HTML checked attribute instead of the aria-checked property.

#### Disclosures (accordions)
[...]

PR #3510, opened 2023-06-08
Fides-js accessibility wrap up

allisonking:
Closes https://github.com/ethyca/fides/issues/3309

### Code Changes

The bulk of the a11y work was done while developing, but this should cover the straggling items.

* [x] Makes the banner not come up in the tab order when it's hidden
* [x] Prepend the overlay instead of appending so that the banner is the first tabbed to item

### Steps to Confirm

* Bring up either cookie house or `fides-js-components-demo.html`
* As a keyboard user, tab into the page. You should immediately be tabbing through the banner (before, you'd have to tab through the whole page before getting to the banner. this is far more noticeable on cookie house than the demo page)
* Close the banner. Tab through the page. You should be able to tab through without issue (before, there would be a few tabs where it looked like nothing was happening, since it was tabbing through the hidden banner)

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### Description Of Changes

I went through a few sites, and I don't think there's actual regulation on where a banner should appear in a tab order. [This blog post](https://uxdesign.cc/cookie-banners-and-accessibility-d476bf9ee4fc) recommends having them come up as soon as possible, though I ran into some sites that have them at the very end (where you have to tab through the entire page, including footers, before you can get to the banner. example: https://stackoverflow.com/) but other sites make sure they're the first thing you tab through. example: https://liveramp.com/). I thought it made sense to have it come up as soon as possible, though this means we `document.body.prepend` instead of `document.body.appendChild`. In other words, the overlay becomes the top DOM element. Because of this, I also gave the overlay an overall high z-index to increase the chance we stay on top of the host site's elements (we probably wanted this anyway)

There might be other ways to have the banner receive focus first other than doing a prepend:
* use the `autoFocus` attr on a button in the banner, however this apparently has [a11y drawbacks](https://www.boia.org/blog/accessibility-tips-be-cautious-when-using-autofocus) (vscode warns against it)
* use a [portal](https://github.com/developit/preact-portal). I don't _think_ we need one, since we're already controlling where in the DOM we are injecting our overlay, but I think that might be a standard way of doing this kind of thing. however, adding preact-portal is another >1kb! 

I haven't come across any drawbacks with prepending instead of appending yet, but we'll see!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,9 @@ The types of changes are:
 
 ## [Unreleased](https://github.com/ethyca/fides/compare/2.15.0...main)
 
+### Fixed
+- Remove the `fides-js` banner from tab order when it is hidden and move the overlay components to the top of the tab order. [#3510](https://github.com/ethyca/fides/pull/3510)
+
 
 ## [2.15.0](https://github.com/ethyca/fides/compare/2.14.1...2.15.0)
 
--- a/clients/fides-js/src/components/fides.css
+++ b/clients/fides-js/src/components/fides.css
@@ -77,6 +77,7 @@ div#fides-banner-container {
   transition: transform 1s;
   display: flex;
   justify-content: center;
+  visibility: visible;
 }
 
 div#fides-banner {
@@ -105,6 +106,7 @@ div#fides-banner-container.fides-banner-bottom {
 
 div#fides-banner-container.fides-banner-bottom.fides-banner-hidden {
   transform: translateY(150%);
+  visibility: hidden;
 }
 
 div#fides-banner-container.fides-banner-top {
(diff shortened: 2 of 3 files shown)


eastandwestwind:
Manually testing this- just to confirm, a side effect of this is that the banner no longer animates when it is closed. This is fine by me (and makes more sense than animating it when the user wishes it to be closed), but want to make sure it's expected.

allisonking:
> Manually testing this- just to confirm, a side effect of this is that the banner no longer animates when it is closed. This is fine by me (and makes more sense than animating it when the user wishes it to be closed), but want to make sure it's expected.

ahh good catch, no that wasn't intentional! I put in a quick fix so that we get the animation when closing again. if we don't want it in the future, we can revert https://github.com/ethyca/fides/pull/3510/commits/0998729464af692b59b1aea6fcb84cf296924099 but for now I didn't want to add something unintentionally :)

eastandwestwind (approved):

Issue #3514, opened 2023-06-09
Consent modal doesn't show saved preferences until reloading


NevilleS:
### Bug Description

If you adjust preferences on the consent modal, everything *saves* correctly but if you close and reopen the model it doesn't reflect your updated preferences.

#### Steps to Reproduce

1. Clear all your cookies from cookiehouse-plus-rc.fides-staging.ethyca.com
2. Open https://cookiehouse-plus-rc.fides-staging.ethyca.com/?geolocation=US-CA
3. Close the banner (if it's enabled)
4. Click "Manage Preferences" to open the modal
5. Click "Opt-out of all"
6. Confirm `window.Fides.consent.data_sales_and_sharing === false`, which is true!
7. Reopen the modal
8. Observe that the "Data Sales" toggle incorrectly shows as opted-in
9. Reload the page & reopen the modal
10. Observe that the "Data Sales" toggle shows as opted-out now

### Expected behavior

Whenever a preference changes from saving/accept all/reject all, reloading the modal should show the newly saved preference.

### Screenshots

https://www.loom.com/share/f77701b658d242f7a7574523ea98dcc3

### Environment

* Version: 2.15.0rc0

Roger-Ethyca:
moving to done

PR #3522, opened 2023-06-09
Fix bug where modal does not reflect draft state

allisonking:
Closes https://github.com/ethyca/fides/issues/3514

### Code Changes

* [x] Hoist state to the overlay. The banner can actually also control the state of the toggles, so it's best to have the parent control the state. Refactor some of the save functions to be more streamlined
* [x] Add a test for this case

### Steps to Confirm

* See repro steps in the original ticket

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### Description Of Changes

Incidentally, I believe this is a regression after we incorporated the a11y-dialog library. Before, we were unmounting the Modal when we hid it. The library hides the modal instead, so the component doesn't unmount. However, we had some logic [which checked the window object in the modal](https://github.com/ethyca/fides/blob/main/clients/fides-js/src/components/ConsentModal.tsx#L32-L38) and set the toggle states that way. That bit of logic doesn't retrigger since the component doesn't mount again, so it becomes stale.

Now that we are only hiding the component, we have to be a little more careful with how we control the state. I think this way makes a bit more intuitive sense, or at least it is more explicit about tracking state.

https://github.com/ethyca/fides/assets/24641006/0c111b23-dacc-4900-882d-48c4515f3998

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@ The types of changes are:
 - Fix CLI output showing a version warning for Snowflake [#3434](https://github.com/ethyca/fides/pull/3434)
 - Flaky custom field Cypress test on systems page [#3408](https://github.com/ethyca/fides/pull/3408)
 - Fix NextJS errors & warnings for Cookie House sample app [#3411](https://github.com/ethyca/fides/pull/3411)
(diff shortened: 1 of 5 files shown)


eastandwestwind (approved):
Manually tested and looks great! Thanks @allisonking !

PR #3528, opened 2023-06-11
Remove Deprecated Data Uses

### LINKED ISSUE #3515: Legacy/deprecated data uses still exist post-1.4 migration
### Bug Description

When I update to our `2.15.0rc0` release, we've an automated migration that runs to add the new `fideslang` 1.4 data uses, notably the revised `marketing.*` taxonomy as see here in the system panel:
image

However, I'm also still able to browse and select legacy data uses like `advertising` (which was migrated to `marketing.advertising`):
image

This is likely to attempt to be less destructive, but it's confusing to have both options now.

#### Steps to Reproduce

1. Visit https://fides-plus-rc.fides-staging.ethyca.com/systems/configure/google_ads
2. Add a new Data Use and open the "Data use" select to browse the taxonomy
3. Observe that the `advertising.*` taxonomy still exists

### Expected behavior

The legacy data uses should no longer be available in the system.

### Screenshots

See description

### Environment

* Version: 2.15.0rc0

ThomasLaPiana:
Closes #3515 

### Code Changes

* [x] add a migration to remove the old data uses

### Steps to Confirm

* [ ] `git checkout 2.14.1; nox -s teardown -- volumes; nox -s dev` - let the server get into a running state before the next commands
* [ ] `fides user login; fides get data_use advertising` - the command should pass
* [ ] spin down the docker containers (but _do not teardown the volumes!_)
* [ ] `git checkout ThomasLaPiana-remove-deprecated-taxonomy; nox -s dev`
* [ ] `fides user login;  fides get data_use advertising` - the command should come back with an empty response
* [ ] Finally, `fides get data_use marketing` - the command should pass, and means the new data uses have been properly seeded (this is handled outside of the migration, by `seed.py`)

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### Description Of Changes

Quick PR to remove the "old" default data uses in favor of the new ones

### DIFF
--- a/src/fides/api/ctl/migrations/versions/5307999c0dac_remove_deprecated_data_uses_for_.py
+++ b/src/fides/api/ctl/migrations/versions/5307999c0dac_remove_deprecated_data_uses_for_.py
@@ -1,0 +1,27 @@
+"""remove deprecated data uses for fideslang 1.4
+
+Revision ID: 5307999c0dac
+Revises: 76c02f99eec1
+Create Date: 2023-06-11 11:15:53.386526
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = "5307999c0dac"
+down_revision = "76c02f99eec1"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # Remove all "default" Data uses
+
+    # Insert the new default data uses
+    pass
+
+
+def downgrade():
+    pass
--- a/src/fides/api/ctl/migrations/versions/5307999c0dac_remove_deprecated_data_uses_for_.py
+++ b/src/fides/api/ctl/migrations/versions/5307999c0dac_remove_deprecated_data_uses_for_.py
@@ -6,7 +6,10 @@ Create Date: 2023-06-11 11:15:53.386526
 
 """
 from alembic import op
-import sqlalchemy as sa
+
+from loguru import logger
+from sqlalchemy import text
+from sqlalchemy.engine import Connection
 
 
 # revision identifiers, used by Alembic.
@@ -16,12 +19,12 @@ branch_labels = None
 depends_on = None
 
 
-def upgrade():
-    # Remove all "default" Data uses
+def upgrade() -> None:
+    logger.info("Removing default data uses.")
+    bind: Connection = op.get_bind()
+    bind.execute(text("DELETE FROM ctl_data_uses WHERE is_default = TRUE;"))
 
-    # Insert the new default data uses
+
+def downgrade() -> None:
     pass
-
-
-def downgrade():
-    pass
+    # This is not reversible, because it relies on an older version of Fideslang
--- a/src/fides/api/ctl/migrations/versions/5307999c0dac_remove_deprecated_data_uses_for_.py
+++ b/src/fides/api/ctl/migrations/versions/5307999c0dac_remove_deprecated_data_uses_for_.py
@@ -24,6 +24,8 @@ def upgrade() -> None:
     bind: Connection = op.get_bind()
     bind.execute(text("DELETE FROM ctl_data_uses WHERE is_default = TRUE;"))
 
+    # This leaves the data uses empty, but the step to seed them back comes _after_ these migrations
+
 
 def downgrade() -> None:
     pass


src/fides/api/ctl/migrations/versions/5307999c0dac_remove_deprecated_data_uses_for_.py
```diff
@@ -0,0 +1,32 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    logger.info("Removing default data uses.")
+    bind: Connection = op.get_bind()
+    bind.execute(text("DELETE FROM ctl_data_uses WHERE is_default = TRUE;"))
+
+    # This leaves the data uses empty, but the step to seed them back comes _after_ these migrations
```
ThomasLaPiana, line 27: The other option we have here is to add them back _right here_ in case this would case issues with other migrations down the line?

Although we would theoretically be aware of those problems at the time of adding that new migration, and be able to fix it there

src/fides/api/ctl/migrations/versions/5307999c0dac_remove_deprecated_data_uses_for_.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+def downgrade() -> None:
+    pass
+    # This is not reversible, because it relies on an older version of Fideslang
```
ThomasLaPiana, line 32: I can't really think of a way to downgrade here without reinstalling an older version of Fideslang 🤔 

We don't really support downgrades anyway despite generally writing them, so I'm not sure how far down the rabbit hole we want to go here

ThomasLaPiana:
@NevilleS this is simple but it gets the job done!

Notice the two comments I left

src/fides/api/ctl/migrations/versions/5307999c0dac_remove_deprecated_data_uses_for_.py
ThomasLaPiana: After thinking about this again, I don't think this is an issue logically. Considering the default taxonomy is _always_ seeded at the end of the migrations, this means that we should _never_ be writing migrations that rely on the default taxonomy tables having the default taxonomies in them at migration-time.

Given this fact, I don't think seeding them is necessary or even correct in this case

NevilleS: I think that's fair; migrations should avoid ever depending on any imported modules 👍

src/fides/api/ctl/migrations/versions/5307999c0dac_remove_deprecated_data_uses_for_.py
```diff
@@ -0,0 +1,32 @@
...
+revision = "5307999c0dac"
+down_revision = "76c02f99eec1"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    logger.info("Removing default data uses.")
+    bind: Connection = op.get_bind()
+    bind.execute(text("DELETE FROM ctl_data_uses WHERE is_default = TRUE;"))
```
NevilleS, line 25: Are you sure you really want to delete _all_ the data use table and then recreate them afterwards, instead of attempting a selective delete? In this file you have an explicit list of keys that you attempt to rename: https://github.com/ethyca/fides/blob/0d721d1b2e02653406d50ea5ef071024e6789d4f/src/fides/api/ctl/migrations/versions/63b482f5b49b_upgrade_to_fideslang_1_4.py#L24

NevilleS: In other words: you're sure that both:
1. There aren't any foreign-key relationships that might cause a SQL error here on delete
2. All the unaffected data uses will safely be recreated afterwards
3. Nothing else is risky here?

NevilleS (changes requested):
Couple open questions!

src/fides/api/ctl/migrations/versions/5307999c0dac_remove_deprecated_data_uses_for_.py
ThomasLaPiana: guaranteed 🙂 check the `steps to confirm` up above

We have no foreign key relationships at all to `data use`, all validation is done via the API which doesn't get touched until after the migrations/seeding is done. I can write up more test cases to confirm this

Correct that in that other migration we _are_ doing selective updates, but that's because we're touching tables outside of `data use`. Here we're only deleting rows that we know we added ourselves (default taxonomy) and that will get re-populated correctly as soon as the migrations finish.

No, but again our test coverage of version-over-version is not broad, so we should do more manual testing to verify. I can work on more test cases to be a little safer?

NevilleS: OK, sounds good. Let's ship it then, and it feels like the only risk might be on the `CustomField` model which is pretty manageable?

NevilleS (approved):
Thanks!

src/fides/api/ctl/migrations/versions/5307999c0dac_remove_deprecated_data_uses_for_.py
ThomasLaPiana: Yep, we'll do some special test cases there and I'll follow up with the relevant individuals 🙂

Issue #3443, opened 2023-06-02
Render acknowledge button when there are only `notice_only` notices


allisonking:
### Is your feature request related to a specific problem?

Our `fides-js` banner currently always renders 3 buttons—Manage preferences, Reject, and Accept, no matter what kind of notice we have. 

Notice only notices also have to be disabled and permanently set to `true`

### Describe the solution you'd like

If there are only `notice_only` notices, we should render an "acknowledge" button instead, and wire up the user preference to pass "acknowledge"

### Describe alternatives you've considered, if any

_A description of any alternative solutions or features you've considered._

### Additional context

Design: https://www.figma.com/file/dmEwdK3xZwjKfGVQ9t66xe/Fides-v.2-Master-Working-Files?type=design&node-id=6980-351571&t=6PGfme4Uozj1lLtz-0

allisonking:
A few questions:
* Do we still render the modal if there are only notice only notices? There's no "Manage preferences" link in the design
  * If yes: they wouldn't be able to change anything, so do we just render one acknowledgment button?
  * If no: then do we also keep a user's link that they use to open up the modal hidden?
* Same question for the privacy center, what should that page look like when there are only notice only notices?

allisonking:
@mfbrown answered in standup today: yes, we should still show the modal, and then we need the modal to also the conditional acknowledge button

Roger-Ethyca:
moving to done

PR #3546, opened 2023-06-12
Notice only rendering

allisonking:
Closes https://github.com/ethyca/fides/issues/3443

### Code Changes

* [x] Conditionally render acknowledge button in the banner
* [x] Conditionally render acknowledge button in the modal
* [x] Make notice-only notices disabled and set to true in both fides-js and privacy center
* [x] Update tests
* [x] Refactored the button component out of the banner and modal since they have similar logic. A few style tweaks ensued. It also cuts down on our bundle size by about ~.1kb now that we are using more shared components and styles!
  * [x] This diff is pretty big, but is mostly contained to this commit https://github.com/ethyca/fides/pull/3546/commits/6dfec803b5190ce95ba973615b8467b585030e65

### Steps to Confirm

* Edit your `fides-js-demo-components.html` such that both notices are `notice_only` and have a default preference of `opt_in` (which all `notice_only` notices should have)
* `turbo dev` in the `fides-js` folder and visit http://localhost:3000/fides-js-components-demo.html
* You should see the banner render "OK" instead of accept/reject
* Click the modal link to view the notice-only notices which can't be toggled off. You should also see "OK" here instead of the 3 buttons
* Clicking "OK" should send an acknowledge preference
* Visit localhost:3000. You'll need to populate your notices with notice only ones. You should see the radio button becomes disabled

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### Description Of Changes

image

image

When only one notice is notice-only
image

Privacy center if we make a notice notice-only
image

### DIFF
--- a/clients/fides-js/src/components/ConsentBanner.tsx
+++ b/clients/fides-js/src/components/ConsentBanner.tsx
@@ -11,7 +11,7 @@ interface BannerProps {
   onClose: () => void;
   bannerIsOpen: boolean;
   /** Shows acknowledge button instead of Accept all/ Reject all */
-  showAcknowledge: boolean;
+  isAcknowledgeOnly: boolean;
 }
 
 const ConsentBanner: FunctionComponent<BannerProps> = ({
@@ -21,7 +21,7 @@ const ConsentBanner: FunctionComponent<BannerProps> = ({
   onManagePreferences,
   onClose,
   bannerIsOpen,
-  showAcknowledge,
+  isAcknowledgeOnly,
 }) => {
   const {
     title = "Manage your consent",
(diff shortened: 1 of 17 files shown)


clients/fides-js/src/components/ConsentBanner.tsx
```diff
@@ -1,88 +1,39 @@
...
 import CloseButton from "./CloseButton";
 
 interface BannerProps {
   experience: ExperienceConfig;
-  onAcceptAll: () => void;
-  onRejectAll: () => void;
-  onManagePreferences: () => void;
   onClose: () => void;
   bannerIsOpen: boolean;
+  buttonGroup: VNode;
```
allisonking, line 9: this pattern is known as react slots, and is handy for a few things. in this case, it's nice because `ConsentBanner` and `ConsentModal` both have these buttons that need to know how to save and a few other things. our component tree looks like

1. Overlay // knows how to save
2. ConsentModal / ConsentBanner // doesn't care about saving
3. ConsentButtons // needs to call save

so rather than prop drilling through the modal and banner, we can actually instantiate the grandchild in the parent element, and consolidate code that way

eastandwestwind (approved):
looks great, thanks for adding e2e tests too!

PR #3629, opened 2023-06-20
Adding array support for unflatten_dict

### LINKED ISSUE #3493: SaaS connector masked_object_fields aren't unflattened correctly
### Bug Description

For SaaS connectors, there are scenarios when we need to mask nested JSON data
```
{
  "emails": [
    {"original": "abc@example.com"}
  ]
}
```

The result we get at the moment after masking is this
```
{
  "emails": {
    "0": {
      "original": "MASKED_VALUE"
    }
}
```
The array is unflattened incorrectly and is represented as an object with each value corresponding to its index. The expected value is this
```
{
  "emails": [
    {"original": "MASKED_VALUE"}
  ]
}
```
The issue is with the `saas_util.unflatten_dict` function, it needs to handle this nested array use case.

galvana:
Closes #3493 

### Description Of Changes

This ticket updates the `unflatten_dict` function to be able to "unflatten" array values

### Code Changes

* [ ] Updated the `unflatten_dict` function and updated the test cases

### Pre-Merge Checklist

* [ ] Update `CHANGELOG.md`


tests/ops/util/test_saas_util.py
```diff
@@ -329,50 +329,63 @@ def test_regex_is_not_greedy(self):
...
-    with pytest.raises(FidesopsException):
-        unflatten_dict({"A.B.C": 1, "A": {"B.D": 2}})
-
-    # unflatten_dict shouldn't be called with a None separator
-    with pytest.raises(IndexError):
-        unflatten_dict({"": "1"}, separator=None)
+
+    def test_multi_value_array(self):
+        assert unflatten_dict(
+            {"A.0.B": "C", "A.0.D": "E", "A.1.F": "G", "A.1.H": "I"}
```
adamsachs, line 372: i guess if we _really_ wanted to go crazy, you could test _nested lists_, i.e. lists of lists. but that feels too contrived i think 😄

adamsachs (approved):
this looks good, tricky change! but great test coverage, the tests make clear the behavior of the change and confirms it's working on a number of different edge cases.

i still have trouble fully tracking all the flattened dict transformations that come thru with the masking engine codepath, so i just want to confirm that we're not going to hit an edge case here where an integer-like string, e.g. `"0"` or `"1"`, is actually intended to be a dict key? rather than an indicator of the index of the element in the list as we're assuming/treating it now.

tests/ops/util/test_saas_util.py
galvana: 😆 agreed, let's draw the line here for now

PR #3630, opened 2023-06-20
[Backend] EU Location Updates [#3611]

### LINKED ISSUE #3611: Location Improvements: EU Formatting

We have a mixture of formats in our PrivacyNoticeRegions. For example, in the us, it is country_state (us_ca) but in the EU, it is countrygroup_country (eu_fr).  As a first round of location improvements, make all eu-prefixed countries just be their country prefix (fr).  Allow filtering on either region or country.

- Switch all eu prefixed countries to just be the country prefix
- Keep API query params the same: GET /privacy_experience?region=fr_idg, or ?region=fr.  
   - Add new filtering requirements: first prioritize matching "fr_idg" - if this doesn't exist, then search for any matching "fr".
   - In other words, allow the API to fall back to a more generic country match if the region match isn't found.
 - We need a data migration for any existing notices, notice histories, and experiences with these region codes.
 - Update out of the box notices to have new country codes

Further context:  https://ethyca.slack.com/archives/C04S7K0T2V7/p1687204241866249

### Additional context

pattisdr:
Closes #3611 

❗ Contains data migration; bump downrev before merge

### Code Changes

* [x] Update out of the box default notices with new country/region codes for the EU
* [x] Allow GET {{host}}/privacy-experience/?region=fr_idg to drop back to returning experiences with just "fr" if "fr_idg" doesn't exist
* [x] Relax privacy experience filtering so it doesn't have to be a region we support
* [x] Add a data migration to update privacy notices, templates, privacy notice history records, privacy experiences, and privacy preference history records to use the new eu codes in the db.
* [x] Update the PrivacyNoticeRegion Pydantic schema with new country/region codes and also define using the functional syntax so country code "is" (which is a reserved word in python) can be used.
* [x] When saving consent preferences, don't require the "user_geography" to be a privacy notice region we support

### Steps to Confirm

* [ ] Main thing: Search {{host}}/privacy-experience/?region=fr_idg, and it will return experiences for "france" since we don't have "fr_idg" region.

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### Description Of Changes

Update eu regions to drop the eu prefix, so everything is structured "country_region" or just "country".  The eu was originally structured like "countrygroup_country".

On a related note, drop back to filtering experiences by country if their specific detected region isn't found.


src/fides/api/models/privacy_preference.py
```diff
@@ -155,7 +154,7 @@ class PrivacyPreferenceHistory(Base):
         ),
     )
 
-    user_geography = Column(EnumColumn(PrivacyNoticeRegion), index=True)
+    user_geography = Column(String, index=True)
```
pattisdr, line 157: This is already a character varying column at the db level so no db migration needed

src/fides/api/models/privacy_notice.py
```diff
@@ -24,99 +24,96 @@ class UserConsentPreference(Enum):
...
+        ("pt", "pt"),  # portugal
+        ("ro", "ro"),  # romania
+        ("si", "si"),  # slovenia
+        ("sk", "sk"),  # slovakia
+        ("fi", "fi"),  # finland
+        ("se", "se"),  # sweden
+        ("gb_eng", "gb_eng"),  # england
+        ("gb_sct", "gb_sct"),  # scotland
+        ("gb_wls", "gb_wls"),  # wales
+        ("gb_nir", "gb_nir"),  # northern ireland
```
pattisdr, line 111: Should I just use the "gb" country code for now? When I originally put this in, I was trying to adhere to the country, region format -

src/fides/api/models/privacy_preference.py
```diff
@@ -42,7 +41,7 @@ class RequestOrigin(Enum):
 class ConsentMethod(Enum):
     button = "button"
     gpc = "gpc"
-    individual_notice = "api"
+    individual_notice = "individual_notice"
```
pattisdr, line 44: Old copy past error - @allisonking are we even using this field? It's the "method" when saving consent preferences and I think it's subject to change with the new consent reporting work

src/fides/api/models/privacy_notice.py
```diff
@@ -24,99 +24,96 @@ class UserConsentPreference(Enum):
...
+        ("is", "is"),  # iceland
```
pattisdr, line 112: Defining this enum using functional notation instead lets me define "is" here which is a reserved python keyword

src/fides/data/privacy_notices/privacy_notice_templates.yml
```diff
@@ -5,39 +5,39 @@ privacy_notices:
...
+    - fi
+    - se
     - gb_eng
     - gb_sct
     - gb_wls
     - gb_nir
-    - isl
-    - nor
+    - is
+    - "no"
```
pattisdr, line 40: Both "is" and "no" were problematic when I first put this in, so I used "isl" and "nor" instead.  I knew that "is" was failing because it was a python keyword, but "no" was actually failing in the yaml, not python because it was being transformed to a boolean. Quoting this helps.

src/fides/api/models/privacy_preference.py
allisonking: oh interesting, no we aren't using this field in the frontend, though maybe we should be? Is `individual_notice` meant to be when a user toggles different notices then clicks "Save", while "button" is for "accept all"/"reject all"?

allisonking: perfect! although just to be clear, we _are_ using `ConsentMethod` in the frontend and using `button` and `gpc`, but not `individual_notice`

pattisdr: got it okay 👍

pattisdr:
@adamsachs I was hoping you could review the data migration if you have time - @allisonking and @eastandwestwind can you verify this solution works for the frontend?

adamsachs (approved):
lgtm! focused on the data migration - smoke test across two branches looked good, values were updated both in tables with region lists and with single-region columns. migration code itself is clean and DRY - very easy to follow 👍 

took a brief look at other code changes and nothing stood out to me but didn't do a comprehensive review there. if that's needed, please let me know and i can circle back!

src/fides/api/alembic/migrations/versions/d2996381c4dd_eu_locations_migration.py
```diff
@@ -0,0 +1,165 @@
...
+    )
+
+
+def upgrade():
+    migration_eu_regions(region_mapping=region_map)
+
+
+def downgrade():
+    migration_eu_regions(
+        region_mapping={value: key for key, value in region_map.items()}
```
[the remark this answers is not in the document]
pattisdr: I copied @ThomasLaPiana - he did this in the data use 1.4 migration!

pattisdr:
thanks very much for reviewing the data migration @adamsachs ⭐

src/fides/api/alembic/migrations/versions/d2996381c4dd_eu_locations_migration.py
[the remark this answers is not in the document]
adamsachs: ah i knew i'd seen it somewhere, 🙇 @ThomasLaPiana

allisonking (approved):
this works great! I had existing notices but they all migrated to the new regions no problem. I also confirmed admin-ui, fides-js, and privacy center are still working 👍 

I switched my vpn location to france, which sends `fr_pac` and got back notices just fine

image

NevilleS:
@pattisdr: I gave this a quick test drive on my local E2E test setup, and this worked like a charm 👍 

In particular, my previously failing test case where I would switch to the `FR-IDG` location in the Cookie House sample app now passes. Before, that wouldn't match the EU-FR notice, but now that returns the `FR` experience and displays the opt-in banner as expected

NevilleS:
(TL;DR what @allisonking said 😄)

PR #959, opened 2022-08-04
Re-add deprecated API settings for backwards compatibility

ThomasLaPiana:
Closes #957 

### Code Changes

* [x] add a special function to handle deprecated config values
* [x] Update tests to confirm expected behavior of deprecated values
* [x] fix the `fides view config` command

### Steps to Confirm

* [x] removing the `database` section and using the `api` section instead works
* [x] run `fides view config` and confirm that 1) it works and that 2) no `api` section is present

### Pre-Merge Checklist

* [x] All CI Pipelines Succeeded
* Documentation Updated:
  - [ ] documentation complete, or draft/outline provided (tag docs-team to complete/review on this branch)
  - [ ] documentation issue created (tag docs-team to complete issue separately)
* [x] Issue Requirements are Met
* [ ] ~~Relevant Follow-Up Issues Created~~ (We already have one for deprecation procedures)
* [x] Update `CHANGELOG.md`

### Description Of Changes

This PR adds a special handler function that will update new config values with older, deprecated ones. This is a short-term stop-gap solution until we implement a proper procedure for deprecating config values or other code features.

Note specifically that if the `database` section exists, this overwrite will get skipped, however if the `api` section (the deprecated one) exists and the `database` section does _not_ exist than the API values will get mapped to the new database section.

This PR is designed to also be minimally invasive, so this should be a pretty safe change to make to support our users. Tests have been updated accordingly to make sure things are being written as intended.

As a side note, I realized that `fides view config` was broken and that we weren't testing for it, so I fixed it and added a test to check it.

PR #3654, opened 2023-06-22
[Backend] Add Systems Applicable Filter to Privacy Experience List 

pattisdr:
Closes [#959](https://github.com/ethyca/fidesplus/issues/959)

### Description Of Changes

If there are no systems in Fides with data uses that are mapped to privacy notices, then those notices shouldn't show up in the experiences.  

Adds a `?systems_applicable` query param to Privacy Experience List (`GET {{host}}/privacy-experience/?systems_applicable=True`)

We have this filter on GET Privacy Notices but not on experiences themselves.

Note that the experiences themselves still show up if they exist.  If you don't want to see the experience either, use the separate existing flag, `?has_notices=True`

### Code Changes

- If systems_applicable query param is passed to the GET PrivacyExperienceList endpoint, pass that filter onto the embedded privacy notices to only return notices whose data use corresponds to a system.

### Steps to Confirm

* [ ] Verify that you have some privacy notices whose data uses match a system's data use and some that don't
* [ ] Compare results of `GET {{host}}/privacy-experience/` and `GET {{host}}/privacy-experience/?systems_applicable=True`
     - Note: data uses "match" if they are either an exact match, or the privacy notice's data use is a parent of the system's data use.  For example, if the privacy notice has a data use of `marketing` and the system has a data use of `marketing.advertising` - that is considered to match.  The system's data use is under the scope of a privacy notice.  The reverse case is not considered a match.

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`


pattisdr:
@allisonking when you have time could you take a look and add the corresponding frontend to this?  There's just a new `systems_applicable` filter for experiences, not just notices!

clients/privacy-center/features/consent/consent.slice.ts
```diff
@@ -67,6 +67,8 @@ export const consentApi = baseApi.injectEndpoints({
           component: ComponentType.PRIVACY_CENTER,
           has_notices: true,
           show_disabled: false,
+          has_config: true,
```
allisonking, line 70: I think this became optional at some point, but adding it in to stay consistent with fides-js and to be safer

allisonking (approved):

PR #3669, opened 2023-06-23
Allow Privacy Action Cards to Grow

### LINKED ISSUE #3668: Privacy Center - Long Action Titles Lead to Cutoff Description Text
### Bug Description

When the title for one of the actions titles is long enough to span two lines, the description text is clipped (note the tails of the `y` and `p` in the below.

image

This is made even worse when card has focus/hover
image

Additionally, longer descriptions are truncated without any way to access the remaining text.

#### Steps to Reproduce

```
    {
      "icon_path": "/download.svg",
      "title": "Download your personal information",
      "description": "We will provide you a with report containing of all your personal data stored in our system.",
      "description_subtext": [],
      "identity_inputs": {
        "email": "required"
      },
      "confirmButtonText": "Continue",
      "cancelButtonText": "Cancel"
    },
```

### Expected behavior

_A description of what you expected to happen._

### Screenshots

### Environment

* Version:
* OS:
* Python Version:
* Docker Version:

### Additional context

RobertKeyser:
Closes #3668 

### Description Of Changes

This PR changes the privacy center, allowing the privacy action cards to grow if there is a long description (now unlimited lines) or a title that spans two lines. Before, they were static and would lead to clipping and truncation.

### Code Changes

* Replace the `h` property on the flex element with `minH`
* Remove max noOfLines on chakra `text` element

### Steps to Confirm

Default text, everything should look the same
image

Long descriptions should now truncate at 5 lines instead of 3, grows flexbox
image

Title that spans two lines now grows flexbox
image

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`


RobertKeyser:
Hypothetically speaking, if we wanted to eliminate the limit altogether, we could remove the `noOfLines` prop on the `Text` element, which would enable the following monstrosity: 

image

There are a few approaches here:
1. We let the user do whatever they want and have the box grow to an infinite length. 
2. We impose a limit for the sake of prescriptive usability.

clients/privacy-center/components/Card.tsx
```diff
@@ -57,13 +57,7 @@ const Card: React.FC<CardProps> = ({
...
       {title}
     </Text>
-    <Text
-      color="gray.600"
-      fontSize="xs"
-      fontWeight="normal"
-      lineHeight="16px"
-      noOfLines={3}
-    >
+    <Text color="gray.600" fontSize="xs" fontWeight="normal" lineHeight="16px">
```
seanpreston, line 60: Nit: Could you leave these on separate lines as it's much easier to see the diffs at that point

RobertKeyser: I think Prettier forced it to go to one line.

allisonking, line 60: the PR description says the number of lines was set to 5, but I'm not seeing that here? seems like this one would be the unlimited size one?

allisonking: if it were up to me, I'd probably go with the unlimited size. and then if a user doesn't like the way it looks, it can be a forcing function to make their description more concise :) 

another option would be to not have the others grow, but then you'll likely have cards of different heights even when there's only a few lines of difference, which wouldn't be very aesthetic

allisonking (approved):

PR #3673, opened 2023-06-26
GPC indication on fides-js overlay

allisonking:
Closes https://github.com/ethyca/fidesplus/issues/958

### Description Of Changes

Adds GPC indicators to both the modal and the banner

| Description | Screenshot |
|-------------|------------|
|Banner|image|
|Toggle when GPC is applied|image|
|If you opt into the opted-out toggle|image|

### Code Changes

* [x] Add `GpcBanner` and `GpcBadge` as components
* [x] Move `GpcStatus` type and `getGpcStatusFromNotice` from privacy-center to `fides-js` so that `fides-js` can use it too
* [x] Render indicators on modal
* [x] Fix bug in `automaticallyApplyGPCPreferences` where the cookie consent would be missing keys of notices that were notice only, or already had a preference
* [x] Render indicator on banner
* [x] Cypress tests

### Steps to Confirm

* `turbo dev` in privacy center
* Visit http://localhost:3000/fides-js-components-demo.html?globalPrivacyControl=true (note the flag that sets the gpc signal!)
* You should see the banner with the indicator
* Open the modal and you should see gpc applied in the first notice, but not the notice-only notice
* Opt into the first notice
* It should now say gpc overridden 
* Edit `fides-js-components-demo.html` so that the first notice `has_gpc_flag=false`
* There should be no gpc indicator in the modal, but there will still be info messages saying gpc was detected

### Pre-Merge Checklist

* [x] Update `CHANGELOG.md`


clients/fides-js/src/fides.ts
```diff
@@ -131,26 +133,40 @@ const automaticallyApplyGPCPreferences = (
...
     return;
   }
 
-  if (!getConsentContext().globalPrivacyControl) {
+  const context = getConsentContext();
+  if (!context.globalPrivacyControl) {
     return;
   }
 
-  const consentPreferencesToSave: Array<SaveConsentPreference> = [];
```
allisonking, line 142: this change fixes a pretty subtle bug, where:
1. let's say there are two notices, one notice-only `first` and one opt-out `second`
2. when fides-js first initializes, it figures out to make the cookie `{first: false, second: true}`
3. Then in this gpc function, because this originally only pushed to `consentPreferencesToSave` if certain conditions were met (`has_gpc_flag` and no current preference), `consentPreferencesToSave` often would not have all the keys. In this case, only `second` would be pushed, since `first` is `notice_only` so would have `has_gpc_flag=false`
4. It would send `consentPreferencesToSave` to `updateConsentPreferences` which sends the update to the backend. However, it also updates the cookie based on this value, so the cookie goes from `{first: false, second: true}` --> `{second: false}`, when really we want `{first: false, second: false}`. This resulted in the notice-only notice looking like it was always opted out!

Instead, we make sure we always map over all privacy notices here, and only set to `false` in certain cases. the important part is that `updateConsentPreferences` gets _all_ the relevant preferences

clients/fides-js/src/lib/consent-value.ts
```diff
@@ -28,6 +28,9 @@ export const resolveConsentValue = (
   if (notice.current_preference) {
     return transformUserPreferenceToBoolean(notice.current_preference);
   }
+  if (notice.consent_mechanism === ConsentMechanism.NOTICE_ONLY) {
```
allisonking, line 31: this is only helpful if the notice is in an invalid state of `notice-only` but `has_gpc_flag=true`. but I added it to be safe!

eastandwestwind (approved):
looks great @allisonking ! Especially, good catch on that bug 🐛 , just 1 request non-blocking, so I'll go ahead and approve.

clients/fides-js/src/lib/consent-value.ts
allisonking: actually, looking at this again, my earlier comment isn't _quite_ accurate—the logic is

1. return user's current preference if they have one
2. for notice-only, return true
3. if gpc is enabled and applicable, return false
4. return the notice's default preference 

for notice only, normally the 4th step would kick in and we'd be okay. but if it's an invalid state of notice-only + `has_gpc_flag=true`, then step 3 would kick in and opt us out of the notice-only notice. so adding step 2 helps prevent that and also returns a lil earlier for notice-only which is nice. I think it's a case where the series of if statements explains it better than a comment (the comment above applies more to why I added it, but I think the end state sense by itself 👍 )

PR #3674, opened 2023-06-26
Include `data_use` and `data_category` metadata in `upload` of access results

### LINKED ISSUE #3266: Update `run_access_request` to include data uses and data category metadata
# Description

The upcoming DSR export package is going to use each connectors data uses and data categories to help organize the data subjects data with the package. Currently, access requests don't add this data to the access package. Access Requests need to be updated to include in data in the package.

[`access_results_post_processing`](https://github.com/ethyca/fides/blob/c063274d1a857c408a64d5bbff52ae2ae3ce9086/src/fides/api/ops/task/graph_task.py#L471) is the function that will updated to include the additional metadata.

## Things to consider
The access package needs to use the metadata that was in the datasets at the time of graph generation. They cannot be looked up at runtime. This is because the execution will likely not match the current datasets exactly. This is because access requests are long running jobs. DSR's already cache the graph to fix this drift issue. 

**Make sure that all of the required package metadata is included at graph creation time**. If it's not already included in the cached graph then add it.

adamsachs:
Closes #3266 

### Description Of Changes

We want to include `data_use` and `data_category` metadata as part of the output package for access requests. In order to do this, we need to make that metadata available to the `upload()` functions that produce and serialize the output packages. This PR updates the request execution internals to pass those pieces of metadata to the appropriate `upload()` function. 

The `data_category` metadata was already available as the `data_category_field_mapping` attribute on the `DatasetGraph` that's used during privacy request execution -- the update here is simply to pass that along as another argument to the `upload()` function.

The `data_use` metadata is not something we'd been keeping track of at all in privacy request execution. A new data structure has been created to maintain that metadata - a `dict` that maps each `Collection` in the graph traversal to a set of `data_use`s associated with the `Collection`, where the association is determined by the `Collection` --> `ConnectionConfig` --> `System` --> `DataUse` relationship chain. That `dict` is populated and stored in the redis cache with a key unique to the particular privacy request as request execution beings; the cache entry is then retrieved right before `upload()` is called at the end of request execution.

### Code Changes

* [x] pass along the `DatasetGraph.data_category_field_mapping` as an argument to `upload()`
* [x] populate a "data use map" and store it in the redis cache, before request execution, with a key that's unique to the privacy request
* [x] retrieve the "data use map" that's stored in the redis cache and pass it along as an argument to `upload()`

### Steps to Confirm

* [ ] hard to confirm this functionality manually until we actually make use of the new metadata in the output package
* [x] in the meantime, we should just confirm manually that request execution does not regress with these additions!

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### DIFF
--- a/src/fides/api/models/privacy_request.py
+++ b/src/fides/api/models/privacy_request.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 import json
 from datetime import datetime, timedelta
 from enum import Enum as EnumType
-from typing import Any, Dict, List, Optional, Union
+from typing import Any, Dict, List, Optional, Set, Union
 
 from celery.result import AsyncResult
 from loguru import logger
@@ -658,6 +658,24 @@ class PrivacyRequest(IdentityVerificationMixin, Base):  # pylint: disable=R0904
         value_dict: Optional[
             Dict[str, Optional[GraphRepr]]
         ] = cache.get_encoded_objects_by_prefix(f"ACCESS_GRAPH__{self.id}")
+        return list(value_dict.values())[0] if value_dict else None
+
+    def cache_data_use_map(self, value: Dict[str, Set[str]]) -> None:
+        """
+        Cache a dict of collections traversed in the privacy request
+        mapped to their associated data uses
+        """
+        cache: FidesopsRedis = get_cache()
+        cache.set_encoded_object(f"DATA_USE_MAP__{self.id}", value)
+
+    def get_cached_data_use_map(self) -> Optional[Dict[str, Set[str]]]:
+        """
+        Fetch the collection -> data use map cached for this privacy request
+        """
+        cache: FidesopsRedis = get_cache()
+        value_dict: Optional[
+            Dict[str, Optional[Dict[str, Set[str]]]]
+        ] = cache.get_encoded_objects_by_prefix(f"DATA_USE_MAP__{self.id}")
         return list(value_dict.values())[0] if value_dict else None
 
     def trigger_policy_webhook(
--- a/src/fides/api/service/privacy_request/request_runner_service.py
+++ b/src/fides/api/service/privacy_request/request_runner_service.py
@@ -231,6 +231,7 @@ def upload_access_results(  # pylint: disable=R0912
                 privacy_request=privacy_request,
                 data=filtered_results,
                 storage_key=storage_destination.key,  # type: ignore
+                data_category_field_mapping=dataset_graph.data_category_field_mapping,
             )
             if download_url:
                 download_urls.append(download_url)
--- a/src/fides/api/service/privacy_request/request_runner_service.py
+++ b/src/fides/api/service/privacy_request/request_runner_service.py
@@ -232,6 +232,7 @@ def upload_access_results(  # pylint: disable=R0912
                 data=filtered_results,
                 storage_key=storage_destination.key,  # type: ignore
                 data_category_field_mapping=dataset_graph.data_category_field_mapping,
+                data_use_map=privacy_request.get_cached_data_use_map(),
             )
             if download_url:
                 download_urls.append(download_url)
--- a/src/fides/api/service/storage/storage_uploader_service.py
+++ b/src/fides/api/service/storage/storage_uploader_service.py
@@ -5,6 +5,7 @@ from loguru import logger
 from sqlalchemy.orm import Session
 
 from fides.api.common_exceptions import StorageUploadError
+from fides.api.graph.graph import DataCategoryFieldMapping
 from fides.api.models.privacy_request import PrivacyRequest
 from fides.api.models.storage import StorageConfig
 from fides.api.schemas.storage.storage import (
@@ -17,7 +18,11 @@ from fides.api.tasks.storage import upload_to_local, upload_to_s3
 
 
 def upload(
-    db: Session, *, privacy_request: PrivacyRequest, data: Dict, storage_key: FidesKey
+    db: Session,
+    privacy_request: PrivacyRequest,
+    data: Dict,
+    storage_key: FidesKey,
+    data_category_field_mapping: Optional[DataCategoryFieldMapping] = None,
 ) -> str:
     """
     Retrieves storage configs and calls appropriate upload method
@@ -35,7 +40,7 @@ def upload(
         logger.warning("Storage type not found: {}", storage_key)
         raise StorageUploadError(f"Storage type not found: {storage_key}")
     uploader: Any = _get_uploader_from_config_type(config.type)  # type: ignore
-    return uploader(db, config, data, privacy_request)
+    return uploader(db, config, data, privacy_request, data_category_field_mapping)
 
 
 def get_extension(resp_format: ResponseFormat) -> str:
@@ -76,7 +81,11 @@ def _get_uploader_from_config_type(storage_type: StorageType) -> Any:
 
 
 def _s3_uploader(
-    _: Session, config: StorageConfig, data: Dict, privacy_request: PrivacyRequest
+    _: Session,
+    config: StorageConfig,
+    data: Dict,
+    privacy_request: PrivacyRequest,
+    data_category_field_mapping: Optional[DataCategoryFieldMapping],
 ) -> str:
     """Constructs necessary info needed for s3 before calling upload"""
     file_key: str = _construct_file_key(privacy_request.id, config)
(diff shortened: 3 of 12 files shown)


adamsachs:
@galvana let me know if this looks good, based on what we discussed.

@pattisdr figured i'd tag you in as well, if you have a moment, to make sure you don't notice any red flags with the enhancement to privacy request execution and a new cache entry!

pattisdr:
starting review!

src/fides/api/service/privacy_request/request_runner_service.py
```diff
@@ -231,6 +231,8 @@ def upload_access_results(  # pylint: disable=R0912
                 privacy_request=privacy_request,
                 data=filtered_results,
                 storage_key=storage_destination.key,  # type: ignore
+                data_category_field_mapping=dataset_graph.data_category_field_mapping,
+                data_use_map=privacy_request.get_cached_data_use_map(),
```
pattisdr, line 235: So the best I can tell, you're caching a data use map on the privacy request instead of formatting this more similar to dataset_graph.data_category_field_mapping because you need the GraphTask -> Connector -> ConnectionConfig -> System.

If that's the case, can you explain this in the docstring of `_format_data_use_map_for_caching`? Looks like it's not something we can calculate after the fact easily.

src/fides/api/task/graph_task.py
```diff
@@ -716,6 +743,7 @@ def termination_fn(
             )
         )
         privacy_request.cache_access_graph(format_graph_for_caching(env, end_nodes))
```
pattisdr, line 745: This is my omission, but can you add a note that `privacy_request.cache_access_graph` is for analytics purposes only? I believe we added this to fideslog last summer, after the fact, to track how graphs changed between retry to figure out how to prioritize some of the edge cases.  We don't use this in any way for the access package.

src/fides/api/service/storage/storage_uploader_service.py
```diff
@@ -85,13 +98,18 @@ def _s3_uploader(
     auth_method = config.details[StorageDetails.AUTH_METHOD.value]
 
     return upload_to_s3(
-        config.secrets, data, bucket_name, file_key, config.format.value, privacy_request, auth_method  # type: ignore
+        config.secrets, data, bucket_name, file_key, config.format.value, privacy_request, auth_method, data_category_field_mapping  # type: ignore
```
pattisdr, line 101: should we pass on the data_use_map too to `upload_to_s3`?

src/fides/api/service/storage/storage_uploader_service.py
```diff
@@ -85,13 +98,18 @@ def _s3_uploader(
...
+    config: StorageConfig,
+    data: Dict,
+    privacy_request: PrivacyRequest,
+    data_category_field_mapping: Optional[DataCategoryFieldMapping] = None,
+    data_use_map: Optional[Dict[str, Set[str]]] = None,
 ) -> str:
     """Uploads data to local storage, used for quick-start/demo purposes"""
     file_key: str = _construct_file_key(privacy_request.id, config)
-    return upload_to_local(data, file_key, privacy_request, config.format.value)  # type: ignore
+    return upload_to_local(data, file_key, privacy_request, config.format.value, data_category_field_mapping)  # type: ignore
```
pattisdr, line 115: should we pass on the data_use_map too to `upload_to_local`?

src/fides/api/task/graph_task.py
```diff
@@ -658,6 +666,25 @@ def update_mapping_from_cache(
...
+    Create a map of `Collection`s mapped to their associated `DataUse`s
+    to be stored in the cache.
+
+    `DataUse`s are associated with a `Collection` by means of the `System`
+    that's linked to a `Collection`'s `Connection` definition.
+
+    Example:
+    {
+       <collection1>: ["data_use_1", "data_use_2"],
+       <collection2>: ["data_use_1"],
```
pattisdr, line 682: Minor, example looks like the value is a list instead of a set -

tests/ops/service/privacy_request/test_request_runner_service.py
```diff
@@ -311,6 +312,121 @@ def get_privacy_request_results(
...
+
+    # sanity check that acccess results returned as expected
+    results = pr.get_results()
+    assert len(results.keys()) == 11
+
+    # what we're really testing - ensure data_category_field_mapping arg is well-populated
+    args, kwargs = upload_mock.call_args
+    data_category_field_mapping = kwargs["data_category_field_mapping"]
+
+    # make sure the catergory field mapping generally looks as we expect
```
pattisdr, line 352: minor category typo

tests/ops/task/test_graph_task.py
```diff
@@ -669,6 +679,211 @@ def test_update_erasure_mapping_from_cache_with_data(self, dsk, task_resource):
...
+    )
+    def test_data_use_map(
+        self,
+        connection_config_fixtures,
+        expected_data_use_map,
+        db,
+        privacy_request,
+        policy,
+        request,
+    ):
```
pattisdr, line 849: 🏆 wonderful test

tests/ops/service/privacy_request/test_request_runner_service.py
```diff
@@ -311,6 +312,121 @@ def get_privacy_request_results(
...
+        "postgres_example_test_dataset:customer": "{'marketing.advertising'}",
+        "postgres_example_test_dataset:service_request": "{'marketing.advertising'}",
+        "postgres_example_test_dataset:visit": "{'marketing.advertising'}",
+        "postgres_example_test_dataset:address": "{'marketing.advertising'}",
+        "postgres_example_test_dataset:login": "{'marketing.advertising'}",
+        "postgres_example_test_dataset:orders": "{'marketing.advertising'}",
+        "postgres_example_test_dataset:payment_card": "{'marketing.advertising'}",
+        "postgres_example_test_dataset:order_item": "{'marketing.advertising'}",
+        "postgres_example_test_dataset:product": "{'marketing.advertising'}",
+    }
```
pattisdr, line 427: ah right, highlighting that every node in the dataset will have the same data use.

src/fides/api/models/privacy_request.py
```diff
@@ -660,6 +660,24 @@ def get_cached_access_graph(self) -> Optional[GraphRepr]:
         ] = cache.get_encoded_objects_by_prefix(f"ACCESS_GRAPH__{self.id}")
         return list(value_dict.values())[0] if value_dict else None
 
+    def cache_data_use_map(self, value: Dict[str, Set[str]]) -> None:
+        """
+        Cache a dict of collections traversed in the privacy request
+        mapped to their associated data uses
+        """
+        cache: FidesopsRedis = get_cache()
+        cache.set_encoded_object(f"DATA_USE_MAP__{self.id}", value)
```
pattisdr, line 669: We also have a .fides/redis_dataset.yml that stores the fields in our redis cache that we can add to here - there's nothing that automatically checks this, it was just something I was asked to make when we were merging ctl and ops.

src/fides/api/task/graph_task.py
```diff
@@ -716,6 +743,7 @@ def termination_fn(
             )
         )
         privacy_request.cache_access_graph(format_graph_for_caching(env, end_nodes))
+        privacy_request.cache_data_use_map(_format_data_use_map_for_caching(env))
```
pattisdr, line 746: When I initially saw this, I felt like this was happening in the wrong place, since it was mimicking the format of something we send to fideslog, that we don't actually use for the access package.

But thinking more, it makes sense to put this here, I think code comments can help clarify.

src/fides/api/task/graph_task.py
```diff
@@ -174,6 +175,13 @@ def __init__(
         self.connector: BaseConnector = resources.get_connector(
             self.traversal_node.node.dataset.connection_key  # ConnectionConfig.key
         )
+        self.data_uses: Set[str] = (
+            System.get_data_uses(
+                [self.connector.configuration.system], include_parents=False
+            )
```
pattisdr, line 181: Hey, getting to use this again, great!

pattisdr (commented):
good work, all comments are minor. I like the incremental nature of this PR, first collecting this, before we actually start using the new metadata in the output package.

src/fides/api/service/storage/storage_uploader_service.py
adamsachs: yup, good catch! missed that.

src/fides/api/models/privacy_request.py
adamsachs: got it, thank you for calling that out! i wonder if/how we could document that in the codebase 🤔

src/fides/api/task/graph_task.py
adamsachs: sure! and with your below comment, i get why my addition of the new cache operation here can confuse things. i can def add some comments, i'll also give a bit of thought to see if there's any other minor tweaks to make this a bit clearer.

src/fides/api/service/privacy_request/request_runner_service.py
adamsachs: yeah, well this is more about the graph being subject to change during request execution -  so we want to cache it early on. the issue description mentions this a bit. i'm going to add in some context to the docstring and a code comment here to hopefully make that clearer. let me know if it looks better after that, happy to iterate on this to make sure it's abundantly clear going forward - i think it's probably the least self-explanatory piece of this update 👍

pattisdr: ah okay got it

pattisdr (approved):

galvana (approved):
What a quick turnaround! This looks good and is consistent with what we discussed. The only part we might have to change in the future is the structure of the data use and data category maps but we won't know the ideal structure until we finalize the design for DSR package (collection address as key instead of the data category for the data category map).

adamsachs:
> What a quick turnaround! This looks good and is consistent with what we discussed. The only part we might have to change in the future is the structure of the data use and data category maps but we won't know the ideal structure until we finalize the design for DSR package (collection address as key instead of the data category for the data category map).

thanks @galvana, that makes sense! should be relatively easy to switch that around if/when needed.

Issue #3682, opened 2023-06-26
system `PUT` API fails if `privacy_declaration.name` property is `null`


adamsachs:
### Bug Description

The BE throws a `422` if a `privacy_declaration.name` property on the `PUT /system` request body is set to `null`.

#### Steps to Reproduce

1. _Do this..._
1. _Then This..._

### Expected behavior

The API is inconsistent. We return the property as `null` on a `GET` (depending on how the `system` was initially created), so we either need to handle `null` in the request or make sure we never return it on a `GET` (or both).

### Screenshots

_If applicable, add screenshots to help explain your problem._

### Environment

* Version:
* OS:
* Python Version:
* Docker Version:

### Additional context

_Add any other context about the problem here._

adamsachs:
https://github.com/ethyca/fides/commit/acf4513b3a042c00a19cd3db00d423b1b2712f7f has a test case that covers/exposes the bug, still need to debug it further to figure out where this is getting thrown from on the BE and how to address -- but this should be a decent starting point.

rsilvery:
@adamsachs , did we resolve this?

adamsachs:
> @adamsachs , did we resolve this?

not that i know of. for context, it's an edge case that's not exposed when using the admin UI to manage privacy declarations.  it's only is exposed if you're hitting our API manually in a certain way OR potentially in some fides CLI use cases too, i think.

PR #3683, opened 2023-06-26
Transform null names to an empty string before submission

allisonking:
FE fix to get around https://github.com/ethyca/fides/issues/3682

### Description Of Changes

https://github.com/ethyca/fides/pull/3571 refactored the privacy declaration form a bit so that we could include `cookies` fields. As part of that, there was a change where we only transformed the declaration was being edited, whereas previously we had transformed _all_ declarations prior to submission.

This PR restores doing an additional transform to all declarations right before sending to the API.

### Code Changes

* [x] Add one more transform function

### Steps to Confirm

* Run `nox -s fides_env -- prod`
* On one of the existing systems that already has a data use, try to add another data use
* You should be able to add one without getting any errors

### Pre-Merge Checklist

~* [ ] Issue Requirements are Met~ --> no they are not! but this is a smaller FE workaround 
* [x] Update `CHANGELOG.md`


adamsachs (approved):
manually tested locally and looks good, nice work on the super speedy patch!!!

PR #3688, opened 2023-06-27
Handle Keys and Collections with a double underscore

### LINKED ISSUE #3687: Filter errors for Collections with a double underscore (`__`) in the name
### Bug Description

When processing a DSR, if a Collection has a double underscore in the name (`__`) it appears to cause an error when filtering on data categories.

#### Steps to Reproduce

1. Create a Collection with a double underscore in the name (or multiple)
1. Seed the data with data that would be processed via a DSR (i.e. email)
2. Execute a DSR
3. In the logs, see just a partial name for the Collection with an error
4. Verify all connections seem to pass

### Expected behavior

All of the connections pass, and the Collection name is preserved

### Screenshots

n/a

### Environment

* Version: 2.14.3
* OS: Darwin
* Python Version: 3.10
* Docker Version:

### Additional context

n/a

SteveDMurphy:
Closes #3687

### Description Of Changes

Found by a user with naming conventions that use a double underscore, which we weren't handling previously

### Code Changes

* [x] Adds a test for the edge case
* [x] Create and use a function that extracts the address information and includes the edge case 

### Steps to Confirm

* [ ] While you can set up a table named something like `my__table__with__underscores` add some data, and run a DSR against it, running the failing test against `main` will also show the result where only `underscores` would be returned

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`


src/fides/api/task/graph_task.py
```diff
@@ -733,6 +733,7 @@ def get_cached_data_for_erasures(
     value_dict = cache.get_encoded_objects_by_prefix(
         f"PLACEHOLDER_RESULTS__{privacy_request_id}"
     )
+    # TODO: Determine if this requires handling as well
```
SteveDMurphy, line 736: @seanpreston this is the main question I still had but wanted to get checks running on the PR. If needed, should be very easy to implement 👍🏽

src/fides/api/task/task_resources.py
```diff
@@ -147,7 +147,22 @@ def get_all_cached_objects(self) -> Dict[str, Optional[List[Row]]]:
...
+        values.
+
+        Handles an edge case where double underscores exist in either the fides_key
+        of the Dataset or the Collection name.
+        """
+        request_id_dataset, collection = full_request_id.split(":")
+        number_of_expected_items_to_remove = 2
+        request_list = request_id_dataset.split("__")
+        address_list = request_list[number_of_expected_items_to_remove:]
+        dataset = "__".join(address_list)
```
adamsachs, line 164: this looks like it should get the job done, but i _think_ we can use the `maxsplit` arg of `split()` to make this a bit more concise/straightforward. would something like the below work? it seems to pass the updated test.

```suggestion
        dataset = request_id_dataset.split("__", 2)[-1]
```

(may need some a comment or two to clarify why we're splitting this way)

src/fides/api/task/graph_task.py
adamsachs: hmm, it _seems_ like it does, because i think we're also referencing the collection address when storing the placeholder results, so presumably we need to handle the same edge case? https://github.com/ethyca/fides/pull/3688/files#diff-f1bb24bdf53cb62b5d5ab785e4cda217afa21e57780fbc2dbd04ff674ee4fafbR134

adamsachs (commented):
nice fix - generally looks good to me, just had a small suggestion for what i think is a slightly simpler implementation! i'm also not 100% sure on your open question, so it may be good to get @pattisdr or @seanpreston's input on that one.

src/fides/api/task/graph_task.py
pattisdr: Yes, the same edge case needs to be handled here! What would be an issue for access requests would be an issue for erasure requests.

src/fides/api/task/task_resources.py
SteveDMurphy: Ya I think your idea here makes sense 👍🏽 

My thoughts here were around trying to avoid a magic number scenario like we hit with all of the custom fields changes by being overly verbose with what each step was doing and why (thinking it easier to debug in the future if/when we hit another edge case)

The only other change I might make here is keeping `number_of_expected_items_to_remove` to self document why the number 2 is there if that works?

src/fides/api/task/graph_task.py
SteveDMurphy: Thank you both! Will get that sorted in here for both then 👍🏽

src/fides/api/task/task_resources.py
SteveDMurphy: Fixed in https://github.com/ethyca/fides/pull/3688/commits/7371272ecaec02b7a52df15286a83a668db62d60

src/fides/api/task/graph_task.py
SteveDMurphy: Added in https://github.com/ethyca/fides/pull/3688/commits/fe4d1a7bd32607938ac1871619db3f50e58d7627

SteveDMurphy: Updated in https://github.com/ethyca/fides/pull/3688/commits/b57bf1e08fbfca7693d5eb2ed091609e50464a39 due to differences in the structure of the string. Parameterized the number of strings we strip from the front, hope that feels ok

src/fides/api/task/task_resources.py
adamsachs: very nice, i like where you landed!

adamsachs (approved):
i like where you landed on this @SteveDMurphy! the `number_of_leading_strings_to_exclude` arg makes sense, and the docstring of that function helps to make clear the functionality that otherwise may be a bit confounding for someone coming back to this code. nice work! 👍

PR #3692, opened 2023-06-27
avoid circular deps caused within `api.api.v1`

### LINKED ISSUE #3691: `postgres_setup.py` fails due to circular import error
### Bug Description

when running `nox -s quickstart` or when passing `postgres` as a flag to `nox -s dev`, we execute the `postgres_setup.py` script. that script is hitting a circular dependency issue that prevents it from running successfully:

```
  File "/fides/tests/ops/integration_tests/setup_scripts/postgres_setup.py", line 86, in <module>
    setup()
  File "/fides/tests/ops/integration_tests/setup_scripts/postgres_setup.py", line 51, in setup
    from fides.api.service.connectors.sql_connector import PostgreSQLConnector
  File "/fides/src/fides/api/service/connectors/__init__.py", line 12, in <module>
    from fides.api.service.connectors.consent_email_connector import (
  File "/fides/src/fides/api/service/connectors/consent_email_connector.py", line 44, in <module>
    from fides.api.util.consent_util import (
  File "/fides/src/fides/api/util/consent_util.py", line 10, in <module>
    from fides.api.api.v1.endpoints.utils import transform_fields
  File "/fides/src/fides/api/api/v1/__init__.py", line 8, in <module>
    from .endpoints.admin import ADMIN_ROUTER
  File "/fides/src/fides/api/api/v1/endpoints/admin.py", line 7, in <module>
    from fides.api.db.database import configure_db, reset_db
  File "/fides/src/fides/api/db/database.py", line 17, in <module>
    from fides.api.db.seed import load_default_resources, load_samples
  File "/fides/src/fides/api/db/seed.py", line 12, in <module>
    from fides.api.api.v1.endpoints.saas_config_endpoints import (
  File "/fides/src/fides/api/api/v1/endpoints/saas_config_endpoints.py", line 58, in <module>
    from fides.api.util.connection_util import validate_secrets
  File "/fides/src/fides/api/util/connection_util.py", line 40, in <module>
    from fides.api.service.privacy_request.request_runner_service import (
  File "/fides/src/fides/api/service/privacy_request/request_runner_service.py", line 25, in <module>
    from fides.api.graph.analytics_events import (
  File "/fides/src/fides/api/graph/analytics_events.py", line 16, in <module>
    from fides.api.task.task_resources import TaskResources
  File "/fides/src/fides/api/task/task_resources.py", line 17, in <module>
    from fides.api.service.connectors import (
ImportError: cannot import name 'BigQueryConnector' from partially initialized module 'fides.api.service.connectors' (most likely due to a circular import) (/fides/src/fides/api/service/connectors/__init__.py)
```

#### Steps to Reproduce

1. `nox -s quickstart` (i think? still haven't repro'd myself yet! cc @galvana @pattisdr @Kelsey-Ethyca )

### Expected behavior

`quickstart` session and `postgres` flag should work

### Additional context

noticed by @Kelsey-Ethyca during `2.15.0` release process

adamsachs:
Closes #3691 

‼️ 
**note that we will need to update `fidesplus` accordingly if we move `urn_registry.py` as is proposed here currently!**
‼️ 

### Description Of Changes

The root cause here is that the `__init__.py` of our `api.api.v1` module brings in a lot of dependencies via `ADMIN_ROUTER`. this means we need to be judicious when to depend on code _within_ `api.api.v1` so as to keep it as self-contained as possible. in this case, the immediate cause of the problem is that `postgres_setup.py` depends on our `connectors` module, and our `connectors` module was bringing in `api.api.v1` via some utilities (namely `consent_util` which brought in `endpoint.utils`.) 

it seems like an unambiguously good update to move our `api.api.v1.endpoint.utils` up a few levels into `api.util.endpoint_utils`, alongside other API related utilities that are not only used by code within `api.api.v1`, per the general principle above. i've made that change in https://github.com/ethyca/fides/pull/3692/commits/5bee96312ab1331df7337db1d9c42f9095454d02, and that does help us resolve the initial error here.

but we quickly hit another error (stack trace below), this time because our `fides_connector` (specifically the `fides_client` it is highly dependent on), which is also initialized by the `connectors` module, relies directly on the `urn_registry` that's within `api.api.v1`. the solution to this feels less straightforward to me -- i see 3 potential paths forward but none seem like an obvious best choice:
1. we could move `urn_registry` out of `api.api.v1`, ~but that doesn't feel right, because it's pretty strictly related/coupled to the code in that module~ -- @ThomasLaPiana also suggested this, and given some more thought, this feels like it might be the right path forward. my proposed change here is to move it to a `common.api.v1` module/directory, so that it's clear it refers to our `api.v1` endpoints/code.
2. we could update the `__init__.py` of `api.api.v1` to not bring in so many dependencies via the `ADMIN_ROUTER`, but this feels like a tall order and larger scope than what we're doing here. i'm not even sure it's desirable, it may lead to other problems.
3. we could move `fides_connector` (and `fides_client`) out of the `connectors` module, or at least not have it initialized as part of the module's `__init__.py`, so that it's not initialized with `postgres_setup.py`. this feels simpler and better, it's not crazy to have the `fides_connector` special-cased, since it's a pretty _special_ connector in how self-referential it is. but where's the better spot for it so that we're sure it gets properly initialized on normal app runtime/startup, but it's not causing dep issues?
4. we could update `fides_client` to not rely on the `urn_registry` and instead just use string literals for its URL/URN references. this would be simple, but it feels like it makes that code materially _worse_, and it just doesn't feel like the _right_ solution here - constants are good :)

new error with the update contained in https://github.com/ethyca/fides/pull/3692/commits/5bee96312ab1331df7337db1d9c42f9095454d02: 
```
  File "/fides/tests/ops/integration_tests/setup_scripts/postgres_setup.py", line 15, in <module>
    from fides.api.service.connectors.sql_connector import PostgreSQLConnector
  File "/fides/src/fides/api/service/connectors/__init__.py", line 23, in <module>
    from fides.api.service.connectors.fides_connector import (
  File "/fides/src/fides/api/service/connectors/fides_connector.py", line 19, in <module>
    from fides.api.service.connectors.fides.fides_client import FidesClient
  File "/fides/src/fides/api/service/connectors/fides/fides_client.py", line 9, in <module>
    from fides.api.api.v1 import urn_registry as urls
  File "/fides/src/fides/api/api/v1/__init__.py", line 8, in <module>
    from .endpoints.admin import ADMIN_ROUTER
  File "/fides/src/fides/api/api/v1/endpoints/admin.py", line 7, in <module>
    from fides.api.db.database import configure_db, reset_db
  File "/fides/src/fides/api/db/database.py", line 17, in <module>
    from fides.api.db.seed import load_default_resources, load_samples
  File "/fides/src/fides/api/db/seed.py", line 12, in <module>
    from fides.api.api.v1.endpoints.saas_config_endpoints import (
  File "/fides/src/fides/api/api/v1/endpoints/saas_config_endpoints.py", line 58, in <module>
    from fides.api.util.connection_util import validate_secrets
  File "/fides/src/fides/api/util/connection_util.py", line 40, in <module>
    from fides.api.service.privacy_request.request_runner_service import (
  File "/fides/src/fides/api/service/privacy_request/request_runner_service.py", line 25, in <module>
    from fides.api.graph.analytics_events import (
  File "/fides/src/fides/api/graph/analytics_events.py", line 16, in <module>
    from fides.api.task.task_resources import TaskResources
  File "/fides/src/fides/api/task/task_resources.py", line 17, in <module>
    from fides.api.service.connectors import (
ImportError: cannot import name 'BigQueryConnector' from partially initialized module 'fides.api.service.connectors' (most likely due to a circular import) (/fides/src/fides/api/service/connectors/__init__.py)
```

### Code Changes

* [x] moved `api.api.v1.endpoint.utils.py` to `api.util.endpoint_utils.py` and updated references
* [x] figure out something for above problem with `fides_connector.py` / `fides_client.py` when running `postgres_setup.py` --> proposed update here is to move `urn_registry.py` to `common.api.v1` rather than `api.api.v1`

### Steps to Confirm

* [x] `nox -s quickstart` and `nox -s dev -- shell postgres` should both work fine
* [x] general smoke testing of application with `fides_env(test)` (just given the number of files touched) - privacy request execution and system management still worked fine.

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`


adamsachs:
@ThomasLaPiana @pattisdr @galvana i've made some progress but my brain's a bit stuck here as i wrap up my day - wrote up where i am and the problem as i see it.

maybe there's a totally different approach here that's either simpler or preferred - i went down this path and didn't really look back!

adamsachs:
> @ThomasLaPiana @pattisdr @galvana i've made some progress but my brain's a bit stuck here as i wrap up my day - wrote up where i am and the problem as i see it.
> 
> maybe there's a totally different approach here that's either simpler or preferred - i went down this path and didn't really look back!

[proposed update](https://github.com/ethyca/fides/pull/3692/commits/809428b208cbc668b48f554bf54971b3cbcbec95) is to move `urn_registry.py` to `common.api.v1` rather than `api.api.v1`

src/fides/common/api/v1/urn_registry.py
pattisdr, line 1: The only other reference to this I found is in docs/fides/docs/development/development_tips.md!

src/fides/api/util/utils.py
```diff
@@ -0,0 +1,164 @@
+from __future__ import annotations
```
pattisdr, line 1: This file looks like a duplicate of renamed src/fides/api/util/endpoint_utils.py? I think we could get rid of this and there's maybe one test pointing to it.

pattisdr (commented):
Local testing worked well, thanks for digging into these circular dependency issues - they can be so time-consuming 😓

src/fides/api/util/utils.py
adamsachs: whoops, that's my mistake -- thank you for catching it!!!

src/fides/common/api/v1/urn_registry.py
adamsachs: addressed in https://github.com/ethyca/fides/pull/3692/commits/8e7406b2427256b04f949f615600b7a2ab8c0a87

src/fides/api/util/utils.py
adamsachs: addressed in https://github.com/ethyca/fides/pull/3692/commits/8e7406b2427256b04f949f615600b7a2ab8c0a87

pattisdr (approved):
🏆

ThomasLaPiana:
I'm just saying this out loud...the majority of these issues were caused specifically by the admin router due to it's need to directly modify the database. While we probably could have avoided most of these changes by directly moving that endpoint into `main.py` or something, I still feel these refactors still actually make the code more navigable 🙂 

Thanks for the changes here @adamsachs !

adamsachs:
> I'm just saying this out loud...the majority of these issues were caused specifically by the admin router due to it's need to directly modify the database. While we probably could have avoided most of these changes by directly moving that endpoint into `main.py` or something (like 4 PRs ago, haha), I still feel these refactors meaningfully contribute to the code being more navigable 🙂
> 
> Thanks for the changes here @adamsachs !

ha - i also made that observation @ThomasLaPiana , thank you for calling it out here! i felt a bit hesitant to move it and "special-case" the endpoint for lack of consistency, but i think you're right to make a case for it. i will say that even that router imports from `from fides.api.api.v1.endpoints import API_PREFIX` so we'd need some further adjustments along the lines of what i've done in this PR, but it does feel like a good change to simplify the dependency tree longer-term!

PR #3694, opened 2023-06-28
Update front-end to treat fields with default values as required

galvana:
Closes https://github.com/ethyca/fidesplus/issues/961

### Description Of Changes

Updating the conditions to validate and flag a field as required. Using the presence of a default value to determine that a field is required.

### Code Changes

* [ ] Updated the `ConnectorParametersForm` under the system page. Skipped making the change for the previous connector manager page since we're phasing that out.

### Steps to Confirm

* [ ] I used Aircall as an example, the `domain` field should be required even though it has a default value.

### Pre-Merge Checklist

* [ ] Update `CHANGELOG.md`


clients/admin-ui/src/features/datastore-connections/system_portal_config/forms/ConnectorParametersForm.tsx
```diff
@@ -142,7 +142,12 @@ const ConnectorParametersForm: React.FC<ConnectorParametersFormProps> = ({
       name={key}
       key={key}
       validate={
-        secretsSchema?.required?.includes(key) || item.type === "integer"
+        secretsSchema?.required?.includes(key) ||
+        Object.prototype.hasOwnProperty.call(
```
galvana, line 146: Not sure if this is the best way to check that this object has a property called `default`.

TheAndrewJackson: This seems like a good spot to use a function since the same check is happening in two places. 

I _think_ you can check if a property exists within an object using the `in` keyword. 


The function might look something like this

```typescript
  const isRequiredSecretValue = (key: string): boolean =>
    secretsSchema?.required?.includes(key) ||
    (secretsSchema?.properties?.[key] !== undefined &&
      "default" in secretsSchema.properties[key]);
```


then the check could be this

```jsx
    <Field
      id={key}
      name={key}
      key={key}
      validate={
       isRequiredSecretValue(key) ||
        item.type === "integer"
          ? (value: string) =>
              validateField(item.title, value, item.allOf?.[0].$ref)
          : false
      }
    >
        {...}
    </Field>

```

Issue #3231, opened 2023-05-05
Prefetch API calls as part of privacy-center in /fides-js API route


eastandwestwind:
Right now our assumption is that the `Fides.js` package will be responsible for making the following API calls during initialization:
1. Fetching it's "geolocation" using a location API
2. Fetching it's "experience" configuration using the Fides API, filtered to the component's location

As an optimization, it should be possible to do both of these API calls server-side and bundle the results into the `/fides.js` package returned to the client. This would allow the `Fides.js` package to initialize instantly, instead of delaying to wait for the API calls to complete.

Kelsey-Ethyca:
@Kelsey-Ethyca to add doc

Roger-Ethyca:
moving to done

PR #3698, opened 2023-06-28
Prefetch geolocation and experience API calls

eastandwestwind:
Closes https://github.com/ethyca/fides/issues/3231

### Description Of Changes

Prefetches geolocation and experience API calls if possible, to reduce round-trip API calls made from the client.

### Code Changes

* [ ] Uses shared fides-js methods to obtain geolocation / experience server-side as part of the /fides-js endpoint
* [ ] Adds unit and e2e tests

### Steps to Confirm

* [ ] Make sure `IS_GEOLOCATION_ENABLED` is `true` and `GEOLOCATION_API_URL` is set up
* [ ] Using an env with notices/experiences, nav to `http://localhost:3000/fides-js-demo.html`
* [ ] See `Consent Experience` and `Geolocation` sections of the page already pre-populated with data

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`


clients/privacy-center/pages/api/fides-js.ts
```diff
@@ -82,7 +110,8 @@ export default async function handler(
       privacyCenterUrl: environment.settings.PRIVACY_CENTER_URL,
       fidesApiUrl: environment.settings.FIDES_API_URL,
     },
-    geolocation,
+    experience: experience || undefined,
```
allisonking, line 113: do we still need this call now that we have prefetching happening?
https://github.com/ethyca/fides/blob/9f081676c6426d6ce7a9d3be3f3adacd52e67df6/clients/fides-js/src/fides.ts#L245-L250

it seems like that would cause 2 calls to the server in the event that no experience exists (prefetch would call once, would say `experience: undefined`, and then fides-js would try to make the call again)

clients/privacy-center/pages/api/fides-js.ts
```diff
@@ -67,6 +70,31 @@ export default async function handler(
...
+    if (fidesRegionString && environment.settings.IS_OVERLAY_ENABLED) {
+      let fidesUserDeviceId = null;
+      if (Object.keys(req.cookies).length) {
+        const fidesCookie = req.cookies[CONSENT_COOKIE_NAME];
+        if (fidesCookie) {
+          fidesUserDeviceId =
+            JSON.parse(fidesCookie)?.identity?.fides_user_device_id;
+        }
+      }
+      console.log("Fetching relevant experiences from server-side...");
```
allisonking, line 88: do we want this? it might spam the logs with how often this endpoint can get called. or maybe we also want to only log depending on `settings.DEBUG`

allisonking (commented):
this looks great! I tested it out and could see the cacheing happening, very cool. just a few questions!

clients/privacy-center/pages/api/fides-js.ts
eastandwestwind: good point, I'll wrap this based on `settings.DEBUG`

clients/privacy-center/pages/api/fides-js.ts
eastandwestwind: `experience: undefined` would be returned in the case where for some reason a customer has overlay enabled but has not setup the necessary ENV vars nor notices to successfully obtain data from this endpoint. So then 2 calls to the server would be made, one from prefetch and one from `fides-js`.

Otherwise, the `fides-js` calls for geolocation and experience currently function as a safeguard in case the pre-fetch would fail for some reason. I'm a little torn whether to remove this now. @NevilleS any opinions here?

NevilleS: Yeah, I don't think I see a way we could fully remove this, since we can't *guarantee* prefetching will work. So making the "standard" case be one where the JS client self-locates and fetches it's experience data makes sense, but then we support an "optimized" case where everything is prefetched and cached server-side to improve page-load times 👍 

As a relatively trivial counter-example: we can only prefetch if we can determine the geolocation server-side. We will never be able to *fully* guarantee that, so let's keep the client code (and tests for this case!) live forever

allisonking (approved):
:shipit:

clients/privacy-center/common/geolocation.ts
```diff
@@ -20,14 +21,15 @@ export const LOCATION_HEADERS = [
...
- * NOTE: This specifically *does not* include performing a geo-IP lookup... yet!
  */
-export const getGeolocation = (
-  req: NextApiRequest
-): UserGeolocation | undefined => {
+export const lookupGeolocation = async (
+  req: NextApiRequest,
+  settings: PrivacyCenterClientSettings
+): Promise<UserGeolocation | null> => {
   // DEFER: read headers to determine & return the request's IP address
```
NevilleS, line 33: @eastandwestwind this is a really important `DEFER` - without doing this, the geoIP is just going to be the IP of the privacy center

NevilleS (commented):
Late review here, but I believe there's a bug here

PR #3700, opened 2023-06-28
3601: test connection results toast

### LINKED ISSUE #3601: Add success and failure toasts when clicking test connection
Currently you get an inline success/failure message with some details about the connection test. We should be using our standard toast to indicate the test was successful or failed. Will handle test status in another ticket.

Steps to reproduce 
1. run fides test env
2. login as root user 
3. click data map in top nav 
4. click the ellipsis of any system 
5. click edit 
6. click integration tab 
7. click the dropdown and choose amplitude 
8. enter amplitude creds from 1password 
9. click save 
10. click test connection 
11. see no toast

jpople:
Closes #3601 

### Description Of Changes

Removed test connection box below "Test connection" button and replaced it with success or failure toasts, depending on result of test.

Success example:

Screenshot 2023-06-28 at 5 42 38 PM

Failure example:

Screenshot 2023-06-28 at 5 41 10 PM

### Code Changes

- Repurposed old `TestConnection` component to be a standalone component for the toast
- Changed `HandleTestConnectionClick` method in `ConnectorParametersForm` to prevent the method from being run repeatedly; errors are now handled by toasting in the `ConnectorParameters` above

### Steps to Confirm

1. From the "Configure your system" view in "Data map", navigate to the "Integrations" tab
2. Add valid credentials to the connection, then click "Test connection"; note toast and no test result box at bottom of page
3. Change connection credentials to be invalid, then click "Test connection" again; note toast and no test result box at bottom of page

### Pre-Merge Checklist

* [X] Update `CHANGELOG.md`


clients/admin-ui/src/features/datastore-connections/system_portal_config/forms/ConnectorParametersForm.tsx
```diff
@@ -238,23 +235,10 @@ const ConnectorParametersForm: React.FC<ConnectorParametersFormProps> = ({
   };
 
   const handleTestConnectionClick = async () => {
-    try {
-      await trigger(connectionConfig!.key).unwrap();
-    } catch (error) {
-      handleError(error);
-    }
+    const result = await trigger(connectionConfig!.key);
+    onTestConnectionClick(result);
```
allisonking, line 239: right now `onTestConnectionClick` is typed as `(value: any) => void` which isn't great. I know that already existed before this PR, but maybe while we are here, we can clean that up a bit 🧹 

unfortunately that type is difficult to track down. we can make our own type though which captures the fields we need (in this case `data` and `fulfilledTimeStamp`)

```typescript
export interface TestConnectionResponse {
  data?: DatastoreConnectionStatus;
  fulfilledTimeStamp?: number;
}
```
so then we can update to `onTestConnectionClick: (value: TestConnectionResponse) => void;`

and also in `ConnectorParameters.tsx`, `response` no longer has to be of type `any`, we can import our new `TestConnectionResponse` type instead. it does yield a few typing errors though, but that's a good thing! since we should handle the case where `fulfilledTimeStamp` is undefined, etc.

clients/admin-ui/src/features/datastore-connections/system_portal_config/forms/ConnectorParametersForm.tsx
```diff
@@ -238,23 +235,10 @@ const ConnectorParametersForm: React.FC<ConnectorParametersFormProps> = ({
   };
 
   const handleTestConnectionClick = async () => {
-    try {
```
allisonking, line 241: woof okay I think I may have given you not great advice removing this `unwrap`. the case that this breaks is if the fetch itself fails, for example if the backend server is down, then an error will be thrown and nothing catches it. (you can test this by getting your admin UI to this screen so it's ready to test the connection, then shutting your backend down, then clicking "Test connection"). adding the try catch back doesn't fix things, since I'm guessing the `unwrap` is forcing the error to throw. but I don't want to add the `unwrap` back since then we lose `fulfilledTimeStamp` 😐 

it seems to originate in the [redux slice here](https://github.com/ethyca/fides/blob/3601-test-connection-toasts/clients/admin-ui/src/features/datastore-connections/datastore-connection.slice.ts#L311). all of that logic under `onQueryStarted` is very unusual IMO (most of our other slice files when defining endpoints are much simpler) and should probably be refactored. I think we can remove the `try catch` inside `datastore-connection.slice` and then that'll let the error bubble up to the toast to be handled. but it might mess up `ConnectionGrid` which also uses that query.

it might be that this is too complicated to roll into this ticket, in which case let's make a follow up issue for it 👍

allisonking (changes requested):
this turned out to be pretty complicated for a toast, didn't it? 😄 nice job tackling it, just a few suggestions, and one bigger thing that might need to be rolled out into its own ticket

clients/admin-ui/src/features/datastore-connections/system_portal_config/forms/ConnectorParametersForm.tsx
```diff
@@ -35,12 +34,18 @@ import {
   SystemType,
 } from "~/types/api";
 
+import { DatastoreConnectionStatus } from "../../types";
```
allisonking, line 37: nit: prefer using an absolute import (i.e. `~/features/...`)

allisonking (approved):
nice job turning this around!! left a very nit comment. other than that, can you make a follow up issue out of [this bug](https://github.com/ethyca/fides/pull/3700#discussion_r1246742483) we talked about? you can add it to the roadmap team project, and also assign it to the next sprint (doesn't mean it'll get worked on next sprint, but gives product a chance to see the ticket and decide its priority!)

PR #3717, opened 2023-06-30
Treat Underscore as a Special Character for Passwords

### LINKED ISSUE #3716: Underscore not a valid Special Character in Passwords
### Bug Description

When creating a password for a user in Fides, the underscore is not recognized as a special character, which renders passwords using it to also need additional special characters.

#### Steps to Reproduce

1. Browse to http://localhost:3000/user-management/new
2. Enter a password like `ADQ8pxz8ahb2dwc_hpb`

image

Note this is not a real password 😄  

### Expected behavior

`_` should be considered a special character

### Environment

* Version: Tested on `2.15.0`
* OS:
* Python Version:
* Docker Version:

### Additional context

RobertKeyser:
Closes #3716 

### Description Of Changes

This PR adds `_` to the regex that tests for special characters in password on both the client and server.

### Code Changes

* Update validation regex on Admin UI
* Update validation regex on Fides Webserver

### Steps to Confirm

Note: 

* Create a new user with a valid password with the only special character being `_`
* Reset an existing user's password to a valid password with the only special character being `_`

### Pre-Merge Checklist

* Documentation:
  * ~documentation complete, [PR opened in fidesdocs](https://github.com/ethyca/fidesdocs/pulls)~
  * ~documentation [issue created in fidesdocs](https://github.com/ethyca/fidesdocs/issues/new/choose)~
* ~Relevant Follow-Up Issues Created~
* [x] Update `CHANGELOG.md`
* ~For API changes, the [Postman collection](https://github.com/ethyca/fides/blob/main/docs/fides/docs/development/postman/Fides.postman_collection.json) has been updated~


RobertKeyser:
I would have added a test or two, but I couldn't find anywhere we test for valid passwords, only bad password.

src/fides/api/schemas/user.py
```diff
@@ -46,7 +46,7 @@ def validate_password(cls, password: str) -> str:
             raise ValueError("Password must have at least one capital letter.")
         if re.search("[a-z]", decoded_password) is None:
             raise ValueError("Password must have at least one lowercase letter.")
-        if re.search(r"[\W]", decoded_password) is None:
+        if re.search(r"[\W_]", decoded_password) is None:
```
pattisdr, line 49: Great, can you add another test similar to `test_create_user`?
https://github.com/ethyca/fides/blob/main/tests/ops/api/v1/endpoints/test_user_endpoints.py#L145

That test has a request body of `body = {"username": "test_user", "password": str_to_b64_str("TestP@ssword9")}`

Your test can use a password instead that has an underscore -

RobertKeyser: Thanks for the suggestion, just added it!

pattisdr (approved):
nice @RobertKeyser 

I didn't really consider underscore as a special character, good to know! https://owasp.org/www-community/password-special-characters

PR #3719, opened 2023-06-30
[Backend] Fix Validation Error where Current Privacy Preferences might be None

pattisdr:
Closes https://github.com/ethyca/fidesplus/issues/979

### Description Of Changes

Under higher loads, saving Privacy Preferences could return Pydantic Validation Errors because the CurrentPrivacyPreferenceSchema is not supposed to be None. 

```
2023-06-27 15:53:55.491 [ERROR] (logger:_log_exception:34): 1 validation error for CurrentPrivacyPreferenceSchema
response -> 0
  none is not an allowed value (type=type_error.none.not_allowed)
```

The root cause is that if a single user saves their preferences in rapid succession, the older PrivacyPreferenceHistory history record will no longer have a relationship with the CurrentPrivacyPreference record because a newer PrivacyPreferenceHistory had superseded it.  

### Code Changes

* [x] Add a new method to PrivacyPreferenceHistory that returns both the created  PrivacyPreferenceHistory and upserted CurrentPrivacyPreference record instead of relying on the relationship between the PrivacyPreferenceHistory and the CurrentPrivacyPreference to fetch the latter after the fact before returning a response.   The response should return the CurrentPrivacyPreference upserted at that moment, even if the FK has since been replaced.

### Steps to Confirm

* [x] I would just validate that saving preferences still works - PATCH {{host}}/privacy-preferences

Example request:
```
{
   "browser_identity": {
       "fides_user_device_id": "75495a74-fc09-4e21-8250-0d53b854877d"
   },
   "preferences": [{
       "privacy_notice_history_id": "pri_d6bbc55d-bc06-4b7c-858b-016ed280e668",
       "preference": "opt_out"
   }],
   "user_geography": "us_ca",
   "privacy_experience_id": "pri_475641b6-75e4-4e77-8a30-d9640c3616e7",
   "method": "button"
}
```

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`


pattisdr:
@adamsachs hoping you could review this small change - it was the [source of the first error we were running into here](https://ethyca.atlassian.net/wiki/spaces/EN/pages/2730426372/2023-06-29+Consent+Load+Testing+Results), creating higher than expected failure rates.

adamsachs (approved):
nice simple fix! i'm sure it was tougher to track down the timing problem than it was to actually fix it once you found it 😄 

manually verified no obvious regressions in the `PATCH /api/v1/privacy-preferences` endpoint per your steps.

have you confirmed this fixes the problem you'd seen in load testing? not a prerequisite for merging, just curious if that's something you'd been able to do before merging.

pattisdr:
@adamsachs yes! you're able to hit this error locally if you send requests at a high enough rate, and I did verify that this particular error has gone away.

Thanks for reviewing this!

pattisdr:
Failing test not related and ticketed here: https://github.com/ethyca/fides/issues/3736

PR #3734, opened 2023-07-05
3632 clickable system tiles

### LINKED ISSUE #3632: Make tiles clickable in View Systems page
AC 
- When clicking the tiles of the View Systems page nav to `/systems/configure/{system-fides-key}` as if you clicked on the ellipsis then edit 
- The ellipsis should remain with the menu items edit and delete, functioning as it does today 

@rsilvery if we want to remove the ellipsis menu we will need to add a delete option in the system config

jpople:
Closes #3632 

### Description Of Changes

Makes system cards on "view systems" page clickable and link to that system's `configure/[id]` page.

### Steps to Confirm

- Go to "view systems" page
- Click on a system's box in the grid

### Pre-Merge Checklist

* [x] Update `CHANGELOG.md`


clients/admin-ui/src/features/datastore-connections/system_portal_config/forms/ConnectorParametersForm.tsx
```diff
@@ -25,7 +25,6 @@ import { useLazyGetDatastoreConnectionStatusQuery } from "datastore-connections/
 import DSRCustomizationModal from "datastore-connections/system_portal_config/forms/DSRCustomizationForm/DSRCustomizationModal";
 import { Field, FieldInputProps, Form, Formik, FormikProps } from "formik";
 import React from "react";
-import { DatastoreConnectionStatus } from "src/features/datastore-connections/types";
```
allisonking, line 28: hm let's undo this change? seems unrelated, and I prefer the absolute import

clients/admin-ui/src/features/system/SystemCard.tsx
```diff
@@ -57,7 +57,7 @@ const SystemCard = ({ system }: SystemCardProps) => {
 
   return (
     <Box display="flex" data-testid={`system-${system.fides_key}`}>
-      <Box flexGrow={1} p={4}>
+      <Box flexGrow={1} p={4} as="button" onClick={handleEdit}>
```
allisonking, line 60: this works, but technically isn't semantically correct: https://stackoverflow.com/questions/12982269/is-it-semantically-incorrect-to-put-a-div-or-span-inside-of-a-button

that's not the biggest deal, but it also changes the rendering a bit:
| Before | After |
|-------|-------|
| image| image|

I think a way that could solve both problems is, instead of using a button component, wrap the box in a `NextLink`. We do a similar thing in the home page: https://github.com/ethyca/fides/blob/f526d9ffb176006d701493c9d0eff6b4884e811f/clients/admin-ui/src/home/HomeContent.tsx#L33

Incidentally, I think [this line](https://github.com/ethyca/fides/blob/f526d9ffb176006d701493c9d0eff6b4884e811f/clients/admin-ui/src/features/system/SystemCard.tsx#L41) may be redundant now, since we also set the active system in the configure page:
https://github.com/ethyca/fides/blob/f526d9ffb176006d701493c9d0eff6b4884e811f/clients/admin-ui/src/pages/systems/configure/%5Bid%5D.tsx#L37-L39
so you don't necessarily have to rely on `handleEdit`, it should be enough to just pass the route to a `NextLink`

allisonking (changes requested):
ah, it's nice to be able to click directly on the card 😄 I think we should try implementing this through links though, let me know what you think?

also, this would be a good opportunity to add a cypress test!

allisonking (approved):
💯 nice work!

PR #3746, opened 2023-07-07
ignore type checker on `click` decorators to bypass known library issue

### LINKED ISSUE #3745: Fix mypy failures in `main` caused by updated `click` library
### Bug Description

Changes introduced in the newly released `click` [version `8.1.4`](https://github.com/pallets/click/releases/tag/8.1.4), which now gets pulled into freshly baked images as a dependency, are seemingly causing mypy failures (e.g. https://github.com/ethyca/fides/actions/runs/5487381560/jobs/9998767180) in our CI runs. From an initial analysis, it seems like [these changes](https://github.com/pallets/click/pull/2401/files) specifically are what's responsible for the new mypy failures.

UPDATE: found https://github.com/pallets/click/issues/2558 which confirms the above, and also that many others are seeing this same problem.

https://github.com/ethyca/fides/issues/3749 is a follow-up tech debt ticket that can be worked on if/when the underlying `click` issue above is fixed and released.

#### Steps to Reproduce

1. `nox -s clean`
2. `nox -s dev -- shell`
3. in the container shell, run `mypy` and notice failures

### Expected behavior

mypy should run successfully on our codebase! :)

### Environment

The only thing to note here is that local docker cache can seemingly prevent this from appearing by not getting the latest `click` version and relying on a cached, older version instead. Running `nox -s clean` clears out this cache, and allows the error to be reproducible.

Most importantly, the error is consistently reproducible on CI runs.

### Additional context

This is purely a CI/static check bug, so although it's annoying, it doesn't impact application runtime (as far as i can tell). Still, we should get it fixed.

adamsachs:
Closes #3745 

### Description Of Changes

Ignore type checker on problematic function calls with decorator wrappers. This is due to a change in `click` version `8.1.4` that many others are also hitting: https://github.com/pallets/click/issues/2558

See issue descriptions (ours and theirs) for more detail/context.

Follow up tech debt ticket to revert this eventually, when possible: https://github.com/ethyca/fides/issues/3749

### Code Changes

* [ ] add comment to ignore type checker on problematic function calls

### Steps to Confirm

* [ ] `mypy` static check passes in CI

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@ The types of changes are:
 - Treat underscores as special characters in user passwords [#3717](https://github.com/ethyca/fides/pull/3717)
 - Allow Privacy Notices banner and modal to scroll as needed [#3713](https://github.com/ethyca/fides/pull/3713)
 - Make malicious url test more robust to environmental differences [#3748](https://github.com/ethyca/fides/pull/3748)
+- Ignore type checker on click decorators to bypass known issue with `click` version `8.1.4` [#3746](https://github.com/ethyca/fides/pull/3746)
 
 ### Changed
 
--- a/src/fides/cli/__init__.py
+++ b/src/fides/cli/__init__.py
@@ -64,7 +64,7 @@ APP = fides.__name__
 PACKAGE = "ethyca-fides"
 
 
-@group(
+@group(  # type: ignore
     context_settings=CONTEXT_SETTINGS,
     invoke_without_command=True,
     name="fides",
--- a/src/fides/cli/commands/annotate.py
+++ b/src/fides/cli/commands/annotate.py
@@ -14,7 +14,7 @@ def annotate(ctx: click.Context) -> None:
     """
 
 
-@annotate.command(name="dataset")
+@annotate.command(name="dataset")  # type: ignore
 @click.pass_context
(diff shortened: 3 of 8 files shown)


adamsachs:
> Is there a way to just ignore all typing error from a library in the pyproject.toml?
> 
> For example I think this works to ignore a bunch of mypy failures for all our tests: https://github.com/ethyca/fides/blob/main/pyproject.toml#L30

hmm, i think there _should_ be in theory -- i'd expect it to look like this: 
```
[[tool.mypy.overrides]]
module = ["click.*"]
ignore_errors=true
``` 
but that's not working when i add it. as a sanity check, when i added this
```
[[tool.mypy.overrides]]
module = ["fides.cli.options"]
ignore_errors=true
```
mypy did stop complaining.

i can keep on digging on it, perhaps i'm just doing something silly, but even if i get it right, it feels a bit heavy-handed to me, i'd prefer not to ignore all type checking for the module if possible? although it's certainly an easier change to revert if/when the underlying problem is fixed 😄

galvana (approved):
Maybe not be as pretty as we'd like but it works. It avoids side-effects that come with more general error suppression or version pinning. Let's go!

PR #3758, opened 2023-07-10
Refactor logging configuration

### LINKED ISSUE #3725: Enable serializing logs as JSON

### LINKED ISSUE #3729: Logging fails on 500s because the process dies, so we get the stack trace but not the proper log message
This will require a general refactor of how things are processed, meaning everything should get wrapped in `try` statements so we can make sure we log things out even if an exception gets thrown

ThomasLaPiana:
Closes #3725 
Closes #3729 

### Description Of Changes

This PR updates the configuration of `loguru` to be more transparent and user-friendly. It also fixes some outstanding issues with the current logging implementation.

### Code Changes

* [x] update how logging is configured to be more transparent
* [x] verify that JSON serialization works as expected
* [x] remove file/func/line logging from prod mode
* [x] make sure that logging succeeds on 500 errors

### Steps to Confirm

Testing the logging changes
* [ ] remove the `colorize` env var from the `docker-compose.yml` file to confirm it will turn color logging off
* [ ] remove `FIDES__DEV_MODE` and confirm that the file name, function name and line number go away in the logs
* [ ] with colorization still disabled, add the `FIDES__LOGGING__SERIALIZATION: "json"` env var to the compose file and run the server again, confirm it is serialized as JSON

Testing the 500 fix
* [ ] `nox -s dev`
* [ ] `fides user login` in another terminal
* [ ] open the `.fides_credentials` file and remove some characters from the key (this triggers a `JWEError`)
* [ ] run any CLI command such as `fides ls data_category`
* [ ] confirm the logs from `nox -s dev` show the exception as well as the failed request

Colorized Dev Mode
image

Prod Mode
image

JSON
image

### Pre-Merge Checklist

* Documentation:
* [x] ~~Relevant Follow-Up Issues Created~~ Already part of an epic
* [x] Update `CHANGELOG.md`


src/fides/api/api/v1/endpoints/dataset_endpoints.py
```diff
@@ -69,6 +66,10 @@
     YAML_DATASETS,
 )
 
+from fides.api.models.sql_models import (  # type: ignore[attr-defined] # isort: skip
```
ThomasLaPiana, line 69: @pattisdr ok I think this _finally_ fixes the isort/mypy issue!

src/fides/api/main.py
```diff
@@ -133,7 +133,13 @@ async def prepare_and_log_request(
 async def log_request(request: Request, call_next: Callable) -> Response:
     """Log basic information about every request handled by the server."""
     start = datetime.now()
-    response = await call_next(request)
+
+    # If the request fails, we still want to log it
+    try:
```
ThomasLaPiana, line 138: this fixed the bug where we would log the exception but not the endpoint, etc.

src/fides/api/util/logger.py
```diff
@@ -42,77 +41,54 @@ def _log_warning(exc: BaseException, dev_mode: bool = False) -> None:
         logger.error(exc)
 
 
-class FidesAPIHandler(logging.Handler):
```
ThomasLaPiana, line 45: this class felt overly complicated to me...why do with classes what can be done with functions? No magic overriding of the handler function, just building things transparently and adding them to `loguru`

adamsachs:
@ThomasLaPiana only had time to go thru steps to confirm today, and those generally looked good! i will get to an actual code review tomorrow my time.

one thing that i did notice was a rather odd JSON log serialization -- see below. this happened on initial server bootup and didn't reappear afterwards, so may not be too big of a deal, but figured i'd flag it anyway. if it's not something particular about my env, i'd say we should probably get it sorted, it's a bit ugly. are you able to repro?  

image

tests/ctl/api/test_admin.py
```diff
@@ -26,12 +26,13 @@ def test_db_reset_dev_mode_disabled(
     test_config: FidesConfig,
     test_config_dev_mode_disabled: FidesConfig,  # temporarily switches off config.dev_mode
     test_client: TestClient,
+    loguru_caplog,
 ) -> None:
-    with pytest.raises(
```
ThomasLaPiana, line 30: Since we're now catching exceptions instead of letting them exit the process, this still gets logged but doesn't blow things up (which the test expected)

ThomasLaPiana:
> @ThomasLaPiana only had time to go thru steps to confirm today, and those generally looked good! i will get to an actual code review tomorrow my time.
> 
> one thing that i did notice was a rather odd JSON log message serialization coming from `setup_server` -- see below. after some poking around and doing some basic application things, i didn't see these types of logs coming from anywhere else in the codebase besides this one function on server startup, so may not be too big of a deal - but figured i'd flag it anyway. if it's not something particular about my env, i'd say we should probably get it sorted, it's a bit ugly. are you able to repro?
> 
> image

This is the ASCII `Fides` banner 😂 I noticed that too but wasn't quite sure what to do...seemed weird to add a conditional check on whether to print it or not but I guess that makes the most sense?

adamsachs (approved):
this looks really good to me! code changes make sense and leave the logging setup very easy to follow. all the new functionality looks good from an end-user perspective, too. thanks again for making the change to handle the ascii art!

PR #3775, opened 2023-07-11
show/hide connector values

### LINKED ISSUE #3380: Update front-end to show values and hide sensitive values
Blocked by design decision needed in https://github.com/ethyca/fides/issues/3537

TheAndrewJackson:
Closes #3380 

### Description Of Changes

Update fides to return connection config secrets by default. It will **ALWAYS** mask sensitive secret values by utilizing the secret schema. Sensitive values will be return as `**********`.

This allows the admin ui to display the non sensitive secrets and the masked sensitive secrets in the integration configuration page. It improves the UX by showing the user that the secrets are configured instead of always showing blank input fields.

### Code Changes

* [ ] Return connection config secrets. The sensitive secrets are masked!
* [ ] Add a new secrets patch endpoint nested under `/system` that only updates secret values that were included in the payload. This is so sensitive secrets don't get overwritten with `**********` values.
* [ ] Update the admin to show secrets and in the UI and use the new patch endpoint

### Steps to Confirm

* [ ] run fides and the admin ui
* [ ] create a system
* [ ] add an integration to that system that have sensitive values. Like Aircall for  example
* [ ] Confirm that the sensitive values are masked in the response

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`


src/fides/api/schemas/connection_configuration/connection_config.py
```diff
@@ -47,43 +49,34 @@ class Config:
...
+    print("secret_schema\n\n")
+    pprint(secret_schema)
+    if isinstance(connection_secrets, dict):
+        for key, val in connection_secrets.items():
+            prop = secret_schema["properties"].get(key, {})
+            if isinstance(val, dict):
+                mask_sensitive_fields(val, prop)
+            elif prop.get("sensitive", False):
+                connection_secrets[key] = "**********"
+    return connection_secrets
```
galvana, line 79: Let's update this to omit fields that are not in the schema. This is a tricky one to catch but for OAuth connections, we store `access_token` and `refresh_token` in the `connectionconfig.secrets`. They're not in the schema because these aren't values we need the user to provide, they're added as part of the OAuth flow. Without this omission, these tokens are returned to the front-end in plain text.

TheAndrewJackson: great catch. I'll update it to handle that now

galvana (approved):
Thanks for making the suggested changes, I'm glad this feature is finally going in!

src/fides/api/schemas/connection_configuration/connection_config.py
```diff
@@ -47,43 +49,41 @@ class Config:
...
+def mask_sensitive_fields(
+    connection_secrets: Dict[str, Any], secret_schema: Dict[str, Any]
+) -> Dict[str, Any]:
     """
-    Describes the returned schema for connection types
+    Mask sensitive fields in the given secrets based on the provided schema.
+    This function traverses the given secrets dictionary and uses the provided schema to
+    identify fields that have been marked as sensitive. The function replaces the sensitive
+    field values with a mask string ('********').
+    Note: This function modifies the input dictionary in-place.
```
galvana, line 60: Let's remove this from the docstring since the input isn't being modified in place anymore.

Issue #3800, opened 2023-07-17
build broken due to upstream libraries and `cython==3.0` compatibilities


adamsachs:
### Bug Description

https://github.com/yaml/pyyaml/issues/601 seems to be causing build problems for any project pinning `pyyaml` below version `6`, as we are (it seems that the problem relates specifically to `5.4.1`, which is the highest `<6` version?).

see e.g. failed output from latest CI build on main: https://github.com/ethyca/fides/actions/runs/5577529613/jobs/10190410537

additionally - as far as i can tell, there is a separate (but presumably related) issue with `pymssql` and its compatibility with `cython` version `3.0`, which we hit once we work around the above `pyyaml` issues. i don't see the `pymssql` issue explicitly reported, though it seems like it may be similar to https://github.com/pymssql/pymssql/issues/819 and https://github.com/pymssql/pymssql/issues/763. i don't feel confident enough diagnosing the `pymssql` problem yet to file an issue in that repo
    - UPDATE: it seems like this is indeed the case, see https://github.com/pymssql/pymssql/issues/826

#### Steps to Reproduce

try running e.g. `nox -s dev` on `main`

### Expected behavior

our build needs to work

### Additional context

see slack discussion https://ethyca.slack.com/archives/CBKK8TS74/p1689616525035169 for different issues that have come up when trying to resolve...

Issue #217, opened 2021-11-09
Loosen installation requirements


ThomasLaPiana:
Right now the versions we have pinned are too specific and could make Fides difficult to use in existing Python projects. We need to make them looser and give ranges instead of exact matches.

Main issue here is how to test it

earmenda:
Peeked over at fidesops and they also have some strict requirements at the moment but also use `~=` for some:
https://github.com/ethyca/fidesops/blob/main/requirements.txt

I would think that adding wildcards to minor and patch versions would be safe. We could confirm that for each library they follow semantic versioning for releases.

ThomasLaPiana:
fidesops is different in that they are meant to be a standalone application that is run in a docker container, but there are situations in which someone would want to add fidesctl to their own python project dev requirements.

I agree on the wildcard for minor/patch, lets start there as a first iteration and figure out if there is any extra testing we want to do around that

ThomasLaPiana:
My method of attack here is going to be to pin everything to the lowest minor/major version that they're at now, confirm they all work, and then loosen the requirements and test it again with the highest version

PR #3802, opened 2023-07-17
fix broken build (`pyyaml` and `pymssql` issues related to `cython` version `3.0`)

adamsachs:
works on #3800 

### Description Of Changes

the `pyyaml` update here is relatively straightforward and understood - `pyyaml>=6.0` seems to fix the compatibility issues that `pyyaml<6.0` had with `cython>=3.0`, namely use of a `setup.cfg` file. upgrading to `pyyaml>=6.0` seems like it _should_ be fine for us - at first glance, i can't find any concrete reason we're constrained lower, besides probably a general caution when first unpinning the dependency (see https://github.com/ethyca/fides/issues/217#issuecomment-985018791 for what i think is the context there). and upgrading to `pyyaml>=6.0` seems like a viable workaround from at least some others (e.g. https://github.com/yaml/pyyaml/issues/601#issuecomment-1638573950).

i also have here a temporary removal of the `pymssql` dependency, since that _also_ was causing build issues. as far as i can tell, those issues are _separate_ from the above `pyyaml` changes, but they evidently have the same root cause, i.e. the new `cython 3.0` incompatibilities. this hypothesis is based partly on the error messages i'm getting from `pymssql` during our build, as well as the fact that it clearly has a `setup.cfg` [file](https://github.com/pymssql/pymssql/blob/master/setup.cfg). but i'm still not 100% sure on my analysis there...the important thing is that i could _not_ get `pymssql` to behave no matter what i did. i even tried a specific `RUN pip install "Cython<3.0" "pymssql" --no-build-isolation ` build phase in our `Dockerfile` to no avail. so for now, i've just removed the dependency to get us unblocked a bit...
 --> UPDATE: looks like there are some explicit complaints now coming to `pymssql`: https://github.com/pymssql/pymssql/issues/826

### Code Changes

* [x] update to _not_ constraint `pyyaml<6.0`
    * [x] this also requires an update to `fideslang==1.4.3` since we'd been constraining/pinning `pyyaml` there too in `fideslang==1.4.2`
* [x] temporarily remove `pymssql` dependency

### Steps to Confirm

* [ ] get CI running (_mostly_) successfully
* [x] make sure our app functions basically
    * note: eventually we need to ensure no other regressions with the `pyyaml` upgrade, but for now, let's just get things working on a basic level...

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -13,7 +13,7 @@ expandvars==0.9.0
 fastapi[all]==0.89.1
 fastapi-caching[redis]==0.3.0
 fastapi-pagination[sqlalchemy]~= 0.10.0
-fideslang==1.4.2
+fideslang==1.4.3
 fideslog==1.2.10
 firebase-admin==5.3.0
 GitPython==3.1.31
@@ -39,7 +39,7 @@ pymongo==3.13.0
 pymssql==2.2.7
 PyMySQL==1.0.2
 python-jose[cryptography]==3.3.0
-pyyaml>=5,<6
+pyyaml>=5,<7
 redis==3.5.3
 RestrictedPython==6.0.0
 rich-click==1.6.1
--- a/requirements.txt
+++ b/requirements.txt
@@ -36,7 +36,7 @@ pydantic==1.10.9
 pydash==6.0.2
 PyJWT==2.4.0
 pymongo==3.13.0
-pymssql==2.2.7
+# pymssql==2.2.7 -- temporarily commented out due to what seems like cython 3.0 issues - see https://github.com/ethyca/fides/issues/3800
 PyMySQL==1.0.2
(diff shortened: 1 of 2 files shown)


adamsachs:
ok @pattisdr @ThomasLaPiana  - this is obviously not perfect, but i think it can at least help get us closer to normal on `main` in an interim period, so i think it is worth considering a quick merge...

pattisdr (approved):
thanks @adamsachs

adamsachs:
merging for now, we'll need to follow up with remaining items soon but this should help unblock at least some of our build/CI...

ThomasLaPiana:
issue here: https://github.com/pymssql/pymssql/pull/803

PR #3804, opened 2023-07-18
TCF Backend

### LINKED ISSUE #3817: Polish Backend TCF Data Model + Endpoint Changes

Early backend TCF POC completed last sprint.  https://github.com/ethyca/fides/pull/3804/files - very rough, code needs to be cleaned up, made more DRY, edge cases handled - 

Polish this into something mergeable.   This increment does not return real data yet.

- Endpoint that lets you turn on TCF
- Ship with new TCF Overlay experiences
- Embed TCF Data uses, vendors, and features in the TCF Overlay Experience response 
- Allow saving preferences against data uses, vendors, and features specifically
- Retrieve previously saved and served preferences for data uses, vendors, and features when retrieving an experience for a given user.

### Additional context

### LINKED ISSUE #3822: Connect TCF Backend to Return Proper Data

TCF POC Branch here https://github.com/ethyca/fides/pull/3804 pulls a lot of the contents for the TCF Experience from a yaml file with TCF data uses that have mostly null fields.

When building the TCF Overlay Experience, we need to pull from a a proper mapping of TCF data uses as well as pull some of these details from the system itself.

### Additional context

### LINKED ISSUE #3876: Backend: Add Remaining TCF Overlay Contents

This PR adds the starting point for a TCF overlay: https://github.com/ethyca/fides/pull/3804, with details about purposes and special purposes returned, some minimal info about vendors, and a framework for features, and special features but no data returned yet. 

We don't yet have all the details we need on systems yet to be able to populate this.

Big picture, we need to add vendor, features, special features, and custom purposes to the TCF overlay:

- Pull vendor field directly from System instead of from attached integration
- Populate Vendor sections with top-level description, is_gvl, etc.  Also populate Vendor special feature, feature, and custom purpose embedded sections
- Query for matching features and special features on systems and return those contents in the TCF overlay
- Populate is_gvl information

### Additional context

pattisdr:
Closes #3817 
Closes #3822 
Closes #3876 
Closes https://github.com/ethyca/fides/issues/3935
Closes https://github.com/ethyca/fidesplus/issues/1051

❗ Contains migration; bump downrev before merge

### Description Of Changes

Adds the backend for being able to build a TCF overlay.

Importantly:
- Adds a new type of experience: the TCF Overlay, that is only returned when the customer has TCF enabled and the user is in an EEA country.
- The contents for the TCF overlay are generated at runtime by looking at what data uses are on current systems (and soon) what features are on systems, to return Purposes, Special Purposes, Vendors (where we have a vendor_id, in other words, we know what type of system we have), Features and Special Features, and Systems (no vendor id).  
   - There are no Privacy Notices associated with a TCF experience.  You save preferences against TCF components directly.
   - Data uses need to be an "exact match" unlike Privacy Notices where we allow hierarchy-matching on data use
 - Adds the ability to save preferences with respect to individual TCF purposes, special purposes, vendors, special features, and features, and systems.
 - Adds the ability to save when individual TCF components were served.
 - When building the TCF Experience, surfaces any previously-saved or previously-served details alongside the TCF data. 
- Relaxes constraints on PrivacyPreferenceHistory, ServedNoticeHistory, CurrentPrivacyPreference, and LastServedNotice to use these tables to store when preferences were saved and served against not just notices, but TCF components as well.
- 

### Code Changes

* [x] Add a new ConsentSettings table and endpoints to enable TCF - this table will have gpp settings added in the future
* [x] Update GET Privacy Experience to be able to return a TCF overlay instead of a regular overlay provided TCF is enabled and the customer is in an EEA country
    * [x] Existing `has_notices` param's meaning has been extended to mean "does the experience have content?" - so does it have notices _or_ TCF content
    * [x] Method added: `embed_experience_details` that either populates the Experience with either TCF content, or with notices (existing behavior), depending on the type of experience
    * [x] New method: `cache_saved_and_served_on_consent_record` that pulls previous records where consent was saved and served both for privacy notices or individual TCF components
* [x] `purpose`, `special_purpose`, `feature`, `special_feature`, `system` and `vendor` columns added to PrivacyPreferenceHistory (tracks all saved preferences), ServedNoticeHistory (tracks all served records), CurrentPrivacyPreference (tracks most current preference across version/time/multiple identities for the same user), and LastServedNotice (tracks most current served across version/time/multiple identities) tables to record serving or saving consent against individual TCF components, not just privacy notices
    * [x] Makes privacy notice related fields nullable since these tables are no longer just for privacy notices exclusively
    * [x] Updates PrivacyPreferenceHistory.create to save preferences against privacy notices _or_ individual TCF components
    * [x] Updates ServedNoticeHistory.create to save that notices _or_ individual TCF components were served
    * [x] Updates `upsert_last_saved_record` to be more generic - we have records of when consent was _last_ saved or served to make for a faster query when building Experiences.  These records are upserted every time consent is saved or served.  Modify to share between notices and individual TCF components
    * [x] Changes to internal method around saving when notices were served as these tables are now shared with preferences.  Table names and endpoints have the same name still. 
* [x] Updates the meaning of the `PrivacyPreferenceHistory.relevant_systems` column to be any systems related to that preference calculated at runtime, removing the logic that it needed to be a system for which we were also propagating backend consent
* [x] Ensure that preferences saved against TCF components are not propagated to backend systems for now

### Steps to Confirm

- [x] You need to be coming in from a region in the EEA like France
- [x] Also, tcf_enabled needs to be set to True: make a PATCH to /consent-settings and set tcf_enabled: true in the request body
- [ ] You need a system with at least one data use that belongs to a GVL purpose or special purpose.  For example, [data_use: "essential.fraud_detection" maps to Special Purpose #1](https://github.com/ethyca/fideslang/compare/main...asachs/tcf-mappings#diff-a4adb03db889d2228be2f2adbcfe3ee6a2964e133c37f190bc34ae8cf8b0a7fbR117)
- [ ] That data use also needs to have a legal basis for processing of "Consent" or "Legitimate interests"
- [ ] If you want a system to show up under "tcf_vendors", add a `vendor_id` to the system in the backend. Otherwise it will show up under "tcf_systems"

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`


src/fides/api/api/v1/endpoints/consent_settings_endpoints.py
```diff
@@ -0,0 +1,44 @@
...
+
+router = APIRouter(tags=["Consent Settings"], prefix=urls.V1_URL_PREFIX)
+
+
+@router.get(
+    urls.CONSENT_SETTINGS,
+    dependencies=[Security(verify_oauth_client, scopes=[scopes.CONSENT_SETTINGS_READ])],
+    response_model=ConsentSettingsResponseSchema,
+)
+def get_consent_settings(*, db: Session = Depends(deps.get_db)) -> ConsentSettings:
```
adamsachs, line 25: a very specific comment (this is obviously one of the lest foundational pieces of this PR), but this caught my attention while watching your (very well done and informative!) loom: it may make sense to make this just any other "application setting" and [expose the setting](https://github.com/ethyca/fides/blob/main/src/fides/api/schemas/application_config.py) to our `[Config` API](https://github.com/ethyca/fides/blob/main/src/fides/api/api/v1/endpoints/config_endpoints.py#L41) so that it's settable via API?

i can discuss how that works if it's not clear from the code (code still needs to  be improved to make _all_ of our application settings configurable via API automatically, but that's been a longer-term TODO)!

anyway, it's not a big deal at all -- and i promise to go thru the more substantive pieces of this code soon! -- but figured i'd point it out quickly in case it's something you feel is appropriate to adjust!

pattisdr: good callout, thank you Adam! I had forgotten about that work you did!

adamsachs: ah right OK - that makes sense! seems to totally fine to keep it in its own table as you've got it then and keep things open for now. 

to be clear, the config API was more about making config properties settable via API (to allow updates without an app restart), rather than the reverse - so it's there's no real concern here!

pattisdr:
@allisonking just tagging your review to get your thoughts as you're integrating with this branch - let me know what else you need from a POC perspective - again this branch mostly just returns purposes/special purposes, and very minimal vendor info at the moment.

pattisdr:
OK! let's resurrect this branch

pattisdr:
@adamsachs looking for a review on the initial POC particularly around model/application changes for serving and saving consent preferences for TCF components.  There's a similar change for returning and saving that consent was served.

Mainly, the work we did for Privacy Notices has been refactored to be shared with TCF components.  Just like we can opt into a CA Advertising Privacy Notice, we can similarly opt into a TCF purpose.  We're sharing the same tables (your suggestion, that ended up working well I think) - so work involved relaxing constraints on original tables too.

As far as the _contents_ of the overlay I would purely treat that as a POC. I am heavily refactoring that on a follow-up branch now that our dictionary fields are in place.

src/fides/api/models/privacy_preference.py
```diff
@@ -146,6 +197,18 @@ def provided_identity_id(cls) -> Column:
 
     user_geography = Column(String, index=True)
 
+    vendor = Column(
+        String, index=True
+    )  # When saving privacy preferences with respect to a vendor directly
```
pattisdr, line 202: The name and meaning of this field is something I want to clarify earlier rather than later. I've been assuming the whole time that preferences would be saved against something like sendgrid's vendor id instead of a system fides key, given that multiple systems can point to the same vendor.

https://ethyca.slack.com/archives/C05E74TR0BB/p1692371205716419?thread_ts=1692369966.244759&cid=C05E74TR0BB

But it sounds like system fides key is on the table instead, I just want to make sure we're all on the same page here - this matters a lot for how we aggregate the results and causes potential confusing edge cases.

adamsachs (commented):
haven't gotten super far through this, but i need to temporarily switch off it and i'm not sure i'll get back to it before the weekend, so leaving some very preliminary comments here.

i think i started to get a general sense of where it's going, and i like it! but i'll need to circle back as soon as i can to dig deeper

src/fides/api/schemas/tcf.py
```diff
@@ -0,0 +1,140 @@
...
+    name: str
+
+
+class TCFPurposeRecord(MappedPurpose, TCFSavedandServedDetails):
+    """Schema for a TCF Purpose or a Special Purpose: returned in the TCF Overlay Experience"""
+
+    vendors: List[EmbeddedVendor] = []  # Vendors that use this purpose
+
+
+class EmbeddedLineItem(FidesSchema):
```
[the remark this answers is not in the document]
pattisdr: In this POC branch, I'd kept it minimal! The last time I looked at the designs, it seemed as if there was just the name here, but I'll add the description in the follow-up if needed. You're right that this should come from the BE!

I'm going to start keeping a running list of items to make sure I address in the follow-up: https://github.com/ethyca/fides/issues/3876

src/fides/api/schemas/tcf.py
```diff
@@ -0,0 +1,140 @@
...
+    special_features: List[EmbeddedLineItem] = []
+
+
+class TCFFeatureRecord(TCFSavedandServedDetails):
+    """Schema for a TCF Feature or a special feature: returned in the TCF Overlay Experience
+    TODO: TCF Flesh out TCFFeatureRecord
+    """
+
+    id: int
+    name: Optional[str]
```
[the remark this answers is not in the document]
pattisdr: yes! we're on the same page.  In the follow-up PR I'm working on now beyond the POC, [I'm starting to point at the Feature I've added to fideslang.](https://github.com/ethyca/fideslang/pull/153/files#diff-4ff722ddf999b44a3d96cf0498b091eed453650d8f02dcfba4f757474ce7f5c7R34)  This branch just has a shell for features at the moment.

src/fides/api/models/privacy_preference.py
pattisdr: In the followup, we'll need to collapse vendor information together in the overlay, but also support saving preferences against systems that don't have a specific vendor. Adam suggested an additional column, so we have one for vendor id and one for system id.  

For retrieving the last saved preference (in a separate table for easier retrieval across time, tcf versions, etc, multiple identities of the user, etc.) my early thoughts are that we should save against the vendor id *only* in that table if it exists, otherwise, the system id. This might solve the unique constraint issue.

adamsachs (commented):
left some comments inline on relatively minor points, but high level this is looking great!!! 👍 

awesome work here @pattisdr, there is a _lot_ of complexity that you've unpacked. i especially appreciate the attention to detail in the PR overview to help summarize and keep track of all the changes.

excited to see where this progresses as we flesh out the MVP!

src/fides/api/models/privacy_preference.py
```diff
@@ -69,6 +103,11 @@ class ConsentReportingMixin:
             padding="pkcs5",
         ),
     )
+
+    feature = Column(
```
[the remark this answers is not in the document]
pattisdr: > understand how these columns correspond to some of the mappings and enums that drive behavior closer to the API level of the stack.

yes that is a great point.  In general I need to figure out how to strike a balance between being concise but understandable.  Originally, in the first drafts where I didn't know about "special purposes or special features", I was much more explicit and spelled out every variable. But then as the number of variables grew then it became too easy to miss a location that needed to change when adding a new TCF component so I started adding these mappings and being more dynamic.  And now it's hard to follow!

src/fides/api/api/v1/endpoints/privacy_experience_endpoints.py
```diff
@@ -165,38 +187,83 @@ def privacy_experience_list(
...
+
+    Returns whether content exists on the experience.
+    """
+    # Reset any temporary cached items just in case
+    privacy_experience.privacy_notices = []
+    for component in TCF_COMPONENT_MAPPING:
+        setattr(privacy_experience, component, [])
+
+    # Temporarily add TCF Contents to the Experience object if applicable
+    tcf_contents: TCFExperienceContents = privacy_experience.get_related_tcf_contents(
```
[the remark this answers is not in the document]
pattisdr: That's a good idea, thanks!

src/fides/api/models/privacy_preference.py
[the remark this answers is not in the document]
pattisdr: I can group these together, that makes more sense -

pattisdr (commented):
thanks very much for your comments @adamsachs  👍

src/fides/api/models/privacy_preference.py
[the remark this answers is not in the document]
adamsachs: oh i definitely like the dynamic mappings - didn't mean to suggest those make it hard to follow in and of themselves. in some ways i think it helps draw some boundaries around what is/isn't TCF-related, and it "centralizes" those concepts. 

but yeah, maybe as the work progresses, there are some steps to take to make sure that centralization stretches across the different references in the codebase, e.g. grouping the column definitions together here, etc. 👍

src/fides/api/api/v1/endpoints/privacy_experience_endpoints.py
```diff
@@ -165,38 +187,83 @@ def privacy_experience_list(
...
-                for notice in privacy_notices
-            ]
-        # Temporarily save privacy notices on the privacy experience object
-        privacy_experience.privacy_notices = privacy_notices
-        # Temporarily save "show_banner" on the privacy experience object
-        privacy_experience.show_banner = privacy_experience.get_should_show_banner(
-            db, show_disabled
-        )
-        if not (has_notices and not privacy_notices):
+        if not (has_notices and not content_exists):
```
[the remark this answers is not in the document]
adamsachs: ahh ok, thanks for spelling that out. understand you want to keep the API stable, so i agree that updating the variable name could be a good step.

i think the biggest impact though would be just a quick explainer of what the variable in the function docstring! and then it's also self-documenting in the generated open API UI 😄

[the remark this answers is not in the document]
pattisdr: docstring good point 👍

pattisdr:
@adamsachs hoping you could review TCF backend this week! I've added vendors, features/special features, and did a sweep to see what remaining items needed to be addressed regarding code sharing with notices/experiences and the new tcf components.  I'll annotate the PR with areas to focus on tomorrow 9/5

src/fides/api/models/privacy_experience.py
```diff
@@ -506,17 +550,11 @@ def upsert_privacy_experiences_after_config_update(
...
-
-        existing_experience: Optional[PrivacyExperience] = (
-            overlay_experience
-            if experience_config.component == ComponentType.overlay
-            else privacy_center_experience
+        existing_experience: Optional[
+            PrivacyExperience
+        ] = PrivacyExperience.get_experience_by_region_and_component(
+            db=db, region=region, component=experience_config.component  # type: ignore[arg-type]
         )
```
pattisdr, line 557: This method updates existing Experiences when ExperienceConfigs (the copy) have new regions added to them.  It assumed that only an overlay and a privacy center experience existed, but we've added a new type.

src/fides/api/util/consent_util.py
```diff
@@ -57,12 +59,17 @@ def filter_privacy_preferences_for_propagation(
...
+    Only applies to preferences saved for privacy notices here, not against individual TCF components.
+    """
 
     propagatable_preferences: List[PrivacyPreferenceHistory] = [
         pref
         for pref in privacy_preferences
-        if pref.privacy_notice_history.enforcement_level == EnforcementLevel.system_wide
+        if pref.privacy_notice_history
+        and pref.privacy_notice_history.enforcement_level
+        == EnforcementLevel.system_wide
```
pattisdr, line 72: Putting some extra checks here just to be comprehensive.  Currently, consent is only propagated to backend systems for notices, not TCF components.  Any privacy preference that hits here should have a privacy notice history attribute.  However, now privacy preferences are not required to have privacy notice histories, so here I'm verifying that one exists. 

Further work would need to be done if/when we get to a point where saving TCF preferences are propagated via API to third party systems.

src/fides/api/schemas/privacy_preference.py
```diff
@@ -27,16 +76,47 @@ class ConsentOptionCreate(FidesSchema):
...
     browser_identity: Identity
     code: Optional[SafeStr]
-    preferences: conlist(ConsentOptionCreate, max_items=50)  # type: ignore
+    preferences: conlist(ConsentOptionCreate, max_items=200) = []  # type: ignore
+    purpose_preferences: conlist(TCFPurposeSave, max_items=200) = []  # type: ignore
+    special_purpose_preferences: conlist(TCFSpecialPurposeSave, max_items=200) = []  # type: ignore
+    vendor_preferences: conlist(TCFVendorSave, max_items=200) = []  # type: ignore
+    feature_preferences: conlist(TCFFeatureSave, max_items=200) = []  # type: ignore
+    special_feature_preferences: conlist(TCFSpecialFeatureSave, max_items=200) = []  # type: ignore
+    system_preferences: conlist(TCFVendorSave, max_items=200) = []  # type: ignore
```
pattisdr, line 95: Updated request body for saving preferences.  You can send in a list of preferences for various privacy notices per usual, but now you can send in preferences against multiple TCF components simultaneously.

src/fides/api/schemas/privacy_experience.py
```diff
@@ -188,6 +189,31 @@ class PrivacyExperienceResponse(PrivacyExperienceWithId):
...
+    tcf_features: Optional[List[TCFFeatureRecord]] = Field(
+        description="For TCF Experiences, the TCF Features that appear on your Systems"
+    )
+    tcf_special_features: Optional[List[TCFFeatureRecord]] = Field(
+        description="For TCF Experiences, the TCF Special Features that appear on your Systems"
+    )
+    tcf_systems: Optional[List[TCFVendorRecord]] = Field(
+        description="For TCF Experiences, Systems with TCF components that do not have an official vendor id "
+        "(identified by system id)"
+    )
```
pattisdr, line 216: Updated schema for fetching privacy experiences.  Previously, it just returned the notices that were relevant for that experience, but now we can optionally return TCF components that are relevant at runtime as well.

pattisdr:
TODO:

- Revert making consent settings public https://github.com/ethyca/fides/issues/3957

src/fides/api/api/v1/endpoints/consent_settings_endpoints.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+@router.get(
+    urls.CONSENT_SETTINGS,
+    response_model=ConsentSettingsResponseSchema,
+)
+def get_consent_settings(*, db: Session = Depends(deps.get_db)) -> ConsentSettings:
+    """Public endpoint that returns organization-wide consent settings."""
+    logger.info("Getting organization-wide consent settings")
+    return ConsentSettings.get_or_create_with_defaults(db)
```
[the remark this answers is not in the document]
pattisdr: Ah! Looks like I can revert this, I will update to make it private https://github.com/ethyca/fides/issues/3957

src/fides/api/util/tcf_util.py
```diff
@@ -0,0 +1,637 @@
...
+        purpose_record.legal_bases = legal_bases
+    else:
+        if legal_basis_for_processing not in purpose_record.legal_bases:
+            purpose_record.legal_bases.extend(legal_bases)
+
+    purpose_record.legal_bases.sort()
+
+
+@lru_cache()
+def get_tcf_contents(
```
adamsachs, line 400: TODO for me in review -- high-level, this looks great. but there's a lot to dig into here and i haven't had the time to go thru in the level of detail i'd want to.

adamsachs (commented):
@pattisdr high level this is looking great! nothing stands out to me as particularly problematic.

i still haven't been able to get down to the level of detail i'd like to in a few areas, namely around the TCF overlay rendering logic. i do think that's worth a thorough review, i just didn't get the time for it today. i'll look to circle back to that on monday, but in the mean time wanted to leave you with some of the comments i had in the meantime!

src/fides/api/util/tcf_util.py
```diff
@@ -0,0 +1,637 @@
...
+                )
+            )
+
+            _embed_purpose_or_feature_under_system(
+                embedded_tcf_record=embedded_purpose_or_feature_record,
+                system_section=getattr(
+                    system_map[system_identifier], tcf_component_name
+                ),
+                legal_basis_for_processing=privacy_declaration_row.legal_basis_for_processing,
+            )
```
adamsachs, line 362: nit: if i'm following correctly, this is a bit redundant - the `privacy_declaration_row.legal_basis_for_processing` has already been added to the `embedded_purpose_or_feature_record` object when it was instantiated within ` _clone_top_level_record_then_add_legal_bases`. 

when reading over ` _clone_top_level_record_then_add_legal_bases`, it struck me as a bit of a "random" assortment of operations. i think you could probably simplify things by removing that function and slightly refactoring things here:

```suggestion
            # copy top level record to preserve its initial state to use for embedded record
            embedded_tcf_record: TCFPurposeOrFeature = top_level_tcf_record.copy()
            # Override legal_bases on the embedded purpose or special purpose record
            _extend_legal_bases(embedded_tcf_record, legal_basis_for_processing, replace=True)
            
            _embed_purpose_or_feature_under_system(
                embedded_tcf_record=embedded_purpose_or_feature_record,
                system_section=getattr(
                    system_map[system_identifier], tcf_component_name
                ),
            )
            
            # Append to the existing legal_bases on the top-level record if applicable for purpose and special purpose sections.
            # The top-level record is updated in-place
            _extend_legal_bases(top_level_tcf_record, legal_basis_for_processing, replace=False)
```

and then within `_embed_purpose_or_feature_under_system` just decide whether to add the `legal_basis_for_processing` that's on the provided embedded purpose to the _existing_ embedded purpose (if it exists), or to add it to the "new" embedded purpose (i.e. the provided one) to the system -- basically the same as it does currently.

does that makes sense? if not, happy to provide a commit that makes the change, maybe that'd be simpler. if it makes sense and you think i'm mistaken, then sorry for the rant 😄 but i do think this may deserve a slight refactor to make it easier to follow, in any case!

tests/ops/util/test_tcf_util.py
```diff
@@ -0,0 +1,578 @@
...
+from fideslang.models import LegalBasisForProcessingEnum
+
+from fides.api.models.sql_models import PrivacyDeclaration
+from fides.api.schemas.tcf import EmbeddedVendor
+from fides.api.util.tcf_util import get_tcf_contents
+
+
+class TestTCFContents:
+    def test_load_tcf_data_uses_no_applicable_systems(self, db):
+        get_tcf_contents.cache_clear()
```
adamsachs, line 12: nit: could we make this operation an auto-use class fixture so that it's automatically run before every test in the class? just to make this a bit more robust going forward! i think that should be simple, if it's not then just ignore me.

tests/ops/util/test_tcf_util.py
adamsachs, line 1: great test coverage here! very thorough 👍 

maybe we can think of a way to parametrize a bit in a future iteration, though it's certainly not a straightforward refactor - we'd need to take some care to not add more complexity than we save 😅

adamsachs (commented):
OK! i took a deeper dive into the TCF overlay population logic and it looks great 👍 really nice job taking a lot of care to get that right, there's a lot packed in there.

i just had some minor comments that may help to improve readability, though by and large this is quite readable for the complexity involved.

i'd also like to get your thoughts on the broader comment i have about a future refactor, it'd be good to align on how we see this shaping up for future iterations!

src/fides/api/schemas/consent_settings.py
```diff
@@ -0,0 +1,16 @@
+from datetime import datetime
+
+from fides.api.schemas.base_class import FidesSchema
+
+
+class ConsentSettingsRequestSchema(FidesSchema):
+    """Response schema for consent settings update"""
```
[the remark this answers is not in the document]
pattisdr: ignoring - as I'm making scopes required again to access this endpoint!

src/fides/api/api/v1/endpoints/privacy_experience_endpoints.py
```diff
@@ -114,15 +134,26 @@ def privacy_experience_list(
...
+    :param has_notices: Return if the Experience has content
+    :param has_config: If True, returns Experiences with copy. If False, returns just Experiences without copy.
+    :param fides_user_device_id: Supplement the response with current saved preferences of the given user
+    :param systems_applicable: Only return embedded Notices associated with systems.
+    :param request:
+    :param response:
+    :return:
     """
     logger.info("Finding all Privacy Experiences with pagination params '{}'", params)
+    content_required: bool = bool(has_notices)  # Renaming confusing query param
```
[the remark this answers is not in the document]
pattisdr: As a first step, I've added an alias, which makes it read more clearly on the backend.

I don't think this is worth changing unless the level of effort is minimal. @allisonking what would be the effort required to update the frontend to use a new query param called `content_required` instead of `has_notices`?

src/fides/api/api/v1/endpoints/privacy_experience_endpoints.py
```diff
@@ -152,8 +183,15 @@ def privacy_experience_list(
         )
 
     if component is not None:
+        # If searching for "overlay" - return both overlay types: the regular overlay and the TCF Overlay,
```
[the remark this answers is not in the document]
pattisdr: But we can return both here, that was why I added the code comment!  The frontend requested that the backend be a little more generous about what it meant when we filtered on "component=overlay".  Because the frontend doesn't know whether to filter on `overlay` or `tcf_overlay`, it just knows where it needs some sort of overlay.    

So here, I am basically excluding the privacy center experience from being returned.  And then later, it is the presence of the `region` filter that further restricts whether a TCF or regular overlay is returned.

src/fides/api/api/v1/endpoints/privacy_preference_endpoints.py
```diff
@@ -346,42 +343,106 @@ def save_privacy_preferences_with_verified_identity(
...
+            },
+            check_name=False,
+        )
+        return current_preference
+
+    # Save TCF preferences individually in the database, with respect to purpose, special purpose, vendor,
+    # feature, special feature, and/or system id.
+    # Currently, we don't attempt to propagate these preferences to third party systems.
+    for tcf_preference_field, field_name in TCF_PREFERENCES_FIELD_MAPPING.items():
+        for preference in getattr(request_data, tcf_preference_field):
```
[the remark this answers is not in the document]
pattisdr: Correct, `getattr(request_data, tcf_preference_field)` returns a list of preferences for a given TCF component, and that list may be empty, in which case nothing is saved. I'll try to add a giant List[Union..] for type hints and improve docstrings.

src/fides/api/alembic/migrations/versions/66df7d9b8103_add_tcf_columns.py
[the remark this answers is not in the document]
pattisdr: I don't see a lot of value in DRYing up an automatically-generated migration! How about a compromise of leaving code comments about what each section is doing.   I did have code comments originally, but I've had to regenerate the migration 5 or 6 times as more fields were added/updated in the model as I learned about new requirements so dropped that.

src/fides/api/util/tcf_util.py
```diff
@@ -0,0 +1,637 @@
...
+    # Collect special features and update system map
+    (
+        special_feature_map,
+        updated_system_map,
+    ) = build_purpose_or_feature_section_and_update_system_map(
+        [feature.name for feature in GVL_SPECIAL_FEATURES.values()],
+        tcf_component_name="special_features",
+        system_map=updated_system_map,
+        matching_privacy_declaration_query=matching_privacy_declarations,
+    )
```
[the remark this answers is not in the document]
pattisdr: Sure, this could be moved to fideslang in some form eventually.  One thing I want to highlight that is happening here though is that each Feature and Vendor, for example, are not being created independently.  We might create two Features and update the corresponding Vendor to nest those features beneath the vendor.  Likewise the individual Features have a vendors section where the vendor id is added.  The TCF components affect each other's construction.  

Not to say there isn't value in a Privacy Declaration -> Purpose/Feature/Vendor mapping, just that those constructs can't live in isolation. Other PrivacyDeclarations will also update the constructs created by a previous PrivacyDeclaration.

src/fides/api/alembic/migrations/versions/66df7d9b8103_add_tcf_columns.py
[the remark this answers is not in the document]
adamsachs: ah ok ignore me then, I didn't quite realize this was totally autogen'd! it's worth the convenience. sorry for that digression!!

src/fides/api/api/v1/endpoints/privacy_preference_endpoints.py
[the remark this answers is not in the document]
adamsachs: ok yes just some comments/additional notes in the docstring is plenty!

src/fides/api/util/tcf_util.py
pattisdr: I agree with you this could be more readable for sure.  I hard time layering in the legal basis for processing bit in a clean way:

The main motivation for passing in the legal basis here when I embed the purpose under a vendor, is that the purpose might already exist under that vendor.   If the purpose is already there, then I would consolidate the legal_bases across both purposes. It's a bit contrived but seems possible, especially given that users can set up their own systems, it is not just pristine GVL systems that appear.

tests/ops/util/test_tcf_util.py
pattisdr: That's definitely fair - these tests could be cleaned up some! I kind of added to them over time as understanding and capabilities for TCF expanded.   I have an idea for a top-level helper that can clean it up a little.

to be honest, I don't love parametrization, unless you're just asserting a few, sparse things. while application code should be DRY, I don't mind when tests are a little more repetitive but clear and readable. I find a giant parametrized test really hard to parse.

tests/ops/util/test_tcf_util.py
pattisdr: Sure I can look into that.  What do you think about the `lru_cache` I'm using here to begin with though? Any concerns?

requirements.txt
```diff
@@ -13,7 +13,7 @@ expandvars==0.9.0
 fastapi[all]==0.89.1
 fastapi-caching[redis]==0.3.0
 fastapi-pagination[sqlalchemy]==0.11.4
-fideslang==1.4.5
+fideslang==1.4.6a0
```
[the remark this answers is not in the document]
pattisdr: I am back to pinning to a specific fideslang commit here that is based off of 1.4.6, instead of 1.4.5. I am hoping tomorrow we get the fideslang 2.0 work in, and I can just move forward on getting the proper release instead of bothering to make an alpha tag. we are close here!

src/fides/api/api/v1/endpoints/privacy_experience_endpoints.py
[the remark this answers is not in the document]
allisonking: it's pretty minimal! we'd change it in two places:

fides-js
https://github.com/ethyca/fides/blob/094325a0f7b66c7f8d4207384ba3a60e55920883/clients/fides-js/src/services/fides/api.ts#L39

privacy center
https://github.com/ethyca/fides/blob/094325a0f7b66c7f8d4207384ba3a60e55920883/clients/privacy-center/features/consent/consent.slice.ts#L70

I can do it as part of `aking/3809/tcf-modal` (the main FE feature branch) if you'd like!

tests/ops/util/test_tcf_util.py
adamsachs: sounds good, and makes sense!

tests/ops/util/test_tcf_util.py
adamsachs: oh thanks for bringing my attention back to that! 👍 

on the face of it, it feels like a good idea, to repeating those many operations involved in computing the overlay. but on closer look i'm questioning whether the cache would be invalidated appropriately? my understanding of the LRU cache is that its caching based on the hash of the function arguments. in this case, the lone argument is a DB `Session` - would its hash be updated when the relevant DB state (in this case, datamap contents) changes? because i do think we'd want to rerun the function on basically any update/addition/removal of a `System` in our datamap, right?

maybe i'm thinking about this wrong! don't have a ton of experience with LRU caches in python. i suppose one thing that would settle my concerns indefinitely, and would be nice to have moving forward, is an explicit test that the cache is _not_ getting in the way of providing up-to-date, accurate results. what do you think?

src/fides/api/util/tcf_util.py
adamsachs: right i think i'm following that, i don't think it's wrong that you're account for that contrived use case. i realize now though i messed up the wording in my comment a bit and i wasn't very clear 🤦 

i'm getting into the weeds and it's probably not worth much more of our time, since this is a stylistic concern, not functional. your most recent changes are close to what i was suggesting, but could we remove the `legal_basis_for_processing` arg from the ` _embed_purpose_or_feature_under_system` function, and instead within that function just read the legal basis from the provided `embedded_tcf_record`? or alternatively we could remove the 
```
 _extend_legal_bases(
                embedded_tcf_record, legal_basis_for_processing, replace=True
            )
```
that's done before the `_embed_purpose_or_feature_under_system` call and handle extending the legal bases _within_ the `_embed_purpose_or_feature_under_system` function. 

i'll make a code suggestion on your most recent revision, if it doesn't make sense then i think we can just leave it. i find the current code a _bit_ confusing but i don't think this will have any functional impact and it's a nit in any case.

src/fides/api/util/tcf_util.py
```diff
@@ -0,0 +1,623 @@
...
+            )
+
+            # Embed the purpose/feature under the system if it doesn't exist, and/or consolidate legal bases
+            _embed_purpose_or_feature_under_system(
+                embedded_tcf_record=embedded_tcf_record,
+                system_section=getattr(
+                    system_map[system_identifier], tcf_component_name
+                ),
+                legal_basis_for_processing=legal_basis_for_processing,
+            )
```
adamsachs, line 342: per our previous discussion, could this look something like 
```suggestion

            # Embed the purpose/feature under the system if it doesn't exist, and/or consolidate legal bases
            _embed_purpose_or_feature_under_system(
                embedded_tcf_record=top_level_tcf_record.copy(),
                system_section=getattr(
                    system_map[system_identifier], tcf_component_name
                ),
                legal_basis_for_processing=legal_basis_for_processing,
            )
```

and then see corresponding suggestion on ` _embed_purpose_or_feature_under_system` to see how i'd imagine that being tweaked...

src/fides/api/util/tcf_util.py
```diff
@@ -0,0 +1,623 @@
...
+    if existing_embedded_purpose_or_feature_record:
+        # Update legal_bases on existing embedded purpose or feature record beneath system if applicable
+        _extend_legal_bases(
+            existing_embedded_purpose_or_feature_record,
+            legal_basis_for_processing,
+            replace=False,
+        )
+        return
+    # Nest new cloned TCF purpose or feature record beneath system otherwise
+    system_section.append(embedded_tcf_record)  # type: ignore[arg-type]
```
adamsachs, line 234: potential update: 
```suggestion
def _embed_purpose_or_feature_under_system(
    embedded_tcf_record: TCFPurposeOrFeature,
    system_section: Union[List[TCFPurposeRecord], List[TCFFeatureRecord]],
    legal_basis_for_processing: Optional[str],
) -> None:
    """
    Embed a second-level TCF purpose/feature under the systems section.

    The systems section is updated in-place.
    """
    existing_embedded_purpose_or_feature_record: Optional[TCFPurposeOrFeature] = next(
        (
            tcf_sub_record
            for tcf_sub_record in system_section
            if tcf_sub_record.id == embedded_tcf_record.id
        ),
        None,
    )

    if existing_embedded_purpose_or_feature_record:
        # Update legal_bases on existing embedded purpose or feature record beneath system if applicable
        _extend_legal_bases(
            existing_embedded_purpose_or_feature_record,
            legal_basis_for_processing,
            replace=False,
        )
        return
    # Nest new cloned TCF purpose or feature record beneath system otherwise
    # Ensuring it has the appropriate legal basis
    _extend_legal_bases(
                embedded_tcf_record, legal_basis_for_processing, replace=True
      )
    system_section.append(embedded_tcf_record)  # type: ignore[arg-type]
```

adamsachs, line 234: potential update: 
```suggestion
    # Nest new cloned TCF purpose or feature record beneath system otherwise
    # Ensuring it has the appropriate legal basis
    _extend_legal_bases(
                embedded_tcf_record, legal_basis_for_processing, replace=True
      )
    system_section.append(embedded_tcf_record)  # type: ignore[arg-type]
```

tests/ops/util/test_tcf_util.py
pattisdr: I've been mulling this over,:I'm thinking I should just remove the caching altogether.  A `session` argument is not a great one to determine if the cache should be invalidated.  It's hard to unit test because we're literally passing around the same session within the test. The session is the same, so the cache isn't invalidated, hence the `cache_clear()`.

However, testing locally against a running instance, subsequent requests seem to recalculate system updates just fine, and I think it has nothing to do with the system changing, and something about the session object changing. I imagine it's recalculating it every time, so having little practical effect.

adamsachs: yeah that makes a lot of sense to me, feels like at this point it'd only add confusion, and we can target any optimizations as needed in a future iteration, via an LRU cache or other mechanism 👍

src/fides/api/util/tcf_util.py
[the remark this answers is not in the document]
pattisdr: I agree with you happy to help figure out how to consolidate more of this in fideslang as a next step!

src/fides/api/util/tcf_util.py
pattisdr: oh sure I think I follow what you mean now

tests/ops/util/test_tcf_util.py
pattisdr: lru cache removed! This was something I'd added very early on and hadn't thought about it since, so I'm glad we revisited before merge. thanks for the comments on this test that reminded me!

src/fides/api/util/tcf_util.py
pattisdr: ah ok thanks

src/fides/api/api/v1/endpoints/privacy_experience_endpoints.py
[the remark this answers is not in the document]
pattisdr: Thanks for looking into this Allison - at this point I say we merge this, and can add a small follow-up this sprint

pattisdr:
great thanks for your comments @adamsachs plan to merge tomorrow after 2.20 is released

requirements.txt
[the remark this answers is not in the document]
pattisdr: now pointed to the correct fideslang version!

pattisdr:
Getting up to date with main again, error with migration heads -

PR #3839, opened 2023-07-20
Fix Fideslang Typing

ThomasLaPiana:
Closes #N/A

### Description Of Changes

I inadvertently fixed typing over in `fideslang`, which means we now have a bunch of things to fix here!

This uncovered a lot of legitimate issues in our type annotations.

### Code Changes

* [x] get `mypy` passing
* [x] please see code comments for more specifics

### Steps to Confirm

### Pre-Merge Checklist


src/fides/api/db/system.py
```diff
@@ -142,11 +141,17 @@ async def upsert_cookies(
...
+    """
+    Upsert cookies for the given privacy declaration: retrieve cookies
+    by name/system/privacy declaration.
+
     Remove any existing cookies that aren't specified here.
     """
-    cookie_list: List[CookieSchema] = cookies or []
-    for cookie_data in cookie_list:
+
+    if not cookies:
```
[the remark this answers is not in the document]
pattisdr: @ThomasLaPiana, you've changed the behavior of this function: previously, if cookies were specified as an empty list on an existing declaration, I would remove those cookies from the declaration altogether.  Because you exit early now, there's no opportunity to remove cookies. 

The request and response declarations are not matching because the request is saying "cookies should be an empty list" on this existing declaration and the response is showing the cookies are still there (because of the change to return if not cookies)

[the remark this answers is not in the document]
ThomasLaPiana: @pattisdr exactly what I was looking for, thank you! I'll make that fix (er...revert my breaking change)

src/fides/api/db/system.py
```diff
@@ -137,7 +137,7 @@ async def upsert_privacy_declarations(
 
 async def upsert_cookies(
     async_session: AsyncSession,
-    cookies: Optional[List[Dict]],  # CookieSchema
+    cookies: List[Dict],  # CookieSchema
```
pattisdr, line 140: There might be some places in the codebase where we're accessing cookies and if they're not there, returning None as a default - can you check that, and make it an empty list if needed?

ThomasLaPiana: If this is supposed to be a cookie schema, why isn't it typed as such? I want to make sure I don't undo anything that has a specific reason

src/fides/api/db/system.py
```diff
@@ -138,15 +137,18 @@ async def upsert_privacy_declarations(
...
     system: System,
 ) -> None:
-    """Upsert cookies for the given privacy declaration: retrieve cookies by name/system/privacy declaration
+    """
+    Upsert cookies for the given privacy declaration: retrieve cookies
+    by name/system/privacy declaration.
+
     Remove any existing cookies that aren't specified here.
     """
-    cookie_list: List[CookieSchema] = cookies or []
```
ThomasLaPiana, line 148: I think this was incorrect? It lists it as `CookieSchema` but further down I see dict syntax being used

src/fides/api/db/system.py
pattisdr: > There might be some places in the codebase where we're accessing cookies and if they're not there, returning None as a default - can you check that, and make it an empty list if needed?

Is just related to how below you've stopped doing "cookies or []" so I just want to make sure we always have something to iterate through.

ThomasLaPiana: understood! For the sake of type cleanliness, I'll do an explicit conversion here

ThomasLaPiana:
update here: looks like everything is good except for the unsafe checks. I suspect at least the `ctl` ones are actually related here but these are much more difficult to troubleshoot. Will start poking at them locally

ThomasLaPiana:
@galvana can you give me confirmation that the failing tests here are expected? I looked at `main` and they look like the same failures but I want to be sure!

src/fides/api/schemas/dataset.py
```diff
@@ -36,17 +38,17 @@ def valid_data_categories(
...
-        fields: Optional[List["FieldDataCategoryValidation"]]
+        fields: Optional[List["FieldDataCategoryValidation"]]  # type: ignore[assignment]
 
     FieldDataCategoryValidation.update_forward_refs()
 
     class CollectionDataCategoryValidation(
         DatasetCollection, DataCategoryValidationMixin
     ):
-        fields: List[FieldDataCategoryValidation] = []
+        fields: Sequence[FieldDataCategoryValidation] = []  # type: ignore[assignment]
```
pattisdr, line 48: Where is this coming from?

src/fides/api/oauth/system_manager_oauth_util.py
```diff
@@ -88,20 +88,22 @@ async def verify_oauth_client_for_system_from_request_body(
...
+    return system
 
 
 async def verify_oauth_client_for_system_from_fides_key(
     security_scopes: SecurityScopes,
     authorization: str = Security(oauth2_scheme),
     db: Session = Depends(get_db),
     system_auth_data: SystemAuthContainer = Depends(_get_system_from_fides_key),
-) -> str:
+) -> Union[str, System]:
```
pattisdr, line 106: I believe this method always returns a string even though the underlying method is capable of returning multiple types, same for the method above which I believe always returns a system.

This type change makes it harder to tell what is actually returned here, I would prefer an ignore in this case.

(Granted, my implementation here was confusing.  When I did something similar in plus, it was implemented in a cleaner way).

pattisdr (commented):
great @ThomasLaPiana thanks for the work to allow us to return on automated mypy checks again! so useful

adamsachs (approved):
thank you for not losing sight of this and staying diligent with maintenance! nothing stands out to me as particularly problematic. i do think @pattisdr had some good questions that may be nice to get answers to before we merge.

🎊 

EDIT -- just wanted to note a reminder that we need to follow up with `fidesplus` - both to make any needed changes based on dependencies on what we've changed here (not sure if there's anything applicable), and just generally getting `mypy` fixed there! i'll see if i can help push that forward once this is merged 👍

pyproject.toml
```diff
@@ -43,7 +43,6 @@ module = [
   "dask.*",
   "deepdiff.*",
   "defusedxml.ElementTree.*",
-  "fideslang.*",
```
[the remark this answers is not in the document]
ThomasLaPiana: this section tells `mypy` to ignore missing type imports from imported libraries. Everything from fideslang is typed so this is redundant

at least that's my understanding

tests/lib/test_system_oauth_util.py
```diff
@@ -56,7 +56,7 @@ async def test_owner_role_can_always_update_system(self, owner_user, db, system)
             ),
         )
 
-        assert response == system.fides_key
+        assert True
```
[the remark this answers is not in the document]
ThomasLaPiana: not sure, re-added the original assertion and it is passing

src/fides/api/schemas/dataset.py
ThomasLaPiana: This is required due to Lists in Python being invariant, as a result of their mutability

src/fides/api/oauth/system_manager_oauth_util.py
ThomasLaPiana: Ok, I'll take a look here. If it always returns a string, then it should be typed as such

src/fides/api/util/endpoint_utils.py
```diff
@@ -69,6 +85,14 @@ async def forbid_if_editing_is_default(
     """
     if sql_model in models_with_default_field:
         resource = await get_resource(sql_model, fides_key, async_session)
+
+        assert isinstance(
+            resource, (SQLDataCategory, SQLDataQualifier, SQLDataSubject, SQLDataUse)
```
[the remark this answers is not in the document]
ThomasLaPiana: @adamsachs I think I have just the thing for this 😉

PR #3844, opened 2023-07-20
Make taxonomy items able to be enabled or disabled

### LINKED ISSUE #3710: FE: add toggle to disable taxonomy elements to Taxonomy Manager
We need to add a toggle that enables our users to mark a particular element in the taxonomy as `disabled` so that these elements don't show up when labeling systems.

Image

jpople:
Closes #3710 

Adds a toggle to the hover menu in the taxonomy view for taxonomy items which enables or disables them after displaying a modal.  Disabled items will not be shown in system form dropdowns [not implemented yet].

### Steps to Confirm

- view taxonomy
- click "disable" on some taxonomy items
- verify that they no longer show up in system, data use, or privacy notice form dropdowns for that item type

### Pre-Merge Checklist

* [x] Update `CHANGELOG.md`


clients/admin-ui/src/features/taxonomy/hooks.tsx
```diff
@@ -163,6 +167,24 @@ export const useDataCategory = (): TaxonomyHookData<DataCategory> => {
...
+    const payload = {
+      ...entity,
+      active: isDisabled,
+    };
+
+    const result = updateDataCategoryMutationTrigger(payload);
+
+    if (customFields.isEnabled) {
+      await customFields.upsertCustomFields(payload);
+    }
```
allisonking, line 183: do we need this here? does toggling the taxonomy affect the custom fields?

clients/admin-ui/src/features/taxonomy/hooks.tsx
```diff
@@ -293,6 +316,28 @@ export const useDataUse = (): TaxonomyHookData<DataUse> => {
     };
   };
 
+  const handleToggleEnabled = async (
```
allisonking, line 319: I think you might be able to refactor this a bit such that there's less duplicate code. it looks like the only difference between these is that they each need a different `update...MutationTrigger`, so perhaps the update function can be passed in as an argument to a more general function, something like `handleEdit`. 

or maybe TS will be cranky about this sort of thing 🙃 worth trying though to cut down on duplicate code!

clients/admin-ui/src/types/api/models/DataCategory.ts
```diff
@@ -28,4 +28,8 @@ export type DataCategory = {
    * Denotes whether the resource is part of the default taxonomy or not.
    */
   is_default?: boolean;
+  /**
```
allisonking, line 31: we'll want to make sure that this is generated from the `npm run openapi:generate` command. you'd have to run it against fidesplus (followed by an `npm run format` to apply prettier to all the new files)

allisonking (commented):
looking good so far!

clients/admin-ui/src/features/taxonomy/hooks.tsx
jpople: I tried messing with it a bit but I don't think I can get it to work without giving the method an `any` type (and even then not easily), because its actual type is something insane.  If you'd like we could take a look at it together sometime tomorrow?  This was also templated after the existing create, edit, and delete functions which are all duplicated as well, so if there is a possible refactor we could do it for those too.

allisonking: ah got it, I vaguely recalled this being a tricky thing so I probably backed out and just copy/pasta'd 😄 don't worry about it then! I might tinker with it at some point but it's not a big deal 👍

clients/admin-ui/src/features/system/privacy-declarations/PrivacyDeclarationStep.tsx
```diff
@@ -15,6 +15,23 @@ const PrivacyDeclarationStep = ({ system }: Props) => {
     includeDatasets: true,
   });
 
+  const allEnabledDataCategories = dataProps.allDataCategories.filter(
```
allisonking, line 18: I think it might make more sense to do this filtering in `usePrivacyDeclarationData` instead:

https://github.com/ethyca/fides/blob/3710-enabled-toggle-in-taxonomy/clients/admin-ui/src/features/system/privacy-declarations/hooks.ts#L24

that's the "source" of the data, so I think it'd be best to transform it as early as we can

clients/admin-ui/src/features/taxonomy/hooks.tsx
```diff
@@ -293,6 +312,26 @@ export const useDataUse = (): TaxonomyHookData<DataUse> => {
...
+    entity: TaxonomyEntity,
+    isDisabled: boolean
+  ) => {
+    // const payload = needToCreateFunction(initialValues, isDisabled);
```
allisonking, line 319: can remove this now :) 
```suggestion
```

clients/admin-ui/src/features/taxonomy/hooks.tsx
```diff
@@ -448,6 +488,26 @@ export const useDataSubject = (): TaxonomyHookData<DataSubject> => {
...
```
allisonking, line 495:
```suggestion
```

clients/admin-ui/src/types/api/index.ts
```diff
@@ -148,6 +148,7 @@ export type { Identity } from "./models/Identity";
 export type { IdentityTypes } from "./models/IdentityTypes";
```
allisonking, line 148: nice, thank you for regenerating this!

clients/admin-ui/src/features/dataset/EditCollectionOrFieldForm.tsx
```diff
@@ -41,19 +41,21 @@ const EditCollectionOrFieldForm = ({ values, onSubmit, dataType }: Props) => {
     data_qualifier: values.data_qualifier,
     data_categories: values.data_categories,
   };
-  const allDataCategories = useSelector(selectDataCategories);
+  const allEnabledDataCategories = useSelector(selectDataCategories).filter(
```
allisonking, line 44: it might be slightly nicer to have a separate selector in the slice file called something like `selectEnabledDataCategories` which looks something like:

```typescript
export const selectEnabledDataCategories = createSelector(selectDataCategories, (dataCategories) =>
  dataCategories.filter(dc => dc.active)
);
```
and then just call `selectEnabledDataCategories` wherever you need it (instead of having the component call a filter each time).

this really isn't a big deal right now though, so you can consider it optional!

allisonking (approved):
nice job working through a complicated feature that touched a lot of parts @jpople !! left a few comments, though I wouldn't consider any of them blocking 👍

allisonking:
oh but one question: do we need privacy notices to be filtered too? they also have a `data_uses` field: 
image

Issue #3780, opened 2023-07-13
Add calls for consent reporting to `fides-js` and `privacy-center`


allisonking:
### Is your feature request related to a specific problem?

https://github.com/ethyca/fides/pull/3777 adds endpoints related to consent reporting which `fides-js` and `privacy-center` need to be updated to call.

#### For fides-js

When notices are shown, `PATCH http://localhost:8080/api/v1/notices-served`. Example payload:
```
{
  "browser_identity": {
    "fides_user_device_id": "3e2e5f7c-5367-47d4-b672-9427626c71db"
  },
  "privacy_notice_history_ids": [
    "pri_52008ca3-406e-420f-8509-9429c73ef859"
  ],
  "privacy_experience_id": "pri_6ec7166a-28a8-49d9-9984-9b429d393d11",
  "user_geography": "us_ca",
  "acknowledge_mode": false,
  "serving_component": "overlay"
}
```
This will return an array of `served_notice_history_id`

Then when we do a PATCH http://localhost:8080/api/v1/privacy-preferences, include the `served_notice_history_id`:
```
{
  "browser_identity": {
    "fides_user_device_id": "3e2e5f7c-5367-47d4-b672-9427626c71db"
  },
  "preferences": [
    {
      "privacy_notice_history_id": "pri_52008ca3-406e-420f-8509-9429c73ef859",
      "preference": "opt_in",
      "served_notice_history_id": "ser_e83dd23c-cc82-44bc-9046-824089f80e2e"
    }
  ],
  "policy_key": "default_consent_policy",
  "privacy_experience_id": "pri_6ec7166a-28a8-49d9-9984-9b429d393d11",
  "user_geography": "us_ca",
  "method": "button"
}
```

#### For privacy center
PATCH `http://localhost:8080/api/v1/consent-request/<consent_request_id>/notices-served`
```
{
  "browser_identity": {
    "fides_user_device_id": "3e2e5f7c-5367-47d4-b672-9427626c71db"
  },
  "privacy_notice_history_ids": [
    "pri_52008ca3-406e-420f-8509-9429c73ef859"
  ],
  "privacy_experience_id": "pri_6ec7166a-28a8-49d9-9984-9b429d393d11",
  "user_geography": "us_ca",
  "acknowledge_mode": false,
  "serving_component": "privacy_center"
}
```
which returns a list of `served_notice_history_id`

`PATCH http://localhost:8080/api/v1/consent-request/<consent_request_id>/privacy-preferences`
[...]

PR #3845, opened 2023-07-20
Fides-js consent reporting calls

allisonking:
Closes https://github.com/ethyca/fides/issues/3780

### Description Of Changes

Adds calls to consent reporting in fides-js! We need fides-js to make one extra PATCH call to a new `/notices-served` endpoint when the modal is showing.

This patch returns a list of served notice IDs, which needs to go into the existing PATCH to user preferences.

### Code Changes

* [x] Update typescript types 
* [x] Add a new Fides event `FidesUIShown`. I thought this would be a clean way to implement this feature since we only want to patch when the UI is shown (specifically, the modal). Also, I think we'll need this event for TCF work, so I thought it made sense to try it out now.
  * [x] Since this event needed some more data, I updated the payload of a `FidesEvent` to take the whole cookie object as well as `extraDetails`
* [x] Call this event when banner or modal are opened up
* [x] Add a hook `useConsentServed` which checks if this was from the modal, then issues the patch
* [x] Add served notices to the PATCH to user preferences
* [x] Cypress tests

### Steps to Confirm

* Run the privacy center via `turbo dev`
* Visit localhost:3000/fides-js-demo.html and open the network tab
* Open the modal
* You should see a call to `/notices-served`. The payload should have a `serving_component` of `overlay` and some other parts that match your data (`fides_user_device_id`, user geography, etc)
* Now opt out of notices
* Inspect the PATCH to `/privacy-preferences` and make sure that the `preferences` has a `served_notice_history_id`, i.e.

```
{
  "privacy_notice_history_id": "pri_1a4b36b3-f92a-48f4-9e72-a54ee86e10bb",
  "preference": "opt_out",
  "served_notice_history_id": "ser_d49cb43d-2d42-4e0b-8224-86ef1b527858"
}
```

### Pre-Merge Checklist

* [x] Update `CHANGELOG.md`


clients/fides-js/src/components/ConsentButtons.tsx
```diff
@@ -14,22 +14,21 @@ const ConsentButtons = ({
...
   enabledKeys: NoticeKeys;
+  isAcknowledge: boolean;
   isInModal?: boolean;
 }) => {
   if (!experience.experience_config || !experience.privacy_notices) {
     return null;
   }
 
   const { experience_config: config, privacy_notices: notices } = experience;
-  const isAllNoticeOnly = notices.every(
```
allisonking, line 30: moved this up into `Overlay.tsx` so that we can send a value to `acknowledge_mode` in the patch notices-served request

eastandwestwind:
Leaving a note that I've manually tested this PR and it works for me:

Screenshot 2023-07-26 at 6 51 43 PM

eastandwestwind (approved):

pattisdr:
This looks really nice @allisonking!  I verified that all of the appropriate columns are being filled out in the backend - saving that a notice was served creates two records, a ServedNoticeHistory (for every single time a notice served) as well as upserts a LastServedNotice that consolidates the last time a notice was served across versions/across time/across different identities of the same user.  All of that looked good to me here -  Verified that things look good on the backend too when you save preferences -

How do you keep everything matched up?

allisonking:
> This looks really nice @allisonking! I verified that all of the appropriate columns are being filled out in the backend - saving that a notice was served creates two records, a ServedNoticeHistory (for every single time a notice served) as well as upserts a LastServedNotice that consolidates the last time a notice was served across versions/across time/across different identities of the same user. All of that looked good to me here - Verified that things look good on the backend too when you save preferences -
> 
> How do you keep everything matched up on the FE?

Whew, nice! The general flow is:

* when the modal opens, we hit the PATCH `/notices-served` then we save the response in state. This state is cleared on page refresh 
* when the user saves their consent, we pass that saved state along to PATCH `/privacy-preferences` ([matching based on `privacy_notice_history_id`](https://github.com/ethyca/fides/pull/3845/files#diff-9b3c7a81c5f7d8251528befb21fddb259c1f939368322db0559da8b5aaa7d09aR61)

let me know if that answers your question?

pattisdr:
> let me know if that answers your question?

Got it that makes sense!

PR #3857, opened 2023-07-25
vendor list dropdown

TheAndrewJackson:
Closes fidesplus#985

### Description Of Changes
Add a new vendor list dropdown for fidesplus that gives users the ability to link a system with an entry from the dictionary service

image

### Code Changes

* [ ] Add a new `CustomOption` component to the `CustomSelect` form component
* [ ] Update system form to conditionally display the vendor dropdown

### Steps to Confirm
OSS
* [ ] run the admin ui and fides
* [ ] create a system
* [ ] verify that the vendor dropdown isn't there
* [ ] edit the system and save. Make sure the saving works

Plus
message @TheAndrewJackson for fidesplus setup
* [ ] run the admin ui and fidesplus
* [ ] create a system
* [ ] verify that the vendor dropdown is there
* [ ] select an option from the vendor dropdown
* [ ] save. make sure it correctly saves.
* [ ] refresh the page and make sure it populates the vendor dropdown

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`


clients/admin-ui/src/features/common/form/inputs.tsx
```diff
@@ -125,10 +130,29 @@ export const ErrorMessage = ({
...
 
+const ClearIndicator = () => null;
+
 export interface Option {
   value: string;
   label: string;
+  description?: string;
 }
+
+const CustomOption: React.FC<
```
TheAndrewJackson, line 141: The `SelectComponentsConfig<Option, boolean, GroupBase<Option>>` type for `components` added below caused the linter to start throwing errors because it requires a `ReactElement` type. I needed to refactor it out into a dedicated component to fix those errors.

TheAndrewJackson: I also decided to not make `CustomOption` component a prop that can be passed down to keep things simple. It we start requiring more custom option components it can be updated in the future to accommodate that

clients/admin-ui/src/features/system/form.ts
```diff
@@ -91,4 +96,12 @@ export const transformFormValuesToSystem = (formValues: FormValues): System => {
         ? undefined
         : formValues.administrating_department,
   };
+  if (features.plus) {
+    if (features.dictionaryService && formValues?.meta?.vendor?.id) {
```
TheAndrewJackson, line 100: I broke it up it to two nested ifs just in case we other additional plus only features that could be toggled on or off

clients/admin-ui/src/features/plus/types.ts
```diff
@@ -0,0 +1,34 @@
+export type Page<T> = {
```
allisonking, line 1: is there a reason why this doesn't come from `types/api`?

clients/admin-ui/src/features/system/SystemInformationForm.tsx
```diff
@@ -83,10 +85,27 @@ const SystemInformationForm = ({
     [passedInSystem, customFields.customFieldValues]
   );
 
+  const features = useFeatures();
+
   const [createSystemMutationTrigger, createSystemMutationResult] =
     useCreateSystemMutation();
   const [updateSystemMutationTrigger, updateSystemMutationResult] =
     useUpdateSystemMutation();
+  const { data } = useGetAllDictionaryEntriesQuery();
```
allisonking, line 94: nit: since this is a system form, can we rename `data` to something like `dictionaryData`? otherwise if I were skimming the file I'd assume `data` is system data.

allisonking: should we `skip` this get if we are in plus? or maybe only do it if the dictionary service is healthy?

clients/admin-ui/src/features/system/form.ts
```diff
@@ -5,6 +5,8 @@ import {
 import { DEFAULT_ORGANIZATION_FIDES_KEY } from "~/features/organization";
 import { DataProtectionImpactAssessment, System } from "~/types/api";
 
+import { Features } from "../common/features";
```
allisonking, line 8: nit: let's use an absolute import here

clients/admin-ui/src/features/system/form.ts
```diff
@@ -91,4 +96,12 @@ export const transformFormValuesToSystem = (formValues: FormValues): System => {
         ? undefined
         : formValues.administrating_department,
   };
+  if (features.plus) {
+    if (features.dictionaryService && formValues?.meta?.vendor?.id) {
+      payload.meta = {};
+      payload.meta.vendor = {};
+      payload.meta.vendor.id = formValues.meta.vendor.id;
```
allisonking, line 103: would this work?
```suggestion
      payload.meta = { vendor: { id: formValues.meta.vendor.id } };
```

allisonking (commented):
looking good! left some comments, but I haven't tested yet (waiting for instructions for fidesplus setup!)

clients/admin-ui/src/features/common/form/inputs.tsx
allisonking: can we leave a comment about what this custom option is? is it that we conditionally render a description if one exists?

clients/admin-ui/src/features/plus/types.ts
TheAndrewJackson: yes, the reason is the plus endpoint this data comes from is currently typed as `Dict[any,any]`. It's serving as a proxy to the dictionary service. 

I'm assuming it was typed as `Dict[any,any]` to avoid copy/pasta-ing the same pydantic models between the two services. I can ask to see if we can copy those models over so we get automated types for those models.

clients/admin-ui/src/features/system/SystemInformationForm.tsx
TheAndrewJackson: both great ideas. I'll add those in now

clients/admin-ui/src/features/system/form.ts
TheAndrewJackson: I think this needs to be handled a little differently. both of our impls will overwrite everything that is within meta. So something like `meta.connector` would be deleted. I'll update it to handle that scenario.

clients/admin-ui/src/features/plus/types.ts
TheAndrewJackson: In the short term we're going to keep using the handmade types because the schema isn't fully defined yet on the BE. Once it's fully defined we'll remove the handmade types and used the generated ones.

TheAndrewJackson:
@allisonking I think I've addressed everything from the first round of feedback 👍

allisonking (approved):
tested it out and it works great! 😎 would also be nice to have cypress tests for this, though it could wait until things are more stable if you'd prefer!

Issue #3808, opened 2023-07-18
Separately bundle TCF code for `fides-js`


allisonking:
### Is your feature request related to a specific problem?

We want to architect fides-js such that TCF can be optionally included. This is so that our bundle size for vanilla fides-js can stay reasonably small, since TCF features are relatively large.

### Describe the solution you'd like

TCF can be toggled on or off in the Fides server. If it is toggled on, then the privacy center should return a script `fides-tcf.js` instead of vanilla `fides.js`.

`fides-tcf.js` should be our core `fides.js` but also include all TCF logic, encapsulated in a file `tcf.ts`. 

This would likely entail:
* [ ] Put core TCF logic in a new file `tcf.ts`
* [ ] Separate "main" function that is similar to `fides.ts` but called `fides-tcf.ts`, and includes imports from `tcf.ts`
* [ ] Modify the rollup file to also build a `fides-tcf.js` bundle
* [ ] Add logic to the privacy center which fetches whether TCF is enabled in the Fides server. If it is enabled, then send the `fides-tcf.js` bundle instead of the `fides.js` bundle.

### Describe alternatives you've considered, if any

There are probably other ways to do this, though I'm not sure what they are yet. Would love thoughts from @NevilleS + @eastandwestwind !

### Additional context

Original design doc: https://ethyca.atlassian.net/wiki/spaces/EN/pages/2737111165/2023-07-07+TCF+frontend+planning

eastandwestwind:
Hey @allisonking this strategy overall looks good. I've created a POC to test this way of conditional bundling: https://github.com/ethyca/fides/compare/main...poc-tcf-bundle

To test, run `npm run dev-pc` and see that the tcf file is included:
<img width="497" alt="Screenshot 2023-07-18 at 8 13 13 PM" src="https://github.com/ethyca/fides/assets/7697292/1b531ee4-d518-4bc9-ac1d-cdb16c8159c9">


Update the `tcf_enabled` vars to `false` (in 2 places for now, but we should update to an env var) to see that tcf is not included:
<img width="463" alt="Screenshot 2023-07-18 at 8 16 55 PM" src="https://github.com/ethyca/fides/assets/7697292/447a1b32-e031-4573-9d2c-af0bc00e669d">

Is this along the lines of your thinking?

Roger-Ethyca:
moving to done

Issue #3811, opened 2023-07-18
Transform privacy experience data to TCF data to generate the TC string


allisonking:
### Is your feature request related to a specific problem?

The backend is going to give us privacy experience data which needs to be converted to data that can be marshalled into a TCF string.

### Describe the solution you'd like

We'll be trying to use the iabtcf-es library https://github.com/InteractiveAdvertisingBureau/iabtcf-es. This ticket should cover going from a privacy experience payload to something the library can handle (it has its own TCF Model object).

### Describe alternatives you've considered, if any

_A description of any alternative solutions or features you've considered._

### Additional context

This requires the backend data to be pretty stable.

PR #3859, opened 2023-07-25
Separately bundle TCF

allisonking:
Closes https://github.com/ethyca/fides/issues/3808

### Description Of Changes

Extends @eastandwestwind 's poc to bundle TCF separately. There are a few areas that are a little murky, so I'm not 100% sure this is the right approach, but I think it's worth a discussion! 

The thing that this _doesn't_ do is separately bundle TCF components. I'm not quite sure how to do that—we could certainly have i.e. a `TcfOverlay` vs `Overlay` but I'm not sure if that's the right approach. Dynamic imports/lazy loading might be the direction we want to go, but more research is needed!

### Code Changes

* [x] Refactor rollup to be able to generate a new script `fides-tcf.js`
* [x] Refactor `fides.ts` so that its core functionality is in a new `initialize.ts` file so that `fides-tcf.ts` can use the same functions
* [x] `tcfEnabled` is no longer a `FidesConfig` option that can be passed into `init`, or at least it won't do what you think it will. See the comments for an in depth reason why, but basically it's because that would determine which script is used (`fides.ts` vs `fides-tcf.ts`) and `init` is on one of those scripts! So now it's only configurable via environment variable.

### Steps to Confirm

* `turbo dev` in privacy center and make sure `localhost:3000/fides-js-demo.html` still works okay
* `FIDES_PRIVACY_CENTER__TCF_ENABLED=true turbo dev` and visit `localhost:3000/fides-js-demo.html`
* You should see in the console log at the very end `tcf loaded`

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`


clients/fides-js/rollup.config.mjs
```diff
@@ -21,70 +25,91 @@ const preactAliases = {
...
+          console.log(
+            `✅ ${fileName} gzipped size passed maximum size checks (${gzipSize} < ${gzipErrorSizeKb} KB)`
+          );
+        }
+      },
+    ],
+  }),
+];
+
+const SCRIPTS = [
```
allisonking, line 76: refactored so that we didn't have to copy/paste as much. in the future, we can add more scripts to this array, and add whatever props we need to make the scripts work 👍

clients/fides-js/src/fides-tcf.ts
```diff
@@ -0,0 +1,131 @@
+/**
```
allisonking, line 1: Refactored the original `fides.ts` so that much of its functional logic is in the new `initialize.ts` file. Now `fides-tcf` and `fides.ts` both use those functions

clients/fides-js/src/fides-tcf.ts
```diff
@@ -0,0 +1,131 @@
...
+  // Dispatch the "FidesInitialized" event to update listeners with the initial
+  // state. Skip if we already initialized due to an existing cookie.
+  // For convenience, also dispatch the "FidesUpdated" event; this allows
+  // listeners to ignore the initialization event if they prefer
+  if (isNewFidesCookie(cookie)) {
+    dispatchFidesEvent("FidesInitialized", cookie, config.options.debug);
+  }
+  dispatchFidesEvent("FidesUpdated", cookie, config.options.debug);
+
+  tcf();
```
allisonking, line 91: this line is the only difference between `fides.ts` and `fides-tcf.ts` right now, but we could certainly add more.

This refactor was a bit difficult since we use this `_Fides` global variable which stays up to date with `window.Fides`. that's not functional, and so it's hard to just pull all of the original initialize logic into functions directly, especially if we want to dispatch events after changes to the global variable. so I broke the initialization logic into a few functions which return updates to the Fides object, and then this function is in charge of applying those updates (via `Object.assign`)

clients/privacy-center/pages/api/fides-js.ts
```diff
@@ -70,6 +70,11 @@ export default async function handler(
     }));
   }
 
+  // DEFER: The Fides server controls whether or not TCF is enabled, and so we
+  // should prefetch that value here. Until the endpoint is ready, we use an
+  // environment variable
+  const tcfEnabled = environment.settings.TCF_ENABLED;
```
allisonking, line 76: This is a little weird and we might want to think about this more—but whether TCF is enabled will be determined by the backend, not via an an option or an environment variable. Of course, we can have an option to override it, but I was thinking the PC would need to prefetch if we are using TCF from the Fides server, then determine which bundle to use. Since that determines which bundle to use, the bundles themselves don't respect the `tcfEnabled` option since they can't just switch themselves out with another bundle...

clients/privacy-center/public/fides-js-components-demo.html
```diff
@@ -119,7 +119,6 @@
           modalLinkId: null,
           privacyCenterUrl: "http://localhost:3000",
           fidesApiUrl: "http://localhost:8080/api/v1",
-          tcfEnabled: false,
```
allisonking, line 122: I removed this so that it's not misleading that this does anything right now (sending the option to the `window.Fides.init` function is too late, since by then, we've decided if we're using `fides.js` or `fides-tcf.js`). it's a little confusing to think about, so maybe that means this isn't quite the right approach...

eastandwestwind (commented):
High level looking good, but one thing we'll need to think about is how this architecture might support triggering TCF-specific callbacks / events from common code?

clients/fides-js/src/fides-tcf.ts
```diff
@@ -0,0 +1,138 @@
...
+import { tcf } from "./lib/tcf";
+import {
+  getInitialCookie,
+  getInitialFides,
+  initialize,
+} from "./lib/initialize";
+import type { Fides } from "./lib/initialize";
+import { dispatchFidesEvent } from "./lib/events";
+import { isNewFidesCookie } from "./fides";
+import { renderOverlay } from "./lib/tcf/renderOverlay";
```
allisonking, line 63: update 8/16: this is the second line that is different between `fides-tcf.ts` and `fides.ts`. I chose to use two separate files for `renderOverlay` since that's a `tsx` file, and I thought changing `fides.ts` to `fides.tsx` would cause more downstream changes

clients/fides-js/src/lib/tcf/renderOverlay.tsx
```diff
@@ -0,0 +1,7 @@
+import { ContainerNode, render, h } from "preact";
+import TcfOverlay from "../../components/tcf/TcfOverlay";
+import { OverlayProps } from "../../components/types";
+
+export const renderOverlay = (props: OverlayProps, parent: ContainerNode) => {
+  render(<TcfOverlay {...props} />, parent);
```
allisonking, line 6: this file is the same as `lib/renderOverlay` except we return the `TcfOverlay` component instead of the `NoticeOverlay` component

allisonking:
I changed this to branch off of https://github.com/ethyca/fides/pull/3870 instead. That PR does the work of splitting the `Overlay` component into two components: `NoticeOverlay` and `TcfOverlay`. That addresses some of my concerns from earlier about how we'd bundle certain components but not others between `fides` and `fides-tcf`. Now when you build, you'll get two dists:

```
fides-js:build: ┌────────────────────────────────┐
fides-js:build: │                                │
fides-js:build: │   Destination: dist/fides.js   │
fides-js:build: │   Bundle Size:  59.27 KB       │
fides-js:build: │   Minified Size:  59.28 KB     │
fides-js:build: │   Gzipped Size:  18.63 KB      │
fides-js:build: │                                │
fides-js:build: └────────────────────────────────┘
```

and 

```
fides-js:build: ┌────────────────────────────────────┐
fides-js:build: │                                    │
fides-js:build: │   Destination: dist/fides-tcf.js   │
fides-js:build: │   Bundle Size:  68.31 KB           │
fides-js:build: │   Minified Size:  68.38 KB         │
fides-js:build: │   Gzipped Size:  20.42 KB          │
fides-js:build: │                                    │
fides-js:build: └────────────────────────────────────┘
```

clients/fides-js/src/lib/consent.ts
```diff
@@ -41,30 +42,16 @@ export const initOverlay = async ({
         document.body.prepend(parentElem);
       }
 
-      // DEFER(fides#3859): use a separate TCF bundle instead of this if statement
-      // so that the vanilla fides-js bundle does not need to know about TCF
```
eastandwestwind, line 45: wooooot! Feels so good removing `DEFER`s

eastandwestwind (approved):
looks great @allisonking ! I think we should merge this first, then I can work on putting https://github.com/ethyca/fides/issues/3811 on top of this infra

Issue #3809, opened 2023-07-18
Render TCF modal UI based on data from backend


allisonking:
### Is your feature request related to a specific problem?

https://github.com/ethyca/fides/pull/3803 handled generic UI components based on the TCF designs. This ticket is responsible for integrating with the backend. It is split into two parts:
1. Rendering the information from the backend
2. Updating data back to the backend after user takes their actions

This ticket is part 1.
Part 2 is #3810.

### Describe the solution you'd like

While backend work is in progress, we can likely still make good progress if we have a stubbed JSON response. This ticket covers:
* Retrieving information from the backend
* Rendering all toggles + extra information (examples, descriptions, cookie table, etc.) 

### Describe alternatives you've considered, if any

_A description of any alternative solutions or features you've considered._

### Additional context

Design: https://www.figma.com/file/dmEwdK3xZwjKfGVQ9t66xe/Fides-v.2-Master-Working-Files?node-id=7853%3A207836&mode=dev

PR #3870, opened 2023-07-27
Refactor overlay to allow components to not necessarily be notice driven

allisonking:
Half of https://github.com/ethyca/fides/issues/3809

### Description Of Changes

Refactor of `fides-js` components so they can handle something other than privacy notices. See [this message for more context](https://ethyca.slack.com/archives/C05E74TR0BB/p1690407064504709)

Mostly this involves breaking existing components down so that their shared parts are their own component, but notice specific logic can be pulled into a different component. Everything should still work as it did before, but now if it receives an experience with `component=tcf_overlay`, it'll render the `<TcfTabs/>` instead!

### Code Changes

* [x] Add TCF types based on https://github.com/ethyca/fides/pull/3804
* [x] Add cypress scaffolding so we can use a mock payload for in development TCF work
* [x] Refactor `Overlay` into `NoticeOverlay` and `TcfOverlay`
* [x] Refactor consent buttons and notice toggles to separate out privacy notice logic
* [x] Render TcfTabs (and remove their notice dependency as well)

### Steps to Confirm

* Everything works the same
* Use cypress to see the TCF version

### Pre-Merge Checklist

* [ ] Update `CHANGELOG.md`


clients/privacy-center/cypress/e2e/consent-banner.cy.ts
```diff
@@ -2,141 +2,16 @@ import {
...
-  ConsentMechanism,
-  EnforcementLevel,
-  FidesOptions,
-  PrivacyExperience,
-  UserConsentPreference,
-  UserGeolocation,
-} from "fides-js/src/lib/consent-types";
-import { FidesEndpointPaths } from "fides-js/src/services/fides/api";
-
-enum OVERRIDE {
```
allisonking, line 21: moved this over to `stubs.ts` so that my new cypress test could share these handy functions

clients/privacy-center/cypress/e2e/consent-banner-tcf.cy.ts
```diff
@@ -0,0 +1,17 @@
...
+    cy.fixture("consent/experience_tcf.json").then((experience) => {
+      stubConfig({
+        options: {
+          isOverlayEnabled: true,
+        },
+        experience: experience.items[0],
+      });
+    });
+  });
+  it("can render the TCF modal", () => {});
```
allisonking, line 16: doesn't test for anything right now, but this is a convenient way to "see" the tcf modal

clients/fides-js/src/lib/consent.tsx
```diff
@@ -39,10 +41,12 @@ export const initOverlay = async ({
         document.body.prepend(parentElem);
       }
 
+      // DEFER: use a separate TCF bundle instead of this if statement
```
allisonking, line 44: okay SO the main idea here is that there is a different entrypoint component for TCF vs Notices. Rather than just our handy `Overlay` component, we now have `NoticeOverlay` and `TcfOverlay`, both of which wrap the original `Overlay` component. this is handy for code splitting, because (once we move away from the if/else logic, and instead have `fides-tcf.ts` render `TcfOverlay`) rollup should be able to figure out which components we need for each bundle

clients/fides-js/src/components/index.ts
```diff
@@ -1,2 +1 @@
-export { default as NoticeToggles } from "./NoticeToggles";
```
allisonking, line 1: this was from my experiments of trying to use the toggles in privacy center, and isn't used

clients/fides-js/src/components/Tcf/TcfOverlay.tsx
```diff
@@ -0,0 +1,95 @@
...
+import { TcfConsentButtons } from "../ConsentButtons";
+import { OverlayProps } from "../types";
+import TcfTabs from "./TcfTabs";
+
+const TcfOverlay: FunctionComponent<OverlayProps> = ({
+  experience,
+  options,
+  cookie,
+}) => {
+  // TODO: how will we initialize TCF data?
```
allisonking, line 18: all of these TCF specific TODOs I figured should go in a separate PR since this refactor is already large

clients/fides-js/src/components/Overlay.tsx
```diff
@@ -1,58 +1,51 @@
-import { h, FunctionComponent } from "preact";
```
allisonking, line 1: Overlay has gone through a big refactor to remove everything that made it PrivacyNotice dependent. Instead, `NoticeOverlay` controls all notice related logic, while `Overlay` is mostly just in charge of controlling showing banner/modal logic

clients/fides-js/src/components/DataUseToggle.tsx
```diff
@@ -0,0 +1,78 @@
+import { ComponentChildren, VNode, h } from "preact";
+import { useDisclosure } from "../lib/hooks";
+import Toggle from "./Toggle";
+
+interface DataUse {
+  key: string;
+  name?: string;
+}
+
+const DataUseToggle = ({
```
allisonking, line 10: changed `NoticeToggle` to `DataUseToggle` to make it less specific to notices. It takes an obj with a `key` and `name` instead of a notice object so that the TCF models can in theory use it as well

clients/fides-js/src/components/notices/NoticeToggles.tsx
```diff
@@ -0,0 +1,57 @@
+import { h } from "preact";
+
+import { ConsentMechanism, PrivacyNotice } from "../../lib/consent-types";
+
+import Divider from "../Divider";
+
+import { GpcBadgeForNotice } from "../GpcBadge";
+import DataUseToggle from "../DataUseToggle";
+
+const NoticeToggles = ({
```
allisonking, line 10: Like `NoticeOverlay`, this is a notice specific version of a list of an existing component (`NoticeToggle` --> `DataUseToggles`)

clients/fides-js/src/components/ConsentButtons.tsx
```diff
@@ -5,29 +5,78 @@ import {
   PrivacyExperience,
   ConsentMechanism,
   PrivacyNotice,
+  ExperienceConfig,
 } from "../lib/consent-types";
 
-type NoticeKeys = Array<PrivacyNotice["notice_key"]>;
-
 const ConsentButtons = ({
```
allisonking, line 11: there are now 3 components in this file:

* `ConsentButtons`
* `NoticeConsentButtons`
* `TcfConsentButtons`

The original `ConsentButtons` had a lot of notice specific logic in it, so a refactoring as needed!

eastandwestwind (approved):
This is a big one, nice work @allisonking !

Also a note that I've regression tested this PR

Issue #3877, opened 2023-07-31
Recording interactions with TCF notices


allisonking:
### Is your feature request related to a specific problem?

https://github.com/ethyca/fides/pull/3845 adds the proper consent reporting calls to the notice driven flow. We need to add this to the TCF flow as well.

### Describe the solution you'd like

We'll need to add or modify the `/notices-served` PATCH to include TCF data, then also send those notice served IDs to the `/privacy-preferences` endpoint

### Describe alternatives you've considered, if any

_A description of any alternative solutions or features you've considered._

### Additional context

https://ethyca.atlassian.net/wiki/spaces/EN/pages/2744320004/2023-07-17+Backend+TCF+2.0+POC

allisonking:
I chatted with @galvana today, and I think we should hold off on this one until https://github.com/ethyca/fides/issues/4209 (which relies on https://github.com/ethyca/fidesplus/issues/1128) since the endpoint is about to change! hopefully we can get those two PR's dealt with in the next 1-2 days though 👍

PR #3879, opened 2023-07-31
TCF modal display data

allisonking:
Closes https://github.com/ethyca/fides/issues/3809

### Description Of Changes

https://www.loom.com/share/0c8b1c05f6be4f99aab109435d808a8a?sid=ebaa2503-68b3-43ff-aa07-b925982eddb5

#### Important caveats
* This work does not yet touch the `fides_consent` cookie. We likely want to initialize values in the cookie the same way we do for notices. We haven't talked too much about what the `notice_key`s for these will be yet though. https://github.com/ethyca/fides/issues/3886
* This also does not integrate with the new consent reporting flow. That should be done in https://github.com/ethyca/fides/issues/3877

### Code Changes

* [x] Display purposes, features, and vendors in tabs
* [x] Hook up toggles

### Steps to Confirm

* Run nox -s dev in fides to start the backend
* `tcf_enabled` needs to be set to True: make a PATCH to `/consent-settings` and set `tcf_enabled: true` in the request body
* Use a VPN to go to an EU location, i.e. France
* To add purposes, you need a system with at least one data use that belongs to a GVL purpose or special purpose.  For example, [data_use: "essential.fraud_detection" maps to Special Purpose #1](https://github.com/ethyca/fideslang/compare/main...asachs/tcf-mappings#diff-a4adb03db889d2228be2f2adbcfe3ee6a2964e133c37f190bc34ae8cf8b0a7fbR117)
* You can do this through the admin-ui: run `turbo dev` in `clients/admin-ui` and then visit localhost:3000
* In `clients/privacy-center`, run `turbo dev`. If you have the admin-ui running, this will start the privacy center at localhost:3001
* Visit localhost:3001/fides-js-demo.html
* The banner should appear and you should be able to click “Manage preferences” to see the TCF modal
* To add vendors, attach an integration to a system that has a GVL data use (i.e. set up an amplitude integration with the system from step 5)

### Pre-Merge Checklist

  * [x] https://github.com/ethyca/fides/issues/3885
* [x] Update `CHANGELOG.md`


clients/fides-js/src/components/ConsentButtons.tsx
```diff
@@ -122,41 +122,3 @@ export const NoticeConsentButtons = ({
     />
   );
 };
-
-interface TcfConsentButtonProps {
```
allisonking, line 126: moved this to its own file

clients/fides-js/src/components/Toggle.tsx
```diff
@@ -28,7 +28,7 @@ const Toggle = ({
         onChange={() => {
           onChange(id);
         }}
-        defaultChecked={checked}
+        checked={checked}
```
allisonking, line 31: this is kind of a weird thing, but it turns out `defaultChecked` only provides the initial value. this started not to work when I added the "select all/deselect toggles" which wanted to control _other_ toggles' state. This seems to work well, though I had to update the cypress tests now that the `checked` prop is a lil different

more info: https://stackoverflow.com/questions/33131413/react-defaultchecked-dont-render-check-attribute-on-second-load

clients/fides-js/src/components/tcf/TcfOverlay.tsx
```diff
@@ -2,47 +2,110 @@ import { h, FunctionComponent } from "preact";
...
+import { TcfConsentButtons } from "./TcfConsentButtons";
 import { OverlayProps } from "../types";
 import TcfTabs from "./TcfTabs";
+import {
+  TCFFeatureRecord,
+  TCFPurposeRecord,
+  TCFVendorRecord,
+} from "../../lib/tcf/types";
+
+const resolveConsentValueFromTcfModel = (
```
allisonking, line 22: this is the kind of thing that can be taken out of this component and put in `lib` instead if we need to initialize the cookie value before rendering the overlay https://github.com/ethyca/fides/issues/3886

clients/fides-js/src/components/tcf/TcfOverlay.tsx
```diff
@@ -2,47 +2,110 @@ import { h, FunctionComponent } from "preact";
...
+  model: TCFPurposeRecord | TCFFeatureRecord | TCFVendorRecord
+) => {
+  if (model.current_preference) {
```
allisonking, line 25: we don't handle notice-only yet—I originally didn't think there were "notice-only" TCF purposes, but I think there might be?

clients/fides-js/src/components/tcf/TcfVendors.tsx
```diff
@@ -0,0 +1,107 @@
...
+  onChange,
+}: {
+  allVendors: PrivacyExperience["tcf_vendors"];
+  enabledIds: string[];
+  onChange: (payload: UpdateEnabledIds) => void;
+}) => {
+  const [isFiltered, setIsFiltered] = useState(false);
+
+  if (!allVendors || allVendors.length === 0) {
+    // TODO: empty state?
```
eastandwestwind, line 46: do we need product input on this?

clients/fides-js/src/components/tcf/TcfOverlay.tsx
eastandwestwind: I'm unsure. @rsilvery - are you aware of any TCF purposes being notice-only?

PR #3884, opened 2023-08-01
Update Healthcheck logic

### LINKED ISSUE #2517: When Workers are Enabled Webserver Health Check Takes a Long Time to Respond
### Bug Description

When workers are enabled, the `/health` route on the webserver takes > 1 second to respond. I tried with 1, 5, and 10 workers and didn't notice any significant difference in the time it took. 

Results of 10 trials each:

Worker Count | Min Time | Median Time | Max Time
-- | -- | -- | --
1 | 1.13 | 2.48 | 4.29
5 | 1.17 | 2.84 | 4.84
10 | 1.1 | 1.735 | 4.28

#### Steps to Reproduce

1. Set up workers
2. Check health page of webserver route

### Expected behavior

< 1 second http response time

### Screenshots

### Environment

* Version:
* OS:
* Python Version:
* Docker Version:

### Additional context

Kubernetes default timeout for probes is 1 second, meaning with these times, Kubernetes will kill the pods if a liveness probe does not have a `timeoutSeconds` >= ~5 seconds

ThomasLaPiana:
Closes #2517 

### Description Of Changes

The `database` health check step was found to be extremely heavy, sometimes spiking response times to ~1 sec when running locally with nothing else hitting the server

After doing some investigation of the `worker` checks as well, I couldn't find evidence that this was an issue. It only increased processing time by ~6ms. After checking everything I believe that the database was the primary culprit here, but we'll continue to monitor this fix as it is rolled out and verify that conclusion.

### Code Changes

* [x] simplify the database health check

### Steps to Confirm

* [ ] open up `ipython` and run some tests using `requests` to get those logs populated, verify the numbers look right
* [ ] go to `database.py` and force the url to be wrong for the healthcheck function. Ensure that the status is 503 and the error is logged

### Pre-Merge Checklist

* [x] Update `CHANGELOG.md`

### DIFF
--- a/src/fides/api/api/v1/endpoints/health.py
+++ b/src/fides/api/api/v1/endpoints/health.py
@@ -80,9 +80,7 @@ def get_cache_health() -> str:
         },
     },
 )
-async def health(
-    db: Session = Depends(get_db),
-) -> Dict:  # Intentionally injecting the ops get_db
+async def health() -> Dict:
     """Confirm that the API is running and healthy."""
     database_health = get_db_health(CONFIG.database.sync_database_uri)
     cache_health = get_cache_health()
--- a/src/fides/api/api/v1/endpoints/health.py
+++ b/src/fides/api/api/v1/endpoints/health.py
@@ -97,6 +97,7 @@ async def health(
     fides_is_using_workers = not celery_app.conf["task_always_eager"]
     if fides_is_using_workers:
         response["workers_enabled"] = True
+        # Figure out a way to make this faster
         response["workers"] = get_worker_ids()
 
     for _, value in response.items():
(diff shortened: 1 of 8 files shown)


ThomasLaPiana:
After refactoring the database check:

image

ThomasLaPiana:
with a worker enabled:

image

ThomasLaPiana:
I don't believe any of these test failures are related to these changes

src/fides/api/api/v1/endpoints/health.py
```diff
@@ -84,7 +84,7 @@ async def health(
     db: Session = Depends(get_db),
```
RobertKeyser, line 84: Is this parameter still needed now that the `get_db_health` call doesn't have `db=db` in it?

SteveDMurphy (approved):
Nice work @ThomasLaPiana ! Looks good and nice use of the existing db health check in `get_db_engine`

PR #3890, opened 2023-08-01
Fixing OAuth2 redirect URL

galvana:
### Description Of Changes

When a user configures an integration that uses the OAuth2 authorization code flow, they must follow an authorization URL to enter their credentials and authenticate the connection. When the authorization URL is generated, we also store the referrer to be able to redirect the user back to the system configuration page after authorization. There is a scenario where the referrer URL is the generic system URL (http://localhost:3000/add-systems/manual) and not the unique system path (http://localhost:3000/systems/configure/salesforce). This causes the user to be redirected to a generic system creation page after authorization instead of the system's configuration page.

### Code Changes

* [ ] Only use the schema (http) and netloc (localhost:3000) and generate the rest of the URL

### Steps to Confirm

* [ ] Navigate to Data Map > Add Systems
* [ ] Create a Salesforce system and hit save and **do not follow the link in the toast**
image
* [ ] Enter the credentials
* [ ] Verify the URL you are on is still http://base-url/add-systems/manual
* [ ] Click `Authorize connection`
* [ ] Verify you are redirected to the system configuration page for the system you just created

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### DIFF
--- a/src/fides/api/api/v1/endpoints/oauth_endpoints.py
+++ b/src/fides/api/api/v1/endpoints/oauth_endpoints.py
@@ -1,4 +1,5 @@
 from typing import Dict, List, Optional
+from urllib.parse import urlparse
 
 from fastapi import Body, Depends, HTTPException, Request, Response, Security
 from fastapi.responses import PlainTextResponse, RedirectResponse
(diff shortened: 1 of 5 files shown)


src/fides/api/api/v1/endpoints/oauth_endpoints.py
```diff
@@ -254,8 +255,16 @@ def oauth_callback(code: str, state: str, db: Session = Depends(get_db)) -> Resp
...
+            # and omit the path. This is done to ensure the correct redirection URL, as the
+            # user might not be on the system's unique page when they follow the
+            # authorization link out of Fides.
+
+            parsed_url = urlparse(authentication_request.referer)
+            base_url = f"{parsed_url.scheme}://{parsed_url.netloc}"
+            system_key = connection_config.system.fides_key
             return RedirectResponse(
-                url=f"{authentication_request.referer}?status={test_status_value}"
+                url=f"{base_url}/systems/configure/{system_key}?status={test_status_value}"
```
TheAndrewJackson, line 267: Can you add a unit test for this to double check that it's always generating URLs correctly?

TheAndrewJackson (approved):

Issue #3851, opened 2023-07-24
Server-side (Privacy Center) API calls are blocked when testing locally


NevilleS:
### Bug Description

When running the app locally using any `nox` commands - but most importantly, using `nox -s "fides_env(test)"` or `fides deploy`, there is a `FIDES_PRIVACY_CENTER__FIDES_API_URL` config var we default to `http://localhost:8080/api/v1`

This works fine for client-side calls from the browser (both Privacy Center client and the Fides.js client), but these fail when the same call is attempted from the server since it tries to connect to `localhost` from within the `privacy-center` docker container.

#### Steps to Reproduce

1. Run `nox -s "fides_env(test)"`
2. Open http://localhost:3001/fides.js
3. Observe an `Internal Server Error` 

### Expected behavior

Should perform all the server-side initialization and return a `fides.js` bundle with `geolocation` and `experience` prefetched 👍 

### Environment

* Version: 2.17.0rc1

### Additional context

We might need to support an optional `FIDES_PRIVACY_CENTER__SERVER_SIDE_FIDES_API_URL` (or similar) to workaround this, as I can't think of a nicer way to get the client & server to agree on the API location when they are _technically_ on different networks.

Roger-Ethyca:
moving to done

Issue #3854, opened 2023-07-24
Re-enable server-side prefetching of geolocation and experience


eastandwestwind:
### Is your feature request related to a specific problem?

With server-side prefetching was disabled here https://github.com/ethyca/fides/pull/3850, but we need to properly implement it

In this ticket, we need to:

1. Add a config var like `FIDES_PRIVACY_CENTER__IS_PREFETCH_ENABLED` and have that false to just toggle this behaviour off
2. When enabled, we should detect ip address from the server header, and forward it to CloudFront using their `x-forwarded-for` header- https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/RequestAndResponseBehaviorCustomOrigin.html#RequestCustomIPAddresses

### Additional context

See https://github.com/ethyca/fides/pull/3698#pullrequestreview-1543377197 for context

Roger-Ethyca:
moving to done

PR #3895, opened 2023-08-03
3851 Fix ability to make server side API calls from privacy-center

eastandwestwind:
Closes https://github.com/ethyca/fides/issues/3851 and https://github.com/ethyca/fides/issues/3854

### Description Of Changes

This change allows us to, most importantly, be able to fetch experience data on the server-side. 

### Code Changes

* [ ] Add new env var `FIDES_PRIVACY_CENTER__SERVER_SIDE_FIDES_API_URL` to set server-side url host (i.e. from one docker container to another)
* [ ] Add new env var `FIDES_PRIVACY_CENTER__IS_PREFETCH_ENABLED` to determine whether or not to make prefetch calls from server

### Steps to Confirm

* [ ] Regression test only
* [ ] Ensure these env vars are set:
```
FIDES_PRIVACY_CENTER__IS_PREFETCH_ENABLED=false
FIDES_PRIVACY_CENTER__GEOLOCATION_API_URL=https://cdn-api.ethyca.com/location
FIDES_PRIVACY_CENTER__PRIVACY_CENTER_URL=http://localhost:3000
```
* [ ] Ensure you have notices / experiences / systems set up for your testing region
* [ ] Nav to `http://localhost:3001/fides-js-demo.html` and see that appropriate experience pops up

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`


NevilleS (changes requested):
Quick comment- but haven't finished reviewing yet 👍

clients/fides-js/src/services/external/geolocation.ts
```diff
@@ -31,6 +32,11 @@ export const getGeolocation = async (
   const fetchOptions: RequestInit = {
     mode: "cors",
   };
+  if (forwarded_ip) {
+    const headers: Headers = new Headers();
+    headers.set("cloudfront-viewer-address", forwarded_ip);
```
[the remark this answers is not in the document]
eastandwestwind: You're right about this not working with our existing lambda API. I also tried `x-forwarded-for` which didn't seem to work either. We may have to add a query param (or new custom header) to our lambda API instead of relying on any existing headers to forward IP address. What do you think?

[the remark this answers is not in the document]
NevilleS: Due to the way that API functions, I think we're just not going to be able to get it to work - it's incredibly fast and performant _because_ it leverages edge CDNs, so it doesn't really have the capability to perform a slower, more traditional geoIP lookup.

We can instead use ipinfo.io or similar for this. I'd recommend taking this part of the code out for now, and shipping the enhanced prefetch for experience data in the _special case_ where the geolocation is specified up front (e.g. `?geolocation=us_ca`) and we can handle the IP lookup upstream in the infra

clients/privacy-center/app/server-environment.ts
```diff
@@ -250,6 +254,9 @@ export const loadPrivacyCenterEnvironment =
       FIDES_API_URL:
         process.env.FIDES_PRIVACY_CENTER__FIDES_API_URL ||
         "http://localhost:8080/api/v1",
+      SERVER_SIDE_FIDES_API_URL:
+        process.env.FIDES_PRIVACY_CENTER__SERVER_SIDE_FIDES_API_URL ||
+        "http://fides:8080/api/v1",
```
NevilleS, line 259: 🤔 

Setting a really smart default value here will save us a lot of headaches. Right now the default will be `http://fides:8080/api/v1`, which will work in a local, `fides deploy`-esque setup.

However, the engineer deploying this to a hosted environment will need to set both `FIDES_PRIVACY_CENTER__FIDES_API_URL` and `FIDES_PRIVACY_CENTER__SERVER_SIDE_API_URL` to make it work there.

My instinct says you should do this instead:
```suggestion
      SERVER_SIDE_FIDES_API_URL:
        process.env.FIDES_PRIVACY_CENTER__SERVER_SIDE_FIDES_API_URL ||
        null
```

...and wherever this variable is used, do something like: `let apiUl = settings.SERVER_SIDE_FIDES_API_URL || settings.FIDES_API_URL`. In other words, I'd recommend we make this an *optional* variable that, when set, will be used for server-side fetches. But if it's not set, we use the value of the `FIDES_API_URL` config by default.

To make that work and still support local testing, you'll need to manually override this value on the docker-compose files in this repo, the sample app, and the demo repo... but you'd need to do that anyways to control `IS_PREFETCH_ENABLED` too!

The benefit here is:
* for the production use case, the engineer deploying this sets one variable (FIDES_API_URL) and it'll work
* in special cases, the engineer can use this variable to have different API URLs for client-side and server-side fetches
* we will need to use that "special case" for local deployments where the API URLs are different

clients/privacy-center/app/server-environment.ts
```diff
@@ -264,6 +271,9 @@ export const loadPrivacyCenterEnvironment =
       IS_OVERLAY_ENABLED: process.env.FIDES_PRIVACY_CENTER__IS_OVERLAY_ENABLED
         ? process.env.FIDES_PRIVACY_CENTER__IS_OVERLAY_ENABLED === "true"
         : false,
+      IS_PREFETCH_ENABLED: process.env.FIDES_PRIVACY_CENTER__IS_PREFETCH_ENABLED
+        ? process.env.FIDES_PRIVACY_CENTER__IS_PREFETCH_ENABLED === "true"
+        : true,
```
NevilleS, line 276: I'd play it safe here - there's enough things that can go wrong that we should default this to false:
```suggestion
      IS_PREFETCH_ENABLED: process.env.FIDES_PRIVACY_CENTER__IS_PREFETCH_ENABLED
        ? process.env.FIDES_PRIVACY_CENTER__IS_PREFETCH_ENABLED === "true"
        : false,
```

clients/privacy-center/common/geolocation.ts
```diff
@@ -11,9 +13,13 @@ const VALID_ISO_3166_LOCATION_REGEX = /^\w{2,3}(-\w{2,3})?$/;
 // (see https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/adding-cloudfront-headers.html#cloudfront-headers-viewer-location)
 const CLOUDFRONT_HEADER_COUNTRY = "cloudfront-viewer-country";
 const CLOUDFRONT_HEADER_REGION = "cloudfront-viewer-country-region";
+const CLOUDFRONT_VIEWER_ADDRESS = "cloudfront-viewer-address";
+const X_FORWARDED_FOR = "x-forwarded-for";
 export const LOCATION_HEADERS = [
   CLOUDFRONT_HEADER_COUNTRY,
   CLOUDFRONT_HEADER_REGION,
+  CLOUDFRONT_VIEWER_ADDRESS,
+  X_FORWARDED_FOR,
```
NevilleS, line 22: These should be removed:
```suggestion
```

Including those headers in this array has an important effect: they are used to control how an upstream cache should vary it's cache keys, etc. It's hard to explain that succinctly here, but I don't think you quite understand that impact so let's play it safe 👍

clients/privacy-center/common/geolocation.ts
```diff
@@ -26,7 +32,8 @@ export const LOCATION_HEADERS = [
  *
  */
 export const lookupGeolocation = async (
-  req: NextApiRequest
+  req: NextApiRequest,
+  settings: PrivacyCenterClientSettings
```
NevilleS, line 36: nit: for code cleanliness, I think it's better to have the caller not pass in a settings hash like this... it feels like the business logic is spidering out a lot and not very maintainable here and you'll get some subtle-but-painful logic errors.

NevilleS: The alternative is to hoist the `if (settings.***) ...` type calls out of this function, and require the caller be specific and provide the params like `geolocationApiUrl?: string`

Not sure if this makes sense, can explain live!

clients/privacy-center/pages/api/fides-js.ts
```diff
@@ -66,7 +72,35 @@ export default async function handler(
 
   // Check if a geolocation was provided via headers, query param, or obtainable via a geolocation URL;
   // if so, inject into the bundle, along with privacy experience
-  const geolocation = await lookupGeolocation(req);
+  const geolocation = await lookupGeolocation(req, environment.settings);
+
+  let experience;
```
NevilleS, line 77: nit: this is really important (and powerful!) logic so a code comment is always a friend.
```suggestion
  // If a geolocation can be determined, "prefetch" the experience from the Fides API immediately.
  // This allows the bundle to be fully configured server-side, so that the Fides.js bundle can initialize instantly!
  let experience;
```

clients/privacy-center/pages/api/fides-js.ts
```diff
@@ -66,7 +72,35 @@ export default async function handler(
 
   // Check if a geolocation was provided via headers, query param, or obtainable via a geolocation URL;
   // if so, inject into the bundle, along with privacy experience
```
NevilleS, line 74: nit:
```suggestion
```

Move this comment down a bit... see next nit!

src/fides/data/sample_project/docker-compose.yml
```diff
@@ -74,6 +74,7 @@ services:
       - ${FIDES_DEPLOY_ENV_FILE:-sample.env}
     environment:
       - FIDES_PRIVACY_CENTER__FIDES_API_URL=http://localhost:8080/api/v1
+      - FIDES_PRIVACY_CENTER__SERVER_SIDE_FIDES_API_URL=http://fides:8080/api/v1
```
NevilleS, line 77: 👍

NevilleS (changes requested):
Some nits here, but the main two points to get this ready to merge are:
1. We should remove the server-side geolocation logic and test this on some feature branches instead
2. We should set prefetch to `false` by default, for safety

clients/privacy-center/app/server-environment.ts
eastandwestwind: ahh so smart. Such a good idea to fall back on `FIDES_API_URL` if the `SERVER_SIDE_FIDES_API_URL` is not provided. Then it's also clearer that it is a special case when you have to set 2 different API urls 🌟

clients/privacy-center/common/geolocation.ts
eastandwestwind: It does look like `Vary` headers are complex 😬 I'll remove for now

clients/privacy-center/common/geolocation.ts
eastandwestwind: yeah this makes total sense 💯

NevilleS (approved):
Sorry for forgetting this! Ship it, it's disabled by default, so let's test it out with some staging and production (test) sites

Issue #3833, opened 2023-07-19
Backend work to support expanded system forms


rsilvery:
Need to determine which backend changes will be needed to support saving fields added in #3827
Also need to store features from the GVL

rsilvery:
@adamsachs  some thoughts:
- We need to add features for GVL
- We need to add a place to link logo

adamsachs:
very initial proposed updates to `fideslang` are commited on this draft PR: https://github.com/ethyca/fideslang/pull/143/files

pattisdr:
New Fideslang PR here to replace previous: https://github.com/ethyca/fideslang/pull/146

Roger-Ethyca:
moving to done

PR #3901, opened 2023-08-04
Systems/Privacy Declaration Dictionary Migration 

pattisdr:
Partially closes [#3833](https://github.com/ethyca/fides/issues/3833) and https://github.com/ethyca/fideslang/issues/139

❗ Contains schema and data migrations; check downrev before merging 
❗ ~Temporarily pinned to fideslang commit.  Need fideslang release as a prerequisite~
### Description Of Changes

Pins to a new [fideslang](https://github.com/ethyca/fideslang/pull/143) branch and adds an additive data migration that only adds new fields to Systems and Privacy Declarations and migrates data from old locations where appropriate.  New fields are not yet removed. 

Context is here: [Confluence Dictionary Forms](https://ethyca.atlassian.net/wiki/spaces/PM/pages/2742648890/Dictionary+forms) and [Fideslang Updates](https://ethyca.atlassian.net/wiki/spaces/PM/pages/2740027499/Fideslang+updates)

### Code Changes

* [x] Fideslang upgrade
* [x] New fields added to System and PrivacyDeclaration models
* [x] 2 migrations: schema migration that adds new fields followed by data migration that migrates/transforms data from previous location without altering original location
* [x] Tests

### Steps to Confirm

For testing the data migration: See [Migrations and Changes to Data Models] to note the data that is being migrated (https://ethyca.atlassian.net/wiki/spaces/PM/pages/2742648890/Dictionary+forms)
- Checkout main
- `nox -s dev -- shell`
- In separate terminal: cd /fides/clients/admin-ui and run `turbo run dev`
- Modify a data use in the taxonomy by adding special category, legal basis, recipient, and legitimate interest.
- Similarly modify a dataset by adding retention period, geographic location, and identifiability
- Create a system, and add a privacy declaration.  Attach the data use and datasets above.  Try to fill out all the fields on the system and privacy declaration
- Checkout this branch
- Bring down fides, and bring it back up again.  
- Navigate back to the system and look at the system response in the network tab and note any values that have been moved 
- Also note brand new fields in system repsonse

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`


pattisdr:
@TheAndrewJackson and @jpople marking you for review as you're integrating with this; let me know if anything seems off!

.fides/systems.yml
```diff
@@ -86,3 +86,58 @@ system:
           # - fides_db.fidesuserpermissions
           # - fides_db.providedidentity
           # - fides_db.privacyrequest
+
+  # System Info
+  - fides_key: new_dictionary_system
```
ThomasLaPiana, line 91: Why is it "new"?

.fides/systems.yml
```diff
@@ -86,3 +86,58 @@ system:
...
+    dpa_location: https://www.example.com/dpia
+    dpa_progress: pending
+    privacy_policy: https://www.example.com/privacy_policy
+    legal_name: Texas Toast
+    legal_address: 103 Trailblazing Rd. Austin, TX 29492
+    responsibility:
+      - Controller
+      - Processor
+    dpo: John Doe, CIPT
+    joint_controller_info: Jane Doe; 104 Test Road, Houston, TX 30525
```
ThomasLaPiana, line 118: this is our proper `systems.yml` for Fides, should this be in a testing file?

src/fides/api/alembic/migrations/versions/fd52d5f08c17_system_dictionary_data_migration.py
```diff
@@ -0,0 +1,374 @@
...
+            joint_controller_data.get("address"),
+            joint_controller_data.get("email"),
+            joint_controller_data.get("phone"),
+        ]
+
+        new_joint_controller_info = "; ".join(
+            [elem for elem in separated_fields if elem and elem != "N/A"]
+        )
+        if not new_joint_controller_info:
+            # If this is an empty string, let's set it back to None
```
ThomasLaPiana, line 76: Nit: Functionally an empty string and `None` produce the same value in a truthiness check, so this seems redundant unless it's for type clarity and warning the user it could be a falsey value?

src/fides/api/alembic/migrations/versions/fd52d5f08c17_system_dictionary_data_migration.py
```diff
@@ -0,0 +1,374 @@
...
+        separated_fields = [
+            joint_controller_data.get("name"),
+            joint_controller_data.get("address"),
+            joint_controller_data.get("email"),
+            joint_controller_data.get("phone"),
+        ]
+
+        new_joint_controller_info = "; ".join(
+            [elem for elem in separated_fields if elem and elem != "N/A"]
+        )
```
ThomasLaPiana, line 74: alternatively could add this:

```suggestion
        ) or None
```

image

src/fides/api/alembic/migrations/versions/fd52d5f08c17_system_dictionary_data_migration.py
```diff
@@ -0,0 +1,374 @@
...
+
+
+def _system_calculate_does_international_transfers(
+    system_third_country_transfers: Optional[List[str]],
+    dataset_third_country_transfers: Optional[List[str]],
+) -> bool:
+    """Migrating System.third_country_transfers and Dataset.third_country_transfers to System.does_international_transfers
+    If any countries are listed, regardless of which country it is, System.does_international_transfers will be set to true.
+    """
+    return bool(system_third_country_transfers) or bool(dataset_third_country_transfers)
```
ThomasLaPiana, line 89: Nit: `any` seems better here?

```suggestion
    return any((system_third_country_transfers, dataset_third_country_transfers))
```

src/fides/api/alembic/migrations/versions/fd52d5f08c17_system_dictionary_data_migration.py
```diff
@@ -0,0 +1,374 @@
...
+
+
+def _system_transform_data_responsibility_title_type(
+    dept_str: Optional[str],
+) -> List[str]:
+    """Migrating System.data_responsibility_title (str) to System.responsibility(list).  There can now be potentially many
+    items listed here"""
+    if not dept_str:
+        return []
+    return [dept_str]
```
ThomasLaPiana, line 99: Nit:

```suggestion
    return [dept_str] if dept_str else []
```

src/fides/api/alembic/migrations/versions/fd52d5f08c17_system_dictionary_data_migration.py
```diff
@@ -0,0 +1,374 @@
...
+    """
+    if not data_protection_impact_assessment or not isinstance(
+        data_protection_impact_assessment, dict
+    ):
+        return False, None, None
+
+    requires_data_protection_assessments = data_protection_impact_assessment.get(
+        "is_required", False
+    )
+    dpa_progress = data_protection_impact_assessment.get("progress", None)
```
ThomasLaPiana, line 116: Mega nit: `None` is the default fallback

```suggestion
    dpa_progress = data_protection_impact_assessment.get("progress")
```

src/fides/api/alembic/migrations/versions/fd52d5f08c17_system_dictionary_data_migration.py
```diff
@@ -0,0 +1,374 @@
...
+            dpa_progress,
+            dpa_location,
+        ) = _flatten_data_protection_impact_assessment(
+            row["data_protection_impact_assessment"]
+        )
+
+        system_data = {
+            JOINT_CONTROLLER_INFO: _system_transform_joint_controller_info(
+                json.loads(row["joint_controller_decrypted"] or "null"),
+                [  # Potentially multiple datasets on systems that have joint controller info so this is a list
```
ThomasLaPiana, line 154: This feels like a lot of nesting to me and obfuscates the function calls in my opinion

Maybe assigning these to variables before passing them would be more readable?

src/fides/api/alembic/migrations/versions/fd52d5f08c17_system_dictionary_data_migration.py
```diff
@@ -0,0 +1,374 @@
...
+def _get_third_party_data(
+    recipients: Optional[List[str]],
+) -> Tuple[bool, Optional[str]]:
+    """Migrate data from DataUse.recipients field to new PrivacyDeclaration.third_parties and
+    PrivacyDeclaration.data_shared_with_third_parties be true
+    """
+    data_shared_with_third_parties: bool = False
+    third_parties: Optional[str] = None
+
+    if recipients and isinstance(recipients, list):
```
ThomasLaPiana, line 245: is there a case where `recipients` would evaluate to `True` but wouldn't be a list?

src/fides/api/alembic/migrations/versions/fd52d5f08c17_system_dictionary_data_migration.py
```diff
@@ -0,0 +1,374 @@
...
+    return None
+
+
+def _get_third_party_data(
+    recipients: Optional[List[str]],
+) -> Tuple[bool, Optional[str]]:
+    """Migrate data from DataUse.recipients field to new PrivacyDeclaration.third_parties and
+    PrivacyDeclaration.data_shared_with_third_parties be true
+    """
+    data_shared_with_third_parties: bool = False
```
ThomasLaPiana, line 242:
```suggestion
    data_shared_with_third_parties: bool = recipients and isinstance(recipients, list)
```

src/fides/api/alembic/migrations/versions/fd52d5f08c17_system_dictionary_data_migration.py
```diff
@@ -0,0 +1,374 @@
...
+    third_parties: Optional[str] = None
+
+    if recipients and isinstance(recipients, list):
+        data_shared_with_third_parties = True
+        third_parties = "; ".join([elem for elem in recipients if elem])
```
ThomasLaPiana, line 247:
```suggestion
        third_parties = "; ".join([elem for elem in recipients if elem]) or None
```

src/fides/api/models/sql_models.py
```diff
@@ -321,22 +335,54 @@ class System(Base, FidesBase):
     meta = Column(JSON)
     fidesctl_meta = Column(JSON)
     system_type = Column(String)
-    data_responsibility_title = Column(String)
-    joint_controller = Column(PGEncryptedString, nullable=True)
-    third_country_transfers = Column(ARRAY(String))
+    joint_controller = Column(
+        PGEncryptedString, nullable=True
+    )  # Will be deprecated in favor of System.joint_controller_info
```
ThomasLaPiana, line 340: Is there a meaningful distinction between "will be deprecated" and "deprecated"?

tests/ctl/core/test_api.py
```diff
@@ -571,15 +597,62 @@ async def test_system_create(
...
-        assert result.json()["privacy_declarations"][0]["cookies"] == [
+        assert json_results["privacy_declarations"][0]["cookies"] == [
             {"name": "essential_cookie", "path": "/", "domain": "example.com"}
         ]
-        assert result.json()["privacy_declarations"][1]["cookies"] == []
-        assert len(result.json()["privacy_declarations"]) == 2
+        assert json_results["privacy_declarations"][1]["cookies"] == []
+        assert len(json_results["privacy_declarations"]) == 2
+
+        assert json_results["vendor_id"] == "test_vendor"
```
ThomasLaPiana, line 611: I think there's a cleaner way to test these

maybe something like:

```python
for key, value in system.items():
    assert json_results[key] = value
```

tests/ctl/core/test_api.py
```diff
@@ -590,6 +663,50 @@ async def test_system_create(
             cookie.name for cookie in systems[0].privacy_declarations[0].cookies
         ] == ["essential_cookie"]
         assert systems[0].privacy_declarations[1].cookies == []
+        system = systems[0]
+
+        assert system.vendor_id == "test_vendor"
```
ThomasLaPiana, line 668: something similar here:

```python
for key, value in system.items():
    assert get_attr(system, key) == value
```

tests/ctl/core/test_api.py
```diff
@@ -733,6 +850,49 @@ async def test_system_create_invalid_privacy_declarations(
...
+        assert result.status_code == 200
+        assert result.json()["fides_key"] == system.fides_key
+
+        data_stewards = result.json()["data_stewards"]
+        assert len(data_stewards) == 1
+        steward = data_stewards[0]
+
+        assert steward["id"] == system_manager.id
+        assert steward["username"] == system_manager.username
+        assert steward["first_name"] is None
```
ThomasLaPiana, line 893: nit:

```suggestion
        assert not steward["first_name"]
```

tests/ctl/core/test_api.py
```diff
@@ -1151,6 +1368,119 @@ def test_system_update_privacy_declaration_invalid_duplicate(
...
+            headers=auth_header,
+            resource_type="system",
+            json_resource=system_update_request_body_with_new_dictionary_fields.json(
+                exclude_none=True
+            ),
+        )
+
+        json_results = result.json()
+        assert result.status_code == HTTP_200_OK
+        assert json_results["vendor_id"] == "test_vendor"
```
ThomasLaPiana, line 1393: same as above, it would be nice to see a more programmatic approach

ThomasLaPiana (commented):
I left some nits and suggestions, but functionally this is looking good!

I think tests need a bit of cleaning up

pattisdr:
Thanks very much for your review @ThomasLaPiana!

.fides/systems.yml
pattisdr: I misunderstood the purpose of this file! I'll just remove this.

.fides/systems.yml
pattisdr: apologies, I thought this _was_ for testing, I'll revert back to what it was before.

src/fides/api/alembic/migrations/versions/fd52d5f08c17_system_dictionary_data_migration.py
pattisdr: My intent here was to avoid writing an empty string to the database.  In the event customer data looked like this: {"name": "N/A", "address": "N/A", "email": "N/A"}, `joint_controller_info` would be an empty string at this point.  Rather than write that to the db, I wanted to convert back to None.

src/fides/api/alembic/migrations/versions/fd52d5f08c17_system_dictionary_data_migration.py
pattisdr: Yes that is a good alternative, I'll use that instead!

src/fides/api/alembic/migrations/versions/fd52d5f08c17_system_dictionary_data_migration.py
pattisdr: that reads better, thanks!

src/fides/api/alembic/migrations/versions/fd52d5f08c17_system_dictionary_data_migration.py
pattisdr: sure I can remove the default

src/fides/api/alembic/migrations/versions/fd52d5f08c17_system_dictionary_data_migration.py
pattisdr: yes, good callout!

src/fides/api/alembic/migrations/versions/fd52d5f08c17_system_dictionary_data_migration.py
pattisdr: No I don't think so, just doing a list check before a list operation.  Habit for data migrations

src/fides/api/alembic/migrations/versions/fd52d5f08c17_system_dictionary_data_migration.py
pattisdr: so how about bool(recipients) here and I'll leave the list check below

src/fides/api/models/sql_models.py
pattisdr: I'll change to deprecated, thanks

tests/ctl/core/test_api.py
pattisdr: sure, I can compare the json response to the system values in a programmatic way, but I'd like to keep the system model assertions.

tests/ctl/core/test_api.py
pattisdr: the main thing I'm trying highlight here though is just that the first name and last name are in this response, they just happen to be null for the data steward. I'll rework with that intent!

tests/ctl/core/test_api.py
pattisdr: compromise - dry up one set, leave system assertions as-is, but compare selected system response values to system?

tests/ctl/core/test_api.py
pattisdr: Stepping back for a moment, my preference is for tests to be on the verbose side. I've seen too many programmatic/clever tests that miss when the code changes out from under it. I also think it's valuable to work through each field and verify the results are what you expect.  Of course, there's room for compromise here! I agree with you that here that my tests should be consolidated in part.

pattisdr:
Thanks again for your comments Thomas, ready for another pass when you are -

src/fides/api/alembic/migrations/versions/fd52d5f08c17_system_dictionary_data_migration.py
```diff
@@ -0,0 +1,372 @@
...
+    )
+    if joint_controller_data and isinstance(joint_controller_data, dict):
+        separated_fields = [
+            joint_controller_data.get("name"),
+            joint_controller_data.get("address"),
+            joint_controller_data.get("email"),
+            joint_controller_data.get("phone"),
+        ]
+
+        # Defaults to None value if the join ends up being an empty string
```
ThomasLaPiana, line 72:
```suggestion
```

redundant comment

src/fides/api/alembic/migrations/versions/fd52d5f08c17_system_dictionary_data_migration.py
```diff
@@ -0,0 +1,372 @@
...
+    system_third_country_transfers: Optional[List[str]],
+    dataset_third_country_transfers: Optional[List[str]],
+) -> bool:
+    """Migrating System.third_country_transfers and Dataset.third_country_transfers to System.does_international_transfers
+    If any countries are listed, regardless of which country it is, System.does_international_transfers will be set to true.
+    """
+    return any([system_third_country_transfers, dataset_third_country_transfers])
+
+
+def _system_transform_data_responsibility_title_type(
```
ThomasLaPiana, line 91: a single line function seems like a bit much, maybe a good candidate for an anonymous function wherever this gets used?

tests/ctl/core/test_api.py
ThomasLaPiana: I'll take it! Thanks 🙂

tests/ctl/core/test_api.py
```diff
@@ -1151,6 +1453,88 @@ def test_system_update_privacy_declaration_invalid_duplicate(
...
+                exclude_none=True
+            ),
+        )
+
+        assert result.status_code == HTTP_200_OK
+
+        db.refresh(system)
+
+        assert system.name == self.updated_system_name
+        assert system.vendor_id == "test_vendor"
```
ThomasLaPiana, line 1481: Not asking for a change here, just giving my thoughts

When the asserts are hardcoded like so, it can make it very hard to add additional test cases down the road using something like parametrization

If we figure out there is an edge case, instead of adding another system that catches it and letting the test do its thing, we now have to copy/pasta the entire test

Keeping tests DRY isn't a priority, verbosity is helpful, but in this case I think it decreases future testability

tests/ops/migration_tests/test_system_dictionary_data_migration.py
```diff
@@ -0,0 +1,212 @@
...
+    _get_privacy_declaration_legal_basis_for_processing,
+    _get_retention_period,
+    _get_third_party_data,
+    _system_calculate_does_international_transfers,
+    _system_transform_data_responsibility_title_type,
+    _system_transform_joint_controller_info,
+)
+
+
+class TestSystemDictionaryDataMigrationFunctions:
```
ThomasLaPiana, line 12: Very thorough, nice work here!

tests/ops/migration_tests/test_system_dictionary_data_migration.py
```diff
@@ -0,0 +1,212 @@
...
+        joint_controller_info = _system_transform_joint_controller_info(
+            joint_controller_decrypted={}, datasets_joint_controller_decrypted=[{}]
+        )
+        assert joint_controller_info is None
+
+        joint_controller_info = _system_transform_joint_controller_info(
+            joint_controller_decrypted={"address": "233 Test Town"},
+            datasets_joint_controller_decrypted=[],
+        )
+        assert joint_controller_info == "233 Test Town"
```
ThomasLaPiana, line 25: Test Town is my _favorite_ town

ThomasLaPiana (approved):
Really complex change here, you broke things down nicely

.fides/systems.yml
```diff
@@ -85,4 +85,4 @@ system:
           # - fides_db.fidesuser
           # - fides_db.fidesuserpermissions
           # - fides_db.providedidentity
-          # - fides_db.privacyrequest
+          # - fides_db.privacyrequest
```
adamsachs, line 88: impact but if you're gonna make updates to the branch/PR anyway (i.e. to update the fideslang version dependency), it'd be nice to clean up so we don't get this file in the diff!

(i realize there'd been other changes in this file initially that were reverted and seems like this just got included accidentally.)

pattisdr: sure thing, thanks

src/fides/api/alembic/migrations/versions/fd52d5f08c17_system_dictionary_data_migration.py
ThomasLaPiana: ah, makes sense!

.fides/systems.yml
adamsachs: lol i copy and pasta'd my comment wrong and it's nonsensical, i guess you were able to get the gist of it though 😆

PR #3906, opened 2023-08-07
Adding new scenario to refine_target_path

### LINKED ISSUE #3905: Null object incorrectly overwritten with masked value
### Bug Description

The output of `query_config.update_value_map` generates an incorrect value map for a null object field. The `string` value of "MASKED" generated by `query_config` is not compatible with the expected data type of `object`.

**Current output**
`{"shipping": "MASKED"}`

**Expected output**
Empty map `{}` 

### Steps to Reproduce

**Minimal dataset**
```
collections:
  - name: customer
    fields:
      - name: shipping
        fields:
          - name: address
            fields:
              - name: city
                data_categories: [user.contact.address.city]
                fidesops_meta:
                  data_type: string                  
          - name: name
            data_categories: [user.name]
            fidesops_meta:
              data_type: string
          - name: phone
            data_categories: [user.contact.phone_number]
            fidesops_meta:
              data_type: string
```

**Data to mask**
```
{
  "shipping": null
}
```

A policy targeting `user.name` fields

### Additional context
This was discovered in the Stripe connector. The item returned for the `customer` collection contains a null `shipping` field. If the `shipping` field were populated, Fides would attempt to mask the nested `name` field. In this case the object field is null so we shouldn't generate any values to replace this null field value.

galvana:
Closes #3905

### Description Of Changes

Updating `refine_target_path` to return an empty path if the path is invalid, meaning the data doesn't exist at the level indicated.

### Code Changes

* [x] Added None value scenario to `refine_target_path` function

### Steps to Confirm

* [x] Added new test cases to `test_refine_target_path.py`

### Pre-Merge Checklist

* [x] Update `CHANGELOG.md`

### DIFF
--- a/src/fides/api/task/refine_target_path.py
+++ b/src/fides/api/task/refine_target_path.py
@@ -116,6 +116,10 @@ def refine_target_path(
         next_levels = _enter_array(current_elem, target_path[1:], only)
         return _update_path(current_level, next_levels)
 
+    # stop recursion if current_elem is None
+    if current_elem is None:
+        return []
+
     # Simple case - value is a scalar
     return [current_level] if _match_found(current_elem, only) else []
 
--- a/src/fides/api/task/refine_target_path.py
+++ b/src/fides/api/task/refine_target_path.py
@@ -108,17 +108,13 @@ def refine_target_path(
         logger.warning("Could not locate target path {} on row", target_path)
         return []
 
-    if isinstance(current_elem, dict):
-        next_levels = refine_target_path(current_elem, target_path[1:], only)
-        return _update_path(current_level, next_levels)
-
     if isinstance(current_elem, list):
         next_levels = _enter_array(current_elem, target_path[1:], only)
         return _update_path(current_level, next_levels)
 
-    # stop recursion if current_elem is None
-    if current_elem is None:
-        return []
+    if len(target_path[1:]):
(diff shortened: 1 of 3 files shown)


src/fides/api/task/refine_target_path.py
```diff
@@ -116,6 +116,10 @@ def refine_target_path(
         next_levels = _enter_array(current_elem, target_path[1:], only)
         return _update_path(current_level, next_levels)
 
+    # stop recursion if current_elem is None
+    if current_elem is None:
+        return []
+
```
pattisdr, line 122: This exit condition doesn't feel right - the target path wouldn't be built even if the current element legitimately pointed to None: 

For example, this is what I would expect here in a test:
```
    def test_refine_target_path_with_none_value(self):
        data = {"A": {"B": None}}
        assert refine_target_path(data, ["A", "B"]) == ["A", "B"]
```

pattisdr: early thoughts, I'm wondering instead something like this, that is run after the "is the current_elem a list?" check.
```
 if len(target_path[1:]):
        next_levels = refine_target_path(current_elem, target_path[1:], only)
        return _update_path(current_level, next_levels)
```

pattisdr (commented):
ah great improvement here

pattisdr (approved):

PR #3917, opened 2023-08-11
Forcing focus on integration search input

### LINKED ISSUE #3681: Bug: Integration search focus jumps after two chars have been entered.
To reproduce:
1. Type 3 letters into search
2. Focus will have jumped down to list after 2

https://www.loom.com/share/1a88eaad1c494a8a813f92b3400134eb?sid=c9cabf66-6794-4eb4-9c38-533e552495b2

galvana:
Closes #3681 
### Description Of Changes

Keeping a reference to the integration search input so that the focus can be restored after a re-render.

### Steps to Confirm

* [x] Navigate to the system integration tab and search for an integration, you should be able to type more than 2 characters and still keep focus on the input

### Pre-Merge Checklist

* [x] Update `CHANGELOG.md`

### DIFF
--- a/clients/admin-ui/src/features/datastore-connections/system_portal_config/ConnectionListDropdown.tsx
+++ b/clients/admin-ui/src/features/datastore-connections/system_portal_config/ConnectionListDropdown.tsx
@@ -16,7 +16,7 @@ import {
   Tooltip,
 } from "@fidesui/react";
 import { debounce } from "common/utils";
-import { useCallback, useMemo, useState } from "react";
+import { useCallback, useMemo, useState, useRef } from "react";
 
 import { useAppSelector } from "~/app/hooks";
 import {
@@ -143,6 +143,8 @@ const ConnectionListDropdown: React.FC<SelectDropdownProps> = ({
   onChange,
   selectedValue,
 }) => {
+  const inputRef = useRef(null);
+
   // Hooks
   const [isOpen, setIsOpen] = useState(false);
   const [searchTerm, setSearchTerm] = useState("");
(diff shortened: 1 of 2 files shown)


clients/admin-ui/src/features/datastore-connections/system_portal_config/ConnectionListDropdown.tsx
```diff
@@ -143,6 +143,8 @@ const ConnectionListDropdown: React.FC<SelectDropdownProps> = ({
   onChange,
   selectedValue,
 }) => {
+  const inputRef = useRef(null);
```
TheAndrewJackson, line 146:
```suggestion
  const inputRef = useRef<HTMLInputElement>(null);
```

clients/admin-ui/src/features/datastore-connections/system_portal_config/ConnectionListDropdown.tsx
```diff
@@ -167,7 +169,8 @@ const ConnectionListDropdown: React.FC<SelectDropdownProps> = ({
   const handleSearchChange = useCallback(
     (event: React.ChangeEvent<HTMLInputElement>) => {
       if (event.target.value.length === 0 || event.target.value.length > 1) {
-        setSearchTerm(event.target.value);
+        setSearchTerm(event.target.value.toLowerCase());
+        setTimeout(() => inputRef.current?.focus(), 0);
```
TheAndrewJackson, line 173: I like this fix. It's clever workaround.

I think we're using the menu list component for more than it was originally designed for. That's probably why we're running into the re-rendering issue. For context, here are the docs for the `Menu` component: https://v1.chakra-ui.com/docs/components/overlay/menu


None of those examples have same dynamic behavior that requires being focused on something.

TheAndrewJackson (commented):
Once all the CI checks are passing it should be good to go 👍

clients/admin-ui/src/features/datastore-connections/system_portal_config/ConnectionListDropdown.tsx
galvana: Implemented!

TheAndrewJackson (approved):

PR #3933, opened 2023-08-17
[DANGER] Update to Fideslang 2.0

### LINKED ISSUE #3795: Fideslang taxonomy migrations for version 2.0.0
In order to support the GVL taxonomy as needed for TCF v2.2, we'll need to map all of the GVL purposes to FidesLang data uses. CK, MB, and I did this exercise and, as a result, we need the following data use updates:
https://ethyca.atlassian.net/wiki/spaces/PM/pages/2740027499/Taxonomy+updates#Data-uses%3A

ThomasLaPiana:
Closes #3795

❗ Customer `config.json` files should be updated as they are migrated to the accompanying release.

### Description Of Changes

This PR is focused on bumping the `Fideslang` version and ensuring that the application is working as expected when set u as a new instance. There is a separate PR for handling migrations for current users, as they are both sufficiently complex pieces to warrant handling separately.

Much of the inspiration/work here is inspired by [this PR](https://github.com/ethyca/fides/pull/3343) that handled an earlier, similar change

The Data Migration work is happening here: #4030

_*WARNING: Merging this change as-is will break any existing instance due to default taxonomy model conflicts*_

### Code Changes

* [x] bump `requirements.txt`
* [x] add a migration for the new `version` fields
* Update `DataCategories`
  * [x] datasets (including collections/fields)
  * [x] policy rule **(tested previously, code is copy/pasta)**
  * [x] privacy declaration **(tested previously, code is copy/pasta)**
  * [x] data flows (egress and ingress on `System` models)
* Update `DataUse`s
  * [x] privacy declaration **(tested previously, code is copy/pasta)**
  * [x] policy rule **(tested previously, code is copy/pasta)**
* [x] do find/replace for the old categories/uses

### Steps to Confirm

Because this PR is focused on getting _new_ instances working instead of migrating data, the `Steps to Confirm` here are very simple. Tests passing is what we care about here! [This PR](https://github.com/ethyca/fides/pull/4030) is where the complex data migration will take place

* [ ] after running `nox -s teardown -- volumes` and `nox -s dev`, everything works as intended

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`


ThomasLaPiana:
@adamsachs update for you before I sign off!

In the description I've linked the `fideslang` PR where I'm making changes as needed

I'm currently working through a local dev env and shell to run specific tests and try to diagnose why some of the breaks are happening. Currently trying to figure out why some `422` are getting thrown in some of the tests

src/fides/api/api/v1/endpoints/user_endpoints.py
```diff
@@ -151,7 +151,7 @@ def update_user_password(
     """
     _validate_current_user(user_id, current_user)
 
-    if not current_user.credentials_valid(
+    if not current_user.authorization.credentials_valid(
```
adamsachs, line 154: this looks like a find/replace mistake, going to correct!

src/fides/api/db/seed.py
```diff
@@ -92,7 +92,9 @@ def create_or_update_parent_user() -> None:
         )
 
         if user and CONFIG.security.parent_server_password:
-            if not user.credentials_valid(CONFIG.security.parent_server_password):
+            if not user.authorization.credentials_valid(
```
adamsachs, line 95: same here

src/fides/api/api/v1/endpoints/user_endpoints.py
adamsachs: addressed all of these instances in 18c493c4c4764f59b39fd785544ae95ca27996db

src/fides/api/db/seed.py
```diff
@@ -401,7 +401,7 @@ async def load_default_taxonomy(async_session: AsyncSession) -> None:
             await upsert_resources(
                 sql_model_map[resource_type], resources, async_session
             )
-        except QueryError:  # pragma: no cover
+        except QueryError as e:  # pragma: no cover
```
adamsachs, line 404: were you going to make use of `e` or can we remove? i'm going to remove for now and feel free to re-add, just because it's throwing a pylint error.

adamsachs: removed in 9c4fe119b49793f9919a5e1d1361bc74ce8618e1

adamsachs:
@ThomasLaPiana stepping away for a bit now, made a few changes that i think should help get things greener...tried to break it up reasonably by commit so that my changes are easy to follow -- everything is pretty straightforward cleanup so far.

will see if i can circle back a bit later this evening to knock out some more, but leaving this update here in case i don't 👍

src/fides/api/api/v1/endpoints/user_endpoints.py
ThomasLaPiana: oof, yep. unfortunate

Thank you for the catch and fix!

ThomasLaPiana:
@adamsachs Thank you! picking it up now

ThomasLaPiana:
there's a final system update test failing here that I suspect @pattisdr might understand, other than that the `ctl` tests are looking good!

adamsachs:
> there's a final system update test failing here that I suspect @pattisdr might understand, other than that the `ctl` tests are looking good!

🙌 nice @ThomasLaPiana! i think i wrote those system update tests so i can take a look, i'll loop in @pattisdr if needed.

@ThomasLaPiana are you working on the ops tests or should i also take a look at that? i've got some bandwidth so i'm happy to, but if you're already on it then i'll let you continue

ThomasLaPiana:
@adamsachs Go for it! I'm signing off for now

ThomasLaPiana:
@adamsachs once we get tests passing, we'll merge and ship fideslang 2.0.1 and pin this to a real version

adamsachs:
OK! i think we've got all non-external tests passing now!

@pattisdr would you mind taking a quick peak at the ops tests i fixed in https://github.com/ethyca/fides/pull/3933/commits/8ec6a87331fde0b328cab0045950ec361340c6ce, since some of them are touching consent-related tests and i want to be extra sure i'm not misunderstanding the tests and invalidating them 🙏 

if that looks good, i think we've got our non-external/unsafe tests good to go! the remaining items to get this ready to merge would be:
- [ ] look into external test failures to determine whether these are expected/in `main` already or caused by some code changes here
- [ ] holistic review of this PR (i've focused on getting CI green up to this point)
- [ ] cut `fideslang 2.0.1` per @ThomasLaPiana's note above, and update our dependency here

i should be able to make some progress on the first two items shortly...

pattisdr:
Yes, starting review!

src/fides/api/alembic/migrations/versions/1e750936a141_migrate_to_fideslang_2_0_uses_and_.py
```diff
@@ -0,0 +1,127 @@
...
+    "user.profiling": "user.demographic.profile",  # Verified in 2.0
+    "user.race": "user.demographic.race_ethnicity",  # Verified in 2.0
+    "user.religious_belief": "user.demographic.religious_belief",  # Verified in 2.0
+    "user.sexual_orientation": "user.demographic.sexual_orientation",  # Verified in 2.0
+    "user.financial.account_number": "user.financial.bank_account",  # Verified in 2.0
+    "user.genetic": "user.health_and_medical.genetic",  # Verified in 2.0
+}
+data_category_downgrades: Dict[str, str] = {
+    value: key for key, value in data_use_upgrades.items()
+}
```
pattisdr, line 60: I understand there's an outstanding TODO to handle data categories, but that brings up a similar question for data uses as well - do we need to migrate the existing `ctl_data_uses` and `ctl_data_categories` tables as well? Just trying to understand what's needed.  It's tricky when customers have also started building off of this taxonomy too.

src/fides/api/alembic/migrations/versions/1e750936a141_migrate_to_fideslang_2_0_uses_and_.py
```diff
@@ -0,0 +1,127 @@
...
+down_revision = "708a780b01ba"
+branch_labels = None
+depends_on = None
+
+###############
+## Data Uses ##
+###############
+# The `key` is the old value, the `value` is the new value
+data_use_upgrades: Dict[str, str] = {
+    "improve.system": "functional.service.improve",  # Verified in 2.0
```
pattisdr, line 28: what if a customer had branched their taxonomy off of "improve.system," and they now have "improve.system.maintenance" and added that to a privacy declaration.  We didn't handle this sort of thing last time, is this just handled on a customer by customer basis?

src/fides/api/alembic/migrations/versions/1e750936a141_migrate_to_fideslang_2_0_uses_and_.py
pattisdr: okay! Adam pointed me towards `load_default_taxonomy` which we already run on startup! I forgot about that. Brought fides down brought it back, verified the new ones were loaded, the old ones were left untouched. Nice to see that new `version_added` with `2.0.0` populated too, that was a good idea.

src/fides/api/alembic/migrations/versions/1e750936a141_migrate_to_fideslang_2_0_uses_and_.py
pattisdr: OK learning more, since we don't remove the old uses from the db, it's fine if their privacy declarations continue to reference their custom uses that are based off of data uses we've technically deprecated.  I assume.

The custom taxonomies are an interesting use case.

pattisdr (commented):
ops consent test changes look fine to me, thanks Adam!

src/fides/api/alembic/migrations/versions/1e750936a141_migrate_to_fideslang_2_0_uses_and_.py
```diff
@@ -0,0 +1,127 @@
...
+    "user.date_of_birth": "user.demographic.date_of_birth",  # Verified in 2.0
+    "user.gender": "user.demographic.gender",  # Verified in 2.0
+    "user.political_opinion": "user.demographic.political_opinion",  # Verified in 2.0
+    "user.profiling": "user.demographic.profile",  # Verified in 2.0
+    "user.race": "user.demographic.race_ethnicity",  # Verified in 2.0
+    "user.religious_belief": "user.demographic.religious_belief",  # Verified in 2.0
+    "user.sexual_orientation": "user.demographic.sexual_orientation",  # Verified in 2.0
+    "user.financial.account_number": "user.financial.bank_account",  # Verified in 2.0
+    "user.genetic": "user.health_and_medical.genetic",  # Verified in 2.0
+}
```
pattisdr, line 57: Oh interesting the pending data category migration, data categories are on privacy declarations, datasets at the dataset, collection, and field level...anywhere else?

tests/ctl/api/test_seed.py
```diff
@@ -86,17 +86,17 @@ def test_filter_data_categories_excluded(self) -> None:
         excluded_data_categories = [
             "user.financial",
             "user.payment",
-            "user.credentials",
+            "user.authorization",
```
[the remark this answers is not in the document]
ThomasLaPiana: technically it is `user.authorization.credentials` but apparently this feature doesn't work with anything past the 2nd level, so all of these tests were failing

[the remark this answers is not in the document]
ThomasLaPiana: maybe this is something to be worried about...if we're expecting it to work with that _specific_ category. Otherwise, should be fine

ThomasLaPiana:
@adamsachs the major final step here is updating the categories in the data migration. I only did the uses because I could copy/pasta the logic from a previous migration, but now I also need to write them for data categories which live in a lot more places

src/fides/api/alembic/migrations/versions/1e750936a141_migrate_to_fideslang_2_0_uses_and_.py
ThomasLaPiana: I _think_ that's it but I'm going to look at the models to confirm

src/fides/api/alembic/migrations/versions/1e750936a141_migrate_to_fideslang_2_0_uses_and_.py
ThomasLaPiana: Correct, these data migrations only care about "renames", not removals/additions

src/fides/api/alembic/migrations/versions/1e750936a141_migrate_to_fideslang_2_0_uses_and_.py
ThomasLaPiana: yep, those tables should be good! That's what we have the most control over so we made it easier on ourselves

src/fides/api/alembic/migrations/versions/1e750936a141_migrate_to_fideslang_2_0_uses_and_.py
```diff
@@ -0,0 +1,241 @@
...
+        bind.execute(
+            update_label_query,
+            {"dataset_id": row["id"], "updated_labels": updated_labels},
+        )
+
+        # Update the collections objects
+        collections = row["collections"]
+
+        for key, value in data_label_map.items():
+            collections.replace(key, value)
```
adamsachs, line 146: do you feel comfortable with a blind string replace here? hard to imagine the data category strings being used other than as `data_categories` on the collection object or its child fields, but i wonder if there's anything quick we could do to make this a bit more targeted/safer? i'll defer to your judgement though.

if we do keep it as-is, then may be good to at least put a comment here explaining the assumptions we're making in this operation.

src/fides/api/alembic/migrations/versions/1e750936a141_migrate_to_fideslang_2_0_uses_and_.py
```diff
@@ -0,0 +1,241 @@
...
+
+        update_collections_query: TextClause = text(
+            "UPDATE datasets SET collections = :updated_labels WHERE id= :dataset_id"
+        )
+        bind.execute(
+            update_collections_query,
+            {"dataset_id": row["id"], "updated_labels": collections},
```
adamsachs, line 153: nit: `updated_labels` feels like a weird name for this var? if just a c+p for now and you're in the process of cleaning up, please ignore me

src/fides/api/alembic/migrations/versions/1e750936a141_migrate_to_fideslang_2_0_uses_and_.py
ThomasLaPiana: I plan on doing more testing to validate here

tests/ctl/api/test_seed.py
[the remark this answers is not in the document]
adamsachs: well wait a sec, isn't the update here specifically updating the invocation within `seed.py` to have `user.authorization.credentials` in the passed arg? https://github.com/ethyca/fides/pull/3933/files#diff-2b86b1162fde67daa1075cd9c2a1b10da9a6bd1a5a235cbd588508536dcec83aR330

do we want to update that invocation to just use `user.authorization`, if it doesn't work with categories past the second level?

i may be missing something though...

src/fides/api/db/seed.py
```diff
@@ -327,7 +327,7 @@ def load_default_dsr_policies() -> None:
         excluded_data_categories = [
             "user.financial",
             "user.payment",
-            "user.credentials",
+            "user.authorization.credentials",
```
adamsachs, line 330: having a bit of trouble reconciling this update with what i understand [this comment](https://github.com/ethyca/fides/pull/3933/files#r1299152122) to mean - just would like to confirm this is going to work as we expect

tests/ctl/core/test_api.py
```diff
@@ -1715,7 +1715,7 @@ def test_system_update_updates_declarations(
         "update_declarations",
         [
             (
-                [  # update a dec matching one existing dec
+                [  # Check 1: update a dec matching one existing dec
```
adamsachs, line 1718: nit/just a thought - the pytest output actually starts the parameterization at `0`, it _could_ be a bit easier to line this up with testing output if we also started our indexing at `0` in the comments here? it just tripped me up for a quick second while i was debugging one of the failing parameterizations 😄

tests/ops/util/test_consent_util.py
```diff
@@ -1196,7 +1196,6 @@ def custom_data_use(self, db):
                 name="New data use",
                 description="A test data use",
                 parent_key=None,
-                is_default=True,
```
adamsachs, line 1199: noting here for posterity: i removed this because it didn't seem intended and was causing problems 👍

tests/ctl/core/test_api.py
ThomasLaPiana: rookie mistake on my part! You're totally right

We could also pass an `id` to pytest to make the parametrized run more clear

ThomasLaPiana:
@adamsachs I think I've just about gotten this into a place where we can start testing it, but it's definitely tricky

src/fides/api/alembic/migrations/versions/1e750936a141_migrate_to_fideslang_2_0_uses_and_.py
adamsachs: ok, that sounds good 👍 

but just to make sure my concern is understood: some of the risk here seems to be based on whether we think there's any legitimate chance that users have put one of the old data category strings within the strings for the `name` property on their `collection`s or `field`s -- i.e, in that case, we'd accidentally update the `name`s of those resources, which i don't think would be an intended affect and could potentially lead to problems. i don't think any testing would really mitigate this risk?

that being said, i realize this is contrived, and it seems unlikely any current users would actually hit this - so i'm not saying that we _need_ to handle this case. i just want to make sure we're on the same page in recognizing this!

adamsachs (commented):
@ThomasLaPiana took a fuller look at this, and generally it's looking good to me!

left some nits/thoughts sprinkled in, but the only thing that i really wanna double down on is re: the `filter_data_categories` [invocation](https://github.com/ethyca/fides/pull/3933/files#r1300752656) - just having trouble reconciling the change in `seed.py` with your comments in the test.

i know you're still wrapping up the last bits of data category migration - the only other thing i'd like to be sure on is some of the external test failures. i can try digging on that a bit if we haven't already 👍

tests/ctl/core/test_api.py
adamsachs: oooh cool, that's a nice protip...

ThomasLaPiana:
working on testing steps now

src/fides/api/alembic/migrations/versions/1e750936a141_migrate_to_fideslang_2_0_uses_and_.py
ThomasLaPiana: Ah, I see your point. Given that the name field is purely to make things human-readable, I think at most we would inconvenience people in this case? Maybe matching on something more specific like `fides_key: {}` would be more accurate? Worth testing for sure

ThomasLaPiana:
this is almost getting ready for merge, I'll polish the steps to test and assign some reviewers

pattisdr:
Starting review..

src/fides/api/alembic/migrations/versions/1e750936a141_migrate_to_fideslang_2_0_uses_and_.py
```diff
@@ -0,0 +1,289 @@
...
+        update_label_query: TextClause = text(
+            "UPDATE privacydeclaration SET data_categories = :updated_label WHERE id= :declaration_id"
+        )
+        bind.execute(
+            update_label_query,
+            {"declaration_id": row["id"], "updated_label": labels},
+        )
+
+
+def update_ctl_policy_data_categories(
```
pattisdr, line 220: There's a new field `PrivacyDeclaration.shared_categories` going out in the same release that is also a data category field- I assume only staging environments would have a potentially old value populated, so maybe not worth migrating but just pointing this out.

src/fides/api/alembic/migrations/versions/1e750936a141_migrate_to_fideslang_2_0_uses_and_.py
```diff
@@ -0,0 +1,289 @@
...
+    "user.political_opinion": "user.demographic.political_opinion",  # Verified in 2.0
+    "user.profiling": "user.demographic.profile",  # Verified in 2.0
+    "user.race": "user.demographic.race_ethnicity",  # Verified in 2.0
+    "user.religious_belief": "user.demographic.religious_belief",  # Verified in 2.0
+    "user.sexual_orientation": "user.demographic.sexual_orientation",  # Verified in 2.0
+    "user.financial.account_number": "user.financial.bank_account",  # Verified in 2.0
+    "user.genetic": "user.health_and_medical.genetic",  # Verified in 2.0
+}
+data_category_downgrades: Dict[str, str] = {
+    value: key for key, value in data_use_upgrades.items()
```
pattisdr, line 107: copy/paste data category downgrades aren't working, should be data_category_upgrades

src/fides/api/alembic/migrations/versions/1e750936a141_migrate_to_fideslang_2_0_uses_and_.py
```diff
@@ -0,0 +1,289 @@
...
+            bind.execute(
+                update_label_query,
+                {"dataset_id": row["id"], "updated_labels": updated_labels},
+            )
+
+        # Update the collections objects
+        collections: str = json.dumps(row["collections"])
+
+        for key, value in data_label_map.items():
+            collections = collections.replace(key, value)
```
pattisdr, line 142: This string replace can cause unexpected behavior.  For example, say an old data category is `user.browsing_history`. I upgrade, and that becomes `user.behavior.browsing_history`.  I downgrade and that becomes the incorrect `user.observed.browsing_history`, because `user.behavior` was replaced with `user.observed`.  Suggestions are changing the order of the data categories dictionary - more specific first and more broad second, or to convert to a dictionary and do more targeted upgrades instead of string replace

src/fides/api/alembic/migrations/versions/1e750936a141_migrate_to_fideslang_2_0_uses_and_.py
```diff
@@ -0,0 +1,289 @@
...
+    update_privacy_declaration_data_uses(bind, data_use_upgrades)
+
+    logger.info("Upgrading ctl policy rule data uses")
+    update_ctl_policy_data_uses(bind, data_use_upgrades)
+
+    logger.info("Upgrading data category on privacy declaration")
+    update_privacy_declaration_data_categories(bind, data_category_upgrades)
+
+    logger.info("Upgrading ctl policy rule data categories")
+    update_ctl_policy_data_uses(bind, data_category_upgrades)
```
pattisdr, line 261: Needs to be `update_ctl_policy_data_categories` instead

src/fides/api/alembic/migrations/versions/1e750936a141_migrate_to_fideslang_2_0_uses_and_.py
```diff
@@ -0,0 +1,289 @@
...
+    update_privacy_declaration_data_uses(bind, data_use_downgrades)
+
+    logger.info("Downgrading ctl policy rule data uses")
+    update_ctl_policy_data_uses(bind, data_use_downgrades)
+
+    logger.info("Downgrading data category on privacy declaration")
+    update_privacy_declaration_data_uses(bind, data_category_downgrades)
+
+    logger.info("Downgrading ctl policy rule data categories")
+    update_ctl_policy_data_uses(bind, data_category_downgrades)
```
pattisdr, line 283: Assume update_ctl_policy_data_categories needed instead

src/fides/api/alembic/migrations/versions/1e750936a141_migrate_to_fideslang_2_0_uses_and_.py
```diff
@@ -0,0 +1,289 @@
...
+    bind: Connection = op.get_bind()
+
+    logger.info("Downgrading data use on privacy declaration")
+    update_privacy_declaration_data_uses(bind, data_use_downgrades)
+
+    logger.info("Downgrading ctl policy rule data uses")
+    update_ctl_policy_data_uses(bind, data_use_downgrades)
+
+    logger.info("Downgrading data category on privacy declaration")
+    update_privacy_declaration_data_uses(bind, data_category_downgrades)
```
pattisdr, line 280: Assume update_privacy_declaration_data_categories needed instead

src/fides/api/alembic/migrations/versions/1e750936a141_migrate_to_fideslang_2_0_uses_and_.py
ThomasLaPiana: code suggestion 🙏

src/fides/api/alembic/migrations/versions/1e750936a141_migrate_to_fideslang_2_0_uses_and_.py
ThomasLaPiana: I'm not really worried about the downgrade case...would it be better to remove the downgrades altogether then to prevent it from causing bugs? Great catch here btw, I didn't think about the downgrade case

src/fides/api/alembic/migrations/versions/1e750936a141_migrate_to_fideslang_2_0_uses_and_.py
ThomasLaPiana: thank you for calling it out!

src/fides/api/alembic/migrations/versions/1e750936a141_migrate_to_fideslang_2_0_uses_and_.py
pattisdr, line 107:
```suggestion
    value: key for key, value in data_category_upgrades.items()
```

pattisdr: OK still needs to be changed though!

src/fides/api/alembic/migrations/versions/1e750936a141_migrate_to_fideslang_2_0_uses_and_.py
pattisdr: Can we just move `"user.observed": "user.behavior",  # Verified in 2.0` to the last entry in the dictionary - looking closer, that's probably the only one? I think we need to get better at supporting downgrades generally

src/fides/api/alembic/migrations/versions/1e750936a141_migrate_to_fideslang_2_0_uses_and_.py
```diff
@@ -0,0 +1,289 @@
...
+        bind.execute(
+            update_label_query,
+            {"declaration_id": row["id"], "updated_label": labels},
+        )
+
+
+def update_ctl_policy_data_categories(
+    bind: Connection, data_label_map: Dict[str, str]
+) -> None:
+    """Upgrade or downgrade data uses from fideslang 1.4 for ctl policies"""
```
pattisdr, line 223:
```suggestion
    """Upgrade or downgrade data uses from fideslang 2.0 for ctl policies"""
```

src/fides/api/alembic/migrations/versions/1e750936a141_migrate_to_fideslang_2_0_uses_and_.py
```diff
@@ -0,0 +1,289 @@
...
+}
+data_use_downgrades: Dict[str, str] = {
+    value: key for key, value in data_use_upgrades.items()
+}
+
+
+def update_privacy_declaration_data_uses(
+    bind: Connection, data_use_map: Dict[str, str]
+) -> None:
+    """Upgrade or downgrade data uses from fideslang 1.4 for privacy declarations"""
```
pattisdr, line 40:
```suggestion
    """Upgrade or downgrade data uses from fideslang 2.0 for privacy declarations"""
```

src/fides/api/alembic/migrations/versions/1e750936a141_migrate_to_fideslang_2_0_uses_and_.py
```diff
@@ -0,0 +1,289 @@
...
+                "UPDATE privacydeclaration SET data_use = :updated_use WHERE id= :declaration_id"
+            )
+            bind.execute(
+                update_data_use_query,
+                {"declaration_id": row["id"], "updated_use": updated_data_use},
+            )
+
+
+def update_ctl_policy_data_uses(bind: Connection, data_use_map: Dict[str, str]) -> None:
+    """Upgrade or downgrade data uses from fideslang 1.4 for ctl policies"""
```
pattisdr, line 59:
```suggestion
    """Upgrade or downgrade data uses from fideslang 2.0 for ctl policies"""
```

src/fides/api/alembic/migrations/versions/1e750936a141_migrate_to_fideslang_2_0_uses_and_.py
```diff
@@ -0,0 +1,289 @@
...
+    logger.info("Downgrading data category on privacy declaration")
+    update_privacy_declaration_data_categories(bind, data_category_downgrades)
+
+    logger.info("Downgrading ctl policy rule data categories")
+    update_ctl_policy_data_categories(bind, data_category_downgrades)
+
+    logger.info("Downgrading data categories in datasets")
+    update_datasets_data_categories(bind, data_category_downgrades)
+
+    logger.info("Downgrading the System egress/ingress data cateogries")
```
pattisdr, line 288:
```suggestion
    logger.info("Downgrading the System egress/ingress data categories")
```

src/fides/api/alembic/migrations/versions/1e750936a141_migrate_to_fideslang_2_0_uses_and_.py
```diff
@@ -0,0 +1,289 @@
...
+    logger.info("Upgrading data category on privacy declaration")
+    update_privacy_declaration_data_categories(bind, data_category_upgrades)
+
+    logger.info("Upgrading ctl policy rule data categories")
+    update_ctl_policy_data_categories(bind, data_category_upgrades)
+
+    logger.info("Upgrading data categories in datasets")
+    update_datasets_data_categories(bind, data_category_upgrades)
+
+    logger.info("Upgrading the System egress/ingress data cateogries")
```
pattisdr, line 266:
```suggestion
    logger.info("Upgrading the System egress/ingress data categories")
```

src/fides/api/alembic/migrations/versions/1e750936a141_migrate_to_fideslang_2_0_uses_and_.py
```diff
@@ -0,0 +1,289 @@
...
+data_use_downgrades: Dict[str, str] = {
+    value: key for key, value in data_use_upgrades.items()
+}
+
+
+def update_privacy_declaration_data_uses(
+    bind: Connection, data_use_map: Dict[str, str]
+) -> None:
+    """Upgrade or downgrade data uses from fideslang 1.4 for privacy declarations"""
+    existing_ctl_policies: ResultProxy = bind.execute(
```
pattisdr, line 41: I'd change the variable name here -

src/fides/api/alembic/migrations/versions/1e750936a141_migrate_to_fideslang_2_0_uses_and_.py
pattisdr: Actually, can we back up and just make sure that this upgrade shouldn't be more targeted?

For customers that have built off of the taxonomy, say customer is using custom data category: `user.search_history.apples`.  Because we're doing string replacement, not exact matches in our migration,  this becomes, `user.behavior.search_history.apples`.  It is not clear to me that this is expected from a customer perspective.  @adamsachs

ThomasLaPiana: With regards to blunt string replacement...

This is desired behavior according to @SteveDMurphy , as we don't want someone with an extended taxonomy to suddenly have orphaned elements

pattisdr: A couple follow-up questions then, with an understanding that folks have ended up with orphaned elements in the past: 1) why are there orphaned elements if we don't touch the taxonomy itself? I was under the impression that the seed script added to DataCategory and DataUse and not updated. 2). On the other hand, if "blunt string replacement" is in fact desirable, would we need more updates to this migration? Broader string replacement in more locations than just collections on datasets.

ThomasLaPiana: 1) The default taxonomy is completely truncated and re-seeded every time. This causes the orphaning
2) That's a great point, let me look here and convert to `replace` instead of dict lookups where it makes sense

ThomasLaPiana:
I ran through the `Steps to Confirm` with the latest code and it all looks good, ready for another review

src/fides/api/alembic/migrations/versions/1e750936a141_migrate_to_fideslang_2_0_uses_and_.py
ThomasLaPiana: @pattisdr I went through and updated the logic for find/replaces

src/fides/api/db/seed.py
adamsachs: @ThomasLaPiana just want to confirm here that we're OK having the default policy include the second-level `user.authorization`, because that's effectively what we're going to end up with here. and then i'm pretty sure this means the default policy will, in effect, also target `user.authorization.credentials`, since we include the children of data category targets when filtering DSR results, as far as i can tell ([see `filter_results.filter_data_categories`](https://github.com/ethyca/fides/blob/main/src/fides/api/task/filter_results.py#L65)). so this seems undesirable.

my suggestion is that we either update this to exclude `user.authorization` (simpler update, but not sure if it's exactly the behavior we're looking for) or we update the logic of `filter_data_categories` here in `seed.py` to change the second-level logic (more complex update that i think would deserve some more thought/discussion). simple code suggestion below for the former:
```suggestion
            "user.authorization",
```

src/fides/api/alembic/migrations/versions/1e750936a141_migrate_to_fideslang_2_0_uses_and_.py
adamsachs: i'll be taking a closer look at all of this locally with some manual testing, but just wanted to ask about this in case you see it before i get to that:

> The default taxonomy is completely truncated and re-seeded every time

from a quick look at `seed.py`, i don't see where that truncation is done? i'll try this out locally and debug further to confirm, but all i see are upserts of the new taxonomy elements, which presumably do not _remove_ any of the outdated taxonomy elements?

src/fides/api/alembic/migrations/versions/1e750936a141_migrate_to_fideslang_2_0_uses_and_.py
```diff
@@ -230,10 +247,11 @@ def update_ctl_policy_data_categories(
         for i, rule in enumerate(rules or []):
             data_labels: Dict = rule.get("data_categories", {})
             for j, val in enumerate(data_labels.get("values", [])):
-                new_data_label: Optional[str] = data_label_map.get(val, None)
-                if new_data_label:
-                    rules[i]["data_categories"]["values"][j] = new_data_label
-                    needs_update = True
```
pattisdr: Looks like you dropped the `needs_update=True` with this change, so data category updates to ctl_policies will no longer work

src/fides/api/alembic/migrations/versions/1e750936a141_migrate_to_fideslang_2_0_uses_and_.py
```diff
@@ -0,0 +1,307 @@
...
+        egress = row["egress"]
+
+        # Do a blunt find/replace
+        if ingress:
+            for item in ingress:
+                item["data_categories"] = [
+                    label.replace(key, value)
+                    for key, value in data_label_map.items()
+                    for label in item.get("data_categories", [])
+                    if key in label
```
pattisdr, line 182: same here

src/fides/api/alembic/migrations/versions/1e750936a141_migrate_to_fideslang_2_0_uses_and_.py
```diff
@@ -0,0 +1,307 @@
...
+## Data Categories ##
+#####################
+"""
+The `key` is the old value, the `value` is the new value
+These are ordered specifically so that string replacement works on both parent and child items
+"""
+data_category_upgrades: Dict[str, str] = {
+    "user.financial.account_number": "user.financial.bank_account",
+    "user.credentials": "user.authorization.credentials",
+    "user.observed": "user.behavior",
```
pattisdr, line 97: Re-surfacing an earlier comment, I'd really like to see this `user.observed` moved to the last entry in this dictionary

pattisdr (changes requested):
This latest round of changes introduced some bugs: https://github.com/ethyca/fides/pull/3933#discussion_r1306872795

but I also think we haven't yet fully settled this is the path we should go

src/fides/api/alembic/migrations/versions/1e750936a141_migrate_to_fideslang_2_0_uses_and_.py
pattisdr: good catch Adam, it does seem like old taxonomy elements are not actually being removed.  So we need to get on the same page regarding whether they _should_ be removed here and if so, make that happen.

src/fides/api/db/seed.py
ThomasLaPiana: I think we want `user.authorization.credentials` to match what we had before, but I primarily made this change with find/replace, I'm not sure what broader implications might be

src/fides/api/alembic/migrations/versions/1e750936a141_migrate_to_fideslang_2_0_uses_and_.py
ThomasLaPiana: I remembered that we were dropping taxonomy elements but it seems I was mistaken/we no longer are...

If we don't drop the existing taxonomy elements, we'll need to update them to all show `is_default=False`. This should fix the `version` validation issue.

src/fides/api/alembic/migrations/versions/1e750936a141_migrate_to_fideslang_2_0_uses_and_.py
```diff
@@ -0,0 +1,307 @@
...
+    """Upgrade or downgrade data uses from fideslang 2.0 for data categories"""
+    existing_privacy_declarations: ResultProxy = bind.execute(
+        text("SELECT id, data_categories FROM privacydeclaration;")
+    )
+    for row in existing_privacy_declarations:
+        labels = [
+            label.replace(key, value)
+            for key, value in data_label_map.items()
+            for label in row["data_categories"]
+            if key in label
```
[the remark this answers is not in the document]
ThomasLaPiana: Ah, you're totally right here, good catch!

src/fides/api/alembic/migrations/versions/1e750936a141_migrate_to_fideslang_2_0_uses_and_.py
ThomasLaPiana: I dropped it but set it to update all regardless

src/fides/api/alembic/migrations/versions/1e750936a141_migrate_to_fideslang_2_0_uses_and_.py
ThomasLaPiana: Done!

src/fides/api/db/seed.py
adamsachs: OK, but the update here effectively makes this line a no-op -- `user.authorization.credentials` isn't included in the default policy regardless, because it is 3 levels deep, and we explicitly only include categories [that are `< 3` levels within `seed.filter_data_categories`](https://github.com/ethyca/fides/blob/ThomasLaPiana-update-fideslang-200/src/fides/api/db/seed.py#L128).

but more than just being a no-op, we're actually going to construct a default access policy here that _does_ include `user.authorization` as a target data category - i've confirmed that locally by looking at the default policy that's created. by including `user.authorization` as a target data category, we also effectively target `user.authorization.credentials` in that policy, so we're effectively changing the behavior here to now _include_ user credentials data in our default policy. i don't think that change in behavior is intended.

my suggestion is to _not_ include `user.authorization` as a target data category in the default access policy. this will effectively keep the existing behavior of _not_ including user credentials data in our default access policy. the only additional side effect is that we'd also be excluding any _other_ `user.authorization` data from the access policy, i.e. any fields tagged broadly with `user.authorization` or with its other subcategories, `user.authorization.biometric`, `user.authorization.password`. IMO those side effects are OK, and worth keeping in-tact the behavior to not include user credentials in our default access policy (though that judgement is a bit subjective).

if there aren't any objections to what i said, then i'll go ahead and commit this update!

src/fides/api/alembic/migrations/versions/1e750936a141_migrate_to_fideslang_2_0_uses_and_.py
pattisdr: that's not true as of your latest commit, it is still using a `needs_update` flag -

data/dataset/bigquery_example_test_dataset.yml
```diff
@@ -136,7 +136,7 @@ dataset:
                   field: address.id
                   direction: to
           - name: ccn
-            data_categories: [user.financial.account_number]
+            data_categories: [user.financial.bank_account]
```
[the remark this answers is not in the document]
pattisdr: Adrian says we do need a SaaS config version bump to update the datasets for the existing connection configs

adamsachs:
sorry for breaking things a bit with merging https://github.com/ethyca/fides/commit/b07cf87cfb95185aed7815597f7e42a6a487dfad into `main`. i can update this branch to resolve the issues, i think we'll just want to:
- [ ] cut a new `2.0.2` release of `fideslang` that includes the same thing as `1.4.5` but is made on top of `2.0.1`, and then update `requirements.txt` here to point to `fideslang==2.0.2`
- [ ] update the `down_revision` of the [first migration to point to `3038667ba898`](https://github.com/ethyca/fides/pull/3933/files#diff-39abd1f52f655ead64e74fd79e75c4b1f224ee343e5160a1d8aabfc4c7ced2fcR13)

src/fides/api/db/seed.py
pattisdr: ah, ok, I agree with you @adamsachs I think we should exclude the higher-level `user.authorization`, which seems to be in line with the intent of that `excluded_data_categories` section, meaning any data category or nested data category in that list will _not_ be returned in the DSR package and _not_ erased by default.

Follow up then, it seems like we need to include a migration of existing RuleTargets on the default access/erasure policies.

adamsachs: oh good point, i'd think just updating _all_ `RuleTargets` would probably be most consistent with the migration/upgrade behavior we're going for here...

src/fides/api/alembic/migrations/versions/1e750936a141_migrate_to_fideslang_2_0_uses_and_.py
adamsachs: OK we reached some general consensus in [slack](https://ethyca.slack.com/archives/CBKK8TS74/p1693323605019939), it seems like we do want to:
- remove the deprecated taxonomy elements (probably should be done here within the migration itself?)
- update _all_ references to those taxonomy elements in-place (i.e. the blunt string replace)
- give some steps for a "fallback" plan upon upgrade if desired, i.e. don't automigrate, take a db backup and migrate manually

src/fides/api/alembic/migrations/versions/1e750936a141_migrate_to_fideslang_2_0_uses_and_.py
ThomasLaPiana: fixed

src/fides/api/alembic/migrations/versions/1e750936a141_migrate_to_fideslang_2_0_uses_and_.py
```diff
@@ -0,0 +1,307 @@
...
+#####################
+## Data Categories ##
+#####################
+"""
+The `key` is the old value, the `value` is the new value
+These are ordered specifically so that string replacement works on both parent and child items
+"""
+data_category_upgrades: Dict[str, str] = {
+    "user.financial.account_number": "user.financial.bank_account",
+    "user.credentials": "user.authorization.credentials",
```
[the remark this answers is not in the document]
ThomasLaPiana: fixed

ThomasLaPiana:
@allisonking do we need to find/replace the typescript references for these now-outdated data categories and uses? Or is there a script that can/will generate them?

ThomasLaPiana:
reposting for visibility: https://github.com/ethyca/fides/pull/3933#discussion_r1309340895

OK we reached some general consensus in [slack](https://ethyca.slack.com/archives/CBKK8TS74/p1693323605019939), it seems like we do want to:

* remove the deprecated taxonomy elements (probably should be done here within the migration itself?)
* update all references to those taxonomy elements in-place (i.e. the blunt string replace)
* give some steps for a "fallback" plan upon upgrade if desired, i.e. don't automigrate, take a db backup and migrate manually

ThomasLaPiana:
@adamsachs @pattisdr sorry for the never-ending back and forth here! I made changes according to what was agreed upon

allisonking:
> @allisonking do we need to find/replace the typescript references for these now-outdated data categories and uses? Or is there a script that can/will generate them?

I don't think there are any references to specific data categories in the code other than in the tests, which it looks like you've already updated! The TS models are generally updated by running `npm run openapi:generate` in `clients/admin-ui` against fidesplus. If that doesn't have any changes, then we should be good to go 👍 (but also it probably will have _some_ changes, as it's probably been a little bit since we last ran it. you can just see if there are any changes directly related to this work)

src/fides/api/alembic/migrations/versions/1e750936a141_migrate_to_fideslang_2_0_uses_and_.py
```diff
@@ -0,0 +1,308 @@
...
+    """Upgrade or downgrade data uses from fideslang 2.0 for data categories"""
+    existing_privacy_declarations: ResultProxy = bind.execute(
+        text("SELECT id, data_categories FROM privacydeclaration;")
+    )
+    for row in existing_privacy_declarations:
+        labels = [
+            label.replace(key, value) if key in label else label
+            for key, value in data_label_map.items()
+            for label in row["data_categories"]
+        ]
```
pattisdr, line 222: This is not formulating the replacement categories correctly, 

if row["data_categories"] = `['user', 'user.financial.account_number']`

labels becomes: 
`
['user', 'user.financial.bank_account', 'user', 'user.financial.account_number', 'user', 'user.financial.account_number', 'user', 'user.financial.account_number', 'user', 'user.financial.account_number', 'user', 'user.financial.account_number', 'user', 'user.financial.account_number', 'user', 'user.financial.account_number', 'user', 'user.financial.account_number', 'user', 'user.financial.account_number', 'user', 'user.financial.account_number', 'user', 'user.financial.account_number', 'user', 'user.financial.account_number', 'user', 'user.financial.account_number', 'user', 'user.financial.account_number', 'user', 'user.financial.account_number']
`

I'll look into rewriting these -

src/fides/api/alembic/migrations/versions/1e750936a141_migrate_to_fideslang_2_0_uses_and_.py
```diff
@@ -0,0 +1,308 @@
...
+    for row in existing_ctl_policies:
+        rules: List[Dict] = row["rules"]
+        for i, rule in enumerate(rules or []):
+            data_labels: Dict = rule.get("data_categories", {})
+            for j, val in enumerate(data_labels.get("values", [])):
+                data_category: str = val
+                for key, value in data_label_map.items():
+                    if key in data_category:
+                        data_category = data_category.replace(key, value)
+                rules[i]["data_uses"]["values"][j] = data_category
```
pattisdr, line 249: ack this replaces matching data use values with data categories!

src/fides/api/alembic/migrations/versions/1e750936a141_migrate_to_fideslang_2_0_uses_and_.py
```diff
@@ -0,0 +1,308 @@
...
+        # Update data categories at the top level
+        labels: Optional[List[str]] = row["data_categories"]
+
+        if labels:
+            # Do a string replace here to catch child items
+            updated_labels: List[str] = [
+                label.replace(key, value) if label.startswith(key) else label
+                for key, value in data_label_map.items()
+                for label in labels
+            ]
```
pattisdr, line 137: ok yeah these sorts of things are causing my migration to crash -fixing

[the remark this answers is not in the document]
pattisdr: here we're using startswith, and for another similar change we're using "in" - I assume I can make both startswith

src/fides/api/alembic/migrations/versions/1e750936a141_migrate_to_fideslang_2_0_uses_and_.py
pattisdr: addressed here https://github.com/ethyca/fides/pull/3933/commits/9308569ad9e9342d9e43256f65a3f9bd266c20cd

src/fides/api/alembic/migrations/versions/1e750936a141_migrate_to_fideslang_2_0_uses_and_.py
pattisdr: addressed https://github.com/ethyca/fides/pull/3933/commits/9308569ad9e9342d9e43256f65a3f9bd266c20cd

src/fides/api/alembic/migrations/versions/1e750936a141_migrate_to_fideslang_2_0_uses_and_.py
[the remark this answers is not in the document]
pattisdr: https://github.com/ethyca/fides/pull/3933/commits/9308569ad9e9342d9e43256f65a3f9bd266c20cd

src/fides/api/alembic/migrations/versions/1e750936a141_migrate_to_fideslang_2_0_uses_and_.py
```diff
@@ -0,0 +1,322 @@
...
+def upgrade() -> None:
+    bind: Connection = op.get_bind()
+
+    # Given that our advice is to turn off auto-migrations and make a db copy,
+    # there is no "downgrade" version of this. It also wouldn't be feasible given
+    # it would require an older version of fideslang.
+    logger.info("Removing old default data categories and data uses")
+    bind.execute(text("DELETE FROM ctl_data_uses WHERE is_default = TRUE;"))
+    bind.execute(text("DELETE FROM ctl_data_categories WHERE is_default = TRUE;"))
+
```
pattisdr, line 283: So we have them upgrade, then they have to restart fides again to get the seed script to run, are those the right steps?

pattisdr:
Thanks for your changes @ThomasLaPiana.  The migration was crashing so I added some changes here: https://github.com/ethyca/fides/pull/3933/commits/9308569ad9e9342d9e43256f65a3f9bd266c20cd to move this forward, but there's still more work to do.  Let's all continue to test and figure out where the gaps are.

Here's some current TODO items I'm seeing:

- [ ] Migrate custom data uses and custom data categories that build off of the affected 2.0 uses and categories.   
    -  For example, say I have a custom data use, "user.search_history.work" that I am using on a PrivacyDeclaration. We migrate that to "user.behavior.search_history.work" on the Declaration.  However, there is no "DataUse" that matches, and you get errors later when trying to update the System as a result. 

- [ ] New column `PrivacyDeclaration.shared_categories` going out in the current release needs to be migrated now that this migration has been delayed for a followup release.  Resurfacing [earlier comment](https://github.com/ethyca/fides/pull/3933#discussion_r1305942158)

- [ ] `PrivacyNotice`, `PrivacyNoticeTemplate`, and `PrivacyNoticeHistory` `data_uses` columns likely need to be migrated.  None of our out of the box notices have data uses at risk, but customers could have added custom notices, or added data uses to existing notices. 

- [ ] Mysterious error on data subjects and sometimes data uses endpoint after running these migrations, usually where we have a data use mismatch somewhere.
```
2023-08-31 17:53:52 2023-08-31 22:53:52.876 | INFO     | fides.api.main:log_request:150 - Request received | {'method': 'GET', 'status_code': 200, 'handler_time': '25.242ms', 'path': '/api/v1/data_category/'}
2023-08-31 17:53:52 2023-08-31 22:53:52.878 | ERROR    | fides.api.util.logger:_log_exception:31 - unhandled errors in a TaskGroup (1 sub-exception)
2023-08-31 17:53:52 Traceback (most recent call last):
2023-08-31 17:53:52   File "/opt/fides/lib/python3.10/site-packages/anyio/streams/memory.py", line 97, in receive
2023-08-31 17:53:52     return self.receive_nowait()
2023-08-31 17:53:52   File "/opt/fides/lib/python3.10/site-packages/anyio/streams/memory.py", line 92, in receive_nowait
2023-08-31 17:53:52     raise WouldBlock
2023-08-31 17:53:52 anyio.WouldBlock
2023-08-31 17:53:52 
2023-08-31 17:53:52 During handling of the above exception, another exception occurred:
2023-08-31 17:53:52 
2023-08-31 17:53:52 Traceback (most recent call last):
2023-08-31 17:53:52   File "/opt/fides/lib/python3.10/site-packages/starlette/middleware/base.py", line 77, in call_next
2023-08-31 17:53:52     message = await recv_stream.receive()
2023-08-31 17:53:52   File "/opt/fides/lib/python3.10/site-packages/anyio/streams/memory.py", line 112, in receive
2023-08-31 17:53:52     raise EndOfStream
2023-08-31 17:53:52 anyio.EndOfStream
2023-08-31 17:53:52 Traceback (most recent call last):
2023-08-31 17:53:52   File "<string>", line 1, in <module>
2023-08-31 17:53:52   File "/usr/local/lib/python3.10/multiprocessing/spawn.py", line 116, in spawn_main
2023-08-31 17:53:52     exitcode = _main(fd, parent_sentinel)
2023-08-31 17:53:52   File "/usr/local/lib/python3.10/multiprocessing/spawn.py", line 129, in _main
2023-08-31 17:53:52     return self._bootstrap(parent_sentinel)
2023-08-31 17:53:52   File "/usr/local/lib/python3.10/multiprocessing/process.py", line 314, in _bootstrap
2023-08-31 17:53:52     self.run()
2023-08-31 17:53:52   File "/usr/local/lib/python3.10/multiprocessing/process.py", line 108, in run
2023-08-31 17:53:52     self._target(*self._args, **self._kwargs)
2023-08-31 17:53:52   File "/opt/fides/lib/python3.10/site-packages/uvicorn/_subprocess.py", line 76, in subprocess_started
2023-08-31 17:53:52     target(sockets=sockets)
2023-08-31 17:53:52   File "/opt/fides/lib/python3.10/site-packages/uvicorn/server.py", line 61, in run
2023-08-31 17:53:52     return asyncio.run(self.serve(sockets=sockets))
2023-08-31 17:53:52   File "/usr/local/lib/python3.10/asyncio/runners.py", line 44, in run
2023-08-31 17:53:52     return loop.run_until_complete(main)
2023-08-31 17:53:52   File "/opt/fides/lib/python3.10/site-packages/starlette/middleware/base.py", line 69, in coro
2023-08-31 17:53:52     await self.app(scope, receive_or_disconnect, send_no_error)
2023-08-31 17:53:52   File "/opt/fides/lib/python3.10/site-packages/starlette/middleware/base.py", line 106, in __call__
2023-08-31 17:53:52     response = await self.dispatch_func(request, call_next)
2023-08-31 17:53:52 > File "/fides/src/fides/api/main.py", line 73, in dispatch_log_request
2023-08-31 17:53:52     response = await call_next(request)
2023-08-31 17:53:52   File "/opt/fides/lib/python3.10/site-packages/starlette/middleware/base.py", line 80, in call_next
2023-08-31 17:53:52     raise app_exc
2023-08-31 17:53:52   File "/opt/fides/lib/python3.10/site-packages/starlette/middleware/base.py", line 69, in coro
2023-08-31 17:53:52     await self.app(scope, receive_or_disconnect, send_no_error)
2023-08-31 17:53:52   File "/opt/fides/lib/python3.10/site-packages/starlette/middleware/cors.py", line 84, in __call__
2023-08-31 17:53:52     await self.app(scope, receive, send)
2023-08-31 17:53:52   File "/opt/fides/lib/python3.10/site-packages/starlette/middleware/base.py", line 104, in __call__
2023-08-31 17:53:52     async with anyio.create_task_group() as task_group:
2023-08-31 17:53:52   File "/opt/fides/lib/python3.10/site-packages/anyio/_backends/_asyncio.py", line 664, in __aexit__
2023-08-31 17:53:52     raise BaseExceptionGroup(
2023-08-31 17:53:52 exceptiongroup.ExceptionGroup: unhandled errors in a TaskGroup (1 sub-exception)
```

ThomasLaPiana:
these failures are being seen elsewhere too...unrelated to this change. I already fixed one in `main` but apparently not the root cause

pattisdr:
but is it possible bad data is still triggering it underneath?

pattisdr:
ah yikes I'm seeing this in unit tests, separate PR.  I wonder if mixing db sessions is related

ThomasLaPiana:
> ah yikes I'm seeing this in unit tests, separate PR. I wonder if mixing db sessions - synchronous/async is related
> 
> [#3990 (comment)](https://github.com/ethyca/fides/pull/3990#discussion_r1313209584)

I'm 90% sure it is caused by an undetected dependency change....I'll post an update in slack so the rest of the team is aware

pattisdr:
In a separate PR https://github.com/ethyca/fides/pull/3990#discussion_r1313209584, this error was happening when we were using a synchronous session in a ctl endpoint - the underlying error was ` sqlalchemy.exc.MissingGreenlet: greenlet_spawn has not been called; can't call await_() here. Was IO attempted in an unexpected place?`

adamsachs:
spoke a bit with @pattisdr and one theory that _could_ explain this `exceptiongroup.ExceptionGroup: unhandled errors in a TaskGroup` type of error from cropping up in all sorts of (seemingly unrelated) places all of a sudden is that some recent change (likely in a dependency of ours?) is causing underlying errors to get swallowed and instead be thrown up with this generic exception type.

@ThomasLaPiana do you have any further leads? i haven't dug into this much, just wanted to put this theory down on paper here. i think @pattisdr is also continuing further investigation on her end 👍

ThomasLaPiana:
> spoke a bit with @pattisdr and one theory that _could_ explain this `exceptiongroup.ExceptionGroup: unhandled errors in a TaskGroup` type of error from cropping up in all sorts of (seemingly unrelated) places all of a sudden is that some recent change (likely in a dependency of ours?) is causing underlying errors to get swallowed and instead be thrown up with this generic exception type.
> 
> @ThomasLaPiana do you have any further leads? i haven't dug into this much, just wanted to put this theory down on paper here. i think @pattisdr is also continuing further investigation on her end 👍

I moved this to a new issue: https://github.com/ethyca/fides/issues/4020

ThomasLaPiana:
I'm moving the migration-specific work to a new PR, to de-clutter this PR and focus it specifically on getting `fides` running with 2.0

pattisdr:
I think we need to update the config.json files we ship with, PR here https://github.com/ethyca/fides/pull/4076 although I'm going to need some FE assistance

pattisdr (approved):
Did another review here, although I didn't go through every single open comment.  

Giving this a tentative 👍 to not hold up the overnight crew, assuming the corresponding migration will be merged into this branch first, this branch will be updated with main, comments will be resolved, and that tests are passing. 

❗ Note that main has had a new migration merged in, we'll need to bump the downrev of these migrations.  

Separately, I think we need the `config.json` files we ship with to be updated, but FE help is needed https://github.com/ethyca/fides/pull/4076. We could also merge this in separately tomorrow.

pattisdr:
OK @ThomasLaPiana, Allison helped me with the FE, so these `config.json` changes are merged here -

ThomasLaPiana:
it looks like `fideslang 2.0` was missing a change that got patched into `1.4.6`, I cut another release of Fideslang (`2.0.3`) that should contain the change and get things working here

src/fides/api/db/seed.py
ThomasLaPiana: Updated as part of the migration and merged into this branch

tests/ctl/api/test_seed.py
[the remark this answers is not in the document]
ThomasLaPiana: that seems logical to me 🤔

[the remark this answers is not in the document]
ThomasLaPiana: done!

adamsachs (approved):
🎊 let's do this!

src/fides/api/alembic/migrations/versions/708a780b01ba_add_version_fields_to_default_types.py
```diff
@@ -0,0 +1,76 @@
+"""add version fields to default types
+
+Revision ID: 708a780b01ba
+Revises: 3038667ba898
```
adamsachs, line 4: nit - we should update the comment here for right `revises` id, just helps to avoid confusion when tracing back migrations!
```suggestion
Revises: 192f23f4c968
```

Issue #3974, opened 2023-08-23
update `system.legal_basis_for_transfers` 


adamsachs:
### Is your feature request related to a specific problem?

Our first cut of the system model updates made to support GVL/dictionary made some assumptions about the `system.legal_basis_for_transfers` field that are inconsistent with the `transferMechanisms` field in the GVL additional vendor info source (https://vendor-list.consensu.org/v2/additional-vendor-information-list.json) that it's based on. We should reconcile these inconsistencies.

Currently our `legal_basis_for_transfers` field:
- is a _single-value_ text field
- enum values: `'Adequacy decision', 'Standard contractual clauses', 'Binding corporate rules', 'Other']`

The GVL additional-vendor-information-list spec (referenced [here](https://github.com/InteractiveAdvertisingBureau/GDPR-Transparency-and-Consent-Framework/blob/master/TCFv2/Additional%20Vendor%20Information%20List%20Specification.md#example-additional-vendor-information-list-object), anywhere else?) defines the `transferMechanisms` field, upon which the `system.legal_basis_for_transfers` field is based (for GVL vendors), differently:
- multivalue (array) field
- enum values: `‘Adequacy decision’, ‘SCCs’, ‘BCRs’, ‘Other’`

### Describe the solution you'd like

We should update the `legal_basis_for_transfers` to allow multiple values (i.e. an array) and (i think) we should update the enum of accepted values to match the GVL source spec exactly. These changes need to be made across 3 repos:
- `fideslang` for  the pydantic/API model definitions
- `fides` for the db models and migrations
- `fidesplus` for any downstream datamap changes (we need to explicitly unpack multivalue fields when generating the datamap response, s/o to @pattisdr for calling this out proactively! 🙏) 

per @rsilvery's guidance, we'll also update the `system.legal_basis_for_profiling` field to be multivalue (with corresponding changes across the 3 repos)

### Describe alternatives you've considered, if any

Updating the accepted values isn't _strictly_ necessary, we could update our upstream utilities to transform the values to our preferred values that we have currently, but IMO this will add complexity and maintenance cost that i don't think is worth it. @pattisdr and i think we should move the transformation logic to the FE when we render the values, if we really want to give the end user the values we've specified currently (e.g. `Standard contractual clauses`).

### Additional context

Needed for `2.19.0` release.

Roger-Ethyca:
moving to done

PR #3977, opened 2023-08-25
fides - update system.legal_basis_for_transfers and system.legal_basis_for_profiling

adamsachs:
Part of #3974 

‼️ contains migration
‼️ ensure `requirements.txt` is pinned to actual `fideslang` version once released

### Description Of Changes

See issue for context/motivation. 

### Code Changes

* [x] bump `fideslang` dependency to integrate associated updates
* [x] update db models:  update `system.legal_basis_for_transfers` and `system.legal_basis_for_profiling` to be multivalue
* [x] associated migration for above changes: also includes data migration that nests previous single value text field as an array element, and then necessary value updates for `system.legal_basis_for_transfers`

### Steps to Confirm

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### DIFF
--- a/src/fides/api/alembic/migrations/versions/3038667ba898_update_system_legal_basis_fields.py
+++ b/src/fides/api/alembic/migrations/versions/3038667ba898_update_system_legal_basis_fields.py
@@ -1,0 +1,154 @@
+"""update system.legal_basis_for_transfers
+
+Revision ID: 3038667ba898
+Revises: 507563f6f8d4
+Create Date: 2023-08-25 14:19:16.798409
+
+"""
+from typing import Dict
+
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.orm.session import Session
+from sqlalchemy.sql import text
+
+# revision identifiers, used by Alembic.
+revision = "3038667ba898"
+down_revision = "507563f6f8d4"
+branch_labels = None
+depends_on = None
+
+legal_basis_transfers_upgrades: Dict[str, str] = {
+    "Adequacy decision": "Adequacy Decision",
+    "Standard contractual clauses": "SCCs",
+    "Binding corporate rules": "BCRs",
+}
+
+legal_basis_transfers_downgrades: Dict[str, str] = {
+    value: key for key, value in legal_basis_transfers_upgrades.items()
+}
+
+
+def upgrade():
+    session = Session(bind=op.get_bind())
+    conn = session.connection()
+    bind = op.get_bind()
+
+    #############################
+    # legal_basis_for_transfers #
+    #############################
+    statement_transfers = text(
+        """SELECT distinct legal_basis_for_transfers, id
+FROM ctl_systems"""
+    )
+    existing_values_transfers = conn.execute(statement_transfers)
+
+    # drop column since we will repopulating later
+    op.drop_column(
+        "ctl_systems",
+        "legal_basis_for_transfers",
+    )
+
+    # add column back, but make it multi-value
+    op.add_column(
+        "ctl_systems",
+        sa.Column(
+            "legal_basis_for_transfers",
+            type_=sa.ARRAY(sa.String()),
+            server_default="{}",
+            nullable=True,
+        ),
+    )
+
+    # data migration - updates to some values since allowable values enum has changed
+    # (enum is not enforced/specified on the DB, just API-level)
+
+    for row in existing_values_transfers:
+        existing_value = row[0]
+        updated_value = legal_basis_transfers_upgrades.get(
+            existing_value, existing_value
+        )
+        updated_value = [
+            updated_value
+        ]  # wrap in a list because column is now array type
+        bind.execute(
+            text(
+                "UPDATE ctl_systems SET legal_basis_for_transfers = :updated_basis WHERE id = :id"
+            ),
+            {"updated_basis": updated_value, "id": row[1]},
+        )
+
+    #############################
+    # legal_basis_for_profiling #
+    #############################
+    statement_profiling = text(
+        """SELECT distinct legal_basis_for_profiling, id
+FROM ctl_systems"""
+    )
+    existing_values_profiling = conn.execute(statement_profiling)
+
+    op.drop_column(
+        "ctl_systems",
+        "legal_basis_for_profiling",
+    )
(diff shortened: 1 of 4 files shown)


src/fides/api/alembic/migrations/versions/3038667ba898_update_system_legal_basis_fields.py
```diff
@@ -0,0 +1,154 @@
...
+    # (enum is not enforced/specified on the DB, just API-level)
+
+    for row in existing_values_transfers:
+        existing_value = row[0]
+        updated_value = legal_basis_transfers_upgrades.get(
+            existing_value, existing_value
+        )
+        updated_value = [
+            updated_value
+        ]  # wrap in a list because column is now array type
```
pattisdr, line 73: Early review comment, apologies if I'm butting in too early - make sure updated_value is not null so you don't get {NULL}

src/fides/api/alembic/migrations/versions/3038667ba898_update_system_legal_basis_fields.py
```diff
@@ -0,0 +1,154 @@
...
+    )
+
+    # data migration - wrap values in a list
+    for row in existing_values_profiling:
+        existing_value = row[0]
+        updated_value = legal_basis_transfers_upgrades.get(
+            existing_value, existing_value
+        )
+        updated_value = [
+            updated_value
```
pattisdr, line 111: and here!

src/fides/api/alembic/migrations/versions/3038667ba898_update_system_legal_basis_fields.py
```diff
@@ -0,0 +1,154 @@
...
+        )
+
+
+def downgrade():
+    """
+    Downgrade from array fields to single-value.
+
+    We do not try to perform a data migration, since we'd have to
+    drop some values arbitrarily. Better to just null the column out
+    to make it clear that manual intervention is needed.
```
pattisdr, line 127: what do you mean? can we still migrate what we can?

pattisdr: ah multiple values potentially, right

pattisdr: maybe a log here

src/fides/api/alembic/migrations/versions/3038667ba898_update_system_legal_basis_fields.py
adamsachs: never too early, thank you!!! good catch 🙏

src/fides/api/alembic/migrations/versions/3038667ba898_update_system_legal_basis_fields.py
adamsachs: added in [1857c7b](https://github.com/ethyca/fides/pull/3977/commits/1857c7b398969554f8d0e0192d674813c7ce0614)

src/fides/api/alembic/migrations/versions/3038667ba898_update_system_legal_basis_fields.py
adamsachs: addressed in [1857c7b](https://github.com/ethyca/fides/pull/3977/commits/1857c7b398969554f8d0e0192d674813c7ce0614)

src/fides/api/alembic/migrations/versions/3038667ba898_update_system_legal_basis_fields.py
adamsachs: addressed in [1857c7b](https://github.com/ethyca/fides/pull/3977/commits/1857c7b398969554f8d0e0192d674813c7ce0614) - does that look good to you? seemed to work for me locally on a manual smoke test.

src/fides/api/alembic/migrations/versions/3038667ba898_update_system_legal_basis_fields.py
pattisdr: thinking more, there's also an argument to be made for keeping the first one? wouldn't most just have one? maybe keep the first and log a list of records where data was lost to reduce downgrade effort. Your call though.

pattisdr (approved):

src/fides/api/alembic/migrations/versions/3038667ba898_update_system_legal_basis_fields.py
adamsachs: yeah, that's a fair argument...i had been weighing it a bit.

one other factor that i'm taking into consideration is that this downgrade is relevant for very few use cases - i don't think we've even officially released a version of the software that has these fields but has them as single-valued, i.e. this downgrade logic really applies to intra-release downgrades only. i think that it basically would have no impact outside of our self-hosted internal instances that have gotten incremental upgrades throughout these past couple of weeks, and i think in the (hopefully unlikely) scenario they need to downgrade, we could live with these fields getting wiped?

so ultimately i felt that it wasn't worth figuring out and testing that logic, but maybe i'm just being a bit lazy 😬

pattisdr: no you're right I forgot that it hadn't been released yet, that's fine!

src/fides/api/alembic/migrations/versions/3038667ba898_update_system_legal_basis_fields.py
```diff
@@ -0,0 +1,158 @@
...
+            type_=sa.ARRAY(sa.String()),
+            server_default="{}",
+            nullable=True,
+        ),
+    )
+
+    # data migration - wrap values in a list
+    for row in existing_values_profiling:
+        existing_value = row[0]
+        updated_value = legal_basis_transfers_upgrades.get(
```
pattisdr, line 108: taking another look 👀 

there was this copy paste issue but that shouldn't have had a practical effect.  The profiling value wouldn't have existed in the legal basis for transfers dictionary, so we would have used the original profiling value and added it to the list

PR #3991, opened 2023-08-29
Populate 'data stewards' field in system form

jpople:
### Description Of Changes

Populates "data stewards" field with assigned users to system.

### Steps to Confirm

* Assign a user to a system, then enter the system details view for that system-- user's username should be shown in "data stewards" field

### Pre-Merge Checklist

### DIFF
--- a/clients/admin-ui/src/features/common/form/inputs.tsx
+++ b/clients/admin-ui/src/features/common/form/inputs.tsx
@@ -502,6 +502,7 @@ export const CustomTextInput = ({
         </Flex>
         <TextInput
           {...field}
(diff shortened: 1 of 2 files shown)


clients/admin-ui/src/features/common/form/inputs.tsx
```diff
@@ -502,6 +502,7 @@ export const CustomTextInput = ({
         </Flex>
         <TextInput
           {...field}
+          value={props.value ? props.value : field.value}
```
TheAndrewJackson, line 505: Can this value be included in the initial values instead of through a prop? That would remove the need to update the base input field.

PR #3997, opened 2023-08-30
Add profiling/benchmarks to CI suite

ThomasLaPiana:
Closes #<issue>

### Description Of Changes

This is a first iteration on adding a suite of benchmarks and profiling tasks to our CI suite to better track performance improvements/degradation over time. Additionally, it gives us the ability to profile _any_ endpoint when using the `profile=true` query param.

* `nox -s performance_tests` will spin up the application and run stress testing as well as resource usage checks. This is helpful for having a baseline and keeping track of things via CI logs or when testing locally
* Hitting any endpoint with `profile=true` as a query parameter _while in dev mode_ will return an HTML-formatted performance trace as opposed to the expected return object. i.e. `localhost:8080/health/database?profile=true`

### Code Changes

* [x] add a new `nox` command to do CPU/memory usage via Docker [docker stats](https://docs.docker.com/engine/reference/commandline/stats/)
* [x] add a load testing `nox` command using [drill](https://github.com/fcsonline/drill)
* [x] add the ability to benchmark any endpoint in dev mode

### Steps to Confirm

* [ ] `nox -s dev` and send an API request with the `profile-request: true` header

### Pre-Merge Checklist

* [x] Update `CHANGELOG.md`
* [x] Update docs with steps/tips for benchmarking/profiling


src/fides/api/main.py
```diff
@@ -50,6 +51,21 @@
 
 app = create_fides_app()
 
+if CONFIG.dev_mode:
+
+    @app.middleware("http")
+    async def profile_request(request: Request, call_next: Callable) -> Response:
+        profiling = request.query_params.get("profile", False)
```
adamsachs, line 58: i really like this middleware design, we get a big bang for our buck: super lightweight but very powerful.

one nit: could this be a request header rather than query param? i don't have many concrete reasons to suggest that, more of just a gut feeling. if i searched my gut (gross) for reasons, here's what i found: 
- i usually think of a query-param as parameterization local to a given endpoint and the endpoint's specific business logic, whereas headers typically hold general metadata that's more common throughout all (or most) of the endpoints in a given API? 
- i think there's also a _slight_ chance that a param like `profile` could actually be used for another purpose by a specific API at some point, whereas a `profile-request` header feels a lot less ripe for ambiguity or overloading.

if you feel convinced that the query param here is the better option, then i'll defer to you - i don't feel too strongly about this!

```suggestion
        profiling = request.headers.get("profile-request", False)
```

noxfiles/ci_nox.py
```diff
@@ -216,6 +216,52 @@ def minimal_config_startup(session: nox.Session) -> None:
...
+## Performance ##
+#################
+@nox.session()
+def performance_tests(session: nox.Session) -> None:
+    """Compose the various performance checks into a single uber-test."""
+    session.notify("teardown")
+    session.run(*START_APP, external=True, silent=True)
+    samples = 2
+    for i in range(samples):
+        session.log(f"Sample {i} of {samples}")
```
adamsachs, line 229: :trollface: 

```suggestion
        session.log(f"Sample {i + 1} of {samples}")
```

it does read a bit better in the output, i swear!

adamsachs (approved):
this is great, two minor nits that you can choose to ignore!

really liking that middleware as a quick way to give us a lot of good info on an arbitrary endpoint 👍

src/fides/api/main.py
ThomasLaPiana: More of a smart design decision than a nit 🙂 good call

noxfiles/ci_nox.py
ThomasLaPiana: my computer brain saw "0" and thought "this is fine" lol

Issue #3992, opened 2023-08-29
Show all privacy notices and disable by default


Kelsey-Ethyca:
AC:

- All privacy notices appear whether connected to a data use or not
- All privacy notices are disabled by default
- When a privacy notice is attached to a data use that exists in the map, a suggestion indicator appears.

allisonking:
To do this, we would need to:

Frontend:
* Remove the `systems_applicable` filter from the endpoint call
* Render "suggestion indicator" based on a new prop from the backend? (see below)

Backend:
* Make all privacy notices disabled by default (update `privacy_notice_templates.yml`)
* Make `systems_applicable` a prop on PrivacyNotices? This would allow us to render the suggestion indicator

@pattisdr can you let me know if you think this is the right approach? I'm happy to take a stab at the backend if you don't think it'd be too complicated :)

allisonking:
Perfect, thanks for the notes! I'll do my best 🫡

@Kelsey-Ethyca let me know if you have any thoughts on what the "suggestion indicator" should look like? I didn't see it in the linked design. otherwise I'll make something up and we can iterate 👍

PR #4010, opened 2023-08-31
Update displaying privacy notices logic

allisonking:
Closes https://github.com/ethyca/fides/issues/3992

### Description Of Changes

https://www.loom.com/share/30bea7fc48d047d8a4aeb629b5fd4fdd

We no longer filter by `systems_applicable` on the frontend. Instead, we get all notices, but mark which ones are `systems_applicable` through a new "Status" column. Some backend work was needed in order for the notice to have a prop `systems_applicable`.

### Code Changes

* [x] Disable privacy notices by default (in the templates)
* [x] Add a new `systems_applicable` property to privacy notices
* [x] Update backend tests
* [x] Update frontend TS types (autogenerated)
* [x] Add suggestion indicator to privacy notice table
* [x] Add a new cypress tests for the indicator

### Steps to Confirm

* Clear your fides db (i.e. `docker-compose down -v`) so that we'll get the new privacy notice templates
* `nox -s dev` to start up fides
* `turbo dev` in `clients/admin-ui` to start the admin UI
* Edit `features.slice.ts` in `admin-ui` to force the constant `plus = true` to simulate a plus environment (alternatively, build a tag off of this branch and run fidesplus with that tag)
* Navigate to http://localhost:3000/consent/privacy-notices and confirm all notices are there but disabled. They should all have the status "Inactive" and their "Enable" toggles should be disabled.
* Add an applicable system (i.e. `marketing.advertising.first_party.targeted`)
* The relevant notice should now have status "Available"
* Enabling it should change its status "Enabled"
* Disabled and no applicable systems: None
* You should still be able to disable this toggle.
* If you change to a user who does not have the permission `privacy-notice:update`, then the Enable column should disappear

### Pre-Merge Checklist

* [x] Update `CHANGELOG.md`


clients/admin-ui/src/types/api/index.ts
```diff
@@ -27,6 +27,7 @@ export type { Body_upload_data_api_v1_storage__request_id__post } from "./models
 export type { BulkPostPrivacyRequests } from "./models/BulkPostPrivacyRequests";
```
allisonking, line 27: all the files in `types/api` were autogenerated. it's been a while since we generated them, so the diff is larger than just what this PR changed

src/fides/api/models/privacy_notice.py
```diff
@@ -296,6 +296,15 @@ def cookies(self) -> List[Cookies]:
             )
         ).all()
 
+    @property
+    def systems_applicable(self) -> bool:
```
allisonking, line 300: I'm unsure about this—my instinct was to use a `property` or a `hybridproperty`, although I had trouble getting the `hybridproperty` to work. I was hoping once I had this field in place, I could use this field for the `systems_applicable` query param filter to simplify that logic a bit, but that didn't work out. I'm not sure if there's a way to make this property figure out `systems_applicable`, and then have the filter param `systems_applicable` just filter on this property, vs doing the logic again here in the endpoint? 

https://github.com/ethyca/fides/blob/a0544e57d42f58425b27e0b58b843e489ad1a671/src/fides/api/api/v1/endpoints/privacy_notice_endpoints.py#L58-L60

clients/admin-ui/src/features/privacy-notices/PrivacyNoticesTable.tsx
```diff
@@ -83,28 +82,7 @@ export const PrivacyNoticesTable = () => {
       </Flex>
     );
   }
-  if (privacyNotices.length === 0) {
```
allisonking, line 86: now that we are always showing notices, and notices are seeded into the db, it didn't seem like we needed this empty state anymore

src/fides/api/models/privacy_notice.py
pattisdr: I think this looks good to me as-is! From a code-sharing perspective, I can understand the motivation to want to have `systems_applicable` use that property to filter as well, but I think that would be less efficient than querying the `PrivacyNotice.data_uses` database column directly as we're doing now.   Granted, I can't imagine the PrivacyNotice table every being very big, so the practical difference between the two would be small.  But generally when querying an entire table, I try not to filter on a property but query the column directly to build the most straightforward and optimized sql query.

pattisdr (commented):
Bravo! 

Noting this here, we chatted about in the issue, but this will not affect pre-existing notices already created, just new out-of-the-box notices going forward.  They will still be enabled.  I think that makes sense, they should disable themselves.

src/fides/api/models/privacy_notice.py
allisonking: ah got it, thanks for explaining!

allisonking:
Confirmed that the failing backend tests are unrelated to this PR (they're [in main too](https://github.com/ethyca/fides/actions/runs/6117056104/job/16603296644)!) 

@TheAndrewJackson do you mind reviewing the FE portion of this PR? 🙏

TheAndrewJackson:
@allisonking not at all! I'll take a look now

clients/admin-ui/src/features/privacy-notices/cells.tsx
```diff
@@ -10,6 +11,29 @@ export const MechanismCell = (
   cellProps: CellProps<typeof MECHANISM_MAP, string>
 ) => <MapCell map={MECHANISM_MAP} {...cellProps} />;
 
+type TagNames = "available" | "enabled" | "inactive";
+
+const systemsApplicableTags: Record<TagNames, TagProps & { tooltip: string }> =
```
TheAndrewJackson, line 16: I haven't seen this use of `Record` before. Nice!

TheAndrewJackson (approved):
:shipit:

clients/admin-ui/src/features/privacy-notices/cells.tsx
allisonking: ah yeah, it's a nice way to declare a dictionary, esp if you already know what the keys will be!

PR #4013, opened 2023-09-01
Fix Parallel Docker Platform Builds

ThomasLaPiana:
Closes #

### Description Of Changes

I previously had tried to get parallel Docker builds working using Python's own multiprocessing framework, however this appeared to not function as expected in GitHub Actions.

This PR iterates on that and shifts the parallel execution responsibility to GitHub Actions with a strategy matrix

Used this for reference when redesigning how the `outputs` work: [GitHub Output Docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idoutputs)

### Code Changes

* [x] Update the Nox command to only run one build command at a time (this makes parallelization easier)
* [x] update the Nox tests
* [x] refactor the `Publish Docker Images` workflow to facilitate a strategy matrix

### Steps to Confirm

* [x] pushed an `alpha` [tag](https://github.com/ethyca/fides/releases/tag/2.18.1a21) to this branch and confirmed that the resulting [CI job](https://github.com/ethyca/fides/actions/runs/6050503276) to publish docker images completed successfully and with the expected parallelized execution

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### DIFF
--- a/.github/workflows/publish_docker.yaml
+++ b/.github/workflows/publish_docker.yaml
@@ -12,8 +12,69 @@ env:
   DOCKER_TOKEN: ${{ secrets.DOCKER_TOKEN }}
 
 jobs:
-  push-fides:
+  ParseTags:
     runs-on: ubuntu-latest
+    outputs:
+      prod_tag: { { steps.check-prod-tag.outputs.match } }
+      rc_tag: { { steps.check-rc-tag.outputs.match } }
+      alpha_tag: { { steps.check-alpha-tag.outputs.match } }
+      beta_tag: { { steps.check-beta-tag.outputs.match } }
+    steps:
+      - name: Check Prod Tag
+        id: check-prod-tag
+        run: |
+          if [[ ${{ github.event.ref }} =~ ^refs/tags/[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+            echo "match=true" >> $GITHUB_OUTPUT
+          else
+            echo "match=false" >> $GITHUB_OUTPUT
+          fi
+
+      - name: Check RC Tag
+        id: check-rc-tag
+        run: |
+          if [[ ${{ github.event.ref }} =~ ^refs/tags/[0-9]+\.[0-9]+\.[0-9]+rc[0-9]+$ ]]; then
+            echo "match=true" >> $GITHUB_OUTPUT
+          else
+            echo "match=false" >> $GITHUB_OUTPUT
+          fi
+
+      - name: Check alpha Tag
+        id: check-alpha-tag
+        run: |
+          if [[ ${{ github.event.ref }} =~ ^refs/tags/[0-9]+\.[0-9]+\.[0-9]+a[0-9]+$ ]]; then
+            echo "match=true" >> $GITHUB_OUTPUT
+          else
+            echo "match=false" >> $GITHUB_OUTPUT
+          fi
+
+      - name: Check beta Tag
+        id: check-beta-tag
+        run: |
+          if [[ ${{ github.event.ref }} =~ ^refs/tags/[0-9]+\.[0-9]+\.[0-9]+b[0-9]+$ ]]; then
+            echo "match=true" >> $GITHUB_OUTPUT
+          else
+            echo "match=false" >> $GITHUB_OUTPUT
+          fi
+
+      # if an RC git tag, also notify Fidesinfra to trigger a redeploy of rc env, to pick up our newly published images
+      - name: Send Repository Dispatch Event (RC redeploy)
+        if: steps.check-rc-tag.outputs.match == 'true'
+        uses: peter-evans/repository-dispatch@v2
+        with:
+          event-type: trigger-fidesinfra-deploy-fides-rc
+          repository: ethyca/fidesinfra
+          token: ${{ secrets.DISPATCH_ACCESS_TOKEN }}
+
+  Push:
+    runs-on: ubuntu-latest
+    needs: ParseTags
+    strategy:
+      # This matrix will effectively _try_ to run every permutation in parallel,
+      # skipping all of the tasks that don't match. This leaves a ton of "skipped" jobs
+      # but is the fastest way to get this working without overhauling the tag check logic.
+      matrix:
+        application: ["fides", "sample_app", "privacy_center"]
+        platform: ["x86", "ARM"]
     steps:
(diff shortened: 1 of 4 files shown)


.github/workflows/publish_docker.yaml
```diff
@@ -86,13 +65,52 @@ jobs:
           repository: ethyca/fidesinfra
           token: ${{ secrets.DISPATCH_ACCESS_TOKEN }}
 
+  Push:
```
ThomasLaPiana, line 68: @adamsachs I'll definitely need your help checking this logic, also thinking we need some way to test this if possible but maybe it isn't possible 🤷

ThomasLaPiana:
would love to get an "early" review here! This is the direction that seemed best but also still a bit janky, so would appreciate some feedback

adamsachs (commented):
this looks really good! i know you preferred the parallelization within the command but from my POV this is quite a bit easier to follow now, in addition to (hopefully 🤞) actually getting us the parallelization in CI that we're looking for! i guess the tradeoff is slightly harder debugging if/when things go wrong (since we've outsourced the parallelization to GH/CI) but i think it's worth the trade 👍 

before approving, i'm just going to try pushing an alpha tag to your latest because i think that'll give us a decent test of the new functionality and should expose any fundamental issues.

adamsachs:
> before approving, i'm just going to try pushing an alpha tag to your latest because i think that'll give us a decent test of the new functionality and should expose any fundamental issues.

OK i pushed `2.18.1a19` and hit a [syntax error](https://github.com/ethyca/fides/actions/runs/6050404801), looking closer 👀

adamsachs:
OK! got things working after a few small fixes to syntax in the worfklow file.

the good news is that things look nicely parallelized as we were hoping. the not-so-good news is that it seems like (lack of) parallelization wasn't really too much of a problem in terms of overall time in the push job. the `fides` and `privacy-center` ARM builds seem to just take a _lonnnggg_ time on their own - ~50 minutes. [more info here](https://github.com/ethyca/fides/actions/runs/6050503276/job/16419970687?pr=4013)

image

to be fair, the published ARM image builds of core OSS `fides` aren't _that_ heavily used afaik. most people running OSS fides are probably building their images locally from source, right? maybe it's worth splitting that job out into its own separate GH action - that way we could decouple it from the x86 builds that are more important for hosted envs and downstream CI jobs generally, for more efficient tracking/event firing. we could even consider having the ARM image publishing for core `fides` be an action that's invoked on-demand, rather than after every tag/commit to `main`?

curious to get your thoughts @ThomasLaPiana and @SteveDMurphy .

adamsachs (approved):
comment above for [some thoughts](https://github.com/ethyca/fides/pull/4013#issuecomment-1702925574) about why this may not be quite the improvement we were hoping for/expecting, but at least it is an incremental improvement and there's no real downside i see! the new code and resulting GH actions are IMO easier to follow.

thanks for continuing to follow up on this in a sensible way! 👍

ThomasLaPiana:
> comment above for [some thoughts](https://github.com/ethyca/fides/pull/4013#issuecomment-1702925574) about why this may not be quite the improvement we were hoping for/expecting, but at least it is an incremental improvement and there's no real downside i see! the new code and resulting GH actions are IMO easier to follow.
> 
> thanks for continuing to follow up on this in a sensible way! i think this is ready to go 👍

On the contrary it seems like we're getting exactly what we wanted here! The `x86` and `ARM` jobs are now completely decoupled, so even though it takes ARM ~1 hour to finish (as expected considering the need for virtualization) the other jobs are succeeding in mere minutes. We test images using the `x86` version, so this means we can start testing as soon as those jobs finish without waiting for the ARM images to finish. Therefore, it seems to me like this _does_ actually solve the problem we were having?

adamsachs:
> > comment above for [some thoughts](https://github.com/ethyca/fides/pull/4013#issuecomment-1702925574) about why this may not be quite the improvement we were hoping for/expecting, but at least it is an incremental improvement and there's no real downside i see! the new code and resulting GH actions are IMO easier to follow.
> > thanks for continuing to follow up on this in a sensible way! i think this is ready to go 👍
> 
> On the contrary it seems like we're getting exactly what we wanted here! The `x86` and `ARM` jobs are now completely decoupled, so even though it takes ARM ~1 hour to finish (as expected considering the need for virtualization) the other jobs are succeeding in mere minutes. We test images using the `x86` version, so this means we can start testing as soon as those jobs finish without waiting for the ARM images to finish. Therefore, it seems to me like this _does_ actually solve the problem we were having?

OK yeah fair enough! it solves the primary problem. i think i'd just had a bad expectation that none of the jobs individually were taking so long, but that was clearly mistaken. agreed that this PR does everything it's supposed to.

>  so this means we can start testing as soon as those jobs finish without waiting for the ARM images to finish

that's right, and totally agree. i guess i just wonder whether we should start splitting this up further to make it clearer to anyone waiting for the x86 images, as well as to potentially trigger downstream CI actions (that rely only on the x86 images) sooner. we can address that in a follow up PR 👍 

> P.S. Thank you for the awesome contributions and helping me get this over the line!

my pleasure, very happy to get this into `main` ASAP - big thanks to you for taking the initiative to come back to this and fix it up 💯

.github/workflows/publish_docker.yaml
```diff
@@ -62,21 +56,6 @@ jobs:
...
-        if: steps.check-prod-tag.outputs.match == 'true'
-        run: nox -s "push(prod)"
-
-      # if an RC git tag, then we run the rc job to publish images with an ":rc" tag
-      - name: Push Fides RC Tags
-        if: steps.check-rc-tag.outputs.match == 'true'
-        run: nox -s "push(rc)"
-
       # if an RC git tag, also notify Fidesinfra to trigger a redeploy of rc env, to pick up our newly published images
       - name: Send Repository Dispatch Event (RC redeploy)
```
adamsachs, line 60: whoooops, i missed this...this needed to move down below, it needs to run _after_ the `Push` job completes. i'll create a follow-up.

adamsachs: here's a follow up/fix: https://github.com/ethyca/fides/pull/4017

Kelsey-Ethyca:
Temporarily reverting this so we can run cypress tests in fidesplus

PR #4024, opened 2023-09-05
Build Multiplatform Images Simultaneously

ThomasLaPiana:
Closes #<issue>

### Description Of Changes

This is actually a redo of #4013 but with a better understanding of how `buildx` wants to handle multiplatform builds

### Code Changes

* [x] copy pasta changes from the previous PR: #4013
* [x] Update the code to run platforms as a single `build` command
* [x] move the `notify fidesinfra` step to the end

### Steps to Confirm

* [ ] trigger a build?
* [ ] run `nox -s "push(fides, dev)"` and verify that it is building both images

### Pre-Merge Checklist

* [ ] ~~Issue Requirements are Met~~
* [x] Update `CHANGELOG.md`


ThomasLaPiana:
@adamsachs brought over all of the changes of the previous PR but fixed the platform issue and also fixed the issue where the `rc-env` was notified too early

adamsachs:
pushed a `2.19.1a8` tag to the HEAD of this branch to test out the image build - resulting action [here](https://github.com/ethyca/fides/actions/runs/6138087149). results to follow...

.github/workflows/publish_docker.yaml
```diff
@@ -62,37 +56,60 @@ jobs:
...
+      # if not a prod git tag, then we run the git-tag job to publish images with a git tag
+      # if one exists on the current commit. the job is a no-op if the commit hasn't been tagged
+      - name: Push Fides Commit Tags
+        if: needs.ParseTags.outputs.prod_tag == 'false'
+        run: nox -s "push(${{ matrix.application }},git-tag)"
 
       # if an RC git tag, also notify Fidesinfra to trigger a redeploy of rc env, to pick up our newly published images
       - name: Send Repository Dispatch Event (RC redeploy)
-        if: steps.check-rc-tag.outputs.match == 'true'
+        if: needs.ParseTags.outputs.rc_tag == 'true'
```
adamsachs, line 113: i think we need to move this out into its own separate job, or else it will be triggered by each published application image within the matrix, and that's too many dispatch notifications; it should just be one notification when all application images have been pushed. i'll push up a commit to adjust this. i think i'd gotten this wrong in my original PR, sorry for that!

adamsachs: updated in [5b46063](https://github.com/ethyca/fides/pull/4024/commits/5b46063e2af95ea168cd5765619c75efacb53c84)

adamsachs:
> pushed a `2.19.1a8` tag to the HEAD of this branch to test out the image build - resulting action [here](https://github.com/ethyca/fides/actions/runs/6138087149). looks like we're still getting long build times (~2 hrs) in the case where we publish git commit tags (see screenshot of results form above job run). i don't think there's any reason we can't push the git-tag images alongside the other tags, that should all be the same command (since it's the same image!). i'll push up a slight proposed refactor -- [cb5adcf](https://github.com/ethyca/fides/pull/4024/commits/cb5adcf2bc3230f97671f677fefb0e2c17352d4c) -- to make this improvement, let me know what you think!
> 
> image

OK with my most recent changes, this is looking better to me! a bit over an hour for a successful push of an alpha tag, previously was >2 hrs. we still get both architecture images for both image tags. i still would love to be able to have the decoupling of amd64/arm64 archs that the original PR had, but like you say, that just seems incompatible with how `buildx` works 😢 .

@ThomasLaPiana let me know what you think of my updates - still not sure the implementation is the cleanest, but i think the end result is something we should target, i.e. not having to re-repeat the whole build for the "git-tag" image push.

image

image

ThomasLaPiana:
@adamsachs hard to argue with a 2x speed increase 🙂 nice changes here! If you're good to merge so am I

PR #4049, opened 2023-09-08
Fides: Relax Legal Basis for Transfers

pattisdr:
Closes https://github.com/ethyca/fides/issues/4057

❗ Needs 1.4.6 bump

### Description Of Changes

Allow System Legal Basis for Transfers to be any string.  

### Code Changes

* [ ] Bumps fideslang to 1.4.6 https://github.com/ethyca/fideslang/releases/tag/1.4.6

### Steps to Confirm

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### DIFF
--- a/clients/admin-ui/src/features/system/SystemInformationFormSelectOptions.ts
+++ b/clients/admin-ui/src/features/system/SystemInformationFormSelectOptions.ts
@@ -7,13 +7,14 @@ export const legalBasisForProfilingOptions = [
   label: opt,
 }));
 
+// Backend technically allows any string
 export const legalBasisForTransferOptions = [
   {
     value: "Adequacy Decision",
     label: "Adequacy decision",
   },
-    {
-    value: "Supplementary Measures",
+  {
+    value: "Supplementary measures",
     label: "Supplementary measures",
   },
   {
--- a/clients/admin-ui/src/features/system/SystemInformationFormSelectOptions.ts
+++ b/clients/admin-ui/src/features/system/SystemInformationFormSelectOptions.ts
@@ -12,6 +12,10 @@ export const legalBasisForTransferOptions = [
     value: "Adequacy Decision",
     label: "Adequacy decision",
   },
+    {
+    value: "Supplementary Measures",
+    label: "Supplementary measures",
+  },
   {
     value: "SCCs",
     label: "Standard contractual clauses",
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,7 +12,7 @@ expandvars==0.9.0
 fastapi[all]==0.89.1
 fastapi-caching[redis]==0.3.0
 fastapi-pagination[sqlalchemy]==0.11.4
-fideslang==1.4.5
+fideslang @ git+https://github.com/ethyca/fideslang.git@4ecf5f590d54b9a519457eceb106a8eef5ee779a#egg=fideslang
 fideslog==1.2.10
 firebase-admin==5.3.0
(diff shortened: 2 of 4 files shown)


requirements.txt
```diff
@@ -12,7 +12,7 @@ expandvars==0.9.0
 fastapi[all]==0.89.1
 fastapi-caching[redis]==0.3.0
 fastapi-pagination[sqlalchemy]==0.11.4
-fideslang==1.4.5
+fideslang @ git+https://github.com/ethyca/fideslang.git@4ecf5f590d54b9a519457eceb106a8eef5ee779a#egg=fideslang
```
pattisdr, line 15: Will switch to 1.4.6 when I get that out -

clients/admin-ui/src/features/system/SystemInformationFormSelectOptions.ts
```diff
@@ -11,6 +11,10 @@ export const legalBasisForTransferOptions = [
   {
     value: "Adequacy Decision",
     label: "Adequacy decision",
+  },
+    {
+    value: "Supplementary Measures",
+    label: "Supplementary measures",
```
adamsachs, line 17: i'm just not 100% sure what's gonna happen here because the value coming from the GVL (and from the dictionary) is actually `Supplementary measures` whereas for al the other entries here, the value from the GVL/dctionary matches the `value` field here...

maybe @jpople can clarify what the behavior is gonna be if we get `"Supplementary measures"` populated from the dictionary?

jpople: I _believe_ it will just fail to populate if it tries to give it an invalid value, but let me test it really quick.  What's the desired behavior here?

pattisdr: Should we just change to `Supplementary measures`, `Supplementary measures`?

adamsachs: hmm we could probably just update the `value` to be `"Supplementary measures"` just to match what they're giving us. our fides API should allow _anything_ to be submitted now, so that's one option that's probably the simplest to implement.

i think a slightly nicer approach, more in-line with what we've got on the BE now, would be to allow the dictionary to populate _any_ value here and we'd just show exactly what the dictionary gives us if it doesn't match an enumerated "option" here? but then these enumerated options would be all that's allowed for someone _manually_ entering in the values here in the form. that feels like it could be a big change though. it'd make us a bit more resilient to these upstream changes coming in without us knowing, but if it's too much work for now, i think we could just do the first suggestion.

@pattisdr do you agree?

adamsachs: whoops, hadn't seen your comment when i wrote mine @pattisdr . i think we landed in the same place, let's just go with that for now 👍

requirements.txt
pattisdr: Bumped to 1.4.6! https://github.com/ethyca/fideslang/releases/tag/1.4.6

clients/admin-ui/src/features/system/SystemInformationFormSelectOptions.ts
pattisdr: updated! let's hope no FE tests fail

adamsachs (approved):
💨

PR #4073, opened 2023-09-12
CORS domain config api and config page

TheAndrewJackson:
Closes fidesplus#1065

### Description Of Changes

Add CORS configuration to the admin ui and config proxy api.

https://github.com/ethyca/fides/assets/17103888/84d68bb5-506b-4f25-b0f7-0b5fad42cb77

### Code Changes

* [ ] Add CORS configuration page
* [ ] Update config proxy to have the `cors_origin` security field
* [ ] Update config api to update `cors_origin`s within the CORS middleware at runtime

### Steps to Confirm

* [ ] run `nox -s dev`
* [ ] update this [line](https://github.com/ethyca/fides/blob/b9acaa785578cadf0f5271041d972fcfd6e8ec12/clients/admin-ui/src/features/common/features/features.slice.ts#L151) to `const plus = true;` so the admin ui thinks plus is running
* [ ] run the admin ui
* [ ] go to `/management/cors-configuration`
* [ ] add a new domain(s) to the form and save
* [ ] restart the server and ensure the values were saved properly
* [ ] remove the new domain(s) and save

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`


clients/admin-ui/src/features/privacy-requests/privacy-requests.slice.ts
```diff
@@ -440,10 +460,26 @@ export const {
...
+  useGetConfigurationSettingsQuery,
   useGetMessagingConfigurationDetailsQuery,
   useGetActiveMessagingProviderQuery,
   useGetActiveStorageQuery,
   useCreateMessagingConfigurationMutation,
   useCreateMessagingConfigurationSecretsMutation,
   useCreateTestConnectionMessageMutation,
 } = privacyRequestApi;
+
+const EMPTY_CORS_DOMAINS = { domains: [] } as { domains: string[] };
```
allisonking, line 472: I think something like this would be a little more strict (`as` does a cast, whereas `:` is defining it)

```suggestion
const EMPTY_CORS_DOMAINS : { domains: string[] } = { domains: [] };
```

clients/admin-ui/src/features/privacy-requests/privacy-requests.slice.ts
```diff
@@ -440,10 +460,26 @@ export const {
...
+      (state) => state,
+      privacyRequestApi.endpoints.getConfigurationSettings.select(),
+    ],
+    (state, { data }) => {
+      // the type checker was having a really hard time with `data`
+      const properlyTypedData = data as ApplicationConfig;
+      const domains = properlyTypedData?.security?.cors_origins;
+      return domains ? { domains } : EMPTY_CORS_DOMAINS;
+    }
+  );
```
allisonking, line 485: is there a particular reason why this is in the privacy requests slice file? it probably works just fine here, but seems a little strange to me

clients/admin-ui/src/features/privacy-requests/privacy-requests.slice.ts
```diff
@@ -440,10 +460,26 @@ export const {
...
+export const selectCORSDomains = () =>
+  createSelector(
+    [
+      (state) => state,
+      privacyRequestApi.endpoints.getConfigurationSettings.select(),
+    ],
+    (state, { data }) => {
+      // the type checker was having a really hard time with `data`
+      const properlyTypedData = data as ApplicationConfig;
+      const domains = properlyTypedData?.security?.cors_origins;
```
allisonking, line 482: hmm I suppose it's because `getConfigurationSettings` has [a lot of things it could be](https://github.com/ethyca/fides/blob/cors-domain-config-api/clients/admin-ui/src/features/privacy-requests/privacy-requests.slice.ts/#L358-L362), and the backend only types it as [`Dict[str, Any]`](https://github.com/ethyca/fides/blob/cors-domain-config-api/src/fides/api/api/v1/endpoints/config_endpoints.py#L28)

the proper thing to do would likely be to set the return type of `getConfigurationSettings` to `Record<string, any>`, and then use something like the `narrow` lib or some other type predicate function to make sure the config has `security` and `cors_origins` (and so we know it's an `ApplicationConfig` type). that said, I don't think it provides all that much more safety than what you already have, so I won't be a stickler about it 😄

clients/admin-ui/src/pages/management/cors-configuration.tsx
```diff
@@ -0,0 +1,172 @@
...
+  useGetConfigurationSettingsQuery,
+} from "~/features/privacy-requests/privacy-requests.slice";
+
+type FormValues = {
+  domains: string[];
+};
+
+const CORSConfigurationPage: NextPage = () => {
+  const { isLoading: isLoadingGetQuery } = useGetConfigurationSettingsQuery();
+  const corsDomains = useAppSelector(selectCORSDomains());
```
allisonking, line 37: I'm kind of surprised this works, I thought it had to be `useAppSelector(selectCORSDomains)`!

clients/admin-ui/src/pages/management/cors-configuration.tsx
```diff
@@ -0,0 +1,172 @@
...
+  const [patchConfigSettingsTrigger, { isLoading: isLoadingPatchMutation }] =
+    useCreateConfigurationSettingsMutation();
+
+  const toast = useToast();
+  if (isLoadingGetQuery) {
+    return <div>loading</div>;
```
allisonking, line 43: how about a `<Spinner/>` instead?

clients/admin-ui/src/pages/management/cors-configuration.tsx
```diff
@@ -0,0 +1,172 @@
...
+          </Text>
+          <Text mb={10} fontSize="sm">
+            Please visit{" "}
+            <Link
+              color="complimentary.500"
+              href="https://docs.ethyca.com"
+              isExternal
+            >
+              docs.ethyca.com
+            </Link>{" "}
```
allisonking, line 95: there's actually a [DocsLink component](https://github.com/ethyca/fides/blob/cors-domain-config-api/clients/admin-ui/src/features/common/DocsLink.tsx) we could use here

clients/admin-ui/src/pages/management/cors-configuration.tsx
```diff
@@ -0,0 +1,172 @@
...
+            <Formik<FormValues>
+              initialValues={corsDomains}
+              enableReinitialize
+              onSubmit={handleSubmit}
+            >
+              {({ dirty, values, isValid }) => (
+                <Form>
+                  <FieldArray
+                    name="domains"
+                    render={(arrayHelpers) => (
```
allisonking, line 110: arrayHelpers, neat!

clients/admin-ui/src/pages/management/cors-configuration.tsx
```diff
@@ -0,0 +1,172 @@
...
+                            rightIcon={<AddIcon />}
+                          >
+                            Add CORS domain
+                          </Button>
+                        </Flex>
+                      </Flex>
+                    )}
+                  />
+
+                  <Box mt={6}>
```
allisonking, line 151: the empty state looks a little weird since the save button appears. could we remove it if there are no domains?
image

allisonking (commented):
nice job, this seems to work great! I left a few minor comments. I'll leave the backend review to Adam :)

clients/admin-ui/src/features/privacy-requests/privacy-requests.slice.ts
TheAndrewJackson: I ended up finding a pretty clean typing solution for this line that's different from the suggestion. I'm going to resolve this comment. Let me know what you think of the new version

clients/admin-ui/src/pages/management/cors-configuration.tsx
TheAndrewJackson: I _think_ I did this because it didn't work without the `()`. I could be incorrect on that. Either way it still works with the function being called 🤷‍♂️

clients/admin-ui/src/pages/management/cors-configuration.tsx
TheAndrewJackson: good idea! I updated it to use a spinner within the form section

clients/admin-ui/src/features/privacy-requests/privacy-requests.slice.ts
TheAndrewJackson: I found it strange too. The patch route was already here so I added the get route alongside it. I think it was put there originally because it was used for configuring something related to privacy requests

clients/admin-ui/src/pages/management/cors-configuration.tsx
TheAndrewJackson: It does look a little weird but someone may want to remove all of they're CORS domains so removing it would prevent them from doing that.

allisonking: ah true!

clients/admin-ui/src/features/privacy-requests/privacy-requests.slice.ts
allisonking: oh great, I'm glad! looks great 😎

adamsachs (commented):
@TheAndrewJackson nice work on getting this working! there are a few stylistic/design points i'd like to refine, some more important than others. let me know if anything doesn't make sense or you'd like to discuss further - i think some of my comments may be a bit dense and hard to parse 😄

src/fides/config/utils.py
```diff
@@ -81,3 +84,15 @@ def get_dev_mode() -> bool:
         "active_default_storage_type",
     ],
 }
+
+
+def load_updated_cors_domains(domains: List[str], resetDomains=False) -> None:
+    from fides.api.main import app
```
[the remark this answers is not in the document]
TheAndrewJackson: update here: https://github.com/ethyca/fides/pull/4073/commits/2b8b7f1da4568c9c3140b1697003f4743e37fbf2

src/fides/api/api/v1/endpoints/config_endpoints.py
```diff
@@ -56,10 +57,17 @@ def update_settings(
...
-    update_config: ApplicationConfig = ApplicationConfig.update_api_set(
-        db, data.dict(exclude_none=True)
-    )
+    update_config: ApplicationConfig = ApplicationConfig.update_api_set(db, pruned_data)
+
+    if (
+        "security" in pruned_data.keys()
+        and "cors_origins" in pruned_data.get("security").keys()
+    ):
+        load_updated_cors_domains(pruned_data["security"]["cors_origins"])
```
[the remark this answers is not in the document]
TheAndrewJackson: That's great suggestion. So after this update `load_updated_cors_domains` will:

- accept a DB session and an app instance as parameters
- pull in the most up to date cors domains from `ConfigProxy` (which handles figuring out the config set and api set for us)
- update the current middleware domains based on what `ConfigProxy` returns

I'll update it to do this now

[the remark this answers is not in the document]
TheAndrewJackson: This was accomplished with a slight modification here: https://github.com/ethyca/fides/pull/4073/commits/ccbbfbc157c6a15ce6a372f79a0d31984e6dc1fd

The `load_updated_cors_domains` was moved onto `ConfigProxy` since the two were couple together. Now the DB session doesn't need to be passed in.

src/fides/config/utils.py
```diff
@@ -81,3 +84,15 @@ def get_dev_mode() -> bool:
         "active_default_storage_type",
     ],
 }
+
+
+def load_updated_cors_domains(domains: List[str], resetDomains=False) -> None:
```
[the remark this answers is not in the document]
TheAndrewJackson: I ended up removing the `resetDomains` block altogether. That was there from when the domains were being passed in as a param. Now that we're loading in whatever is in `ConfigProxy` there isn't really a need to "reset" the domains. We're always going to load what's coming in `ConfigProxy`.

Done here: https://github.com/ethyca/fides/pull/4073/commits/bd4cd1fc0f980219e2b41310d59fb424a72d2edb

src/fides/config/utils.py
```diff
@@ -81,3 +84,15 @@ def get_dev_mode() -> bool:
...
+    from fides.api.main import app
+
+    for mw in app.user_middleware:
+        if mw.cls is CORSMiddleware:
+            if resetDomains:
+                mw.options["allow_origins"] = domains
+                break
+
+            mw.options["allow_origins"] = [str(domain) for domain in domains]
```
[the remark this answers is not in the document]
TheAndrewJackson: Good callout, it's not. I removed `resetDomains` in this commit https://github.com/ethyca/fides/commit/bd4cd1fc0f980219e2b41310d59fb424a72d2edb.

The set idea makes sense. In the event that there's a delta between what's in the `CORSMiddleware` and the `ConfigProxy` that will make it so no domains get lost. I'll update it to use a set there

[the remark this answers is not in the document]
TheAndrewJackson: set applied here: https://github.com/ethyca/fides/pull/4073/commits/665a653f4e097d55ff4c6a51349596c4872a5c40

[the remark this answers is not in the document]
TheAndrewJackson: thanks! I feel the same. If an issue crops up we'll patch it 👍

TheAndrewJackson:
@adamsachs this should be good to go for another pass 👍

src/fides/config/config_proxy.py
```diff
@@ -85,3 +94,27 @@ def __init__(self, db: Session) -> None:
...
+                current_config_proxy_domains = (
+                    self.security.cors_origins
+                    if self.security.cors_origins is not None
+                    else []
+                )
+                patch_domains = set(
+                    [*current_middleware_domains, *current_config_proxy_domains]
+                )
+
+                mw.options["allow_origins"] = [str(domain) for domain in patch_domains]
```
adamsachs, line 120: i know i sorta pointed you toward this implementation on your previous iteration, but i hadn't thought of a consequence of this additive logic that now seems like it _could_ be problematic, unless i'm thinking about this wrong. 

with the way this is written, won't we never be able to _remove_ a domain that's been added at some point via config setting?  initially a domain will be added to the `allow_origins` list/set, and then wouldn't it essentially stay there forever, since it'd continue to be retrieved as a `current_middleware_domains` on all subsequent updates (until the service is restarted).

would there be a significant downside to switching the logic here to _replace_ the existing `allow_origins` list, so that it's exactly what's been specified in the `ConfigProxy`? i'm not sure if there's an initial set of `allow_origins` that we need to keep in place from the get-go...if so, i think this problem could be a bit tricky to solve. if not, then i feel like we'd be better off just _replacing_ the `cors_origins` with whatever's been specified in the `ConfigProxy`. 

i guess ultimately i'm wondering on the motivation for making this additive rather than just a replace. it seems to me like the additive logic is a bit at-odds with the way we set our config properties generally, which assumes a replace sort of functionality. i get worried that trying to reconcile those two can be a bit tricky.

sorry i hadn't thought of this on the previous iteration! but just wanna make sure we get this right 👍 

```suggestion
            if mw.cls is CORSMiddleware:
                current_config_proxy_domains = (
                    self.security.cors_origins
                    if self.security.cors_origins is not None
                    else []
                )
                
                mw.options["allow_origins"] = [str(domain) for domain in current_config_proxy_domains]
```
^ may not be exactly how you'd wanna code it, but just to illustrate what i think would lead to a bit more consistent and transparent behavior, if we can get away with it.

tests/ops/api/v1/endpoints/test_config_endpoints.py
```diff
@@ -539,6 +607,39 @@ def test_reset_application_config_non_existing(
...
+            url,
+            headers=auth_header,
+        )
+        assert response.status_code == 200
+        # Check the app config direcitly to make sure it's reset and
+        # by proxy check that the CORS domains are reset.
+        # The middleware isn't checked directly because it's
+        # possible that the config set will still have some 
+        # domains and when the `api_set` is reset the middleware
+        # will still have the `config_set` domains loaded in.
```
adamsachs, line 638: right so this is the behavior that i think would be desirable, but per my above comment i'm worried that this isn't actually exactly what's happening. happy to talk through this more because it gets a bit tricky (and perhaps counterintuitive) with the config proxy, but i actually think that if we're going for this behavior, we can accomplish it with what i suggested above 👍

adamsachs (commented):
@TheAndrewJackson the code re-org looks great! i think having that sort of side-effect function on the `ConfigProxy` could be a nice pattern generally.

i just have one concern now on the particular update logic for the cors origins property - i should have realized this before i'd suggested this logic on my previous review 🤦 

let me know if it's clear from my comments, i realize things are a bit tricky here and i would be more than happy to talk thru it, since it may be more efficient and i don't wanna hold you up on getting this across the line!

src/fides/api/alembic/migrations/versions/f17f92237383_disable_inactive_notices.py
```diff
@@ -13,7 +13,6 @@
 from sqlalchemy.engine import Connection, ResultProxy
 from sqlalchemy.sql.elements import TextClause
 
-
```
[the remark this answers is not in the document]
adamsachs: actually seeing this same change now on another PR so i'm assuming it's an auto-formatter thing that we're gonna keep on hitting across other PRs until we remove it, so i'm OK with commit this in this case, sorry for the unnecessary nit!

src/fides/config/config_proxy.py
```diff
@@ -85,3 +94,27 @@ def __init__(self, db: Session) -> None:
         self.notifications = NotificationSettingsProxy(db)
         self.execution = ExecutionSettingsProxy(db)
         self.storage = StorageSettingsProxy(db)
+        self.security = SecuritySettingsProxy(db)
+
+    def load_current_cors_domains_into_middleware(self, app: FastAPI) -> None:
```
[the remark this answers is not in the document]
TheAndrewJackson: me too!

src/fides/api/api/v1/endpoints/config_endpoints.py
```diff
@@ -78,19 +78,23 @@ def patch_settings(
...
+    pruned_data = data.dict(exclude_none=True)
+    logger.info("PUTing application settings")
     update_config: ApplicationConfig = ApplicationConfig.update_api_set(
         db,
-        data.dict(exclude_none=True),
+        pruned_data,
         merge_updates=False,
     )
+
+    ConfigProxy(db).load_current_cors_domains_into_middleware(request.app)
```
TheAndrewJackson, line 97: Nice catch!

adamsachs (approved):
BE is looking nice after all the iterations 😄 

i have a sneaking suspicion we'll be coming back to the `/config` API again sometime soon since i feel like edge cases abound (not the concern of this PR), but i feel comfortable this is serving our current purposes well 👍 

thanks for all the back and forth on this, i think we landed in a nice spot and are pushing some more generic functionality ⏩

PR #4078, opened 2023-09-13
Privacy notice table updates

### LINKED ISSUE #4077: Update privacy notice table
1. Hide descriptions & last update columns
2. Make locations a box with a number where # of locations > 1
3. Swap the locations of the last two columns
4. Ideally match style although that could be follow-up work

Kelsey-Ethyca:
Closes #4077 

### Description Of Changes

Privacy notice table updates - remove created and description columns, update location and consent mechanism tabs, and rearrange column order

### Code Changes

* [ ] update the privacy notice table and cells
* [ ] update the privacy notice response type so region and consent mechanism is not optional 

### Steps to Confirm

* [ ] confirm no descriptions & last update columns
* [ ] If locations > 1 then show a tag with # of locations
* [ ] if locations = 1 then show a tag with the location 
* [ ] column order should be title, mechanism, location, last update, status, enable 
* [ ] style matches design 

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### DIFF
--- a/clients/admin-ui/src/features/privacy-notices/cells.tsx
+++ b/clients/admin-ui/src/features/privacy-notices/cells.tsx
@@ -9,35 +9,72 @@ import { usePatchPrivacyNoticesMutation } from "~/features/privacy-notices/priva
 import { PrivacyNoticeResponse } from "~/types/api";
 
 export const MechanismCell = (
-  cellProps: CellProps<typeof MECHANISM_MAP, string>
-) => <MapCell map={MECHANISM_MAP} {...cellProps} />;
+  cellProps: CellProps<PrivacyNoticeResponse, string>
+) => {
+  const tagValue = MECHANISM_MAP.get(cellProps.row.original.consent_mechanism);
+  return (
+    <Badge
+      size="sm"
+      width="fit-content"
+      data-testid="status-badge"
+      textTransform="uppercase"
+      fontWeight="400"
+      color="gray.600"
+      px={2}
+    >
+      {tagValue}
+    </Badge>
+  )
+}
+{/* <MapCell map={MECHANISM_MAP} {...cellProps} />; */ }
 
 type TagNames = "available" | "enabled" | "inactive";
 
 const systemsApplicableTags: Record<TagNames, TagProps & { tooltip: string }> =
-  {
-    available: {
-      backgroundColor: "orange.100",
-      color: "orange.800",
-      tooltip:
-        "This notice is associated with a system + data use and can be enabled",
-    },
-    enabled: {
-      backgroundColor: "green.100",
-      color: "green.800",
-      tooltip: "This notice is active and available for consumers",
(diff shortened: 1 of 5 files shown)


clients/admin-ui/src/features/privacy-notices/cells.tsx
```diff
@@ -3,14 +3,30 @@ import React from "react";
...
 
 export const MechanismCell = (
-  cellProps: CellProps<typeof MECHANISM_MAP, string>
-) => <MapCell map={MECHANISM_MAP} {...cellProps} />;
+  cellProps: CellProps<PrivacyNoticeResponse, string>
+) => {
+  const { row } = cellProps;
+  const tagValue = MECHANISM_MAP.get(row.original.consent_mechanism);
+  return (
+    <Badge
```
allisonking, line 17: I think it might make sense to change `MapCell` to use this badge style instead of the tag it originally used. that seems to be the way the new designs are leaning. then you wouldn't need to make this change to `MechanismCell` and it can still be a wrapper around `MapCell`, and we wouldn't need to repeat this `<Badge />` code for locations either

`MapCell` is used in one other place, which is in the custom fields table but I feel like that could probably use this new design too?
image

if custom fields does need to stay the same, then we can still go with this solution, it's just not as dry as it could be :)

clients/admin-ui/src/types/api/models/PrivacyNoticeResponse.ts
```diff
@@ -16,8 +16,8 @@ export type PrivacyNoticeResponse = {
   description?: string;
   internal_description?: string;
   origin?: string;
-  regions?: Array<PrivacyNoticeRegion>;
-  consent_mechanism?: ConsentMechanism;
+  regions: Array<PrivacyNoticeRegion>;
+  consent_mechanism: ConsentMechanism;
```
allisonking, line 20: hmmm we shouldn't be modifying this file directly, since the next time we try to autogenerate types, it'll get overwritten. that said, these fields are _probably_ required, not optional, but the backend [does a workaround](https://github.com/ethyca/fides/blob/kt-privacy-notice-table-updates/src/fides/api/schemas/privacy_notice.py/#L20-L21) which I guess ends up making these fields optional. 

the proper solution would probably be for the backend to return the types properly, but in lieu of that, we should probably try to handle the optional cases (even though we know they won't be optional 🥲 )

allisonking (changes requested):

clients/admin-ui/src/features/privacy-notices/cells.tsx
Kelsey-Ethyca: You're right I reverted MechanismCell back to use MapCell and updated MapCell

allisonking (approved):
woot thanks for making those changes! the table looks great now!

Issue #4092, opened 2023-09-14
No error shown in UI when enabling/disabling a privacy notice errors


allisonking:
### Bug Description

There is no indication in the UI when enabling/disabling a privacy notice errors. Now that the backend checks if regions can overlap with privacy notices, it is possible to error if you enable a privacy notice that has the same notice key as another notice in the same region.

#### Steps to Reproduce

1. Add a privacy notice in `us_ca` with notice key `test`. Disable it.
1. Add another privacy notice in `us_ca` with notice key `test`. Enable it
2. Try to enable the first privacy notice you made
3. It won't enable, but you won't see an error either.

### Expected behavior

Show a toast that something has gone wrong!

### Screenshots

_If applicable, add screenshots to help explain your problem._

### Environment

* Version:
* OS:
* Python Version:
* Docker Version:

### Additional context

_Add any other context about the problem here._

pattisdr:
Moving to done looks great

PR #4095, opened 2023-09-14
Notice toggle error toast

allisonking:
Closes https://github.com/ethyca/fides/issues/4092

### Description Of Changes

image

New error toast for if an error comes up while enabling/disabling a privacy notice. Since this is a shared component, privacy experiences and custom fields also get the same upgrade!

### Code Changes

* [x] Inspect the result of the PATCH for errors, and render a toast if needed
* [x] Cypress test

### Steps to Confirm

* See steps in the issue

### Pre-Merge Checklist

* [x] Update `CHANGELOG.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,9 +17,6 @@ The types of changes are:
 
 ## [Unreleased](https://github.com/ethyca/fides/compare/2.20.0...main)
 
-### Fixed
-- Enable toggles in various tables now render an error toast if an error occurs [#4095](https://github.com/ethyca/fides/pull/4095)
-
 ## [2.20.0](https://github.com/ethyca/fides/compare/2.19.1...2.20.0)
 
 ### Added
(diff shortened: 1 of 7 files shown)


TheAndrewJackson (approved):
🪃

CHANGELOG.md
```diff
@@ -17,6 +17,9 @@ The types of changes are:
 
 ## [Unreleased](https://github.com/ethyca/fides/compare/2.20.0...main)
 
+### Fixed
+- Enable toggles in various tables now render an error toast if an error occurs [#4095](https://github.com/ethyca/fides/pull/4095)
```
pattisdr, line 21: If going out in 2.20 you could put this in the changelog below instead!

allisonking: ah right!

clients/admin-ui/src/features/privacy-notices/form.ts
```diff
@@ -21,27 +25,25 @@ export const defaultInitialValues: PrivacyNoticeCreation = {
 
 export const transformPrivacyNoticeResponseToCreation = (
   notice: PrivacyNoticeResponse
-): PrivacyNoticeCreation => {
-  // Remove the fields not needed for editing/creation
```
allisonking, line 25: this is the second time this bug has come up! I thought adding a test would fix it, but it turns out not since the text fixture didn't have the new field 🥲 

so no more removing fields aka opting out—we're moving to opt in only! in other words, we create a new object from scratch, rather than starting with an original object and removing the fields we think we don't need.

TheAndrewJackson: Good idea! opt-in should be more resilient

Issue #4094, opened 2023-09-14
Migration: Disable Notices that are not Systems Applicable


pattisdr:
### Is your feature request related to a specific problem?

Followup to: https://github.com/ethyca/fides/pull/4010

Disable privacy notices that are not applicable to at least one system

### Describe the solution you'd like

_A description of what you want to happen._

### Describe alternatives you've considered, if any

_A description of any alternative solutions or features you've considered._

### Additional context

_Add any other context or screenshots about the feature request here._

Roger-Ethyca:
testing

Roger-Ethyca:
moving to done

PR #4096, opened 2023-09-14
Fix: Disable Privacy Notices that are not Systems Applicable

pattisdr:
Closes https://github.com/ethyca/fides/issues/4094

❗ Needed for release 2.20 

### Description Of Changes
This migration gets notices in a better state for new UI.  We originally shipped with out of the box notices that were enabled by default, but new UI prevents you from enabling a notice unless it is applicable for a system.  So we have some "enabled" notices that are "inactive" which our new UI is supposed to prevent.

### Code Changes

* [ ] This data migration disables and bumps the version of enabled Privacy Notices that are not applicable for a current system. A new Privacy Notice History record is created to have a snapshot of the changes made here.

#### Get Environment in Expected State
- `nox -s dev -- shell`
- In admin `features.slice.ts` set plus=True, and run turbo run dev 
- Create one system with two declarations. One with a more specific use than our notice: `personalize.content` and one that matches: `analytics`
- In a postgres shell, run `update privacynotice set disabled=false` and `update privacynoticehistory set disabled=false` to mimic the state of a customer that is running fides when we used to ship with notices enabled by default
- Privacy Notices UI correctly shows Analytics and Functional notices enabled, but the other three are also enabled, but "inactive".  Our migration wants to set these enabled but "inactive" notices to be disabled.
Screenshot 2023-09-14 at 11 40 22 AM

#### Migrate

- ` cd src/fides/api/alembic`
-  `alembic downgrade -1`
- `alembic upgrade head`
- Privacy Notices UI should now show these 3 inactive notices as disabled and their versions have been bumped
Screenshot 2023-09-14 at 11 40 44 AM

```
fides=# select count(*) from privacynotice where disabled=true and version=2;
 count 
-------
     3
```
- Corresponding PrivacyNoticeHistory records were created:
```
fides=# select count(*) from privacynoticehistory where disabled=true and version=2;
 count 
-------
     3

```

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### DIFF
--- a/src/fides/api/alembic/migrations/versions/f17f92237383_disable_inactive_notices.py
+++ b/src/fides/api/alembic/migrations/versions/f17f92237383_disable_inactive_notices.py
@@ -1,0 +1,151 @@
+"""disable_inactive_notices
+
+Revision ID: f17f92237383
+Revises: 671358247251
+Create Date: 2023-09-14 13:35:38.474968
+
+"""
+import uuid
+from typing import Set
+
+from alembic import op
+from sqlalchemy import text
+from sqlalchemy.engine import Connection, ResultProxy
+from sqlalchemy.sql.elements import TextClause
+
+
+# revision identifiers, used by Alembic.
+from fides.api.models.sql_models import DataUse
+
+revision = "f17f92237383"
+down_revision = "671358247251"
+branch_labels = None
+depends_on = None
+
+
+def get_expanded_system_data_uses(bind: Connection) -> Set:
+    """Get data uses across all systems and those data uses' parents
+
+    Same results as System.get_data_uses(System.all(db), include_parents=True), but
+    we don't want to use the System Sqlalchemy model within a migration
+    """
+    system_data_uses: ResultProxy = bind.execute(
+        text("SELECT data_use from privacydeclaration;")
+    )
+
+    expanded_data_uses = set()
+    for row in system_data_uses:
+        data_use: str = row["data_use"]
+        # Expand system data uses so we also include their parents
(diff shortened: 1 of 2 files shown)


src/fides/api/alembic/migrations/versions/f17f92237383_disable_inactive_notices.py
```diff
@@ -0,0 +1,151 @@
...
+    bind.execute(
+        disable_privacy_notice_query,
+        {
+            "privacy_notice_id": privacy_notice_id,
+            "new_version": int(current_version) + 1,
+        },
+    )
+
+
+def create_corresponding_historical_record(
```
pattisdr, line 60: Rather than just disabling the current record , I am disabling, bumping the version, and creating a historical record, to keep a good history for consent reporting.  (This is what would happen if we disabled in the UI, too)

It is definitely possible preferences have been saved against enabled privacynoticehistory records that I am disabling.  If I just disable here, without any audit trail, consent reporting will look like we made it possible for a user to save a preference against a disabled notice.

adamsachs (approved):
looks good! if you've done the manual testing steps locally i'm good with merging this, we should get quick feedback as we deploy to staging envs 👍

src/fides/api/alembic/migrations/versions/f17f92237383_disable_inactive_notices.py
allisonking: wow, good thinking!!

pattisdr:
great, tested pretty thoroughly locally!

PR #4097, opened 2023-09-14
consider rc tags as latest release when generating beta tags

adamsachs:
Closes n/a

### Description Of Changes

A slight change to our tagging utility so that it takes into account existing `rc` tags when generating a `beta` tag on `main`.  So if we are in _the process_ of releasing e.g. `2.20.0` and therefore have `2.20.0rc0` but _not_ an official `2.20.0` tag, when we run `nox -s tag` on `main` it will generate `2.20.1b0` as the tag.

This makes it easy for us to ensure our packages built from the latest of `main` will always be _ahead_ of our published `rc` packages, which is important in keeping the `double-edge-slim` images (and therefore our `plus-nightly` environment!) based off of `main` of both repos.

As part of this, we'll also want to tweak our release process to generate a tag on `main` as soon as we cut release branches (i.e. beginning of the release process), rather than after we publish the release as we do now. 

The primary motivation here was to avoid `plus-nightly` from accidentally pulling in the `rc` versions of OSS `fides` during the release process, which has been happening and causing some confusion and unpredictability.

### Code Changes

* [x] when generating a `beta` tag (i.e. when on `main`) and determining the next release version increment, check for `rc` tags and evaluate their "release version", and compare it against the latest _official_ "release version" tag

### Steps to Confirm

* [ ] automated test coverage is solid here, otherwise we'll need to wait till the next release process to test live 👍 

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### DIFF
--- a/noxfiles/git_nox.py
+++ b/noxfiles/git_nox.py
@@ -4,12 +4,14 @@ import re
 from enum import Enum
 from typing import List, Optional
 
-import nox
 from packaging.version import Version
+
+import nox
 
 RELEASE_BRANCH_REGEX = r"release-(([0-9]+\.)+[0-9]+)"
 RELEASE_TAG_REGEX = r"(([0-9]+\.)+[0-9]+)"
 VERSION_TAG_REGEX = r"{version}{tag_type}([0-9]+)"
+RC_TAG_REGEX = r"(([0-9]+\.)+[0-9]+)rc([0-9]+)"
 GENERIC_TAG_REGEX = r"{tag_type}([0-9]+)$"
 
 INITIAL_TAG_INCREMENT = 0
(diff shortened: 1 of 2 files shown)


noxfiles/git_nox.py
```diff
@@ -115,9 +122,23 @@ def next_release_increment(session: nox.Session, all_tags: List) -> Version:
...
+    if treat_rc_as_release:
+        rc_releases = sorted(  # sorted by Version - what we want!
+            (
+                Version(tag.name)
+                for tag in all_tags
+                if re.fullmatch(RC_TAG_REGEX, tag.name)
+            ),
+            reverse=True,
+        )
+        latest_rc_release = rc_releases[0] if rc_releases else None
```
ThomasLaPiana, line 138: Is it checking for rc releases of only the latest version? In what case could it be `None`?

ThomasLaPiana (approved):
Good git tag wizardry 👍

noxfiles/git_nox.py
adamsachs: hmm no actually this'll get all `rc` tags, i guess the `None` is there just as a safeguard to prevent an unhandled error. pretty sure the real reason is that i just c+p'd the function from another area here where it was possible that we'd not have a match 😄 

if you think it's problematic it'd probably be OK to remove and have an unhandled error, it's not like we should ever hit this if things are generally working fine

adamsachs: gonna go ahead and merge as-is, lmk if you see a problem here though and i'm happy to address in a follow up 👍

ThomasLaPiana: definitely not worth holding up the PR! As a future note though if there is a case that you really don't expect and could potentially signify an issue, something like `assert rc_releases is not None` is helpful documentation to other developers that that really is an unexpected state

PR #4104, opened 2023-09-15
Updating DatasetTable to return all datasets

### LINKED ISSUE #4103: Not all datasets are shown in the dataset page
### Bug Description

The datasets created automatically during SaaS connector instantiation do not show up in the dataset page. This prevents the user from being able to edit data categories from SaaS connectors from the UI.

#### Steps to Reproduce

1. Create a new system and save
2. Navigate to the integrations tab and create any SaaS integration
3. Navigate to the Manage datasets page
4. Notice the dataset for the newly created connector isn't there

### Expected behavior

The dataset for the new connector should be visible
image

galvana:
Closes #4103

### Description Of Changes

Updating `DatasetTable` to display all datasets and not omit the SaaS datasets.

### Code Changes

* [ ] Renamed `DatasetTable` to match filename
* [ ] Updated dataset selector

### Steps to Confirm

* [ ] Create a system
* [ ] Add a SaaS integration
* [ ] Navigate to Dataset manager and verify the SaaS dataset is present

### Pre-Merge Checklist

* [ ] Update `CHANGELOG.md`


clients/admin-ui/src/features/dataset/DatasetTable.tsx
```diff
@@ -10,15 +10,13 @@ import { Dataset, GenerateTypes } from "~/types/api";
...
-  useGetAllFilteredDatasetsQuery,
+  useGetAllDatasetsQuery,
 } from "./dataset.slice";
 
-const DatasetsTable = () => {
+const DatasetTable = () => {
   const dispatch = useDispatch();
   const router = useRouter();
-  const { data: datasets } = useGetAllFilteredDatasetsQuery({
-    onlyUnlinkedDatasets: false,
```
TheAndrewJackson, line 20: Did the requirements change on this? I think we wanted this behavior when shipping this initially. Or was this filtering option accidentally placed where it shouldn't be?

What problem is this solving?

galvana: We ran into an issue where we wanted to be able to edit the data category for a dataset created from SaaS connector template being instantiated. Do you remember a reason why we wouldn't want to return those datasets in this table?

TheAndrewJackson: I do, the reason was we wanted to prevent users from editing SaaS connector template datasets 😅 At the time our (you and I) big push was to not let users change the datasets that are linked to SaaS template based connectors. Or not display the option to them in the UI rather. 

After that we established that, we talked about how some parts of a SaaS dataset are okay to modify but not all of them. Since that UI, AFAIK, doesn't limit a user to only editing the "safe" properties of a SaaS dataset we decided to prevent them from doing it all together until the UI was updated to restricting users to editing the "safe" properties for SaaS datasets.

Assuming, this is the same area of the codebase that I'm thinking it is and we're okay with users have the ability to potentially corrupt a SaaS template dataset then this should be good to ship. Otherwise the we'd need to update the UI to specifically handle SaaS datasets.

Issue #4108, opened 2023-09-15
No caching/prefetch available for regions without privacy notice(s)


SteveDMurphy:
### Bug Description

When not in a location that has a privacy notice, the experience is returned as an empty object. This empty object is not cached, resulting in the privacy experience endpoint being hit for any user coming from a region that doesn't have a privacy notice.

#### Steps to Reproduce

1. Use a prefetch enabled privacy center and `fides.js` via the CDN
1. Ensure the location you are in does not have a valid privacy notice
2. Go to the URL where fides.js is installed (i.e. Cookie House) and inspect the Network responses, filtering for `fides`
3. View the experiences and see that none are returned. Additionally, note that the cache cannot be used where no experience exists

### Expected behavior

Be able to use the cache for a region without a privacy notice.

### Screenshots

n/a

### Environment

* Version:
* OS:
* Python Version:
* Docker Version:

### Additional context

This was found while investigating an issue, it turned out the privacy notice itself had been disabled because the sole system/data use driving it had been removed. This could be another related symptom of not having a privacy notice returned by default

NevilleS:
Pretty sure it's just this behaviour: https://github.com/ethyca/fides/blob/0f74db23b3686215cde609075b61935426efe927/clients/fides-js/src/services/fides/api.ts#L61

For the purposes of prefetching, any `response.ok` should be treated as success, but since we're (effectively) mapping an successful result that found an empty list to `null`, it acts as if the prefetch failed.

Roger-Ethyca:
moving to done

PR #4113, opened 2023-09-18
Allows CDN to cache empty experiences responses from fides.js API

eastandwestwind:
Closes https://github.com/ethyca/fides/issues/4108

### Description Of Changes

Updates experience API to return empty obj instead of undefined when no experience was found for a specific location. This allows cache to work properly under this scenario, meaning that requests to the server will decrease overall.

### Code Changes

* [ ] Return empty obj for fides.js experience API when no experience was found
* [ ] Update e2e tests

### Steps to Confirm

* [ ] Use a prefetch enabled privacy center and fides.js via the CDN. Ensure that debug mode is on.
* [ ] Ensure the location you are in does not have a valid privacy notice
* [ ] Go to the URL where fides.js is installed (i.e. Cookie House) and inspect the console
* [ ] View the line that starts with `Got experience response from Fides API, returning:` and see that `{}` is returned instead of being undefined or null.
Screenshot 2023-09-18 at 5 02 57 PM

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### DIFF
--- a/clients/fides-js/src/fides.ts
+++ b/clients/fides-js/src/fides.ts
@@ -72,10 +72,11 @@ import {
 import {
   constructFidesRegionString,
   debugLog,
-  experienceIsValidAndHasNotices,
+  experienceIsValid,
   experienceHasNotices,
   transformConsentToFidesUserPreference,
   validateOptions,
+  isEmptyExperience,
 } from "./lib/consent-utils";
 import { dispatchFidesEvent } from "./lib/events";
 import { fetchExperience } from "./services/fides/api";
@@ -217,7 +218,7 @@ const init = async ({
     _Fides.geolocation = geolocation;
     _Fides.options = options;
     _Fides.initialized = true;
-    if (experienceHasNotices(experience)) {
+    if (experience && !isEmptyExperience(experience)) {
       // at this point, pre-fetched experience contains no user consent, so we populate with the Fides cookie
       updateExperienceFromCookieConsent(
         experience as PrivacyExperience,
@@ -254,7 +255,7 @@ const init = async ({
         `User location could not be obtained. Skipping overlay initialization.`
       );
       shouldInitOverlay = false;
-    } else if (!experienceHasNotices(experience)) {
+    } else if (!experience || isEmptyExperience(experience)) {
       effectiveExperience = await fetchExperience(
         fidesRegionString,
         options.fidesApiUrl,
@@ -263,7 +264,7 @@ const init = async ({
       );
     }
 
-    if (experienceIsValidAndHasNotices(effectiveExperience, options)) {
+    if (experienceIsValid(effectiveExperience, options)) {
       // Overwrite cookie consent with experience-based consent values
       cookie.consent = buildCookieConsentForExperiences(
         effectiveExperience as PrivacyExperience,
--- a/clients/fides-js/src/fides.ts
+++ b/clients/fides-js/src/fides.ts
@@ -74,10 +74,9 @@ import {
   constructFidesRegionString,
   debugLog,
   experienceIsValid,
-  experienceHasNotices,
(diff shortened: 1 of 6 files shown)


clients/fides-js/src/fides.ts
```diff
@@ -133,7 +134,7 @@ const automaticallyApplyGPCPreferences = (
   fidesRegionString: string | null,
   fidesApiUrl: string,
   debug: boolean,
-  effectiveExperience?: PrivacyExperience | null
+  effectiveExperience?: PrivacyExperience
```
allisonking, line 137: hmmmm to be really TS correct this should probably be `PrivacyExperience | {}`. I suspect that could cause other errors, but I think it'd be worth it since `{}` is kind of a tricky case (it's not falsy)

clients/fides-js/src/fides.ts
```diff
@@ -216,16 +217,20 @@ const init = async ({
     _Fides.geolocation = geolocation;
     _Fides.options = options;
     _Fides.initialized = true;
-    if (experience) {
+    if (experienceHasNotices(experience)) {
```
allisonking, line 220: is this basically to check if experience is not `{}`? if so, to keep this compatible with TCF, can we check that it's not `{}` (and that it is defined)? (TCF won't have notices!). I guess you can't really do `{} === {}` since those are technically different objects, but you could check if [it's an empty object](https://stackoverflow.com/questions/679915/how-do-i-test-for-an-empty-javascript-object)

if you're feeling _really_ fancy 🎩 you could use a TS predicate which might look something like...

```typescript
type EmptyExperience = {};
const isEmptyExperience = (obj: unknown): obj is EmptyExperience =>
  typeof obj === "object" && obj != null && Object.keys(obj).length === 0;
```

if you do some sort of check and change the TS type in the comment above, I don't think you'd need the cast to `PrivacyExperience` below

clients/fides-js/src/fides.ts
```diff
@@ -249,7 +254,7 @@ const init = async ({
         `User location could not be obtained. Skipping overlay initialization.`
       );
       shouldInitOverlay = false;
-    } else if (!effectiveExperience) {
+    } else if (!experienceHasNotices(experience)) {
```
allisonking, line 257: same thing here, perhaps we just compare to undefined and to `{}`?

allisonking (changes requested):
ah this is a tricky problem!! 

can we try to make the types stricter? I think once we make it stricter, some other pieces will fall into place, and we might not need some of the helper methods 🤞 of course, TS can often be tricky, so maybe my theory won't pan out, but I think it's worth trying, and will be less likely to bite us in the future

clients/fides-js/src/fides.ts
eastandwestwind: ah so I call the `experienceHasNotices` method before calling `automaticallyApplyGPCPreferences`, which will check for `Object.keys(experience).length >= 0`. Unless you think I should move that check internal to `automaticallyApplyGPCPreferences` instead?

clients/fides-js/src/fides.ts
eastandwestwind: oh thank you, I'd forgotten about TCF not using `notices`. We really have to get the TCF branch is `main` soon...

allisonking (approved):
🎉

clients/fides-js/src/lib/consent-utils.ts
```diff
@@ -26,6 +27,25 @@ export const debugLog = (
   }
 };
 
+/**
+ * Returns true if privacy experience is null or empty
```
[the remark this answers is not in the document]
allisonking: oh these comments still went through since I had them pending before we paired 😅 will resolve 'em now!

PR #4115, opened 2023-09-18
Pause manual erasure requests

galvana:
### Description Of Changes

Updated the privacy request workflow to pause any erasure requests with manual webhook connection configs before completing the request. The privacy request is flagged as `requires input` and waits for a user to mark which fields have been manually deleted before the request can be completed.

I kept the changes very light, I just made enough changes to pause the request in the UI and give the user a way to track which fields have been deleted. The manual erasure inputs are not used for any request execution. The changes are also not very DRY, often just erasure duplicates of existing manual access functionality. The requirements around this ticket were a bit loose (originally scoped as a defect) so I didn't want to integrate this too closely with the existing manual access workflow.

### Code Changes

* [ ] Updated the `run_privacy_request` function of `request_runner_service.py` to pause erasure privacy requests if there are any manual webhook connection configs enabled.
* [ ] Updated the privacy request detail page to use different side menus for manual access and erasure
* [ ] Added new endpoints to cache and retrieve manual erasure inputs

### Steps to Confirm

* [ ] Create a new system
* [ ] Add a `Manual Process` integration
* [ ] Click `Customize DSR` and add some fields
* [ ] Submit an erasure request via the Privacy Center
* [ ] Navigate to the Privacy request tab in the Admin UI and verify that the privacy request is in a `Requires Input` state, either immediately of after clicking `approve` (based on your test environment's settings).

### Pre-Merge Checklist

* [ ] Update `CHANGELOG.md`


src/fides/api/models/privacy_request.py
```diff
@@ -637,11 +637,27 @@ def cache_manual_webhook_input(
         parsed_data = manual_webhook.fields_schema.parse_obj(input_data)
 
         cache.set_encoded_object(
-            f"WEBHOOK_MANUAL_INPUT__{self.id}__{manual_webhook.id}",
+            f"WEBHOOK_MANUAL_ACCESS_INPUT__{self.id}__{manual_webhook.id}",
             parsed_data.dict(),
         )
 
-    def get_manual_webhook_input_strict(
+    def cache_manual_webhook_erasure_input(
```
adamsachs, line 644: i know we're following the example from the access input as a first pass, and i think that's probably fine. but i'd want to make sure we flag this, since i could see it leading to problems if users (or our product team) doesn't realize the implication here.

specifically, there's no sort of persisted record of the confirmation that the records have been deleted from the manual integration. there'd be no way of tracking that admin X confirmed at time Y that person Z's records were deleted from system A (whoops ran out of letters) -- at least, not one that we can rely on since we have to treat the cache as ephemeral. i think it's true that we don't persist directly in our db the manual input data for access requests (though maybe we should?) but it's a bit different there because that input is packaged up as part of the access result package and that package is put into storage and/or sent back to the end user -- so it _is_ persisted somewhere.

this is certainly not listed out as an explicit requirement here, but it feels like a reporting/audit requirement that could potentially have legal implications if/when this gets implemented. not saying we need to solve it here in a first pass, i'd just want this to be flagged before we release!

src/fides/api/models/privacy_request.py
```diff
@@ -668,15 +684,44 @@ def get_manual_webhook_input_strict(
...
+        """
+        cached_results: Optional[Dict[str, Any]] = _get_manual_erasure_input_from_cache(
+            privacy_request=self, manual_webhook=manual_webhook
+        )
+
+        if cached_results:
+            data: Dict[str, Any] = manual_webhook.erasure_fields_schema.parse_obj(
+                cached_results
+            ).dict(exclude_unset=True)
+            if set(data.keys()) != set(
```
adamsachs, line 705: i could be following wrong, but is there a narrow (and perhaps too contrived) edge case where the value has been set to `False` for a given field (i.e. someone actually checks and unchecks a field) and we essentially judge this as OK, we're ready to proceed, when in reality they still have one or more fields outstanding? maybe that's a desired behavior, i.e. if an admin has explicitly unchecked a box then its up to their discretion to allow the request to proceed...

anyway, sorta just curious whether that's a legitimate possibility and if so whethere you'd thought thru it at all...

src/fides/api/api/v1/endpoints/privacy_request_endpoints.py
```diff
@@ -1381,18 +1382,67 @@ def upload_manual_webhook_data(
...
+    )
+
+
+@router.patch(
+    PRIVACY_REQUEST_MANUAL_WEBHOOK_ERASURE_INPUT,
+    status_code=HTTP_200_OK,
+    dependencies=[Security(verify_oauth_client, scopes=[PRIVACY_REQUEST_UPLOAD_DATA])],
+    response_model=None,
+)
+def upload_manual_webhook_erasure_data(
```
adamsachs, line 1410: i know you specifically mentioned not DRYing this code up, and that's fine, but this is basically identical to the above endpoint besides `cache_manual_webhook_erasure_input` instead of `cache_manual_webhook_access_input` and then one word differences in log messages. IMO it's worth some minimal effort on DRYing this up for readability and so that we can be sure not to have any drift if/when any of this logic needs to be updated.

adamsachs (approved):
nice work @galvana , looks good for a first pass! like you said, def some code consolidation opportunities but that's fine to defer, especially since i think this is still pretty straightforward to follow - though there's one i think it'd be worth addressing for now.

there are a couple other points that i'd just like to make sure we've considered and/or are being communicated to product (and potentially documented in release notes/user guides), especially the one about persisting the "confirmation" signal.

src/fides/api/api/v1/endpoints/privacy_request_endpoints.py
galvana: Good point about DRYing this portion up to clearly show these two endpoints are related. I extracted the shared logic into a function so now we can just do this for each endpoint
```
    _handle_manual_webhook_input(
        action="access",
        connection_config=connection_config,
        privacy_request_id=privacy_request_id,
        db=db,
        input_data=input_data,
    )
```

src/fides/api/models/privacy_request.py
galvana: I think this is ok, but that's only because it's hard for us to determine when it's acceptable to proceed with a manual erasure. There are cases when certain fields might not exist in the manual data source so we'll still want to proceed even if no fields are checked.

src/fides/api/models/privacy_request.py
Kelsey-Ethyca: We do not need a persisted record of the confirmation since a successfully completed deletion requires the confirmation we can infer that a confirmation was made. CC @rsilvery

src/fides/api/models/privacy_request.py
adamsachs: OK fair enough, i mean i think there's a bit of tension between that and what we're saying [here](https://github.com/ethyca/fides/pull/4115#discussion_r1337649410), i.e. that completed deletion infers a confirmation was made. if we're allowing deletion to proceed without requiring confirmation of all fields, someone who is auditing things really has no way of determining what was deleted on a particular DSR request.

totally acknowledge that this is an edge case and the requirement generally is perhaps beyond MVP scope. i just think that a robust/long term implementation probably includes some sort of persisted tracking of what was confirmed on a particular DSR 👍

src/fides/api/models/privacy_request.py
adamsachs: OK, that's fine for me in terms of an MVP, but there's some tension with the other piece of this implementation as noted [here](https://github.com/ethyca/fides/pull/4115#discussion_r1337661024). 

maybe we can ticket for a follow up, just to track it at least?

src/fides/api/api/v1/endpoints/privacy_request_endpoints.py
adamsachs: awesome, thank you @galvana!

src/fides/api/models/privacy_request.py
galvana: Placeholder ticket https://github.com/ethyca/fides/issues/4164

PR #4116, opened 2023-09-18
Add fields to System PUT payload 

Kelsey-Ethyca:
### Code Changes

* [ ] add organization_fides_key, registry_id, meta, fidesctl_meta, and dpa_progress to system api so DB is not overwritten with null values

### Steps to Confirm

* [ ] create systems using Data flow scan 
* [ ] update a system 
* [ ] using the network tab see the System PUT call's payload 
* [ ] verify these fields are in the payload organization_fides_key, registry_id, meta, fidesctl_meta, and dpa_progress
* [ ] meta should be populated since the scanner was used `"meta": { "kubernetes_namespace": "plc" }`

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### DIFF
--- a/clients/admin-ui/src/features/system/form.ts
+++ b/clients/admin-ui/src/features/system/form.ts
@@ -102,6 +102,11 @@ export const transformFormValuesToSystem = (formValues: FormValues): System => {
     vendor_id: formValues.vendor_id,
     ingress: formValues.ingress,
     egress: formValues.egress,
+    meta: formValues.meta,
+    fidesctl_meta: formValues.fidesctl_meta,
+    registry_id: formValues.registry_id,
+    organization_fides_key: formValues.organization_fides_key,
+    dpa_progress: formValues.dpa_progress
   };
 
   if (!formValues.processes_personal_data) {


clients/admin-ui/src/features/system/form.ts
```diff
@@ -102,6 +102,11 @@ export const transformFormValuesToSystem = (formValues: FormValues): System => {
     vendor_id: formValues.vendor_id,
     ingress: formValues.ingress,
     egress: formValues.egress,
+    meta: formValues.meta,
+    fidesctl_meta: formValues.fidesctl_meta,
+    registry_id: formValues.registry_id,
```
pattisdr, line 107: We have a weird thing with the registry_id looks like. I have no idea what this field is for, but validation expects an integer, but database needs it to be a string.  If FE is sending an integer here, that will fail at the db level.  Maybe we leave out this one.

pattisdr (approved):
thanks for addressing @Kelsey-Ethyca!

clients/admin-ui/src/features/system/form.ts
ThomasLaPiana: registries are an extremely old and never-used feature...honestly should just remove it

It was conceived as a way to help organizations organize their fides resources into sub-units. Specifically you could assign systems to a registry to logically "group" them. This use-case has never materialized

pattisdr: ah okay thanks for the background Thomas!

PR #4117, opened 2023-09-19
add a global async scheduler

adamsachs:
Closes n/a

### Description Of Changes

As part of https://github.com/ethyca/fidesplus/pull/1087, we'd like to put an async function on a scheduler. i had been working around this by wrapping the function with a `@sync` decorator but i think a cleaner solution is to just instantiate, start, and expose a global `AsyncIOScheduler` here in `fides`, alongside the `BackgroundScheduler` we already expose.

I could have put this directly in `fidesplus`, but it feels like something we may want to make more use of generally if/when we need to schedule other async functions, and generally feels like a better design to define and manage this directly alongside the `BackgroundScheduler` here in `fides` 👍 

### Code Changes

* [x] instantiate, expose and start an `AsyncIOScheduler` that can be used globally to schedule async functions
* [x] also update our `conftest.py` to start the scheduler in the test runtime to be comprehensive, even if we are not making use of it there yet

### Steps to Confirm

* [ ] will just need to confirm that this works over on `fidesplus`, but i did confirm that an `AsyncIOScheduler` defined directly there (just as a test!) was successful for my purposes here

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### DIFF
--- a/src/fides/api/main.py
+++ b/src/fides/api/main.py
@@ -32,7 +32,7 @@ from fides.api.schemas.analytics import Event, ExtraData
 from fides.api.service.privacy_request.email_batch_service import (
     initiate_scheduled_batch_email_send,
 )
-from fides.api.tasks.scheduled.scheduler import scheduler
+from fides.api.tasks.scheduled.scheduler import async_scheduler, scheduler
 from fides.api.ui import (
     get_admin_index_as_response,
     get_path_to_admin_ui_file,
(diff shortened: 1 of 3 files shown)


pattisdr (approved):
On a high level, this seems fine!  Didn't do any testing but trust your plus scheduling test is sufficient -

PR #4120, opened 2023-09-19
avoid n+1 queries on bulk `GET /system` endpoint

### LINKED ISSUE #4028: improve `GET /system` API performance with many integrations

@Roger-Ethyca has [noticed](https://ethyca.slack.com/archives/C04S7K0T2V7/p1693940640080529) that the "view systems" admin UI tab is slower to load as he adds more integrations to his local fides instance.

Following up on this observation, i noticed that the response time for the`GET /system` endpoint seems to increase linearly with the # of integrations configured on the server, with ~500ms of response time added per configured integration (i tested up to 8 integrations): https://docs.google.com/spreadsheets/d/1mDo7XDRmB4xUYWZXAaeWeLU0oEjeFHJIcg6ou_NfSm4/edit#gid=0. this isn't really a sustainable trend - it's not inconceivable that a fides user would have ~100 integrations configured on their server!

I'm _pretty_ sure that our [lazy loading](https://docs.sqlalchemy.org/en/14/orm/loading_relationships.html#lazy-loading) of the [`connection_config` relationship on a `System`](https://github.com/ethyca/fides/blob/main/src/fides/api/models/sql_models.py#L394-L400) is effectively causing an n+1 query problem when it's applied to the `GET /system` endpoint that lists _all_ `System` records, since the `connection_config` population requires an additional query for each `System`.

I don't think we need the `connection_configs` relationship/object field populated on the `GET /system` endpoint response at all, since we're just trying to get the basic system records for all systems - so the simplest approach would probably be just to ensure the response schema for that endpoint _doesn't_ include the `connection_config` field.

I quickly POC'd locally this quick and dirty fix by swapping out the endpoint's `response_model` [reference](https://github.com/ethyca/fides/blob/main/src/fides/api/api/v1/endpoints/system.py#L364) to `fides.api.schemas.system.SystemResponse` with `fides.api.schemas.system.System` (which does _not_ have the `connection_configs` field defined) and it reduced the 8-integration response time from ~4s to ~200ms so it seems effective!

Slightly more complex solutions that i can think of, but i'm not sure are worth the added complexity:
- provide a query param/option for returning the embedded `connection_config` relationships on the response records of this endpoint
- improving the `connection_config` SQLAlchemy relationship so that it generates a more performant query pattern

### Additional context

See slack thread for some more context!

adamsachs:
Closes #4028 

### Description Of Changes

Narrow down what we return on the bulk `GET /system` endpoint to not have relationship fields that can lead to n+1 query problems - in this case, it was specifically the `connection_configs` property/relationship. See issue for more context.

### Code Changes

* [ ] define a narrower `BasicSystemResponse` model without any relationship-backed fields, used for the bulk `GET` endpoint. 

### Steps to Confirm

* [ ] ran locally with `nox -s "fides_env(test)"` with 4 systems, i added an additional integration onto one of those systems (i think two systems come seeded with integrations already).
    * [ ] without the change, i was seeing the `GET /system` endpoint take ~1s to respond
    * [ ] with the change, that was down to <200ms
* [ ] confirmed that `connection_configs` was no longer being returned on the bulk `GET /system` endpoint response records
* [ ] confirmed that all admin UI around this still seemed to be functional 👍 

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ The types of changes are:
 
 ### Fixed
 - Allows CDN to cache empty experience responses from fides.js API  [#4113](https://github.com/ethyca/fides/pull/4113)
+- Avoid un-optimized query pattern in bulk `GET /system` endpoint [#4120](https://github.com/ethyca/fides/pull/4120)
 
 ## [2.20.0](https://github.com/ethyca/fides/compare/2.19.1...2.20.0)
 
--- a/src/fides/api/api/v1/endpoints/system.py
+++ b/src/fides/api/api/v1/endpoints/system.py
@@ -50,7 +50,7 @@ from fides.api.schemas.connection_configuration.connection_secrets import (
 from fides.api.schemas.connection_configuration.saas_config_template_values import (
     SaasConnectionTemplateValues,
 )
-from fides.api.schemas.system import SystemResponse
+from fides.api.schemas.system import BasicSystemResponse, SystemResponse
 from fides.api.util.api_router import APIRouter
 from fides.api.util.connection_util import (
     connection_status,
@@ -361,7 +361,7 @@ async def create(
             scopes=[SYSTEM_READ],
         )
     ],
-    response_model=List[SystemResponse],
+    response_model=List[BasicSystemResponse],
     name="List",
 )
 async def ls(  # pylint: disable=invalid-name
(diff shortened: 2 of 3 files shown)


galvana (approved):
The system management page can still access the fields it needs, looks good to me!

pattisdr (approved):
nice one

Issue #4046, opened 2023-09-08
TCF mode does not prefetch properly


allisonking:
### Bug Description

When both TCF and prefetching are enabled, fides.js does not populate user consents properly due to changes in https://github.com/ethyca/fides/pull/4009.

#### Steps to Reproduce

This would be the same bug as https://github.com/ethyca/fides/issues/4001 but for TCF. Since the solution was to not prefetch with the `fides_device_user_id` and instead rely on the cookie for user preferences, the function `updateExperienceFromCookieConsent` needs to adapt to handle TCF as well.

This is nontrivial because we need to parse the TCF string, then assign values accordingly. This would also involve a slight refactor since we do not want TCF parsing logic in our vanilla fides-js code (since that uses a 3rd party lib which would increase bundle size).

### Expected behavior

User consents to persist 

### Screenshots

_If applicable, add screenshots to help explain your problem._

### Environment

* Version:
* OS:
* Python Version:
* Docker Version:

### Additional context

https://ethyca.slack.com/archives/C05E74TR0BB/p1694184770802259

Issue #4011, opened 2023-08-31
Write tests for new updateExperienceFromCookieConsent method in cookie.ts


eastandwestwind:
https://github.com/ethyca/fides/issues/4001 add a `updateExperienceFromCookieConsent` method in `clients/fides-js/src/lib/cookie.ts` that requires unit tests in `clients/fides-js/__tests__/lib/cookie.test.ts`

allisonking:
Going to tackle this as part of https://github.com/ethyca/fides/issues/4046 since that one edits this function quite a bit and will also need tests!

Roger-Ethyca:
moving to done

PR #4124, opened 2023-09-20
Initialize prefetched TCF experience preferences from cookie

allisonking:
Closes https://github.com/ethyca/fides/issues/4046
Closes https://github.com/ethyca/fides/issues/4011

### Description Of Changes

Makes prefetch work for TCF—the thing that was making this not work was that we weren't saving TCF preferences to a cookie, and so we couldn't initialize a user's preferences from a cookie (since a prefetched experience won't have a user's individual preferences). 

### Code Changes

* [x] Adds function to help save TCF preferences to a cookie
* [x] Updates assigning cookie values to an experience object function handle the TCF parts
* [x] Re-enable prefetching with TCF in the privacy center endpoint (there was a protective if statement that made it not possible to do both prefetch and TCF)
* [x] Unit tests
* [x] Cypress tests

### Steps to Confirm

* Add the following to a `.env` file at the root of your `fides` repo (I'm actually not sure that _all_ of these are needed, but this works!)
```sh
FIDES_SAMPLE_APP__PRIVACY_CENTER_URL="http://localhost:3001"
FIDES_PRIVACY_CENTER__TCF_ENABLED=true
FIDES_PRIVACY_CENTER__GEOLOCATION_API_URL="https://cdn-api.ethyca.com/location"
FIDES_PRIVACY_CENTER__PRIVACY_CENTER_URL="http://localhost:3001"
```
* Run `nox -s "fides_env(test)"
* Once everything comes up, visit `localhost:8080/docs` and turn TCF on via the PATCH to `consent-settings`. You'll need to authorize with [these creds ](https://github.com/ethyca/fides/blob/aking/prod-1063/tcf-prefetch/src/fides/data/sample_project/fides.toml/#L19-L20) first
image
* Now that TCF is enabled, visit `localhost:8080` and edit one of the existing systems.
* Add a data use such as `analytics.reporting.campaign_insights` (which is one of the TCF data uses). Give it a legal basis of "Consent" and save.
* Now visit cookie house at `localhost:3000` and click on "Frankfurt" from the geolocation dropdown. 
* Open the network tab and refresh the page. You should _not_ see a call to `/privacy-experience` since it is being prefetched by the privacy center! But you should still see the banner pop up
* Click on "manage preferences" in the banner, then "manage preferences" again to move past the initial layer. Toggle the the purpose `Understand audiences through statistics or combinations of data from different sources` ON
* Refresh the page. The banner should not pop up this time since there is nothing new to agree to
* Open the modal by clicking "Manage preferences"  at the bottom and "manage preferences" again in the modal to get to the second layer
* Your preference should still be saved (opted in)
* You can also look at your saved cookie, which should look something like
```
{
  "consent": {},
  "identity": {
    "fides_user_device_id": "9a82a38d-f5d4-477e-92c0-79e4233e2a2e"
  },
  "fides_meta": {
    "version": "0.9.0",
    "createdAt": "2023-09-21T15:21:31.361Z",
    "updatedAt": "2023-09-21T15:21:45.363Z"
  },
  "tcfConsent": {
    "purpose_preferences": { "9": true },
    "special_purpose_preferences": {},
    "feature_preferences": {},
    "special_feature_preferences": {},
    "vendor_preferences": {},
    "system_preferences": { "ctl_af18ae4b-8c7d-47e7-b3db-a13e5c319662": true }
  },
  "tcString": "CPyc0sAPyc0sAAMABBENASEAAACAAAAAAAAAAAAAAAAA.IgoMV_H3_bX9v8Tq16ft0eYFX99j77uQxBhfJs-4FyDvW_JwVn2EyNAUytqYKmRIEuzZBIQFsHJnURVChQogVryHsYkGMhTEAJ-BkgHMRa2dYCFxvmYtjEQIZ5_p_czf52R_9_dn-3czjT9lnnzVtO2dlcKidS59tH_l_bRKb-7Ied-7W_4v09tfrkW_eTVu_9evvp9-u_s_v__934AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAQ"
}

```

### Pre-Merge Checklist

* [x] Update `CHANGELOG.md`


clients/fides-js/__tests__/lib/cookie.test.ts
```diff
@@ -323,3 +338,289 @@ describe("removeCookiesFromBrowser", () => {
...
+
+describe("updateExperienceFromCookieConsent", () => {
+  const baseCookie = makeFidesCookie();
+
+  // Notice test data
+  const notices = [
+    { notice_key: "one" },
+    { notice_key: "two" },
+    { notice_key: "three" },
+  ] as PrivacyExperience["privacy_notices"];
```
allisonking, line 398: I normally try to avoid this type of casting if possible, but I've found it's super handy for tests where a func might only care about a few attributes so we only have to build mock data to cover those attrs

clients/fides-js/src/fides.ts
```diff
@@ -83,7 +83,13 @@ const init = async (config: FidesConfig) => {
     dispatchFidesEvent("FidesInitialized", cookie, config.options.debug);
     dispatchFidesEvent("FidesUpdated", cookie, config.options.debug);
   }
-  const updatedFides = await initialize({ ...config, cookie, renderOverlay });
+  const experience = initialFides?.experience ?? config.experience;
```
allisonking, line 86: this is actually kind of a consequential change—basically what was originally happening was...

1. `getInitialFides` returns an object which may or may not have an updated experience based on cookie values. we'll call the updated experience `updatedExperience`
2. `_Fides`'s `experience` object is updated to `updatedExperience`
3. `initialize` is called with **the original experience object**

I believe that's a bug and we actually want to pass in the `updatedExperience` to `initialize`. Luckily, I think this only affects TCF since the notice overlay already builds its preferences off of cookies, not the experience object. But the TCF overlay builds it off of the experience object.

clients/fides-js/src/lib/cookie.ts
```diff
@@ -252,30 +258,75 @@ export const buildCookieConsentForExperiences = (
...
+    tcf_purposes: experience.tcf_purposes,
+    tcf_special_purposes: experience.tcf_special_purposes,
+    tcf_features: experience.tcf_features,
+    tcf_special_features: experience.tcf_special_features,
+    tcf_vendors: experience.tcf_vendors,
+    tcf_systems: experience.tcf_systems,
+  };
+
+  if (cookie.tcfConsent) {
+    TCF_COOKIE_KEY_TO_EXPERIENCE_KEY.forEach(({ cookieKey, experienceKey }) => {
```
allisonking, line 291: this is a little painful because we have a `PrivacyExperience` object (what we get from the server), a `PrivacyPreferencesRequest` object (what we save to the server), and now also a cookie object. `PrivacyExperience` has one set of keys (`tcf_purposes`, `tcf_features`, etc) and `PrivacyPreferencesRequest` has another set of keys (`purpose_preferences`, `feature_preferences`, etc). I chose to make the cookie object follow the `PrivacyPreferencesRequest` keys, so the cookie looks like

```js
tcfConsent: {
  purpose_preferences: { 1: true }
}
```

because the `PrivacyExperience` keys and the `PrivacyPreferencesRequest` keys don't map to each other, I added this extra `TCF_COOKIE_KEY_TO_EXPERIENCE_KEY` array which essentially defines the mapping. Convincing TS the keys do work is not super fun—the alternative is to do each mapping individually, but then we'd have quite a bit of repeated code. If we wanted the cookie to be a separate key (i.e. `purposes` instead of `purpose_preferences`) we'd need another mapping.

We should probably decide if this cookie key looks okay though, since it is an API of sorts that a user could interact with

clients/privacy-center/public/fides-js-demo.html
```diff
@@ -4,7 +4,7 @@
     <title>fides-js script demo page</title>
     <meta name="viewport" content="width=device-width, initial-scale=1" />
 
-    <script src="/fides.js"></script>
+    <script src="/fides-js"></script>
```
allisonking, line 7: this allows testing against the PC endpoint instead of just the bundle (which we need in order for prefetch to happen!)

clients/fides-js/src/lib/cookie.ts
```diff
@@ -52,7 +56,8 @@ export interface FidesCookie {
   consent: CookieKeyConsent;
   identity: CookieIdentity;
   fides_meta: CookieMeta;
-  tcString?: string;
+  tc_string?: string;
```
allisonking, line 59: I realized we had a mix of camel and snake case in this cookie... I went with snake case since that makes things a little easier to work with on the tc string side. `createdAt` and `updatedAt` in `fides_meta` are still camel case though, but I think we'd have to do a migration to change those and that doesn't sound fun! I figured I'd change `tcString` to snake case now, before anyone is using this feature yet

clients/privacy-center/public/fides-js-demo.html
eastandwestwind: wait, could you explain this a bit more? I thought `fides.js` includes the prefetch code as part of the bundle. Since our customers use `fides.js` instead of calling the PC endpoint directly I want to ensure we're best mimicking their setup on our `fides-js-demo.html` page

eastandwestwind (commented):
Overall looks good 👍 , main Q is around the switch from `/fides.js` to `/fides-js` on our test site. I want to be sure I understand that

clients/privacy-center/public/fides-js-demo.html
allisonking: you know what, you're right... I could've sworn it wasn't working before I changed it to `fides-js`, but now that I think about it, that doesn't really make sense either. I'll revert!

allisonking: okay I fixed, and also added a treat where you can now add a query param to force a geolocation (similar to what we do for cookiehouse) https://github.com/ethyca/fides/pull/4124/commits/cc1c94d11e5bbe64e36aa2e5dbd6c7dfcba84b26

clients/privacy-center/public/fides-js-demo.html
```diff
@@ -4,8 +4,19 @@
...
+      Pass along any params to the fides.js script. For example, visiting
+      http://localhost:3000/fides-js-demo.html?geolocation=fr-id
+      will pass a geolocation query param to fides.js
+    -->
+    <script>
+      (function () {
+        const params = window.location.search;
+        const src = `/fides.js${params ?? ""}`;
+
+        document.write('<script src="' + src + '"><\/script>');
```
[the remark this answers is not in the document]
allisonking: I think this should be ok since it's a test page but maybe I'm not being paranoid enough!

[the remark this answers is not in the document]
eastandwestwind: Hm, since this is a public-facing page, I think we should treat this as legit and be extra careful. I'm unable to perform a XSS attack manually, but I think we should at least use our `VALID_ISO_3166_LOCATION_REGEX` regex here.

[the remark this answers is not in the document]
allisonking: good point, let me try that!

eastandwestwind (approved):

PR #4128, opened 2023-09-21
Removing the connector template properties

### LINKED ISSUE #4123: `GET /connection` endpoint seems slow to respond with many connections present
### Bug Description

I'm seeing on our plus-rc environment that the `GET /connection` endpoint is taking an exceedingly long time to respond, up to ~10 seconds. Some of our admin UI relies on this endpoint, e.g. the legacy "connection manager" page but also the "link integration" button on the "integrations" tab of a given system edit page. neither of these UI elements are particularly "mission critical", so we're probably not severely impacting most user experience currently, but any long-running API call is a potential problem lurking!

the specific API call that the UI is making to populate the "link integration" button is `GET api/v1/connection?size=25&page=1&orphaned_from_system=true`, but i've tried the same API call _without_ the `orphaned_from_system` parameter and it actually _increases_ latency. 

one guess is that it's a similar problem to https://github.com/ethyca/fides/issues/4028, and that the relationships on the `ConnectionConfig` model are somehow causing the latency here.

#### Steps to Reproduce

1. create many (5? 10? not sure...) connections/integrations on your fides instance
2. try hitting the above endpoint or going to the admin UI pages mentioned above

### Expected behavior

We should have a more performant endpoint. haven't confirmed this is strictly tied to the number of connections, but seems like it may be, in which case our endpoint latency shouldn't scale linearly with the number of connections we have!

One potential solution to try is to narrow down the response model on this endpoint so that it doesn't include all the relationships it has currently. That's what we did to solve https://github.com/ethyca/fides/issues/4028. It's a bit of a workaround and maybe not getting at the core of the inefficiencies, but could be a good near-term mitigation. Still TBD if that is effective in this case though!

### Screenshots

### Environment

* Version:
* OS:
* Python Version:
* Docker Version:

### Additional context

Found when testing fix for https://github.com/ethyca/fides/issues/4028, funny enough. But i don't think it's strictly related in any way.

galvana:
Closes #4123

### Description Of Changes

Removing the `authorization_required` and `user_guide` properties from `ConnectorTemplate`. Calculating these values on the fly required the parsing of the SaaS config YAML and caused other threads to block.

### Code Changes

* [ ] Refactored the ConnectorTemplate model

### Steps to Confirm

* [ ] Pull up the `Salesforce Classic` connector in the System Integration tab, this connector has `authorization_required` and `user_guide`, The page should show an `Authorize integration` button and have a link to external documentation.

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`


src/fides/api/service/connectors/saas/connector_registry_service.py
```diff
@@ -147,11 +160,23 @@ def _register_template(
         Registers a custom connector template by converting it to a ConnectorTemplate
         and adding it to the loader's template dictionary.
         """
+
+        config = SaaSConfig(**load_config_from_string(template.config))
+        authentication = config.client_config.authentication
+        authorization_required = (
+            authentication.strategy
```
ThomasLaPiana, line 167: This was a bit hard for me to read because it's doing a nested comparison instead of using `and`, I think something like this might be a bit more readable:

```
In [1]: foo = None

In [2]: bar = True if foo and foo.test else False
```

src/fides/api/service/connectors/saas/connector_registry_service.py
```diff
@@ -71,6 +74,14 @@ def _load_connector_templates(self) -> None:
                 config_dict = load_config(config_file)
                 connector_type = config_dict["type"]
                 human_readable = config_dict["name"]
+                user_guide = config_dict.get("user_guide")
+                authentication = config_dict["client_config"].get("authentication")
+                authorization_required = (
```
ThomasLaPiana, line 79: same not as below

src/fides/api/service/connectors/saas/connector_registry_service.py
galvana: Good call!

ThomasLaPiana (approved):
LGTM, not sure if external/unsafe checks need to be run though

Issue #3957, opened 2023-08-22
Dynamically send TCF bundle based on location


allisonking:
### Is your feature request related to a specific problem?

We currently have a `/consent-settings` endpoint which allows turning TCF on/off. With the new [TCF bundling work](https://github.com/ethyca/fides/pull/3859), the privacy center needs to know whether TCF is on/off in order to determine whether to serve `fides.js` or `fides-tcf.js`. As a temporary measure, we use an environment variable `FIDES_PRIVACY_CENTER__TCF_ENABLED=true turbo dev` to run the privacy center. However, this should really be derived from the backend instead.

### Describe the solution you'd like

The backend `/consent-settings` endpoint is not currently public, which I believe we would need in order for the privacy center to be able to ping it (unless the privacy center has some oauth? I don't think it currently does, but I might be wrong). 

If `/consent-settings` becomes public, then the privacy center should be able to prefetch the settings like we do with experience and geolocation. 

This approach would involve both backend and frontend work.

### Describe alternatives you've considered, if any

Continue using an environment variable?

### Additional context

_Add any other context or screenshots about the feature request here._

pattisdr:
great, no problem @allisonking, this was a small change, I will revert to make it private in the TCF backend which will be safer anyway going forward

PR #4131, opened 2023-09-21
TCF bundling based on geography

allisonking:
Closes https://github.com/ethyca/fides/issues/3957

### Description Of Changes

We can now allow both a notice experience and a TCF experience!

### Code Changes

* [x] Sort out some cookie logic which became much more apparently not quite right when switching between notices and TCF
* [x] Remove the `TCF_ENABLED` environment variable from the PC—now we determine if it's TCF purely from the backend 🎉 
* [x] Cypress tests

### Steps to Confirm

* Run `nox -s "fides_env(slim)" -- prod_prerelease core_version=2.20.2a2`
* Bring down the privacy center container that the test env creates
* Bring up the privacy center on this branch using `turbo dev` in `clients/privacy-center`. I used these env variables (note: no more `TCF_ENABLED` 🎉  )
```sh
FIDES_PRIVACY_CENTER__GEOLOCATION_API_URL=https://cdn-api.ethyca.com/location
FIDES_PRIVACY_CENTER__PRIVACY_CENTER_URL=http://localhost:3000
```
* Turn TCF on in the backend via `localhost:8080/docs`
* Now that TCF is enabled, visit localhost:8080 and edit one of the existing systems.
* Add a data use such as analytics.reporting.campaign_insights (which is one of the TCF data uses). Give it a legal basis of "Consent" and save.
* Visit the consent page and enable all the privacy notices that are `Available`.
* Visit cookie house at `localhost:3000`. Choose California from the geo picker. You should see "Manage preferences" show up at the bottom. Click it and you should see "Data sales and sharing". The banner won't come up since that's not required for CA
* Now choose Frankfurt from the geo picker. The banner should come up, and you can click "manage preferences" to open what should be the TCF modal! 
* Interact with both modals, switching between geographies, to make sure your choices persist and the cookie looks right

### Pre-Merge Checklist

* [x] Update `CHANGELOG.md`


clients/fides-js/src/components/Overlay.tsx
```diff
@@ -38,16 +34,6 @@ interface Props {
   renderModalContent: (props: RenderModalContent) => VNode;
 }
 
-/**
```
allisonking, line 41: I realized this wasn't needed anymore once I fixed how the updating cookie logic in place worked

clients/fides-js/src/fides-tcf.ts
```diff
@@ -136,7 +141,8 @@ const updateCookie = async (
   };
 
   const tcString = await generateTcString({ tcStringPreferences, experience });
-  return { ...oldCookie, tc_string: tcString };
+  const tcfConsent = transformTcfPreferencesToCookieKeys(tcStringPreferences);
+  return { ...oldCookie, tc_string: tcString, tcf_consent: tcfConsent };
```
allisonking, line 145: this probably should have been in https://github.com/ethyca/fides/pull/4124 but I didn't realize it was missing until I started switching between TCF and non-TCF

clients/fides-js/src/lib/initialize.ts
```diff
@@ -252,18 +262,12 @@ export const initialize = async ({
       isPrivacyExperience(effectiveExperience) &&
       experienceIsValid(effectiveExperience, options)
     ) {
-      // Overwrite cookie consent with experience-based consent values
-      // eslint-disable-next-line no-param-reassign
-      cookie.consent = buildCookieConsentForExperiences(
```
allisonking, line 257: this was messing up when going from Notices --> TCF because TCF doesn't have any notices, so this was making the `consent` object `{}` which would erase if someone had made choices in CA. So this logic got moved out to only notices

clients/fides-js/src/lib/preferences.ts
```diff
@@ -46,36 +45,26 @@ export const updateConsentPreferences = async ({
...
+  updateCookie: (oldCookie: FidesCookie) => Promise<FidesCookie>;
 }) => {
-  // Derive the CookieKeyConsent object from privacy notices
-  const noticeMap = new Map<string, boolean>(
-    consentPreferencesToSave.map(({ notice, consentPreference }) => [
-      notice.notice_key,
-      transformUserPreferenceToBoolean(consentPreference),
-    ])
-  );
-  const consentCookieKey: CookieKeyConsent = Object.fromEntries(noticeMap);
```
allisonking, line 58: similar to https://github.com/ethyca/fides/pull/4131/files#r1333616066, this moves the cookie assignment logic out of the shared function, and into a notice-specific function. this is because otherwise, when we save TCF (which has no notices), we would overwrite `cookie.consent` to equal `{}` which isn't right!

clients/fides-js/src/lib/preferences.ts
```diff
@@ -46,36 +45,26 @@ export const updateConsentPreferences = async ({
...
-        privacy_notice_history_id: notice.privacy_notice_history_id,
-        preference: consentPreference,
-        served_notice_history_id: servedNotice?.served_notice_history_id,
-      };
-    });
-
-  // Update the cookie object
-  // eslint-disable-next-line no-param-reassign
-  cookie.consent = consentCookieKey;
+  const fidesUserPreferences: Array<ConsentOptionCreate> | undefined =
```
allisonking, line 51: this is the same, but wrapped in a ternary to be a little more careful about what preferences we pass to the backend. we don't want to accidentally overwrite notice preferences when working with tcf data!

clients/fides-js/src/lib/preferences.ts
```diff
@@ -89,29 +78,31 @@ export const updateConsentPreferences = async ({
...
-  // 2. Update the window.Fides.consent object
+  // 2. Update the cookie object based on new preferences
+  const updatedCookie = await updateCookie(cookie);
+  Object.assign(cookie, updatedCookie);
+
+  // 3. Update the window.Fides object
   debugLog(debug, "Updating window.Fides");
   window.Fides.consent = cookie.consent;
+  window.Fides.tc_string = cookie.tc_string;
+  window.Fides.tcf_consent = cookie.tcf_consent;
```
allisonking, line 89: also probably should have been in https://github.com/ethyca/fides/pull/4124

clients/fides-js/src/lib/preferences.ts
```diff
@@ -89,29 +78,31 @@ export const updateConsentPreferences = async ({
   };
   patchUserPreferenceToFidesServer(privacyPreferenceCreate, fidesApiUrl, debug);
 
-  // 2. Update the window.Fides.consent object
+  // 2. Update the cookie object based on new preferences
+  const updatedCookie = await updateCookie(cookie);
+  Object.assign(cookie, updatedCookie);
```
allisonking, line 83: I learned something weird— assigning to the function param with something like

```
param = updatedParam;
```
doesn't work! (`param` in the parent scope is unchanged). However, doing an `Object.assign` does work, and so does assigning to individual attributes such as

```
param.consent = updatedParam.consent
```

clients/fides-js/src/lib/initialize.ts
allisonking: This does have an interesting side effect which I'm not sure if we can avoid. Let's say our cookie starts as
```js
{
  consent: {
    data_sales_and_sharing: false
  }
}
```
Now we switch to TCF. when TCF initializes, there are no notices, so it only has the consent defaults, which it merges with the existing cookie 

```js
{
  consent: {
    data_sales: true, 
    tracking: true, 
    analytics: true, 
    data_sales_and_sharing: true
  }
}
```
I'm not sure if it's ok to have the consent defaults hanging around like that. That's from [this function](https://github.com/ethyca/fides/blob/aking/prod-1070/tcf-geography-bundling/clients/fides-js/src/lib/cookie.ts/#L169-L175) which I think is there so that we continue support for the legacy config. We could potentially say just ignore those defaults if we are in TCF mode?

eastandwestwind: Hmm, thinking beyond one impl on a customer's site, suppose a web user were to visit 2 different websites, site A uses Fides non-TCF and the site B uses Fides TCF. The user has submitted consent for site A, then navigates to site B. I'd think we'd want to preserve any existing consent the user has given on site A, rather than overwrite with defaults from TCF, right?

eastandwestwind (commented):
2 questions for you!

clients/privacy-center/pages/api/fides-js.ts
```diff
@@ -101,6 +98,13 @@ export default async function handler(
     }
   }
 
+  // We determine server-side whether or not to send the TCF bundle, which is based
+  // on whether or not the experience is marked as TCF. This means for TCF, we *must*
+  // be able to prefetch the experience.
```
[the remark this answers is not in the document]
allisonking: this is a bit of a chicken and egg problem, where the backend will never return that the experience is TCF if prefetch is not enabled 😕 the flow is

1. Get the user's location 
2. Prefetch the experience, passing in the user location
3. Check if it's TCF, which can only happen if both 1 and 2 work

clients/fides-js/src/lib/initialize.ts
allisonking: site A and site B would use different cookies, right?

eastandwestwind: I didn't think so. It should just be the same `fides_consent` cookie

eastandwestwind (approved):

PR #4135, opened 2023-09-21
add version_added to TaxonomyEntity

### LINKED ISSUE #4134: 422 error when updating a taxonomy entity
### Bug Description

When updating a default data use, data subject, or data category in the taxonomy an error is thrown and the update is not saved.

#### Steps to Reproduce

1. run fides or use the rc env
2. go to the taxonomy page https://fides-plus-rc.fides-staging.ethyca.com/taxonomy
3. hover over an entity 
4. click the pencil 
5. make an edit 
6. click Update entity 
7. see the error 

### Expected behavior

Updates should save without error 

### Screenshots

image

### Environment

* Version: 2.20.1

Kelsey-Ethyca:
Closes #4134 

### Description Of Changes

add version_added to TaxonomyEntity

### Steps to Confirm

- [ ] run fides 
- [ ] go to the taxonomy page
- [ ] hover over an entity
- [ ] click the pencil
- [ ] make an edit
- [ ] click Update entity
- [ ] verify there is no error and the update is saved 

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### DIFF
--- a/clients/admin-ui/src/features/taxonomy/hooks.tsx
+++ b/clients/admin-ui/src/features/taxonomy/hooks.tsx
@@ -79,6 +79,8 @@ const transformTaxonomyBaseToInitialValues = (
   parent_key: t.parent_key ?? "",
   is_default: t.is_default ?? false,
   version_added: t.version_added ?? "",
+  version_deprecated: t.version_deprecated ?? "",
+  replaced_by: t.replaced_by ?? "",
   customFieldValues,
 });
 
--- a/clients/admin-ui/src/features/taxonomy/hooks.tsx
+++ b/clients/admin-ui/src/features/taxonomy/hooks.tsx
@@ -78,6 +78,7 @@ const transformTaxonomyBaseToInitialValues = (
   description: t.description ?? "",
   parent_key: t.parent_key ?? "",
   is_default: t.is_default ?? false,
+  version_added: t.version_added ?? "",
   customFieldValues,
 });
 
--- a/clients/admin-ui/src/features/taxonomy/hooks.tsx
+++ b/clients/admin-ui/src/features/taxonomy/hooks.tsx
@@ -78,9 +78,9 @@ const transformTaxonomyBaseToInitialValues = (
   description: t.description ?? "",
   parent_key: t.parent_key ?? "",
   is_default: t.is_default ?? false,
-  version_added: t.version_added ?? null,
-  version_deprecated: t.version_deprecated ?? null,
-  replaced_by: t.replaced_by ?? null,
+  version_added: t.version_added ?? undefined,
+  version_deprecated: t.version_deprecated ?? undefined,
+  replaced_by: t.replaced_by ?? undefined,
   customFieldValues,
 });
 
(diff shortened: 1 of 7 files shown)


clients/admin-ui/src/types/api/models/DataCategory.ts
```diff
@@ -32,4 +32,8 @@ export type DataCategory = {
    * Indicates whether the resource is currently 'active'.
    */
   active?: boolean;
+  /**
+   * This is for tracking when the daulft entity was added
```
pattisdr, line 36:
```suggestion
   * This is for tracking when the default entity was added
```

clients/admin-ui/src/types/api/models/DataSubject.ts
```diff
@@ -47,4 +47,8 @@ export type DataSubject = {
    * Indicates whether the resource is currently 'active'.
    */
   active?: boolean;
+  /**
+   * This is for tracking when the daulft entity was added
```
pattisdr, line 51:
```suggestion
   * This is for tracking when the default entity was added
```

clients/admin-ui/src/types/api/models/DataUse.ts
```diff
@@ -55,4 +55,8 @@ export type DataUse = {
    * Indicates whether the resource is currently 'active'.
    */
   active?: boolean;
+  /**
+   * This is for tracking when the daulft entity was added
```
pattisdr, line 59:
```suggestion
   * This is for tracking when the default entity was added
```

clients/admin-ui/src/features/taxonomy/hooks.tsx
```diff
@@ -78,6 +78,7 @@ const transformTaxonomyBaseToInitialValues = (
   description: t.description ?? "",
   parent_key: t.parent_key ?? "",
   is_default: t.is_default ?? false,
+  version_added: t.version_added ?? "",
```
pattisdr, line 81: Looking more closely at the db here, I think we should need to write back the values of `version_deprecated` and `replaced_by`, or we could end up overwriting those if they are populated.

pattisdr (commented):
Tested this works well nice find @Kelsey-Ethyca! I just have a comment about adding additional fields that aren't throwing errors but we would also be writing back.

- Tested updating default uses, categories, and subjects, as well as creating/updating uses, categories and subjects.


@ThomasLaPiana FYI this validation over in fideslang seems to now be requiring this field in the request body when updating. In general, it seems like it would be useful if the CTL endpoints ignored fields that were not in the request body.  We had a similar thing last week, where we realized fields not in the System request were getting set to null https://github.com/ethyca/fides/pull/4116.  What are your thoughts on that approach?

pattisdr (approved):
great @Kelsey-Ethyca thank you

ThomasLaPiana:
> Tested this works well nice find @Kelsey-Ethyca! I just have a comment about adding additional fields that aren't throwing errors but we would also be writing back.
> 
> * Tested updating default uses, categories, and subjects, as well as creating/updating uses, categories and subjects.
> 
> @ThomasLaPiana FYI this validation over in fideslang seems to now be requiring this field in the request body when updating. In general, it seems like it would be useful if the CTL endpoints ignored fields that were not in the request body. We had a similar thing last week, where we realized fields not in the System request were getting set to null #4116. What are your thoughts on that approach?

Sounds logical to me! I think that would be a pretty easy fix

PR #4136, opened 2023-09-22
Custom fides.css

galvana:
### Description Of Changes

Updating `fides.js` to include any custom css uploaded to the fidesplus-only `/custom-assets` endpoint. The custom styles are injected into `fides.js` and take priority over any of the base `fides.css` styles. The `fides.js` file from the Privacy Center now has the following properties

- The Privacy Center attempts to fetch a custom stylesheet from fidesplus via the `/custom_asset` endpoint
- If a custom stylesheet is available, it will be cached for 1 hour
- If the link is unreachable or returns an empty response, the Privacy Center will just serve `fides.js` with the default styles in `fides.css` and the last cached version of the custom stylesheet
- `/fides.js` is a virtual link, so it acts like an API endpoint, this allows us to specify a `refresh` query param to trigger a manual refresh of the custom css styles from fidesplus
- When fidesplus receives a new custom stylesheet, it will call `/fides.js?refresh` to trigger the Privacy Center to invalidate its cached version of the custom styles and retrieve the new version.

#### Fides
* [ ] New `plus_custom_asset` table to store the custom stylesheets
* [ ] Added `updateCustomAsset` mutator function to upload custom stylesheets to fidesplus
* [ ] Added an "Upload Stylesheet" button to the privacy experiences table restricted to the `CUSTOM_ASSET_UPDATE` scope

### Privacy Center
* [ ] Updated `fides.js` with the custom stylesheet retrieval logic and a `refresh` query param

### Steps to Confirm

* [ ] Needs fidesplus to run an end to end test but for this standalone change, verify that `http://privacy-center-url/fides.js` contains the custom css uploaded to fidesplus.

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`


clients/fides-js/src/components/Overlay.tsx
```diff
@@ -142,6 +141,14 @@ const Overlay: FunctionComponent<Props> = ({
     }
   }, [showBanner, cookie, options.debug, bannerIsOpen]);
 
+  useEffect(() => {
+    const cssUrl = new URL("fides.css", options.privacyCenterUrl);
+    const link = document.createElement("link");
+    link.rel = "stylesheet";
+    link.href = cssUrl.href;
+    document.head.appendChild(link);
+  }, [options.privacyCenterUrl]);
```
allisonking, line 150: actually, now that I'm thinking about this more, I think we actually want this css _in_ the `fides.js` bundle (that comes from the `fides-js` API endpoint). see [this comment](https://ethyca.atlassian.net/wiki/spaces/EN/pages/2774499406/2023-08-31+Customizing+fides.js+and+privacy+center+CSS?focusedCommentId=2775449606)

in other words, I think the flow would be something like...

1. user uploads some CSS which gets saved to the Fides server
2. the fides-js endpoint queries to see if there's any custom CSS
3. if there is, it stuff it into the fides.js payload. we used to do this with the fides.css file but not anymore. you can still find that code [here](https://github.com/ethyca/fides/commit/bbbf31982b7ec0119910c0500bc825007707c4d5) though (search for `fidesCss`)
4. then **in theory**, no changes are needed in fides.js. but I wouldn't be surprised if I'm wrong 😄 

that does mean there doesn't seem to be as much of a need for your new `fides-css` endpoint, except for perhaps that `refresh` logic. that said, the `fides-css` endpoint might be useful for the privacy center, which doesn't get bundled javascript

clients/admin-ui/src/types/api/models/UploadCustomAsset.ts
```diff
@@ -0,0 +1,6 @@
+import { CustomAssetType } from "./CustomAssetType";
+
+export type UploadCustomAsset = {
+  assetType: CustomAssetType;
+  file: File;
```
allisonking, line 5: I definitely appreciate the typing! however, we need this to be autogenerated, otherwise the next time we run `npm run openapi:generate` these types will be overwritten 😢 

`npm run openapi:generate` can only do as well as the python typing in the fastapi docs. I took a peek at https://github.com/ethyca/fidesplus/compare/custom-asset-endpoints and I think that means in that PR, the `key` param has to be a bit more strict—it'll need to be an Enum in order to get something like the `CustomAssetType` you already have. 

if you want types that are only for the FE, then they should live outside of `types/api/models`. Sometimes we have random `type` files in different folders that handle that kind of case.

more info here: https://github.com/ethyca/fides/blob/main/clients/admin-ui/src/types/api/README.md

allisonking (changes requested):
this is a great start! I'd maybe get a second opinion from Neville or Catherine on the refresh flow / which endpoint this should live in, but hopefully my comments make sense, happy to expand on them if you'd like!

clients/admin-ui/src/features/custom-assets/CustomAssetUploadModal.tsx
```diff
@@ -0,0 +1,159 @@
...
+
+  const [updateCustomAsset, { isLoading }] = useUpdateCustomAssetMutation();
+
+  const handleSubmit = async () => {
+    if (uploadedFile) {
+      try {
+        await updateCustomAsset({ assetType, file: uploadedFile }).unwrap();
+        toast(successToastParams("Stylesheet uploaded successfully"));
+        onClose();
+      } catch (error) {
```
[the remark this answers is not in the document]
galvana: Updated to use this pattern instead

clients/fides-js/src/components/Overlay.tsx
```diff
@@ -8,7 +8,6 @@ import {
 
 import { debugLog, hasActionNeededNotices } from "../lib/consent-utils";
 
-import "./fides.css";
```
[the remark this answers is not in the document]
galvana: Do you have a preference for what this fides.css override should be named?

galvana:
Thanks @allisonking for the initial review. The main thing I changed in this second pass is the requirements for `custom-fides.css`. In my original design, the contents of the custom stylesheet replaced `fides.css` completely but now the custom styles are treated as an override to the default `fides.css` values. I also implemented the more minor recommendations.

clients/admin-ui/src/types/api/models/CustomAssetType.ts
```diff
@@ -0,0 +1,10 @@
+/* istanbul ignore file */
+/* tslint:disable */
+/* eslint-disable */
+
+/**
+ * An enumeration.
+ */
+export enum CustomAssetType {
+  CUSTOM_FIDES_CSS = "custom-fides.css",
```
galvana, line 9: I wasn't sure of our policy around `npm run openapi:generate` so I ran it to make sure this would autogenerate correctly and I removed the other autogenerated types.

clients/privacy-center/pages/api/fides-js.ts
```diff
@@ -160,6 +224,11 @@ export default async function handler(
     // Include generic fides.js script
     ${fidesJS}
 
+    // Include custom-fides.css
+    const style = document.createElement('style');
+    style.innerHTML = ${JSON.stringify(customFidesCss)};
+    document.head.append(style);
```
NevilleS, line 230: We should put this in a guard like `if (customFidesCss is not an empty string)`, otherwise I think this'll do some strange things when no custom CSS exists

clients/privacy-center/pages/api/fides-js.ts
```diff
@@ -9,9 +9,64 @@ import {
...
+import { LOCATION_HEADERS, lookupGeolocation } from "~/common/geolocation";
 
-const FIDES_JS_MAX_AGE_SECONDS = 60 * 60; // one hour
+// one hour, how long the client should cache fides.js for
+const FIDES_JS_MAX_AGE_SECONDS = 60 * 60;
+// one hour, how long until the custom-fides.css is refreshed
+const CUSTOM_FIDES_CSS_TTL_MS = 3600 * 1000;
+
+let cachedCustomFidesCss: string = "";
+let lastFetched: number = 0;
```
NevilleS, line 20: Some comments here would help 👍

clients/privacy-center/pages/api/fides-js.ts
```diff
@@ -9,9 +9,64 @@ import {
...
+
+async function fetchCustomFidesCss(
```
NevilleS, line 22: Styling nit, but I think this would be better if you could move it's definition below the main entrypoint for this file which is the API `handler` function itself?

clients/privacy-center/pages/api/fides-js.ts
```diff
@@ -9,9 +9,64 @@ import {
...
+      const environment = await loadPrivacyCenterEnvironment();
+      const fidesUrl =
+        environment.settings.SERVER_SIDE_FIDES_API_URL ||
+        environment.settings.FIDES_API_URL;
+      const response = await fetch(
+        `${fidesUrl}/plus/custom-asset/custom-fides.css`
+      );
+      const data = await response.text();
+
+      console.log("Successfully retrieved custom-fides.css");
```
NevilleS, line 42: In general all the `console.log` entries are guarded by `process.env.NODE_ENV` to avoid too much log spam (this endpoint is called _a lot_ in production if caches aren't configured in front of it, so the log spam is real)

clients/privacy-center/pages/api/fides-js.ts
```diff
@@ -148,6 +209,9 @@ export default async function handler(
   if (!fidesJS || fidesJS === "") {
     throw new Error("Unable to load latest fides.js script from server!");
   }
+
+  const customFidesCss = await fetchCustomFidesCss(req);
```
NevilleS, line 213: If this fails at runtime when we launch this because of a version mismatch or something similar, what do we do? That'd be a lot of production issues!

I'd recommend we let this fail open- if _anything_ goes wrong fetching the custom CSS, log an error and continue rendering the generic script result as-is today 👍

NevilleS (changes requested):
This feels pretty reasonable, but throwing a couple flags here for safety

allisonking (approved):
this worked 💯 for me, awesome job!!

image

this was the CSS file I uploaded:

```css
:root {
  --fides-overlay-primary-color: tomato;
}
```

I was also able to swap out the CSS and the refresh flow worked 👍  (after PUTting a privacy center URL to the `fides-cloud` endpoint

clients/privacy-center/pages/api/fides-js.ts
galvana: I agree, and that's the current behavior. All the error handling happens inside `fetchCustomFidesCss` we'll either get the custom CSS string or null, both of which are handled appropriately when we inject that a few lines down.

clients/privacy-center/pages/api/fides-js.ts
galvana: The custom styles are cached so this `console.log` should only happen if the custom styles were retrieved successfully. This happens every hour, or when a manual refresh is triggered.

clients/privacy-center/pages/api/fides-js.ts
galvana: I tested this use case and in the event of an empty custom stylesheet `sytle.innerHTML` will just be assigned `""`. No side-effects reported.

clients/privacy-center/pages/api/fides-js.ts
galvana: Added comments

clients/privacy-center/pages/api/fides-js.ts
galvana: Moved function

clients/privacy-center/pages/api/fides-js.ts
galvana: Updated the logic to conditionally include the custom styles

clients/privacy-center/pages/api/fides-js.ts
NevilleS: Hmm, I *almost* trust you, but I'm going to look a second time now... 🤔

NevilleS: OK I see the `try`/`catch`, you've convinced me 👍

.fides/db_dataset.yml
```diff
@@ -2691,3 +2691,25 @@ dataset:
...
+    data_qualifier: aggregated.anonymized.unlinked_pseudonymized.pseudonymized.identified
+    fields:
+    - name: id
+      data_categories: [system.operations]
+      data_qualifier: aggregated.anonymized.unlinked_pseudonymized.pseudonymized.identified
+    - name: key
+      data_categories: [system.operations]
+      data_qualifier: aggregated.anonymized.unlinked_pseudonymized.pseudonymized.identified
+    - name: filename
+      data_categories: [user.unique_id]
```
NevilleS, line 2706:
```suggestion
      data_categories: [system.operations]
```

clients/admin-ui/src/features/custom-assets/CustomAssetUploadModal.tsx
```diff
@@ -0,0 +1,159 @@
...
+        <ModalHeader tabIndex={-1} ref={initialRef}>
+          Upload stylesheet
+        </ModalHeader>
+        <ModalBody>
+          <Text fontSize="sm" mb={4}>
+            To customize the appearance of your consent experiences, you may
+            upload a CSS stylesheet. To download a template as a helpful
+            starting point, click{" "}
+            <DocsLink
+              href="https://raw.githubusercontent.com/ethyca/fides/main/clients/fides-js/src/components/fides.css"
```
NevilleS, line 99: Side note, but this template isn't _super_ helpful... so maybe we should just have another `fid.es/customize-styles-template` link here that we can swap out?

NevilleS (approved):
Thanks! One or two open nits for you to incorporate but I leave it in your capable hands

Issue #4137, opened 2023-09-22
Features and special purposes should not be toggle-able in the TCF UI


allisonking:
### Bug Description

Per the [IAB spec](https://iabeurope.eu/iab-europe-transparency-consent-framework-policies/),users should not be given a choice regarding features and special purposes.

#### Steps to Reproduce

1. Open up our TCF overlay
1. You'll see you can toggle features and special purposes. These don't get saved to the TC string, but they do get saved to the backend

### Expected behavior

No toggles should appear for either of these

### Screenshots

![image](https://github.com/ethyca/fides/assets/24641006/c3aff6b8-ebae-427c-8c23-185dbf3355e9)

![image](https://github.com/ethyca/fides/assets/24641006/e949a211-75fd-42eb-8f23-f6ceb4080723)

### Environment

* Version:
* OS:
* Python Version:
* Docker Version:

### Additional context

_Add any other context about the problem here._

PR #4139, opened 2023-09-22
Remove features and special purposes toggles

allisonking:
Closes https://github.com/ethyca/fides/issues/4137

### Description Of Changes

Removes toggles from Features and Special Purposes since users aren't supposed to be able to change those preferences. 

image
image

Can still expand them
image

### Code Changes

* [x] Add a flag for conditionally showing toggles
* [x] Remove payload for saving to Features or Special Purposes
* [x] Update cypress tests

### Steps to Confirm

* `nox -s dev`
* Add a system with a purpose data use and a special purpose data use. I used `analytics.reporting.campain_insights` for purpose, and `essential.fraud_detection` for special purpose. Make sure they have legal bases (campaign insights should probably be `Consent` and fraud detection `legitimate interest`)
* Add a feature and a special feature. I used `Link different devices` for feature, and `Use precise geolocation data` for special feature. Note that this is a select input, and you have to hit enter after adding the feature 
image
* Start up the privacy center via `FIDES_PRIVACY_CENTER__TCF_ENABLED=true turbo dev`
* Visit `/fides-js-demo.html`. You should see the features and special purposes rendered but without a toggle

### Pre-Merge Checklist

* [x] Update `CHANGELOG.md`

### DIFF
--- a/clients/fides-js/src/components/tcf/TcfPurposes.tsx
+++ b/clients/fides-js/src/components/tcf/TcfPurposes.tsx
@@ -12,15 +12,22 @@ const PurposeToggle = ({
   purpose,
   onToggle,
   checked,
+  includeToggle = true,
 }: {
   purpose: TCFPurposeRecord;
   onToggle: () => void;
   checked: boolean;
+  includeToggle?: boolean;
 }) => {
   const dataUse = { key: purpose.name, name: purpose.name };
   const vendors = [...(purpose.vendors || []), ...(purpose.systems || [])];
   return (
-    <DataUseToggle dataUse={dataUse} checked={checked} onToggle={onToggle}>
+    <DataUseToggle
+      dataUse={dataUse}
+      checked={checked}
+      onToggle={onToggle}
+      includeToggle={includeToggle}
+    >
       <div>
         <p className="fides-tcf-toggle-content">{purpose.description}</p>
         <p className="fides-tcf-illustration fides-background-dark">
@@ -48,11 +55,13 @@ const PurposeBlock = ({
   allPurposes,
   enabledIds,
   onChange,
+  hideToggles,
 }: {
   label: string;
   allPurposes: TCFPurposeRecord[] | undefined;
   enabledIds: string[];
   onChange: (newIds: string[]) => void;
+  hideToggles?: boolean;
 }) => {
   if (!allPurposes || allPurposes.length === 0) {
     return null;
@@ -84,6 +93,7 @@ const PurposeBlock = ({
         onToggle={handleToggleAll}
         checked={allChecked}
         isHeader
+        includeToggle={!hideToggles}
       />
       {allPurposes.map((p) => (
         <PurposeToggle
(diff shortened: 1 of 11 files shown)


clients/fides-js/src/lib/consent-utils.ts
```diff
@@ -226,16 +226,12 @@ const hasActionNeededTcfPreference = (
...
   hasCurrentPreference(experience.tcf_purposes) ||
-  hasCurrentPreference(experience.tcf_special_purposes) ||
-  hasCurrentPreference(experience.tcf_features) ||
   hasCurrentPreference(experience.tcf_special_features) ||
   hasCurrentPreference(experience.tcf_vendors) ||
   hasCurrentPreference(experience.tcf_systems);
 
 export const hasActionNeededTcfPreferences = (experience: PrivacyExperience) =>
   hasActionNeededTcfPreference(experience.tcf_purposes) ||
-  hasActionNeededTcfPreference(experience.tcf_special_purposes) ||
```
eastandwestwind, line 237: What about the state where an experience _only_ has `tcf_special_purposes` and `tcf_features`? We still want to show the banner in this case, and I believe removing these will prevent that from happening.

clients/fides-js/src/components/tcf/TcfPurposes.tsx
```diff
@@ -138,6 +149,7 @@ const TcfPurposes = ({
         onChange={(newEnabledIds) =>
           onChange({ newEnabledIds, modelType: "specialPurposes" })
         }
+        hideToggles
```
eastandwestwind, line 152: I'm confused where this var comes from? It's not passed into the `TcfPurposes` const...

allisonking: oh this is a JSX shortcut, it's equivalent to 

```jsx
<Component hideToggles={true} />
```
but you can leave off the `={true}`. so `hideToggles` is the name of the prop on the component it's being passed into

clients/fides-js/src/lib/consent-utils.ts
allisonking: oh hmm that's a good point! let me think about it...

clients/fides-js/src/components/tcf/TcfPurposes.tsx
eastandwestwind: ohhh ok thanks for explaining!

clients/fides-js/src/lib/consent-utils.ts
allisonking: okay I think you are right, but this would require https://github.com/ethyca/fides/issues/4141. we basically want to treat them as notice-only, and instead of rendering "Save", "Opt out of all" "Opt into all" we just want the OK button, same as we have for notices. (it would be weird to surface something that only has special purposes and features and then have those buttons, since those aren't toggle-able!).

eastandwestwind: hah yeah bit of a catch 22 but I'll go ahead and approve this, and we can follow-up with #4141, adding back these lines in that PR.

eastandwestwind (approved):
🌟

Issue #3941, opened 2023-08-21
consume "pristine" GVL JSON from new API


adamsachs:
### Is your feature request related to a specific problem?

in https://github.com/ethyca/fidesplus/issues/1044 we are adding to `fidesplus` an API/endpoint that will serve the up-to-date, "pristine" GVL information. we should switch the FE code to get its GVL "source" from this endpoint, rather than using the hardcoded, static asset that we'd started using as an initial cut.

### Describe the solution you'd like

- API details still TBD pending https://github.com/ethyca/fidesplus/issues/1044 work
- API should be assumed to have minimal latency
- should be a relatively "clean" swap out of our existing code, i.e. the response from the API should look identical to the static GVL JSON we're using currently -- at least on our first iteration this.

### Describe alternatives you've considered, if any

continue using static JSON asset, but then we'd need to update that ~every week

### Additional context

blocked by https://github.com/ethyca/fidesplus/issues/1044

allisonking:
Closed by https://github.com/ethyca/fides/pull/4143

PR #3486, opened 2023-06-07
Include default consent templates in Python Package

ThomasLaPiana:
Closes #3484 

### Code Changes

* [x] move privacy notice defaults into `src/fides/data`
* [x] fix the hardcoded, absolute paths for the consent files

### Steps to Confirm

* [x] pinned `fidesplus` locally to this commit of `fides` via the `requirements.txt`, ran `nox -s "build(slim)" -- dev`, saw that by default we had 5 privacy notices stored in the db, also saw a privacy notice appear in the UI once i added a `System` with a `analytics` data use

### Pre-Merge Checklist

* [ ] All CI Pipelines Succeeded
* Documentation:
  * [ ] documentation complete, [PR opened in fidesdocs](https://github.com/ethyca/fidesdocs/pulls)
  * [ ] documentation [issue created in fidesdocs](https://github.com/ethyca/fidesdocs/issues/new/choose)
* [ ] Issue Requirements are Met
* [ ] Relevant Follow-Up Issues Created
* [ ] Update `CHANGELOG.md`

### Description Of Changes

add default consent files to the Python package, which previously only worked when used in the Docker image

PR #4143, opened 2023-09-22
Returning GVL.JSON through Experiences API

pattisdr:
- Have copied file into src/fides/data - assume FE is going to remove the file of the same name on that end.

Closes #<issue>

### Description Of Changes

### Code Changes

### Steps to Confirm

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### DIFF
--- a/src/fides/api/api/v1/endpoints/privacy_experience_endpoints.py
+++ b/src/fides/api/api/v1/endpoints/privacy_experience_endpoints.py
@@ -34,7 +34,11 @@ from fides.api.util.consent_util import (
     get_fides_user_device_id_provided_identity,
 )
 from fides.api.util.endpoint_utils import fides_limiter, transform_fields
-from fides.api.util.tcf_util import TCF_COMPONENT_MAPPING, TCFExperienceContents
+from fides.api.util.tcf_util import (
+    TCF_COMPONENT_MAPPING,
+    TCFExperienceContents,
+    load_gvl,
+)
 from fides.common.api.v1 import urn_registry as urls
 from fides.config import CONFIG
 
@@ -264,6 +268,9 @@ def embed_experience_details(
     has_tcf_contents: bool = any(
         getattr(tcf_contents, component) for component in TCF_COMPONENT_MAPPING
     )
+    if has_tcf_contents:
+        privacy_experience.gvl = load_gvl()
+
     # Add fetched TCF contents to the Privacy Experience if applicable
     for component in TCF_COMPONENT_MAPPING:
         setattr(privacy_experience, component, getattr(tcf_contents, component))
--- a/src/fides/api/util/tcf_util.py
+++ b/src/fides/api/util/tcf_util.py
@@ -1,4 +1,6 @@
+import json
 from enum import Enum
+from os.path import dirname, join
 from typing import Callable, Dict, List, Optional, Set, Tuple, Type, Union
 
 from fideslang.gvl import (
@@ -14,6 +16,7 @@ from fideslang.gvl import (
 from fideslang.gvl.models import Feature, Purpose
 from fideslang.models import LegalBasisForProcessingEnum
 from fideslang.validation import FidesKey
+from loguru import logger
 from sqlalchemy.engine.row import Row  # type:ignore[import]
 from sqlalchemy.orm import Query, Session
 
@@ -27,6 +30,16 @@ from fides.api.schemas.tcf import (
     TCFPurposeRecord,
     TCFVendorRecord,
 )
+from fides.config.helpers import load_file
+
+_gvl: Optional[Dict] = None
+
+GVL_PATH = join(
+    dirname(__file__),
+    "../../data",
+    "gvl.json",
+)
+
 
 TCFPurposeOrFeature = Union[TCFPurposeRecord, TCFFeatureRecord]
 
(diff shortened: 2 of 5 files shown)


src/fides/api/api/v1/endpoints/privacy_experience_endpoints.py
```diff
@@ -264,6 +268,9 @@ def embed_experience_details(
     has_tcf_contents: bool = any(
         getattr(tcf_contents, component) for component in TCF_COMPONENT_MAPPING
     )
+    if has_tcf_contents:
+        privacy_experience.gvl = load_gvl()
```
pattisdr, line 272: 😞 this is way too slow. I'm not reloading from the file each time.  Is it something about serializing all of this in FastAPI?

pattisdr: oh it might be worse in the swagger UI!

adamsachs: the swagger UI does slow this down a ton, it's really bad with rendering the large payload. What does it look like with a cURL?

fwiw this is a general concern of mine with including this as an embedding object in the payload, we're bloating this endpoint with data that doesn't necessarily belong here IMO.

that being said, I think we should be able to get this to run pretty quickly, it's not _that_ much data, and outside of the swagger UI, the plus endpoint has never been noticeably slow to me. We can try comparing directly?

adamsachs: when cURLing the `/privacy-experiences` endpoint with a `gvl` property populated in the response, i'm getting a response times in the range of ~100-300ms. that's certainly slower than without the `gvl` data in the response (~50ms) but that feels like it could be unavoidable overhead in serializing and sending that much response data back - the response size goes up by over 100x when including the `gvl` data (from ~30k to ~500kb in my case).

if those response times aren't acceptable for this endpoint, then i'm not sure we should be putting this data here (not suggesting that the response times are acceptable or not, i'm legitimately unsure!). if we _do_ want to embed the data here, i think we should certainly add a query param to enable/disable whether the `gvl` data is included, so that clients have the option of not including it if/when it's not needed.

pattisdr: thank you Adam. Discussed internally, I'll be adding an `include_gvl` query param that is only applicable if we also have TCF contents

src/fides/api/util/tcf_util.py
```diff
@@ -27,6 +30,16 @@
...
     TCFVendorRecord,
 )
+from fides.config.helpers import load_file
+
+_gvl: Optional[Dict] = None
+
+GVL_PATH = join(
+    dirname(__file__),
+    "../../data",
+    "gvl.json",
```
allisonking, line 40: I remember there being some trickery when importing static files like this once we get to fidesplus—any chance that will be a problem here?

allisonking (commented):
this looks good to me! I'll work on https://github.com/ethyca/fides/issues/3941 today to get the FE to start consuming this and make sure all is good, then we can remove its version of gvl.json. I can work off of this branch, and then we can merge that one with this PR? so that we don't pollute `main` with two GVL.json files 😄 but if that's not a concern then merging this one first would be fine too!

pattisdr:
>  I can work off of this branch, and then we can merge that one with this PR? 

That sounds perfect @allisonking!

allisonking:
Oh actually, while you're here, let's use the latest GVL: https://vendor-list.consensu.org/v3/vendor-list.json (updated 9/21)

src/fides/api/util/tcf_util.py
pattisdr: that's a good callout, we'll need to make sure this works!  I tried to avoid hardcoded absolute paths similar to https://github.com/ethyca/fides/pull/3486

adamsachs: oh this is a good point, we need to see whether this is packaged up with the python project and if it'd resolve from this relative path. without looking closer, it feels like it may not be, considering that we have to explicitly add the static saas config files to `fidesplus`?

we should be able to give this a test by pushing up an alpha tag to this branch

allisonking: confirmed this works with an alpha tag 🎉

allisonking:
Frontend piece: https://github.com/ethyca/fides/pull/4159

adamsachs (approved):
preemptively approving, once the FE piece gets merged in here, assuming CI passes, let's get this merged into `main`.

just a general note/disclaimer that the hardcoded `gvl.json` is a workaround until this API is moved over to `fidesplus` and we leverage the dynamically fetched vendor list that's maintained there!

CHANGELOG.md
```diff
@@ -27,11 +27,13 @@ The types of changes are:
 - Fides-js can now display preliminary TCF data [#3879](https://github.com/ethyca/fides/pull/3879)
 - Fides-js can persist TCF preferences to the backend [#3887](https://github.com/ethyca/fides/pull/3887)
 - TCF modal now supports setting legitimate interest fields [#4037](https://github.com/ethyca/fides/pull/4037)
+- Embed the GVL in the GET Experiences response [#4143](https://github.com/ethyca/fides/pull/4143)
 
 ### Changed
 - Added further config options to customize the privacy center [#4090](https://github.com/ethyca/fides/pull/4090)
 - CORS configuration page [#4073](https://github.com/ethyca/fides/pull/4073)
 - Refactored `fides.js` components so that they can take data structures that are not necessarily privacy notices [#3870](https://github.com/ethyca/fides/pull/3870)
+- Use hosted GVL.json from the backend [#4159](https://github.com/ethyca/fides/pull/4159)
```
pattisdr, line 36: @allisonking do we need to remove the FE GVL file?

allisonking: yes, it's gone! it looks like it's showing up on GH as a rename

pattisdr: ah you're right thank you!

pattisdr: could you also reconcile this with the PR just merged?

pattisdr:
Thank you for your updates Allison!

PR #4144, opened 2023-09-22
Add number of vendors with link to vendor tab

allisonking:
Closes https://github.com/ethyca/fidesplus/issues/1110

### Description Of Changes

https://github.com/ethyca/fides/assets/24641006/6a08827d-9406-42a7-bd46-74a8b1c75334

### Code Changes

* [x] Refactor `TcfModalContent` which doesn't really need to be its own component anymore. This makes it easier to hoist tab state up one component, which allows us to open the modal with a tab preselected
* [x] Adds a button which shows the number of vendors, and if you click on it, it will open the modal straight to the vendor tab
* [x] Removes the "Manage prefereneces" links on stacks which make less sense now that it's in the banner
* [x] Cypress test

### Steps to Confirm

* Set up a TCF environment
* When you visit `/fides-js-demo.html`, you should see the banner pop up with a new button "View our # partners", where # is the number of systems + vendors you have in your data map.
* Clicking on the button should open the modal, dismiss the banner, and bring you straight to the vendor tab.

### Pre-Merge Checklist

* [x] Update `CHANGELOG.md`


clients/fides-js/src/components/tcf/TcfOverlay.tsx
```diff
@@ -228,19 +249,17 @@ const TcfOverlay: FunctionComponent<OverlayProps> = ({
...
+            <InitialLayer experience={experience} />
+            <button
+              type="button"
+              className="fides-link-button"
+              onClick={() => {
+                onManagePreferencesClick();
+                goToVendorTab();
+              }}
+            >
+              View our {numVendors} partner{numVendors > 1 ? "s" : ""}
```
eastandwestwind, line 261: will this button show if there are 0 systems + vendors?

clients/fides-js/src/components/tcf/TcfOverlay.tsx
```diff
@@ -252,17 +271,34 @@ const TcfOverlay: FunctionComponent<OverlayProps> = ({
...
+              activeTabIndex={activeTabIndex}
+              onTabChange={setActiveTabIndex}
+            />
+            <TcfConsentButtons
+              experience={experience}
+              onSave={onSave}
+              firstButton={
+                <Button
+                  buttonType={ButtonType.SECONDARY}
+                  label={experience.experience_config?.save_button_label}
```
eastandwestwind, line 294: what's the default if there's no `save_ button_label` configured?

eastandwestwind (commented):
Looks great, just 2 nit questions!

clients/fides-js/src/components/tcf/TcfOverlay.tsx
allisonking: it will! but in theory it shouldn't be possible to have 0 vendors since then no purposes/features would show up. but also I asked Rachel and she said she's ok with it saying 0 :) I suppose in that case it should say 0 vendor`s`...

clients/fides-js/src/components/tcf/TcfOverlay.tsx
allisonking: oh hm I guess there isn't one. I'll check if we have a fallback elsewhere!

allisonking: it doesn't look like we have a fallback for "save" anywhere. do you think we should put one? I don't mind not having one since it'd be a reminder for the user they need to fill out their experience config if it shows up blank, but maybe that's not a good customer experience :) I can see it being a tricky thing to debug if it always says i.e. "Save" and a user expects it to say something else

eastandwestwind: Looking a bit, I see that the BE _should_ always populate the labels with defaults in `fides/data/privacy_notices/privacy_experience_config_defaults.yml`, so I think it's improbable we run into a case where it's undefined. We can keep as blank 👍

eastandwestwind (approved):

PR #4159, opened 2023-09-25
Use backend GVL json

allisonking:
Closes https://github.com/ethyca/fides/issues/3941

### Description Of Changes

Removes the FE gvl.json in favor of getting it from the backend.

### Code Changes

* [x] Update types so that privacy experience has an optional `gvl` type
* [x] Rework references to the static gvl.json file
* [x] Lower TCF warning/error kb in rollup config
* [x] Add the `include_gvl` param to the `fetchExperience` calls. This should be okay to always include in `fides-js` since the backend won't include GVL if there's no TCF to begin with
* [x] Update cypress tests to include a minimal gvl set to get tests working. Minimal to keep the fixture lean, but we could consider including the whole thing...

### Steps to Confirm

* Everything should work as well as before! In particular, make sure stacks render, and that the GVL tags render

### Pre-Merge Checklist

* [x] Update `CHANGELOG.md`


clients/fides-js/rollup.config.mjs
```diff
@@ -14,8 +14,8 @@ const GZIP_SIZE_ERROR_KB = 20; // fail build if bundle size exceeds this
 const GZIP_SIZE_WARN_KB = 15; // log a warning if bundle size exceeds this
 
 // TCF
-const GZIP_SIZE_TCF_ERROR_KB = 100;
-const GZIP_SIZE_TCF_WARN_KB = 90;
+const GZIP_SIZE_TCF_ERROR_KB = 40;
```
eastandwestwind, line 17: 🎉

clients/fides-js/src/lib/tcf.ts
```diff
@@ -62,7 +56,7 @@ export const generateTcString = async ({
   // due to TCF library not yet supporting latest GVL (https://vendor-list.consensu.org/v3/vendor-list.json).
   // We'll need to update this with our own hosted GVL once the lib is updated
   // https://github.com/InteractiveAdvertisingBureau/iabtcf-es/pull/389
```
eastandwestwind, line 58: we can remove this comment right?

clients/fides-js/src/lib/consent-types.ts
```diff
@@ -85,6 +86,7 @@ export type PrivacyExperience = {
   tcf_systems?: Array<TCFVendorRecord>;
   tcf_features?: Array<TCFFeatureRecord>;
   tcf_special_features?: Array<TCFFeatureRecord>;
+  gvl?: GVLJson;
```
eastandwestwind, line 89: When is gvl undefined? When tcf mode is disabled?

clients/fides-js/src/lib/tcf.ts
```diff
@@ -62,7 +56,7 @@ export const generateTcString = async ({
...
-  const tcModel = new TCModel(new GVL(gvlJson as VersionOrVendorList));
+  const tcModel = new TCModel(new GVL(experience.gvl));
```
eastandwestwind, line 59: Can gvl ever be undefined in TCF mode? should we catch / log in the case of potentially hitting this state?

clients/fides-js/src/lib/consent-types.ts
allisonking: yes, that's right!

clients/fides-js/src/lib/tcf.ts
allisonking: yeah that's a good idea—I'll add something!

clients/fides-js/src/lib/tcf.ts
allisonking: ah right!

clients/fides-js/src/lib/tcf.ts
allisonking: added a try catch: https://github.com/ethyca/fides/pull/4159/commits/18a9c2c26a6965bfd99bbc8aed7329dd4075794e

eastandwestwind (approved):

Issue #4141, opened 2023-09-22
Make Features and Special Purposes "notice only" in TCF


allisonking:
### Is your feature request related to a specific problem?

Related to https://github.com/ethyca/fides/issues/4137, these TCF entities should effectively be notice-only.

### Describe the solution you'd like

We should change the backend so that if the entity is "Feature" or "Special Purpose", then the consent mechanism should be "notice only".

We also need to change the frontend to respect this. If the mechanism is notice only, then most of the FE should fall into place, but there will probably be a few tweaks needed (and tests/fixtures updated)

### Describe alternatives you've considered, if any

_A description of any alternative solutions or features you've considered._

### Additional context

As of https://github.com/ethyca/fides/pull/4139, we do not send any preferences for Features or Special Purposes

pattisdr:
Just noting here, Allison already noticed this, https://github.com/ethyca/fides/pull/4201 updates the backend to make the default preferences "acknowledge" for feature and special purpose.  That PR updates the default preferences for many of the TCF sections so I just did these altogether -

PR #4161, opened 2023-09-26
Recording interactions with TCF notices

galvana:
### Description Of Changes

Adding `served_notice_history_id` to the TCF fields of the `PATCH /privacy-preference` request.

### Code Changes

* [ ] Created a distinction between `banner` and `tcf_banner` for `FidesUIShown` events
* [ ] `PATCH /notices-served` is called from the `tcf_banner` and the consent `overlay`
* [ ] Updated Cypress tests for the TCF banner worfklows

### Steps to Confirm

* [ ] Set `FIDES__CONSENT__TCF_ENABLED` to `true`
* [ ] Start fidesplus
* [ ] Navigate to `http://localhost:3000/fides-js-demo.html?geolocation=fi-18` to view the TCF banner
* [ ] Look for the `Dispatching event type FidesUIShown from tcf_banner` in the browser console
* [ ] Opt-in or opt-out and verify a call to `PATCH /privacy-preferences` is made, the TCF fields should have the `served_notice_history_id`
image

### Pre-Merge Checklist

* [ ] Update `CHANGELOG.md`


clients/fides-js/src/components/tcf/TcfOverlay.tsx
```diff
@@ -182,9 +226,30 @@ const TcfOverlay: FunctionComponent<OverlayProps> = ({
...
   );
 
+  const privacyNotices = useMemo(
+    () => experience.privacy_notices ?? [],
+    [experience.privacy_notices]
+  );
+
+  const isAllNoticeOnly = privacyNotices.every(
+    (n) => n.consent_mechanism === ConsentMechanism.NOTICE_ONLY
+  );
```
allisonking, line 236: we don't need this right now—all TCF payloads will return an empty `privacy_notices` array, so we can just pass `[]` to `notices`. also, TCF doesn't have a concept of notice-only _yet_ (see https://github.com/ethyca/fides/issues/4141), so for now, let's just pass in false for `acknowledgeMode`

clients/fides-js/src/components/tcf/TcfOverlay.tsx
```diff
@@ -182,9 +226,30 @@ const TcfOverlay: FunctionComponent<OverlayProps> = ({
...
+    options,
+    userGeography: fidesRegionString,
+    acknowledgeMode: isAllNoticeOnly,
+    privacyExperience: experience,
+  });
+
   const handleUpdateAllPreferences = useCallback(
-    (enabledIds: EnabledIds) => {
-      const tcf = createTcfSavePayload({ experience, enabledIds });
+    (enabledIds: EnabledIds, servedNotices: LastServedNoticeSchema[]) => {
```
allisonking, line 247: I don't think `servedNotices` needs to be an arg here—I think you can just reference it above (since it's function inside the component, it has access to its vars)

allisonking (commented):
nice job @galvana this is looking like the right approach so far! we'll want to make sure the types are right—they make sense to me, but we should verify with a `npm run openapi:generate`. let me know if you'd like help with that!

clients/fides-js/src/components/tcf/TcfOverlay.tsx
galvana: That's what I tried originally but when I logged out the value of `servedNotices` inside `handleUpdateAllPreferences` it was empty. The only way I could get it to work is by passing it in as a param. Could it be something to do with `useConsentServed`?

galvana: Never mind, this is working without the `servedNotices` arg

clients/fides-js/src/components/tcf/TcfOverlay.tsx
```diff
@@ -87,12 +91,51 @@ export interface UpdateEnabledIds {
...
 
+type Category =
+  | "purpose_consent"
+  | "purpose_legitimate_interests"
+  | "special_feature"
+  | "vendor_consent"
+  | "vendor_legitimate_interests"
+  | "system_consent"
+  | "system_legitimate_interests";
+type NoticeSubMap = {
```
allisonking, line 102: these aren't technically Notices (we typically don't use notices for TCF). maybe we can change this to be `TcfEntitiesMap`?

clients/fides-js/src/components/tcf/TcfOverlay.tsx
```diff
@@ -87,12 +91,51 @@ export interface UpdateEnabledIds {
...
+  [key: string]: string | number;
+};
+
+type NoticeMap = Record<Category, NoticeSubMap>;
+
+const noticeMap = (servedNotices: LastServedConsentSchema[]): NoticeMap => {
```
allisonking, line 108: heh I guess these shouldn't really be called `servedNotices` either (you probably noticed the backend type changed from `LastServedNoticeSchema` --> `LastServedConsentSchema`) so I guess this should be `servedConsent`. it's definitely a nit, but I think not conflating notices and TCF is a good distinction that helps with understanding all that's going on here

clients/fides-js/src/components/tcf/TcfOverlay.tsx
```diff
@@ -87,12 +91,51 @@ export interface UpdateEnabledIds {
...
+    special_feature: {},
+    vendor_consent: {},
+    vendor_legitimate_interests: {},
+    system_consent: {},
+    system_legitimate_interests: {},
+  };
+
+  servedNotices.forEach((notice) => {
+    (Object.keys(map) as Category[]).forEach((key) => {
+      const value = notice[key as keyof LastServedConsentSchema];
```
allisonking, line 121: it _might_ help TS to do something like this https://github.com/ethyca/fides/blob/main/clients/fides-js/src/lib/tcf/types.ts/#L177-L186 for `Category` and `LastServedConsentSchema`. that said, it might not get you as far as you want, sometimes TS is picky about this kind of thing. but it _might_ save you an `as` here or there

clients/fides-js/src/components/tcf/TcfOverlay.tsx
```diff
@@ -104,17 +147,21 @@ const transformTcfModelToTcfSave = ({
     return {
       id: model.id,
       preference,
+      served_notice_history_id: noticeSubMap[String(model.id)],
```
allisonking, line 150: ah I guess this type is still called `notice` in the backend 😅 ok I take back what I said about distinguishing between notices and TCF...

clients/fides-js/src/lib/hooks.ts
```diff
@@ -64,41 +65,79 @@ export const useDisclosure = ({ id }: { id: string }) => {
   };
 };
 
+/**
+ * Extracts the id value of each object in the list and returns a list
+ * of IDs, either strings or numbers based on the IDs' type.
+ */
+const extractIds = <T extends { id: string | number }[]>(
```
allisonking, line 72: can we reuse this function here? https://github.com/ethyca/fides/blob/main/clients/fides-js/src/components/tcf/TcfConsentButtons.tsx/#L15-L20 maybe extract it somewhere else, then share it between the two files?

(but also, nice use of `T`!)

clients/privacy-center/cypress/e2e/consent-banner-tcf.cy.ts
```diff
@@ -34,10 +41,12 @@ const SPECIAL_PURPOSE_1 = {
 const SYSTEM_1 = {
   id: "ctl_b3dde2d5-e535-4d9a-bf6e-a3b6beb01761",
   name: "Fides System",
+  served_notice_history_id: "ser_5b1bc497-b4ba-489d-b39c-9ff352d460b0",
```
allisonking, line 44: woo thank you for combing through this test!

clients/fides-js/src/components/tcf/TcfOverlay.tsx
```diff
@@ -246,14 +312,14 @@ const TcfOverlay: FunctionComponent<OverlayProps> = ({
         userLocationString: fidesRegionString,
         cookie,
         debug: options.debug,
-        servedNotices: null, // TODO: served notices
+        servedNotices: null,
```
allisonking, line 315: I think the thing to do here is to actually refactor NoticeOverlay a little too... I know that's not what this ticket calls for, but I think it'd be a good thing to do if you're up for it. basically, before TCF ever came along, we would:
1. In `NoticeOverlay` get the `servedNotices`
2. When the user saves, pass the `servedNotices` to `updateConsentPreferences` along with the `consentPreferencesToSave`
3. `updateConsentPreferences` would then do the unenviable task of [mapping `servedNotices` to `consentPreferencesToSave`](https://github.com/ethyca/fides/blob/main/clients/fides-js/src/lib/preferences.ts/#L54-L60)

Now that we have TCF, we:
1. In `TcfOverlay` get the `servedNotices`
2. `TcfOverlay` does the unenviable task of mapping `servedNotices` to each of the tcf entities (what your `noticeSubMap` is doing
3. We pass the `tcf` obj directly to `updateConsentPreferences` and set `servedNotices=null`

This is a good approach, and I think we should do the same for notices. In other words, I think `updateConsentPreferences` should no longer take a `servedNotices`, and we should pass an object that already has the `servedNotices` matched into `updateConsentPreferences`, same as we do for TCF. so then we can remove some code from `preferences.ts`. this might involve updating the type that `updateConsentPreferences` takes but hopefully it's not arduous. what do you think?

allisonking (commented):
awesome job figuring this out @galvana ! I left a comment for something I think we can improve, curious to hear your thoughts!

clients/fides-js/src/lib/hooks.ts
galvana: I thought about it but `getAllIds` casts the result to a string and for my use case, I need to preserve the original type

```
const getAllIds = (modelList: TcfModels) => {
  if (!modelList) {
    return [];
  }
  return modelList.map((m) => `${m.id}`);
};
```

clients/fides-js/src/components/tcf/TcfOverlay.tsx
galvana: Let's discuss this next week, I couldn't figure out a way to remove the `as keyof` without adding more code than we currently have for the related type definitions.

clients/fides-js/src/components/tcf/TcfOverlay.tsx
galvana: Sounds good! I gave it a shot, let me know if that's what you had in mind.

allisonking: nice job, I think you handled it perfectly! 💯

clients/fides-js/src/components/tcf/TcfOverlay.tsx
allisonking: okay no worries—I figured TS might get picky about it. let's leave it as-is then!

clients/fides-js/src/lib/hooks.ts
```diff
@@ -64,41 +65,79 @@ export const useDisclosure = ({ id }: { id: string }) => {
...
-      // is the only time we show all notices
+      // is the only time we show all notices.
+      // When TCF is enabled, report notices served as soon as the banner is shown
       if (
         !event.detail.extraDetails ||
-        event.detail.extraDetails.servingComponent !== ServingComponent.OVERLAY
+        (event.detail.extraDetails.servingComponent !==
+          ServingComponent.OVERLAY &&
+          event.detail.extraDetails.servingComponent !==
+            ServingComponent.TCF_BANNER)
```
allisonking, line 108: now that I'm testing this out, I think there's something we should ask product for clarity on... we had a discussion back when we did this for notices of "when" to say consent was served. for notices, we decided only for when the modal pops up, since the banner doesn't actually render notices.

for TCF, as you know, we decided the banner _does_ render what the user is consenting to (via the "stacks") and so we put it on the banner. however, nothing happens if only the modal is opened. for example, if I've already opted in to everything, the banner doesn't render the next time I visit the site. but if I decide I want to make a change and open the modal (by clicking "Manage Consent" on the page in the screenshot), then we don't send another notices-served patch.

image

I suspect we want to send it for both banner and overlay for TCF, but that we don't want to send it twice, if the user goes through the flow of 1. banner opens 2. user clicks "manage preferences" in the banner 3. the modal pops up. that was actually my first implementation of notices back when I was POC-ing it—we can add a state to the hook and it should sort itself out: https://github.com/ethyca/fides/pull/3845/commits/5ad7ebcb6e79d74ad56f0568ce7e03b6b0c0f702

but let's confirm with product first, I can ask...

allisonking (commented):
I'm pretty much ready to approve this as I tested locally and it looks great, and I super appreciate the refactor you did for notices! I'm just going to go ask product one more question, and then hopefully we can get this merged 💥

galvana:
I was finally able to get back to this. I updated the conditions to trigger the notices served to include the TCF_OVERLAY. The only time we don't trigger the notice is for the consent banner, since that's the only component where notices aren't shown directly.

clients/fides-js/src/lib/hooks.ts
```diff
@@ -97,15 +97,12 @@ export const useConsentServed = ({
...
+      // The only time a notices served API call isn't triggered is when
+      // the BANNER is shown. Calls can be triggered for
+      // TCF_BANNER, TCF_OVERLAY, and OVERLAY
       if (
         !event.detail.extraDetails ||
-        (event.detail.extraDetails.servingComponent !==
-          ServingComponent.OVERLAY &&
-          event.detail.extraDetails.servingComponent !==
-            ServingComponent.TCF_BANNER)
+        event.detail.extraDetails.servingComponent === ServingComponent.BANNER
```
allisonking, line 105: ah nice, this is clean, and the comment helps a lot 👍

allisonking (approved):
whew, way to persist through this PR! I know it's probably been haunting you for a while now 😄 great job! ⭐

galvana:
> whew, way to persist through this PR! I know it's probably been haunting you for a while now 😄 great job! ⭐

I'm free!

PR #4163, opened 2023-09-26
trigger prerelease build in fidesplus on beta tag

adamsachs:
Closes n/a

### Description Of Changes

slight tweak/enhancement to our integrated CI workflows. Currently, we don't trigger any `fidesplus` builds when a `beta` tag is pushed to `fides` -- we only on do this on `alpha` tags to `fides`. This PR would update the logic so that we build "prerelease" versions of `fidesplus` whenever we push a `beta` tag (i.e. on `fides:main`). 

we'll get a published image with an explicit tag based off the `beta` tag in `fides` and the version of `fidesplus:main` at that point. this image will in effect look similar to a `double-edge` image but it will be explicitly tagged, so it's a bit safer and more stable to use in e.g. demo environments that we want to keep a bit more constant!

the only real downside is more images/noise on dockerhub. i don't know of that causing any significant operational or financial cost at this point, but if it does, then we should look to scale back the publishing that's done by these sort of automated build pipelines

### Code Changes

* [x] tweak logic/conditional for triggering a prerelease `fidesplus` build based on `fides` tags

### Steps to Confirm

* [ ] once we merge this to `main`, we can push up a `beta` tag to `fides:main` and ensure it publishes the type of image we expect
* [x] we can also push an `alpha` tag to this branch to confirm there aren't any regressions with that workflow: https://hub.docker.com/layers/ethyca/fidesplus/prerelease_core_2.20.2a13-slim/images/sha256-5cb3082e585cea86b4d5ef1cd65f837e14bd52d26690cab81ef6e232441ea25d?context=explore looks good 👍 

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`


allisonking (approved):
nice, yeah I think this will be very helpful!

Issue #4175, opened 2023-09-28
Split vendor toggles into separate toggles for consent and legitimate interest


allisonking:
### Is your feature request related to a specific problem?

Vendor toggles need to be two separate toggles instead of one toggle.

### Describe the solution you'd like

Render two toggles instead of just one toggle for vendors. One controls vendor legitimate interest, and one controls vendor consent. Not all vendors will have both though.

We will need to save this to the TC string/cookie and read from it since the backend won't be quite ready yet.

### Describe alternatives you've considered, if any

_A description of any alternative solutions or features you've considered._

### Additional context

https://www.figma.com/file/dmEwdK3xZwjKfGVQ9t66xe/Fides-v.2-Master-Working-Files?node-id=10555%3A605238&mode=dev

Split out of this ticket: https://github.com/ethyca/fides/issues/4166

Roger-Ethyca:
moving to done

PR #4189, opened 2023-09-29
Double vendor toggles

allisonking:
Closes https://github.com/ethyca/fides/issues/4175

### Description Of Changes

https://github.com/ethyca/fides/assets/24641006/4ed03ebb-a174-4b35-b3dd-2a22f2dba903

### Code Changes

* [x] Add double toggles to the vendor view
* [x] Separate out vendor consent and vendor legints from just `vendor_preferences`
* [x] Read these values from the TC string on initialization
* [x] Update cypress tests

### Steps to Confirm

* See video!

### Pre-Merge Checklist

* [x] Update `CHANGELOG.md`


clients/fides-js/src/components/DataUseToggle.tsx
```diff
@@ -17,6 +17,7 @@ const DataUseToggle = ({
   disabled,
   isHeader,
   includeToggle = true,
+  secondToggle,
```
allisonking, line 20: this is not a good pattern! it's weird that there'd be a toggle defined in the component, and a separate one passed in. but it works quite well, and time is short 🥲 

eventually, I think we should totally rework `TcfVendors` to be a table, and maybe not even use `DataUseToggle` which was never really meant to be a table.

clients/fides-js/src/components/tcf/TcfVendors.tsx
```diff
@@ -157,35 +162,39 @@ const StorageDisclosure = ({ vendor }: { vendor: Vendor }) => {
...
 }) => {
   const [isFiltered, setIsFiltered] = useState(false);
 
-  // Vendors and Systems are the same for the FE, but are 2 separate
-  // objects in the backend. We combine them here but keep them separate
-  // when patching preferences
-  const vendors = [...(allVendors || []), ...(allSystems || [])];
-  const enabledIds = [...enabledVendorIds, ...enabledSystemIds];
-
-  if (vendors.length === 0) {
```
allisonking, line 182: I snuck in a refactor here because it was getting unwieldy to pass in `enabledVendorConsentIds`, `enabledVendorLegintIds`, `enabledSystemLegintIds`, `enabledSystemConsentIds`. instead, the parent component combines vendors and systems (this component used to do that). the parent component is also responsible for separating back out into vendors and systems to save to the backend.

clients/fides-js/src/components/tcf/TcfVendors.tsx
```diff
@@ -223,14 +238,47 @@ const TcfVendors = ({
...
+                vendor,
+                LegalBasisForProcessingEnum.LEGITIMATE_INTERESTS
+              );
             }}
-            checked={enabledIds.indexOf(vendor.id) !== -1}
+            checked={enabledVendorLegintIds.indexOf(vendor.id) !== -1}
             badge={gvlVendor ? "IAB TCF" : undefined}
+            secondToggle={
+              <div
+                style={{ width: "50px", display: "flex", marginLeft: ".2em" }}
```
allisonking, line 267: because we aren't using a table, we force positioning this way 😬 follow up ticket to refactor to make things into proper tables: https://github.com/ethyca/fides/issues/4190

clients/fides-js/src/fides-tcf.ts
```diff
@@ -109,30 +111,31 @@ const updateCookie = async (
...
-    vendor_preferences: experience.tcf_vendors?.map((vendor) => ({
-      id: vendor.id,
-      preference: getInitialPreference(vendor),
-    })),
-    system_preferences: experience.tcf_systems?.map((system) => ({
-      id: system.id,
-      preference: getInitialPreference(system),
-    })),
-  };
+  // const tcStringPreferences: TcfSavePreferences = {
```
allisonking, line 114: looking for a second opinion on the commented out code in this file! the purpose of this logic was for the use case of we have an experience that has user preferences filled in, and now we are populating the cookie/tc string based on those values.

however, with our recent prefetch work, and the decision that TCF needs prefetch (where the cookie determines user preference rather than the backend) I don't think this makes sense anymore. In particular, it is difficult to work with in this PR because the info we get from the backend doesn't have `vendorConsents`/`vendorLegints` yet. 

I'd like to just get rid of it to simplify code paths, but that feels like a big decision!

clients/fides-js/src/lib/tcf.ts
```diff
@@ -48,7 +48,7 @@ export const generateTcString = async ({
   experience,
   tcStringPreferences,
 }: {
-  tcStringPreferences?: TcfSavePreferences;
+  tcStringPreferences?: EnabledIds;
```
allisonking, line 51: the logic in this func should be mostly the same as before—we just get a different input obj, so things are a little different (`EnabledIds` actually maps well to this use case, because, like TC strings, it only stores opt ins)

clients/fides-js/src/components/DataUseToggle.tsx
allisonking: follow up ticket: https://github.com/ethyca/fides/issues/4190

clients/fides-js/src/components/tcf/TcfOverlay.tsx
```diff
@@ -78,14 +83,24 @@ const getEnabledIds = (modelList: TcfModels) => {
...
+const getVendorEnabledIds = (
+  modelList: TCFVendorRecord[] | undefined,
+  legalBasis: LegalBasisForProcessingEnum
+) => {
+  if (!modelList) {
+    return [];
+  }
+  const records = vendorRecordsWithLegalBasis(modelList, legalBasis);
+  if (legalBasis === LegalBasisForProcessingEnum.LEGITIMATE_INTERESTS) {
+    // TODO: the backend should eventually return legint fields with a default preference of OPT_IN
```
eastandwestwind, line 95: do we have a ticket number this is coming in?

clients/fides-js/src/fides-tcf.ts
eastandwestwind: yes you're right! Since TCF requires prefetch, this code path isn't necessary. Let's go ahead and remove this code and leave a code comment describing something like:  for TCF we don't need to update the cookie on init, because we populate the pre-fetched experience with consent from a pre-existing cookie

eastandwestwind (commented):
Looking good overall @allisonking ! Left a couple Qs for you

eastandwestwind:
Oh, and could you add a `CHANGELOG` entry?

clients/fides-js/src/components/tcf/TcfOverlay.tsx
allisonking: I think that'd fall under this one: https://github.com/ethyca/fides/issues/4141 I'll add it to the comment 👍

clients/fides-js/src/components/tcf/TcfOverlay.tsx
```diff
@@ -170,15 +220,55 @@ const TcfOverlay: FunctionComponent<OverlayProps> = ({
...
+      LegalBasisForProcessingEnum.CONSENT
+    );
+    let vendorsLegint = getVendorEnabledIds(
+      vendorsAndSystems,
+      LegalBasisForProcessingEnum.LEGITIMATE_INTERESTS
+    );
+
+    // Initialize vendor values from the TC string if it's available. The cookie only
+    // stores what the backend has, while the TC string has a little more right now.
+    // (fidesplus#1128)
```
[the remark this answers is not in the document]
allisonking: Yes—you have it right! Let me reword the comment to make that clearer.

eastandwestwind (approved):
Thanks for making these changes, looks great!

PR #4228, opened 2023-10-05
fides: add `System` model support for new tcf fields

### LINKED ISSUE #4227: fides: support new TCF related fields on `System` model

There are some TCF-related fields we need to add to our `System` model. We also want to surface them in the TCF overlay.

Follow up to https://github.com/ethyca/fideslang/issues/172.

We need to update our corresponding DB model here in `fides` with these additional fields. We also need to ensure these fields are surfaced appropriately in the TCF overlay that we generate.

We can try to ensure (thru automated test coverage) that the new fields in the `fideslang`-defined pydantic model are appropriately exposed via the system CRUD API.

adamsachs:
Closes #4227 

### Description Of Changes

Adds support in fides for the new fields on the `System` and `PrivacyDeclaration` model added to fideslang.

- `System.cookie_max_age_seconds`
- `System.cookie_refresh`
- `System.uses_cookies`
- `System.uses_non_cookie_access`
- `System.legitimate_interest_disclosure_url`
- `PrivacyDeclaration.flexible_legal_basis_for_processing`

### Code Changes

* [x] update `System` db model
* [x] DB migration to update `ctl_systems` table
* [x] update system API tests to cover new fields
* [x] update TCF overlay generation to include and populate these fields in the vendor records
* [ ] ~(maybe) update admin UI forms to expose these new fields?~ split into https://github.com/ethyca/fides/issues/4248
* [ ] ~(maybe) update TCF overlay UI to leverage these new fields?~ split into https://github.com/ethyca/fides/issues/4255

### Steps to Confirm

* [x] additional fields can't be set yet in the datamap UI - but i ensured they could be set manually via API using the following payload with a `PUT /system`

```
{"fides_key":"ts1","organization_fides_key":"default_organization","tags":[],"name":"ts1","description":"","registry_id":null,"meta":null,"fidesctl_meta":null,"system_type":"","data_responsibility_title":null,"egress":null,"ingress":null,"privacy_declarations":[{"name":"pd1", "flexible_legal_basis_for_profiling": true, "data_categories":["system.operations"],"data_use":"analytics.reporting.ad_performance","data_subjects":[],"features":[],"impact_assessment_location":"","retention_period":"","processes_special_category_data":false,"data_shared_with_third_parties":false,"customFieldValues":{},"cookies":[],"id":""}],"joint_controller":null,"third_country_transfers":null,"administrating_department":"","data_protection_impact_assessment":null,"vendor_id":null,"dataset_references":[],"processes_personal_data":true,"exempt_from_privacy_regulations":false,"reason_for_exemption":null,"uses_profiling":false,"legal_basis_for_profiling":[],"does_international_transfers":false,"legal_basis_for_transfers":[],"requires_data_protection_assessments":false,"dpa_location":null,"dpa_progress":null,"privacy_policy":null,"legal_name":"","legal_address":"","responsibility":[],"dpo":"","joint_controller_info":"","data_security_practices":"","cookie_max_age_seconds":355,"uses_cookies":true,"cookie_refresh":true,"uses_non_cookie_access":true,"legitimate_interest_disclosure_url":"http://testurl.com","created_at":"2023-10-11T16:15:54.925759+00:00","connection_configs":null,"data_stewards":[],"cookies":[]}
```
* [ ] confirm that a sample TCF overlay has these fields populated

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### DIFF
--- a/src/fides/api/alembic/migrations/versions/81886da90395_add_legal_basis_dimension.py
+++ b/src/fides/api/alembic/migrations/versions/81886da90395_add_legal_basis_dimension.py
@@ -4,7 +4,7 @@ We've previously been saving preferences against purposes, vendors, and systems,
 need to instead save preferences against a purpose/vendor AND a legal basis.
 
 Revision ID: 81886da90395
-Revises: 4cb3b5af4160
+Revises: 9b98aba5bba8
 Create Date: 2023-09-30 17:39:46.251444
 
 """
--- a/src/fides/api/util/tcf/tcf_experience_contents.py
+++ b/src/fides/api/util/tcf/tcf_experience_contents.py
@@ -426,49 +426,47 @@ def build_purpose_or_feature_section_and_update_vendor_map(
 
 
 def populate_vendor_relationships_basic_attributes(
-    vendor_map,
+    vendor_map: Dict[str, TCFVendorRelationships],
     matching_privacy_declarations: Query,
-):
+) -> Dict[str, TCFVendorRelationships]:
     """Populates TCFVendorRelationships records for all vendors that we're displaying in the overlay.
     Ensures that these TCFVendorRelationships records have the "basic" TCF attributes populated.
     """
     for privacy_declaration_row in matching_privacy_declarations:
         vendor_id, system_identifier = get_system_identifiers(privacy_declaration_row)
-        if vendor_id:  # this only applies to GVL vendors
-            # Add top-level entry to the vendor relationship map
-            if system_identifier not in vendor_map:
-                vendor_map[system_identifier] = TCFVendorRelationships(
-                    id=system_identifier,  # Identify system by vendor id if it exists, otherwise use system id.
-                    name=privacy_declaration_row.system_name,
-                    description=privacy_declaration_row.system_description,
-                    has_vendor_id=bool(
-                        vendor_id
-                    ),  # Has_vendor_id will let us separate data between systems and vendors
-                    cookie_max_age_seconds=privacy_declaration_row.system_cookie_max_age_seconds,
-                    uses_cookies=privacy_declaration_row.system_uses_cookies,
-                    cookie_refresh=privacy_declaration_row.system_cookie_refresh,
-                    uses_non_cookie_access=privacy_declaration_row.system_uses_non_cookie_access,
-                    legitimate_interest_disclosure_url=privacy_declaration_row.system_legitimate_interest_disclosure_url,
-                )
 
-            else:
-                # otherwise just ensure that our basic attributes are populated on the vendor relationship entry
-                vendor_relationships = vendor_map[system_identifier]
-                vendor_relationships.cookie_max_age_seconds = (
-                    privacy_declaration_row.system_cookie_max_age_seconds
-                )
-                vendor_relationships.uses_cookies = (
-                    privacy_declaration_row.system_uses_cookies
-                )
-                vendor_relationships.cookie_refresh = (
-                    privacy_declaration_row.system_cookie_refresh
-                )
-                vendor_relationships.uses_non_cookie_access = (
-                    privacy_declaration_row.system_uses_non_cookie_access
-                )
-                vendor_relationships.legitimate_interest_disclosure_url = (
-                    privacy_declaration_row.system_legitimate_interest_disclosure_url
-                )
+        # Get the existing TCFVendorRelationships record or create a new one.
+        # Add to the vendor map if it wasn't added in a previous section.
+        vendor_relationship_record = vendor_map.get(system_identifier)
+        if not vendor_relationship_record:
+            vendor_relationship_record = TCFVendorRelationships(
+                id=system_identifier,  # Identify system by vendor id if it exists, otherwise use system id.
+                name=privacy_declaration_row.system_name,
+                description=privacy_declaration_row.system_description,
+                has_vendor_id=bool(
+                    vendor_id  # This will let us separate data between systems and vendors later
+                ),
+            )
+            vendor_map[system_identifier] = vendor_relationship_record
+
+        # Now add basic attributes to the VendorRelationships record
+        vendor_relationship_record.cookie_max_age_seconds = (
+            privacy_declaration_row.system_cookie_max_age_seconds
+        )
+        vendor_relationship_record.uses_cookies = (
+            privacy_declaration_row.system_uses_cookies
+        )
+        vendor_relationship_record.cookie_refresh = (
+            privacy_declaration_row.system_cookie_refresh
+        )
+        vendor_relationship_record.uses_non_cookie_access = (
+            privacy_declaration_row.system_uses_non_cookie_access
+        )
+        vendor_relationship_record.legitimate_interest_disclosure_url = (
+            privacy_declaration_row.system_legitimate_interest_disclosure_url
+        )
+
+    return vendor_map
 
 
 def get_tcf_contents(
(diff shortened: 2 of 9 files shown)


src/fides/api/alembic/migrations/versions/81886da90395_add_legal_basis_dimension.py
```diff
@@ -4,7 +4,7 @@
 need to instead save preferences against a purpose/vendor AND a legal basis.
 
 Revision ID: 81886da90395
-Revises: 4cb3b5af4160
+Revises: 9b98aba5bba8
```
adamsachs, line 7: just patching up migration comment in previous migration to keep a clean paper trail 👍

adamsachs:
i've got a changelog commit queued up locally, just didn't want to push to re-trigger CI - writing this so i don't forget to push it up!

pattisdr:
starting review -

src/fides/api/alembic/migrations/versions/81886da90395_add_legal_basis_dimension.py
pattisdr: Ah thank you 🙏

tests/ctl/core/test_api.py
```diff
@@ -654,6 +659,14 @@ async def test_system_create(
             system.data_security_practices
             == "We encrypt all your data in transit and at rest"
         )
+        assert system.cookie_max_age_seconds == 31536000
+        assert system.uses_cookies == True
+        assert system.cookie_refresh == True
+        assert system.uses_non_cookie_access == True
```
pattisdr, line 665: minor, in Python, you'd typically want to do the comparison like: `system.uses_cookies is True` for a boolean, like you did below with `is None` but only worth changing if you're pushing another commit for CI

tests/ops/util/test_tcf_experience_contents.py
```diff
@@ -233,6 +244,94 @@ def test_system_exists_with_tcf_purpose_and_vendor(self, db):
...
+
+        assert tcf_contents.tcf_purpose_consents[0].id == 8
+        assert tcf_contents.tcf_purpose_consents[0].data_uses == [
+            "analytics.reporting.content_performance"
+        ]
+        assert tcf_contents.tcf_purpose_consents[0].vendors == [
+            EmbeddedVendor(id="sendgrid", name="TCF System Test")
+        ]
+
+        assert tcf_contents.tcf_vendor_consents[0].id == "sendgrid"
```
pattisdr, line 297: This is something that will be affected with my PR: https://github.com/ethyca/fides/pull/4256, I stop making the vendor_id: "sendgrid".  Just something to sort out on merge, for whoever merges second 😄

src/fides/api/util/tcf/tcf_experience_contents.py
```diff
@@ -398,6 +405,11 @@ def build_purpose_or_feature_section_and_update_vendor_map(
                     has_vendor_id=bool(
                         vendor_id
                     ),  # Has_vendor_id will let us separate data between systems and vendors
+                    cookie_max_age_seconds=privacy_declaration_row.system_cookie_max_age_seconds,
+                    uses_cookies=privacy_declaration_row.system_uses_cookies,
+                    cookie_refresh=privacy_declaration_row.system_cookie_refresh,
+                    uses_non_cookie_access=privacy_declaration_row.system_uses_non_cookie_access,
+                    legitimate_interest_disclosure_url=privacy_declaration_row.system_legitimate_interest_disclosure_url,
```
pattisdr, line 412: This makes sense given how we're currently structuring the TCF overlay contents, but worth noting this creates duplication when we intentionally created a separate "vendor relationships" section to reduce duplication in the first place.  I think your implementation is fine, provided we're not making it harder for the FE to combine vendors.


---
More info below:

The TCF overlay has three vendor sections: 1) Vendor Consent 2) Vendor Legitimate Interests and 3) Vendor Relationships.  The same vendor can show up in one or many sections, provided it has associated content for those sections.  For example, if a vendor has a purpose it's collecting with a consent legal basis, it will show up in 1) Vendor Consent section.  If a Vendor has a Special Feature (plus other criteria) it will show up in 3) Vendor Relationships.    

3) Vendor Relationships surfaces Vendors with Special Purposes, Features, and Special Features to reduce duplicating this across 1) and 2).  Now we're duplicating Cookie info across 1) 2) and 3).  

Other alternatives might be to have the Vendor always show up in 3) if it appears in 1) or 2) and only collect cookie information there.

pattisdr (approved):
Looks great @adamsachs. I know we're not ready to bump a fideslang requirement here, but did you test locally with a fideslang alpha release here to make sure this will work as expected?

tests/ctl/core/test_api.py
adamsachs: gosh i don't know why i always struggle to get this right 😄 i'll update, it's sloppy - this is what happens when developing in a rush. thanks for catching 👍

src/fides/api/util/tcf/tcf_experience_contents.py
adamsachs: hmmm, i still don't totally understand this even after reading it a few times, and i know you'd also tried to point me to it last week! 😆 

i feel like we should probably defer to what's best for the FE as long as it's reasonable, and let's just get it in place here while we're at it... @allisonking, thoughts? all 3 of us could sync on this to get the right path forward quickly?

adamsachs:
> Looks great @adamsachs. I know we're not ready to bump a fideslang requirement here, but did you test locally with a fideslang alpha release here to make sure this will work as expected?

yup! the base branch is pointed at a fideslang alpha tag, so this branch is too, it's just not showing up in the diff 👍

pattisdr:
> yup! the base branch is pointed at a fideslang alpha tag, so this branch is too, it's just not showing up in the diff 👍

Ah! 😅 missed that thank you

src/fides/api/util/tcf/tcf_experience_contents.py
pattisdr: Updating here - we've pushed ahead with the fact that a vendor should always show up in VendorRelationships every time, and we are only collecting the basic attributes in that section

src/fides/api/util/tcf/tcf_experience_contents.py
```diff
@@ -523,6 +564,11 @@ def get_tcf_contents(
         matching_privacy_declaration_query=matching_privacy_declarations,
     )
 
+    populate_vendor_relationships_basic_attributes(
+        vendor_map=updated_vendor_relationships_map,
+        matching_privacy_declarations=matching_privacy_declarations,
+    )
```
pattisdr: Nice Adam, this was a faster way forward. I was thinking of updating `build_purpose_or_feature_section_and_update_vendor_map` to add special handling for the vendor relationships section but that would add bloat to an already confusing method to parse.  I think this extra loop through the privacy declarations makes sense for readability.

pattisdr:
OK I think this is ready when CI passes - I'm waiting on whether separate https://github.com/ethyca/fides/pull/4256 should be merged to main since these sets of changes need to be reconciled. Wanted to double check since I've missed some context around the integration branch the last couple of days.

src/fides/api/util/tcf/tcf_experience_contents.py
adamsachs: yup, i think there were a few ways to skin the cat, probably pros/cons to each of them. glad what i arrived at reads easy enough and seems to get the job done, we can/should come back to it for performance optimizations or refactoring generally!

src/fides/api/util/tcf/tcf_experience_contents.py
```diff
@@ -564,7 +562,11 @@ def get_tcf_contents(
         matching_privacy_declaration_query=matching_privacy_declarations,
     )
 
-    populate_vendor_relationships_basic_attributes(
+    # Finally, add missing TCFVendorRelationships records for vendors that weren't already added
+    # via the special_features, features, and special_purposes section.  Every vendor in the overlay
+    # should show up in this section. Add the basic attributes to the vendor here to avoid duplication
+    # in other vendor sections.
```
adamsachs, line 568: 🙏 thank you @pattisdr for the extra care in adding this comment, will help make sense of things moving forward!

adamsachs:
> OK I think this is ready when CI passes - I'm waiting on whether separate #4256 should be merged to main since these sets of changes need to be reconciled. Wanted to double check since I've missed some context around the integration branch the last couple of days.

great, i'll go ahead and merge this one first since i think it's got more dependencies on it 🤷 

we need to make both work, and i don't think the conflicts will be different no matter which we merge first...

Issue #4210, opened 2023-10-02
Handle legal basis dimension for `purposes`


allisonking:
### Is your feature request related to a specific problem?

https://github.com/ethyca/fidesplus/issues/1128 adds a legal basis dimension for purposes. There's already been some work to handle this for vendors, but we need to do the same work for purposes.

### Describe the solution you'd like

After the backend is updated, we need to update all the fides-js code to split purposes up by legal basis. This will require:
* updates to the cookie saving logic
* updates to the UI to render multiple toggles for purposes

### Describe alternatives you've considered, if any

_A description of any alternative solutions or features you've considered._

### Additional context

https://github.com/ethyca/fides/issues/4209

Do we still want the legal basis filter to work in the same way for this? It has always felt like a filter dropdown to me, but it seems like it's really meant to totally switch the view.

allisonking:
Closed by https://github.com/ethyca/fides/pull/4234

Roger-Ethyca:
moving to done

PR #4234, opened 2023-10-05
Purpose legal basis

allisonking:
Closes https://github.com/ethyca/fides/issues/4210

### Description Of Changes

Purposes now have double vendor toggles! Also did the work of enabling all for each column, as well as the All On / All Off buttons.

https://github.com/ethyca/fides/assets/24641006/2a9d8866-edac-4e05-9216-4625c3f35128

### Code Changes

* [x] Make a new component `DoubleToggleTable` to deal with double toggles for both vendors and purposes
* [x] Refactor purposes to use the new `DoubleToggleTable`
* [x] Add a header option to toggle all legitimate interests or all consents on/off at once
* [x] Add "All on"/"All off" buttons
* [x] Update cypress tests

### Steps to Confirm

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### DIFF
--- a/clients/fides-js/src/components/fides.css
+++ b/clients/fides-js/src/components/fides.css
@@ -724,6 +724,10 @@ div#fides-modal .fides-modal-button-group {
   color: var(--fides-overlay-body-font-color);
 }
 
+.fides-primary-text-color {
+  color: var(--fides-overlay-primary-color);
+}
+
 /* Link */
 .fides-external-link {
   color: var(--fides-overlay-primary-color);
@@ -768,4 +772,15 @@ div#fides-modal .fides-modal-button-group {
 
 .fides-info-box p {
   margin: 0;
+}
+
+/* All on off buttons */
+.fides-all-on-off-buttons {
+  display: flex;
+  justify-content: end;
+  margin-bottom: 1em;
+}
+
+.fides-all-on-button {
+  margin-right: 0.9em;
 }
--- a/clients/fides-js/src/components/tcf/DoubleToggleTable.tsx
+++ b/clients/fides-js/src/components/tcf/DoubleToggleTable.tsx
@@ -38,7 +38,7 @@ const DoubleToggleTable = <T extends Item>({
   consentModelType,
   legintModelType,
 }: Props<T>) => {
-  const toggleAllId = `toggle-all-${title}`;
+  const toggleAllId = `all-${title}`;
   const toggleAllConsentId = `${toggleAllId}-consent`;
 
   const { allConsentChecked, allLegintChecked } = useMemo(() => {
--- a/clients/fides-js/src/components/tcf/DoubleToggleTable.tsx
+++ b/clients/fides-js/src/components/tcf/DoubleToggleTable.tsx
@@ -1,8 +1,12 @@
-/** This isn't a table _yet_, but it probably should be fides#4190 */
+/** This isn't a table _yet_, but it is aspirationally named.
+ * We should probably convert it to a table at some point fides#4190 */
 
 import { VNode, h } from "preact";
+import { useMemo } from "react";
 import DataUseToggle from "../DataUseToggle";
 import Toggle from "../Toggle";
+import { EnabledIds, LegalBasisEnum } from "../../lib/tcf/types";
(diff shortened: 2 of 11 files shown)


clients/fides-js/src/components/fides.css
```diff
@@ -317,23 +317,6 @@ div#fides-modal .fides-modal-button-group {
   background: var(--fides-overlay-hover-color);
 }
 
-.fides-back-button {
```
allisonking, line 320: I hit the 20kb limit for `fides-js` 😵 so I went and looked through the CSS file for classes that aren't used anymore. that reduced by a whole .2kb!

allisonking: our CSS classes are all in this file. it might make sense to have a separate TCF css file which only `TcfOverlay` imports. that should reduce our vanilla `fides-js` even further (FYI @galvana , who has been doing a ton of CSS work!)

clients/fides-js/src/components/tcf/DoubleToggleTable.tsx
```diff
@@ -0,0 +1,166 @@
...
+  enabledConsentIds: string[];
+  enabledLegintIds: string[];
+  renderToggleChild: (item: T) => VNode;
+  onToggle: (payload: UpdateEnabledIds) => void;
+  renderBadgeLabel?: (item: T) => string | undefined;
+  consentModelType: keyof EnabledIds;
+  legintModelType: keyof EnabledIds;
+}
+
+const DoubleToggleTable = <T extends Item>({
```
allisonking, line 30: typescript templating!! super handy for this case, where I wanted the table to handle both vendors and purposes

allisonking: otherwise, this file is mostly the same as what used to be in `TcfVendors`, just generic'd a bit

clients/fides-js/src/components/tcf/DoubleToggleTable.tsx
```diff
@@ -0,0 +1,166 @@
...
+
+  return (
+    <div>
+      {/* DEFER: ideally we use a table object, but then DataUseToggles would need to be reworked
+      or we would need a separate component. */}
+      <div className="fides-legal-basis-labels">
+        <span className="fides-margin-right">Legitimate interest</span>
+        <span>Consent</span>
+      </div>
+      <DataUseToggle
```
allisonking, line 110: the one thing that did get added which `TcfVendors` didn't have is this title row which also has toggles:

image

clients/fides-js/src/components/tcf/TcfOverlay.tsx
```diff
@@ -234,14 +234,6 @@ const TcfOverlay: FunctionComponent<OverlayProps> = ({
     [experience]
   );
 
-  const handleUpdateDraftState = useCallback(
-    ({ newEnabledIds, modelType }: UpdateEnabledIds) => {
-      const updated = { ...draftIds, [modelType]: newEnabledIds };
-      setDraftIds(updated);
-    },
-    [draftIds]
-  );
```
allisonking, line 243: this is kind of a fun thing—`draftIds` is an object that looks something like

```js
{
  purposesConsent: [1,2,3],
  purposesLegint: [4,5,6]
  ...
}
```
and this function is perfect for updating one field at a time, since it can say "set purposesConsent to [1,2,3]". However, for the all on/all off buttons, I wanted to update _two_ fields at once. in other words, I wanted to say "set purposesConsent to [1,2,3] and purposesLegint to [4,5,6]". my first instinct was to just continue using this function but call it twice. BUT that doesn't work because setting react state is asynchronous! therefore the first call fires, but since we do a `{...draftIds, [modelType]: newEnabledIds }`, when the second call fires, `draftIds` is still the old state (not updated yet) and therefore it looks like the first state never fired! 

so we really want to be able to set both states at the same time. therefore, I now pass the entire `setDraftIds` down to `TcfTabs`. most things only want to set one thing at a time (so this function moved to `TcfTabs` but now the all on/all off buttons can set multiple at once)

eastandwestwind (approved):
This looks solid, thanks @allisonking !

PR #4243, opened 2023-10-09
Fix double toggle bug

allisonking:
Closes bug reported by @galvana !

### Description Of Changes

When a vendor has two toggles available, it was always setting consent instead of the proper one.
 
toggle

### Code Changes

* [x] Pass in a legal basis for individual toggles (we had this before I moved to the refactored DoubleVendorToggle component, and then I made a bad assumption while generalizing it 🙃 )
* [x] Add a cypress test

### Steps to Confirm

* [ ] Set up TCF with a vendor that has both legint and consent purposes
* [ ] Toggle its legitimate interest purpose
* [ ] It should work as expected

### Pre-Merge Checklist

### DIFF
--- a/clients/privacy-center/cypress/e2e/consent-banner-tcf.cy.ts
+++ b/clients/privacy-center/cypress/e2e/consent-banner-tcf.cy.ts
@@ -344,6 +344,43 @@ describe("Fides-js TCF", () => {
           });
         });
       });
+
+      it("can toggle double toggles individually", () => {
+        cy.fixture("consent/experience_tcf.json").then((payload) => {
+          const experience = payload.items[0];
+          // Add a vendor legitimate interest which is the same as vendor consent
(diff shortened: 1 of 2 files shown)


clients/privacy-center/cypress/e2e/consent-banner-tcf.cy.ts
```diff
@@ -359,10 +396,9 @@ describe("Fides-js TCF", () => {
               { id: PURPOSE_7.id, preference: "opt_in" },
               { id: PURPOSE_9.id, preference: "opt_in" },
             ]);
-            // TODO: fides#4210
```
allisonking, line 362: I realized I left all these TODOs which I should've uncommented out in https://github.com/ethyca/fides/pull/4234 but forgot to!

galvana (approved):

galvana:
Thanks for the quick fix @allisonking, it works as expected!

Issue #4212, opened 2023-10-02
Support AC string in fides-tcf


allisonking:
### Is your feature request related to a specific problem?

We need to support [Google's Additional Consent Mode](https://support.google.com/admanager/answer/9681920?hl=en) which includes its own list of vendors, and its own string.

### Describe the solution you'd like

* [x] Add a function `vendorIsAC` similar to our existing `vendorIsGvl` function. Or, depending on how the BE gives us AC data, we may not need this if the BE is already calculating this.
* [x] Add a function `generateAcString` to generate an AC string (similar to the existing `generateTcString`)
* [x] Add [custom commands](https://github.com/InteractiveAdvertisingBureau/iabtcf-es/tree/master/modules/cmpapi#built-in-and-custom-commands) to enable sending along the `addtlConsent` field
* [x] Include the AC string as part of the TC string when we save, but remove it when we send it via the CMP API
~~* [ ] Render an AC badge on vendors that are part of the AC. Also render its AC id (design needed?)~~

### Describe alternatives you've considered, if any

_A description of any alternative solutions or features you've considered._

### Additional context

Design doc: https://ethyca.atlassian.net/wiki/spaces/EN/pages/2801303987/fides-js+AC+string

allisonking:
The BE is going to update its vendor id's to be prefixed by its data source, i.e. `gvl.2` or `ac.42`. that will let us figure out if a vendor is AC or not. 

See this ticket for updating for GVL: https://github.com/ethyca/fides/issues/4242

Issue #4165, opened 2023-09-26
Only list vendors in the data map as `vendors_disclosed` in the FE


allisonking:
### Bug Description

After another reread of the spec, we think it makes sense to only list the vendors in both the datamap and the GVL in the `vendors_disclosed` section of the TC string. See the discussion here: https://github.com/ethyca/fides/pull/4160#discussion_r1337694296

The backend has gone ahead and made this change, so we should do something similar for the FE!

#### Steps to Reproduce

1. [Set up your TCF environment](https://ethyca.atlassian.net/wiki/spaces/EN/pages/2795438113/Setting+up+TCF) 
1. Opt in or opt out of all 
2. Check your TC string—it'll have a huge vendors disclosed section since it lists every vendor in the GVL when it should really only list the vendors we are showing in the vendor tab.

### Expected behavior

The vendors disclosed section should only show the vendors in the vendor tab.

### Screenshots

_If applicable, add screenshots to help explain your problem._

### Environment

* Version:
* OS:
* Python Version:
* Docker Version:

### Additional context

_Add any other context about the problem here._

allisonking:
Actually, it looks like we should not be including this string at all when the CMP API surfaces it 
 🤔 https://ethyca.slack.com/archives/C05E74TR0BB/p1696341554803559

allisonking:
Okay we no longer surface the `vendors_disclosed` string as part of the CMP API: https://github.com/ethyca/fides/pull/4244/files

However, I am still confused about how to actually properly set the `vendors_disclosed` string, as the library seems to insist that we set all the GVL vendors. more investigation needed!

PR #4244, opened 2023-10-09
Frontend AC string support

allisonking:
Closes https://github.com/ethyca/fides/issues/4212

### Description Of Changes

Support AC string in fides-js!

### Code Changes

* [x] Add a function `decodeVendorId` based on [upcoming backend refactor](https://github.com/ethyca/fidesplus/issues/1129#issue-1916394861)
* [x] Add a function vendorIsAc
* [x] Add a function generateAcString to generate an AC string (similar to the existing generateTcString)
* [x] Add [custom commands](https://github.com/InteractiveAdvertisingBureau/iabtcf-es/tree/master/modules/cmpapi#built-in-and-custom-commands) to enable sending along the `addtlConsent` field
* [x] Rename `tc_string` in the FidesCookie to `fides_tc_string`
* [x] Cypress tests

### Steps to Confirm

* This is difficult to confirm outside of Cypress tests right now since we don't have universal vendor IDs on the backend yet. But if you play around with TCF as usual, you should notice a few things:
  * When saving the cookie, we now save the tc string to `fides_tc_string`, and it'll end with `,1~"`. in theory, once we have AC vendor ids working, that string will fill in those AC ids so it'll be more like `1~[id].[id2]` etc.
  * Whenever we print out Fides events, the event will include this composite string
  * When looking at tcData events, there should not be the composite string (but it'll be included in `tcData` under `addtlConsents`. To look at tcdata events, you can paste this into your console:
  
```js
__tcfapi("getTCData", 2, (data) => console.log({data}))
```
and you should see something like

```json
{
    "data": {
        "cmpId": 407,
        "tcString": "CPzYJkAPzYJkAGXABBENAUEIAFKAAAAAAAAAABEAAAAA",
        "addtlConsent": "1~"
    }
}
```

### Pre-Merge Checklist

* [x] Update `CHANGELOG.md`

### DIFF
--- a/clients/fides-js/src/fides-tcf.ts
+++ b/clients/fides-js/src/fides-tcf.ts
@@ -86,7 +86,7 @@ const updateCookie = async (
 ): Promise<FidesCookie> => {
   // First check if the user has never consented before
   if (!hasSavedTcfPreferences(experience)) {
-    return { ...oldCookie, tc_string: "" };
+    return { ...oldCookie, fides_tc_string: "" };
   }
 
   // Usually at this point, we'd look at the Experience from the backend and update
--- a/clients/fides-js/src/fides-tcf.ts
+++ b/clients/fides-js/src/fides-tcf.ts
@@ -52,7 +52,7 @@ import { shopify } from "./integrations/shopify";
 
 import { FidesConfig, PrivacyExperience } from "./lib/consent-types";
 
-import { tcf } from "./lib/tcf";
+import { initializeCmpApi } from "./lib/tcf";
 import {
   getInitialCookie,
   getInitialFides,
@@ -103,6 +103,8 @@ const updateCookie = async (
 const init = async (config: FidesConfig) => {
   const cookie = getInitialCookie(config);
   const initialFides = getInitialFides({ ...config, cookie });
+  // Initialize the CMP API early so that listeners are established
+  initializeCmpApi();
   if (initialFides) {
     Object.assign(_Fides, initialFides);
     dispatchFidesEvent("FidesInitialized", cookie, config.options.debug);
@@ -126,8 +128,6 @@ const init = async (config: FidesConfig) => {
     dispatchFidesEvent("FidesInitialized", cookie, config.options.debug);
   }
   dispatchFidesEvent("FidesUpdated", cookie, config.options.debug);
-
-  tcf();
 };
 
 // The global Fides object; this is bound to window.Fides if available
--- a/clients/fides-js/src/lib/tcf.ts
+++ b/clients/fides-js/src/lib/tcf.ts
@@ -155,7 +155,7 @@ export const generateTcString = async ({
  * `vendors_disclosed` and our own AC string addition.
  */
 const fidesEventToTcString = (event: FidesEvent) => {
-  const { tc_string: cookieString } = event.detail;
+  const { fides_tc_string: cookieString } = event.detail;
   if (cookieString) {
     // Remove the AC portion which is separated by FIDES_SEPARATOR
     const [tcString] = cookieString.split(FIDES_SEPARATOR);
@@ -181,7 +181,8 @@ export const initializeCmpApi = () => {
       /*
        * If using with 'removeEventListener' command, add a check to see if tcData is not a boolean. */
       if (typeof tcData !== "boolean") {
-        const stringSplit = window.Fides.tc_string?.split(FIDES_SEPARATOR);
+        const stringSplit =
+          window.Fides.fides_tc_string?.split(FIDES_SEPARATOR);
         const addtlConsent = stringSplit?.length === 2 ? stringSplit[1] : "";
         next({ ...tcData, addtlConsent }, status);
       }
--- a/clients/fides-js/src/lib/tcf.ts
+++ b/clients/fides-js/src/lib/tcf.ts
@@ -42,7 +42,7 @@ const generateAcString = ({
         .map((id) => decodeVendorId(id).id)
     )
   );
-  const vendorIds = uniqueIds.join(".");
+  const vendorIds = uniqueIds.sort().join(".");
 
   return `${AC_SPECIFICATION_VERSION}~${vendorIds}`;
 };
--- a/clients/fides-js/src/lib/tcf.ts
+++ b/clients/fides-js/src/lib/tcf.ts
@@ -165,7 +165,7 @@ const fidesEventToTcString = (event: FidesEvent) => {
     // We only want to return the first part of the tcString, which is separated by '.'
     // This means Publisher TC is not sent either, which is okay for now since we do not set it.
     // However, if we do one day set it, we would have to decode the string and encode it again
-    // without vendor_consents
+    // without vendorsDisclosed
     return tcString.split(".")[0];
   }
   return cookieString;
--- a/clients/fides-js/src/lib/tcf.ts
+++ b/clients/fides-js/src/lib/tcf.ts
@@ -10,12 +10,40 @@ import { TCModel, TCString, GVL } from "@iabtechlabtcf/core";
 import { makeStub } from "./tcf/stub";
 
 import { EnabledIds } from "./tcf/types";
-import { vendorIsGvl } from "./tcf/vendors";
+import { vendorIsAc, vendorIsGvl } from "./tcf/vendors";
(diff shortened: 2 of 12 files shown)


clients/fides-js/src/fides-tcf.ts
```diff
@@ -126,8 +128,6 @@ const init = async (config: FidesConfig) => {
     dispatchFidesEvent("FidesInitialized", cookie, config.options.debug);
   }
   dispatchFidesEvent("FidesUpdated", cookie, config.options.debug);
-
-  tcf();
```
allisonking, line 130: moving this farther up in the lifecycle is actually kind of important, otherwise the CMP API doesn't get initialized (since originally it wasn't called until after the `FidesInitialized` events). this resulted in a state where if you opened the IAB's chrome extension before opting in or out of anything, it wouldn't have any information, even though it should if we had previously agreed to something

clients/fides-js/src/lib/tcf.ts
```diff
@@ -119,37 +153,73 @@ export const generateTcString = async ({
...
-export const tcf = () => {
+const fidesEventToTcString = (event: FidesEvent) => {
+  const { fides_tc_string: cookieString } = event.detail;
+  if (cookieString) {
+    // Remove the AC portion which is separated by FIDES_SEPARATOR
+    const [tcString] = cookieString.split(FIDES_SEPARATOR);
+    // We only want to return the first part of the tcString, which is separated by '.'
+    // This means Publisher TC is not sent either, which is okay for now since we do not set it.
+    // However, if we do one day set it, we would have to decode the string and encode it again
+    // without vendor_consents
```
allisonking, line 168: this sneaks in a fix that is part of https://github.com/ethyca/fides/issues/4165 which says we should not send `vendors_disclosed` in the CMP API

clients/fides-js/src/lib/tcf.ts
```diff
@@ -119,37 +153,73 @@ export const generateTcString = async ({
...
+ * Initializes the CMP API, including setting up listeners on FidesEvents to update
+ * the CMP API accordingly.
+ */
+export const initializeCmpApi = () => {
   makeStub();
   const isServiceSpecific = true; // TODO: determine this from the backend?
-  const cmpApi = new CmpApi(CMP_ID, CMP_VERSION, isServiceSpecific);
+  const cmpApi = new CmpApi(CMP_ID, CMP_VERSION, isServiceSpecific, {
+    // Add custom command to support adding `addtlConsent` per AC spec
+    getTCData: (next, tcData: TCData, status) => {
```
allisonking, line 183: adapted from the docs: https://github.com/InteractiveAdvertisingBureau/iabtcf-es/tree/master/modules/cmpapi#built-in-and-custom-commands

clients/privacy-center/cypress/e2e/consent-banner-tcf.cy.ts
```diff
@@ -749,4 +749,158 @@ describe("Fides-js TCF", () => {
...
+          current_preference: null,
+          outdated_preference: null,
+          current_served: null,
+          outdated_served: null,
+          purpose_consents: [
+            {
+              id: 4,
+              name: "Use profiles to select personalised advertising",
+            },
+          ],
```
NevilleS, line 776: Can you ensure there are some tests _without_ any purposes?

clients/fides-js/src/lib/tcf/constants.ts
```diff
@@ -1,5 +1,7 @@
 import { TcfExperienceRecords, TcfModelType } from "./types";
 
+export const FIDES_SEPARATOR = ",";
```
NevilleS, line 3: This probably need a comment, just to explain why the "Fides TC String" may include both the IAB TCF string and the AC string, together. The TL;DR of that is because we want a single-source-of-truth for offline storage & syncing of these kinds of preference strings, but we can't wait for GPP to solve that so we can approximate it here.

Important to also state that this doesn't affect the standards-compliant JS APIs that fetch TCF and AC strings separately!

clients/privacy-center/cypress/e2e/consent-banner-tcf.cy.ts
```diff
@@ -749,4 +749,158 @@ describe("Fides-js TCF", () => {
...
+        .then(([tcData, success]) => {
+          expect(success).to.eql(true);
+          expect(tcData.eventStatus).to.eql("useractioncomplete");
+          // This TC string should not be a composite—should just be the tc string
+          const { tcString } = tcData;
+          const parts = tcString.split(",");
+          expect(parts.length).to.eql(1);
+          expect(parts[0]).to.not.contain("1~");
+          // But we can still access the AC string via `addtlConsent`
+          expect(tcData.addtlConsent).to.eql(acceptAllAcString);
```
NevilleS, line 882: Nice 👍

clients/fides-js/src/lib/tcf.ts
```diff
@@ -5,18 +5,48 @@
...
+import { FIDES_SEPARATOR } from "./tcf/constants";
+import { FidesEvent } from "./events";
 
+// TCF
 const CMP_ID = 407;
 const CMP_VERSION = 1;
 const FORBIDDEN_LEGITIMATE_INTEREST_PURPOSE_IDS = [1, 3, 4, 5, 6];
 
+// AC
+const AC_SPECIFICATION_VERSION = 1;
```
pattisdr, line 24: How do we know this is 1?

clients/fides-js/src/lib/tcf.ts
```diff
@@ -5,18 +5,48 @@
...
+      [
+        ...tcStringPreferences.vendorsConsent,
+        ...tcStringPreferences.vendorsLegint,
+      ]
+        .filter((id) => vendorIsAc(id))
+        // Convert ac.42 --> 42
+        .map((id) => decodeVendorId(id).id)
+    )
+  );
+  const vendorIds = uniqueIds.join(".");
```
pattisdr, line 45: do VendorIds need to be sorted ascending? I don't see this specifically in the spec, but see that they're sorted in examples.

clients/fides-js/src/lib/tcf.ts
allisonking: oh right, I meant to revisit this—that's what they had in their docs as an example and I couldn't find anywhere else that had a version. let me ask if anyone else has come across one!

clients/fides-js/src/lib/tcf.ts
allisonking: I'll sort, doesn't hurt!

clients/fides-js/src/lib/tcf.ts
allisonking: okay @NevilleS confirmed we should use `1`!

allisonking: yay! luckily it's much easier than the TC string 😄

clients/fides-js/src/lib/tcf.ts
```diff
@@ -119,37 +153,73 @@ export const generateTcString = async ({
 };
 
 /**
- * Call tcf() to configure Fides with tcf support (if tcf is enabled).
+ * Extract just the TC string from a FidesEvent. This will also remove parts of the
+ * TC string that we do not want to surface with our CMP API events, such as
+ * `vendors_disclosed` and our own AC string addition.
  */
-export const tcf = () => {
+const fidesEventToTcString = (event: FidesEvent) => {
```
eastandwestwind, line 160: awesome, I think this means I'll have to make minimal changes, if any, on my TC str override work 🌟

clients/fides-js/src/lib/tcf/vendors.ts
```diff
@@ -17,6 +22,20 @@ export const vendorIsGvl = (
   return gvl.vendors[vendor.id];
 };
 
+/**
+ * Given a vendor id such as `gvl.2`, return {source: "gvl", id: "2"};
+ */
+export const decodeVendorId = (vendorId: TCFVendorRelationships["id"]) => {
```
eastandwestwind, line 28: can we rename this to `createSourceToVendorIdMap` or `splitSourceAndVendorId` or something?

eastandwestwind (approved):
Looks really solid! One naming nit, but not blocking if you disagree on it 😄

PR #4250, opened 2023-10-10
Narrow vendors disclosed to only the vendors we show in the UI

allisonking:
Closes https://github.com/ethyca/fides/issues/4165

### Description Of Changes

The library we use wants to put every vendor in the GVL as part of the `vendors_disclosed` string. According to the docs, we should only put the vendors we've disclosed, i.e. the ones we've shown in the UI (makes sense!). There's a `vendors_disclosed` field we can set, but that appears to always be overwritten by the library to set the whole GVL. I believe the way around this is to use the [library's `narrowVendorsTo` function](https://github.com/InteractiveAdvertisingBureau/iabtcf-es/tree/master/modules/core#narrow-the-list-of-vendors)

Part of this ticket is also to not surface the vendor disclosed part of the string to the CMP API—that's covered in https://github.com/ethyca/fides/pull/4244

### Code Changes

* [x] Narrow the GVL to just the vendors we show in the UI
* [x] Add test

### Steps to Confirm

* Set up your TCF environment
* Opt in to some vendors. Inspect your cookie's `tc_string` object. it should have two parts, separated by a period. The second part should be very short (it used to be a very long string that ended with `AAAAAAAAAAAAAAAAQ`)
* If you decode this string [here](https://iabtcf.com/#/decode) you should see that there are only a few vendors disclosed (the ones shown in the UI)

### Pre-Merge Checklist

* [x] Update `CHANGELOG.md`

### DIFF
--- a/clients/fides-js/src/lib/tcf.ts
+++ b/clients/fides-js/src/lib/tcf.ts
@@ -39,17 +39,22 @@ export const generateTcString = async ({
     tcModel.cmpVersion = CMP_VERSION;
     tcModel.consentScreen = 1; // todo- On which 'screen' consent was captured; this is a CMP proprietary number encoded into the TC string
 
+    // Narrow the GVL to say we've only showed these vendors provided by our experience
+    const vendorIds = [
+      ...(experience.tcf_vendor_consents?.map((v) => +v.id) || []),
+      ...(experience.tcf_vendor_legitimate_interests?.map((v) => +v.id) || []),
+    ];
+    tcModel.gvl.narrowVendorsTo(vendorIds);
+
     if (tcStringPreferences) {
-      if (
-        tcStringPreferences.vendorsConsent &&
-        tcStringPreferences.vendorsConsent.length > 0
-      ) {
-        tcStringPreferences.vendorsConsent.forEach((vendorId) => {
-          if (vendorIsGvl({ id: vendorId }, experience.gvl)) {
-            tcModel.vendorConsents.set(+vendorId);
-          }
-        });
-        tcStringPreferences.vendorsLegint.forEach((vendorId) => {
+      // Set vendors on tcModel
+      tcStringPreferences.vendorsConsent.forEach((vendorId) => {
+        if (vendorIsGvl({ id: vendorId }, experience.gvl)) {
+          tcModel.vendorConsents.set(+vendorId);
+        }
+      });
+      tcStringPreferences.vendorsLegint.forEach((vendorId) => {
+        if (vendorIsGvl({ id: vendorId }, experience.gvl)) {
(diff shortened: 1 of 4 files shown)


clients/fides-js/src/lib/tcf.ts
```diff
@@ -39,17 +39,22 @@ export const generateTcString = async ({
...
+      ...(experience.tcf_vendor_consents?.map((v) => +v.id) || []),
+      ...(experience.tcf_vendor_legitimate_interests?.map((v) => +v.id) || []),
+    ];
+    tcModel.gvl.narrowVendorsTo(vendorIds);
+
     if (tcStringPreferences) {
-      if (
-        tcStringPreferences.vendorsConsent &&
-        tcStringPreferences.vendorsConsent.length > 0
-      ) {
```
allisonking, line 46: I don't think we need these `if`s: it's always an array, and even if it's empty, the `forEach` below won't do anything

clients/fides-js/src/lib/tcf.ts
```diff
@@ -39,17 +39,22 @@ export const generateTcString = async ({
     tcModel.cmpVersion = CMP_VERSION;
     tcModel.consentScreen = 1; // todo- On which 'screen' consent was captured; this is a CMP proprietary number encoded into the TC string
 
+    // Narrow the GVL to say we've only showed these vendors provided by our experience
+    const vendorIds = [
+      ...(experience.tcf_vendor_consents?.map((v) => +v.id) || []),
+      ...(experience.tcf_vendor_legitimate_interests?.map((v) => +v.id) || []),
```
allisonking, line 45: ah, Dawn's commit here https://github.com/ethyca/fides/pull/4256/commits/7b86c153cafb9701aef72642d0589badfa802aa1 reminded me that in addition we should filter to just the IDs that are GVL, now that we are supporting AC. I will do that shortly

clients/fides-js/src/lib/tcf.ts
```diff
@@ -39,17 +39,22 @@ export const generateTcString = async ({
...
+    ];
+    tcModel.gvl.narrowVendorsTo(vendorIds);
```
pattisdr, line 47: ah! narrowVendorsTo, good find

> narrowVendorsTo - narrows vendors represented in this GVL to the list of ids passed in

pattisdr (approved):
Looks great @allisonking, verified your vendors disclosed sections for accept all/reject all also match what the backend is generating.

Just that gvl filter left -

allisonking:
thank you @pattisdr !! 
> Just that gvl filter left -

thank you for the reminder—I knew there was something I left off of a PR somewhere but totally forgot it was this one! will address that now

PR #4256, opened 2023-10-11
Backend: Assume Universal Vendor Ids

pattisdr:
Closes https://github.com/ethyca/fidesplus/issues/1129

### Description Of Changes

Update all-things TCF on the BE (namely, overlay generation and preferences saving) to account for new "universal vendor ID" scheme.

In other words, we assume that TCF vendor preferences are saved against universal vendor ids, and also that universal vendor ids are returned in the experience.

Therefore, the main thing that needs to change on the backend is the translation to and from the TC string itself.

### Code Changes

- Vendor ID -> GVL Universal Vendor ID: Save privacy preferences under universal vendor id when unpacking vendor preferences from a TC string by adding the gvl prefix
- Universal Vendor ID -> Vendor ID: When building a TCModel from the TCF Experience (a prerequisite to building TC strings), convert universal vendor ids to vendor ids (integers)

### Steps to Confirm
* [ ] Set tcf_enabled=true in `fides.toml`
* [ ] `nox -s dev -- shell`
* [ ] Create [GVL system locally, Instructions here] - this creates a system with vendor id "gvl.8"(https://ethyca.atlassian.net/wiki/spaces/EN/pages/2809724964/API+How-to+Create+GVL+System+Locally)
* [ ] PATCH Privacy Preferences using the following accept-all TC string for the given system above:
PATCH http://localhost:8080/api/v1/privacy-preferences
```json
{
  "browser_identity": {
    "fides_user_device_id": "932209c6-e765-472c-b614-68f9b95b3278"
  },
  "tc_string": "CPzevcAPzevcAGXABBENAUEAALAAAEOAAAAAAEAEACACAAAA.IAEAEAAA"
}
```
* [ ] Verify vendor consent and legitimate preferences were saved against id "gvl.8" and tc_mobile_data > IABTCF_VendorConsents="00000001" and IABTCF_VendorLegitimateInterests="00000001":
```json
  "vendor_consent_preferences": [
    {
      "purpose_consent": null,
      "purpose_legitimate_interests": null,
      "special_purpose": null,
      "vendor_consent": "gvl.8",
      "vendor_legitimate_interests": null,
      "feature": null,
      "special_feature": null,
      "system_consent": null,
      "system_legitimate_interests": null,
      "id": "cur_dc981346-1dcb-4c21-adf1-7f8e0a88746e",
      "preference": "opt_in",
      "privacy_notice_history": null,
      "privacy_preference_history_id": "pri_25d3af5e-520b-4ae1-8903-0f6c0e0888ae"
    }
```
* [ ] Fetch the TCF Privacy Experience for EEA with include_meta=True and the same fides user device id from above (to retrieve currently saved preferences)
```bash
curl -X 'GET' \
  'http://localhost:8080/api/v1/privacy-experience?show_disabled=true&region=eea&fides_user_device_id=932209c6-e765-472c-b614-68f9b95b3278&systems_applicable=false&include_gvl=false&include_meta=true&page=1&size=50' \
  -H 'accept: application/json'
```
* [ ] Assert vendor consents and vendor legitimate interests are what is expected ("00000001") in the tc mobile data section, that the vendor consents and legitimate interests have "gvl.8" in them, and the previously saved preferences of opt-in for all are surfaced.

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### DIFF
--- a/src/fides/api/util/tcf/tc_model.py
+++ b/src/fides/api/util/tcf/tc_model.py
@@ -21,6 +21,17 @@ FORBIDDEN_LEGITIMATE_INTEREST_PURPOSE_IDS = [1, 3, 4, 5, 6]
 gvl: Dict = load_gvl()
 
 
+def universal_vendor_id_to_id(universal_vendor_id: str) -> int:
+    """Converts a universal vendor id to a vendor id
+
+    For example, converts "gvl.42" to integer 42.
+    Throws a ValueError if the id cannot be converted to an integer.
+
+    We store vendor ids as a universal vendor id internally, but need to strip this off when building TC strings.
+    """
+    return int(universal_vendor_id.lstrip("gvl.").lstrip("ac."))
+
+
 class TCModel(FidesSchema):
     """Base internal TC schema to store and validate key details from which to later build the TC String"""
 
@@ -288,7 +299,7 @@ def _build_vendor_consents_and_legitimate_interests(
 
     for vendor_consent in vendor_consents:
         try:
-            int(vendor_consent.id)
+            vendor_consent_id: int = universal_vendor_id_to_id(vendor_consent.id)
         except ValueError:
             # Early check that filters out non-integer vendor ids.  Later we'll run a separate
             # check that ensures this id is also in the gvl.
@@ -297,11 +308,11 @@ def _build_vendor_consents_and_legitimate_interests(
         if vendor_consent.purpose_consents:
             # This vendor should only be in "vendor_consents" in the first place, if consent_purposes is populated.
             # This is just to check
-            vendor_consent_ids.append(int(vendor_consent.id))
+            vendor_consent_ids.append(vendor_consent_id)
 
     for vendor_legitimate_interest in vendor_legitimate_interests:
         try:
-            int(vendor_legitimate_interest.id)
+            vendor_li_id: int = universal_vendor_id_to_id(vendor_legitimate_interest.id)
         except ValueError:
             # Early check that filters out non-integer vendor ids.  Later we'll run a separate
             # check that ensures this id is also in the gvl.
@@ -316,7 +327,7 @@ def _build_vendor_consents_and_legitimate_interests(
         if leg_int_purpose_ids and not bool(
             set(leg_int_purpose_ids) & set(FORBIDDEN_LEGITIMATE_INTEREST_PURPOSE_IDS)
         ):
-            vendor_legitimate_interest_ids.append(int(vendor_legitimate_interest.id))
+            vendor_legitimate_interest_ids.append(vendor_li_id)
 
     return vendor_consent_ids, vendor_legitimate_interest_ids
 
@@ -339,7 +350,7 @@ def _build_vendors_disclosed(tcf_contents: TCFExperienceContents) -> List[int]:
     The DisclosedVendors is an optional TC String segment that records which vendors have been disclosed to a
     given user by a CMP. It may be used by a CMP while storing TC Strings, but must not be included in the TC String
     when returned by the CMP API."""
-    all_vendor_ids: List[str] = [vendor_id for vendor_id in gvl.get("vendors", {})]
+    all_vendor_ids: List[int] = [int(vendor_id) for vendor_id in gvl.get("vendors", {})]
     vendors_disclosed = set()
 
     vendor_main_lists: List[List] = [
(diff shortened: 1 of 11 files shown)


src/fides/api/util/tcf/tc_model.py
```diff
@@ -288,7 +299,7 @@ def _build_vendor_consents_and_legitimate_interests(
 
     for vendor_consent in vendor_consents:
         try:
-            int(vendor_consent.id)
+            vendor_consent_id: int = universal_vendor_id_to_id(vendor_consent.id)
```
adamsachs, line 302: it's not strictly needed for our release increment, since we're not committing to AC support yet. but we'll need to remove AC vendors from compass if we're not supporting them. what would you think of updating this to check for whether the vendor is GVL, based on the prefix? we don't have to make this super robust/clean, but even a rudimentary check would mean that we can feel confident that an AC vendor that sneaks into a fides instance wouldn't make the TC string building here invalid -- to not have to worry about this in the interim until we have proper AC support.

does that make sense? am i understanding things correctly? let me know what you think.

src/fides/api/util/tcf/tc_string.py
```diff
@@ -214,13 +219,19 @@ def convert_to_fides_preference(
...
         preference: bool = tc_string_preferences.get(identifier, False)
 
-        if isinstance(preference_class, TCFVendorSave):
-            # Vendors are currently saved as strings in our db
-            identifier = str(identifier)
+        if preference_class == TCFVendorSave:
+            # Vendors are currently saved as strings in our db.
+            # We also need to add the gvl prefix here to turn this into
+            # a universal vendor id.
+            vendor_id: str = add_gvl_prefix(str(identifier))
```
adamsachs, line 226: yup, this assumption is good, i.e. all TC string specified vendors can be assumed to be GVL vendors 👍

adamsachs (commented):
@pattisdr this looks good! thank you for the quick turnaround 👏 

just one question along the peripheries with GVL/AC support and trying to keep things as "safe" as possible until we've got the more proper support in next release - let me know if you want to talk it over, or if i've misunderstood something!

src/fides/api/util/tcf/tc_model.py
pattisdr: thank you, I missed that nuance here in the requirements.

Do you think we should fail right now if there's an "ac." prefixed vendor at the moment or just skip it @adamsachs

pattisdr: I'm going to assume an "ac."-prefixed vendor should just be skipped here -

pattisdr: @allisonking @adamsachs included a test where there's a system with an AC id but I am excluding it from showing up in the vendor sections of the TC string.  However, its relevant purpose does show up, which I thought was in line with out we treat regular systems currently.  What do you all think?

See test here https://github.com/ethyca/fides/pull/4256/commits/7b86c153cafb9701aef72642d0589badfa802aa1

adamsachs: sorry for delay in response - i think what you landed on is good! basically it means we _could_ support an AC vendor if they happen to have a privacy declaration, for whatever reason. that paves the way nicely to the "real" support, which will allow it to show up even without a declaration (the other ticket, not being done here).

@allisonking does this work for you on the FE? i think you already have some AC logic you're building in, so just wanna make sure this won't break anything...

adamsachs (approved):
this looks good!

let's just wait to get a FE perspective from @allisonking on https://github.com/ethyca/fides/pull/4256#discussion_r1355767091 before calling it done 👍

src/fides/api/util/tcf/tc_model.py
allisonking: > @allisonking @adamsachs included a test where there's a system with an AC id but I am excluding it from showing up in the vendor sections of the TC string. However, its relevant purpose does show up, which I thought was in line with out we treat regular systems currently. What do you all think?

yes, I agree that makes sense! I'm not doing a whole lot specific to AC right now other than generating the string from IDs. any purposes that show up will be treated the same way we treat purposes for GVL/systems 👍

pattisdr: wonderful thank you both!

adamsachs:
merged https://github.com/ethyca/fides/pull/4228 into integration branch, then merged integration branch here and adjusted tests accordingly in https://github.com/ethyca/fides/pull/4256/commits/1b2d3f8854fa5e371da35846767b35d43052a104

i'll let CI run - if it's green, i think we're good to go on this?

PR #4262, opened 2023-10-12
1117 Support and implement additional configs through Fides.init()

eastandwestwind:
Closes https://github.com/ethyca/fidesplus/issues/1117

### Description Of Changes

Adds support for new config options, using Fides.init(), for exaxmple:

```
window.Fides.init({
    consent: {
      options: [{
         fidesEmbed: true,
         fidesDisableSaveApi: true,
         fidesTcString: "",
      }]
    },
});
```

**Details:**
1. Supports a `fides_embed` boolean config option. If this option is true, we "embed" the fides.js overlay UI (ie. “Layer 2”) into a web page instead of as a pop-up overlay, and never render the banner (ie. “Layer 1”).  That page will be loaded in either the website or via a native app webview for managing privacy preferences instead of in a pop-up overlay. Use `<div id="fides-embed-container"></div>` as the parent DOM node for the embedded overlay.
2. Supports a `fides_disable_save_api` boolean config option. If this option is true, we should disable the save to API, but still trigger the FidesUpdated event and save preferences to the cookie.
3. Supports a `fides_tc_string` config option. If it's set, `fides_tc_string` should supercede the cookie and any experience preferences, whether pre-fetched or client-side fetched.

### Code Changes

* [ ] Adds all new config options to init()
* [ ] Implements all new config options according to the above acceptance criteria
* [ ] Write full-coverage e2e tests to test how user preferences take precedence from the cookie, experience (both client-side fetched and pre-fetched), and explicit fides_tc_string

### Steps to Confirm

* [ ] Run e2e tests to confirm: `consent-banner-tcf.cy.ts` and `consent-banner.cy.ts`

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`


clients/fides-js/src/fides-tcf.ts
```diff
@@ -80,29 +103,110 @@ declare global {
...
+    purposesLegint: [],
+    specialPurposes: [],
+    features: [],
+    specialFeatures: [],
+    vendorsConsent: [],
+    vendorsLegint: [],
+  };
+
+  TCF_KEY_MAP.forEach(({ experienceKey, cookieKey, enabledIdsKey }) => {
+    if (experienceKey) {
```
allisonking, line 145: `experienceKey` always exists so I don't think you need the `if` here

clients/fides-js/src/fides-tcf.ts
```diff
@@ -80,29 +103,110 @@ declare global {
...
+      tcSavePrefs[cookieKey] = [];
+      experience[experienceKey]?.forEach((record) => {
+        const pref: UserConsentPreference = getInitialPreference(record);
+        // map experience to tcSavePrefs (same as cookie keys)
+        tcSavePrefs[cookieKey]?.push({
+          // @ts-ignore
+          id: record.id,
+          preference: getInitialPreference(record),
```
allisonking, line 153: I think you can use `pref` defined above here, right?

clients/fides-js/src/lib/tcf/utils.ts
```diff
@@ -0,0 +1,37 @@
+import { PrivacyExperience } from "../consent-types";
+import { EnabledIds, TcfCookieConsent, TcfCookieKeyConsent } from "./types";
+import { TCF_KEY_MAP } from "./constants";
+import { generateTcString } from "../tcf";
+
+export const generateTcStringFromCookieTcfConsent = async (
+  experience: PrivacyExperience,
+  tcfConsent: TcfCookieConsent
+): Promise<string> => {
+  const enabledIds: any | EnabledIds = {
```
allisonking, line 10: I don't think you need `any` here either?

clients/fides-js/src/lib/tcf/utils.ts
```diff
@@ -0,0 +1,37 @@
...
+    purposesLegint: [],
+    specialPurposes: [],
+    features: [],
+    specialFeatures: [],
+    vendorsConsent: [],
+    vendorsLegint: [],
+  };
+
+  TCF_KEY_MAP.forEach(({ cookieKey, enabledIdsKey }) => {
+    if (cookieKey) {
```
allisonking, line 21: `cookieKey` is always defined

clients/fides-js/src/fides-tcf.ts
```diff
@@ -80,29 +103,110 @@ declare global {
...
+  };
+
+  TCF_KEY_MAP.forEach(({ experienceKey, cookieKey, enabledIdsKey }) => {
+    if (experienceKey) {
+      tcSavePrefs[cookieKey] = [];
+      experience[experienceKey]?.forEach((record) => {
+        const pref: UserConsentPreference = getInitialPreference(record);
+        // map experience to tcSavePrefs (same as cookie keys)
+        tcSavePrefs[cookieKey]?.push({
+          // @ts-ignore
```
allisonking, line 151: hmmm this part of the code starts to have a bit of overlap with some of the initialization logic that `TcfOverlay` does. we might want to look at it as a whole and extract some common functions out at some point, but we don't have to do that now

allisonking (commented):
whew, nice work @eastandwestwind , lots of new features here! left some comments, but I don't think any of them are that big!

clients/fides-js/src/components/Overlay.tsx
```diff
@@ -55,6 +55,7 @@ const Overlay: FunctionComponent<Props> = ({
   const { instance, attributes } = useA11yDialog({
     id: "fides-modal",
     role: "alertdialog",
+    clazz: options.fidesEmbed ? "fides-embed" : "",
```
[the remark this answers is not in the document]
eastandwestwind: ahah not sure why I didn't think of that 😆

clients/fides-js/src/fides-tcf.ts
```diff
@@ -80,29 +103,110 @@ declare global {
 // eslint-disable-next-line no-underscore-dangle,@typescript-eslint/naming-convention
 let _Fides: Fides;
 
+/** Helper function to determine the initial value of a TCF object */
+const getInitialPreference = (
+  tcfObject:
+    | TCFFeatureRecord
+    | TCFVendorConsentRecord
+    | TCFVendorLegitimateInterestsRecord
```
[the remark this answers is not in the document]
eastandwestwind: `TcfModels` defines a `[]` of each record, so I've instead created a new type that echoes `TcfModels`: `TcfModelsRecord`

clients/fides-js/src/lib/consent.ts
```diff
@@ -28,18 +28,29 @@ export const initOverlay = async ({
...
-          `Parent element not found (#${overlayParentId}), creating and appending to body...`
-        );
-        // Create our own parent element and append to body
-        parentElem = document.createElement("div");
-        parentElem.id = overlayParentId;
-        document.body.prepend(parentElem);
+      let parentElem;
+      if (options.fidesEmbed) {
+        // Embed mode requires an existing element by which to embed the consent overlay
+        parentElem = document.getElementById("fides-embed-container");
```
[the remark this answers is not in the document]
eastandwestwind: eh, good idea regardless. I'll udpate!

clients/fides-js/src/fides-tcf.ts
eastandwestwind: This was more for a possible scenario in the future where the `TCF_KEY_MAP` would contain an item that exists in the cookie keys but we don't track in an experience. Is this something we should anticipate you think?

clients/fides-js/src/fides-tcf.ts
eastandwestwind: oh wow you're right 😅

clients/fides-js/src/lib/tcf/utils.ts
eastandwestwind: same Q as above- https://github.com/ethyca/fides/pull/4262#discussion_r1358129434 where we'd possibly have items within the `TCF_KEY_MAP` in the future that don't contain a `cookieKey`

clients/fides-js/src/components/tcf/TcfOverlay.tsx
```diff
@@ -82,6 +86,32 @@ const getEnabledIds = (modelList: TcfModels) => {
     .map((model) => `${model.id}`);
 };
 
+export const transformTcStringToCookieKeys = (
```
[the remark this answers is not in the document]
eastandwestwind: yes good call!

clients/fides-js/src/components/Overlay.tsx
```diff
@@ -75,6 +76,20 @@ const Overlay: FunctionComponent<Props> = ({
     }
   }, [instance, dispatchCloseEvent]);
 
+  const fidesEmbed = useMemo(
+    () =>
+      experience.show_banner &&
+      hasActionNeededNotices(experience) &&
```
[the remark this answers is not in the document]
eastandwestwind: Good call, I think you're right. All we'll need for embed mode is a valid experience with notices, but they don't have to "have action needed". And we can remove the `show_banner` check

clients/fides-js/src/fides-tcf.ts
eastandwestwind: yeah totally. It's tricky bc for `fides-tcf.ts` we don't have `enabledIds` upfront like we do for `TcfOverlay`. So we'd have to loop through the `experience` and generate those first. 

But we actually do use a core shared method down the line `transformTcfPreferencesToCookieKeys` which is something...

clients/fides-js/src/fides-tcf.ts
```diff
@@ -80,28 +103,100 @@ declare global {
...
+    );
+    debugLog(
+      config.options.debug,
+      "tc_string was missing from cookie, so it has been generated based on tcf_consent",
+      cookie.tc_string
+    );
+  }
   const initialFides = getInitialFides({ ...config, cookie });
   // Initialize the CMP API early so that listeners are established
   initializeCmpApi();
```
eastandwestwind, line 202: @allisonking after rebasing with your work, `initializeCmpApi();` ended up down here, but I'm guessing we shouldn't put it above `getInitialFides()`?

clients/fides-js/src/components/tcf/TcfOverlay.tsx
```diff
@@ -192,7 +192,7 @@ const updateCookie = async (
   });
   return {
     ...oldCookie,
-    fides_tc_string: tcString,
+    fides_string: tcString,
```
allisonking, line 195: oh nice, you're getting ahead of https://github.com/ethyca/fides/issues/4265!

clients/fides-js/src/fides-tcf.ts
allisonking: hmm hard to say! I hope not... I'd prefer to keep things matching with how they are now rather than building towards a potential future, but in this case it's not a huge thing

clients/fides-js/src/fides-tcf.ts
```diff
@@ -80,28 +103,100 @@ declare global {
...
+  return { ...oldCookie, fides_string: fidesString, tcf_consent: tcfConsent };
 };
 
 /**
  * Initialize the global Fides object with the given configuration values
  */
 const init = async (config: FidesConfig) => {
   const cookie = getInitialCookie(config);
+  if (config.options.fidesString) {
+    // If a TC str is explicitly passed in, we override the associated cookie props, which are then used to
```
allisonking, line 171:
```suggestion
    // If a fidesString is explicitly passed in, we override the associated cookie props, which are then used to
```

clients/fides-js/src/fides-tcf.ts
allisonking: hm yes I think what you have is right

clients/fides-js/src/lib/consent.ts
```diff
@@ -5,6 +5,9 @@ import { debugLog } from "./consent-utils";
 
 import { OverlayProps } from "../components/types";
 
+const FIDES_EMBED_CONTAINER_ID = "fides-embed-container";
+const FIDES_OVERLAY_DEFAULT_ID = "fides-overlay";
```
allisonking, line 9: 🙌

clients/fides-js/src/lib/cookie.ts
```diff
@@ -302,11 +283,43 @@ export const updateExperienceFromCookieConsent = ({
               Boolean(cookieConsent[item.id]),
               ConsentMechanism.OPT_IN
             )
-          : undefined;
+          : // if experience contains a tcf entity not defined by tcfEntities, we override experience current pref with the default pref
+            item.default_preference;
```
allisonking, line 287: I know we talked about this live, but it worries me a little. it's okay right now since all of our default preferences are "opt out", but if you imagine that one became "opt in", this might pose a problem. for example, if Vendor A was default "opt in", but not included in the `fidesString`, this means either:
1. the user opted out of Vendor A (since `fidesString` only saves opt ins)
2. the user never saw Vendor A
this poses a problem in case 1️⃣ because the user may have actually opted out, but now, since we don't see it in the string, we opt them in

again, this won't happen right now because everyone's default preference is "opt out" but I feel like it could bite us one day without us realizing it

clients/fides-js/src/lib/cookie.ts
```diff
@@ -302,11 +283,43 @@ export const updateExperienceFromCookieConsent = ({
...
+      ? transformConsentToFidesUserPreference(
+          Boolean(cookie.consent[notice.notice_key]),
+          notice.consent_mechanism
+        )
+      : undefined;
+    return { ...notice, current_preference: preference };
+  });
+
+  // Handle the TCF case, which has many keys to query
+  const tcfEntities = buildTcfEntitiesFromCookie(experience, cookie);
```
allisonking, line 322: nice refactoring of this function 👍

clients/fides-js/src/lib/initialize.ts
```diff
@@ -264,13 +272,35 @@ export const initialize = async ({
...
+            cookie
+          );
+          Object.assign(effectiveExperience, tcfEntities);
+        }
+      } else {
+        const updatedCookie = await updateCookie(
+          cookie,
+          effectiveExperience,
+          options.debug,
+          Boolean(options.fidesString)
```
allisonking, line 295: this line isn't needed anymore is it? `updateCookie` can always run, no need to check if there's an override prop? as in, can we remove the override prop?

allisonking (approved):
nice job @eastandwestwind !! I think it's looking good—I added a discussion question which might be worth thinking about, but it doesn't necessarily block this PR. 

also, I think this PR also closes https://github.com/ethyca/fides/issues/4265 but can you confirm? if it does, go ahead and assign it to yourself :)

clients/fides-js/src/lib/tcf/utils.ts
```diff
@@ -0,0 +1,64 @@
...
+import { EnabledIds, TcfCookieConsent, TcfCookieKeyConsent } from "./types";
+import { TCF_KEY_MAP } from "./constants";
+import { generateFidesString } from "../tcf";
+import { debugLog } from "../consent-utils";
+
+export const transformFidesStringToCookieKeys = (
+  fidesString: string,
+  debug: boolean
+): TcfCookieConsent => {
+  // Defer: to fully support AC string, we need to split out TC from AC string https://github.com/ethyca/fides/issues/4265
```
eastandwestwind, line 12: @allisonking I went ahead and updated more usages of `tcString` to `fidesString` to complete most of the work needed here- https://github.com/ethyca/fides/issues/4265, but there's still the question of splitting out the TC portion from the `fidesString`, along with adding tests and some validation checks to fully support AC str here. 

Do we have a separate ticket for functionality that's separate from just naming?

allisonking: I think that can be this one? https://github.com/ethyca/fides/issues/4263 can you add a comment to that ticket if you think there's stuff not captured there?

clients/fides-js/src/lib/cookie.ts
allisonking: okay, thanks for looking into that!

clients/fides-js/src/lib/tcf/utils.ts
eastandwestwind: oh perfect, thanks!

PR #4266, opened 2023-10-12
Surface AC Systems in Vendor Consents section of the TCF Overlay

pattisdr:
Closes https://github.com/ethyca/fidesplus/issues/1155

### Description Of Changes

AC systems should show up in the TCF overlay and need vendor consent toggles.  They are not required to have privacy declarations though, which currently drives what is in the TCF overlay.

### Code Changes

- Surface AC Systems (systems whose vendor_id starts with an "ac.*" prefix) in the TCF overlay under the Vendor Consents section, even if it has no privacy declarations.

- Separately update the "relevant systems" when vendor consent preferences are saved to include ac systems in the lookup 

### Steps to Confirm

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`


src/fides/api/util/tcf/tcf_experience_contents.py
```diff
@@ -422,9 +438,47 @@ def build_purpose_or_feature_section_and_update_vendor_map(
                 privacy_declaration_row=privacy_declaration_row,
             )
 
+        # Separately, add AC Vendor to Vendor Consent Map if applicable
+        add_ac_vendor_to_vendor_consent_map(
+            vendor_map=vendor_map,
+            tcf_vendor_component_type=tcf_vendor_component_type,
+            privacy_declaration_row=privacy_declaration_row,
+        )
```
pattisdr, line 446: This adds AC vendors to the Vendor Consents section if they weren't there already. What it *doesn't do* is prevent this from showing up in other sections.  If the AC vendor had privacy declarations with certain data uses and appropriate legal bases it could show up elsewhere too.  It didn't seem like I needed to *prevent* that specifically, and this seems more theoretical than something we'd actually encounter

allisonking: yup I agree—that makes sense to me!

src/fides/api/util/tcf/tcf_experience_contents.py
```diff
@@ -740,36 +792,33 @@ def get_relevant_systems_for_tcf_attribute(  # pylint: disable=too-many-return-s
...
             tcf_value
             # type: ignore[arg-type]
         )
 
     if tcf_field == TCFComponentType.system_legitimate_interests.value:
         return systems_that_match_system_id(
             starting_privacy_declarations.filter(
-                PrivacyDeclaration.legal_basis_for_processing
-                == LegalBasisForProcessingEnum.LEGITIMATE_INTEREST
+                LEGITIMATE_INTEREST_LEGAL_BASIS_FILTER
```
allisonking, line 821: declaring the filter as a variable is really handy!

allisonking (commented):
this looks great—very clean and easy to follow, well done! will test it out with the FE now to see if things Just Work and report back

allisonking:
ah I guess this will also need the prefix change before I can test E2E with local compass

pattisdr:
I'll start on that now!

src/fides/api/util/tcf/tcf_experience_contents.py
```diff
@@ -196,15 +210,20 @@ def get_matching_privacy_declarations(db: Session) -> Query:
...
+        .outerjoin(PrivacyDeclaration, System.id == PrivacyDeclaration.system_id)
         .filter(
-            PrivacyDeclaration.data_use.in_(ALL_GVL_DATA_USES),
-            PrivacyDeclaration.legal_basis_for_processing.in_(
-                [
-                    LegalBasisForProcessingEnum.CONSENT,
-                    LegalBasisForProcessingEnum.LEGITIMATE_INTEREST,
-                ]
-            ),
+            or_(
```
adamsachs, line 215: nice, this is pretty straightforward to follow!

src/fides/api/util/tcf/tcf_experience_contents.py
adamsachs: agreed, seems like the right way to handle things, at least for now. we'll see if we need to circle back if people start adding declarations to these systems and it hits unexpected edge cases here 👍

src/fides/api/util/tcf/tcf_experience_contents.py
```diff
@@ -422,9 +441,47 @@ def build_purpose_or_feature_section_and_update_vendor_map(
...
+    vendor_id, system_identifier = get_system_identifiers(privacy_declaration_row)
+
+    if not tcf_vendor_component_type == TCFVendorConsentRecord:
+        return
+
+    if not system_identifier.startswith(AC_PREFIX):
+        return
+
+    if system_identifier in vendor_map:
+        return
```
adamsachs, line 473: out of curiosity - do you think this reads easier as 3 separate conditionals rather than joining them in a single conditional with `or`s? just caught my eye since it's not the way i would have written it, but i'm guessing this was deliberate so i'm curious to hear! no need to adjust obviously.

src/fides/api/util/tcf/tcf_experience_contents.py
```diff
@@ -422,9 +441,47 @@ def build_purpose_or_feature_section_and_update_vendor_map(
...
+    vendor_map: Dict[str, VendorRecord],
+    tcf_vendor_component_type: VendorSectionType,
+    privacy_declaration_row: Row,
+) -> None:
+    """
+    Add systems with ac.*-prefixed vendor records to the Vendor Consent section if applicable.
+
+    FE shows Consent toggle only for AC vendors, and they are not required to have Privacy Declarations
+    """
+    vendor_id, system_identifier = get_system_identifiers(privacy_declaration_row)
```
adamsachs, line 464: nit: do we even want to support `system_id` specifying an AC vendor? not that it's causing any real harm, and it could potentially leave things more flexible. but i do feel like the flexibility and ambiguity here could potentially cause confusion going forward, and i'm inclined to just be a little bit more restrictive and say `vendor_id` is what we care about, and that's that -- at least within the scope of AC support.

more broadly, i know we'd kept system_id/vendor_id open ended in the past for good reason - maybe there are some other points in the overlay generation where we could start to hone in on the vendor_Id and remove use of the system_id. that's not within scope of this PR, and if you think that's not a good idea then no need to follow up on it. 

the point of this comment was regarding the AC use case specifically and i just thought it was worth mentioning that broader thought, without doing much to check whether it makes sense at this point! 😄

tests/fixtures/application_fixtures.py
```diff
@@ -2831,7 +2831,7 @@ def tcf_system(db: Session) -> System:
 
 
 @pytest.fixture(scope="function")
-def ac_system(db: Session) -> System:
+def ac_system_with_privacy_declaration(db: Session) -> System:
     """Test AC System - will be fleshed out further later"""
```
adamsachs, line 2835: nit: may be time to adjust this docstring now for clarity?

adamsachs, line 2835: nit: may be time to adjust this docstring now for clarity?

tests/ops/util/test_tcf_experience_contents.py
```diff
@@ -942,3 +942,39 @@ def test_add_same_data_use_to_different_systems(
             .id
             == 4
         )
+
+    @pytest.mark.usefixtures("ac_system_without_privacy_declaration")
+    def test_ac_systems_without_privacy_declarations(self, db):
```
adamsachs, line 947: nice - this is good coverage.

a nit, but would it be worth covering the case where the AC vendor does have a "valid GVL" data use declaration (i.e. your `ac_system_with_privacy_declaration` fixture), even though we don't expect that to be common? just to make sure the expected overlay content is returned then. would also be a nice way to demonstrate/document what the expected overlay behavior is when that edge case (inevitably) rears its head in the wild 😄

adamsachs (approved):
the changes look great here @pattisdr, this is a lot less painful than i expected, which owes a lot to your clean implementation! just had some minor nits, none of which i'd consider blocking - but would be curious to get your thoughts on them.

i didn't get a chance to smoke test, but if you've done some manual testing, then i think that's sufficient since the changes are straightforward enough. i can circle back and do some testing though if you'd like that extra confirmation!

src/fides/api/util/tcf/tcf_experience_contents.py
pattisdr: I do think it's easier to parse this way 😎

src/fides/api/util/tcf/tcf_experience_contents.py
pattisdr: that's a good callout.  I'll restrict so we're ignoring if there's no vendor id.

tests/ops/util/test_tcf_experience_contents.py
pattisdr: This is tested end-to-end in the `test_ac_system_not_in_tc_string`!   This test builds the TCF contents, and then builds the TC string from that.  There's an AC system and I am showing that while the vendor doesn't show up in the Vendor Consents section of the TC string, because it has valid GVL data uses, it's Purpose shows up in the purpose consents section.

pattisdr: It was not clear though! I will add a separate section to this file where I'm asserting the breakdown of the TCF experience specifically.

pattisdr:
Thanks for the review Adam - I always do manual testing!  I'll go ahead and merge and get the follow-up PR up that is building AC strings on the backend, building off of this work -

Issue #4257, opened 2023-10-11
Fix issue with stack accordion and single purposes


allisonking:
### Bug Description

In the first layer, there are cases where you will not display purposes: when a purpose stands alone without a stack or in the case of stack 1, where there are special features but no purposes. In this case, we should render this better

#### Steps to Reproduce

Set up TCF so that stack 1 shows up, and so that there are some single purposes. Observe the accordion.

### Expected behavior

Design?

### Screenshots

_If applicable, add screenshots to help explain your problem._

### Environment

* Version:
* OS:
* Python Version:
* Docker Version:

### Additional context

https://ethyca.atlassian.net/browse/PROD-1133?atlOrigin=eyJpIjoiNTVmZGY2MjMwOGRkNDgxN2FmZjI0NTc3OGNiYzk2NGIiLCJwIjoiaiJ9

PR #4278, opened 2023-10-13
Fix accordion stack purposes

allisonking:
Closes https://github.com/ethyca/fides/issues/4257

### Description Of Changes

Prevents a stack that only has special features from showing up with empty purposes.

| Before | After |
|--------|--------|
| image  | image |

### Code Changes

* [x] it was easy, just needed to make sure the `purposes` had a length
* [x] while I was here, I also sorted the purposes (see Dawn's catch here: https://ethyca.slack.com/archives/C05E74TR0BB/p1697215553763889)

### Steps to Confirm

* Go to the admin ui and for a data use that shows up in the TCF overlay, add two special features: `Use precise geolocation data` and `Actively scan device characteristics for identification`
* These will combine in the banner stack to "Precise geolocation data, and identification through device scanning". Clicking on it should not show an empty "Purposes" block

### Pre-Merge Checklist

* [x] Update `CHANGELOG.md`


clients/fides-js/src/lib/tcf.ts
```diff
@@ -47,7 +47,7 @@ const generateAcString = ({
         .map((id) => decodeVendorId(id).id)
     )
   );
-  const vendorIds = uniqueIds.sort().join(".");
+  const vendorIds = uniqueIds.sort((a, b) => Number(a) - Number(b)).join(".");
```
allisonking, line 50: it turns out JS sorts alphabetically by default, even for numbers. this means `.sort()` doesn't always work, i.e. it'll put `10` before `9`! but if we explicitly give it a sort function, it'll figure itself out

galvana:
Looks good to me
toggle

galvana (approved):
Tested locally and it works as expected!

allisonking:
Failing test is a known issue: https://github.com/ethyca/fides/issues/4288

Issue #4255, opened 2023-10-11
TCF overlay FE: support new TCF related fields on System model 


adamsachs:
### Is your feature request related to a specific problem?

TCF overlay should respect these coming from the BE, rather than relying on raw GVL as it does now.

needs https://github.com/ethyca/fides/issues/4227 to be done - https://github.com/ethyca/fides/pull/4228 will have added these properties to "vendor" records in the overlay payload

### Describe the solution you'd like

can get a sample payload from the BE once that's been implemented!

### Describe alternatives you've considered, if any

_A description of any alternative solutions or features you've considered._

### Additional context

_Add any other context or screenshots about the feature request here._

PR #4286, opened 2023-10-16
Render new TCF fields in the overlay using the data map

allisonking:
Closes https://github.com/ethyca/fides/issues/4255

### Description Of Changes

Renders the following TCF fields from data from the data map as opposed to querying gvl.json directly:
* cookie storage disclosure copy
* retention periods for purposes in a vendor
* privacy policy/legitimate interest disclosure urls

Not all of these fields were fully passed over from the backend, so there are some backend changes in this PR as well.

### Code Changes

* [x] Update TS types
* [x] Simplify some logic now that vendor relationships contain _all_ vendors
* [x] Derive storage disclosure info from `TCFVendorRelationships`
* [x] Add `privacy_policy_url` to `TCFVendorRelationships` in the backend
* [x] Derive urls from `TCFVendorRelationships`
* [x] Add retention period to `EmbeddedPurposes` in vendor data
* [x] Remove no longer needed GVL types 
* [x] Cypress tests

### Steps to Confirm

* Set up TCF and on one of those systems, fill out the "Cookie properties" section in admin-ui for the system
* Spin up the TCF overlay and you should see these cookie properties reflected in the vendor tab, when you click on a vendor
image
* Edit the system's privacy policy and legitimate interest disclosure. This should reflect in the TCF overlay
image
* In one of the system's data uses, change its "Retention period" to a different number. You should see this reflect in the "Purposes" section in the vendor
image

### Pre-Merge Checklist

* [x] Update `CHANGELOG.md`


tests/fixtures/application_fixtures.py
```diff
@@ -2807,6 +2807,7 @@ def tcf_system(db: Session) -> System:
             "legal_basis_for_processing": "Consent",
             "egress": None,
             "ingress": None,
+            "retention_period": "3-5 days",
```
adamsachs, line 2810: nit: i know we want to accommodate "arbitrary" string values for this property, i'm just a bit dubious of putting this value in a fixture that's meant to be representative of a TCF/GVL system. the GVL systems should really only have integer-like values (e.g. `"397"`) on this field. i'm a bit torn though because i know we want to have test coverage for values like `"3-5 days"` since that's possible for non-GVL systems. i just get worried about establishing a fixture here that's specifically different from the data we're expecting upstream that it's meant to represent.

would it be too much to create another fixture perhaps that's a non-GVL system? it may not be bad to have test coverage for that generally on the BE...although to be fair, i'm not 100% sure it's gonna show up in the overlay at all _yet_ 😄

src/fides/api/schemas/tcf.py
```diff
@@ -29,7 +29,13 @@ class NonVendorSection(UserSpecificConsentDetails):
     systems: List[EmbeddedVendor] = []  # Systems that use this TCF attribute
 
 
-class TCFPurposeConsentRecord(NonVendorSection, MappedPurpose):
+class CommonPurposeFields(FidesSchema):
```
adamsachs, line 32: nice, i think this is a good abstraction! we may end up adding more fields to it as we go...

adamsachs (commented):
@allisonking backend is looking good from my POV - nice work! i can see now it wasn't _quite_ as straightforward as it seemed when talking it over. just had a small nit on the fixtures, though it's nuanced enough that i wouldn't blame you if you choose to ignore me 👍 

and i'd like @pattisdr to also confirm things look good, since i'm still pretty fresh in the overlay generation codepath!

pattisdr:
Reviewing now!

src/fides/api/util/tcf/tcf_experience_contents.py
```diff
@@ -388,6 +391,7 @@ def build_purpose_or_feature_section_and_update_vendor_map(
                 non_vendor_record_map=non_vendor_record_map,
                 is_purpose_type=is_purpose_section,
                 use_or_feature=attribute,
+                declaration=privacy_declaration_row,
```
pattisdr, line 394: An edge case - say a system has two data uses, both of which map to the same purpose, but with different retention periods. For example, a system has marketing.advertising.first_party.contextual with a retention period of 100 days and a separate declaration marketing.advertising.frequency_capping with a retention period of 90 days. Both map to Purpose 2. Only one retention period will be kept (I am ordering the previous query, so at least the retention period kept is repeatable).

src/fides/api/util/tcf/tcf_experience_contents.py
```diff
@@ -272,7 +275,7 @@ def _add_top_level_record_to_purpose_or_feature_section(
     # Transform the base gvl record into the TCF record type that has more elements for TCF display.
     # Will be a top-level section.
     top_level_tcf_record: NonVendorRecord = tcf_component_type(
-        **fideslang_gvl_record.dict()
+        **fideslang_gvl_record.dict(), retention_period=declaration.retention_period
```
pattisdr, line 278: If we're not displaying retention periods on the Purpose tab (?), I don't think we should add the retention period here, instead, adding them only to the embedded purposes under vendors, so their retention periods are allowed to vary.

---

This builds a top-level record like a Purpose record, which is also used later as the basis for the EmbeddedPurpose records under Vendors and Systems as well.  If we haven't seen the Purpose thus far, we create a new one, **otherwise we use the existing Purpose**. 

 This hasn't been an issue thus far because the purpose data we're displaying is supposed to be the same everywhere. My concern is that the contents of the Purpose records now can vary.  One System may have a retention period on Purpose 1 of a 100 days. Another System may have a retention period on Purpose 1 of 300 days.  The first data use encountered for a given purpose is the one whose retention period will be used everywhere.  



Even if the GVL enforces that purpose 1 has a retention period of 90 days or similar everywhere across every vendor (does it? I don't know), systems outside of the GVL could define whatever retention period they wish, which could be the value that is used everywhere.

pattisdr (commented):
Noting concerns about differences in retention periods on the same Purpose across Systems or within the same System -

clients/fides-js/src/components/tcf/TcfVendors.tsx
```diff
@@ -158,13 +143,13 @@ const DataCategories = ({
...
+const StorageDisclosure = ({ vendor }: { vendor: VendorRecord }) => {
   const {
     name,
-    usesCookies,
-    usesNonCookieAccess,
-    cookieMaxAgeSeconds,
-    cookieRefresh,
+    uses_cookies: usesCookies,
+    uses_non_cookie_access: usesNonCookieAccess,
+    cookie_max_age_seconds: cookieMaxAgeSeconds,
```
pattisdr, line 151: It's not clear to me in the admin ui that this cookie age is seconds - 
Screenshot 2023-10-17 at 12 27 41 PM

clients/fides-js/src/components/tcf/TcfVendors.tsx
```diff
@@ -158,13 +143,13 @@ const DataCategories = ({
...
+    cookie_refresh: cookieRefresh,
   } = vendor;
   let disclosure = "";
   if (usesCookies) {
```
pattisdr, line 155: Below some unexpected behavior if uses cookies is False, but refreshes cookies is True for example.   Maybe everything should be nested under `usesCookies`?

clients/fides-js/src/components/tcf/TcfVendors.tsx
allisonking: yeah I brought this up here but I don't feel like we really reached a conclusion: https://ethyca.slack.com/archives/C05E74TR0BB/p1697483221897889

clients/fides-js/src/components/tcf/TcfVendors.tsx
```diff
@@ -90,34 +75,34 @@ const PurposeVendorDetails = ({
...
-        }
+        lineItems={purposes}
+        // dataRetention={
+        //   dataRetention
+        //     ? {
+        //         mapping: dataRetention.purposes,
+        //         default: dataRetention.stdRetention,
+        //       }
+        //     : undefined
+        // }
```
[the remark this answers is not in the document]
allisonking: oops, meant to remove!

clients/fides-js/src/components/tcf/TcfVendors.tsx
allisonking: mm good point! I guess you could feasibly have non cookie access along with not using cookies though, so maybe I'll change the logic to...

```js
  if (usesCookies) {
    const days = cookieMaxAgeSeconds
      ? Math.ceil(cookieMaxAgeSeconds / 60 / 60 / 24)
      : 0;
    disclosure = `${name} stores cookies with a maximum duration of about ${days} Day(s).`;
    if (cookieRefresh) {
      disclosure = `${disclosure} These cookies may be refreshed.`;
    }
    if (usesNonCookieAccess) {
      disclosure = `${disclosure} This vendor also uses other methods like "local storage" to store and access information on your device.`;
    }
  } else if (usesNonCookieAccess) {
    disclosure = `${name} uses methods like "local storage" to store and access information on your device.`;
  }
```

clients/fides-js/src/components/tcf/TcfVendors.tsx
allisonking: I got the okay to put `Maximum duration (seconds)` here—I'll do that, but also there's another ticket to make this into a better input one day: https://github.com/ethyca/fides/issues/4293

src/fides/api/schemas/tcf.py
```diff
@@ -29,7 +29,13 @@ class NonVendorSection(UserSpecificConsentDetails):
...
+    """Fields shared between the purpose sections of the TCF Experience"""
+
+    retention_period: Optional[str]
```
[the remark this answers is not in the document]
allisonking: no you're right—they don't show up in the Purpose tab. I put them in Purpose because I thought that was the only way to get them into EmbeddedPurpose but there's probably another way...

clients/fides-js/src/components/tcf/TcfVendors.tsx
pattisdr: that looks good!

src/fides/api/schemas/tcf.py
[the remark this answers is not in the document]
pattisdr: OK great! let's update it downstream to just put them into EmbeddedPurpose - I think it's going to cause problems per some of my other comments.

src/fides/api/util/tcf/tcf_experience_contents.py
allisonking: yeah you're right! okay so I guess I only want `retention_period` on the embedded purpose. I was having trouble figuring out how to get that data over to the embedded purpose though. let me try again...

tests/fixtures/application_fixtures.py
allisonking: based on the convo with Rachel today, I think I'll change these to just a number since it sounds like that's pretty much what we'll expect for the future

src/fides/api/util/tcf/tcf_experience_contents.py
adamsachs: whoops, i totally missed this - @pattisdr you're right, this needs to go on the `EmbeddedPurpose`, the retention period applies to a given purpose/vendor combination, not globally to a purpose, and there's nothing (that i know of) indicating what the retention period should be for a given purpose generally. this is specified per-vendor, so let's make sure we represent that in our overlay too.

thanks for catching this, i completely glossed over it!

src/fides/api/util/tcf/tcf_experience_contents.py
adamsachs: i think this edge case that's only applicable to non-GVL vendors. which i know is a use case we're handling, so this is good to call out - but i think we can live with some "strange" behavior in the overlay (whatever we think it should be -- perhaps the higher value retention period should be chosen?).

this shouldn't be applicable to GVL vendors, since all of the data use declarations that are generated from/based on a common purpose should, necessarily, have the same retention period. we don't enforce that in fides at all, but in the conversion of GVL --> compass records, that's the logic.

@pattisdr @allisonking let me know if i'm misunderstanding, and also curious to hear what you think on the suggestion to use the higher value retention period on this edge case. thanks for calling it out @pattisdr!

allisonking: mm yeah you are both making sense to me :) I agree using the higher retention period makes sense, though I'm not sure how to do that... would I have to change [this order by](https://github.com/ethyca/fides/blob/aking/4255/tcf-fields/src/fides/api/util/tcf/tcf_experience_contents.py/#L212-L214) to be by retention period instead? or maybe I sort it here? https://github.com/ethyca/fides/blob/aking/4255/tcf-fields/src/fides/api/util/tcf/tcf_experience_contents.py#L390

pattisdr: interesting idea that first idea should work fine

allisonking: wait a minute, that's not going to work well if retention period is a string right...

adamsachs: if we think that change is too impactful/risky/cumbersome, i'm also not opposed to us just choosing a retention period arbitrarily, since this is such an edge case at the moment...

whatever we choose, we should probably just communicate the behavior to product, since we're kinda in uncharted territory here and making some (reasonable) assumptions/decisions!

pattisdr: ah right Allison! In that case, I'd leave as-is.   This edge case is very unlikely, I only added it as I was trying to make sure I'd thought through all the scenarios.  The important thing you've already addressed with moving to adding retention at the embedded purpose level

allisonking:
Thanks for all the assistance @pattisdr and @adamsachs ! I think this is ready for review now 👍

pattisdr:
Thanks Allison, starting review!

clients/fides-js/src/components/tcf/TcfVendors.tsx
```diff
@@ -158,26 +119,28 @@ const DataCategories = ({
...
-    usesNonCookieAccess,
-    cookieMaxAgeSeconds,
-    cookieRefresh,
+    uses_cookies: usesCookies,
+    uses_non_cookie_access: usesNonCookieAccess,
+    cookie_max_age_seconds: cookieMaxAgeSeconds,
+    cookie_refresh: cookieRefresh,
   } = vendor;
   let disclosure = "";
   if (usesCookies) {
```
pattisdr, line 131: Do we need to change this to `uses_cookies` or similar, the conditional isn't working as I'd expect.

This value is false, but cookie refresh is true, and I see this:
Screenshot 2023-10-18 at 5 02 04 PM
e,

src/fides/api/util/tcf/tcf_experience_contents.py
pattisdr: We can remove this now since we're no longer adding at the top level: `retention_period=declaration.retention_period`

src/fides/api/util/tcf/tcf_experience_contents.py
```diff
@@ -388,6 +404,7 @@ def build_purpose_or_feature_section_and_update_vendor_map(
                 non_vendor_record_map=non_vendor_record_map,
                 is_purpose_type=is_purpose_section,
                 use_or_feature=attribute,
+                declaration=privacy_declaration_row,
```
pattisdr, line 407: We can remove this now!

src/fides/api/util/tcf/tcf_experience_contents.py
```diff
@@ -253,6 +256,7 @@ def _add_top_level_record_to_purpose_or_feature_section(
     non_vendor_record_map: Dict[int, NonVendorRecord],
     is_purpose_type: bool,
     use_or_feature: str,
+    declaration: PrivacyDeclaration,
```
pattisdr, line 259: We can remove this!

src/fides/api/util/tcf/tcf_experience_contents.py
```diff
@@ -304,8 +310,18 @@ def _embed_purpose_or_feature_under_system(
...
+        system_section.append(
+            EmbeddedPurpose(  # type: ignore[arg-type]
+                id=embedded_tcf_record.id,
+                name=embedded_tcf_record.name,
+                retention_period=retention_period,
+            )
+        )
+    else:
+        # Nest new cloned feature record beneath system otherwise
+        system_section.append(embedded_tcf_record)
```
pattisdr, line 324: Great this is working as I'd expect now, the main thing we fixed is that if the same purpose occurred on multiple vendors, with different retention periods, both of those retention periods show up individually instead of just the first purpose's retention being used.

Can we add a test for this?   Two vendors with the same purpose, but the retention periods differ on both.

clients/admin-ui/src/types/api/models/TCFPurposeConsentRecord.ts
```diff
@@ -9,6 +9,7 @@ import type { UserConsentPreference } from "./UserConsentPreference";
  * Schema for a TCF Purpose with Consent Legal Basis returned in the TCF Overlay Experience
  */
 export type TCFPurposeConsentRecord = {
+  retention_period?: string;
```
pattisdr, line 12: Is this outdated?

clients/fides-js/src/lib/tcf/types.ts
```diff
@@ -31,8 +31,15 @@ export type EmbeddedVendor = {
...
 
+export type EmbeddedPurpose = {
+  id: number;
+  name: string;
+  retention_period?: string;
+};
+
 // Purposes
 export type TCFPurposeConsentRecord = {
+  retention_period?: string;
```
pattisdr, line 42: Do we need to regenerate some of these?

pattisdr (commented):
Nice work Allison, especially moving some of this from the GVL into being pulled from the datamap in the backend. Mostly minor cleanup and another test requested!

src/fides/api/util/tcf/tcf_experience_contents.py
allisonking: ah, yes!

clients/fides-js/src/components/tcf/TcfVendors.tsx
allisonking: I wasn't able to reproduce, but while looking into this I did notice that there was a gap if there are no cookies + no privacy/legint URLs, so I fixed that 
image

clients/fides-js/src/lib/tcf/types.ts
allisonking: yes, thank you!

src/fides/api/util/tcf/tcf_experience_contents.py
allisonking: done! https://github.com/ethyca/fides/pull/4286/commits/38be263dbb0491df54233d1e2941246235a08faa

clients/fides-js/src/components/tcf/TcfVendors.tsx
pattisdr: ah sorry false alarm.  Seeing better handling needed when retention period is null?
Screenshot 2023-10-19 at 12 45 08 PM

allisonking: ah, good catch!

pattisdr: Allison what I think I was doing yesterday was not always hitting save after those toggles 😫.  I need to spend more time in the UI.

allisonking: okay, fixed! https://github.com/ethyca/fides/pull/4286/commits/c49d5e5c172991a04d66b3c5acd706cdd69bf52d

image

pattisdr (approved):
wonderful job 🌞

allisonking:
thanks for the thorough review @pattisdr !

PR #4294, opened 2023-10-17
Add custom fields to system history

### LINKED ISSUE #4273: Add custom system fields to system history tracking

galvana:
Closes #4273

⚠️ Dependent on fidesplus, check out the branch in https://github.com/ethyca/fidesplus/pull/1170 to run the backend

### Description Of Changes

Adding support for `system:information` and `system:data_use` custom fields. Includes a few bug fixes and general code cleanup.

### Code Changes

* [ ] Fixed bug where an incomplete system was set as active after saving the system information form
* [ ] Added new `SystemCustomFieldGroup` component to dynamically render the custom field diffs
* [ ] Fixed typo in the `Data use` diff field
* [ ] Cleaned up `history/helpers.tsx` and added docstrings
* [ ] Fixed typo in `test_system_history.py`

### Steps to Confirm

* [x] Start fidesplus
* [x] Create three custom fields for the `system:information` location (open text, single select, and multi select)
* [x] Create three custom fields for the `system:data_use` location (open text, single select, and multi select)
* [x] Define the custom fields in the system information form for a system
* [x] Define the custom fields in a privacy declaration for the system
* [ ] Navigate to the system history tab, there should be six system history entries (3 for system info and 3 for data use)
* [ ] The open text field should just be text, the single select and multi select fields should show up as badges

### Pre-Merge Checklist

* [x] Update `CHANGELOG.md`


clients/admin-ui/src/features/system/SystemInformationForm.tsx
```diff
@@ -174,6 +177,7 @@ const SystemInformationForm = ({
         formikHelpers.resetForm({ values });
         onSuccess(systemBody);
         dispatch(setSuggestions("hiding"));
+        dispatch(setActiveSystem(result.data));
```
galvana, line 180: Setting the response as the active system instead of `systemBody` since only the `created_on` field will only be available on the system response, not the system form fields.

clients/admin-ui/src/features/system/history/helpers.tsx
```diff
@@ -120,6 +95,29 @@ export const describeSystemChange = (history: SystemHistoryResponse) => {
...
+
+  // if the history contains custom fields, it won't have the main fields
+  // so we can just process the custom_fields object
+  if (before.custom_fields || after.custom_fields) {
+    ({ addedFields, removedFields, changedFields } =
+      categorizeFieldModifications(before.custom_fields, after.custom_fields));
+  } else {
+    // process the main fields
+    ({ addedFields, removedFields, changedFields } =
+      categorizeFieldModifications(before, after));
```
galvana, line 118: Most of the changes in this file are just cleanup but this is the main part to focus on for the custom field diff.

clients/admin-ui/src/features/system/history/modal/SystemDataForm.tsx
```diff
@@ -171,8 +175,8 @@ const SystemDataForm: React.FC<SystemDataFormProps> = ({ initialValues }) => {
                       />
                       <SystemDataTextField
                         name={`privacy_declarations[${index}].data_use`}
-                        label="This system processes personal data"
-                        tooltip="Does this system process personal data?"
+                        label="Data use"
+                        tooltip="For which business purposes is this data used?"
```
galvana, line 179: Fixing copy/paste error

tests/ctl/core/test_system_history.py
```diff
@@ -109,4 +109,4 @@ async def test_multiple_changes(self, db, async_session_temp, system: System):
         ).all()
         assert len(system_histories) == 3
         for system_history in system_histories:
-            system_history.edited_by == CONFIG.security.root_username
+            assert system_history.edited_by == CONFIG.security.root_username
```
galvana, line 112: Fixing old typo

allisonking:
Testing this out now! leaving an early comment—is this expected? the fields don't quite point to the right places
image

clients/admin-ui/src/features/system/history/SystemHistoryTable.tsx
```diff
@@ -23,7 +23,10 @@ import { SystemHistoryResponse } from "~/types/api";
 import { SystemResponse } from "~/types/api/models/SystemResponse";
 
 import {
+  alignCustomFields,
```
allisonking, line 26: ah, CI says this function doesn't exist :)

clients/admin-ui/src/features/system/history/helpers.tsx
```diff
@@ -245,3 +266,96 @@ export const alignPrivacyDeclarations = (
...
+  ]);
+
+  allKeys.forEach((key) => {
+    if (!(key in beforeCustomFields)) {
+      beforeCustomFields[key] = null;
+    }
+    if (!(key in afterCustomFields)) {
+      afterCustomFields[key] = null;
+    }
+  });
```
allisonking, line 290: ah okay—I think this is the problem with the alignment. if I have a `beforeCustomFields` that has only one field, but my `afterCustomFields` has two fields...

```
before: {a: 'field'}
after: {b: 'field', a: 'field'}
```
then we do this `forEach` and that makes it...

```
before: {a: 'field', b: 'null}
after: {b: 'field', a: 'field'}
```

and then we get a misalignment. this is tricky—it appears JS objects [don't guarantee sort order](https://stackoverflow.com/questions/5525795/does-javascript-guarantee-object-property-order). so we may need to sort somewhere else, or create this object a little differently

allisonking: perhaps you can recreate the object entirely, i.e. get all the keys, then iterate and not just fill in nulls—fill in the real values too. something like... (this is untested!)

```javascript
const before = Object.fromEntries(allKeys.map(key => { 
  [key, history.before.customFields[key] ?? null]
}))
const after = Object.fromEntries(allKeys.map(key => { 
  [key, history.after.customFields[key] ?? null]
}))
```

clients/admin-ui/src/features/system/history/helpers.tsx
```diff
@@ -245,3 +266,96 @@ export const alignPrivacyDeclarations = (
...
+    history.after.privacy_declarations && history.after.privacy_declarations[0]
+      ? { ...history.after.privacy_declarations[0].custom_fields }
+      : {};
+
+  const allKeys = new Set([
+    ...Object.keys(beforeCustomFields),
+    ...Object.keys(afterCustomFields),
+  ]);
+
+  allKeys.forEach((key) => {
```
allisonking, line 331: I think we may have the same sort order problem here, though I didn't encounter it specifically

allisonking (commented):
I think I found a bug!

clients/admin-ui/src/features/system/history/SystemHistoryTable.tsx
galvana: Fixed!

clients/admin-ui/src/features/system/SystemInformationForm.tsx
galvana: That makes sense! I shuffled things around to what we had before but with the correct param for onSuccess

clients/admin-ui/src/features/system/history/modal/fields/SystemCustomFieldGroup.tsx
```diff
@@ -0,0 +1,60 @@
...
+      (value) => value.name === name && !!value.allow_list_id
+    );
+
+  const prefix =
+    resourceType == ResourceTypes.SYSTEM
+      ? "custom_fields"
+      : "privacy_declarations[0].custom_fields";
+
+  // to ensure the order in the diff lists is the same
+  const sortedFieldNames = Object.keys(customFields).sort();
```
galvana, line 37: This is the fix for the out-of-order fields, this will handle both the system info and privacy declaration fields.

Kelsey-Ethyca:
Testing notes- 
1. I have found a bug - 400 is thrown when updating a custom field if you have a multiselect custom field but do not update it then it is not written is an array https://www.loom.com/share/e5aa1abebfe04a58ba82624e97c31d17?sid=2d397d78-eff5-4684-9d0a-bcb1f011370e 
2. I need to look into the failing tests 
3. everything else looks to be working 
-creating a system and adding custom fields; displays properly in system info tab and history tab ✔️ 
-creating a data use and adding custom fields; displays properly in system info tab and history tab ✔️ 
-updating all system:info custom fields; displays properly in system info tab and history tab  ✔️ 
-updating only one system:info custom field; displays properly in system info tab and history tab ✔️ 
-updating all system:data use custom fields; displays properly in data use tab and history tab  ✔️ 
-updating only one system:system data use custom field; displays properly in data use tab and history tab ✖️ This is the bug listed above as #1

Kelsey-Ethyca (approved):
🦖

Issue #4263, opened 2023-10-12
Need to be able to take a custom composite `fides_tc_string` and parse out TC and AC strings


allisonking:
### Is your feature request related to a specific problem?

https://github.com/ethyca/fides/pull/4262 adds the ability to take in a custom `fides_tc_string` in order to overwrite initial settings. This string will actually be a composite string, comma separated into `{TC_STRING},{AC_STRING}` and so we need to be able to handle that.

### Describe the solution you'd like

In theory we already have the logic to initialize from `TC_STRING`, but we'll need to make sure to parse it first. For `AC_STRING`, we'll likely have to initialize our vendor (and maybe system?) fields anew.

Also, add validation for format of fides_string. We accept `TC`, `TC,` and `TC,AC`, but let's reject `,AC`.

### Describe alternatives you've considered, if any

_A description of any alternative solutions or features you've considered._

### Additional context

https://ethyca.atlassian.net/wiki/spaces/PM/pages/2784788587

eastandwestwind:
Let's think about: when a custom fides_tc_string is passed in, what are the possible states? `TC`, `TC,AC`? It won’t be possible to just have `AC` or `,AC` correct?

PR #4295, opened 2023-10-18
Build and Accept Backend AC Strings

pattisdr:
Closes https://github.com/ethyca/fidesplus/issues/1162

### Description Of Changes

Add proper AC string support from the backend, both in saving privacy preferences from a string, and in returning meta information when building developer-friendly privacy experiences.  [Behavior nuances documented here](https://ethyca.atlassian.net/wiki/spaces/EN/pages/2819358794/2023-10-20+Documented+Backend+AC+String+Behavior).  

❗ IMPORTANT: This adds a new AC Enabled setting!  This must be set to True for AC functionality to work.

### Code Changes

**Saving Privacy Preferences:**
- Allows preferences to be saved from the AC string portion of a Fides string if AC mode enabled
    - Allows a Fides string to include both a TC string and an AC string. New format is a TC string and an AC string joined by a comma:  `<tc_string>,<ac_string>`.
   - TC string and AC string are pulled from the string.  TC string preferences saved per usual.  Big change here is that we figure out Vendor consent preferences from the AC string and persist those as well
   - Response includes the AC string under the  IABTCF_AddtlConsent key in the mobile data

**Retrieving Privacy Experiences**
- Updates to building developer-friendly privacy experiences meta section:
    - When building accept all and reject all fides strings, return these in fides_string format, which includes the TC string and AC string in the same representation if applicable and AC mode is enabled
    - Accept all and reject all mobile data also has populated IABTCF_AddtlConsent sections if AC mode is enabled

**General**
- When building a TCModel from the TCF Experience, which has been a prerequisite for building a TC string, add a new `ac_vendor_consents` section to support building an AC string too.
- Stricter handling to prevent AC systems defined in an invalid way from showing up in the Experience
- New AC Enabled environment variable

### Steps to Confirm

Note these instructions do not require you to be running the dictionary

* [ ] Set both `tcf_enabled` AND `ac_enabled` to True in your toml file or use an env var
* [ ] Run nox -s dev -- shell
* [ ] Add a GVL system [following these instructions](https://ethyca.atlassian.net/wiki/spaces/EN/pages/2809724964/API+How-to+Create+GVL+System+Locally)
* [ ] Add an AC system locally [following these instructions](https://ethyca.atlassian.net/wiki/spaces/EN/pages/2816999425/API+How-to+Create+AC+System+Locally)

#### GET Experiences with AC String

* [ ] Fetch an EEA Privacy Experience, with `include_meta=true` query param
```curl
http://localhost:8080/api/v1/privacy-experience?show_disabled=true&region=fr&systems_applicable=false&include_gvl=false&include_meta=true&page=1&size=50
```
* From a previous PR, the AC system shows up under the Vendor Consents section of the experience
* Note the `meta` section of the response.  The `accept_all_fides_string` concatenates the TC string and the AC string. The TC string still goes under the `IABTCF_TCString` section and the `IABTCF_AddtlConsent` is now populated with the AC String.  Reject all string looks like this: `1~`.  This is also what the accept-all string looks like if there are no AC systems.
```json
 "meta": {
        "version_hash": "a2e85860c68b",
        "accept_all_fides_string": "CPz_s0APz_s0AGXABBENAWEAALAAAEOAAAAAAEAEACACAAAA.IAEAEAAA,1~100",
        "accept_all_fides_mobile_data": {
          "IABTCF_CmpSdkID": 407,
          "IABTCF_CmpSdkVersion": 1,
          "IABTCF_PolicyVersion": 4,
          "IABTCF_gdprApplies": 1,
          "IABTCF_PublisherCC": "AA",
          "IABTCF_PurposeOneTreatment": 0,
          "IABTCF_UseNonStandardTexts": 0,
          "IABTCF_TCString": "CPz_s0APz_s0AGXABBENAWEAALAAAEOAAAAAAEAEACACAAAA.IAEAEAAA",
          "IABTCF_VendorConsents": "00000001",
          "IABTCF_VendorLegitimateInterests": "00000001",
          "IABTCF_PurposeConsents": "101100000000000000000000",
          "IABTCF_PurposeLegitimateInterests": "010000111000000000000000",
          "IABTCF_SpecialFeaturesOptIns": "000000000000",
          "IABTCF_PublisherConsent": null,
          "IABTCF_PublisherLegitimateInterests": null,
          "IABTCF_PublisherCustomPurposesConsents": null,
          "IABTCF_PublisherCustomPurposesLegitimateInterests": null,
          "IABTCF_AddtlConsent": "1~100"
        },
        "reject_all_fides_string": "CPz_s0APz_s0AGXABBENAWEAAAAAAAAAAAAAAAAAAAAA.IAEAEAAA,1~",
        "reject_all_fides_mobile_data": {
          "IABTCF_CmpSdkID": 407,
          "IABTCF_CmpSdkVersion": 1,
          "IABTCF_PolicyVersion": 4,
          "IABTCF_gdprApplies": 1,
          "IABTCF_PublisherCC": "AA",
          "IABTCF_PurposeOneTreatment": 0,
          "IABTCF_UseNonStandardTexts": 0,
          "IABTCF_TCString": "CPz_s0APz_s0AGXABBENAWEAAAAAAAAAAAAAAAAAAAAA.IAEAEAAA",
          "IABTCF_VendorConsents": "",
          "IABTCF_VendorLegitimateInterests": "",
          "IABTCF_PurposeConsents": "000000000000000000000000",
          "IABTCF_PurposeLegitimateInterests": "000000000000000000000000",
          "IABTCF_SpecialFeaturesOptIns": "000000000000",
          "IABTCF_PublisherConsent": null,
          "IABTCF_PublisherLegitimateInterests": null,
          "IABTCF_PublisherCustomPurposesConsents": null,
          "IABTCF_PublisherCustomPurposesLegitimateInterests": null,
          "IABTCF_AddtlConsent": "1~"
        }
      }
    }
```

#### PATCH Preferences with AC strings

- [ ] Let's take the accept-all fides string from the previous request and pass that in to save preferences
- [ ] PATCH http://localhost:8080/api/v1/privacy-preferences
```json
{
  "browser_identity": {
    "fides_user_device_id": "d81b51c4-10b1-4fc5-b283-1bb5fe3659d5"
  },
  "fides_string": "CPz10AAPz10AAGXABBENAWEAALAAAEOAAAAAAEAEACACAAAA.IAEAEAAA,1~100"
}
```
- [ ] Both the gvl.8 and the gacp.100 vendors show up in the Vendor Consent section:

System `gacp.100` was pulled off of the AC string and saved as a Vendor Consent.  gvl.8 was saved per previous releases.

```json
"vendor_consent_preferences": [
    {
      "purpose_consent": null,
      "purpose_legitimate_interests": null,
      "special_purpose": null,
      "vendor_consent": "gvl.8",
      "vendor_legitimate_interests": null,
      "feature": null,
      "special_feature": null,
      "system_consent": null,
      "system_legitimate_interests": null,
      "id": "cur_05f2e200-caa3-4ada-8a51-1c9a8d9bbbf4",
      "preference": "opt_in",
      "privacy_notice_history": null,
      "privacy_preference_history_id": "pri_aa7910f5-75e6-46e4-8906-23785f33f693"
    },
    {
      "purpose_consent": null,
      "purpose_legitimate_interests": null,
      "special_purpose": null,
      "vendor_consent": "gacp.100",
      "vendor_legitimate_interests": null,
      "feature": null,
      "special_feature": null,
      "system_consent": null,
      "system_legitimate_interests": null,
      "id": "cur_1a7ab22b-832a-46bf-831a-4d0000e5b309",
      "preference": "opt_in",
      "privacy_notice_history": null,
      "privacy_preference_history_id": "pri_de1201be-154c-494d-9396-67c695f67dd7"
    }
  ]
```
- [ ] The AC string shows up in the fides mobile data in the response under IABTCF_AddtlConsent:
```json
"fides_mobile_data": {
    "IABTCF_CmpSdkID": 407,
    "IABTCF_CmpSdkVersion": 1,
    "IABTCF_PolicyVersion": 4,
    "IABTCF_gdprApplies": 1,
    "IABTCF_PublisherCC": "AA",
    "IABTCF_PurposeOneTreatment": 0,
    "IABTCF_UseNonStandardTexts": 0,
    "IABTCF_TCString": "CPz10AAPz10AAGXABBENAWEAALAAAEOAAAAAAEAEACACAAAA.IAEAEAAA",
    "IABTCF_VendorConsents": "00000001",
    "IABTCF_VendorLegitimateInterests": "00000001",
    "IABTCF_PurposeConsents": "101100000000000000000000",
    "IABTCF_PurposeLegitimateInterests": "010000111000000000000000",
    "IABTCF_SpecialFeaturesOptIns": "000000000000",
    "IABTCF_PublisherConsent": null,
    "IABTCF_PublisherLegitimateInterests": null,
    "IABTCF_PublisherCustomPurposesConsents": null,
    "IABTCF_PublisherCustomPurposesLegitimateInterests": null,
    "IABTCF_AddtlConsent": "1~100"
  }
```

- [ ] Passing in reject all string opts out of any AC vendors:
```bash
curl -X 'PATCH' \
  'http://localhost:8080/api/v1/privacy-preferences' \
  -d '{
  "browser_identity": {
    "fides_user_device_id": "d81b51c4-10b1-4fc5-b283-1bb5fe3659d5"
  },
  "fides_string": "CPz_s0APz_s0AGXABBENAWEAAAAAAAAAAAAAAAAAAAAA.IAEAEAAA,1~"

}'
```

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`


src/fides/api/util/tcf/experience_meta.py
```diff
@@ -79,8 +80,14 @@ def build_experience_tcf_meta(tcf_contents: TCFExperienceContents) -> Dict:
...
         version_hash=build_tcf_version_hash(tcf_contents),
-        accept_all_fides_string=accept_all_mobile_data.IABTCF_TCString,
-        reject_all_fides_string=reject_all_mobile_data.IABTCF_TCString,
+        accept_all_fides_string=build_fides_string(
+            accept_all_mobile_data.IABTCF_TCString,
+            accept_all_mobile_data.IABTCF_AddtlConsent,
+        ),
+        reject_all_fides_string=build_fides_string(
+            reject_all_mobile_data.IABTCF_TCString,
+            reject_all_mobile_data.IABTCF_AddtlConsent,
```
pattisdr, line 89: Because the AC string encodes opt-ins, for a reject all string I am returning null here.  Although there may be alternative formats desired, like "1~" but I thought null made sense.     Generally, I expect AC strings in this format "1~1.100.333", or none supplied at all.

src/fides/api/util/tcf/tc_model.py
```diff
@@ -188,6 +191,12 @@ class TCModel(FidesSchema):
 
     num_pub_restrictions: int = 0  # Hardcoded here for now
 
+    ac_vendor_consents: List[int] = Field(
+        default=[],
+        description="A list of Google's Additional Consent Vendors for which the customer has opted in. These "
+        "are consented Google Ad Tech Providers that are not registered with IAB",
+    )
```
pattisdr, line 198: The TC Model is an intermediate structure that I use to convert the TCF Experience contents into a TC string.  I'm doing some prerequisite steps to get the data in the expected format, plus adding a lot of validation, similar to what the frontend library does to build strings. 

To parallel the other constructs in this model, I'm now adding a list of AC vendors here as well when doing this translation that I will later use to build the AC string.

I left the name as `TCModel` even though it now has an AC string section because I wanted it to stay similarly named to what the FE is using - FidesModel also feels way too vague.

src/fides/api/util/tcf/tc_model.py
```diff
@@ -446,6 +455,7 @@ def convert_tcf_contents_to_tc_model(
         vendor_consents=vendor_consents if consented else [],
         vendor_legitimate_interests=vendor_legitimate_interests if consented else [],
         vendors_disclosed=_build_vendors_disclosed(tcf_contents),
+        ac_vendor_consents=build_ac_vendor_consents(tcf_contents, preference),
```
pattisdr, line 458: The AC Vendors from the vendor_consents section of the TCF Experience are added to the intermediate TC model.

src/fides/api/api/v1/endpoints/privacy_preference_endpoints.py
```diff
@@ -700,29 +704,39 @@ def validate_served_record(
...
+            decoded_tc_str_request_body: FidesStringFidesPreferences = (
+                decode_tc_string_to_preferences(tc_str, tcf_contents)
+            )
+            decoded_ac_str_request_body: FidesStringFidesPreferences = (
+                decode_ac_string_to_preferences(ac_str, tcf_contents)
+            )
+            # We combine Vendor Consent Preferences from the TC String and AC String if applicable
+            decoded_tc_str_request_body.vendor_consent_preferences = (
+                decoded_tc_str_request_body.vendor_consent_preferences
+                + decoded_ac_str_request_body.vendor_consent_preferences
```
pattisdr, line 727: When saving preferences, I assume a fides_string can be passed in and split the string accordingly. It's backwards compatible. You can pass in just a TC string.  A comma-combined TC String and AC string. Or just an AC string even, provided a comma comes first.  It will fail early however if either the TC string or AC string are in an unexpected format.

The TC string is decoded per usual.  The AC string translation is new here, and I assume all pieces of the AC string will be saved as "vendor consents" specifically.  I combine the vendor consent preferences from the AC string and TC string into one.

The I proceed per usual, saving preferences, just like I would if the user had supplied them directly, rather than via the string.

src/fides/api/util/tcf/ac_string.py
```diff
@@ -0,0 +1,193 @@
...
+    accept_all_ac_str: Optional[str],
+) -> List[TCFVendorSave]:
+    """Helper to convert an AC string into a list of corresponding vendor preferences, for saving preferences
+    from an AC string.
+
+    AC strings only encode opt-ins, so by itself, this doesn't give us explicit opt-outs. We compare the passed-in AC
+    String with the accept-all AC string built at runtime against our current datamap.
+
+    Vendors in both strings are saved as an opt-in.  AC vendors in the datamap, but not in the supplied AC string,
+    get persisted as an explicit opt-out.
```
pattisdr, line 147: Note I use the same strategy here when translating an AC string to fides preferences, as I did for TC strings.  AC strings only encode opt in, but I compare to the datamap directly to also be able to persist explicit opt-outs.

pattisdr:
@NevilleS matched this against the original AC, realized I overlooked `If Google AC Mode` is enabled. Do you want this to be a separate environment variable or can this just be derived from a combination of TCF being enabled + AC systems in the datamap?

pattisdr:
Ah! I need to make this change, it's allowed in my implementation:

https://github.com/ethyca/fides/issues/4263

> We can reject ,AC, that's invalid because the core TC string is always needed for a complete signal

src/fides/api/util/tcf/experience_meta.py
pattisdr: Updated after slack convo, going with `1~` now

src/fides/api/util/tcf/tc_mobile_data.py
```diff
@@ -59,6 +65,9 @@ def _build_binary_string(name: str, num_bits: int) -> str:
         IABTCF_SpecialFeaturesOptIns=_build_binary_string(
             "special_feature_optins", SPECIAL_FEATURE_BITS
         ),
+        IABTCF_AddtlConsent=build_ac_string(tc_model.ac_vendor_consents)
+        if CONFIG.consent.ac_enabled
+        else None,
```
pattisdr, line 70: This is where I exclude the AC strings from showing up in the mobile data if AC is not enabled

src/fides/api/util/tcf/tc_model.py
adamsachs: this makes sense. to play devil's advocate a bit, it could be something like `FidesStringModel` but even that feels vague. we probably need a better field name than `fides_string` (`fides_consent_string` ?!) but that ship has sailed and i'm not going to suggest we go back now.

anyway, i don't see any major issues having this in the `TCModel`, that doesn't feel too problematic!

src/fides/api/util/tcf/tcf_experience_contents.py
```diff
@@ -214,21 +229,28 @@ def get_matching_privacy_declarations(db: Session) -> Query:
...
+                    == LegalBasisForProcessingEnum.CONSENT,
+                ),
+                and_(
+                    GVL_DATA_USE_FILTER,
+                    PrivacyDeclaration.legal_basis_for_processing
+                    == LegalBasisForProcessingEnum.LEGITIMATE_INTEREST,
+                    NOT_AC_SYSTEM_FILTER,
                 ),
-                AC_SYSTEM_FILTER,
+                AC_SYSTEM_NO_PRIVACY_DECL_FILTER,
```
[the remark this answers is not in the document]
adamsachs: ok! i like this, i think it's a nice refinement on the previous iteration, given that we've got some more certainty around the fact that AC vendors can only be considered for consent and not legitimate interest 👍

src/fides/api/util/tcf/tcf_experience_contents.py
```diff
@@ -217,21 +232,28 @@ def get_matching_privacy_declarations(db: Session) -> Query:
...
             )
         )
         .order_by(
             PrivacyDeclaration.created_at.desc()
         )  # Order to get repeatable results when collapsing information
     )
+    if not CONFIG.consent.ac_enabled:
+        # If AC Mode is not enabled, exclude all Privacy Declarations
```
adamsachs, line 253:
```suggestion
        # If AC Mode is not enabled, exclude all Privacy Declarations tied to an AC-based System.
```

src/fides/config/consent_settings.py
```diff
@@ -9,6 +9,9 @@ class ConsentSettings(FidesSettings):
     tcf_enabled: bool = Field(
         default=False, description="Toggle whether TCF is enabled."
     )
+    ac_enabled: bool = Field(
+        default=False, description="Toggle whether Google AC Mode is enabled."
+    )
```
adamsachs, line 14: not a must-have - but is there a relatively straightforward way for us to fail fast if `ac_enabled` is set to `True` while `tcf_enabled` is set to `False`? i _think_ we could have a validator to do that?

just feels like it'd be nice to prevent that misconfiguration before anyone gets too far ahead of themselves. basically, enforcing in the application the assumption you mentioned earlier:
> Assumes TCF must be enabled as a prerequisite.

tests/ops/util/test_ac_string.py
```diff
@@ -0,0 +1,259 @@
...
+    assert ac_str == ""
+
+    # Bad AC format
+    with pytest.raises(DecodeFidesStringError):
+        split_fides_string(
+            "CPz1hddPz1hddDxAAAENCZCgADgAAAAAAAAAAEBcABioAAA.YAAAAAAAAAA,100.1000"
+        )
+
+
+class TestBuildACString:
```
adamsachs, line 110: nice tests here!

tests/ops/util/test_ac_string.py
```diff
@@ -0,0 +1,259 @@
...
+                TCFVendorConsentRecord(id="gacp.8844"),
+            ]
+        )
+        ac_vendor_consents = build_ac_vendor_consents(
+            contents, UserConsentPreference.opt_in
+        )
+        assert build_ac_string(ac_vendor_consents) == "1~231.1313.8844"
+
+
+class TestDecodeACStringToPreferences:
```
adamsachs, line 182: also nice tests here - could maybe use a case with multiple AC vendors specified in the AC string, just to get coverage on the multi-value parsing? (i.e. that the `.` delimiter is well handled). i know that's already implicitly covered above, but couldn't hurt to get a bit more coverage here specifically on this `decode_ac_string_to_preferences` function!

adamsachs (approved):
looks great @pattisdr, great work! really just some nits, but i think the functionality is just about good to go from my POV and it's also an easy-to-follow implementation!

src/fides/config/consent_settings.py
pattisdr: ah a validator is a good idea!

pattisdr:
🙏 Thanks so much for your review here @adamsachs, it's my bigger one and unblocks several other pieces.  I'll get this updated and ready to go!

src/fides/api/util/tcf/ac_string.py
```diff
@@ -0,0 +1,193 @@
...
+def build_ac_string(ac_vendor_consents: List[int]) -> Optional[str]:
+    """Returns an AC string following this spec: https://support.google.com/admanager/answer/9681920"""
+    return (
+        SPECIFICATION_VERSION_NUMBER
+        + SEPARATOR_SYMBOL
+        + ".".join([str(vendor_id) for vendor_id in sorted(ac_vendor_consents)])
+    )
+
+
+def build_fides_string(tc_str: Optional[str], ac_str: Optional[str]) -> str:
```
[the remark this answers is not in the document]
pattisdr: good call! I agree I should split this out.

src/fides/api/util/tcf/ac_string.py
```diff
@@ -0,0 +1,193 @@
...
+
+def build_ac_vendor_consents(
+    tcf_contents: TCFExperienceContents, preference: Optional[UserConsentPreference]
+) -> List[int]:
+    """Build a list of integers representing AC vendors with opt-in consent
+
+    This is a prerequisite step for building an accept-all AC string powered by our datamap.
+    """
+    if not tcf_contents.tcf_vendor_consents:
+        # AC Vendors are automatically added to the Vendor Consents section of the TCF Experience if present
```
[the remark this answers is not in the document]
pattisdr: Thank you! I'll get rid of this section altogether and just rely on looping through the `tcf_contents.tcf_vendor_consents` list.  I've gotten into a habit of spelling out the separate exit conditions but I don't really need this one.

src/fides/api/util/tcf/experience_meta.py
```diff
@@ -67,6 +69,9 @@ def build_tcf_version_hash(tcf_contents: TCFExperienceContents) -> str:
 
 def build_experience_tcf_meta(tcf_contents: TCFExperienceContents) -> Dict:
     """Build TCF Meta information to supplement a TCF Privacy Experience at runtime"""
+    if not CONFIG.consent.tcf_enabled:
+        return ExperienceMeta().dict()
```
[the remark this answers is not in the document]
pattisdr: I'm thinking I should pull this out altogether and rely on existing locations upstream to handle if TCF is enabled!  Mainly this method embeds meta in TCF experiences, but upstream has already decided whether we should fetch the TCF experience to begin with -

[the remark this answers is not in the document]
pattisdr: There's so many potential locations we can add these gates, it feels like at the endpoint level rather in individual methods might work better?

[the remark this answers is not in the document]
adamsachs: yes! that makes sense. i like putting the gates pretty close to the "entrypoint" (i.e. as close to the endpoint as possible) so that farther down the stack we don't have to be concerned with it

tests/ops/util/test_ac_string.py
pattisdr: good idea -

PR #4308, opened 2023-10-19
Parse ac string from fides string

allisonking:
Closes https://github.com/ethyca/fides/issues/4263

### Description Of Changes

We are able to take a custom fides string and initialize the `tcf_consent` portion of the cookie. This PR allows us to take a fides string that has an AC string attached, and initialize the AC values on our cookie as well.

### Code Changes

* [x] Util functions to parse fidesString and get IDs from AC strings
* [x] Unit tests for these functions
* [x] Call these functions and initialize `vendor_consents` based on AC string
* [x] Cypress test

### Steps to Confirm

* Set up TCF with some AC vendors (for example, Facebook)
* Opt out of everything in your overlay. Examine your cookie, and you should see under `fides_string` an entry like `CPz5G8APz5G8AGXABBENAWEAAACAAEIAAAAAAAAKhgAgKhAA.IJawBAABAD-AVCBLUAAA,1~`. The first part might differ, but it should end in `,1~`. This is because there are no AC vendors opted in.
* Make a note of the vendor ID for the AC vendor you added (Facebook is `gacp.89`)
* Now tack on a query parameter `fides_string=CPz5G8APz5G8AGXABBENAWEAAACAAEIAAAAAAAAKhgAgKhAA.IJawBAABAD-AVCBLUAAA,1~89` to your URL, i.e. http://localhost:3001/fides-js-demo.html?geolocation=fi-18&fides_string=CPz5G8APz5G8AGXABBENAWEAAACAAEIAAAAAAAAKhgAgKhAA.IJawBAABAD-AVCBLUAAA,1~89. This passes in the fides string as a query parameter for initialization. Replace the `89` with the AC vendor ID you added
* Now when you visit the page and click on the Vendors tab, you should see Facebook opted in

### Pre-Merge Checklist

* [x] Update `CHANGELOG.md`


clients/fides-js/__tests__/lib/tcf/fidesString.ts
```diff
@@ -0,0 +1,86 @@
...
+      },
+    },
+    // With extra unexpected stuff
+    {
+      fidesString:
+        "CPzvOIAPzvOIAGXABBENAUEAAACAAAAAAAAAAAAAAAAA.IAAA,1~2.3.4,extrastuff",
+      expected: {
+        tc: "",
+        ac: "",
+      },
```
allisonking, line 46: is this actually what we expect? or would we rather still be able to parse `tc` and `ac` out? @NevilleS

clients/fides-js/src/lib/tcf/utils.ts
```diff
@@ -24,6 +25,16 @@ export const transformFidesStringToCookieKeys = (
       cookieKeys[cookieKey] = items;
     }
   });
+
+  // Set AC consents, which will only be on vendor_consents
```
allisonking, line 29: I'm actually not sure if this is true. Will AC consents only apply to `vendor_consents`? This might not be true if an AC vendor was assigned a legitimate interest purpose—is that a state we can expect? 

if so, this big of logic gets more complicated to implement because at this point we don't have the experience yet—we can't tell if the AC string is talking about consent or legitimate interest (unless AC is by definition, consent only?)

clients/fides-js/__tests__/lib/tcf/fidesString.ts
pattisdr: good question, my PR on the backend fails if handed a string that looks like this.

clients/fides-js/src/lib/tcf/utils.ts
pattisdr: I'm glad you raised this because I likely need to revisit on the backend too for related reasons.

When someone hands the backend an AC string with which to save preferences, I am saving everything under vendor consents, while it's technically possible an AC system shows up in the Vendor Legitimate Interests section too https://github.com/ethyca/fides/pull/4266#discussion_r1357097200

clients/fides-js/__tests__/lib/tcf/fidesString.ts
NevilleS: Hmm... I'm OK with erroring in that case, although I think I prefer the idea of being (slightly) forwards-compatible here and just ignore the "extrastuff" suffix but still respect the TC & AC segments.

clients/fides-js/src/lib/tcf/utils.ts
NevilleS: Yes, just vendor consents - it's explicitly _only_ for consent basis, that's a very reasonable assumption to make here.

clients/fides-js/__tests__/lib/tcf/fidesString.ts
pattisdr: To follow up from slack Allison, I'll be updating the backend to parse out TC and AC string here successfully and not erroring.

clients/fides-js/src/lib/tcf/utils.ts
pattisdr: Updating here Allison, I now also prevent AC systems from showing up in the Vendor Legitimate interests section of the experience too so everything matches

pattisdr:
Here's some [nuances around how the backend works](https://ethyca.atlassian.net/wiki/spaces/EN/pages/2819358794/2023-10-20+Documented+Backend+AC+String+and+AC+System+Behavior) around both returning AC systems and building AC strings

clients/fides-js/__tests__/lib/tcf/fidesString.ts
allisonking: thanks for the input—I also updated this to ignore `extrastuff` and successfully parse TC and AC out 👍

clients/fides-js/src/lib/tcf/utils.ts
allisonking: thanks, both! will keep this logic as-is then :)

clients/fides-js/src/lib/tcf/fidesString.ts
```diff
@@ -0,0 +1,39 @@
...
+    }
+    return { tc, ac };
+  }
+  return { tc: "", ac: "" };
+};
+
+/**
+ * Given an AC string, return a list of its ids, encoded
+ *
+ * @example 1~1.2.3 --> [gacp.1, gacp.2, gacp.3]
```
eastandwestwind, line 22: Nice! What does the `@example` annotation do exactly?

clients/fides-js/src/lib/tcf/fidesString.ts
```diff
@@ -0,0 +1,39 @@
...
+ */
+export const idsFromAcString = (acString: string) => {
+  const isValidAc = /\d~/;
+  if (!isValidAc.test(acString)) {
+    return [];
```
eastandwestwind, line 27: can we debug log if we have an invalid AC str?

clients/fides-js/src/lib/tcf/utils.ts
```diff
@@ -4,13 +4,13 @@ import { EnabledIds, TcfCookieConsent, TcfCookieKeyConsent } from "./types";
 import { TCF_KEY_MAP } from "./constants";
 import { generateFidesString } from "../tcf";
 import { debugLog } from "../consent-utils";
+import { decodeFidesString, idsFromAcString } from "./fidesString";
 
 export const transformFidesStringToCookieKeys = (
   fidesString: string,
   debug: boolean
 ): TcfCookieConsent => {
-  // Defer: to fully support AC string, we need to split out TC from AC string https://github.com/ethyca/fides/issues/4263
```
eastandwestwind, line 12: 🎉

eastandwestwind (approved):
looks great, just a Q and a small nit, nothing blocking.

clients/fides-js/src/lib/tcf/fidesString.ts
allisonking: oh heh I think I had just seen it around before, but only just looked up the documentation: https://jsdoc.app/tags-example.html

so it actually looks like it should be in the format of calling the function to show it being used as an example... I'll update the docstring, but basically it formats the following line to look like code

image

allisonking:
Failing test is one we know about: https://ethyca.atlassian.net/browse/PROD-1272?atlOrigin=eyJpIjoiNzliMjlkYjU4ZjljNDg4ZWEzMDc2YzUzMDcyZDZmMmUiLCJwIjoiaiJ9

PR #4309, opened 2023-10-20
add multiple systems component

### LINKED ISSUE #4230: Add multiple vendors component
This ticket encompasses adding a new bulk system/vendor adding flow that creates systems with data automatically filled in with data from compass.

The PR has more details about how everything should work: https://github.com/ethyca/fides/pull/4309

TheAndrewJackson:
Closes #4230 

### Description Of Changes

> [!NOTE]  
> There is still a small about of additional work being done in the plus PR. That doesn't effect the outcome of this PR though. This is just for the FE portion and is still ready to review.

This adds a new component that handles the bulk creation of systems/vendors based off of data from compass.

### Code Changes

* [ ] Complete table refactor. This change is additive. The original tables can be migrated over gradually over time as described [here](https://ethyca.atlassian.net/wiki/spaces/EN/pages/2810707987/2023-10-13+Migration+to+react-table+V8)
  * [ ] uses version 8 of react table
  * [ ] refactored logic to make the table more modular
  * [ ] uses the brand new table V3 designs for styling
  * [ ] Generally better DX due to version 8 and refactor
* [ ] Adds two new pages. Their nearly identical in code with the wording being different on each page
  * [ ] Bulk System creation page
  * [ ] Bulk Vendor creation page
* [ ] Upgrade modal that links to a new tab if a required feature isn't enabled.

### Steps to Confirm

* [x] checkout fides on `ajackson/add-multiple-systems-component`
* [x] checkout fidesplus on `ajackson/add_dictionary_routes/PROD-1069` - make sure plus is configured to use compass
* [x] checkout compass on `ajackson/get-all-privacy-declarations` - make sure compass has it's config values
* [x] Run all three
* [x] Navigate to `/add-systems` and make sure the `Add multiple systems` card is there.
* [x] Clicking it should bring you to `/add-systems/multiple`
* [x] Now briefly, turn of fidesplus, and configure compass to be disabled
* [ ] Now go back to `/add-systems` and click on the card. It should bring up a modal that says to either upgrade or create one system at a time. One button links out, the order goes to the single creation page.
* [ ] Now turn off fidesplus and conigure compass to be enabled and turn back on
* [x] Go to `/add-systems/multiple`
* [x] Make sure the search bar works
* [x] Check that paging works
* [x] Click on row checkbox should trigger a purple row to pop up
* [x] Clicking the header checkbox should enable/disable all rows
* [x] With a few rows enable click the `Add Systems` button and add the systems. You should be redirected to the datamap page
* [x] Navigate to the configure page (`/systems/configure/<id>`) for one of the systems you just created and check that everything was filled out.
* [x] Now navigate to `/consent/configure` and click the `Add multiple vendors` button. It'll redirect you to a page above that looks almost the same. 
* [x] Repeat the above steps with adding systems to test out that it works

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`


clients/admin-ui/src/features/common/FixedLayout.tsx
```diff
@@ -9,42 +9,49 @@ import { NavTopBar } from "~/features/common/nav/v2/NavTopBar";
...
+}) => {
+  const style = isDefaultLayoutPadding
+    ? {
+        padding: "40px",
+        paddingRight: "48px",
+      }
+    : {
+        paddingTop: "40px",
+        paddingLeft: "40px",
+      };
```
TheAndrewJackson, line 26: This is the actual change. I'm surprised it diffed it like this.

clients/admin-ui/src/features/common/tablev2/FidesTable.tsx
```diff
@@ -0,0 +1,157 @@
...
+/* eslint-enable */
+
+type Props<T> = {
+  tableInstance: TableInstance<T>;
+  rowActionBar?: ReactNode;
+  footer?: ReactNode;
+  onRowClick?: (row: T) => void;
+};
+
+export const FidesTableV2 = <T,>({
```
TheAndrewJackson, line 47: The biggest change here was moving the `useTable` (now `useReactTable`) hook outside the component. It was causing issues with changing the UI in a parent component. It's easier to pass the table instance in as a prop so the parent/surrounding components can update based on the table state.

clients/admin-ui/src/features/common/tablev2/FidesTable.tsx
```diff
@@ -0,0 +1,157 @@
...
+    overflowY="auto"
+    borderBottomWidth="1px"
+    borderBottomColor="gray.200"
+  >
+    <Table
+      variant="unstyled"
+      style={{
+        borderCollapse: "separate",
+        borderSpacing: 0,
+      }}
```
TheAndrewJackson, line 63: Styling it this way allows for a sticky header that doesn't have its borders move. `borderCollapse: collapse` has a weird issue where the headers borders will scroll away while the row itself remains sticky.

clients/admin-ui/src/features/common/tablev2/PaginationBar.tsx
```diff
@@ -0,0 +1,61 @@
...
+  IconButton,
+  Text,
+} from "@fidesui/react";
+import { Table as TableInstance } from "@tanstack/react-table";
+
+type PaginationBarProps<T> = {
+  tableInstance: TableInstance<T>;
+};
+
+export const PaginationBar = <T,>({ tableInstance }: PaginationBarProps<T>) => {
```
TheAndrewJackson, line 14: I didn't include the surrounding components above and below the table within the `FidesTable` component because it starts to complicate the logic and styling of it a lot.

It made things easier to grok by separating them like this.

clients/admin-ui/src/features/common/tablev2/FidesTable.tsx
```diff
@@ -0,0 +1,157 @@
...
+                  header.column.columnDef.header,
+                  header.getContext()
+                )}
+              </Th>
+            ))}
+          </Tr>
+        ))}
+      </Thead>
+      <Tbody>
+        {rowActionBar}
```
TheAndrewJackson, line 106: The row action bar is rendered separatly from the react table state because, while the library provides a pinning API, it only works nicely for rows that match the column schema defined for the table. 

Since this is a visual row that's separate from the table state it was easier to render it this way.

clients/admin-ui/src/features/common/tablev2/cells.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+export const DefaultCell = ({ value }: { value: string }) => (
+  <Flex alignItems="center" height="100%">
+    <Text fontSize="xs" lineHeight={4} fontWeight="normal">
+      {value}
+    </Text>
+  </Flex>
+);
+
+type IndeterminateCheckboxCellProps = {
```
TheAndrewJackson, line 20: I didn't port over all of the cells. I think it makes more sense to gradually port them over as needed.


We don't need `FidesObject` anymore 🥳

clients/admin-ui/src/features/system/AddMultipleSystems.tsx
```diff
@@ -0,0 +1,249 @@
...
+
+export const VendorSourceCell = ({ value }: { value: string }) => (
+  <Flex alignItems="center" height="100%">
+    <Badge>{value.split(".")[0] === "gvl" ? "GVL" : "AC"}</Badge>
+  </Flex>
+);
+
+type MultipleSystemTable = DictSystems;
+
+const columnHelper = createColumnHelper<MultipleSystemTable>();
```
TheAndrewJackson, line 52: This is a new helper function that react table provides. It makes it to typing new column definitions is really easy. It handles getting all of the types setup correctly.

clients/admin-ui/src/features/system/AddMultipleSystems.tsx
```diff
@@ -0,0 +1,249 @@
...
+        innerRowSelection[index] = true;
+      }
+    });
+    return innerRowSelection;
+  }, [dictionaryOptions]);
+
+  const tableInstance = useReactTable<MultipleSystemTable>({
+    columns,
+    data: dictionaryOptions,
+    getCoreRowModel: getCoreRowModel(),
```
TheAndrewJackson, line 136: hooks no longer exist in the table library. Before the order that hooks ran was dependent on the order you passed them in. 

Now the order that each step (sorting, filtering, grouping, etc) is deterministic. This is why everything is explicitly passed in on a key word.

clients/admin-ui/src/features/system/AddMultipleSystems.tsx
```diff
@@ -0,0 +1,249 @@
...
+    enableRowSelection: true,
+    enableSorting: true,
+    enableGlobalFilter: true,
+    state: {
+      globalFilter,
+    },
+    initialState: {
+      rowSelection,
+      pagination: {
+        pageSize: 25,
```
TheAndrewJackson, line 150: Eventually pageSize will be dynamic but it's hardcoded for now.

clients/admin-ui/src/features/system/AddMultipleSystems.tsx
```diff
@@ -0,0 +1,249 @@
...
+  if (isPostLoading || isPostSuccess) {
+    return (
+      <Flex height="100%" justifyContent="center" alignItems="center">
+        <Spinner />
+      </Flex>
+    );
+  }
+
+  if (isGetLoading) {
+    return <TableSkeletonLoader rowHeight={36} numRows={15} />;
```
TheAndrewJackson, line 186: The skeleton loader is only loaded when the table itself is having its data loaded.

The post route gets a spinner because it redirects to a new page after it finishes.

clients/admin-ui/src/features/system/AddMultipleSystems.tsx
```diff
@@ -0,0 +1,249 @@
...
+            onClick={onOpen}
+            size="xs"
+            variant="outline"
+            disabled={!anyNewSelectedRows}
+          >
+            Add {`${systemText}s`}
+          </Button>
+        </Tooltip>
+      </TableActionBar>
+      <FidesTableV2<MultipleSystemTable>
```
TheAndrewJackson, line 233: another side note: The typing is awesome in version 8 of the table library!! The author definitely put a lot of effort into making the typing experience nice. I'm pretty sure everything is strongly typed.

You get autocomplete for the row fields.

clients/admin-ui/src/pages/add-systems/multiple.tsx
```diff
@@ -0,0 +1,57 @@
...
+      Add systems
+    </Heading>
+  </Box>
+);
+
+const AddMultipleSystemsPage: NextPage = () => {
+  const { systemOrDatamapRoute } = useSystemOrDatamapRoute();
+
+  return (
+    <FixedLayout isDefaultLayoutPadding title="Describe your system">
```
TheAndrewJackson, line 28: The fixed layout is required to make it so the table scrollbar will appear.

clients/admin-ui/src/features/common/FixedLayout.tsx
allisonking: hmm why do we want FixedLayout to have an alternate styling? I'd expect FixedLayout to be standard across everything, but there are probably cases not to do that! just trying to figure out if this is a design discrepancy, or if something else is going on.

also, an approach I'd prefer is to pass additional props as a prop into FixedLayout, i.e.

```typescript
const FixedLayout = ({
  children,
  title,
  mainProps,
}: {
  children: React.ReactNode;
  title: string;
  mainProps?: FlexProps
}) => {
...
  <Flex as="main" ... {...mainProps} />
}
```

I think that makes it a bit more flexible, and doesn't require FixedLayout to be responsible for knowing about how other components need to pad

clients/admin-ui/src/features/common/ConfirmationModal.tsx
```diff
@@ -66,7 +68,7 @@ const ConfirmationModal = ({
           <Button
             variant="outline"
             mr={3}
-            onClick={onClose}
+            onClick={onCancel || onClose}
```
allisonking, line 71: is there ever a case where we want to cancel but _not_ close the modal? if not, I feel like this might be better as

```js
onClick={() => {
  if (onCancel) {
    onCancel();
  }
  onClose();
}}
```

so that components don't have to worry about adding cancel logic _and_ closing logic

clients/admin-ui/src/features/common/tablev2/EmptyTableState.tsx
```diff
@@ -0,0 +1,35 @@
+import { Box, HStack, Text, WarningTwoIcon } from "@fidesui/react";
```
allisonking, line 1: can we namespace these files as `table/v2` as opposed to `tablev2` to mimic what we already have for `nav/v2`?

clients/admin-ui/src/features/common/tablev2/TableSkeletonLoader.tsx
```diff
@@ -0,0 +1,16 @@
+import { Skeleton, Stack } from "@fidesui/react";
+import { FC } from "react";
+
+type Props = {
+  rowHeight: number;
+  numRows: number;
+};
+
+export const TableSkeletonLoader: FC<Props> = ({ rowHeight, numRows }) => {
```
allisonking, line 9: oh cool, I didn't know there was such thing as a skeleton!

clients/admin-ui/src/features/common/tablev2/TableSkeletonLoader.tsx
```diff
@@ -0,0 +1,16 @@
...
+  const rows = [];
+  for (let i = 0; i < numRows; i += 1) {
+    rows.push(<Skeleton height={`${rowHeight}px`} key={i} />);
+  }
+
+  return <Stack>{rows}</Stack>;
```
allisonking, line 15: this feels more react-y to me, and then I don't think you have to do the for loop plus rows array:

```js
<Stack>
  {[...Array(numRows).keys()].map(key => {
  <Skeleton height=`${rowHeight}px`} key={key} />
})}
</Stack>
```

clients/admin-ui/src/features/common/UpgradeModal.tsx
```diff
@@ -0,0 +1,40 @@
...
+  return (
+    <ConfirmationModal
+      isOpen={isOpen}
+      onClose={onClose}
+      onCancel={() => {
+        router.push(nonUpgradeLink);
+      }}
+      isCentered
+      title={`Upgrade to add multiple ${templateWord}s at once!`}
+      message={`You will need to upgrade your subscription to add multiple ${templateWord}s at once. In the meantime you can continue here to add indivual ${templateWord}s manually.`}
```
allisonking, line 32: is there a case we wouldn't want to show this upgrade modal? like if a user is already subscribed to... something? I guess if the dictionary is available?

clients/admin-ui/src/features/system/AddMultipleSystems.tsx
```diff
@@ -0,0 +1,252 @@
...
+import {
+  DictSystems,
+  selectAllDictSystems,
+  useGetAllCreatedSystemsQuery,
+  usePostCreatedSystemsMutation,
+} from "~/features/plus/plus.slice";
+
+export const VendorSourceCell = ({ value }: { value: string }) => (
+  <Flex alignItems="center" height="100%">
+    <Badge>{value.split(".")[0] === "gvl" ? "GVL" : "AC"}</Badge>
```
allisonking, line 46: ah we're going to be doing this kind of operation pretty often I'm imagining—we'll probably want it as its own function eventually (see https://github.com/ethyca/fides/blob/main/clients/fides-js/src/lib/tcf/vendors.ts/#L10-L25 for reference)

clients/admin-ui/src/features/system/AddMultipleSystems.tsx
allisonking: awesome!

clients/admin-ui/src/features/system/AddMultipleSystems.tsx
allisonking: neat, I like that more!

clients/admin-ui/src/features/system/AddMultipleSystems.tsx
```diff
@@ -0,0 +1,252 @@
...
+  const addVendors = async () => {
+    const vendorIds = tableInstance
+      .getSelectedRowModel()
+      .rows.filter((r) => !r.original.linked_system)
+      .map((r) => r.original.vendor_id);
+    if (vendorIds.length > 0) {
+      await postVendorIds(vendorIds);
+      router.push(redirectRoute);
+      toast(
+        successToastParams(
```
allisonking, line 167: should we handle the error case too?

clients/admin-ui/src/features/common/tablev2/FidesTable.tsx
```diff
@@ -0,0 +1,157 @@
...
+        {tableInstance.getRowModel().rows.map((row) => {
+          // @ts-ignore
```
allisonking, line 108: wait I thought we got types now!

clients/admin-ui/src/pages/add-systems/multiple.tsx
```diff
@@ -0,0 +1,57 @@
...
+import FixedLayout from "~/features/common/FixedLayout";
+import { useSystemOrDatamapRoute } from "~/features/common/hooks/useSystemOrDatamapRoute";
+import {
+  ADD_SYSTEMS_ROUTE,
+  DATAMAP_ROUTE,
+} from "~/features/common/nav/v2/routes";
+import { AddMultipleSystems } from "~/features/system/AddMultipleSystems";
+
+const DESCRIBE_SYSTEM_COPY =
+  "Select and add systems directory to your data map. All s ystems available here are from the TCF Global Vendor list, Google's AC list, and Fides Compass. All Systems come pre-configured so there is no need for your to do anything!";
```
allisonking, line 14:
```suggestion
  "Select and add systems directory to your data map. All systems available here are from the TCF Global Vendor list, Google's AC list, and Fides Compass. All Systems come pre-configured so there is no need for you to do anything!";
```

clients/admin-ui/src/pages/consent/configure/add-vendors.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+import FixedLayout from "~/features/common/FixedLayout";
+import {
+  ADD_MULTIPLE_VENDORS_ROUTE,
+  CONFIGURE_CONSENT_ROUTE,
+} from "~/features/common/nav/v2/routes";
+import { AddMultipleSystems } from "~/features/system/AddMultipleSystems";
+
+const DESCRIBE_VENDOR_COPY =
+  "Select and add vendors. All vendors available here are from the TCF Global Vendor list, Google's AC list, and Fides Compass. All Systems come pre-configured so there is no need for your to do anything!";
```
allisonking, line 13:
```suggestion
  "Select and add vendors. All vendors available here are from the TCF Global Vendor list, Google's AC list, and Fides Compass. All Systems come pre-configured so there is no need for you to do anything!";
```

allisonking (commented):
haven't had a chance to test myself yet, but leaving some comments! this is awesome though, react table v8 looks very nice 😎 nice work!

clients/admin-ui/src/features/common/ConfirmationModal.tsx
TheAndrewJackson: I see what you mean. I did this because I needed a separate function from `onClose`. The reason for that is the modal needs an `onClose` for when a user clicks on the blurred overlay that surrounds that modal. I needed to add a custom cancel function because now the cancel will handle doing the other "non upgrade" option in the modal.


I like your proposed solution. I'll update it to do it that way :+1:

clients/admin-ui/src/features/common/FixedLayout.tsx
TheAndrewJackson: It needs different styling because the datamap page (the only other page that uses it AFAIK) is expected to have no right padding and hug the right side of the screen. This new page that uses it is expected to have right padding.

That's a good idea! I'll update it to do it that way.

clients/admin-ui/src/features/common/UpgradeModal.tsx
TheAndrewJackson: Yes, that's correct. I handled that in the `AddSystem` component :slightly_smiling_face: 

https://github.com/ethyca/fides/blob/aecf560ad52d4f3b28da9c21f4b61ce114e68cbc/clients/admin-ui/src/features/config-wizard/AddSystem.tsx#L91-L98

clients/admin-ui/src/features/common/tablev2/FidesTable.tsx
TheAndrewJackson: Good catch, that's a leftover from the refactor

clients/admin-ui/src/features/common/tablev2/TableSkeletonLoader.tsx
TheAndrewJackson: I actually tried that same method. It wasn't working when I tried and had no errors. I fell back to other method that's currently in use.

clients/admin-ui/src/features/system/AddMultipleSystems.tsx
TheAndrewJackson: we should just in case. I'll update it :+1:

clients/admin-ui/src/features/common/tablev2/EmptyTableState.tsx
TheAndrewJackson: we can. good idea!

src/fides/api/alembic/migrations/versions/ddec29f24945_update_cookie_max_age_to_bigint.py
```diff
@@ -0,0 +1,26 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "ddec29f24945"
+down_revision = "c5a218831820"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.alter_column("ctl_systems", "cookie_max_age_seconds", type_=BIGINT)
```
TheAndrewJackson, line 22: @adamsachs here is the migration that we talked about

TheAndrewJackson: I was able to upgrade and downgrade successfully with data in that column. I also manually changed the type back and forth within pgadmin and it worked there too

TheAndrewJackson:
@allisonking one of the tests that's failing is the flaky `system-plus` test. The other one seems unrelated to my work but I can take a look at it

clients/admin-ui/cypress/e2e/systems-plus.cy.ts
```diff
@@ -159,4 +166,60 @@ describe("System management with Plus features", () => {
       });
     });
   });
+
+  describe("bulk system/vendor adding page", () => {
```
TheAndrewJackson, line 170: I added some initial test coverage. It doesn't cover every possible permutation of functionality but it makes sure the page loads and the tables most important functions are working.

I didn't cover every possible scenario because I think this page is likely to be changed soon :sweat_smile:

allisonking:
beginning UAT testing 😎

allisonking:
should these perhaps use different icons? I don't have the link the original design, maybe it is meant to use the same icon?
image

TheAndrewJackson:
@allisonking good callout. I think that's okay for now. I reviewed it with Simon this morning and he didn't mentioned the logos

allisonking:
can this be sorted in any way? they seem a little haphazard 
image

allisonking:
it it's easy, perhaps this purple header can be sticky? 
image

allisonking:
I feel like checking this top checkbox should only check the 25 listed on the page... seems like we have fake paging going on (I'd expect paging to be useful to limit the request we get from the BE, but it looks like we still get the whole obj, which makes for a weird user experience in this case of checking everything)

image

allisonking:
we should pluralize `system` here (conditionally, probably)
image

TheAndrewJackson:
> can this be sorted in any way? they seem a little haphazard image

It can be sorted by clicking the header. It's just not sorted by default. Is that satisfactory? I think it's possible to have something sorted by default. I didn't do it because it wasn't in the requirements.

TheAndrewJackson:
> it it's easy, perhaps this purple header can be sticky? image

That's weird. That header used to be sticky. I'm pretty sure it was sticky when I demoed this morning :sweat_smile: I'll look into that.

TheAndrewJackson:
> I feel like checking this top checkbox should only check the 25 listed on the page... seems like we have fake paging going on (I'd expect paging to be useful to limit the request we get from the BE, but it looks like we still get the whole obj, which makes for a weird user experience in this case of checking everything)
> 
> image

I actually talked through this with Simon and that's the desired behavior. The reason for it is we'll have more filtering at some point the it's useful to have that select everything.

It can be changed in the future if people don't like it though

TheAndrewJackson:
> we should pluralize `system` here (conditionally, probably) image
> 
> The toast afterwards should also be pluralized

Yes, it should. Good catch!

clients/admin-ui/src/features/common/UpgradeModal.tsx
```diff
@@ -0,0 +1,40 @@
...
+      cancelButtonText={`Add ${templateWord}s manually`}
+      continueButtonText="Upgrade"
+      onConfirm={() => {
+        window.open(upgradeLink);
```
allisonking, line 36: hmm I haven't seen `window.open` much before. the nextjs router docs suggest using `window.location` https://nextjs.org/docs/pages/api-reference/functions/use-router

I don't know if that makes a big difference but it feels a little more familiar 🤷‍♀️

clients/admin-ui/src/features/common/helpers.ts
```diff
@@ -148,3 +148,16 @@ export const enumToOptions = (e: { [s: number]: string }) =>
     value: entry[1],
     label: entry[1],
   }));
+
+export enum VendorSources {
```
allisonking, line 152: woot thanks for adding this!

allisonking (approved):
Okay, I've tested locally and think it's reasonable! My main gripe is that checking all checks all 1000+ when I think it should be at the page level, but maybe that's not the intention. otherwise I think this looks good! nice job, I'm excited for react-table v8!

allisonking:
> It can be sorted by clicking the header. It's just not sorted by default. Is that satisfactory? I think it's possible to have something sorted by default. I didn't do it because it wasn't in the requirements.

okay that's fine with me!

> I actually talked through this with Simon and that's the desired behavior. The reason for it is we'll have more filtering at some point the it's useful to have that select everything.
>
> It can be changed in the future if people don't like it though
> 
> edit: also if by fake paging you mean client side only paging then yes it's fake haha. That is known and okay for now

ah okay I didn't see this before my review—in that case, lgtm! no more gripes from me :)

clients/admin-ui/src/features/common/UpgradeModal.tsx
TheAndrewJackson: I used that method because I need to open that link in a new tab.

TheAndrewJackson: AFAIK `window.location` isn't used for that

TheAndrewJackson: although, I did just learn I may need blank to ensure it always opens in a new tab: https://sentry.io/answers/how-to-open-a-url-in-a-new-tab-(and-not-a-new-window)/

PR #4313, opened 2023-10-20
Update Version Hash Contents 

pattisdr:
Closes https://github.com/ethyca/fidesplus/issues/1172

### Description Of Changes

 Update the contents of the version hash that is used to determine when consent should be resurfaced for TCF.  Previously we hashed `purpose x legal basis` and `vendor x legal basis` to match what could be saved in the string.  This largely updates to save three dimensions together `vendor x legal basis x purpose` and also includes AC vendors and other applicable systems in the hash, as well as determination of whether vendors moved from GVL 2 to GVL 3
    
    New sections are:
      - Policy version (no change)
      - Vendor purpose consents
      - Vendor purpose legitimate interests
      - gvl vendors disclosed

### Code Changes

* [x] Updates the fields in `TCFVersionHash` which is the object built prior to constructing the version hash

### Steps to Confirm

* [ ] Set tcf_enabled is True, run `nox -s dev -- shell`
* [ ] [Add GVL System](https://ethyca.atlassian.net/wiki/spaces/EN/pages/2809724964/API+How-to+Create+GVL+System+Locally)
* [ ] GET Privacy Experiences with include_meta query param:
```bash
curl -X 'GET' \
  'http://localhost:8080/api/v1/privacy-experience?show_disabled=true&region=eea&systems_applicable=false&include_gvl=false&include_meta=true&page=1&size=50' \
  -H 'accept: application/json'
```
* [ ] Note stays the same on repeated API calls.  
* [ ] Add new system, or add new GVL purpose to an existing system or update the legal basis of an existing GVL purpose. Note the hash changes.

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`


src/fides/api/util/tcf/experience_meta.py
```diff
@@ -1,32 +1,49 @@
...
+    )
+    gvl_vendors_disclosed: List[int] = Field(
+        description="List of GVL vendors disclosed from the last vendor list"
+    )
 
     @root_validator()
     @classmethod
     def sort_lists(cls, values: Dict) -> Dict:
-        """Verify lists are sorted ascending for repeatability"""
+        """Verify lists are sorted for repeatability"""
```
pattisdr, line 46: This existing validator on the TCFVersionHash sorts the keys in each list. It's currently sorting them as strings, for example, "gacp.100" comes before "gacp.8-1".  Although this might not be how humans would expect this to be sorted, I am not implementing a more complex sorting strategy here because these are just the raw contents of the hash.

I experimented with custom sort functions that treated the letter and numerical sections separately, but was concerned about potential future vendor keys that might break this. I think the important thing is that it is sorted.
```
[
      "gacp.100",
      "gacp.8-1",
      "gvl.2-1,2,3,4,7,9,10",
      "gvl.8-1,3,4",
  ]
```

src/fides/api/util/tcf/experience_meta.py
```diff
@@ -46,18 +63,74 @@ def _build_tcf_version_hash_model(
...
+            vendor_id: str = vendor_record.id
+            purposes_key: str = (
+                "purpose_consents"
+                if isinstance(vendor_record, TCFVendorConsentRecord)
+                else "purpose_legitimate_interests"
+            )
+            purpose_ids = sorted(
+                purpose_record.id
+                for purpose_record in getattr(vendor_record, purposes_key)
+            )
```
pattisdr, line 91: Purposes should already be in order from the base experience but just in case, they are sorted here again -

src/fides/api/util/tcf/experience_meta.py
```diff
@@ -1,32 +1,49 @@
...
+        description="Stores vendor * purpose * consent legal basis in the format "
+        "<universal_vendor_id>-<comma separated purposes>"
+    )
+    vendor_purpose_legitimate_interests: List[str] = Field(
+        description="Stores vendor * purpose * legitimate interest legal basis in the format"
+        " <universal_vendor_id>-<comma separated purposes>"
+    )
+    gvl_vendors_disclosed: List[int] = Field(
+        description="List of GVL vendors disclosed from the last vendor list"
+    )
```
adamsachs, line 41: nice choice to add field descriptions here as this functionality becomes more nuanced!

src/fides/api/util/tcf/experience_meta.py
```diff
@@ -1,32 +1,49 @@
...
+        "re-establish transparency and consent from users."
+    )
+    vendor_purpose_consents: List[str] = Field(
+        description="Stores vendor * purpose * consent legal basis in the format "
+        "<universal_vendor_id>-<comma separated purposes>"
+    )
+    vendor_purpose_legitimate_interests: List[str] = Field(
+        description="Stores vendor * purpose * legitimate interest legal basis in the format"
+        " <universal_vendor_id>-<comma separated purposes>"
+    )
```
adamsachs, line 38: nit - any reason the spacing isn't consistent between the two here? just to make it all a bit less ambiguous, maybe do something like:
```suggestion
        description="Stores vendor * purpose * consent legal basis in the format: "
        "<universal_vendor_id>-<comma separated purposes>"
    )
    vendor_purpose_legitimate_interests: List[str] = Field(
        description="Stores vendor * purpose * legitimate interest legal basis in the format: "
        "<universal_vendor_id>-<comma separated purposes>"
    )
```

i wouldn't normally nitpick this much, but we don't have any "specs" for the values here outside of the code, so i figure it's worth removing any room for ambiguity!

src/fides/api/util/tcf/experience_meta.py
adamsachs: this makes perfect sense and i think your decision is a wise one!

could add in a note
```suggestion
        """Verify lists are deterministically sorted for repeatability"""
```
but maybe that's overkill :)

src/fides/api/util/tcf/experience_meta.py
```diff
@@ -46,18 +63,74 @@ def _build_tcf_version_hash_model(
...
     """
+
+    def build_vendor_by_purpose_strings(
+        vendor_list: Union[
+            List[TCFVendorConsentRecord], List[TCFVendorLegitimateInterestsRecord]
+        ]
+    ) -> List[str]:
+        """Hash helper for building vendor x purposes.  Converts a list of vendor records
+        into a list of strings each consisting of a vendor id, a separator (-), and comma
+        separated purposes (if applicable) with the given legitimate interest.
```
adamsachs, line 74: should be
```suggestion
        separated purposes (if applicable) with the given legal basis.
```
right?

src/fides/api/util/tcf/experience_meta.py
```diff
@@ -46,18 +63,74 @@ def _build_tcf_version_hash_model(
...
     model: TCModel = convert_tcf_contents_to_tc_model(
         tcf_contents, UserConsentPreference.opt_in
     )
-    return TCFVersionHash(**model.dict())
 
+    return TCFVersionHash(
+        policy_version=model.policy_version,
+        vendor_purpose_consents=vendor_consents,
+        vendor_purpose_legitimate_interests=vendor_li,
+        gvl_vendors_disclosed=model.vendors_disclosed,
```
adamsachs, line 120: may be worth commenting here that this is specifically GVL vendors that show up in the "current" GVL reference list. i know that can be found by tracing back and digging into the docstrings/code that is responsible for populating  `model.vendors_disclosed` (i did it!), that's a key nuance here that could easily get overlooked or forgotten about.

```suggestion
        gvl_vendors_disclosed=model.vendors_disclosed, # `vendors_disclosed` field only has GVL vendors present in "current" GVL vendor list
```

tests/ops/util/test_tc_string.py
```diff
@@ -92,30 +137,61 @@ def test_build_tcf_version_hash_removing_declaration(
...
-            vendor_legitimate_interests=[],
+            vendor_purpose_consents=["gvl.2-2,3,4,7,9,10"],
+            vendor_purpose_legitimate_interests=[],
+            gvl_vendors_disclosed=[2],
+        )
+
+        version_hash = build_tcf_version_hash(tcf_contents)
+        assert version_hash == "f2e19b4b3eae"
+
+    def test_build_tcf_version_hash_adding_vendor(self, db, system):
```
adamsachs, line 148: great test cases!

adamsachs (approved):
looks great @pattisdr , nice work! mainly just some documentation nits, but i think the functionality here is spot on and it's a clean update to capture the new requirements 👍

src/fides/api/util/tcf/experience_meta.py
pattisdr: that's good 👍

src/fides/api/util/tcf/experience_meta.py
pattisdr: Thanks for the cleanup, just accidentally broke the lines in two separate places

src/fides/api/util/tcf/experience_meta.py
pattisdr: hahaha yes too much TCF, thank you!

src/fides/api/util/tcf/experience_meta.py
pattisdr: I can add this 👍

PR #4318, opened 2023-10-23
Adds ability to provide custom fn to save user prefs

eastandwestwind:
Closes https://ethyca.atlassian.net/browse/PROD-1239

### Description Of Changes

Adds ability to provide custom fn to save user prefs

### Code Changes

* [ ] Adds new `api` param in Fides.init() options to store all current & future custom fn overrides  
* [ ] Refactor so that we have access to global Fides obj at time of custom fn call
* [ ] Add e2e test to ensure we don't call our internal API on user preference save

### Steps to Confirm

* [ ] Run e2e test OR provide custom fn override in Fides.init() and see that it is called

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### DIFF
--- a/clients/fides-js/src/fides-tcf.ts
+++ b/clients/fides-js/src/fides-tcf.ts
@@ -217,8 +217,8 @@ const init = async (config: FidesConfig) => {
   }
   const experience = initialFides?.experience ?? config.experience;
   const updatedFides = await initialize({
+    ...config,
     cookie,
-    config,
     experience,
     renderOverlay,
     updateCookie,
@@ -255,7 +255,7 @@ _Fides = {
     fidesEmbed: false,
     fidesDisableSaveApi: false,
     fidesString: null,
-    api: null,
+    apiOptions: null,
   },
   fides_meta: {},
   identity: {},
--- a/clients/fides-js/src/fides-tcf.ts
+++ b/clients/fides-js/src/fides-tcf.ts
@@ -217,9 +217,9 @@ const init = async (config: FidesConfig) => {
   }
   const experience = initialFides?.experience ?? config.experience;
   const updatedFides = await initialize({
-    ...config,
-    experience,
     cookie,
+    config,
+    experience,
     renderOverlay,
     updateCookie,
   });
@@ -255,6 +255,7 @@ _Fides = {
     fidesEmbed: false,
     fidesDisableSaveApi: false,
     fidesString: null,
+    api: null,
   },
   fides_meta: {},
   identity: {},
--- a/clients/fides-js/src/lib/consent-types.ts
+++ b/clients/fides-js/src/lib/consent-types.ts
@@ -14,6 +14,7 @@ import type {
   TCFVendorLegitimateInterestsRecord,
   TCFVendorRelationships,
 } from "./tcf/types";
+import { CookieKeyConsent } from "~/lib/cookie";
 
 export type EmptyExperience = Record<PropertyKey, never>;
 
@@ -74,6 +75,27 @@ export type FidesOptions = {
   // An explicitly passed-in TC string that supersedes the cookie, and prevents any API calls to fetch
   // experiences / preferences. Only available when TCF is enabled. Optional.
   fidesString: string | null;
+
+  // Allows for explicit overrides on various internal API calls made from Fides.
+  api: FidesApiOptions | null;
+};
+
+export type FidesApiOptions = {
+  /**
+   * Intake a custom save function to upload the saved `fides_string` alongside
+   * the `version_hash` of the notices it was created for to our own API.
+   *
+   * @param {object} Fides  - the global Fides object with all configured options, state, etc.
+   * @param {object} consent - updated version of Fides.consent with the user's saved preferences for Fides notices
+   * @param {string} fides_string - updated version of Fides.fides_string with the user's saved preferences for TC/AC/etc notices
+   * @param {object} experience - current version of the privacy experience that was shown to the user
+   */
+  savePreferencesFn: (
+    fides: FidesConfig,
+    consent: CookieKeyConsent,
+    fides_string: string | undefined,
+    experience: PrivacyExperience
+  ) => void;
 };
 
 export class SaveConsentPreference {
--- a/clients/fides-js/src/lib/consent-types.ts
+++ b/clients/fides-js/src/lib/consent-types.ts
@@ -77,24 +77,22 @@ export type FidesOptions = {
   fidesString: string | null;
 
   // Allows for explicit overrides on various internal API calls made from Fides.
-  api: FidesApiOptions | null;
+  apiOptions: FidesApiOptions | null;
 };
 
 export type FidesApiOptions = {
   /**
    * Intake a custom function that is called instead of the internal Fides API to save user preferences.
    *
-   * @param {object} Fides  - the global Fides object with all configured options, state, etc.
    * @param {object} consent - updated version of Fides.consent with the user's saved preferences for Fides notices
    * @param {string} fides_string - updated version of Fides.fides_string with the user's saved preferences for TC/AC/etc notices
    * @param {object} experience - current version of the privacy experience that was shown to the user
    */
   savePreferencesFn: (
-    fides: FidesConfig,
     consent: CookieKeyConsent,
     fides_string: string | undefined,
     experience: PrivacyExperience
-  ) => void;
+  ) => Promise<void>;
 };
 
 export class SaveConsentPreference {
--- a/clients/fides-js/src/lib/consent-types.ts
+++ b/clients/fides-js/src/lib/consent-types.ts
@@ -82,8 +82,7 @@ export type FidesOptions = {
 
 export type FidesApiOptions = {
   /**
-   * Intake a custom save function to upload the saved `fides_string` alongside
-   * the `version_hash` of the notices it was created for to our own API.
+   * Intake a custom function that is called instead of the internal Fides API to save user preferences.
    *
    * @param {object} Fides  - the global Fides object with all configured options, state, etc.
    * @param {object} consent - updated version of Fides.consent with the user's saved preferences for Fides notices
--- a/clients/fides-js/src/lib/initialize.ts
+++ b/clients/fides-js/src/lib/initialize.ts
@@ -88,15 +88,13 @@ const retrieveEffectiveRegionString = async (
 const automaticallyApplyGPCPreferences = ({
   cookie,
   fidesRegionString,
-  fidesApiUrl,
-  fidesDisableSaveApi,
   effectiveExperience,
+  fidesConfig,
 }: {
   cookie: FidesCookie;
   fidesRegionString: string | null;
-  fidesApiUrl: string;
-  fidesDisableSaveApi: boolean;
   effectiveExperience?: PrivacyExperience;
+  fidesConfig: FidesConfig;
 }) => {
   if (!effectiveExperience || !effectiveExperience.privacy_notices) {
     return;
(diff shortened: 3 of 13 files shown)


clients/fides-js/src/lib/preferences.ts
```diff
@@ -18,33 +20,30 @@ import { TcfSavePreferences } from "./tcf/types";
 
 /**
  * Updates the user's consent preferences, going through the following steps:
- * 1. Save preferences to Fides API
+ * 1. Update the cookie object based on new preferences
  * 2. Update the window.Fides.consent object
- * 3. Save preferences to the `fides_consent` cookie in the browser
- * 4. Remove any cookies from notices that were opted-out from the browser
- * 5. Dispatch a "FidesUpdated" event
+ * 3. Save preferences to Fides API
```
NevilleS, line 25:
```suggestion
 * 3. Save preferences to Fides API or a custom function (`savePreferencesFn`)
```

clients/fides-js/src/lib/preferences.ts
```diff
@@ -18,33 +20,30 @@ import { TcfSavePreferences } from "./tcf/types";
 
 /**
  * Updates the user's consent preferences, going through the following steps:
- * 1. Save preferences to Fides API
+ * 1. Update the cookie object based on new preferences
  * 2. Update the window.Fides.consent object
```
NevilleS, line 24: For clarity, this updates more than `Fides.consent` now right

clients/privacy-center/cypress/e2e/consent-banner-tcf.cy.ts
```diff
@@ -722,7 +724,83 @@ describe("Fides-js TCF", () => {
           ).to.eql(true);
         });
       });
-
```
NevilleS, line 725: nit: I really don't like the lack of newlines in between specs, is that something your IDE is doing? 😆

clients/privacy-center/cypress/e2e/consent-banner-tcf.cy.ts
```diff
@@ -722,7 +724,83 @@ describe("Fides-js TCF", () => {
...
+              if (error.message.indexOf("Timed out retrying") !== 0) {
+                throw error;
+              }
+            });
+            // check that preferences aren't sent to Fides API
+            cy.wait("@patchPrivacyPreference", {
+              requestTimeout: 500,
+            }).then((xhr) => {
+              assert.isNull(xhr?.response?.body);
+            });
```
NevilleS, line 762: 👍 though I would set the timeout lower for test suite speed!

clients/privacy-center/cypress/e2e/consent-banner-tcf.cy.ts
```diff
@@ -722,7 +724,83 @@ describe("Fides-js TCF", () => {
...
+              isOverlayEnabled: true,
+              tcfEnabled: true,
+              api: {
+                /* eslint-disable @typescript-eslint/no-unused-vars */
+                savePreferencesFn: (
+                  fides: FidesConfig,
+                  consent: CookieKeyConsent,
+                  fides_string: string | undefined,
+                  experience: PrivacyExperience
+                ) => {},
```
NevilleS, line 740: You should add an explicit assertion to ensure this save function gets called. That can be done with some Cypress spying

clients/fides-js/src/lib/preferences.ts
```diff
@@ -68,36 +67,45 @@ export const updateConsentPreferences = async ({
...
+
+  // 2. Update the window.Fides object
+  debugLog(fidesConfig.options.debug, "Updating window.Fides");
+  window.Fides.consent = cookie.consent;
+  window.Fides.fides_string = cookie.fides_string;
+  window.Fides.tcf_consent = cookie.tcf_consent;
+
+  // 3. Save preferences to Fides API
+  if (!fidesConfig.options.fidesDisableSaveApi) {
+    debugLog(fidesConfig.options.debug, "Saving preferences to Fides API");
```
NevilleS, line 82: nit: this log will be confusing if it calls a custom function instead

NevilleS (changes requested):
Overall this is looking good- this'll work. Let's keep going!

clients/fides-js/src/fides-tcf.ts
```diff
@@ -255,6 +255,7 @@ _Fides = {
     fidesEmbed: false,
     fidesDisableSaveApi: false,
     fidesString: null,
+    api: null,
```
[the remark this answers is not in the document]
eastandwestwind: we use the `FidesApiOptions` type here, so maybe `apiOptions` could work. Just should be consistent with whatever we name the type

clients/fides-js/src/lib/consent-types.ts
```diff
@@ -74,6 +75,27 @@ export type FidesOptions = {
...
   fidesString: string | null;
+
+  // Allows for explicit overrides on various internal API calls made from Fides.
+  api: FidesApiOptions | null;
+};
+
+export type FidesApiOptions = {
+  /**
+   * Intake a custom save function to upload the saved `fides_string` alongside
+   * the `version_hash` of the notices it was created for to our own API.
```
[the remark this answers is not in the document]
eastandwestwind: you're right- I'll update!

clients/fides-js/src/lib/initialize.ts
```diff
@@ -258,9 +255,8 @@ export const getInitialFides = ({
  */
 export const initialize = async ({
   cookie,
-  options,
+  config,
   experience,
-  geolocation,
```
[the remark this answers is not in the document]
eastandwestwind: yup, `geolocation` is internally fetched within `initialize.ts` so we can simply use the `geolocation` off the original `config`. `experience`, on the other hand, might have been previously updated with a `fidesString`, so we can't guarantee it's the same one that's passed into `config`. 

I could certainly update the original `config` obj in this case so we can be consistent with just using `config` for both `geolocation` and `experience`, but I need confirmation of this strategy. As it stands, we don't reassign props on `config`, aside from override params.

clients/privacy-center/cypress/e2e/consent-banner-tcf.cy.ts
eastandwestwind: oh, this was a rebase thing I probably should've kept. Will update!

clients/privacy-center/cypress/e2e/consent-banner-tcf.cy.ts
```diff
@@ -722,7 +724,83 @@ describe("Fides-js TCF", () => {
...
+        cy.waitUntilFidesInitialized().then(() => {
+          cy.get("#fides-modal-link").click();
+          cy.getByTestId("consent-modal").within(() => {
+            cy.get("button").contains("Opt out of all").click();
+            // timeout means API call not made, which is expected
+            cy.on("fail", (error) => {
+              if (error.message.indexOf("Timed out retrying") !== 0) {
+                throw error;
+              }
+            });
```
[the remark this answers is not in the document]
eastandwestwind: This is a pattern we use for other tests that require asserting a certain cypress mock was not called. Basically, we set a timeout on the call we `wait` for. That call never happens, as expected, so Cypress triggers a `fail`. 

This function intercepts the Cypress `fail`, and if we don't manually `throw error`, then Cypress won't actually fail the test. So in this case, we still want Cypress to fail the test for any other reason, _except_ when we have a timeout.

 The downfall of this is that we could have a false positive for tests that timeout for other reasons than we intended, so that's why we keep this intercept scoped to just 1 specific assertion.

clients/fides-js/src/lib/consent-types.ts
```diff
@@ -74,6 +75,26 @@ export type FidesOptions = {
...
+  /**
+   * Intake a custom function that is called instead of the internal Fides API to save user preferences.
+   *
+   * @param {object} Fides  - the global Fides object with all configured options, state, etc.
+   * @param {object} consent - updated version of Fides.consent with the user's saved preferences for Fides notices
+   * @param {string} fides_string - updated version of Fides.fides_string with the user's saved preferences for TC/AC/etc notices
+   * @param {object} experience - current version of the privacy experience that was shown to the user
+   */
+  savePreferencesFn: (
+    fides: FidesConfig,
```
NevilleS, line 93: This feels odd, why a `FidesConfig` type here? It doesn't match the comment and also is pretty narrow. Can't this be the `Fides` global?

clients/fides-js/src/lib/preferences.ts
```diff
@@ -68,36 +67,65 @@ export const updateConsentPreferences = async ({
...
+  // 2. Update the window.Fides object
+  debugLog(fidesConfig.options.debug, "Updating window.Fides");
   window.Fides.consent = cookie.consent;
   window.Fides.fides_string = cookie.fides_string;
   window.Fides.tcf_consent = cookie.tcf_consent;
 
+  // 3. Save preferences to Fides API
+  if (!fidesConfig.options.fidesDisableSaveApi) {
+    if (fidesConfig.options.api?.savePreferencesFn) {
+      try {
```
NevilleS, line 83: nit: I find this chunk of logic rather "long" and unwieldy when compared to the rest of the chunks in this `updateConsentPreferences` function. I'd recommend a few tidy-ups to look like this:
```
// 3. Save preferences to API (if not disabled)
if (!fidesConfig.options.fidesDisableSaveApi) {
  try {
    await savePreferencesApi(...);
  } catch (e) {
    debugLog(fidesConfig.options.debug, "Error saving updated preferences to API, continuing. Error: ", e);
  }
}
```

Then encapsulate the main if/else into that `savePreferencesApi` helper. This does a few things:
1. Makes this overall function much more readable, because it's effectively calling 6 helpers
2. Uses the same try/catch for both branches, I don't see much value in slightly different logs for each
3. Forces you to think about the arguments to the two branches and ideally just tighten up the code

clients/privacy-center/cypress/e2e/consent-banner-tcf.cy.ts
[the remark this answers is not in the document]
NevilleS: OK, gotcha- I'm remembering this now... thanks

clients/privacy-center/cypress/e2e/consent-banner-tcf.cy.ts
```diff
@@ -723,6 +725,87 @@ describe("Fides-js TCF", () => {
...
+            }).then((xhr) => {
+              assert.isNull(xhr?.response?.body);
+            });
+          });
+          cy.wait("@mockSavePreferencesFn");
+          // The cookie should still get updated
+          cy.getCookie(CONSENT_COOKIE_NAME).then((cookie) => {
+            const cookieKeyConsent: FidesCookie = JSON.parse(
+              decodeURIComponent(cookie!.value)
+            );
```
NevilleS, line 772: These assertions feel like the wrong thing to test here- yes, the cookie should be updated but we don't need to fully interrogate it here, do we?

Instead, can you do some checks that the `savePreferencesFn` works as expected? Some suggestions:
* assert that the `fides_string` passed in matches what we'd expect it to be after clicking "opt out"
* assert that the `experience` passed in matches the one from the stubConfig
* etc

NevilleS (changes requested):
Few more nits and some bugs (e.g. lets ensure we are using a Promise )

clients/fides-js/src/lib/consent-types.ts
eastandwestwind: `FidesConfig` is the entire obj we use to init Fides in `Fides.init()`. I'm a little hesitant to pass in the `Fides` global as it's something we build during the init and state management could lead to inconsistencies in what we have in `Fides` vs in the cookie, depending on when we check for those vals. But I'll go ahead with refactoring to pass the `Fides` global in for now.

eastandwestwind: Running into some tricky bits here where the `Fides` global isn't updated until _after_ we initialize the overlay:

```
const updatedFides = await initialize({
    cookie,
    _Fides,
    experience,
    renderOverlay,
    updateCookie,
  });
  Object.assign(_Fides, updatedFides);
```

So if we pass it in, it won't be completely up to date, unless we do some refactoring. Additionally there are some cyclical import issues with the `Fides` type as it's defined in `initialize.ts` and it's not as simple as moving it to another types file.

PR #4328, opened 2023-10-23
GET Privacy Experience Meta Endpoint

pattisdr:
Closes https://github.com/ethyca/fidesplus/issues/1179

### Description Of Changes

Add a new GET Privacy Experience Meta endpoint that just returns the meta object for Experiences.  Content is really only built for TCF Experiences.  If non-TCF Experiences are queried, these objects just have null for the experience meta objects for now.  In the future, we could look into building a version hash for experiences with notices too. 

### Code Changes

Added new endpoint:
```
curl -X 'GET' \
  'http://localhost:8080/api/v1/privacy-experience-meta?region=us_ca&component=overlay&page=1&size=50' \
  -H 'accept: application/json'
```

Relevant query params are:
- region
- component
- page
- size

### Steps to Confirm

* [ ] See [doc with expected sample requests and responses](https://ethyca.atlassian.net/wiki/spaces/EN/pages/2820309077/2023-10-23+Backend+Design+Experience+Meta+Only+Endpoint?focusedCommentId=2820472929)
* [ ] Experiment with setting TCF Enabled/AC Enabled, neither, and make sure the meta content looks appropriate.  This is a public endpoint, no authentication required. It's the same content that is returned in GET Privacy Experience List, just segmented into a simpler endpoint.

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`


src/fides/api/api/v1/endpoints/privacy_experience_endpoints.py
```diff
@@ -64,6 +67,25 @@ def get_privacy_experience_or_error(
...
+    component_search_map: Dict = {
+        ComponentType.overlay: [ComponentType.overlay, ComponentType.tcf_overlay]
+    }
+    return experience_query.filter(
+        PrivacyExperience.component.in_(
+            component_search_map.get(component, [component])
+        )
+    )
+
+
```
pattisdr, line 88: This already existed: splitting out to use in my new endpoint

src/fides/api/api/v1/endpoints/privacy_experience_endpoints.py
```diff
@@ -119,6 +141,64 @@ def _filter_experiences_by_region_or_country(
...
+            break
+
+    results: List[PrivacyExperience] = []
+    for experience in experience_query:
+        if experience.component == ComponentType.tcf_overlay:
+            # Attach meta for TCF Experiences only.  We don't yet build meta info for non-TCF experiences.
+            experience.meta = tcf_meta
+        results.append(experience)
+
+    return fastapi_paginate(results, params=params)
```
adamsachs, line 199: at first glance, it struck me as a bit odd that we're paging and wrapping this within a `PrivacyExperience`, but the doc was helpful in understanding rationale, and i think it makes sense!

src/fides/api/api/v1/endpoints/privacy_experience_endpoints.py
```diff
@@ -119,6 +141,64 @@ def _filter_experiences_by_region_or_country(
...
+@fides_limiter.limit(CONFIG.security.public_request_rate_limit)
+async def get_privacy_experience_meta(
+    *,
+    db: Session = Depends(deps.get_db),
+    params: Params = Depends(),
+    region: Optional[str] = None,
+    component: Optional[ComponentType] = None,
+    request: Request,  # required for rate limiting
+    response: Response,  # required for rate limiting
+) -> AbstractPage[PrivacyExperienceMetaResponse]:
```
adamsachs, line 158: nit: i could be wrong, but isn't the return type annotation here technically `AbstractPage[PrivacyExperience]`? and then the `response_model` would be responsible for eliminating in the API response any additional fields on the returned `PrivacyExperience` that aren't part of the `PrivacyExperienceMetaResponse` model/schema?

it's a bit academic, and i may be wrong in understanding the functionality/intention! but i was scratching my head for a few seconds trying to reconcile the return type annotation and the objects actually being returned in the function.

adamsachs (approved):
@pattisdr nice, this looks good, and i think you'd be good to merge it as is!

just to raise the question - what would you think of moving this endpoint over to fidesplus? it's not a priority, so if you need/want to get this in as quick as possible, then you can ignore for now. but it seems like a relatively straightforward "wrapping" of underlying functionality that could possibly be moved over without serious rework required?

src/fides/api/api/v1/endpoints/privacy_experience_endpoints.py
pattisdr: Thank you for catching, it should be PrivacyExperience, this was leftover from when I was still toying with it being just PrivacyExperienceMetaResponse and not wrapped within the PrivacyExperience, but then I thought this should truly be a subset of the PrivacyExperience endpoint contents.

pattisdr:
Discussed - will merge and look into moving later -

Issue #4288, opened 2023-10-17
`systems-plus.cy.ts` has a flaky test


allisonking:
### Bug Description

There is a test in `systems-plus.cy.ts` called `can switch between tabs after populating from dictionary` which is flaky in CI. It sometimes passes but often does not.

#### Steps to Reproduce

* This can fail locally too (but for a different reason than the error in CI) if you throttle your cypress network activity to 3G
* Otherwise, it seems to only happen sometimes in CI, for example:
  * https://github.com/ethyca/fides/actions/runs/6548622848/job/17783709214?pr=4286
  * https://github.com/ethyca/fides/actions/runs/6547821364/job/17782622930?pr=4262#step:6:966

### Expected behavior

No flaky tests!

### Screenshots

See CI logs

### Environment

* Version:
* OS:
* Python Version:
* Docker Version:

### Additional context

https://ethyca.slack.com/archives/C04S7K0T2V7/p1697552900614439

allisonking:
It seems like now there are even more tests that fail—the first few in `systems-plus.cy.ts` also fail in CI, but not locally. Perhaps the page loads too slowly? Need to investigate.

PR #4347, opened 2023-10-26
Add initial state to suggestions

allisonking:
Closes https://ethyca.atlassian.net/browse/PROD-1255

### Description Of Changes

This was a very strange bug, and I think it was introduced by https://github.com/ethyca/fides/pull/4294#discussion_r1362890811 but even that seems to be a bit of a red herring... 

First I found that if I reverted that particular change described in that comment, things worked. I did a diff between `systemBody` and `result.data`, then added all the fields until I found the one that was causing things to break which was mysteriously `third_country_transfers`. There's no rhyme or reason to this as far as I can tell, so I dug further...

Then I realized the only fields that were being reset were the dictionary fields, so I looked specifically at `useDictSuggestion`.

It seems like this line https://github.com/ethyca/fides/blob/aking/prod-1255/fix-system-info-saving/clients/admin-ui/src/features/system/dictionary-form/DictSuggestionInputs.tsx/#L88-L92 was causing problems. The `initialValues` were properly set, but that bit of code would reset all those fields to empty strings so `values` would be incorrect. Again, this only happens when `third_country_transfers` is included, which is very strange, and I suspect is not entirely the root cause and may be a red herring (or a sign of something else).

By adding a new suggestion state `initial` we can get around this—that line will only fire when we actually want "presuggestions" which I think works pretty well.

https://github.com/ethyca/fides/assets/24641006/9d99385c-1b5b-4d14-994b-423a2d73f3df

### Code Changes

* [x] Add a new `initial` state to suggestions

### Steps to Confirm

* [ ] See steps in Roger's bug report.
* [ ] Make sure "pre suggestions" still work

### Pre-Merge Checklist

* [x] Update `CHANGELOG.md`

### DIFF
--- a/clients/admin-ui/cypress/e2e/systems-plus.cy.ts
+++ b/clients/admin-ui/cypress/e2e/systems-plus.cy.ts
@@ -27,8 +27,9 @@ describe("System management with Plus features", () => {
 
   describe("vendor list", () => {
     beforeEach(() => {
-      cy.visit(`${SYSTEM_ROUTE}/configure/demo_analytics_system`);
       stubVendorList();
+      cy.visit(`${SYSTEM_ROUTE}/configure/demo_analytics_system`);
+      cy.wait("@getDictionaryEntries");
     });
 
     it("can display the vendor list dropdown", () => {
--- a/clients/admin-ui/cypress/e2e/systems-plus.cy.ts
+++ b/clients/admin-ui/cypress/e2e/systems-plus.cy.ts
@@ -117,9 +117,9 @@ describe("System management with Plus features", () => {
           fixture: "taxonomy/custom-metadata/custom-field/list.json",
         }
       ).as("getCustomFields");
-      cy.intercept("PUT", `/api/v1/plus/custom-metadata/custom-field`, {
-        fixture: "taxonomy/custom-metadata/custom-field/update-party.json",
-      }).as("updateParty");
+      cy.intercept("POST", `/api/v1/plus/custom-metadata/custom-field/bulk`, {
+        body: {},
+      }).as("bulkUpdateCustomField");
     });
 
     it("can populate initial custom metadata", () => {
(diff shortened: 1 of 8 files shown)


allisonking:
If we are very lucky, perhaps this will fix https://github.com/ethyca/fides/issues/4288 too 🤞

allisonking:
Investigating CI failures which appear to be on `main` now...

TheAndrewJackson (approved):
Thanks for digging into this. It's very weird that that specific field is causing this issue. I'm surprised that this hasn't come up yet already.

The fix itself makes sense though!

:shipit:

clients/admin-ui/cypress/e2e/systems-plus.cy.ts
```diff
@@ -117,9 +117,9 @@ describe("System management with Plus features", () => {
           fixture: "taxonomy/custom-metadata/custom-field/list.json",
         }
       ).as("getCustomFields");
-      cy.intercept("PUT", `/api/v1/plus/custom-metadata/custom-field`, {
-        fixture: "taxonomy/custom-metadata/custom-field/update-party.json",
-      }).as("updateParty");
+      cy.intercept("POST", `/api/v1/plus/custom-metadata/custom-field/bulk`, {
+        body: {},
+      }).as("bulkUpdateCustomField");
```
allisonking, line 122: this is a test that needed to be updated after the changes in https://github.com/ethyca/fides/pull/4294/commits/ba7f66bcba991fa768e36d9f7120771598544fdd

clients/admin-ui/cypress/e2e/taxonomy-plus.cy.ts
```diff
@@ -228,33 +228,26 @@ describe("Taxonomy management with Plus features", () => {
...
-          expect(updatePartyInterception.request.body.resource_id).to.eql(
-            RESOURCE_CHILD.key
-          );
-          expect(updatePartyInterception.request.body.value).to.eql([
-            "Charmander",
-            "Eevee",
-          ]);
-        }
-      );
+      cy.wait("@bulkUpdateCustomField").then((interception) => {
```
allisonking, line 237: also a result of the new bulk update endpoint (which is a lot nicer than what we used to have!)

TheAndrewJackson (approved):

PR #4350, opened 2023-10-27
Handle decoding problems with passed in `fides_string`

allisonking:
Closes https://ethyca.atlassian.net/browse/PROD-1281

### Description Of Changes

Passing in an invalid `fides_string` will now ignore it instead of breaking the app!

In the video, both with and without a fides string give the same result.

https://github.com/ethyca/fides/assets/24641006/ad1103e5-ccb6-4451-aebe-8de27e59b648

### Code Changes

* [x] Add `try/catch` around decoding the tc string
* [x] If the decoding fails, reset the passed in `fides_string` to `null`

### Steps to Confirm

* [ ] See steps in the ticket

### Pre-Merge Checklist

* [x] Update `CHANGELOG.md`

### DIFF
--- a/clients/fides-js/src/fides-tcf.ts
+++ b/clients/fides-js/src/fides-tcf.ts
@@ -68,6 +68,7 @@ import {
 import type { Fides } from "./lib/initialize";
 import { dispatchFidesEvent } from "./lib/events";
 import {
+  buildTcfEntitiesFromCookie,
   debugLog,
   experienceIsValid,
   FidesCookie,
@@ -121,13 +122,43 @@ const getInitialPreference = (
   return tcfObject.default_preference ?? UserConsentPreference.OPT_OUT;
 };
 
-const updateCookie = async (
-  oldCookie: FidesCookie,
-  experience: PrivacyExperience
-): Promise<FidesCookie> => {
-  // ignore server-side prefs if either user has no prefs to override, or TC str override is set
+const updateCookieAndExperience = async ({
+  cookie,
+  experience,
+  debug = false,
+  hasFidesStringOverride,
+  isExperienceClientSideFetched,
+}: {
+  cookie: FidesCookie;
+  experience: PrivacyExperience;
+  debug?: boolean;
+  hasFidesStringOverride: boolean;
+  isExperienceClientSideFetched: boolean;
+}): Promise<{
+  cookie: FidesCookie;
+  experience: Partial<PrivacyExperience>;
+}> => {
+  // If string override exists, the cookie is already okay, but the experience may not be.
+  if (hasFidesStringOverride) {
+    // If we didn't get the experience until the client side fetch, we need to update the fetched
+    // experience based on the cookie here.
+    if (isExperienceClientSideFetched) {
+      debugLog(
+        debug,
+        "Overriding preferences from client-side fetched experience with cookie fides_string consent",
+        cookie.fides_string
+      );
+      const tcfEntities = buildTcfEntitiesFromCookie(experience, cookie);
+      return { cookie, experience: tcfEntities };
+    }
+    // If it's not client side fetched, we don't update anything since the cookie has already
+    // been updated earlier.
+    return { cookie, experience };
+  }
+
+  // ignore server-side prefs if user has no prefs to override
   if (!hasSavedTcfPreferences(experience)) {
-    return { ...oldCookie, fides_string: "" };
+    return { cookie: { ...cookie, fides_string: "" }, experience };
   }
 
   const tcSavePrefs: TcfSavePreferences = {};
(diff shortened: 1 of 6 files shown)


clients/fides-js/src/fides-tcf.ts
```diff
@@ -184,11 +184,18 @@ const init = async (config: FidesConfig) => {
...
       config.options.debug
     );
+    if (!success) {
+      // Treat an unsuccessful transform as if no fidesString were passed in.
+      // eslint-disable-next-line no-param-reassign
+      config.options.fidesString = null;
+    } else {
+      cookie.tcf_consent = cookieKeys;
+      cookie.fides_string = config.options.fidesString;
+    }
```
NevilleS, line 198: Don't we need a similar type of fallback if we fail to decode the data from the `fides_consent` cookie or from the `experience`?

NevilleS (changes requested):
This looks solid, but I have a question in there...

clients/fides-js/src/fides-tcf.ts
allisonking: this was the only place I found that does a call to `TCString.decode`—I don't think we directly parse strings from the cookie or from the experience (we typically use the `cookie.tcf_consent` obj instead)

NevilleS: OK 👍 

I think that'll change pretty quickly, I'm planning to have us load `fides_string`s from API calls promptly. So I think we should plan for that and just refactor this a bit... how about this:
```
if (config.options.fidesString) {
  updateFidesCookieFromString(cookie, config.options.fidesString);
}
```

...and then encapsulate the error handling in that helper

clients/fides-js/src/fides-tcf.ts
```diff
@@ -184,11 +184,18 @@ const init = async (config: FidesConfig) => {
...
-    cookie.fides_string = config.options.fidesString;
-    cookie.tcf_consent = transformFidesStringToCookieKeys(
+    const { cookieKeys, success } = transformFidesStringToCookieKeys(
       config.options.fidesString,
       config.options.debug
     );
+    if (!success) {
+      // Treat an unsuccessful transform as if no fidesString were passed in.
+      // eslint-disable-next-line no-param-reassign
+      config.options.fidesString = null;
```
NevilleS, line 194: I think `eslint` is right here... modifying that argument feels confusing to me, I'd want to be able to _remember_ that there was an override string provided.

Why do we need to set it to `null` here, is it being re-read again later?

NevilleS (changes requested):
Thanks for clarifying - OK, some nits

clients/fides-js/src/fides-tcf.ts
allisonking: yeah that's the route I went first, but it turned out it was being used [here](https://github.com/ethyca/fides/blob/main/clients/fides-js/src/lib/initialize.ts#L315) as well, which right now just checks if a `fidesString` exists, but really needs to also know that the string is valid. we could have it check if it's valid instead, but I wasn't sure if calling decode again would eat into the initialization time... thoughts?

allisonking: I'll see if I can refactor the other function that uses `fidesString` 👍

allisonking:
Okay @NevilleS I've refactored, hopefully it is a little more streamlined now. thank goodness for all the tests you and @eastandwestwind put in 🙏

NevilleS (changes requested):
This works, but I feel like we've hit a level of complexity here that gives me pause- let's take a beat right now and see what we could do to simplify this!

allisonking:
We talked about this separately and decided:
* the initialization logic as a whole is too complicated!
* this PR makes it marginally more understandable
* let's make a separate ticket to simplify the initialization logic
* this PR can continue down its release process (I tried to poke at it more to simplify it, but it ended up requiring larger changes I'd rather do as part of an initialization refactor)

clients/fides-js/src/fides-tcf.ts
```diff
@@ -177,18 +239,18 @@ const init = async (config: FidesConfig) => {
...
-    cookie.tcf_consent = transformFidesStringToCookieKeys(
+    const { cookie: updatedCookie, success } = updateFidesCookieFromString(
+      cookie,
       config.options.fidesString,
       config.options.debug
     );
+    if (success) {
+      Object.assign(cookie, updatedCookie);
+    } else {
+      hasFidesStringOverride = false;
```
eastandwestwind, line 252: this functionally works for now, but perhaps a better name would be `hasValidFidesStringOverride`?

clients/fides-js/src/fides-tcf.ts
```diff
@@ -121,13 +122,43 @@ const getInitialPreference = (
...
+  cookie: FidesCookie;
+  experience: PrivacyExperience;
+  debug?: boolean;
+  hasFidesStringOverride: boolean;
+  isExperienceClientSideFetched: boolean;
+}): Promise<{
+  cookie: FidesCookie;
+  experience: Partial<PrivacyExperience>;
+}> => {
+  // If string override exists, the cookie is already okay, but the experience may not be.
```
eastandwestwind, line 141:
```suggestion
  // If string override exists and is valid, the cookie has already been overridden, but the experience has not.
```

clients/fides-js/src/fides-tcf.ts
```diff
@@ -121,13 +122,43 @@ const getInitialPreference = (
...
+      );
+      const tcfEntities = buildTcfEntitiesFromCookie(experience, cookie);
+      return { cookie, experience: tcfEntities };
+    }
+    // If it's not client side fetched, we don't update anything since the cookie has already
+    // been updated earlier.
+    return { cookie, experience };
+  }
+
+  // ignore server-side prefs if user has no prefs to override
```
eastandwestwind, line 159: I know I wrote this code comment, but realizing this could be clearer...

```suggestion
  // If user has no prefs saved, we don't need to override the prefs on the cookie
```

clients/privacy-center/cypress/e2e/consent-banner-tcf.cy.ts
```diff
@@ -1934,6 +1934,105 @@ describe("Fides-js TCF", () => {
...
+     * TEST CASE #8:
+     * 😬 1) fides_string override option exists but is invalid (via config.options.fidesString)
+     * ❌ 2) DEFER: preferences API (via a custom function)
+     * ❌ 3) local cookie (via fides_consent cookie)
+     * ❌ 4) "prefetched" experience (via config.options.experience)
+     * ✅ 5) experience API (via GET /privacy-experience)
+     *
+     * EXPECTED RESULT: ignore invalid fides_string option and render experience as-is
+     */
+    it("can handle an invalid fides_string option and continue rendering the experience", () => {
```
eastandwestwind, line 1948: thanks for adding these tests!

eastandwestwind (approved):
Overall looks good, nothing blocking, just a couple minor suggestions 🚀

NevilleS (approved):
Thanks for the extra TLC on this one 👍

PR #4356, opened 2023-10-30
Resurface TCF consent (frontend)

allisonking:
Closes https://ethyca.atlassian.net/browse/PROD-1281

### Description Of Changes

Rely on the backend calculated `version_hash` to determine whether or not to resurface the banner.

https://www.loom.com/share/0a3d388b388543428c1b75ede2184e21?sid=44f41194-e921-4217-b42c-70185d4c44c0

### Code Changes

* [x] Autogenerate TS types
* [x] Rework logic to surface the banner (and rename the function to `resurfaceConsent`) to compare TCF hashes and save the hash in our cookie
* [x] Remove redundant `showBanner` check
* [x] Update cypress tests
* [x] Add a `mockCookie` function to help with testing

### Steps to Confirm
See loom video for an example

* [ ] Set up TCF and save your preferences
* [ ] You should see a new field `tcf_version_hash` in your cookie
* [ ] In admin-ui, add another system with TCF properties
* [ ] Refresh your TCF page
* [ ] The banner should resurface
* [ ] Save your preferences again
* [ ] Refresh the page
* [ ] The banner should not resurface

### Pre-Merge Checklist

* [x] Update `CHANGELOG.md`

### DIFF
--- a/clients/admin-ui/src/types/api/models/ExperienceConfigCreate.ts
+++ b/clients/admin-ui/src/types/api/models/ExperienceConfigCreate.ts
@@ -35,7 +35,7 @@ export type ExperienceConfigCreate = {
    */
   privacy_policy_link_label?: string;
   /**
-   * Overlay and Privacy Center 'Privacy policy URl'
+   * Overlay and Privacy Center 'Privacy policy URL
    */
   privacy_policy_url?: string;
   /**
--- a/clients/fides-js/src/components/ConsentBanner.tsx
+++ b/clients/fides-js/src/components/ConsentBanner.tsx
@@ -54,7 +54,7 @@ const ConsentBanner: FunctionComponent<BannerProps> = ({
     if (bannerIsOpen) {
       onOpen();
     }
-  }, [bannerIsOpen]);
+  }, [bannerIsOpen, onOpen]);
 
   return (
     <div
--- a/clients/fides-js/src/components/notices/NoticeOverlay.tsx
+++ b/clients/fides-js/src/components/notices/NoticeOverlay.tsx
@@ -13,7 +13,7 @@ import ConsentBanner from "../ConsentBanner";
 import { updateConsentPreferences } from "../../lib/preferences";
 import {
   debugLog,
-  hasActionNeededNotices,
+  resurfaceConsent,
   transformConsentToFidesUserPreference,
 } from "../../lib/consent-utils";
 
@@ -43,7 +43,7 @@ const NoticeOverlay: FunctionComponent<OverlayProps> = ({
   >(initialEnabledNoticeKeys);
 
   const showBanner = useMemo(
-    () => experience.show_banner && hasActionNeededNotices(experience),
+    () => experience.show_banner && resurfaceConsent(experience),
     [experience]
   );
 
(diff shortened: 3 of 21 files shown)


clients/fides-js/src/components/ConsentBanner.tsx
```diff
@@ -54,7 +54,7 @@ const ConsentBanner: FunctionComponent<BannerProps> = ({
     if (bannerIsOpen) {
       onOpen();
     }
-  }, [bannerIsOpen]);
+  }, [bannerIsOpen, onOpen]);
```
allisonking, line 57: also noticed a few places where the dependency array wasn't complete

clients/fides-js/src/lib/consent-utils.ts
```diff
@@ -242,17 +234,19 @@ export const hasSavedTcfPreferences = (experience: PrivacyExperience) =>
...
-    return hasActionNeededTcfPreferences(experience);
+export const resurfaceConsent = (
+  experience: PrivacyExperience,
+  cookie: FidesCookie
+) => {
+  if (
+    experience.component === ComponentType.TCF_OVERLAY &&
+    experience.meta?.version_hash
+  ) {
+    return experience.meta.version_hash !== cookie.tcf_version_hash;
```
allisonking, line 249: this is the crux of the change

clients/fides-js/src/components/ConsentBanner.tsx
allisonking: this _does_ mean that the `onOpen` function needs to be wrapped in a `useCallback` in order to avoid infinite renders!

clients/fides-js/src/components/notices/NoticeOverlay.tsx
```diff
@@ -125,56 +119,52 @@ const NoticeOverlay: FunctionComponent<OverlayProps> = ({
...
   );
 
-  if (!experience.experience_config) {
-    debugLog(options.debug, "No experience config found");
-    return null;
-  }
-  const experienceConfig = experience.experience_config;
-
-  const dispatchOpenBannerEvent = () => {
+  const dispatchOpenBannerEvent = useCallback(() => {
```
allisonking, line 122: see https://github.com/ethyca/fides/pull/4356/files#r1376843489 for why we wrap it in a `useCallback`. we avoided the infinite renders before by not including `onOpen` in the dependency array, but that's not a good practice

clients/fides-js/src/lib/consent-utils.ts
eastandwestwind: on the BE, have all experiences been migrated to have a `version_hash`? Does this mean that for users that have a Fides consent cookie on their browser, when they navigate to a site that has Fides installed, they will get the banner popup again?

allisonking: good point! this is only for TCF—the logic for notice banners is still the same as before. the backend doesn't have version hashes for notice based experiences yet. So I think we are safe here (since no one is using TCF yet)

eastandwestwind: cool this makes sense, just making sure it's something we've thought about 👍

eastandwestwind (approved):

clients/admin-ui/src/types/api/models/ExperienceConfigCreate.ts
```diff
@@ -35,7 +35,7 @@ export type ExperienceConfigCreate = {
    */
   privacy_policy_link_label?: string;
   /**
-   * Overlay and Privacy Center 'Privacy policy URl'
+   * Overlay and Privacy Center 'Privacy policy URL
```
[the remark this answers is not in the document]
allisonking: womp this was autogenerated from the backend so we'd need to fix it on the backend...

Issue #3171, opened 2023-04-26
Add error-handling and versioned logs to Privacy Center to help admins troubleshoot configuration issues


NevilleS:
### Description
Today, any misconfiguration of the `config.json` file for the Privacy Center tends to result in a mysterious server crash when it starts. With https://github.com/ethyca/fides/pull/3105 we've retooled the configuration to load at runtime, which gives us a lot more leeway to provide some helpful breadcrumbs for admins to troubleshoot!

We don't need to make this bullet-proof, but a bit of error-handling, validation, and helpful log messages would go a long way.

### Acceptance Criteria
- [ ] Log a version number in the `privacy-center` server logs
- [ ] Initialize the Redux store with a safe set of default configuration, but with a "customized = false" state
- [ ] Add a helpful warning in the UI that the configuration hasn't been customized and isn't ready to receive requests 👍 
- [ ] Validate the `config.json`, either on startup or on page load
- [ ] Log helpful validation errors for admins to troubleshoot
- [ ] Show an error page in the UI when configuration errors occur (to avoid confusing end-user UX)
- [ ] Cleanup any other miscellaneous TODOs from the https://github.com/ethyca/fides/pull/3105 refactor

### Implementation Notes
From @TheAndrewJackson:
> [Zod](https://github.com/colinhacks/zod) can be used to validate the config. It's essentially pydantic but for typescript. It's super powerful!
> 
> This might helpful too. There is a package called [t3-env](https://github.com/t3-oss/t3-env) that will uses a zod schema to validate env vars on startup. If any are missing or invalid it will fail startup and let the user know what's missing.

That looks interesting...!

TheAndrewJackson:
copy/pasta of my comment from the PR keep the convo closer to the issue:

-----

Zod can be used to validate the config. It's essentially pydantic but for typescript. It's super powerful! 

This might helpful too. There is a package called t3-env that will uses a zod schema to validate env vars on startup. If any are missing or invalid it will fail the startup and let the user know what's missing.

rsilvery:
@mfbrown  is this captured somewhere with our DSR roadmap work?

PR #4370, opened 2023-10-31
Adding default_value and hidden properties to custom privacy request fields

galvana:
### Description Of Changes

Adding the ability to support hidden custom privacy request fields. This requires the addition of a default value so I added two new properties `default_value` and `hidden`.

- The default value for a hidden field won't be shown in the form, but it will be included in the privacy request payload
- The default value for a non-hidden field will be shown in the form and can be overwritten by the user

I also limited the height of the privacy request form field and added scrolling.

### Code Changes

* [ ] Updated `smoke_test.cy.ts`
* [ ] Added validation to error fields flagged as hidden without a default value
* [ ] Updated styling on the form
* [ ] Updated form logic to account for the new field properties

### Steps to Confirm

* [x] Enable custom privacy request field collection https://docs.ethyca.com/dev-docs/configuration/privacy-requests/custom-request-fields#enable-custom-fields
* [x] Run `nox -s "fides_env(test)"`
* [x] Navigate to the Privacy Center
* [x] Fill out the form for an access request
* [x] Navigate to the Admin UI
* [ ] Verify the Tenant ID is included even though it wasn't shown to the user in the access request form

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`


clients/cypress-e2e/cypress/e2e/smoke_test.cy.ts
```diff
@@ -12,16 +12,49 @@ describe("Smoke test", () => {
...
+              value: "",
+            },
+            color: {
+              label: "Color",
+              value: "blue",
+            },
+            tenant_id: {
+              label: "Tenant ID",
+              value: "123",
+            },
```
galvana, line 51: Verifying a few scenarios with this check

- A non-required field still contains a value, even if it's an empty string
- A field with a `default_value` can be overwritten
- A hidden field with a `default_value` is still included with the other form values in the request

galvana, line 51: Verifying a few scenarios with this check

- A non-required field still contains a value, even if it's an empty string
- A field with a `default_value` can be overwritten in the form
- A hidden field with a `default_value` is still included with the entered form values in the request

clients/privacy-center/app/server-environment.ts
```diff
@@ -178,6 +178,21 @@ export const validateConfig = (
...
   }
+
+  // Validate that each hidden field in custom_privacy_request_fields has a default value
+  config.actions.forEach((action) => {
+    Object.entries(action.custom_privacy_request_fields || {}).forEach(
+      ([key, field]) => {
+        if (field.hidden && field.default_value === undefined) {
+          return {
+            isValid: false,
+            message: `Default value required for hidden field '${key}' in action '${action.policy_key}'`,
```
galvana, line 189: This error message doesn't surface completely, but I see that's a known issue https://github.com/ethyca/fides/issues/3171

clients/privacy-center/components/modals/privacy-request-modal/PrivacyRequestForm.tsx
```diff
@@ -273,11 +273,11 @@ const PrivacyRequestForm: React.FC<PrivacyRequestFormProps> = ({
...
       </ModalHeader>
+      <Text fontSize="sm" color="gray.600" mb={4} ml={6}>
+        {action.description}
+      </Text>
       <chakra.form onSubmit={handleSubmit} data-testid="privacy-request-form">
-        <ModalBody>
-          <Text fontSize="sm" color="gray.600" mb={4}>
-            {action.description}
-          </Text>
+        <ModalBody maxHeight={400} overflowY="auto">
```
galvana, line 280: Updating the popup to scroll if the form body exceeds 400px

clients/privacy-center/app/server-environment.ts
allisonking: this doesn't appear to be working—I set a value to `hidden` without a `default_value` but it still says `isValid = true`! 

I think `forEach` `return`s don't work quite as we expect them to here: https://stackoverflow.com/questions/56129153/how-to-return-a-value-from-foreach

allisonking (changes requested):
confirmed happy path works! 
image

I don't think the validation is working quite right though

allisonking (approved):
🙌

PR #4374, opened 2023-11-02
Adding support for configuring enabled actions

galvana:
### Description Of Changes

⚠️ Requires the corresponding branch in fidesplus to test https://github.com/ethyca/fidesplus/pull/1205

Adding the ability to specify which privacy actions (access, erasure, consent) to enable for a system integration. The new `Enabled actions` field is a multi-select with the options being the actions supported by the integration.

### Code Changes

* [ ] Updated the `ConnectorParamaters` component to call the fidesplus versions of the system integration endpoints if fidesplus is available
* [ ] Updated the `ConnectorParameterForm` to conditionally render the `Enabled actions` field if fidesplus is enabled AND there is more than one action to choose from. The option to disable an action won't be available if it's the only action.
* [ ] Added fidesplus versions of `patchSystemConnectionConfigs` and `createSaasConnectionConfig` to `plus.slice.ts`
* [ ] Updated select models with `supported_actions` and `enabled_actions` fields
* [ ] Refactored the `has_access`, `has_erasure`, `has_consent` logic in `connection_type.py` and moved it to `saas_config.py`
* [ ] Minor cleanup of the `connector_registry_service.py`
* [ ] Minor field name fixes (`SSH required`)

### Steps to Confirm

* [ ] Start fidesplus
* [ ] Start the Admin UI
* [ ] Create a new system
* [ ] Select a new integration from the integration search
* [ ] Note the default values for the `Enabled actions`
  * [ ] Database connectors should have `access` and `erasure`
  * [ ] SaaS connectors will vary based on the supported actions
  * [ ] Not available for email connectors since they only support one action, either `erasure` or `consent`
  * [ ] Manual connectors should have `access` and `erasure`
* [ ] For the integrations with the `Enabled actions` field enabled, make a change and verify the change persists
* [ ] Shut down fidesplus and start fides, verify the field is not available for any of the broad connector types (DB, SaaS, email, manual)

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`


src/fides/api/service/connectors/saas/connector_registry_service.py
```diff
@@ -70,15 +70,15 @@ def _load_connector_templates(self) -> None:
...
-                connector_type = config_dict["type"]
-                human_readable = config_dict["name"]
-                user_guide = config_dict.get("user_guide")
-                authentication = config_dict["client_config"].get("authentication")
+                config_path = os.path.join("data/saas/config", file)
+                config = SaaSConfig(**load_config(config_path))
+
+                connector_type = config.type
+
+                authentication = config.client_config.authentication
```
galvana, line 78: Refactoring to use a `SaaSConfig` instead of a `Dict` this way it's more consistent with the way it's done in other places in this file.

src/fides/api/util/connection_type.py
```diff
@@ -94,38 +94,9 @@ def saas_request_type_filter(connection_type: str) -> bool:
...
-                )
-            )
-            or saas_config.data_protection_request
-        )
-        has_consent = saas_config.consent_requests
-
-        return bool(
-            (ActionType.consent in action_types and has_consent)
-            or (ActionType.access in action_types and has_access)
-            or (ActionType.erasure in action_types and has_erasure)
```
galvana, line 128: This logic was moved to the `SaaSConfig` https://github.com/ethyca/fides/pull/4374/files#diff-e02882cee5e6ee2325e4d613abed561b386e3e15064d0a1cbb41785a5d3f6d1aR483

src/fides/api/util/connection_type.py
```diff
@@ -157,6 +128,7 @@ def saas_request_type_filter(connection_type: str) -> bool:
                     identifier=item,
                     type=SystemType.database,
                     human_readable=ConnectionType(item).human_readable,
+                    supported_actions=[ActionType.access, ActionType.erasure],
```
galvana, line 131: DB connectors will support access and erasure by default

tests/ops/api/v1/endpoints/test_connection_template_endpoints.py
```diff
@@ -1005,7 +1005,7 @@ def test_get_connection_secret_schema_mysql(
                     "type": "string",
                 },
                 "ssh_required": {
-                    "title": "SSH Required",
+                    "title": "SSH required",
```
galvana, line 1008: Changing `SSH required` to use sentence-case to be consistent with the other field labels.

src/fides/api/service/connectors/saas/connector_registry_service.py
```diff
@@ -339,6 +341,10 @@ def create_connection_config_from_template_no_save(
     if template_values.name:
         data["name"] = template_values.name
 
+    enabled_actions = getattr(template_values, "enabled_actions", None)
+    if enabled_actions is not None:
+        data["enabled_actions"] = enabled_actions
```
galvana, line 346: This function is used by both fides and fidesplus endpoints so we need to perform this attribute check to map the `enabled_actions` attribute which is only found in fidesplus.

clients/admin-ui/src/types/api/models/ConnectionConfigurationResponse.ts
galvana, line 1: These two model changes were generated by `openapi:generate` but I didn't keep the other changes to avoid a huge diff.

src/fides/api/service/connectors/saas/connector_registry_service.py
adamsachs: yeah, this is an interesting pattern. i can't say i _love_ it, it feels like a development footgun moving forward - we're implying (in a pretty cryptic manner) that someone developing here in fides knows about something defined over in fidesplus. i'm wouldn't be surprised if we do that in some other places already, but my gut-instinct tells me it's something we should try to avoid if possible. 

i'll make a broader comment on the PR at this point, since i think this is a holistic concern - not tied to this particular code block, even though this is where it's manifested

tests/ops/api/v1/endpoints/test_connection_template_endpoints.py
adamsachs: how dare you!

src/fides/api/util/connection_type.py
adamsachs: nice! like the refactor here 👍

src/fides/api/service/connectors/saas/connector_registry_service.py
adamsachs: that seems good 👍

adamsachs (commented):
@galvana broadly this looks good, i just get concerned with any sort of reverse dependency of OSS on plus. i know it's not a hard dependency, but see my inline comment for the concern - basically it's a maintainability/development concern. having an expectation within the BE OSS codebase that we need to be aware of things defined in plus sets a bad precedent. we may do it in other places, but i don't think it should be encouraged.

i know we're trying to layer in a plus-only feature on top of a lot of OSS plumbing that i'm sure needs to stay in OSS, or at least can't be quickly moved out. so if we need to do this here, i'm not going to totally hold this up. i know we're not purists on this, nor should we be - i just want to see if there's anything we can think of to untangle this. 

what's first popping into my head - how about some sort of override/extension of `create_connection_config_from_template_no_save` in plus that has the enabled action handling tacked on top of the rest of the function defined in OSS? basically just keeping all `enabled_action` logic in plus...

happy to talk this one over. maybe my concern's misplaced/overblown. in any case, i don't think this path forward is totally unviable -- if we need to do it, we can. it just adds some tech debt and feels to me like an antipattern we should strive to avoid, if we can.

adamsachs (commented):
ok from a BE perspective this is looking good to me! i had a few comments on the plus PR of things you may look to tweak, but i don't think any of those would require updates here.

a couple of notes that we just need to consider before merging:
- reminder to update the readme :)
- i haven't done any code review on the FE pieces, so it may be good to have a FE dev specifically review those?
- i haven't gotten a chance to perform any UAT on this - and that feels like that'd probably be a good idea, particularly if there hasn't been any FE code review! i can get to that tomorrow if that's OK?

allisonking:
taking a look! getting an error once I select an integration: 
image

allisonking:
> taking a look! getting an error once I select an integration: image

Never mind, I think I needed to run fidesplus with `dev_prerelease` to get this working!

clients/admin-ui/src/features/datastore-connections/system_portal_config/forms/ConnectorParametersForm.tsx
```diff
@@ -398,6 +410,71 @@ const ConnectorParametersForm: React.FC<ConnectorParametersFormProps> = ({
                     }
                   )
                 : null}
+              {isPlusEnabled &&
+                connectionOption.supported_actions.length > 1 && (
+                  <Field
+                    id="enabled_actions"
+                    name="enabled_actions"
+                    validate={(value: string[]) => {
```
allisonking, line 418: normally it'd be better to do this as part of a validation schema, but I don't think this form has one... it does validation pretty oddly, so this is in line with the component, just making a note!

allisonking (approved):
this is looking good! the only thing that might be odd is when I try to autogenerate TS types off of your fidesplus branch, the `enabled_actions` field gets deleted from `CreateConnectionConfigurationWithSecrets`—is that expected?

galvana:
> this is looking good! the only thing that might be odd is when I try to autogenerate TS types off of your fidesplus branch, the `enabled_actions` field gets deleted from `CreateConnectionConfigurationWithSecrets`—is that expected?

Yes @allisonking, that's expected, `enabled_actions` was moved to the fidesplus model `CreateConnectionConfigurationWithSecretsExtended`

adamsachs:
UAT tested this piece on RC, looks good! still haven't yet confirmed the _effect_ of enabled actions but the form functionality  (including saving/persisting changes) looks good 👍

PR #4375, opened 2023-11-02
PROD-1241 Adds support for custom get preferences fn

eastandwestwind:
Closes https://ethyca.atlassian.net/browse/PROD-1241

### Description Of Changes

Adds support for custom get preferences fn

### Code Changes

* [ ] Implements custom get preferences fn
* [ ] Adds test coverage

### Steps to Confirm

* [ ] Run regression testing on TCF banner / modal.

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### DIFF
--- a/clients/fides-js/src/fides-tcf.ts
+++ b/clients/fides-js/src/fides-tcf.ts
@@ -52,8 +52,7 @@ import { shopify } from "./integrations/shopify";
 
 import {
   FidesConfig,
-  FidesOptionOverrides,
-  GetPreferencesFnResp,
+  FidesOverrides,
   OverrideOptions,
   PrivacyExperience,
   UserConsentPreference,
@@ -63,7 +62,7 @@ import { generateFidesString, initializeCmpApi } from "./lib/tcf";
 import {
   getInitialCookie,
   getInitialFides,
-  getOverrideFidesOptions,
+  getOverrides,
   initialize,
 } from "./lib/initialize";
 import type { Fides } from "./lib/initialize";
@@ -90,7 +89,6 @@ import {
   generateFidesStringFromCookieTcfConsent,
   transformFidesStringToCookieKeys,
 } from "./lib/tcf/utils";
-import { getConsentPreferences } from "./lib/preferences";
 
 declare global {
   interface Window {
@@ -219,16 +217,13 @@ const updateFidesCookieFromString = (
   );
   try {
     const cookieKeys = transformFidesStringToCookieKeys(fidesString, debug);
-    const updatedCookie = {
-      ...cookie,
-      tcf_consent: cookieKeys,
-      fides_string: fidesString,
-    };
-    if (fidesStringVersionHash) {
-      updatedCookie.tcf_version_hash = fidesStringVersionHash;
-    }
     return {
-      cookie: updatedCookie,
+      cookie: {
+        ...cookie,
+        tcf_consent: cookieKeys,
+        fides_string: fidesString,
+        tcf_version_hash: fidesStringVersionHash ?? cookie.tcf_version_hash,
+      },
       success: true,
     };
   } catch (error) {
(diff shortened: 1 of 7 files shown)


clients/fides-js/src/fides-tcf.ts
```diff
@@ -202,22 +204,31 @@ const updateCookieAndExperience = async ({
...
+      ...cookie,
+      tcf_consent: cookieKeys,
+      fides_string: fidesString,
+    };
+    if (fidesStringVersionHash) {
+      updatedCookie.tcf_version_hash = fidesStringVersionHash;
+    }
     return {
-      cookie: { ...cookie, tcf_consent: cookieKeys, fides_string: fidesString },
+      cookie: updatedCookie,
```
allisonking, line 231: rather than this `if`, I think it might be cleaner and work just as well to always set `fidesStringVersionHash`. at worst it's undefined, which is ok since we just compare to the experience version hash which [will be defined when we compare it](https://github.com/ethyca/fides/blob/PROD-1241/clients/fides-js/src/lib/consent-utils.ts/#L247-L249). that makes this bit of code cleaner if we can keep it as...

```typescript
return {
  cookie: {...cookie, tcf_consent: cookieKeys, fides_string: fidesString, tcf_version_hash: fidesStringVersionHash
}
```
mostly thinking of ways to keep fewer branches in our initialization code!

clients/fides-js/src/fides-tcf.ts
```diff
@@ -235,15 +246,25 @@ const updateFidesCookieFromString = (
 const init = async (config: FidesConfig) => {
   const overrideOptions: Partial<FidesOptionOverrides> =
     getOverrideFidesOptions();
+  const customPrefs: GetPreferencesFnResp | null = await getConsentPreferences(
```
allisonking, line 249: again, thinking of streamlining our initialization logic...

these are technically override options, aren't they? can we call this in `getOverrideFidesOptions`? it's hard to follow when these `customPrefs` are being applied in 3 different places in this function. 

in other words, is there a way to call `getConsentPreferences` in `getOverrideFidesOptions` and override all the fields we need in there before the rest of the function runs?

eastandwestwind: I haven't included them in `getOverrideFidesOptions` because a `getConsentPreferences` fn can only be provided in the `init` and not through all of the following: window obj, cookie, query param. So they're a bit different. 

> is there a way to call getConsentPreferences in getOverrideFidesOptions and override all the fields we need in there before the rest of the function runs?

I actually tried this route initially, but since we obtain the values we need to override later in the init lifecycle (e.g. `cookie.tcf_version_hash`, we have to pass down the return vals from `getConsentPreferences`, and cannot override all fields immediately

clients/fides-js/src/fides-tcf.ts
eastandwestwind: hmm won't this potentially overwrite a previously-saved `tcf_version_hash` from the cookie? E.g. scenario: A new user visits a site with Fides. In this site's config, there is no custom get preferences fn configured. The user saves their preferences, which writes the cookie to reflect the user preferences and `tcf_version_hash` from the `experience`. Next time they visit the site, in the init of Fides, their browser cookie's `tcf_version_hash` gets set to `undefined`, since there is no `fidesStringVersionHash`. This causes `shouldResurfaceConsent` to return `true`, which would mean the banner will keep resurfacing for the user.

Is that right, or am I missing something?

I can keep the code more minimal by doing something like this:

```
return {
  cookie: {...cookie, tcf_consent: cookieKeys, fides_string: fidesString, tcf_version_hash: fidesStringVersionHash ?? cookie.tcf_version_hash
}
```

allisonking: oh yeah, that's a good point! I think your suggested change makes sense 👍

clients/fides-js/src/fides-tcf.ts
allisonking: hmm okay, thinking out loud here... so there are 3 fields at play here:
### `consent`
This needs to override part of `cookie`. It's been a while since I looked at this code, but I think if the `config` object has a `consent` object, `getInitialCookie` will just use that (through `makeConsentDefaultsLegacy`)? So instead of 

```js
if (customPrefs?.consent) {
    cookie.consent = customPrefs.consent;
}
```
I think `getOverrideFidesOptions` could also overwrite `cookie.consent` and that'll propagate down okay. I know the override func only has `Partial<FidesOptionOverrides>` right now though... if you're against changing that, I feel like we could still maybe take advantage of the `getInitialCookie` being willing to take a `consent` obj.

### `fides_string`
This can fill in `overrideOptions.fidesString` (and indeed does so right on the next line) so it feels reasonable to me to put it with `getOverrideFidesOptions` 

### `version_hash`
I think `tcf_version_hash` can be written whenever—the only time we write to it is when we save an experience. so there's no other part in the initialization logic that that writes to `tcf_version_hash`

eastandwestwind: So it ended up being easier than I thought to go ahead and update `getOverrideFidesOptions = async(config: FidesConfig): Promise<Partial<FidesOptionOverrides>>` to include all overrides, like:  `getOverrides = async(config: FidesConfig): Promise<Partial<FidesOverrides>>

Thanks for the suggestions!

allisonking (approved):
yay, thanks for trying those changes out! so glad it worked out, things read much cleaner now 💥

Issue #4190, opened 2023-09-29
Refactor TCF vendor tab to use a table


allisonking:
### Is your feature request related to a specific problem?

The new vendor tab design features more toggles and labels, and would make more sense as a table than our current list structure.

### Describe the solution you'd like

We should refactor the `TcfVendors` component to actually use `<table />`. This would solve a number of hacky CSS solutions.

### Describe alternatives you've considered, if any

_A description of any alternative solutions or features you've considered._

### Additional context

https://github.com/ethyca/fides/pull/4189
https://www.figma.com/file/dmEwdK3xZwjKfGVQ9t66xe/Fides-v.2-Master-Working-Files?type=design&node-id=10555-605238&mode=dev

PR #4376, opened 2023-11-02
Various TCF fixes for vendor/purpose double toggle logic

allisonking:
Closes https://ethyca.atlassian.net/browse/PROD-1212
Closes https://github.com/ethyca/fides/issues/4190 since we no longer have a table view going on!

### Description Of Changes

https://www.loom.com/share/2b5f25c3043b43d4b74b8cdfe989040d?sid=3f9eebd6-ea1d-4ece-8633-22999914e97e

This also brings our bundle size from 42.67 KB to 39.67 KB 😌 

### Code Changes

* [x] Refactor DoubleToggleTable to instead be a `RecordsList` which all the tab components can share
* [x] Make everyone use `RecordsList`
* [x] Consent/legitimate interest radio toggle
* [x] Split vendors into IAB list and other list
* [x] Update toggle styling
* [x] Remove parent toggles + all on/off buttons
* [x] Update tests

### Steps to Confirm

* [ ] See video!

### Pre-Merge Checklist

* [x] Relevant Follow-Up Issues Created https://ethyca.atlassian.net/browse/PROD-1318
* [ ] Update `CHANGELOG.md`


clients/fides-js/src/components/ConsentContent.tsx
```diff
@@ -1,5 +1,5 @@
 import { ComponentChildren, VNode, h } from "preact";
-import { HTMLAttributes } from "react";
+import type { HTMLAttributes } from "react";
```
allisonking, line 2: I realized I was mistakenly importing from "react" in another place—I'm not sure if this was contributing too, but ultimately we don't really want to import from "react" because that causes "preact/compat" to kick in, which adds 2kb! I'm pretty sure importing just the `type` makes it safer

clients/fides-js/src/components/tcf/FilterButtons.tsx
```diff
@@ -2,35 +2,36 @@ import { h } from "preact";
 import { useState } from "preact/hooks";
 
 interface Filter {
-  name: string;
+  label: string;
+  value: string;
 }
 
-const FilterButtons = ({
+const FilterButtons = <T extends Filter>({
```
allisonking, line 9: I refactored this component to use an object instead of just the `name` string. makes comparing LegalBasisEnums a little more type safe

clients/fides-js/src/components/tcf/TcfPurposes.tsx
```diff
@@ -130,25 +74,66 @@ const TcfPurposes = ({
...
+    enabledPurposeIds: string[];
+    specialPurposes: TCFSpecialPurposeRecord[];
+    enabledSpecialPurposeIds: string[];
+  } = useMemo(() => {
+    if (activeLegalBasisOption.value === LegalBasisEnum.CONSENT) {
+      return {
+        purposes: uniquePurposes.filter((p) => p.isConsent),
+        purposeModelType: "purposesConsent",
+        enabledPurposeIds: enabledPurposeConsentIds,
+        // TODO: do we need to filter special purposes too?
```
allisonking, line 92: yes—we should make a follow up ticket for this. dunno if it can make it into this sprint though, would need both FE and BE work

allisonking: https://ethyca.atlassian.net/browse/PROD-1318

simon-keane:
This is looking really polished now @allisonking thank you. One thing I'm seeing in the loom video, so it's hard to tell what is happening, there seems to be a jump in the scroll position when toggling between Consent and Legitimate Interest in the new master toggle. Is it possible to retain the scroll position so we don't see that jump?

eastandwestwind (approved):
Wooot nice work on the refactor to use a single component for the records, cutting down on the LOC!

This looks great, just want to be sure we've tested for any styling issues in `?fides_embed=true` mode as well?

allisonking:
> This is looking really polished now @allisonking thank you. One thing I'm seeing in the loom video, so it's hard to tell what is happening, there seems to be a jump in the scroll position when toggling between Consent and Legitimate Interest in the new master toggle. Is it possible to retain the scroll position so we don't see that jump?

Good eye! Extremely conveniently, this got fixed in another PR by moving the Footer out of the ConsentContent, so it's looking good now

https://github.com/ethyca/fides/assets/24641006/c87a1c09-4823-4687-a1b0-a67861598bc6

allisonking:
> Wooot nice work on the refactor to use a single component for the records, cutting down on the LOC!
> 
> This looks great, just want to be sure we've tested for any styling issues in `?fides_embed=true` mode as well?

Just took a look! I think it looks okay, but let me know what you think

https://github.com/ethyca/fides/assets/24641006/44a04b8f-0d93-467d-abdb-e9440752bb91

clients/fides-js/src/components/tcf/TcfPurposes.tsx
allisonking: done, only possible thanks to @pattisdr 's quick backend work! [1543e88](https://github.com/ethyca/fides/pull/4376/commits/1543e88a6603656739de3bbe18011ce54f0c8a54)

PR #4377, opened 2023-11-02
Backend TC String Is Service Specific should be True 

pattisdr:
Closes https://ethyca.atlassian.net/browse/PROD-1311

### Description Of Changes

Google Ads considers this an error and will only accept service-specific scopes

### Code Changes

* [ ] set is_service_specific to true on TC Model which is then converted to a TC string

### Steps to Confirm

* [ ]  Enable TCF, add a system from the GVL like Emerse, and then get Privacy Experiences with include_meta=True:
```
curl -X 'GET' \
  'http://localhost:8080/api/v1/privacy-experience?show_disabled=true&region=fr&systems_applicable=false&include_gvl=false&include_meta=true&page=1&size=50' \
  -H 'accept: application/json'
```
* [ ] The accept-all string that is built should have is_service_specific set to True when decoded. Example: `CP0nQEAP0nQEAGXABBENAYEgALAAAEOAAAAAAEAEACACAAAA.IAEAEAAA,1~`

Screenshot 2023-11-02 at 1 22 24 PM

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### DIFF
--- a/tests/ops/util/test_tc_string.py
+++ b/tests/ops/util/test_tc_string.py
@@ -335,7 +335,7 @@ class TestBuildTCModel:
         m = TCModel(
             vendor_legitimate_interests=[66]
         )  # This vendor doesn't have leg int purposes, or special purposes. It does have flexible purposes,
-        # but this isn't an option right now, as is_service_specific is True
+        # but this isn't yet supported
         assert m.vendor_legitimate_interests == []
 
         m = TCModel(
--- a/tests/ops/util/test_tc_string.py
+++ b/tests/ops/util/test_tc_string.py
@@ -335,7 +335,7 @@ class TestBuildTCModel:
         m = TCModel(
             vendor_legitimate_interests=[66]
         )  # This vendor doesn't have leg int purposes, or special purposes. It does have flexible purposes,
-        # but this isn't an option right now, as is_service_specific is False
+        # but this isn't an option right now, as is_service_specific is True
         assert m.vendor_legitimate_interests == []
 
         m = TCModel(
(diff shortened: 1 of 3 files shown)


tests/ops/util/test_tc_string.py
```diff
@@ -335,7 +335,7 @@ def test_filter_invalid_vendor_legal_basis(self):
         m = TCModel(
             vendor_legitimate_interests=[66]
         )  # This vendor doesn't have leg int purposes, or special purposes. It does have flexible purposes,
-        # but this isn't an option right now, as is_service_specific is False
+        # but this isn't an option right now, as is_service_specific is True
```
allisonking, line 338: hm what does the `is_service_specific` value have to do with flexible purposes?

allisonking (approved):
looks good, just a question!

tests/ops/util/test_tc_string.py
pattisdr: thank you for catching my find and replace mistake!

pattisdr: This does highlight how the backend isn't accounting for flexible purposes when building the TC string though!

pattisdr: I don't have any publisher restrictions-type handling for TC strings on the backend: https://github.com/ethyca/fides/blob/main/src/fides/api/util/tcf/tc_model.py#L297, do you have it on the FE?

allisonking: nope not right now, I haven't really thought about it...

pattisdr: sounds good at least we match and we can add later as needed!

PR #4235, opened 2023-10-06
Update privacy center patch call for new API response

allisonking:
Closes https://github.com/ethyca/fides/issues/4225

### Description Of Changes

The privacy center uses the response of patching notices with user preferences in order to create/update its cookie. The response payload format changed in https://github.com/ethyca/fides/pull/4201 to nest preferences under an object. This PR allows the FE to handle this new expected payload type.

### Code Changes

* [x] Update privacy center types
* [x] Minor code updates to handle new response type
* [x] Update cypress fixtures

### Steps to Confirm

Everything should work as before! The main thing that changed is the API returns a new type, so now the FE handles that. To prove it's working, you'll want to observe that your cookie updates according to your saved preferences. This is for notices only, not TCF.

* Enable "Data sales and sharing" for privacy notices in admin-ui
* VPN to california
* Visit the privacy center and click "Manage your consent"
* Take a look at your cookie, if one exists. Data sales and sharing is default opt-in, but that might not be reflected in your cookie yet (esp if you don't have a cookie yet!)
* Opt out of data sales and sharing
* Your cookie value should now say ```{"consent":{"data_sales_and_sharing":false}}```

### Pre-Merge Checklist

* [ ] All CI Pipelines Succeeded
* [x] Issue Requirements are Met
* [x] Update `CHANGELOG.md`

PR #4383, opened 2023-11-03
Backend System-Level Cookie Support 

pattisdr:
Closes https://ethyca.atlassian.net/browse/PROD-1310

❗ Blocked by fideslang release: https://github.com/ethyca/fideslang/pull/181
❗ Blocked by FE accompaniment: https://ethyca.atlassian.net/browse/PROD-1319

### Description Of Changes

Add system-level cookie support, (cookies that are attached at the System level and not directly to a data use). Remove the functionality that keeps a cookie attached to a System if its PrivacyDeclaration is deleted. Cookies should now be linked to a privacy declaration OR a system, not both.

Separately, while here:
- Make a small change to SystemHistory.edited_by property, to allow automated System Updates to not show null in System History because they don't correspond to a user 
- Add new system "previous_vendor_id" column to Systems history table so we can update fideslang

### Code Changes

- Temporarily pin to fideslang commit that adds "cookies" to System schema.
- Remove cookies from SystemResponse schema now that it is on the inherited System schema
- Update "create_system" to pop off system level cookies from the request and upsert them separately
- Update "update_system" to pop off system level cookies and upsert separately
- Update "upsert_cookies" method to take in a system or a privacy declaration and link to the appropriate resource, not both.
- Now cascade delete Cookies when a Privacy Declaration is updated instead of setting the privacy declaration id to null
- Small data migration to detach Cookies from Systems that are also attached to a Privacy Declaration
- Update SystemHistory.edited_by to return the user_id as-is if there is no matching user
- Add new `System.previous_vendor_id` column to support being able to upgrade fideslang

### Steps to Confirm

* [ ] You can now POST cookies at the System level directly. These are not associated with a data use:
```bash
curl -X 'POST' \
  'http://localhost:8080/api/v1/system' \
  -H 'Authorization: Bearer <REDACTED>' \
  -d '{
  "fides_key": "system_a",
  "system_type": "system",
  "name": "System A",
  "description": "test description",
  "privacy_declarations": [],
  "cookies": [
    {
      "name": "test_cookie",
      "path": "/",
      "domain": "https://www.example.com"
    }
  ]
}'
```
* Also adding Cookies in the UI to a Privacy Declaration and then deleting the Privacy Declaration will now delete the Cookie (verify in the System response that it is not on System.cookies)

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`


pattisdr:
@allisonking you can add the FE here directly if that's easier!

pattisdr:
we might need to push a fideslang alpha tag for this if that not being merged yet blocks you here -

src/fides/api/db/system.py
```diff
@@ -147,57 +150,73 @@ async def upsert_privacy_declarations(
...
-    Upsert cookies for the given privacy declaration: retrieve cookies
-    by name/system/privacy declaration.
+    Upsert cookies for the given system or privacy_declaration.  Cookies can be attached at the system-level
+    or the privacy declaration-level.
 
     Remove any existing cookies that aren't specified here.
     """
+    if not (privacy_declaration or system):
+        # Dev-level error
+        raise Exception(
```
adamsachs, line 164: nit: this may actually be a good use case for `ValueError` :) ?
```suggestion
        raise ValueError(
```

src/fides/api/db/system.py
```diff
@@ -147,57 +150,73 @@ async def upsert_privacy_declarations(
 async def upsert_cookies(
     async_session: AsyncSession,
     cookies: Optional[List[Dict]],
-    privacy_declaration: PrivacyDeclaration,
-    system: System,
+    privacy_declaration: Optional[PrivacyDeclaration],
+    system: Optional[System],
```
adamsachs, line 154: nice updates to the function here, this is easy to follow 👍

src/fides/api/models/system_history.py
```diff
@@ -45,4 +45,4 @@ def edited_by(self) -> Optional[str]:
         db = Session.object_session(self)
         user: Optional[FidesUser] = FidesUser.get_by(db, field="id", value=self.user_id)
 
-        return user.username if user else None
+        return user.username if user else self.user_id
```
adamsachs, line 48: i see this called out as an additional change. just checking, this isn't specifically related to the system-level cookie support, right? that's OK if it's not, just wanted to make sure i'm not misunderstanding/missing any implications

src/fides/api/db/system.py
```diff
@@ -147,57 +150,73 @@ async def upsert_privacy_declarations(
...
     """
+    if not (privacy_declaration or system):
+        # Dev-level error
+        raise Exception(
+            "Either system or privacy declaration must be supplied to upsert cookies"
+        )
+
+    if privacy_declaration and system:
+        # Dev-level error
+        raise Exception(
```
adamsachs, line 170: same as above, just providing a suggestion here to make the change easier if you do wanna make it! 
```suggestion
        raise ValueError(
```

adamsachs (approved):
this looks great to me @pattisdr ! the code reads very cleanly, and the functionality here feels a lot more straightforward and maintainable.

thanks for revisiting this and taking the time to resolve this _the right way_

allisonking:
Did a separate PR for the FE just because a good number of files changed (all very small changes!) https://github.com/ethyca/fides/pull/4395

src/fides/api/models/system_history.py
pattisdr: Could have done better at motivating why here. The whole reason I looked into doing this PR in the first place was to be able to make use of  the system history captured when "update_system" was called.  We are going to be having "compass" update the system, not a user.  This will keep "None" from showing up in the FE as the editing user.

adamsachs: ah got it, thanks for that context! makes sense 👍

requirements.txt
```diff
@@ -12,7 +12,7 @@ defusedxml==0.7.1
 expandvars==0.9.0
 fastapi[all]==0.89.1
 fastapi-pagination[sqlalchemy]==0.11.4
-fideslang==2.2.0
+fideslang @ git+https://github.com/ethyca/fideslang.git@f32c295f72666ab5ee035353c57ea3752e4240e4#egg=fideslang
```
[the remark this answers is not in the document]
pattisdr: yes that would be great if we could get fideslang released when your work is in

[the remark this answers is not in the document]
pattisdr: updated -

pattisdr:
Lots of failing tests since bumping fideslang requirement - investigating.

clients/admin-ui/src/types/api/models/BasicSystemResponse.ts
```diff
@@ -114,6 +114,10 @@ export type BasicSystemResponse = {
    * The unique identifier for the vendor that's associated with this system.
    */
   vendor_id?: string;
+  /**
+   * If specified, the unique identifier for the vendor that was previously associated with this system.
+   */
+  previous_vendor_id?: string;
```
pattisdr, line 120: @allisonking! I got to use the instructions you typed up for me here! https://github.com/ethyca/fides/pull/4235#discussion_r1349288757 🙏

allisonking: amazing!!!

PR #4397, opened 2023-11-07
High Level Track Compass System Sync

pattisdr:
Closes https://ethyca.atlassian.net/browse/PROD-1332

### Description Of Changes

Add table to high-level track automatic compass system syncs and add unique constraint on cookie name/system id.

### Code Changes

- Update _audit_system_changes to return whether the system/privacy declaration was updated.
- Update "update_system" to return a tuple, of the updated system, and then whether the system was updated.
- Add a new SystemCompassSync table with model methods
- Add a unique constraint on cookie name and system id, deleting duplicates first.

### Steps to Confirm

* [ ] [See this in action in Plus](https://github.com/ethyca/fidesplus/pull/1201/commits/adb3524a96960e920b4d0312b204817d3c2b3b24)

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### DIFF
--- a/src/fides/api/alembic/migrations/versions/842790fa918a_add_compass_system_sync.py
+++ b/src/fides/api/alembic/migrations/versions/842790fa918a_add_compass_system_sync.py
@@ -7,7 +7,6 @@ Create Date: 2023-11-07 16:22:01.746436
 """
 import sqlalchemy as sa
 from alembic import op
-from sqlalchemy import text
 from sqlalchemy.dialects import postgresql
 
 # revision identifiers, used by Alembic.
@@ -18,8 +17,6 @@ depends_on = None
 
 
 def upgrade():
-    bind = op.get_bind()
-
     op.create_table(
         "system_compass_sync",
         sa.Column("id", sa.String(length=255), nullable=False),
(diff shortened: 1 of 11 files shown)


src/fides/api/alembic/migrations/versions/842790fa918a_add_compass_system_sync.py
```diff
@@ -0,0 +1,65 @@
...
+
+    # Remove duplicate cookies on system id and system name.
+    bind.execute(
+        text(
+            "DELETE FROM cookies A USING cookies B "
+            "WHERE A.id < B.id AND A.name = B.name "
+            "AND A.system_id = B.system_id "
+            "AND A.system_id IS NOT NULL"
+        )
+    )
```
pattisdr, line 55: I deterministically delete duplicate cookies with same name/system id.  Their data on other fields may differ. I viewed this as acceptable since we are only just adding proper System cookie support in the backend, existing System cookies would have been added in bulk add only, or if a privacy declaration was deleted, and system sync will sync this up if necessary.

adamsachs: hmm - should we put this in its own migration? i'm fine with it in the same PR, but i do feel like overloading a migration could make things more difficult to trace back...

adamsachs (approved):
looks good! just a note about whether to include a different migration for the cookie table update since i think it's unrelated? not a huge deal either way though.

src/fides/api/alembic/migrations/versions/842790fa918a_add_compass_system_sync.py
pattisdr: sure I can split it

PR #4405, opened 2023-11-08
PROD-1348 Fides string should always override experience prefs

eastandwestwind:
Closes https://ethyca.atlassian.net/browse/PROD-1348

### Description Of Changes

This change addresses a root inconsistency in our [sources of truth for user preference](https://github.com/ethyca/fides/blob/4bdbbe59d029c6fe4f6be7dbd84812381db97dc8/clients/privacy-center/cypress/e2e/consent-banner-tcf.cy.ts#L1381) for TCF.

Currently:
1. User saves preference
2. The Fides save prefs API call fails for some reason, but we still save the cookie
3. User returns to site and opens the consent modal. Their preferences are not reflected in the UI because the client-side fetched experience returned no prefs.

Desired:
1. User saves preference
2. The Fides save prefs API call fails for some reason, but we still save the cookie
3. User returns to site and opens the consent modal. Their preferences **are** reflected in the UI because we use the `cookie.fides_string` to get user prefs.

Note: this should also affect non-TCF.

### Code Changes

* [ ] Update TCF logic to use prefs on `cookie.fides_string` if they exist instead of prefs on client-side fetched experience

### Steps to Confirm

* [ ] Save preference on a TCF or non-TCF banner /modal
* [ ] Alter the `cookie.consent` (non-TCF) or `cookie.fides_string` (TCF) to be something different
* [ ] Reload the page, open consent modal link
* [ ] See that consent preferences reflect your manual changes to cookie instead of from the experience

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`


clients/fides-js/src/fides-tcf.ts
```diff
@@ -125,41 +125,39 @@ const updateCookieAndExperience = async ({
...
+      cookie.fides_string
+    );
+    const tcfEntities = buildTcfEntitiesFromCookie(experience, cookie);
+    return { cookie, experience: tcfEntities };
+  }
+
   // If user has no prefs saved, we don't need to override the prefs on the cookie
   if (!hasSavedTcfPreferences(experience)) {
-    return { cookie: { ...cookie, fides_string: "" }, experience };
+    return { cookie, experience };
```
eastandwestwind, line 157: Even though this line shouldn't be hit if we have a `fides_string`, defensively, we don't want to overwrite the `fides_string` based on prefs from experience

allisonking: hmm would this mess with our initialization logic in the case of no fides string and no saved preferences yet? I guess there'd be no fides_string variable, which might be the same as having an empty string...

clients/fides-js/src/fides-tcf.ts
```diff
@@ -125,41 +125,39 @@ const updateCookieAndExperience = async ({
...
   }
 
+  // If the user has prefs on a client-side fetched experience, but there is no fides_string, we
```
allisonking, line 160: incomplete comment?

clients/fides-js/src/fides-tcf.ts
```diff
@@ -125,41 +125,39 @@ const updateCookieAndExperience = async ({
...
-        cookie.fides_string
-      );
-      const tcfEntities = buildTcfEntitiesFromCookie(experience, cookie);
-      return { cookie, experience: tcfEntities };
-    }
+  if (!isExperienceClientSideFetched) {
     // If it's not client side fetched, we don't update anything since the cookie has already
     // been updated earlier.
     return { cookie, experience };
   }
```
allisonking, line 142: hm okay if we don't need the experience to update the cookie anymore, then I don't think we actually need this line anymore, do we? in other words, this function doesn't really care about `isExperienceClientSideFetched`—this function would still return the right thing even without this line (since a server side fetched exp wouldn't have preferences)? we could leave it since it might make things easier to reason about, but it would also be nice to cut out a parameter to this function

allisonking (commented):
I think this is reasonable, but left some comments!

clients/fides-js/src/fides-tcf.ts
eastandwestwind: Fixed!
> If the user has prefs on a client-side fetched experience, but there is no fides_string, we need to user the prefs on the experience to generate a fidesString and cookie.tcf_consent

clients/fides-js/src/fides-tcf.ts
eastandwestwind: We still will need the experience to update the cookie, if we don't have a `fides_string` on the `cookie`, but we have prefs on the `experience`. See updated comment here- https://github.com/ethyca/fides/pull/4405#discussion_r1388472610

Does that make sense?

allisonking: hm okay—this flow is confusing but I think it works!

allisonking (commented):
Okay thinking out loud a little here to help me understand this better... In our TCF initialization logic:

1. If there is a fides string, we update the cookie ✅ 
2. At some point, we need to update the experience based on the cookie. We do this through the `updateCookieAndExperience` function. Before, we only did this if the string was valid and if the experience came from the client. Now, we do this as long as the experience came from the client (I feel like I'm missing something here, since that logic doesn't change...?)
3. The tricky part here is that sometimes we _do_ want the preferences from the experience object, but this is increasingly becoming the nonstandard use case (when it used to be the standard!). So what is the case for using the experience object's preferences (which would only come client side)? 

And now we are also updating notice based initialization to prioritize the cookie over the experience preferences. The tests are failing (I believe legitimately)—so I think this also comes back to the question of when do we want preferences from the exp object? It feels like we are carving out cases for "actually use the cookie as source of truth" when perhaps we should be flipping it since cookie-as-source-of-truth seems to be more common now?

clients/fides-js/src/fides.ts
```diff
@@ -86,8 +90,23 @@ let _Fides: Fides;
...
   experience: PrivacyExperience,
-  debug?: boolean
+  debug?: boolean,
+  isExperienceClientSideFetched?: boolean
 ): Promise<{ cookie: FidesCookie; experience: PrivacyExperience }> => {
+  if (!isExperienceClientSideFetched) {
+    // If it's not client side fetched, we don't update anything since the cookie has already
+    // been updated earlier.
+    return { cookie: oldCookie, experience };
+  }
```
allisonking, line 100: I suspect this might be messing with our tests, which provide server side data with preferences, and we rely on those preferences for the tests... tests might be the only place where our server side data has preferences which we want to respect?

clients/fides-js/src/fides.ts
```diff
@@ -86,8 +90,23 @@ let _Fides: Fides;
...
+  if (oldCookie.consent) {
```
allisonking, line 101: this will always be true I think— `{}` is truthy

allisonking:
Okay, if we follow this rule "if we have cookie prefs, we should always use that over experience prefs", I think there is room for simplification here. It might not make sense to do it in this PR, but leaving a note here (should probably be done when we simplify the initialization logic):

`updateCookieAndExperience` should probably be broken up to (being explicit about the naming for comment purposes...) `updateCookieBasedOnExperience` and `updateExperienceBasedOnCookie`. Right now we essentially do `updateExperienceBasedOnCookie` followed by `updateCookieBasedOnExperience` with some checks in between to see if we should move on to `updateCookieBasedOnExperience`, but we would never (I think??) do both. Perhaps `initialize.ts` could look like:

```js
if (
  isPrivacyExperience(effectiveExperience) &&
  experienceIsValid(effectiveExperience, options)
) {
  // new check to see if there are cookie preferences
  if (cookiePreferencesExist()) {
    updateExperienceBasedOnCookie(); // buildTcfEntities and all that
  } else {
    updateCookieBasedOnExperience(); // transformTcfPreferencesToCookieKeys and all that
  }
}
```

eh, that might not actually be much simpler than what exists... we should prioritize that initialization simplification PR!

clients/fides-js/src/fides.ts
eastandwestwind: yes I actually removed this bit, adding `isExperienceClientSideFetched` to the condition below it, so:

```
if (isExperienceClientSideFetched && oldCookie.consent) {
    // if we already have preferences on the cookie, update client-side experience with those preferences
    const updatedExperience = updateExperienceFromCookieConsent({
      experience,
      cookie: oldCookie,
      debug,
    });
    return { cookie: oldCookie, experience: updatedExperience };
  }
```

clients/fides-js/src/fides-tcf.ts
```diff
@@ -291,7 +287,7 @@ const init = async (config: FidesConfig) => {
     experience,
     renderOverlay,
     updateCookieAndExperience: (props) =>
-      updateCookieAndExperience({ ...props, hasValidFidesStringOverride }),
+      updateCookieAndExperience({ ...props }),
```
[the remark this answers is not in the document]
eastandwestwind: ohhh right!

clients/fides-js/src/fides.ts
eastandwestwind: good catch! I've added a helper fn to help determine whether we have consent set on this obj

clients/privacy-center/cypress/e2e/consent-banner-tcf.cy.ts
```diff
@@ -1487,10 +1487,7 @@ describe("Fides-js TCF", () => {
             8: false,
             9: true,
           });
-          expect(tcData.purpose.legitimateInterests).to.eql({
-            [PURPOSE_2.id]: true,
-            1: false,
-          });
+          expect(tcData.purpose.legitimateInterests).to.eql({});
```
allisonking, line 1490: hm is this because the cookie doesn't have purpose legitimate interests? but this used to expect something because it prioritized the experience obj?

allisonking (approved):
thanks for sticking through with this one, it's a gnarly one!

clients/privacy-center/cypress/e2e/consent-banner-tcf.cy.ts
eastandwestwind: yes exactly, we have to ensure we're using the cookie consents instead of experience, so if it's not set there, we have to ensure it isn't set overall.

PR #4406, opened 2023-11-08
Gracefully handle Foreign Key Constraint Errors

ThomasLaPiana:
Closes https://ethyca.atlassian.net/browse/PROD-1108?focusedCommentId=37809

### Description Of Changes

When someone deletes a `dataset`, it throws an error due to foreign key constraints if there is a referenced `connection`

The solution is to update our `delete` logic to more intelligently catch and surface errors relate to foreign key constraints instead of blanket labeling everything as the same error.

### Code Changes

* [x] catch and surface IntegrityErrors when deleting objects

### Steps to Confirm

* [x] the initial `Steps to Reproduce` no longer work in the Jira ticket

### Pre-Merge Checklist

* [x] Update `CHANGELOG.md`

The new error message:

image


ThomasLaPiana:
Got it working like this so far:

https://github.com/ethyca/fides/assets/5105354/0e69d68e-2c19-4814-a2a9-6b5a751008c7

src/fides/api/db/crud.py
```diff
@@ -324,6 +327,19 @@ async def delete_resource(
                     log.debug("Deleting resource")
                     query = _delete(sql_model).where(sql_model.fides_key == fides_key)
                 await async_session.execute(query)
+            except IntegrityError as err:
+                raw_error_text: str = err.orig.args[0]
+
+                if "violates foreign key constraint" in raw_error_text:
+                    error_message = "Failed to delete resource! Foreign key constraint found, try deleting related objects first."
```
ThomasLaPiana, line 334: @NevilleS we can play with the wording here but generally I thought this was clear

adamsachs: i know you didn't ask me, but it looks pretty clear to me :) my only nit may be to change "object" to "resource"? but i'm not really sure, maybe that's less clear!
```suggestion
                    error_message = "Failed to delete resource! Foreign key constraint found, try deleting related resources first."
```

src/fides/api/db/crud.py
```diff
@@ -324,6 +327,19 @@ async def delete_resource(
                     log.debug("Deleting resource")
                     query = _delete(sql_model).where(sql_model.fides_key == fides_key)
                 await async_session.execute(query)
+            except IntegrityError as err:
```
adamsachs, line 330: i know we don't need to listen to everything codecov says, but it feels like we could have a test case to cover this, i.e. i don't think it should be too bad to provoke the error state in a test? specifically - the string matching in the error text may be a good piece of logic to ensure we've got coverage for.

if it's not an easy situation to get test coverage on for whatever reason then not a big deal to me.

adamsachs (approved):
nice @ThomasLaPiana , this makes sense and looks good to me! i was able to verify with some basic local testing with `fides_env` (see screenshot below).

only thing is whether we'd want a bit of automated test coverage, if that's not too hard to get in place.

image

src/fides/api/db/crud.py
ThomasLaPiana: You're totally right here. I'm deep in some Pydantic stuff but will get this added before merging :)

adamsachs: got coverage for this in https://github.com/ethyca/fides/pull/4406/commits/8512d52fae60b5ac2c9e60bba80661a06ae2579c to help get this across the line! i think we're good to go, i'll let CI run through again...

PR #4410, opened 2023-11-09
PROD-1303: Adds support for custom get experience fn and custom served notices fn

eastandwestwind:
Closes https://ethyca.atlassian.net/browse/PROD-1303

### Description Of Changes

Adds support for 2 new custom functions provided through Fides.init(): 
1. `getPrivacyExperienceFn`- custom get experience fn
2. `patchNoticesServedFn`- custom served notices fn

### Code Changes

* [ ] Adds support for custom get experience fn and custom served notices fn
* [ ] refactors exernal api calls to be more modular
* [ ] adds e2e tests

### Steps to Confirm

* [ ] Run e2e tests

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`


clients/fides-js/src/lib/initialize.ts
```diff
@@ -317,14 +319,21 @@ export const initialize = async ({
...
-      // the Fides API using the current region string
-      effectiveExperience = await fetchExperience(
-        fidesRegionString,
-        options.fidesApiUrl,
-        options.debug,
-        cookie.identity.fides_user_device_id
-      );
+      // If no effective PrivacyExperience was pre-fetched, fetch one using the current region string
+      if (options.apiOptions?.getPrivacyExperienceFn) {
+        effectiveExperience = await customGetPrivacyExperience(
```
allisonking, line 324: I don't feel super strongly about this, but I wonder if it would make things easier or harder if we did the check for if there's a custom function inside the relevant function call, instead of outside of it? for example:

```js
const fetchExperience = (params) => {
  if (options.apiOptions?.getPrivacyExperienceFn) {
    return await customGetPrivacyExperience(...)
  }
  // whatever we already had
}
```

that way the outside logic could stay the same, and it's just the function internals of `fetchExperience` (and `patchNoticesServed`) that is changing?

maybe that complicates other arguments? but I think it'd be nicer in the sense of if we `fetchExperience` or `patchNoticesServed` in more than one place, we are covered

clients/fides-js/src/lib/preferences.ts
```diff
@@ -54,11 +31,11 @@ async function savePreferencesApi(
   userLocationString?: string
 ) {
   if (options.apiOptions?.savePreferencesFn) {
-    debugLog(options.debug, "Calling custom save preferences fn");
-    await options.apiOptions.savePreferencesFn(
+    await customSaveConsentPreferences(
```
allisonking, line 34: actually yeah—this is the pattern I mean in my earlier comment, where `savePreferencesApi` is conditionally calling the custom func, and if it's not there, it continues doing whatever it originally did

clients/fides-js/src/services/external/preferences.ts
```diff
@@ -0,0 +1,58 @@
...
+      e
+    );
+    return null;
+  }
+}
+
+/**
+ * Helper function to save preferences to an external API
+ */
+export async function customSaveConsentPreferences(
```
allisonking, line 35: these all follow the same pattern so could probably be refactored to share more code, though I know that's tricky because they all take different arguments! would maybe use typescript generics—in any case, not asking you to change it now since it is not trivial, but might be an opportunity for refactoring in the future

clients/fides-js/src/lib/initialize.ts
allisonking: having it the way you have it now also makes things a little odd, in that there are two checks for `options.apiOptions?.getPrivacyExperienceFn`—one here, and one inside `customGetPrivacyExperience`

clients/fides-js/src/lib/preferences.ts
eastandwestwind: Yeah, I was back and forth on this. It's a bit tricky since `options.apiOptions.savePreferencesFn()` returns `void`, so this means if that custom fn were successful, we would still try to call the Fides base fn. We need to disable calling the Fides base fn when a custom fn is defined. I'm trying to think of a better way to track whether the fn was defined.

E.g. having `customGetPrivacyExperience` return a `boolean` (to track whether or not the fn was called) or `experience` is problematic because then we have to do an extra check everywhere we want to use the `experience` outside of the fn call to see if it's an actual experience vs `false`.

Let me know if you want me to walk you through a screenshare, as it's tough to fully describe via text 😅

clients/fides-js/src/services/api.ts
```diff
@@ -22,12 +25,31 @@ export const fetchExperience = async (
   userLocationString: string,
   fidesApiUrl: string,
   debug: boolean,
+  apiOptions?: FidesApiOptions | null,
```
allisonking, line 28: we can do this later, but I think we either need to parametrize `fetchExperience`'s arguments, or pass in the whole `options` object. there are too many positional arguments!

allisonking (approved):
woo nice, the refactor turned out nicely, I think! the general logic flow in `initialize` is much cleaner now 💯

PR #4411, opened 2023-11-09
[fides.js] Small CSS updates

allisonking:
Needed for https://ethyca.atlassian.net/browse/PROD-1340

### Description Of Changes

Allows buttons to take different ids as needed. The originally id wasn't being used (and it's a little odd that multiple buttons would get the same id)

### Code Changes

* [x] Allow button to take `id` prop
* [x] Pass in id names for save, accept all, reject all
* [x] Update cypress test

### Steps to Confirm

* [ ] If you start up the TCF overlay and inspect the modal buttons, they will now have IDs that are relevant to what the buttons do. This makes it easier for extensions to select them for styling.

### Pre-Merge Checklist

* [ ] Issue Requirements are Met - companion to https://github.com/ethyca/fidesplus/pull/1220
* [x] Update `CHANGELOG.md`


clients/privacy-center/cypress/e2e/consent-notices.cy.ts
```diff
@@ -118,6 +118,10 @@ describe("Privacy notice driven consent", () => {
         expect(
           preferences.map((p: ConsentOptionCreate) => p.preference)
         ).to.eql(["opt_in", "opt_in", "acknowledge"]);
+        // Wait for toast so that we know cookie is ready to be inspected
```
allisonking, line 121: unrelated to this issue, but the test failed on this PR so I added this which should make the test less flaky

clients/fides-js/src/components/Button.tsx
```diff
@@ -5,16 +5,18 @@ interface ButtonProps {
...
 const Button: FunctionComponent<ButtonProps> = ({
   buttonType,
   label,
   onClick,
+  id,
 }) => (
   <button
     type="button"
-    id={`fides-banner-button-${buttonType.valueOf()}`}
+    id={id}
```
allisonking, line 19: I think this is a good change since html `id`s should be unique in html, and this resulted in duplicates of the same ID. this change does pose a risk if any one was already relying on this id though! they would be able to swap to use the class (of the same name) instead, but that would be a change

eastandwestwind: Hm, so we do have a customer relying on these ids. Is it possible to use a different approach here, perhaps generating a new class instead?

allisonking: okay, I'll keep the id around and work around this!

allisonking: okay, changed it to an additive classname instead 👍

eastandwestwind (approved):
Thanks for adding as a className for now! I know it's not ideal to not be able to change the `id`s 😬

galvana (approved):
Verified locally 👍

clients/fides-js/src/components/Button.tsx
```diff
@@ -5,17 +5,19 @@ interface ButtonProps {
...
 
 const Button: FunctionComponent<ButtonProps> = ({
   buttonType,
   label,
   onClick,
+  className = "",
 }) => (
   <button
     type="button"
     id={`fides-banner-button-${buttonType.valueOf()}`}
```
NevilleS, line 19: Unrelated, but this is still a bug here - we can have multiple types of the same button in the DOM which is invalid HTML 👍

allisonking: okay—I guess we'll have to come up with a plan to fix these ids while also migrating customers who are already using these ids then.

PR #4430, opened 2023-11-14
PROD-1389 for TCF, any consent pref that is not defined on cookie should be assumed opt-out

eastandwestwind:
Closes https://ethyca.atlassian.net/browse/PROD-1389

### Description Of Changes

For TCF, if we have a previously saved `fides_consent` cookie or a `fides_string` override,  any consent pref that is not defined on cookie.tcf_consent should be assumed opt-out, instead of using the defaults on the experience.

### Code Changes

* [ ] Updates default consent for tcf_entities on experience to be opt-out if we have any previously saved consent prefs on either cookie.tcf_consent or a fides_string override (which actually gets put on the `cookie` too).
* [ ] Update e2e tests

### Steps to Confirm

* [ ] Nav to http://localhost:3000/fides-js-components-demo.html with TCF enabled and TCF experience configured for your geo
* [ ] Click reject all, take note of the generated fides_string (window.Fides.fides_string)
* [ ] Clear cookies
* [ ] Reload page
* [ ] Add the `?fides_string` query param to the URL, e.g. http://localhost:3000/fides-js-components-demo.html?fides_string=CP1OzUAP1OzUAGXABBENATEgAAAAAAAAAAAAAAAAAAAA.IABE,1~
* [ ] Open up consent modal
* [ ] See that everything in the modal UI is opted out, instead of showing any defaults from the experience.

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### DIFF
--- a/clients/fides-js/src/lib/cookie.ts
+++ b/clients/fides-js/src/lib/cookie.ts
@@ -316,8 +316,10 @@ export const buildTcfEntitiesFromCookie = (
               Boolean(cookieConsent[item.id]),
               ConsentMechanism.OPT_IN
             )
-          : // if experience contains a tcf entity not defined by tcfEntities, we override experience current pref with the default pref
-            item.default_preference;
+          : // If experience contains a tcf entity not defined by tcfEntities, this means either:
+            // A) Most commonly, user has opted out, and opt-outs are not tracked by TC string. It's safe to assume this case.
+            // B) There is a new tcf entity that requires consent. In this case we would just resurface the banner
+            ConsentMechanism.OPT_OUT;
         return { ...item, current_preference: preference };
       });
     });
--- a/clients/fides-js/src/lib/cookie.ts
+++ b/clients/fides-js/src/lib/cookie.ts
@@ -311,15 +311,18 @@ export const buildTcfEntitiesFromCookie = (
       const cookieConsent = cookie.tcf_consent[cookieKey] ?? {};
       // @ts-ignore the array map should ensure we will get the right record type
       tcfEntities[experienceKey] = experience[experienceKey]?.map((item) => {
+        const defaultPreference = cookie.fides_string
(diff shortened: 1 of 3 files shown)


clients/fides-js/src/lib/cookie.ts
```diff
@@ -316,8 +316,10 @@ export const buildTcfEntitiesFromCookie = (
               Boolean(cookieConsent[item.id]),
               ConsentMechanism.OPT_IN
             )
-          : // if experience contains a tcf entity not defined by tcfEntities, we override experience current pref with the default pref
-            item.default_preference;
+          : // If experience contains a tcf entity not defined by tcfEntities, this means either:
+            // A) Most commonly, user has opted out, and opt-outs are not tracked by TC string. It's safe to assume this case.
+            // B) There is a new tcf entity that requires consent. In this case we would just resurface the banner
+            ConsentMechanism.OPT_OUT;
```
allisonking, line 322: I think we only want this for the fides string override case. in other words, I think this change would break this flow:
* No fides string override
* User opts out of all purposes
* A new legint purpose is added
* On load, the new legint purpose would be opted out, even though in this case, because we have the cookie key consent, we actually know the user has _not_ opted out.

I think this could be fixed with...

```js
const defaultPreference = cookie.fides_string ? ConsentMechanism.OPT_OUT : item.default_preference;
const preference = Object.hasOwn(cookieConsent, item.id) 
                 ? transformConsentToFidesUserPreference(
                      Boolean(cookieConsent[item.id]),
                      ConsentMechanism.OPT_IN
                    )
                 : defaultPreference;
```

eastandwestwind: Ah this makes more sense, thanks @allisonking !

allisonking (approved):
nice job turning this around so quickly! 💯

PR #4431, opened 2023-11-14
GPP stub

allisonking:
Part of https://ethyca.atlassian.net/browse/PROD-1284

Actually using this requires more thought, which is the basis of [this design doc](https://ethyca.atlassian.net/wiki/spaces/EN/pages/2836856888/2023-11-13+Fides.js+bundling+with+GPP)

### Description Of Changes

Creates the stub for the GPP CMP API. GPP is yet another IAB standard. Similar to how there is a TCF CMP API, there is also a GPP CMP API. One requirement for both APIs is that they be immediately callable. Because it takes time to initialize the actual API, the strategy recommended by the IAB is to first create stub functions that can reply with proper defaults on page initialization. The stubs queue up commands which, once the actual API is initialized, the actual API will execute on.

In the case of GPP, this is a function called `__gpp` which can take various "generic" [commands listed here in the spec](https://github.com/InteractiveAdvertisingBureau/Global-Privacy-Platform/blob/main/Core/CMP%20API%20Specification.md#what-required-api-commands-must-a-cmp-support-). This PR stubs out all of the generic commands. 

image

### Code Changes

* [x] Add new stub file for GPP
* [x] Refactor to share code between GPP and TCF
* [x] Unit tests

### Steps to Confirm

* [ ] Unit tests cover this for the most part
* [ ] Since we haven't figured out how we will integrate this yet, you'd have to do some manual integration to trigger GPP. You can do this by first commenting out the call to `initializeGppCmpApi();` in `gpp.ts`. Then, in `fides-tcf.js`, add a call to `initializeGppCmpApi()` right after `initializeTcfCmpApi()`. You'll have to increase the size of your TCF payload to ~55kb for this to build.
* [ ] Once you've set that up, you can start up the privacy center via `turbo dev` and visit `localhost:3000/fides-js-demo.html?geolocation=fr-id`. Open up the console and you should be able to trigger any of the gpp commands, i.e.
```
__gpp("ping", (data) => {console.log({data})});
```

### Pre-Merge Checklist

* [x] Update `CHANGELOG.md`


clients/fides-js/src/lib/tcf/stub.ts
```diff
@@ -15,32 +17,10 @@ const isMessageData = (data: unknown): data is MessageData =>
...
 
 export const makeStub = () => {
-  const TCF_LOCATOR_NAME = "__tcfapiLocator";
   const queue: any[] = [];
   const currentWindow = window;
-  let frameLocator = currentWindow;
-  let cmpFrame;
   let gdprApplies: boolean;
 
-  function addFrame() {
```
allisonking, line 25: moved to `cmp-stubs` so that it can be shared with the gpp stub

clients/fides-js/src/lib/tcf/stub.ts
```diff
@@ -150,29 +130,11 @@ export const makeStub = () => {
    * "__tcfapilLocator" frame on every level. If one exists already then we are
    * not the master CMP and will not queue commands.
    */
-  while (frameLocator) {
```
allisonking, line 153: moved to `cmp-stubs` so that it can be shared with the gpp stub

clients/fides-js/src/lib/gpp/stub.ts
```diff
@@ -0,0 +1,171 @@
+/**
+ * Typescript adaptation of https://github.com/IABTechLab/iabgpp-es/blob/master/modules/stub
+ * Refactored to share code with the TCF version
```
allisonking, line 3: the original code looks like it was adapted from the TCF version but with some different naming conventions, and not quite as many guardrails as the TCF one has (and not typescript). This standardizes the two library stubs, uses TS, and hopefully has the same guardrails.

allisonking: The general format of these stubs is:

1. Look to see if a CMP already exists on the page
2. If not, instantiate the function (`__gpp` or `__tcfapi`) on the window object. Set up the "generic" commands (ping, addEventListener, etc. these are defined by the spec) to return something (likely that the CMP isn't ready yet)
3. Also set up so that we can `postMessage` (how iframes talk to one another)

clients/fides-js/src/lib/gpp/constants.ts
```diff
@@ -0,0 +1 @@
+export const SUPPORTED_APIS = ["2:tcfeuv2"];
```
allisonking, line 1: this will eventually come from the backend, I think

eastandwestwind: from which endpoint exactly?

eastandwestwind (approved):
Nice work extracting out shared code for both the gpp and tcf stubs!

clients/fides-js/src/lib/gpp/constants.ts
allisonking: I think it'll eventually come through the privacy experience endpoint. the backend will start recording it in https://ethyca.atlassian.net/browse/PROD-1361

PR #4434, opened 2023-11-15
Flexible Legal Basis = True Default

pattisdr:
❗ ~Contains migration; check downrev~
❗ ~Dependent on fideslang release of https://github.com/ethyca/fideslang/pull/184~

Closes #PROD-1264

### Description Of Changes

Make "flexible legal basis for processing" non nullable and True by default in the backend.

### Code Changes

- Update PrivacyDeclaration.flexible_legal_basis_for_processing to get rid of the third state. Make it required, and set the server default to True, to match what should happen when systems are created outside of Compass.
- Add a migration that first updates existing flexible_legal_basis_for_processing to False if it is currently null, then adds the new server default and makes it required.

### Steps to Confirm

* [ ] Creating a system without this field specified via the API (POST System) will set this value to True by default in the backend.

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,7 +12,7 @@ defusedxml==0.7.1
 expandvars==0.9.0
 fastapi[all]==0.89.1
 fastapi-pagination[sqlalchemy]==0.11.4
-fideslang==2.2.2a0
+fideslang==2.2.2
 fideslog==1.2.10
 firebase-admin==5.3.0
 GitPython==3.1.35
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,7 +12,7 @@ defusedxml==0.7.1
 expandvars==0.9.0
 fastapi[all]==0.89.1
 fastapi-pagination[sqlalchemy]==0.11.4
-fideslang==2.2.1
+fideslang @ git+https://github.com/ethyca/fideslang.git@d273421ccadb4833c19dc31e30de7ec332dc6a78#egg=fideslang
 fideslog==1.2.10
 firebase-admin==5.3.0
 GitPython==3.1.35
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,7 +12,7 @@ defusedxml==0.7.1
 expandvars==0.9.0
 fastapi[all]==0.89.1
 fastapi-pagination[sqlalchemy]==0.11.4
-fideslang @ git+https://github.com/ethyca/fideslang.git@d273421ccadb4833c19dc31e30de7ec332dc6a78#egg=fideslang
+fideslang==2.2.2a0
 fideslog==1.2.10
 firebase-admin==5.3.0
 GitPython==3.1.35
--- a/src/fides/api/models/sql_models.py
+++ b/src/fides/api/models/sql_models.py
@@ -475,7 +475,9 @@ class PrivacyDeclaration(Base):
 
     features = Column(ARRAY(String), server_default="{}", nullable=False)
     legal_basis_for_processing = Column(String)
-    flexible_legal_basis_for_processing = Column(BOOLEAN())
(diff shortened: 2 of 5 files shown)


src/fides/api/models/sql_models.py
```diff
@@ -475,7 +475,9 @@ class PrivacyDeclaration(Base):
 
     features = Column(ARRAY(String), server_default="{}", nullable=False)
     legal_basis_for_processing = Column(String)
-    flexible_legal_basis_for_processing = Column(BOOLEAN())
+    flexible_legal_basis_for_processing = Column(
+        BOOLEAN(), server_default="t", nullable=False
```
pattisdr, line 479: @adamsachs wanted to get your thoughts here before proceeding - as-is this maybe would have plus changes and compass changes somewhere.

An alternative, so plus/compass doesn't have to change, we still allow this to be nullable, but still make the default=True change.  That was my main objective here, so systems created outside of compass via the API or CLI would be True by default.  And then if compass returns as null, null/false basically mean the same thing.

requirements.txt
```diff
@@ -12,7 +12,7 @@ defusedxml==0.7.1
 expandvars==0.9.0
 fastapi[all]==0.89.1
 fastapi-pagination[sqlalchemy]==0.11.4
-fideslang==2.2.1
+fideslang @ git+https://github.com/ethyca/fideslang.git@d273421ccadb4833c19dc31e30de7ec332dc6a78#egg=fideslang
```
pattisdr, line 15: Temporary

src/fides/api/models/sql_models.py
adamsachs: @pattisdr i think i'm good with the approach you're proposing. the compass changes would be pretty minimal to adjust for this -- we'd simply need to have the webesrvice assume that a null/empty value for this field in airtable should be returned as `False` in the API. i could even get those changes staged very quickly on a branch for us to test out end to end. after thinking about this a bit, i think it may actually make things a bit clearer for compass to model things this way - Airtable doesn't have an explicit `False` state, so there's really no way to distinguish `False`/`null`, so by not allowing `null` across our stack, we can basically _force_ `null` to signify `False` in airtable/compass.

if i'm understanding the alternative you're presenting correctly - i.e. keep it nullable but default to true - my gut tells me it will just lead to more confusion down the road. we'd basically be establishing different result states depending on whether we save the attribute to the DB with an explicit `None` vs. not specifying the attribute on a given save, right? in some ways that _could_ enable some more flexibility, but i think it comes at the expense of predictability/clear behavior down the road. we'd need to constantly keep track of whether the field is specified with `None`, unspecified, defaulted to a particular value, etc. through multiple layers of our stack - and i've always found that difficult. tl;dr is that i think just eliminating the nullable state (as you're proposing here) forces us to explicitly specify true/false across our stack, and i think the slight loss in flexibility is worth the clarity gained...

where do you foresee changes in plus? not doubting, just i don't know off the top of my head and i want to make sure i'm understanding potential impact accurately

pattisdr: Great, just checking, I didn't know what it would take in compass. Let's go with what I have in this PR: I prefer that approach too.

Maybe plus doesn't need anything, maybe it's just compass thinking about it more.

pattisdr:
@adamsachs I think this is ready to go, just need a 👍

adamsachs:
> @adamsachs I think this is ready to go, just need a 👍 

Looks good - does changelog need to be updated?

adamsachs (approved):
looks good, thank you for making quick work of this!

pattisdr:
thank you for the changelog catch, I thought I'd done it, but just in fideslang!

PR #4435, opened 2023-11-17
Adding conditions for legal basis fields

galvana:
### Description Of Changes

Adding **This legal basis is flexible field** field above **Legal basis for processing**. The flexible field is only visible for GVL systems and is used as an indicator for why the **Legal basis for processing** field is enabled or disabled. The **Legal basis for processing** field will always be editable for systems not associated with a vendor.

### Code Changes

* [ ] Updated `PrivacyDeclarationForm.tsx` with the **This legal basis is flexible field** field
* [ ] Added Cypress tests

### Steps to Confirm

* [ ] Startup fidesplus with Compass enabled
* [ ] Create a system from the 1plusX vendor
* [ ] Navigate to data uses
* [ ] Click on Analytics for Advertising Performance
* [ ] The **This legal basis is flexible field** should be set to true and be disabled
* [ ] The legal basis field underneath should be editable
* [ ] Go back to the data use list and click on **Profiling for personalization**
* [ ] The **This legal basis is flexible field** should be set to false and be disabled
* [ ] The legal basis field underneath should not be editable
* [ ] Create new system but don't associate it to a Vendor
* [ ] Navigate to data uses
* [ ] Click on **Add data use**
* [ ] The **This legal basis is flexible field** should not be present and the legal basis field should be editable

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`


allisonking (changes requested):
I can change the flexible legal basis for 1plusX vendor's "Analytics for Advertising Performance" but I can't save it, which I'd expect to be able to do? The save button is gone, from the locked gvl work. 

I think we can probably just put the save button back in, but it'd need to be conditional, and hopefully there aren't gotchas from putting it back....

https://github.com/ethyca/fides/assets/24641006/79e388a1-752e-4357-9b84-b9ab1189a55c

clients/admin-ui/src/features/system/system-form-declaration-tab/PrivacyDeclarationForm.tsx
```diff
@@ -488,7 +490,7 @@ export const PrivacyDeclarationForm = ({
               >
                 Cancel
               </Button>
-              {!lockedForGVL ? (
+              {!lockedForGVL || values.flexible_legal_basis_for_processing ? (
```
galvana, line 493: Enabling the save button if `flexible_legal_basis_for_processing` is true

clients/admin-ui/cypress/e2e/systems-plus.cy.ts
```diff
@@ -87,6 +89,105 @@ describe("System management with Plus features", () => {
...
+                ...origSystem,
+                ...dictSystem,
+                fides_key: origSystem.fides_key,
+                customFieldValues: undefined,
+                data_protection_impact_assessment: undefined,
+              },
+            }
+          ).as("getDictSystem");
+        });
+      });
```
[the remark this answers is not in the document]
galvana: Ah! That makes sense, I was just trying to keep it consistent but thanks for letting me know there's a more standard way 😄

clients/admin-ui/cypress/e2e/systems-plus.cy.ts
```diff
@@ -156,6 +156,81 @@ describe("System management with Plus features", () => {
...
 
+  describe("data use", () => {
+    it("should enable legal basis editing if flexible is true", () => {
+      cy.fixture("systems/system_with_flexible_legal_basis.json").then(
+        (system) => {
+          cy.intercept("GET", "/api/v1/system/*", {
+            body: system,
+          }).as("getSystemWithFlexibleDataUses");
+        }
+      );
```
allisonking, line 167: this can be simplified since you aren't modifying the `system` obj (don't commit this directly as it's not formatted correctly, but something like)

```suggestion
      cy.intercept("GET", "/api/v1/system/*, 
        { fixture: "systems/system_with_flexible_legal_basis.json" }
      ).as("getSystemWithFlexibleDataUses");
```

clients/admin-ui/cypress/e2e/systems-plus.cy.ts
```diff
@@ -156,6 +156,81 @@ describe("System management with Plus features", () => {
...
+    });
+
+    it("should disable legal basis editing if flexible is false", () => {
+      cy.fixture("systems/system_with_flexible_legal_basis.json").then(
+        (system) => {
+          cy.intercept("GET", "/api/v1/system/*", {
+            body: system,
+          }).as("getSystemWithFlexibleDataUses");
+        }
+      );
```
allisonking, line 193: same simplification as above can be done here 👍 . generally, `cy.intercept` with a `fixture` argument is the simplest way to do things. `cy.fixture()` with a nested `cy.intercept` is needed when you want to modify the fixture a little first. but overall, thanks for simplifying the fixture already!

allisonking (approved):
code looks great, and UI looks good too! just a suggestion for simplifying the cypress fixture call even more

galvana:
From @mfbrown 
> Given the new approach we’re taking to overriding purposes instead of changing the legal basis on a vendor by vendor basis, let’s leave this PR and task open for now so we don’t introduce an issue when compass tries to update a vendor.

PR #4442, opened 2023-11-20
Fideslang 3/Pydantic 2.0 Upgrade

ThomasLaPiana:
Closes #<issue>

### Description Of Changes

This will be a very large change encapsulating multiple version changes within very major requirements, such as `FastAPI`, `Pydantic`, and related dev requirements like `mypy` and `pylint`. Also very important to note is the removal of support for `Python 3.8`.

### Warnings

With version 2.0, `Pydantic` is functioning much more as a proper typing system than a validation/coercion library. This means that before, things like `foo: int = "12"` worked just fine, because there would be automatic coercion if possible.

That is no longer the case, which means choices needed to be made about what the "right" types for things should be. Specifically this affects "ports". We have been using `int` but still mostly using strings everywhere. I made the (breaking) change decision that ports are in fact `int`, and that they needed to be provided as such in `TOML` or when calling the relevant code directly.

Old Config
```toml
server = "some_server"
# These both worked!
port = "1234"
port = 1234
```

New Config
```toml
server = "some_server"
# This works!
port = 1234
# This breaks!
port = "1234
``` 

### Code Changes

* [ ] see `Description` for more details

### Steps to Confirm

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`


ThomasLaPiana:
I'm getting a major error here in terms of the OpenAPI docs generation:

```
PydanticInvalidForJsonSchema: Cannot generate a JsonSchema for core_schema.PlainValidatorFunctionSchema ({'type': 'no-info', 'function': <function validate_fides_key at 0x000002518D3CB670>})
```

I'm going to go back to the Fideslang side and add tests to verify JSON schema generation

pattisdr:
Replacing with https://github.com/ethyca/fides/pull/5020 but will be consulting this original work too, thank you for your work here Thomas -

PR #4447, opened 2023-11-20
Call GPP via code splitting

allisonking:
Closes https://ethyca.atlassian.net/browse/PROD-1395

### Description Of Changes

Integrates GPP as a dynamically imported JS package. This means it isn't called until it is needed so isn't included in the initial bundle size. This is also a good proof of concept for how we can implement code splitting in our bundle to split out our UI code from our core initialization code.

#### Caveats

This works pretty well, but there are some warnings in the Privacy Center logs which say `Critical dependency: the request of a dependency is an expression`. After some googling, this appears to be a warning that is okay to ignore: https://github.com/vercel/next.js/issues/10633

One way to remove the warning is to force the variable where we cast the [GPP bundle parameter to a string](https://stackoverflow.com/a/73359606) but this actually causes the app not to compile. It tries to import typescript files which it can't read, and I haven't figured out why that might be. So I left the warning in, though I can't say I totally understand it. 

### Code Changes

* [x] Set up a new route in privacy center called `/gpp-ext` 
* [x] Set up the gpp code as a separate output of rollup which copies to a place the PC expects (in order to host `/gpp-ext`
* [x] New environment variable `FIDES_PRIVACY_CENTER__IS_GPP_ENABLED` to include the GPP bundle dynamically. In the future, this will be done through a config on the backend, but to get this going, I added the env var. We should remove it down the line, similar to what we did for TCF.
* [x] Cypress tests

### Steps to Confirm

* [ ] Run privacy center via `turbo dev`. This will build the new `gpp-ext.js` output and you should see a block along the lines of 

```
fides-js:build: src/extensions/gpp.ts → dist/gpp-ext.js...
fides-js:build: ✅ gpp-ext.js gzipped size passed maximum size checks (2.86 KB < 15 KB)
fides-js:build: ┌──────────────────────────────────┐
fides-js:build: │                                  │
fides-js:build: │   Destination: dist/gpp-ext.js   │
fides-js:build: │   Bundle Size:  11.04 KB         │
fides-js:build: │   Minified Size:  10.94 KB       │
fides-js:build: │   Gzipped Size:  2.86 KB         │
fides-js:build: │                                  │
fides-js:build: └──────────────────────────────────┘
fides-js:build: copied:
fides-js:build:   dist/gpp-ext.js → ../privacy-center/public/lib/gpp-ext.js
fides-js:build: created dist/gpp-ext.js in 122ms
```

* [ ] Make a note of the gzipped size of the `fides-tcf.js` bundle. It should be around `40.45 KB`
* [ ] Now visit `localhost:3000/fides-js-demo.html?geolocation=fr-id` to get the TCF overlay
* [ ] Open up the console and try to ping the GPP API via `__gpp('ping', (data) => {console.log({data})})`. You should get an error that the function `__gpp` is not defined. This is expected since the bundle wasn't included!

image

* [ ] Now stop your privacy center and restart it with `FIDES_PRIVACY_CENTER__IS_GPP_ENABLED=true`
* [ ] Check the gzipped size of the `fides-tcf.js` bundle. It should be the same as it was before (`40.45 KB`).
* [ ] Now visit `localhost:3000/fides-js-demo.html?geolocation=fr-id` to get the TCF overlay
* [ ] Open up the console and try to ping the GPP API via `__gpp('ping', (data) => {console.log({data})})`. This time you should get a `data` object back!

image

* [ ] In the Network tab, you should also see a new call to `gpp-ext.js`

image

Tada, GPP without increasing the initial bundle size! 

### Pre-Merge Checklist

* [x] Update `CHANGELOG.md`


clients/fides-js/src/components/ConsentContent.tsx
```diff
@@ -4,7 +4,7 @@ import { ExperienceConfig } from "../lib/consent-types";
 
 import GpcInfo from "./GpcInfo";
 import ExperienceDescription from "./ExperienceDescription";
-import { getConsentContext } from "../fides";
+import { getConsentContext } from "../lib/consent-context";
```
allisonking, line 7: this import was causing a warning in the build step. I guess it prefers the full path relative import

clients/fides-js/src/extensions/setup.ts
```diff
@@ -0,0 +1,16 @@
...
+
+/**
+ * The path where the GPP extension bundle is located,
+ * likely served by the privacy center
+ */
+const GPP_EXT_PATH = "/gpp-ext.js";
+
+export const setupExtensions = (options: FidesOptions) => {
+  if (options.gppEnabled) {
+    import(GPP_EXT_PATH).catch((e) => {
```
allisonking, line 11: this is the dynamic importing step!

clients/fides-js/src/lib/consent-types.ts
```diff
@@ -14,7 +15,6 @@ import type {
   TCFVendorLegitimateInterestsRecord,
   TCFVendorRelationships,
 } from "./tcf/types";
-import { CookieKeyConsent } from "~/lib/cookie";
```
allisonking, line 17: again, causing a build warning because of the absolute path. by importing `type` instead of the import, we can also get around the circular import

clients/privacy-center/pages/api/gpp-js.ts
```diff
@@ -0,0 +1,52 @@
...
+ *             example: |
+ *               (function(){
+ *                 // gpp.js extension bundle...
+ *               )();
+ */
+export default async function handler(
+  req: NextApiRequest,
+  res: NextApiResponse
+) {
+  const gppJsFile = "public/lib/gpp-ext.js";
```
NevilleS, line 31: FWIW: I feel like this an API route here is more complicated than we'd _want_ here, just because next.js will happily serve static files... but at the same time, it's fairly painful to force cache headers on them...

In other words, a potentially simpler way to solve this is:
1. copy that fides-gpp-ext.js file into `public/` instead of `public/lib`
2. use the default hosting of static files from next
3. add the required headers in the next.config.js or similar

clients/fides-js/rollup.config.mjs
```diff
@@ -86,33 +84,36 @@ const SCRIPTS = [
...
 
 /**
  * @type {import('rollup').RollupOptions}
  */
 const rollupOptions = [];
 
-SCRIPTS.forEach(({ name, gzipErrorSizeKb, gzipWarnSizeKb }) => {
+SCRIPTS.forEach(({ name, gzipErrorSizeKb, gzipWarnSizeKb, isExtension }) => {
+  const input = isExtension ? `src/extensions/${name}.ts` : `src/${name}.ts`;
+  const outputFileName = isExtension ? `dist/${name}-ext` : `dist/${name}`;
```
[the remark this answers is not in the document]
NevilleS: In fact, here's my fully formed nitpick: let's call these `fides-ext-*.js`, like `fides-ext-gpp.js` (and potentially doing things like `fides-ext-ui.js`, `fides-ext-iab.js`, etc.)

NevilleS (changes requested):
This is super promising! I have some nitpicky comments about filenames (mostly!) but my main question is: will this work once it's deployed? 😄

clients/fides-js/src/lib/initialize.ts
```diff
@@ -361,6 +362,11 @@ export const initialize = async ({
     });
   }
 
+  // Call extensions
+  // DEFER(PROD#1238): This is likely too late for the GPP stub.
+  // We should move stub code out to the base package and call it right away instead.
+  setupExtensions(options);
```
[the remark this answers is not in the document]
allisonking: 👍 that should be covered by this one: https://ethyca.atlassian.net/browse/PROD-1439

clients/privacy-center/pages/api/gpp-js.ts
allisonking: oh good point, that makes things a lot simpler!

allisonking: hm actually there seem to be some open issues about adding cache headers in the public folder: https://github.com/vercel/next.js/issues/22319

will look into some of the workarounds on monday...

clients/fides-js/src/extensions/setup.ts
allisonking: hm yeah I think it'll try to use example.com/gpp-ext.js. how do we want to treat this? should we try to force it to look relative, or allow the whole path to be configurable?

clients/privacy-center/pages/api/gpp-js.ts
allisonking: https://stackoverflow.com/questions/62077589/setting-cache-control-header-for-static-file-serving-on-nextjs-default-server

allisonking:
I'm running into some issues merging main because of this bit that was just introduced in a separate PR:

https://github.com/ethyca/fides/blob/1c500ababdf57033d1d1f215c203e3bf2a35cff5/clients/fides-js/src/lib/tcf/events.ts#L19-L22

This PR also has GPP calling the same function, however I'm unclear about how to get the GPP extension file to also have access to `window.Fides`. I know it has something to do with the rollup `output` here

https://github.com/ethyca/fides/blob/fae7acfc9c0e0be9d92b47672369b17254c941f9/clients/fides-js/rollup.config.mjs#L105-L113

But I'm not sure the right way to tell the GPP extension to also use the same `window.Fides` from one of the base `fides.js` files

allisonking:
> I'm running into some issues merging main because of this bit that was just introduced in a separate PR:

Figured it out! I just needed to make sure my GPP ext also had the same window declaration in order to access the `Fides` object.

clients/fides-js/src/lib/initialize.ts
```diff
@@ -361,6 +362,11 @@ export const initialize = async ({
     });
   }
 
+  // Call extensions
+  // DEFER(PROD#1439): This is likely too late for the GPP stub.
+  // We should move stub code out to the base package and call it right away instead.
+  await setupExtensions(options);
```
allisonking, line 368: I just changed this to `await` because otherwise `FidesInitialized` might be called before the extension is ready to listen for it. I'm not sure if this is okay though in terms of initialization times—the first `FidesInitialized` from an existing cookie will already have fired, so this is counting on the second one to fire. It would happen after the UI initializes, which might be far enough away from initial load times?

NevilleS (approved):
Overall: I'd say ship it! I think you should add a config option here though to experiment with the import path.

PR #4450, opened 2023-11-21
Improve Cypress tests for locking system forms for GVL systems

jpople:
Closes PROD-1343

### Description Of Changes

Updates tests for locking system form when a GVL vendor is selected, and expand test coverage to also test whether adding, deleting and editing data uses is locked (it should be).

### Code Changes

* Adds new tests for changes to data uses and for selecting non-GVL vendors
* Add a non-GVL-vendor test fixture and add privacy declarations to the fixture in `dictionary-system.json`

### Pre-Merge Checklist

* [ ] Update `CHANGELOG.md`


clients/admin-ui/cypress/e2e/systems-plus.cy.ts
```diff
@@ -125,6 +125,49 @@ describe("System management with Plus features", () => {
...
+
+    it("locks editing for a GVL vendor when TCF is enabled", () => {
+      cy.getSelectValueContainer("input-vendor_id").type("Aniview{enter}");
+      cy.getByTestId("locked-for-GVL-notice");
+      cy.getByTestId("input-description").should("be.disabled");
+    });
+
+    it("does not allow changes to data uses when locked", () => {
+      cy.getSelectValueContainer("input-vendor_id").type("Aniview{enter}");
+      cy.fixture("systems/dictionary-system.json").then((dictSystem) => {
```
allisonking, line 137: do we need the double fixture call here? I think this was originally in order to simulate saving so we could test the dirty form bug. for the most part we shouldn't need this kind of complicated fixture call except for situations like that, so if we can avoid it here, I would simplify this where possible. @galvana [did this too](https://github.com/ethyca/fides/pull/4435/files#r1399415219), so maybe I'm the one that's wrong and it needs to be here!

jpople: Nope, you're right, these can be safely removed.

allisonking (approved):
🙌

PR #4453, opened 2023-11-22
Initialize TCF earlier and set gdpr to apply always

allisonking:
Closes https://ethyca.atlassian.net/browse/PROD-1433

### Description Of Changes

### Code Changes

### Steps to Confirm

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`


clients/fides-js/src/lib/tcf/stub.ts
```diff
@@ -19,7 +19,11 @@ const isMessageData = (data: unknown): data is MessageData =>
 export const makeStub = () => {
   const queue: any[] = [];
   const currentWindow = window;
-  let gdprApplies: boolean;
+  /**
+   * Fides modification (PROD#1433): gdprApplies defaults to true
+   * since we only return TCF when gdpr does apply!
+   */
+  let gdprApplies: boolean = true;
```
NevilleS, line 26: One ask: for our future benefit, let's do this as an override option `fides_tcf_gdpr_applies` and default it to true there. This gives us an out if we need one 👍

NevilleS (changes requested):
Thanks! Quick functionality ask for some future proofing

clients/fides-js/src/fides-tcf.ts
```diff
@@ -241,6 +242,7 @@ const updateFidesCookieFromString = (
  * Initialize the global Fides object with the given configuration values
  */
 const init = async (config: FidesConfig) => {
+  makeStub();
```
eastandwestwind, line 245: 👍

clients/fides-js/src/lib/tcf.ts
```diff
@@ -192,30 +194,29 @@ export const initializeTcfCmpApi = () => {
     },
   });
 
-  // `null` value indicates that GDPR does not apply
   // Initialize api with TC str, we don't yet show UI, so we use false
   // see https://github.com/InteractiveAdvertisingBureau/iabtcf-es/tree/master/modules/cmpapi#dont-show-ui--tc-string-does-not-need-an-update
   window.addEventListener("FidesInitialized", (event) => {
     const tcString = fidesEventToTcString(event);
-    cmpApi.update(tcString ?? null, false);
```
allisonking, line 200: `fidesEventToTcString` now does the work of defaulting to a value. since if there's a tc string at all (including an empty string) the library sets `gdprApplies = true`. so `fidesEventToTcString` additionally checks if gdpr applies first. this might not actually be what we want though since it can effectively turn off setting a tc string at all. perhaps the override should just be for the initial default value? I did this at first, but I couldn't actually test it since the period where there's a default value until this line was too short for cypress to catch

allisonking: maybe only `FidesInitialized` should do the check for `fides_tcf_gdpr_applies` and the other events should stick to what they were?

clients/privacy-center/cypress/fixtures/consent/test_banner_options.json
```diff
@@ -102,6 +102,7 @@
     "isGeolocationEnabled": false,
     "geolocationApiUrl": "",
     "privacyCenterUrl": "http://localhost:3000",
-    "fidesApiUrl": "http://localhost:8080/api/v1"
+    "fidesApiUrl": "http://localhost:8080/api/v1",
+    "fidesTcfGdprApplies": true
```
allisonking, line 106: since this is an option that defaults to true, and because our test harness doesn't go through `fides-js` the API (which defaults this value to true), we set the default to true here.

clients/fides-js/src/lib/tcf.ts
NevilleS: Hmmm... a philosophical question eh?

I would recommend we check for that option override in _one_ place, so I think it's OK to centralize into the helper function that extracts the string.

It's a bit weird that the `CmpApi` class sets GDPR applies based on a string though. Looking at their docs a bit to form a stronger POV

NevilleS (commented):
Leaving a quick comment now but doing a bit more reading!

clients/fides-js/src/lib/tcf.ts
NevilleS: OK, their docs are super explicit: https://github.com/InteractiveAdvertisingBureau/iabtcf-es/blob/master/modules/cmpapi/README.md#trigger-change-event
> update() may be called either an encoded [TC string](https://github.com/InteractiveAdvertisingBureau/GDPR-Transparency-and-Consent-Framework/blob/master/TCFv2/IAB%20Tech%20Lab%20-%20Consent%20string%20and%20vendor%20list%20formats%20v2.md#creating-a-tc-string) an empty string ('') or null.
> - Encoded TC string, CmpApi will decode the string and respond to [TCData](https://github.com/InteractiveAdvertisingBureau/GDPR-Transparency-and-Consent-Framework/blob/master/TCFv2/IAB%20Tech%20Lab%20-%20CMP%20API%20v2.md#tcdata) with the decoded values.
>    - gdprApplies will be set to true
> - Empty string (''), CmpApi will respond to [TCData](https://github.com/InteractiveAdvertisingBureau/GDPR-Transparency-and-Consent-Framework/blob/master/TCFv2/IAB%20Tech%20Lab%20-%20CMP%20API%20v2.md#tcdata) with the correct structure but all primitive values will be empty.
>    - gdprApplies will be set to true
> - null, CmpApi will respond to [TCData](https://github.com/InteractiveAdvertisingBureau/GDPR-Transparency-and-Consent-Framework/blob/master/TCFv2/IAB%20Tech%20Lab%20-%20CMP%20API%20v2.md#tcdata) with the correct structure but all primitive values will be empty.
>    - gdprApplies will be set to false

In that case I think we can document this more clearly but I agree with your implementation 👍

clients/fides-js/src/lib/tcf/events.ts
```diff
@@ -7,6 +7,10 @@ import { FIDES_SEPARATOR } from "./constants";
  * `vendors_disclosed` and our own AC string addition.
  */
 export const fidesEventToTcString = (event: FidesEvent) => {
+  if (!window.Fides.options.fidesTcfGdprApplies) {
+    // A `null` TC string means gdpr does not apply
+    return null;
+  }
```
NevilleS, line 13: Three nitpicky suggestions:
* rename to `extractTCStringForCmpApi`, to be clear that this is used specifically for the CmpApi
* add a return type of `string | null` for type checks 👍 
* add a comment to link to the IAB docs

Putting those together (no guarantee this compiles!):
```suggestion
* `vendors_disclosed` and our own AC string addition.
 *
 * Returns a string to be used for the `CmpApi` class from `@iabtechlabtcf/cmpapi`, which expects one of three input strings:
 *  - Encoded TC string (e.g. "CP1sGZVP1..."), which will be decoded into `TCData` and set `gdprApplies = true`
 *  - Empty string (e.g. `""`), which will populate an empty `TCData`, but importantly set `gdprApplies = true`
 *  - Null string (e.g. `null`), which will also populate an empty `TCData`, but set `gdprApplies = false`
 * 
 * Therefore we should only use a `null` string when we specifically want to set `gdprApplies = false`!
 *
 * See the `CmpApi` docs for more details: https://github.com/InteractiveAdvertisingBureau/iabtcf-es/blob/master/modules/cmpapi/README.md#trigger-change-event
 */
export const extractTCStringForCmpApi = (event: FidesEvent): string | null => {
  if (!window.Fides.options.fidesTcfGdprApplies) {
    // A `null` TC string will instruct CmpApi to set gdprApplies = false
    return null;
  }
```

NevilleS (approved):
Nitpicks (including a renamed function!) but otherwise this looks good.

PR #4464, opened 2023-11-28
Backend TCF Purpose Override Support

pattisdr:
Closes #PROD-1426

### Description Of Changes

Adds global support for Purpose Overrides.

When enabled, a customer can opt to exclude a purpose altogether or specify a legal basis to be set for that purpose globally.  This should affect the contents of the TCF Experience.  Instead of storing this override at the System/Privacy Declaration level, we calculate it when building the TCF Experience.

### Code Changes
* [ ] Adds a new consent `override_vendor_purposes` config
* [ ] Adds a `tcf_purpose_overrides` table to store overrides for the 11 Purposes
* [ ] Adds default override records to the database on startup (which don't have an effect until enabled later), and even then there is no effect unless customers actually specify override details
* [ ] Adds `purpose` hybrid properties to PrivacyDeclaration since we don't store that directly but this is needed for joins
* [ ] Update the base query that is used to build the TCF Experience.  Join a subquery that contains the legal basis override.  Update the response that this method returns since we also need to return legal basis filters because they touch columns on the subquery. We can't generate them after-the-fact
* [ ] Adds a separate method on the Privacy Declaration for use in returning this in the System response

### Steps to Confirm

* [ ] See in action here https://github.com/ethyca/fidesplus/pull/1240

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`


src/fides/api/models/sql_models.py
```diff
@@ -514,6 +532,124 @@ def create(
...
+                    cls._publisher_override_is_included_join.is_(False),
+                    None,
+                ),
+                (
+                    cls._publisher_override_legal_basis_join.is_(None),
+                    cls.legal_basis_for_processing,
+                ),
+            ],
+            else_=cls._publisher_override_legal_basis_join,
+        )
```
pattisdr, line 651: This is the primary query change that is used to update the Privacy Declaration query which downstream affects the contents of the TCF Experience.   The `@overridden_legal_basis_for_processing.expression` denotes it as a class-level hybrid property.  

This `overridden_legal_basis_for_processing` becomes the legal basis that is filtered for all intents and purposes downstream.  So if we shouldn't override it, we default to the regular legal basis. If it shouldn't be included at all, we default to None.  If we should override it, we return the legal basis specified in the override table.

src/fides/api/models/sql_models.py
```diff
@@ -514,6 +532,124 @@ def create(
...
+
+        overridden_legal_basis: Optional[str] = (
+            await self._publisher_override_legal_basis_join
+        )
+
+        return (
+            overridden_legal_basis
+            if overridden_legal_basis  # pylint: disable=using-constant-test
+            else self.legal_basis_for_processing
+        )
```
pattisdr, line 622: This is the matching instance-level hybrid property that is used for supplementing the system-level response so we can return the overridden legal basis via API. 

Some things I don't like here, since I need to join with the publisher overrides, I need a session, and I only have the async session available to me in the GET* /system contexts.    When FastAPI serializes the response, you get issues with coroutines not being awaited, so instead, I am adding a property of a different name to the PrivacyDeclarationResponse and just stashing it on the PrivacyDeclaration at runtime myself before the response is serialized.

src/fides/api/models/sql_models.py
```diff
@@ -514,6 +532,124 @@ def create(
...
+            TCFPublisherOverride.purpose == self.purpose
+        )
+        async_session: AsyncSession = async_object_session(self)
+        async with async_session.begin():
+            result = await async_session.execute(query)
+            return result.scalars().first()
+
+    @_publisher_override_is_included_join.expression
+    def _publisher_override_is_included_join(cls) -> ScalarSelect:
+        """Returns the class-level indication of whether the purpose should be included"""
```
pattisdr, line 587: These are supporting instance-level and class-level hybrid properties for the primary overridden_legal_basis_for_processing properties below

src/fides/api/api/v1/endpoints/system.py
```diff
@@ -412,3 +426,16 @@ def instantiate_connection_from_template(
...
+    """At runtime, adds a "legal_basis_for_processing_override" to each PrivacyDeclaration"""
+
+    for privacy_declaration in (
+        resp.get("privacy_declarations")
+        if isinstance(resp, Dict)
+        else resp.privacy_declarations
+    ):
+        privacy_declaration.legal_basis_for_processing_override = (
+            await privacy_declaration.overridden_legal_basis_for_processing
+        )
```
pattisdr, line 441: I don't love this but am struggling to get async sessions to work with instance-level properties.   At any rate, the queries aren't efficient in these endpoints already and this doesn't help.  Because trying to return `privacy_declaration.overridden_legal_basis_for_processing` has errors with FastAPI code not awaiting the result, I am awaiting it here instead and stashing it on the PrivacyDeclaration under a different label.

src/fides/api/util/tcf/tcf_experience_contents.py
```diff
@@ -171,21 +175,28 @@ def get_matching_privacy_declarations(db: Session) -> Query:
...
+            PrivacyDeclaration.overridden_legal_basis_for_processing.label(  # pylint: disable=no-member
+                "legal_basis_for_processing"
+            ),
             PrivacyDeclaration.features,
             PrivacyDeclaration.retention_period,
+            PrivacyDeclaration.flexible_legal_basis_for_processing,
+            PrivacyDeclaration.purpose,
+            PrivacyDeclaration.legal_basis_for_processing.label(
+                "original_legal_basis_for_processing"
+            ),
```
pattisdr, line 187: Can remove don't need these in the result

src/fides/api/util/tcf/tcf_experience_contents.py
```diff
@@ -171,21 +175,28 @@ def get_matching_privacy_declarations(db: Session) -> Query:
             System.privacy_policy.label("system_privacy_policy"),
             System.vendor_id,
             PrivacyDeclaration.data_use,
-            PrivacyDeclaration.legal_basis_for_processing,
+            PrivacyDeclaration.overridden_legal_basis_for_processing.label(  # pylint: disable=no-member
+                "legal_basis_for_processing"
+            ),
```
pattisdr, line 180: I am labeling the "overridden legal basis" as the "legal_basis_for_processing" because at this point, that field combines the original field with the override and contains everything we need. It also meant we needed very little changes downstream.

pattisdr: Pros and cons to this though

src/fides/api/util/tcf/tcf_experience_contents.py
```diff
@@ -171,21 +175,28 @@ def get_matching_privacy_declarations(db: Session) -> Query:
...
                 and_(
                     GVL_DATA_USE_FILTER,
-                    PrivacyDeclaration.legal_basis_for_processing
+                    PrivacyDeclaration.overridden_legal_basis_for_processing
                     == LegalBasisForProcessingEnum.CONSENT,
                 ),
                 and_(
                     GVL_DATA_USE_FILTER,
-                    PrivacyDeclaration.legal_basis_for_processing
+                    PrivacyDeclaration.overridden_legal_basis_for_processing
```
pattisdr, line 199: Changes were simple here once the `overridden_legal_basis_for_processing` class-level hybrid property was working.

src/fides/api/models/tcf_publisher_overrides.py
```diff
@@ -0,0 +1,23 @@
...
+
+    @declared_attr
+    def __tablename__(self) -> str:
+        return "tcf_publisher_overrides"
+
+    purpose = Column(Integer, nullable=False)
+    is_included = Column(Boolean, server_default="t", default=True)
+    required_legal_basis = Column(String)
+
+    UniqueConstraint("purpose")
```
pattisdr, line 23: Assumed purposes, not special purposes were being toggled here, from the mockups.  We can't save preferences against a special purpose x legal basis, only purpose x legal basis, so I assume that's the distinction. Still, someone may want to suppress the special purpose altogether? In that case I might add another column here that clarifies if the record is for a purpose or special purpose.

src/fides/api/schemas/system.py
```diff
@@ -19,6 +19,10 @@ class PrivacyDeclarationResponse(PrivacyDeclaration):
     )
     cookies: Optional[List[Cookies]] = []
 
+    legal_basis_for_processing_override: Optional[str] = Field(
+        description="Global overrides for this purpose's legal basis for processing if applicable.  Defaults to the legal_basis_for_processing otherwise."
+    )
```
pattisdr, line 24: This is the property I added to include this value in the System response.  As noted elsewhere, it differs by name from the hybrid property on the Privacy Declaration due to async issues.

src/fides/config/consent_settings.py
```diff
@@ -15,6 +15,11 @@ class ConsentSettings(FidesSettings):
         default=False, description="Toggle whether Google AC Mode is enabled."
     )
 
+    override_vendor_purposes: bool = Field(
+        default=False,
+        description="Whether or not vendor purposes can be globally overridden.",
+    )
```
pattisdr, line 21: Added a temporary config setting here because I needed it in my query although this overlaps with @TheAndrewJackson's work. Happy to update to match what he's doing if he needs to go a different direction -

pattisdr:
@adamsachs when you get a chance, could you do an early pass of this on just the approach? 

Namely, I add a new hybrid class-level property with all the legal basis override logic that adds the appropriate "case" statement to the Privacy Declaration query used by the TCF Experience.   

So as requested, the overrides aren't stored on the System themselves, but the override is still calculated fairly early on in the process so it is used downstream in 1) The TCF Experience 2) Building backend TC strings 3) Calculating "relevant_systems" which are stored for consent reporting when consent is saved.

I like this approach, the main thing I don't like about this PR is the messiness of including the override property in the System response.

It can be tested in this Plus PR https://github.com/ethyca/fidesplus/pull/1240 if interested

src/fides/api/api/v1/endpoints/system.py
adamsachs: hmm. the major performance concerns should hopefully be mitigated pretty substantially by having a paged version of the endpoint that's relied upon for the UI (a separate ticket/unit of work), but yeah i see your qualms.

i do think that providing optionality (i.e. an endpoint query param) about whether this computed property is even returned at all could be another nice gate to put in place.

i'm also not 100% sure whether the `/consent/configure` page in the UI is still using our basic `/system` APIs - i know there had been some talk about making a specific endpoint for that page when we had some advanced filtering requirements. i think those requirements are no longer in place, but it's worth confirming that we're still planning to use this endpoint to back that UI. 

i'm getting a bit carried away in the details, but i think the major points are:
- this seems like a viable approach, at least as a starting point, whichever API this function ends up being used in
- let's keep an eye on the performance concerns, but they probably (hopefully?) will won't be too significant compared to other performance issues already in place on these endpoints, as you mention.
- if it adds nontrivial overhead, may be nice to provide optionality about whether the property is computed via an API param, since this is only needed for very particular purposes
- we should generally make sure we're all in sync on what endpoints are meant to back what APIs, there's been a lot of movement around all of this work over the past couple of weeks (cc @TheAndrewJackson @galvana)

adamsachs: ~also, a bit of a separate point - i could be missing something in the details, but do we have a mechanism for providing whether a purpose is broadly "excluded"? i.e. from what i can tell, here we are just computing the `overridden_legal_basis_for_processing`, but i don't think that provides a signal to the FE in the case where the purpose is just generally _excluded_ via an override?~

nevermind, i think i get it - if this is `None`/null, that indicates an exclusion, right?

src/fides/api/util/tcf/tcf_experience_contents.py
adamsachs: i think i like this, i.e. my intuition is the "earlier" we cut it off the better 👍 but am open to hearing your take on the cons :)

src/fides/api/models/sql_models.py
adamsachs: ok, i think i've got it. let's definitely include this behavior in the docstring for the method and/or other places.

it is a bit counterintuitive to me that the general "exclusion" signal is coming through on the `overridden_legal_basis_for_processing` property, since the "exclusion" applies to the purpose (`PrivacyDeclaration`) as a whole, not its legal basis. but i understand that this is a pretty _efficient_ way to represent the different states here, and it plugs in seamlessly to existing downstream code. i'm just a bit concerned about maintainability, i.e. that we may be hiding some behavior by overloading this property. this is related to my comment below.

i'm still thinking through this! maybe i'm not conceptualizing it properly.

src/fides/api/util/tcf/tcf_experience_contents.py
adamsachs: i see - so the fact that this is `None` in the case of an exclusion override naturally filters those out.

this does seem  _efficient_ (in terms of being a very targeted change), i just get a bit concerned about that behavior being opaque moving forward. relates to my comment above, still thinking about this 🤔

adamsachs (commented):
@pattisdr broadly, i really like the approach. i'll outline the broad points that i liked, and then the couple of concerns i've got. those aren't meant to be prescriptive, just for consideration!

what i like:
- the data model for the overrides looks as i'd expect, you've captured the requirements there efficiently and clearly
- impressive work on the sqlalchemy magic (would it be fair to call it "alchemy"? 😬), especially with some of the async paradigms! i really like how much of this is encapsulated within the model, keeps the logic centralized and easy to parse (or adjust as needed)
- it's great how few changes are needed to the actual TCF overlay generation code. filtering as early as possible, and putting the computations on the model, keeps us from adding complexity to the overlay generation code itself, and i think that's a big win!

the broad concerns i have:
- as you've pointed out, performance! i frankly have no idea how all the (composite) hybrid properties affect performance, or really how they get translated into underlying DB queries. this probably deserves its own discussion on the next iteration of the PR -- and i'd like to get a better understanding of the functionality generally - so i won't try to tackle it all here...but generally:
    - i think we can live with some performance hits on the APIs that return `System` information to be used in the consent mgmt UI, since i think/hope we'll have some paged endpoints to support that UI (need to sync up on that, as a separate matter).
    - i'd focus my initial attention on whether the TCF contents performance is significantly impacted by this update, as that is not an API that we can easily break apart in any way. maybe you've already verified it, or maybe my concern around this is totally misguided because i just don't have a great mental model for how the composite hybrid properties are worked into the db queries in the ORM layer.
- i've got a slight concern for maintainability purposes that we're overloading the `overridden_legal_basis_for_processing` property to communicate information about exclusions on the level of a `PrivacyDeclaration` along with legal basis information. i see there are benefits to how you've implemented this, but (at least in how i'm thinking about it now), i get a bit worried it may be unclear if/when we want to come back to this down the road. some more details in the comments, but would be happy to talk this over separately to get your thoughts.

pattisdr:
Thanks very much for your thoughtful comments @adamsachs ⭐

src/fides/api/util/tcf/tcf_experience_contents.py
pattisdr: Just there’s a lot of logic being encapsulated under this property and then also relabeling it here as an existing attribute can obscure the truth.

But I thought that this fit well with the existing fundamental that nothing shows up in the TCF experience without a legal basis of consent or LI, which made setting it to null fit really nicely with this.  This property covers what would happen if we were using the data map instead to store the override directly. 

Maybe code comments and renaming so the intent is more clear will help.

src/fides/api/models/sql_models.py
pattisdr: In my latest draft, I've removed all the hybrid properties except for "purpose" which I think still makes sense. I'm instead joining the primary privacy declaration query with a new subquery that is very similar to this logic.  So less is living on the model now but the query generated by SQLAlchemy is the one I'd expect.

pattisdr: I still think capturing everything in one place, combining the different states into one new attribute makes sense for the TCF experience, going along with our original rule that a declaration needs a consent or LI legal basis to show up at all.  Because this logic no longer lives on the model, I'm not as concerned about overloading it, since this is now written very specifically for the TCF overlay.

src/fides/config/consent_settings.py
pattisdr: Correct yes, I thought you said that Andrew was doing that as part of https://ethyca.atlassian.net/browse/PROD-1414?

src/fides/api/util/tcf/tcf_experience_contents.py
```diff
@@ -149,12 +143,80 @@ class TCFExperienceContents(
...
+                            False
+                        ),
+                        PrivacyDeclaration.legal_basis_for_processing,
+                    ),
+                    (
+                        TCFPurposeOverride.is_included.is_(False),
+                        None,
+                    ),
+                    (
+                        TCFPurposeOverride.required_legal_basis.is_(None),
```
pattisdr, line 183: This is the new primary location where the purpose overrides are applied.

 Instead of using a hybrid property (which generated an awful SQL query, good call on checking that Adam), I'm joining with a subquery that has similar logic (and generates the SQL query I hoped the hybrid property would generate)

src/fides/api/util/tcf/tcf_experience_contents.py
```diff
@@ -171,22 +233,27 @@ def get_matching_privacy_declarations(db: Session) -> Query:
...
+            PrivacyDeclaration.purpose,
+            PrivacyDeclaration.legal_basis_for_processing.label(
+                "original_legal_basis_for_processing"
+            ),
         )
         .outerjoin(PrivacyDeclaration, System.id == PrivacyDeclaration.system_id)
+        .outerjoin(
+            legal_basis_override_subquery,
+            legal_basis_override_subquery.c.id == PrivacyDeclaration.id,
+        )
```
pattisdr, line 250: This is now the primary change of this PR . I join the base query to the legal basis override subquery, and then below filter on the legal basis override instead of the original legal basis.

src/fides/api/util/tcf/tcf_experience_contents.py
```diff
@@ -201,7 +268,12 @@ def get_matching_privacy_declarations(db: Session) -> Query:
         matching_privacy_declarations = matching_privacy_declarations.filter(
             NOT_AC_SYSTEM_FILTER
         )
-    return matching_privacy_declarations
+
+    return (
+        matching_privacy_declarations,
+        consent_legal_basis_filter,
+        legitimate_interest_legal_basis_filter,
```
pattisdr, line 275: These filters need to be passed forward to further filter the various sections of the TCF Privacy Experience on legal basis.  We can't generate them after the fact because they are filters on the fields in the subquery.  They are later chained onto the base "matching_privacy_declarations" filter. in plus.

src/fides/api/models/sql_models.py
```diff
@@ -514,6 +529,81 @@ def create(
         """Overrides base create to avoid unique check on `name` column"""
         return super().create(db=db, data=data, check_name=check_name)
 
+    @hybrid_property
+    def purpose(self) -> Optional[int]:
```
adamsachs, line 533: i'm sort of curious how these get compiled into the generated SQL queries! if you have something handy i'd love to see, but no need to go out of your way.

src/fides/api/util/tcf/tcf_experience_contents.py
adamsachs: nice, yeah this is still relatively easy to follow (maybe even easier?), especially considering the complex logic that you're encoding!

src/fides/api/util/tcf/tcf_experience_contents.py
```diff
@@ -149,12 +143,80 @@ class TCFExperienceContents(
...
+        .subquery()
+    )
+
+
+def get_tcf_base_query_and_filters(
+    db: Session,
+) -> Tuple[Query, BinaryExpression, BinaryExpression]:
     """
+    Returns the base query that contains the foundations of the TCF Experience as well as
+    two filters to further refine the query when building the Experience.
```
adamsachs, line 203: yeah, the returning these filters seems like probably the hardest thing here to easily digest, especially because it's going "across repos" 😬 

but that's ok, and having the special note here in the docstring helps!

adamsachs (approved):
@pattisdr this is also looking great, i like where it's landed! great job on this, the end result is pretty simple given the trickiness involved in fitting this into our existing framework.

just a couple of nits and some noncritical questions that i'd like to understand at some point, but on the whole, i don't think this needs any significant changes!

src/fides/api/alembic/migrations/versions/5225ea4de265_tcf_purpose_overrides.py
```diff
@@ -0,0 +1,55 @@
...
+        sa.Column("is_included", sa.Boolean(), server_default="t", nullable=True),
+        sa.Column("required_legal_basis", sa.String(), nullable=True),
+        sa.PrimaryKeyConstraint("id"),
+    )
+    op.create_index(
+        op.f("ix_tcf_purpose_overrides_id"),
+        "tcf_purpose_overrides",
+        ["id"],
+        unique=False,
+    )
```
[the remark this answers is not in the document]
pattisdr: I can! I considered it and thought this few rows wasn't worth it, but not a bad thing to add.

src/fides/api/models/sql_models.py
pattisdr: It's just a CASE statement added to the other SELECT attributes, it's not bad!  We can explore adding purpose too on the declaration at some point in the future. Cons are keeping data use/purpose in sync though.

```sql
SELECT
  ...
  CASE 
    WHEN (privacydeclaration.data_use = %(data_use_1)s) THEN %(param_1)s 
    WHEN (privacydeclaration.data_use = %(data_use_2)s) THEN %(param_2)s 
    WHEN (privacydeclaration.data_use = %(data_use_3)s) THEN %(param_3)s 
    WHEN (privacydeclaration.data_use = %(data_use_4)s) THEN %(param_4)s 
    WHEN (privacydeclaration.data_use = %(data_use_5)s) THEN %(param_5)s 
    WHEN (privacydeclaration.data_use = %(data_use_6)s) THEN %(param_6)s 
    WHEN (privacydeclaration.data_use = %(data_use_7)s) THEN %(param_7)s 
    WHEN (privacydeclaration.data_use = %(data_use_8)s) THEN %(param_8)s 
    WHEN (privacydeclaration.data_use = %(data_use_9)s) THEN %(param_9)s 
    WHEN (privacydeclaration.data_use = %(data_use_10)s) THEN %(param_10)s 
    WHEN (privacydeclaration.data_use = %(data_use_11)s) THEN %(param_11)s 
    WHEN (privacydeclaration.data_use = %(data_use_12)s) THEN %(param_12)s 
    WHEN (privacydeclaration.data_use = %(data_use_13)s) THEN %(param_13)s 
    WHEN (privacydeclaration.data_use = %(data_use_14)s) THEN %(param_14)s 
  END AS purpose,
  ....
```

src/fides/api/util/tcf/tcf_experience_contents.py
pattisdr: This could be alleviated some by moving more of this code over to Plus! There are some operations still here that need this information. I have a ticket in the backlog!

pattisdr:
Thank you for review @adamsachs and for your earlier comments that helped shape this!

src/fides/api/alembic/migrations/versions/5225ea4de265_tcf_purpose_overrides.py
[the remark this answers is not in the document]
pattisdr: Added 👍

PR #4470, opened 2023-12-01
Prevent dismissal of banner and overlay

galvana:
### Description Of Changes

Adding the `PREVENT_DISMISSAL` config setting to the Privacy Center to allow an admin to prevent a user from dismissing the consent banner and modal.

### Code Changes

* [ ] Updated the `CloseButton` component to accept a `hidden` prop. We want to hide the component but still have it take up space in the layout to prevent visual issues caused by the close button not being rendered
* [ ] Added an overlay div to `Overlay` to serve as the black background for the banner. I originally tried to use `a11y-dialog` for the banner but there were some challenges (how to trigger the banner to show up, slide in animation didn't work anymore)
* [ ] Added ESC and "click-outside" event handlers to the `ConsentBanner` component

### Steps to Confirm

* [ ] Add some systems/vendors to have data available for the consent banners
* [ ] Run `FIDES_PRIVACY_CENTER__PREVENT_DISMISSAL=false turbo run dev` to allow dismissal (default behavior)
* [ ] Verify the banner and modal show up with and without TCF
* [ ] Verify both the banner and modal can be dismissed by hitting ESC, clicking the x, or clicking outside the banner/modal
* [ ] Run `FIDES_PRIVACY_CENTER__PREVENT_DISMISSAL=true turbo run dev` to prevent dismissal
* [ ] Verify the banner and modal show up with and without TCF
* [ ] Verify both the banner and modal can be dismissed by hitting ESC, clicking the x, or clicking outside the banner/modal. Additionally, the banner shouldn't have a black background if dismissal is allowed
 * [ ] Verify the banner and modal show up with and without TCF
 * [ ] Verify both the banner and modal can not be dismissed by hitting ESC, clicking the x (shouldn't be visible), or clicking outside the banner/modal.

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`


clients/fides-js/src/components/Overlay.tsx
```diff
@@ -58,7 +58,7 @@ const Overlay: FunctionComponent<Props> = ({
 
   const { instance, attributes } = useA11yDialog({
     id: "fides-modal",
-    role: "alertdialog",
+    role: window.Fides.options.preventDismissal ? "alertdialog" : "dialog",
```
allisonking, line 61: ooh this works out nicely 😄

clients/fides-js/src/components/Overlay.tsx
```diff
@@ -140,6 +140,9 @@ const Overlay: FunctionComponent<Props> = ({
 
   return (
     <div>
+      {bannerIsOpen && window.Fides.options.preventDismissal && (
+        <div className="fides-modal-overlay" />
```
allisonking, line 144: Instead of the `useEffect` in ConsentBanner, would it work to give this div an `onClick` which closes the banner? I generally try to avoid `ref`s as I find they can get confusing

allisonking (commented):
testing this PR reminded me how difficult modals are 🥲 they have a lot of a11y reqs which are not straightforward to implement, such as focus trapping. there are more listed here from when we first implemented the modal: https://github.com/ethyca/fides/issues/3309

it seems like our banner has become more of a modal now that we can control dismissing it (though maybe I could be convinced that it's not a modal...) therefore I think your original intuition to use the `a11y-dialog` library was right. we might be able to figure the animation out (there are some docs [here](https://a11y-dialog.netlify.app/advanced/animations), for instance which look similar to our use case) though you probably ran into other things too? 

I'm not an a11y expert, but I don't think this implementation would pass the [dialog/modal pattern](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/) and we'd need to spend more time to make it comply. maybe worth discussing with product?

galvana:
Hi @allisonking, this is ready for another pass. I was having trouble getting rid of the `ref` approach in the `ConsentBanner` component so I moved to the Cypress tests instead. I'll give it another shot tomorrow, if time allows, now that we have the tests in place.

galvana:
I'm also not able to reproduce the formatting errors locally when I run `turbo run format:ci`, is that the right command?

allisonking:
> I'm also not able to reproduce the formatting errors locally when I run `turbo run format:ci`, is that the right command?

Yes, but run it in the `fides-js` directory if you aren't already (or in `clients`)

To just run prettier, do `turbo run format` and that'll apply changes

clients/privacy-center/cypress/e2e/banner-overlay-dismissal.cy.ts
```diff
@@ -0,0 +1,358 @@
...
+          cy.get(".fides-close-button").click();
+        });
+        cy.get("#fides-banner").should("not.be.visible");
+
+        cy.get("@FidesUpdated")
+          .should("have.been.calledOnce")
+          .its("lastCall.args.0.detail.extraDetails.consentMethod")
+          .then((consentMethod) => {
+            expect(consentMethod).to.eql(ConsentMethod.dismiss);
+          });
```
allisonking, line 28: let's also assert that there was a PATCH to the backend here

clients/privacy-center/cypress/e2e/banner-overlay-dismissal.cy.ts
```diff
@@ -0,0 +1,358 @@
...
+        cy.get(".fides-modal-content").should("be.visible");
+        cy.get("body").type("{esc}");
+        cy.get(".fides-modal-content").should("not.be.visible");
+
+        cy.get("@FidesUpdated")
+          .should("have.been.calledOnce")
+          .its("lastCall.args.0.detail.extraDetails.consentMethod")
+          .then((consentMethod) => {
+            expect(consentMethod).to.eql(ConsentMethod.dismiss);
+          });
```
allisonking, line 103: might be worth putting this in a function, along with an assertion that the PATCH was called, since we check for it every time. maybe a function like `assertDismissCalled`?

clients/fides-js/src/components/tcf/TcfOverlay.tsx
```diff
@@ -344,6 +348,7 @@ const TcfOverlay: FunctionComponent<OverlayProps> = ({
       onVendorPageClick={() => {
         setActiveTabIndex(2);
       }}
+      onDismiss={handleDismiss}
```
allisonking, line 351: hmm this prop doesn't exist on `Overlay`?

clients/fides-js/src/lib/a11y-dialog.tsx
```diff
@@ -15,8 +15,19 @@ const useA11yDialogInstance = () => {
         .on("show", () => {
           document.documentElement.style.overflowY = "hidden";
         })
-        .on("hide", () => {
+        .on("hide", (_: Element, event?: Event) => {
           document.documentElement.style.overflowY = "";
+
+          // a11y-dialog natively supports dismissing a dialog by pressing ESC
```
allisonking, line 21: 🙌

clients/privacy-center/cypress/e2e/consent-banner-tcf.cy.ts
```diff
@@ -312,7 +312,12 @@ describe("Fides-js TCF", () => {
       cy.intercept("PATCH", `${API_URL}${FidesEndpointPaths.NOTICES_SERVED}`, {
         fixture: "consent/notices_served_tcf.json",
       }).as("patchNoticesServed");
-      cy.get("#fides-modal-link").click();
+
+      cy.get("div#fides-banner").within(() => {
```
allisonking, line 316: maybe a bit too late now, but could be worth wrapping this whole block in a function `openModalFromBanner` and calling that instead of this whole nested thing

allisonking (approved):
nice job sticking with this ticket that was tougher than it looked, @galvana ! is there a followup ticket for porting the banner to a modal? we should also note in that ticket the issue with clicking `#fides-modal-link` while the banner is still open.

left some comments but they're mostly nits, I think this should be okay to merge!

galvana:
Thanks for the review @allisonking, I created a follow up ticket to track the `a11y-dialog` refactor and some of the other minor changes. https://ethyca.atlassian.net/browse/PROD-1518

PR #4480, opened 2023-12-04
Reduce size of `tcf_consent` payload in `fides_consent` cookie

allisonking:
Closes https://ethyca.atlassian.net/browse/PROD-1413

### Description Of Changes

This PR refactors core TCF logic to rely on Fides str (in `cookie.fides_string`) for most TCF consent preferences instead of using our `tcf_consent` obj in the cookie. This reduces the size of our `tcf_consent` payload we send in every HTTP req between client <---> server which reduces risk for any slowdown due to cookie size.

### Code Changes

* [ ] Removes unneeded props from `tcf_consent`, refactor usages to rely on the fides string instead.
* [ ] Add e2e tests

### Steps to Confirm

* [ ] Configure 1k+ systems by bulk adding the entire GVL list
* [ ] Open the TCF banner in a sample app and accept all vendors
* [ ] Inspect the size of the fides_consent cookie. It should be smaller than before
* [ ] Additionally, this should be fully regression tested, for mostly TCF banner / modal, but also for notice-based consent. Confirm consent selections persist correctly when you make a selection on either the banner/modal, then force the modal to open on next page load via "manage preferences" link.

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`


allisonking:
I've updated tests to no longer expect a `tcf_consent` to contain info for purposes, special features, or vendors. The tests that assert for `tcf_consent.system_*` are still there since that key should still exist. 

There are two tests which are failing but which once we implement this ticket _should_ pass. They're labeled with "TODO: CURRENTLY FAILING!!". That said, I wouldn't be surprised if more things came up during implementation that we need to tweak the tests for!

clients/fides-js/src/fides-tcf.ts
```diff
@@ -128,6 +124,82 @@ const getInitialPreference = (
   return tcfObject.default_preference ?? UserConsentPreference.OPT_OUT;
 };
 
+/**
+ * Populates TCF entities with items from cookie.tcf_consent and Fides string.
+ * Returns TCF entities to be assigned to an experience.
```
eastandwestwind, line 129: Our terminology is a bit confusing I realize. We talk about the cookie vs the Fides string, when here they are both already on the cookie at this point. Can we clarify in the comments that `TCF entities are derived from both cookie.tcf_consent and cookie.fides_string. We must look at both because they contain non-overlapping info that is required for a complete tcf entities object.  `

clients/fides-js/src/lib/tcf/constants.ts
```diff
@@ -18,41 +17,46 @@ export const ETHYCA_CMP_ID = 407;
...
   },
   {
-    cookieKey: "vendor_legitimate_interests_preferences",
     experienceKey: "tcf_vendor_legitimate_interests",
     tcfModelKey: "vendorLegitimateInterests",
     enabledIdsKey: "vendorsLegint",
   },
+];
+
+export const TCF_COOKIE_KEY_MAP: {
```
eastandwestwind, line 56: add a comment, something like: we must store these preferences on the cookie outside of the `fides_string` because they are not supported by the FidesString.

clients/fides-js/src/lib/tcf/constants.ts
```diff
@@ -48,6 +51,12 @@ export const TCF_KEY_MAP: {
...
+  cookieKey: TcfModelType;
+  experienceKey: keyof TcfExperienceRecords;
```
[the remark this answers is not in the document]
allisonking: it doesn't seem like anything uses it, probably since we combine `vendors` with `systems` for enabledIds. this variable should maybe be a different name, but I haven't thought of a great one yet. maybe `FIDES_SYSTEM_COOKIE_KEY_MAP`?

clients/fides-js/src/fides-tcf.ts
```diff
@@ -169,8 +240,11 @@ const updateCookieAndExperience = async ({
   }
 
   // If the user has prefs on a client-side fetched experience, but there is no fides_string,
-  // we need to use the prefs on the experience to generate a fidesString and cookie.tcf_consent
-  const tcSavePrefs: TcfSavePreferences = {};
+  // we need to use the prefs on the experience to generate
+  // 1. a fidesString
+  // 2. a cookie.tcf_consent (which only has system preferences since those are not captured in the fidesString)
+
+  // 1. Generate a fidesString from the experience
```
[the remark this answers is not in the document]
allisonking: ah yeah I was trying to summarize the steps needed, and then label the steps... maybe there's a better way to do that

clients/fides-js/src/lib/cookie.ts
```diff
@@ -408,12 +302,12 @@ export const updateExperienceFromCookieConsent = ({
     return { ...notice, current_preference: preference };
   });
 
-  // Handle the TCF case, which has many keys to query
-  const tcfEntities = buildTcfEntitiesFromCookieAndFidesString(
-    experience,
-    cookie,
-    fidesString
-  );
+  // // Handle the TCF case, which has many keys to query
```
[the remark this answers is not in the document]
allisonking: ah yes!

clients/fides-js/src/fides-tcf.ts
eastandwestwind: updated

clients/fides-js/src/lib/tcf/constants.ts
eastandwestwind: updated

clients/fides-js/src/fides-tcf.ts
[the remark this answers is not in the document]
eastandwestwind: oh I see ! The diff was confusing to look at but makes sense 👍

clients/fides-js/__tests__/lib/cookie.test.ts
```diff
@@ -484,165 +444,6 @@ describe("updateExperienceFromCookieConsent", () => {
       ]);
     });
   });
-
-  describe("tcf", () => {
```
eastandwestwind, line 488: Had to remove these tests because `updateExperienceFromCookieConsent` actually needs to use the new `updateExperienceFromCookieConsentTcf` method.

I had some issues with jest / preact exporting it from `fides-tcf.ts`, but decided to remove it as a) this is already covered by our e2e tests and b) `updateExperienceFromCookieConsentTcf` exists in `fides-tcf.ts` and is therefore not a candidate for a `cookie.ts` unit test.

clients/fides-js/src/fides-tcf.ts
```diff
@@ -263,35 +236,32 @@ const init = async (config: FidesConfig) => {
...
+          overrides.consentPrefsOverrides?.version_hash ??
+          cookie.tcf_version_hash,
+      };
       Object.assign(cookie, updatedCookie);
+    } catch (error) {
+      debugLog(
+        config.options.debug,
+        `Could not decode tcString from ${fidesString}, it may be invalid. ${error}`
+      );
     }
```
NevilleS, line 258: It seems like this got... a lot simpler! That's actually a pretty good sign 👍

clients/fides-js/src/fides-tcf.ts
```diff
@@ -263,35 +236,32 @@ const init = async (config: FidesConfig) => {
...
-    // we should generate fidesString so that our CMP API accurately reflects user preference
-    cookie.fides_string = await generateFidesStringFromCookieTcfConsent(
-      config.experience,
-      cookie.tcf_consent
-    );
-    debugLog(
-      config.options.debug,
-      "fides_string was missing from cookie, so it has been generated based on tcf_consent",
-      cookie.fides_string
-    );
```
NevilleS, line 292: Totally makes sense that this case is removed now. Nice.

NevilleS (approved):
Nice... let's ship it.

clients/fides-js/src/fides-tcf.ts
allisonking: yes, I agree! was a pleasant surprise 😄

PR #4481, opened 2023-12-04
Backend Overhaul: Data Model Changes for Saving Privacy Preferences

pattisdr:
Closes #PROD-1483

❗ Contains schema and data migration; bump downrev before merge
❗ Should wait for all subtasks in https://ethyca.atlassian.net/browse/PROD-1477
👉  When getting this PR up-to-date with main, check what changes and tests need to be moved to Plus.
👉 Corresponding Fidesplus PR https://github.com/ethyca/fidesplus/pull/1249

### Description Of Changes

We're updating how we save **privacy preferences** and **notices served** on the backend to improve performance and reduce the number of records created on save.

This PR adds new models, strips out obsolete code, and moves Served Notice Endpoints, Consent Reporting Endpoints, and TCF schemas and base query to Plus to support this effort.

### Code Changes

* [x] Adds two new tables CurrentPrivacyPreferenceV2 and LastServedNoticeV2 for storing the latest preferences saved for Notices and the notices served for a user 
* [x] Update existing PrivacyPreferenceHistory and ServedNoticeHistory tables to track more details about notices in-line, and track TCF preferences/served directly on the record under a jsonb.
   * [x] New models no longer have FK's to ProvidedIdentities or FK's from the "Last saved" record to the historical record that created.  
   * [x] We now save preferences directly against email/device id/and/or phone number
* [x] Similar v1 tables marked as deprecated
* [x] Gets rid of functionality to supplement Privacy Experience with user's previously-saved preferences for performance and to not migrate over a rarely used code path.
* [x] Gets rid of consent reporting functionality that calculates "relevant systems" to improve performance on privacy preference save (plus we don't need it in the report)
* [x] Moves notices served endpoints, old consent reporting endpoints, purpose endpoints, privacy preference schemas, TCF schemas, and TCF base queries to plus.  It's a lot easier to maintain with TCF and Privacy Preferences/Notices served in one location, not to mention, this move saved at least a day of development in the rewrite.
* [x] Lots of tests moved to plus.
* [x] Removes unused CURRENT_PRIVACY_PREFERENCES_REPORT endpoint
* [x] Data migration https://github.com/ethyca/fides/pull/4511

### Steps to Confirm

* [ ] This should be given a broad testing sweep over in Plus. [My testing notes are here](https://ethyca.atlassian.net/wiki/spaces/EN/pages/2868445187/2023-12-15+Privacy+Preferences+Save+Overhaul+Testing+Notes)

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`


src/fides/api/api/v1/endpoints/privacy_preference_endpoints.py
```diff
@@ -1,422 +0,0 @@
...
-    response_model=Page[ConsentReportingSchema],
-)
-def get_historical_consent_report(
-    *,
-    params: Params = Depends(),
-    db: Session = Depends(get_db),
-    request_timestamp_gt: Optional[datetime] = None,
-    request_timestamp_lt: Optional[datetime] = None,
-) -> AbstractPage[PrivacyPreferenceHistory]:
-    """Endpoint to return a historical record of all privacy preferences saved for consent reporting"""
```
pattisdr, line 312: Endpoint has moved to plus

src/fides/api/models/privacy_experience.py
```diff
@@ -612,70 +534,3 @@ def upsert_privacy_experiences_after_config_update(
...
-        TCFPurposeLegitimateInterestsRecord,
-        TCFSpecialPurposeRecord,
-        TCFFeatureRecord,
-        TCFSpecialFeatureRecord,
-        TCFVendorConsentRecord,
-        TCFVendorLegitimateInterestsRecord,
-        TCFVendorRelationships,
-    ],
-    fides_user_provided_identity: Optional[ProvidedIdentity],
-    record_type: Optional[ConsentRecordType],
```
pattisdr, line 631: Obsolete functionality removed. 

We no longer support `GET /privacy-exerience?fides_user_device=xxxxx`, supplementing an Experience with user's saved TCF preferences.  This is a time-consuming operation and is only used in the privacy center.  Generally we haven't used this functionality in the way we thought we would.

Ticket to remove dependency in FE https://ethyca.atlassian.net/browse/PROD-1485

src/fides/api/models/privacy_experience.py
```diff
@@ -308,62 +285,7 @@ def get_related_privacy_notices(
...
-            notices.append(notice)
-
-        return notices
-
-    def update_with_tcf_contents(
-        self,
-        db: Session,
-        base_tcf_contents: TCFExperienceContents,
-        fides_user_provided_identity: Optional[ProvidedIdentity],
-    ) -> bool:
```
pattisdr, line 331: Method moved out of the model and adapted in Plus - this adds the TCF contents to the Privacy Experience but this move lets us also port more TCF schemas to Plus.

src/fides/api/models/privacy_experience.py
```diff
@@ -308,62 +285,7 @@ def get_related_privacy_notices(
...
-        for notice in privacy_notice_query.order_by(PrivacyNotice.created_at.desc()):
-            cache_saved_and_served_on_consent_record(
-                db=db,
-                consent_record=notice,
-                fides_user_provided_identity=fides_user_provided_identity,
-                record_type=ConsentRecordType.privacy_notice_id,
-            )
-            notices.append(notice)
-
-        return notices
```
pattisdr, line 324: Obsolete functionality removed. We don't add user's previously saved preferences to Notices in the Experience.

src/fides/api/models/privacy_notice.py
```diff
@@ -27,6 +27,7 @@ class UserConsentPreference(Enum):
     opt_in = "opt_in"  # The user wants to opt in to the notice
     opt_out = "opt_out"  # The user wants to opt out of the notice
     acknowledge = "acknowledge"  # The user has acknowledged this notice
+    tcf = "tcf"  # Overall preference set for TCF where there are numerous preferences under the single notice
```
pattisdr, line 30: Now that we capture the save of historical TCF preferences as a single row, instead of one row per TCF attribute, there is no overall "preference" for a TCF record. Neville recommended adding "tcf" so if someone queried this column for the preferences, we'd have no null records - 

To see more detailed preferences you'd look at `PrivacyPreferenceHistoryV2.tcf_preferences` jsonb column

src/fides/api/models/privacy_notice.py
```diff
@@ -178,15 +179,6 @@ class PrivacyNoticeBase:
...
 
-    # Attribute that can be temporarily cached as the result of "get_related_privacy_notices"
-    # for a given user, for surfacing CurrentPrivacyPreferences for the user.
-    current_preference: Optional[str] = None
-    outdated_preference: Optional[str] = None
-    # Attributes that can be temporarily cached on the notice to see if the most
-    # recent version or a previous version of a notice have ever been served to the user
-    current_served: Optional[bool] = None
-    outdated_served: Optional[bool] = None
-
```
pattisdr, line 189: Discussed elsewhere, removed, related to how we used to supplement the Experience with previously-saved preferences.

src/fides/api/models/privacy_notice.py
```diff
@@ -456,18 +448,6 @@ class PrivacyNoticeHistory(PrivacyNoticeBase, Base):
...
-        A system is relevant if their data use is an exact match or a child of the notice's data use.
-        """
-        relevant_systems: List[FidesKey] = []
-        for system in db.query(System):
-            if self.applies_to_system(system):
-                relevant_systems.append(system.fides_key)
-                continue
-        return relevant_systems
-
```
pattisdr, line 470: Obsolete functionality removed.  We no longer calculate "relevant systems" for the consent report upfront. This was removed for performance reasons to improve the speed of saving preferences, but in examining existing functionality, we believe it's not a necessary feature of the consent report.

src/fides/api/models/privacy_preference.py
pattisdr, line 1: This file contains Deprecated Preferences Saved and Notices Served models.  Related work will migrate the data in these tables, but leave the tables in place for now.  These shouldn't be used.

src/fides/api/models/privacy_preference.py
```diff
@@ -328,95 +227,28 @@ class ServedNoticeHistory(ConsentReportingMixin, Base):
...
-    """The DB ORM model for storing PrivacyPreferenceHistory, used for saving
+    The DB ORM model for storing PrivacyPreferenceHistory, used for saving
     every time consent preferences are saved for reporting purposes.
+
+    Soon to be deprecated in favor of PrivacyPreferenceHistoryv2
     """
 
+    @declared_attr
+    def __tablename__(self) -> str:
+        return "privacypreferencehistory"
```
pattisdr, line 250: Prefixed the old models with Deprecated* until I remove them entirely but tablenames have stayed the same.

src/fides/api/api/v1/endpoints/served_notice_endpoints.py
pattisdr, line 1: Endpoints have been moved to plus and rewritten to support new workflow.

src/fides/api/models/tcf_purpose_overrides.py
```diff
@@ -20,4 +20,4 @@ def __tablename__(self) -> str:
     is_included = Column(Boolean, server_default="t", default=True)
     required_legal_basis = Column(String)
 
-    UniqueConstraint("purpose")
+    __table_args__ = (UniqueConstraint("purpose"),)
```
pattisdr, line 23: This omission was flagged by my latest alembic migration. Adjusting here.

src/fides/api/schemas/privacy_notice.py
```diff
@@ -145,37 +153,6 @@ class PrivacyNoticeResponse(PrivacyNoticeWithId):
...
-    ]  # The current saved preference for the given user if it exists
-    outdated_preference: Optional[
-        UserConsentPreference
-    ]  # If no current preference, check if we have a preference saved for a previous version.
-    current_served: Optional[
-        bool
-    ]  # Do we have a record of the most recent version of this notice or TCF component being served to the user?
-    outdated_served: Optional[
-        bool
-    ]  # Have we served an older version of this notice or TCF component to the user?
```
pattisdr, line 167: Obsolete - Plus should now use PrivacyNoticeResponse when it adds Privacy Notices to the Experience, not PrivacyNoticeResponseWithUserPreferences.

src/fides/api/schemas/privacy_preference.py
```diff
@@ -1,334 +1,10 @@
...
-    request_origin: Optional[RequestOrigin]
-    url_recorded: Optional[SafeStr]
-    user_agent: Optional[SafeStr]
 
 
 class MinimalPrivacyPreferenceHistorySchema(FidesSchema):
     """Minimal privacy preference history schema for building consent emails"""
 
     preference: UserConsentPreference
     privacy_notice_history: PrivacyNoticeHistorySchema
```
pattisdr, line 10: Most of the Privacy Preference schemas have moved to Plus for convenience.  This stays here because this is for building consent emails.

src/fides/api/schemas/tcf.py
pattisdr, line 1: Most TCF schemas have moved to Plus for convenience so we have TCF in one repo.  Not all of these schemas are needed now that we no longer save preferences against TCF individually.

src/fides/api/service/connectors/consent_email_connector.py
```diff
@@ -230,12 +232,12 @@ def batch_email_send(self, privacy_requests: Query) -> None:
...
-            # Privacy preferences for new workflow
             filtered_privacy_preference_records: List[
-                PrivacyPreferenceHistory
-            ] = filter_privacy_preferences_for_propagation(
-                self.configuration.system, privacy_request.privacy_preferences
+                PrivacyPreferenceHistoryV2
+            ] = filter_privacy_preferences_for_propagation(  # type: ignore[assignment]
+                self.configuration.system, privacy_request.privacy_preferences_v2
             )
+
```
pattisdr, line 240: Type just changing related to how Privacy Requests have a relationship to the new Privacy Preference V2 table, that is the record that stores the consent details we need to propagate to third parties.

src/fides/api/util/consent_util.py
```diff
@@ -57,27 +57,32 @@
...
         for pref in privacy_preferences
         if pref.privacy_notice_history
         and pref.privacy_notice_history.enforcement_level
         == EnforcementLevel.system_wide
-        and pref.preference != UserConsentPreference.acknowledge
+        and (
+            pref.preference
+            and pref.preference
+            in [UserConsentPreference.opt_in, UserConsentPreference.opt_out]
+        )
```
pattisdr, line 79: Now that we have a new "tcf" preference we needed to update "filter_privacy_preferences_for_propagation" to look at just opt out/opt in, so the new TCF preference was excluded.

src/fides/api/util/tcf/tcf_experience_contents.py
```diff
@@ -1,450 +0,0 @@
...
-    """
-    legal_basis_override_subquery = get_legal_basis_override_subquery(db)
-
-    consent_legal_basis_filter: BinaryExpression = (
-        legal_basis_override_subquery.c.overridden_legal_basis_for_processing
-        == LegalBasisForProcessingEnum.CONSENT
-    )
-    legitimate_interest_legal_basis_filter: BinaryExpression = (
-        legal_basis_override_subquery.c.overridden_legal_basis_for_processing
-        == LegalBasisForProcessingEnum.LEGITIMATE_INTEREST
```
pattisdr, line 217: TCF base query now moved to Plus!  We need to double check that Purpose overrides functionality still works as expected.

src/fides/api/util/tcf/tcf_experience_contents.py
```diff
@@ -1,450 +0,0 @@
...
-) -> List[FidesKey]:
-    """Return a list of the system fides keys that match the given TCF attribute,
-    provided the system is relevant for TCF.
-
-    Used for consent reporting, to take a snapshot of the systems that are relevant at that point in time
-    """
-
-    # For TCF attributes, we need to first filter to systems/privacy declarations that have a relevant GVL data use
-    # as well as a legal basis of processing of consent or legitimate interests
-    (
```
pattisdr, line 313: This hard-to-maintain functionality that calculated relevant systems for TCF for consent reporting has been removed altogether and is not a part of the new consent reporting table.

tests/ops/api/v1/endpoints/test_privacy_preference_endpoints.py
pattisdr, line 1: Moved to plus and adapted!

tests/ops/api/v1/endpoints/test_served_notice_endpoints.py
pattisdr, line 1: Moved to plus and adapted to check for new behavior and response!

tests/ops/models/test_privacy_experience.py
pattisdr, line 1: Obsolete functionality tests removed here

tests/ops/models/test_privacy_notice.py
pattisdr, line 1: Obsolete tests removed.

tests/ops/models/test_privacy_preference.py
pattisdr, line 1: Almost all of the functionality tested here (core functionality related to how we save preferences) is obsolete.  Removed and replaced with "v2" tests in plus.

tests/ops/util/test_tcf_privacy_declarations.py
pattisdr, line 1: Tests moved to plus!

src/fides/api/models/privacy_preference_v2.py
pattisdr, line 1: Core changes are in this file, with new models:

`CurrentPrivacyPreferenceV2` - Upserts the user's last saved preferences
`PrivacyPreferenceHistoryV2` - Consent reporting. Stores every time preferences are saved
`LastServedNoticeV2` - Upserts latest notices served to the user.
`ServedNoticeHistoryV2` - Consent reporting for notices served. Also has a field to calculate conversion between what notices were served and what notices had preferences saved against them.

Similar to their v1 counterparts in src/fides/api/models/privacy_preference.py except we now upsert one "last saved" record and one "historical" record per notice. Multiple preferences/records served are saved under jsonb fields instead of being stored flat.
Followup data migration will migrate old data and application has been pointed to the new tables, but I'll temporarily leave the old tables in place. I recommend we keep something like this "V2" notation until we get rid of the old tables.

tests/ops/models/test_privacy_preference_v2.py
```diff
@@ -0,0 +1,131 @@
...
+                "fides_user_device": "051b219f-20e4-45df-82f7-5eb68a00889f",
+                "method": "accept",
+                "privacy_experience_config_history_id": None,
+                "privacy_experience_id": privacy_experience_france_overlay.id,
+                "preference": "tcf",
+                "request_origin": "tcf_overlay",
+                "user_agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_2) AppleWebKit/324.42 (KHTML, like Gecko) Chrome/108.0.0.0 Safari/425.24",
+                "user_geography": "fr_idg",
+                "url_recorded": "example.com/",
+                "served_notice_history_id": "served_notice_history_test_id",
```
adamsachs, line 70: i was getting a bit tripped up in the relationships between `served_notice_history_id`, `privacy_notice_history_id`, etc. i think i've worked it all out by reading through the code and your comments in the model are helpful. i just wonder if there's any way to reword some of the concepts to make it a bit clearer? i don't have any great suggestions off the top of my head, and i don't think it's worth making things more difficult to implement - your comments are already helpful. just wanted to note that it's a bit of a mental exercise, though that may be unavoidable!

adamsachs (approved):
@pattisdr this side of things looks quite good to me! still haven't gotten through the plus-side, which is obviously highly coupled/related, but nothing major here that looks problematic.

besides some smaller notes, the only major thing i was looking to understand better is the motivation for using the v2 tables and not just replacing more "directly". it's certainly not a blocker, and i'm sure there are good reasons for doing it this way, i just think there are some downsides so i'd like to better understand the motivation there!

src/fides/api/util/tcf/__init__.py
adamsachs, line 1: sorry just coming back to this after publishing my review. nothing major, just wanted to note:

1. maybe we could/should move this utility module over to plus too? it's only being used in plus, and i think it should stay only being used in plus generally, since it seems like we want as much of our TCF functionality to live there as possible...
2. weird how these missing imports got through our linter? i suppose it's because it's an `__init__.py`? anyway, thanks for cleaning up

pattisdr:
Thanks very much for your comments Adam!

> what's the motivation for leaving the old tables in place?

This was a no-brainer for me for the POC! It helped me comprehensively track down all the areas of the codebase that were affected (I would have definitely overlooked some things), and at the time, I wasn't positive of the migration requirements and believed all tables would be more difficult to migrate.  Now that I've gotten approval to only migrate a _subset_ of the preferences, you're right the "historical" tables could be migrated in-place easily, but the "current" tables remain nearly impossible. 

So I could go back and refactor to only create new tables for Current not Historical at this point, but I don't think that's where my focus should be with the remaining time.  I also strongly prefer the extra safety here of these before/after tables, the ease of testing the migrations, clearer diffs in these PR's, etc. I think it's a minor followup in a subsequent release to delete the old models and remove the old tables.

src/fides/api/util/tcf/__init__.py
pattisdr: yep I can add this to my list 👍

src/fides/api/schemas/tcf.py
```diff
@@ -1,235 +1,7 @@
...
-        if value not in GVL_SPECIAL_FEATURES:
-            raise ValueError(
-                f"Cannot save preferences against invalid special feature id: '{value}'"
-            )
-        return value
+from fideslang.gvl.models import MappedPurpose
+from pydantic import BaseModel
 
 
 class PurposesResponse(BaseModel):
```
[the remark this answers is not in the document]
pattisdr: sure I can move this 👍

src/fides/api/schemas/privacy_notice.py
pattisdr: Exactly yes

src/fides/api/service/connectors/consent_email_connector.py
pattisdr: sure, I thought the diff would be more confusing though since basically none of the logic in the original preferences model holds/original endpoints hold!

src/fides/api/models/privacy_preference_v2.py
pattisdr: https://github.com/ethyca/fides/pull/4481#issuecomment-1852825104

tests/ops/models/test_privacy_preference_v2.py
pattisdr: hmm not sure what to do here. 

In short, FE sends to the BE that consent was served and now I generate a single id and return it the FE.  FE should send that id to the backend on a subsequent save of their preferences. Maybe I could change the name?

src/fides/api/models/privacy_preference_v2.py
pattisdr: This was a no-brainer for me for the POC! It helped me comprehensively track down all the areas of the codebase that were affected (I would have definitely overlooked some things), and at the time, I wasn't positive of the migration requirements and believed all tables would be more difficult to migrate. Now that I've gotten approval to only migrate a subset of the preferences, you're right the "historical" tables could be migrated in-place easily, but the "current" tables remain nearly impossible.

So I could go back and refactor to only create new tables for Current not Historical at this point, but I don't think that's where my focus should be with the remaining time. I also strongly prefer the extra safety here of these before/after tables, the ease of testing the migrations, clearer diffs in these PR's, etc. I think it's a minor followup in a subsequent release to delete the old models and remove the old tables.

pattisdr: I guess cons are performance in the migration although it's fairly straightforward?

src/fides/api/util/tcf/__init__.py
pattisdr: OK moved over to Plus

src/fides/api/schemas/tcf.py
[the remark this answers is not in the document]
pattisdr: OK moved!

src/fides/api/models/privacy_preference_v2.py
pattisdr: @adamsachs I'll be looking into the lift of going back and updating the "historical" tables in place today (still migrating "current" preferences to new tables though, that migration is tricky!). Thanks for pushing here.

src/fides/api/service/connectors/consent_email_connector.py
adamsachs: that's a fair point! 🤔

pattisdr: but for this file in particular, I'm looking into your advice of leaving the historical models in place, which will remove the need for this update.  I think I will keep the "updated" historical models in new files though.

src/fides/api/models/privacy_preference_v2.py
adamsachs: @pattisdr ok! but only if you think there's significant benefit to doing so.

if the V2 approach makes this migration (both in the broad sense of the code update, and in the literal db migration(s)) easier and we have a relatively straightforward plan to "phase out" the v2 suffixes, then i'm _totally_ OK with that. facilitating implementation of a big change like this is a totally valid reason from my pov, and i didn't mean to question that - i just wanted to make sure it still feels like the right approach now that we've got a more holistic view on things. your comment about "not where your focus should be with the remaining time" is also totally valid, i.e. this isn't a decision made in a vacuum - so really, don't feel like there's a need to revisit any of this  if (1) it's providing benefit and saving some time/mental effort and (2) this is just an interim state that we've got a clear path for 'reverting' (and we ensure that work is planned!) 

migration performance doesn't feel like something we need to weigh heavily at all imo. i'm actually not sure if you're saying the v2 makes performance better or worse -- but in any case, i don't think we should let that factor influence the decision much at all!

pattisdr: Thanks @adamsachs - I originally didn't think it was worth the effort to go back and refactor to migrate the historical tables in place, but now that I'm in the middle of testing the migrations, the "current privacy preference" migration is quite tricky and time consuming, so I'd like to make the "historical migration" as quick as possible.  

Thanks again for commenting here, I do think it makes sense now to spend a little time on this now.  I think just removing the columns, removing preferences collected against TCF, and adding privacy notice names, etc. will be quicker than migrating over a lot of data into a new table with lots of FK's. Trying to avoid the FK checking.

pattisdr: Updated! Now updating PrivacyPreferenceHistory and ServedNoticeHistory tables in place instead of migrating to new tables.  The updated models are located in new files though to make the diff easier to see.

tests/ops/api/v1/endpoints/test_privacy_notice_endpoints.py
```diff
@@ -764,6 +765,7 @@ def test_get_privacy_notice_by_data_use_wrong_scope(
                             cookies=[
                                 CookieSchema(name="test_cookie", path="/", domain=None)
                             ],
+                            default_preference=UserConsentPreference.opt_out,
```
pattisdr, line 768: I have just unified the PrivacyNoticeResponseSchemas, I don't think there are any issues with this default preference being returned here

pattisdr:
Getting this up-to-date with fideslang!

PR #4503, opened 2023-12-08
general purpose override blocks purpose in all cases

adamsachs:
Closes PROD-1513 (again)

### Description Of Changes

Ensures that purposes are excluded no matter if they are declared with an inflexible legal basis.

### Code Changes

* [x] check first for exclusion in TCF experience generation
* [x]  check first for exclusion on the PrivacyDeclaration model method

### Steps to Confirm

* [x] did some integration testing locally with an alpha tag and the fidesplus release branch and made sure this works with e.g. `33across`: https://www.loom.com/share/af3541cb4e284eddb6d0c1a3eebcf303 

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### DIFF
--- a/src/fides/api/models/sql_models.py
+++ b/src/fides/api/models/sql_models.py
@@ -570,10 +570,7 @@ class PrivacyDeclaration(Base):
 
         Otherwise, we return the override!
         """
-        if not (
-            CONFIG.consent.override_vendor_purposes
-            and self.flexible_legal_basis_for_processing
-        ):
+        if not (CONFIG.consent.override_vendor_purposes):
             return self.legal_basis_for_processing
 
         query: Select = select(
@@ -600,7 +597,7 @@ class PrivacyDeclaration(Base):
 
         return (
             required_legal_basis
-            if required_legal_basis
+            if required_legal_basis and self.flexible_legal_basis_for_processing
             else self.legal_basis_for_processing
         )
 
--- a/src/fides/api/models/sql_models.py
+++ b/src/fides/api/models/sql_models.py
@@ -570,7 +570,7 @@ class PrivacyDeclaration(Base):
 
         Otherwise, we return the override!
         """
-        if not (CONFIG.consent.override_vendor_purposes):
+        if not CONFIG.consent.override_vendor_purposes:
             return self.legal_basis_for_processing
 
         query: Select = select(
--- a/src/fides/api/util/tcf/tcf_experience_contents.py
+++ b/src/fides/api/util/tcf/tcf_experience_contents.py
@@ -170,14 +170,14 @@ def get_legal_basis_override_subquery(db: Session) -> Alias:
             case(
                 [
                     (
+                        TCFPurposeOverride.is_included.is_(False),
+                        None,
+                    ),
+                    (
                         PrivacyDeclaration.flexible_legal_basis_for_processing.is_(
                             False
                         ),
                         PrivacyDeclaration.legal_basis_for_processing,
-                    ),
-                    (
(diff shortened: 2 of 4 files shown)


tests/ops/util/test_tcf_privacy_declarations.py
```diff
@@ -140,13 +161,12 @@ def test_privacy_declaration_purpose_overrides(
             if declaration.purpose
         }
         assert original_legal_basis == {
-            9: "Legitimate interests",
```
adamsachs, line 143: i didn't totally understand why this isn't included in the `original_legal_basis` even after looking closer at the `get_tcf_base_query_and_filters` code, but i was following the example for purpose 2, and this does seem to be the behavior.

i assume it's expected! would be nice for me to understand _why_ though. based on my static analysis of the code, i thought that we'd still get the original legal basis populated in the underlying query, but i must be missing something :)

adamsachs:
@pattisdr i think this is ready for review, previous test runs looked good and i ensured the tests i added passed locally, so i am hopeful this CI run should be green.

i'll work on getting some e2e testing locally running to be extra sure, but in the meantime, mind taking a look at the code to make sure it looks ok to you?

tests/ops/util/test_tcf_privacy_declarations.py
pattisdr: ah, the main data I get from `get_tcf_base_query_and_filters` are the declarations that should show up in the TCF experience.  `original_legal_basis` is showing the original legal basis of declarations that met the criteria to be included in the experience.

Perhaps renaming the test variable would increase clarity around what I was testing here? To see original declarations, you could assert the legal bases on the declarations in the emerse system fixture

src/fides/api/util/tcf/tcf_experience_contents.py
```diff
@@ -169,16 +169,16 @@ def get_legal_basis_override_subquery(db: Session) -> Alias:
...
                     (
                         PrivacyDeclaration.flexible_legal_basis_for_processing.is_(
                             False
                         ),
                         PrivacyDeclaration.legal_basis_for_processing,
                     ),
-                    (
-                        TCFPurposeOverride.is_included.is_(False),
-                        None,
-                    ),
```
pattisdr, line 181: 👍

src/fides/api/models/sql_models.py
```diff
@@ -600,7 +597,7 @@ async def get_purpose_legal_basis_override(self) -> Optional[str]:
 
         return (
             required_legal_basis
-            if required_legal_basis
+            if required_legal_basis and self.flexible_legal_basis_for_processing
```
pattisdr, line 600: thanks for getting this one too. In hindsight since the planned work that uses this method has been postponed, I probably should have removed this altogether -

pattisdr (approved):
great attention to detail in flagging this one!

tests/ops/util/test_tcf_privacy_declarations.py
pattisdr: I can just make a note in my TODO list for moving this work into the override to add clarity to this test, not worth pushing another commit here

src/fides/api/models/sql_models.py
adamsachs: all good! it's not bad to have this here so that subsequent work will have a good starting point.

tests/ops/util/test_tcf_privacy_declarations.py
adamsachs: sure that sounds good! i think i get it :)

Issue #3400, opened 2023-05-30
Import autogenerated typescript types from a shared package


allisonking:
### Is your feature request related to a specific problem?

Right now, whenever one of our frontend apps wants to utilize a backend type, we either:
1. For admin-ui: run `npm run openapi:generate` in the `admin-ui` folder which generates types into [`admin-ui/types/api`](https://github.com/ethyca/fides/tree/main/clients/admin-ui/src/types/api)
2. For privacy-center: copy the relevant types into the `privacy-center` folder in [`privacy-center/types/api/`](https://github.com/ethyca/fides/tree/main/clients/privacy-center/types/api)
3. For fides-js: copy the relevant types into the `fides-js` folder in one file [`consent-types.ts`](https://github.com/ethyca/fides/blob/main/clients/fides-js/src/lib/consent-types.ts)

This results in copies of types often being out of date.

### Describe the solution you'd like

We should keep our autogenerated typescript types (the output from `npm run openapi:generate`) in a shared separate package which all of these projects can import from. Now that we have turbo set up, this should be easier to iterate on.

### Describe alternatives you've considered, if any

_A description of any alternative solutions or features you've considered._

### Additional context

See here for more details: https://github.com/ethyca/fides/pull/3340#discussion_r1210455050

rsilvery:
@allisonking  is this still tech debt we should address?

allisonking:
Yup

PR #4504, opened 2023-12-08
Core GPP string serving for US geographies

allisonking:
Closes https://ethyca.atlassian.net/browse/PROD-1362

### Description Of Changes

Allows serving GPP strings for US regions!

https://www.loom.com/share/38960cec78a34cf6aa198cbc6632ffaa

### Code Changes

* [x] Added notice key map which captures how privacy notices map to GPP fields
* [x] Update calls to CMP API to update GPP fields when privacy notices are served, and when privacy notices are opted in/out of
* [x] Integrates with the new GPP settings backend update https://github.com/ethyca/fidesplus/pull/1248
* [x] Removes the env var for `GPP_ENABLED` now that we can get it from the backend
* [x] Unit tests
* [x] Cypress tests

### Steps to Confirm

* [ ] In the admin-ui, enable "data sales and sharing" privacy notice
* [ ] For convenience, you can also set the privacy experience for an overlay to "always enable" the banner
* [ ] Run the privacy center via `FIDES_PRIVACY_CENTER__IS_GPP_ENABLED=true turbo dev`
* [ ] Visit http://localhost:3000/fides-js-demo.html?geolocation=us-ca
* [ ] In your console, enter `__gpp('ping', (data) => {console.log({data})})`
* [ ] You should get a response back that is something like

```json
{
    "gppVersion": "1.1",
    "cmpStatus": "loaded",
    "cmpDisplayStatus": "visible",
    "signalStatus": "not ready",
    "supportedAPIs": [],
    "cmpId": 407,
    "sectionList": [
        8
    ],
    "applicableSections": [
        2
    ],
    "gppString": "DBABBg~BWAAAAAA.QA",
    "parsedSections": {
        "uscav1": {
            "Version": 1,
            "SaleOptOutNotice": 1,
            "SharingOptOutNotice": 1,
            "SensitiveDataLimitUseNotice": 2,
            "SaleOptOut": 0,
            "SharingOptOut": 0,
            "SensitiveDataProcessing": [
                0,
                0,
                0,
                0,
                0,
                0,
                0,
                0,
                0
            ],
            "KnownChildSensitiveDataConsents": [
                0,
                0
            ],
            "PersonalDataConsents": 0,
            "MspaCoveredTransaction": 0,
            "MspaOptOutOptionMode": 0,
            "MspaServiceProviderMode": 0,
            "GpcSegmentType": 1,
            "Gpc": false
        }
    }
}
```

The important things to note here is that the fields related to `data_sales_and_sharing` "Notice" are marked as `1`:
* "SaleOptOutNotice": 1, meaning notice was provided
* "SharingOptOutNotice": 1, meaning notice was provided
* And the other `*Notice` fields are marked as 2, meaning notice was not provided (since they don't fall under `data_sales_and_sharing` which is the only notice we have enabled

* [ ] Now opt in or out of all notices. If you opt out, if you issue the `ping` command again, you should get a result like

```json
{
    "gppVersion": "1.1",
    "cmpStatus": "loaded",
    "cmpDisplayStatus": "hidden",
    "signalStatus": "ready",
    "supportedAPIs": [],
    "cmpId": 407,
    "sectionList": [
        8
    ],
    "applicableSections": [
        8
    ],
    "gppString": "DBABBg~BWUAAAAA.QA",
    "parsedSections": {
        "uscav1": {
            "Version": 1,
            "SaleOptOutNotice": 1,
            "SharingOptOutNotice": 1,
            "SensitiveDataLimitUseNotice": 2,
            "SaleOptOut": 1,
            "SharingOptOut": 1,
            "SensitiveDataProcessing": [
                0,
                0,
                0,
                0,
                0,
                0,
                0,
                0,
                0
            ],
            "KnownChildSensitiveDataConsents": [
                0,
                0
            ],
            "PersonalDataConsents": 0,
            "MspaCoveredTransaction": 0,
            "MspaOptOutOptionMode": 0,
            "MspaServiceProviderMode": 0,
            "GpcSegmentType": 1,
            "Gpc": false
        }
    }
}
```
Note that `SaleOptOut` and `SharingOptOut` are now both `1`, and the gppString has changed slightly (can be decoded [here](https://iabgpp.com/#))

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`


clients/fides-js/src/lib/gpp/constants.ts
```diff
@@ -1 +1,356 @@
...
+type FidesRegionToGppFields = Record<string, GppFields>;
+type NoticeKeyToFidesRegionGppFields = Record<string, FidesRegionToGppFields>;
+
+/**
+ * Eventually this will be encoded on the privacy notice itself
+ *
+ * Currently based off of this table
+ * https://ethyca.atlassian.net/wiki/spaces/EN/pages/2852945989/2023-11-30+Fides.js+GPP+implementation?focusedCommentId=2861989902
+ */
+export const NOTICE_KEY_TO_FIDES_REGION_GPP_FIELDS: NoticeKeyToFidesRegionGppFields =
```
allisonking, line 35: @mfbrown would you be able to review this JSON? it's meant to be structured like an extension of a privacy notice. let me know if it makes sense, or if you'd like me to walk through it with you?

clients/fides-js/src/lib/gpp/us-notices.ts
```diff
@@ -0,0 +1,92 @@
+/**
```
allisonking, line 1: this file contains the bulk of the new logic!

clients/fides-js/src/lib/gpp/us-notices.ts
```diff
@@ -0,0 +1,92 @@
...
+  cmpApi,
+  experience,
+}: {
+  cmpApi: CmpApi;
+  experience: PrivacyExperience;
+}) => {
+  const sectionsChanged = new Set<{ name: string; id: number }>();
+  const { privacy_notices: notices, region } = experience;
+
+  Object.entries(NOTICE_KEY_TO_FIDES_REGION_GPP_FIELDS).forEach(
```
allisonking, line 29: we iterate through all possible notice types and set the GPP `*Notice` fields according to whether or not that notice appears in the `experience`

clients/fides-js/src/lib/gpp/us-notices.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  return Array.from(sectionsChanged);
+};
+
+/**
+ * Sets the appropriate fields on a GPP CMP API model for user opt-outs
+ *
+ * Returns GPP sections which were updated
+ */
+export const setGppOptOutsFromCookie = ({
```
allisonking, line 56: we iterate through all possible notice types and set the GPP "mechanism" fields (i.e. `SaleOptOut`, _not_ `SaleOptOutNotice`) according to whether or not that notice has been opted into/out of on the cookie

clients/fides-js/src/fides-ext-gpp.ts
```diff
@@ -41,39 +45,87 @@ declare global {
...
+  cmpApi.setFieldValueBySectionId(TcfEuV2.ID, "CmpId", ETHYCA_CMP_ID);
+  cmpApi.setSectionStringById(TcfEuV2.ID, tcString);
+  return true;
+};
+
+/** From our options, derive what APIs of GPP are applicable */
+const getSupportedApis = () => {
+  const supportedApis: string[] = [];
+  // QUESTION: or should this come from GPP config?
+  if (window.Fides.options.tcfEnabled) {
```
allisonking, line 71: the GPP config also has a `regions` param which can include the EU. do we need to consider that as well when we say that TCF is enabled here inside GPP?

clients/fides-js/src/fides-ext-gpp.ts
```diff
@@ -41,39 +45,87 @@ declare global {
...
     if (
       isPrivacyExperience(experience) &&
       !shouldResurfaceConsent(experience, event.detail)
     ) {
-      setTcString(event, cmpApi);
+      const tcSet = setTcString(event, cmpApi);
+      if (tcSet) {
+        cmpApi.setApplicableSections([TcfEuV2.ID]);
+      }
+      setGppNoticesProvidedFromExperience({ cmpApi, experience });
```
allisonking, line 106: this is the case where a user returns and already has a cookie, so we need to initialize the CMP with the right data. we set the opt outs, and I assume we should also set the `*Notices` but I'd like that assumption checked @mfbrown 🙏

clients/fides-js/src/fides-ext-gpp.ts
```diff
@@ -11,6 +11,7 @@ import {
   CmpDisplayStatus,
```
allisonking, line 11: This pattern of relying on Fides events to do GPP updates worked quite well, and I think we should consider porting the TCF code to do the same (right now in the TCF code, we manually call the TCF updates to generate the string, then the listeners just set the string. instead, I think it may make sense to let the listeners do all the string generation too, to keep it one file, and have the components not need to worry about what functions they are importing). this would help separate our bundles a bit more!

clients/fides-js/src/lib/gpp/constants.ts
```diff
@@ -1 +1,356 @@
...
+  {
+    data_sales_and_sharing: {
+      // QUESTION: should national be here AND in "data_sales" + "data_sharing"
```
allisonking, line 38: question for @mfbrown here!

clients/fides-js/src/lib/gpp/constants.ts
```diff
@@ -1 +1,356 @@
...
+        ],
+      },
+      us_ca: {
+        gpp_notice_fields: [UsCaV1Field.SENSITIVE_DATA_LIMIT_USE_NOTICE],
+        gpp_mechanism_fields: [
+          {
+            field: UsCaV1Field.SENSITIVE_DATA_PROCESSING,
+            not_available: Array(9).fill(0),
+            opt_out: Array(9).fill(1),
+            not_opt_out: Array(9).fill(2),
```
allisonking, line 256: this was what I ended up doing to handle [this question](https://ethyca.atlassian.net/wiki/spaces/EN/pages/2852945989/2023-11-30+Fides.js+GPP+implementation?focusedCommentId=2863038640) but this might not be the right assumption @mfbrown

clients/privacy-center/pages/api/fides-js.ts
```diff
@@ -150,8 +150,6 @@ export default async function handler(
       fidesDisableBanner: environment.settings.FIDES_DISABLE_BANNER,
       fidesTcfGdprApplies: environment.settings.FIDES_TCF_GDPR_APPLIES,
       fidesString,
-      // DEFER(PROD#1361): this should come from the backend
-      gppEnabled: environment.settings.IS_GPP_ENABLED,
```
eastandwestwind, line 154: 🚀

clients/fides-js/src/lib/extensions.ts
```diff
@@ -1,7 +1,17 @@
...
-export const setupExtensions = async (options: FidesOptions) => {
-  if (options.gppEnabled) {
+export const setupExtensions = async ({
+  options,
+  experience,
+}: {
+  options: FidesOptions;
+  experience: PrivacyExperience | EmptyExperience | undefined;
+}) => {
+  if (experience?.gpp_settings?.enabled) {
```
eastandwestwind, line 14: ✅

clients/fides-js/src/lib/gpp/us-notices.ts
```diff
@@ -0,0 +1,92 @@
...
+  const sectionsChanged = new Set<{ name: string; id: number }>();
+  const { privacy_notices: notices, region } = experience;
+
+  Object.entries(NOTICE_KEY_TO_FIDES_REGION_GPP_FIELDS).forEach(
+    ([noticeKey, regionGppFields]) => {
+      const gppFields = regionGppFields[region];
+      if (gppFields) {
+        const { gpp_notice_fields: fields } = gppFields;
+        const experienceHasNotice = !!notices?.find(
+          (n) => n.notice_key === noticeKey
```
eastandwestwind, line 35: I'm trying to get a sense for how GPP + TCF will interact. So if a user's experience / cookie has a consent val for a notice_key of "data_sales_and_sharing", then this consent val would apply for both GPP and TCF right?

eastandwestwind (commented):
Couple Qs for you @allisonking !

clients/fides-js/src/lib/gpp/constants.ts
```diff
@@ -1 +1,356 @@
...
+  string,
+  { name: string; id: number; prefix: string }
+> = {
+  us: { name: UsNatV1.NAME, id: UsNatV1.ID, prefix: "usnat" },
+  us_ca: { name: UsCaV1.NAME, id: UsCaV1.ID, prefix: "usca" },
+  us_ct: { name: UsCtV1.NAME, id: UsCtV1.ID, prefix: "usct" },
+  us_co: { name: UsCoV1.NAME, id: UsCoV1.ID, prefix: "usco" },
+  us_ut: { name: UsUtV1.NAME, id: UsUtV1.ID, prefix: "usut" },
+  us_va: { name: UsVaV1.NAME, id: UsVaV1.ID, prefix: "usva" },
+  // DEFER: Iowa isn't part of the GPP spec yet
```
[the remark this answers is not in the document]
allisonking: good question! so first we'd need to see if they're defined by the GPP spec yet. this file, in my mind, will eventually be privacy notice driven. so we'd update the privacy notice template with the mapping from privacy notice region (a fides concept) to gpp fields. in the interim, we'd update the JSON object instead

clients/fides-js/src/lib/gpp/stub.ts
```diff
@@ -52,10 +51,10 @@ export const makeStub = () => {
       cmpStatus: CmpStatus.STUB, // possible values: stub, loading, loaded, error
       cmpDisplayStatus: CmpDisplayStatus.HIDDEN, // possible values: hidden, visible, disabled
       signalStatus: SignalStatus.NOT_READY, // possible values: not ready, ready
-      supportedAPIs: SUPPORTED_APIS,
+      supportedAPIs: [],
       cmpId: 0, // IAB assigned CMP ID, may be 0 during stub/loading
       sectionList: [],
-      applicableSections: [],
+      applicableSections: [-1],
```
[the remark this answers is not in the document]
allisonking: ah this was something I saw on rereading the spec

> The value can be 0 or a Section ID specified by the Publisher / Advertiser, during stub / load. When no section is applicable, the value will be [-1].

actually re-reading it, maybe I should set it to 0 instead since this is in the stub...

clients/admin-ui/src/types/api/models/GPPSections.ts
```diff
@@ -0,0 +1,17 @@
+/* istanbul ignore file */
+/* tslint:disable */
+/* eslint-disable */
+
+/**
+ * An enumeration.
+ */
+export enum GPPSections {
```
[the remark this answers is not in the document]
allisonking: normally admin-ui contains all the types. we have a really old ticket to import all types from a shared lib, in which case admin-ui and fides-js could import from there https://github.com/ethyca/fides/issues/3400

clients/fides-js/src/lib/gpp/us-notices.ts
allisonking: not quite— `data_sales_and_sharing` is a privacy notice, and TCF doesn't use privacy notices (it uses all those `tcf_*` objects instead). so for TCF, we don't get any privacy notices back, AND the region isn't one of the regions defined in that `constants.ts` file, and so `if (gppFields)` wouldn't trigger (there are no gpp fields for TCF) so I guess the flow in this particular bit of logic is...

1. Map through all privacy notices
2. Grab the GPP fields that apply for our region
3. In the case of TCF, no GPP fields apply in i.e. France (this is defined by `NOTICE_KEY_TO_FIDES_REGION_GPP_FIELDS`)
4. so nothing happens in this function

Overall, we can think of GPP as another way to represent TCF _and_ a whole bunch of other policies. They "interact" if TCF is being used (i.e. in an EU region), and in that case, our GPP code subscribes to TCF updates, and sets the TC string that the TCF API gives us. For US regions (non-TCF), each state has a different policy which is encoded more or less in `NOTICE_KEY_TO_FIDES_REGION_GPP_FIELDS`. there's no region (right now...) that uses both TCF and another policy, so in theory they are separate

clients/fides-js/src/lib/gpp/stub.ts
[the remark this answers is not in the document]
eastandwestwind: Hm, yeah that sounds right. But also sounds like we'd possibly later need to set it to `-1` if we find that no sections are applicable. Can you easily make that update?

clients/fides-js/src/lib/gpp/constants.ts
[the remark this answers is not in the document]
eastandwestwind: Got it, thanks!

clients/fides-js/src/lib/gpp/us-notices.ts
eastandwestwind: oh thanks for the clarification! Could you point me to the bit of code where `our GPP code subscribes to TCF updates, and sets the TC string that the TCF API gives us`?

allisonking: yes! I suppose "GPP code subscribes to TCF updates" isn't quite right—it really subscribes to Fides updates, but that includes the TC string. 

https://github.com/ethyca/fides/blob/9470384dd0c37a6eaec8532a3030cd515ce48a44/clients/fides-js/src/fides-ext-gpp.ts#L143-L150

clients/fides-js/src/lib/gpp/stub.ts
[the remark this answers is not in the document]
allisonking: Yes, I'll update to `[0]`!

eastandwestwind (approved):
This is looking great. Just want to confirm that you have everything answered that's needed from @mfbrown ?

clients/fides-js/src/fides-ext-gpp.ts
mfbrown: Im not sure if it will help make this decision, but if a customer has purchased GPP and TCF, then TCF would be enabled as part of GPP.

clients/fides-js/src/lib/gpp/constants.ts
mfbrown: I think having national only on "data sales and sharing" is fine. The practical implementation for US National is basically the same as for California

clients/fides-js/src/fides-ext-gpp.ts
mfbrown: I think it's fair to say that if a user returns with an existing opt-out value, then we can report that they've also seen the notice in the GPP string.

allisonking:
> This is looking great. Just want to confirm that you have everything answered that's needed from @mfbrown ?

@mfbrown and I met today and we should be good here! some takeaways...

* known child consents + personal data consents have been separated out into their own notices
* known child consents + sensitive data are often bitfields which we decided to continue handling the way we have it now (either setting the whole array to 1 or 2) though this may change in the future

PR #4511, opened 2023-12-11
Backend Overhaul: Privacy Preferences Data Migration

pattisdr:
Closes #PROD-1511

❗ This data migration is needed for https://github.com/ethyca/fides/pull/4481

### Description Of Changes

As part of our backend overhaul with how we save privacy preferences we need to migrate existing preferences and notices served to the new format
### Code Changes

* [x] Add test script / generation script 
* [x] Migrate historical Privacy Preferences in-place
* [x] Migrate historical Notices Served in-place
* [x] Migrate current Privacy Preferences to new tables
* [x] Migrate all notices served to new tables

### Steps to Confirm

* [ ]  `nox -s teardown -- volumes`
* [ ] `nox -s dev -- shell`
* [ ] `python scripts/verify_privacy_preferences_overhaul_migration.py --migrate` to create fake artifacts in the db and then run the migration
    * [ ] Make sure you see:
```bash
Upgrading database to migration revision: head
2023-12-14 03:34:24.882 | INFO     | 5a8cee9c014c_privacy_preferences_v2_data_py:upgrade:41 - Migrating PrivacyPreferenceHistory.
2023-12-14 03:34:24.884 | INFO     | 5a8cee9c014c_privacy_preferences_v2_data_py:upgrade:48 - Migrating ServedNoticeHistory.
2023-12-14 03:34:24.885 | INFO     | 5a8cee9c014c_privacy_preferences_v2_data_py:upgrade:58 - Migrating CurrentPrivacyPreference to CurrentPrivacyPreferenceV2
2023-12-14 03:34:24.993 | INFO     | 5a8cee9c014c_privacy_preferences_v2_data_py:upgrade:67 - Migrating LastServedNotice to LastServedNoticeV2
```
* [ ] `python scripts/verify_privacy_preferences_overhaul_migration.py` to verify 

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`


scripts/verify_privacy_preferences_overhaul_migration.py
```diff
@@ -0,0 +1,782 @@
...
+    parser = argparse.ArgumentParser(description="Verify Privacy Preferences Migration")
+    parser.add_argument(
+        "--migrate",
+        dest="reload",
+        action="store_const",
+        const=True,
+        default=False,
+        help="whether or not to redo the migrations and reload the objects",
+    )
+
```
pattisdr, line 770: This is just a temporary test script illustrating the scenarios considered, will probably remove in the follow up work https://ethyca.atlassian.net/browse/PROD-1529

src/fides/api/alembic/migrations/versions/5a8cee9c014c_privacy_preferences_v2_data.py
```diff
@@ -0,0 +1,377 @@
...
+
+decryptor = sqlalchemy_utils.types.encrypted.encrypted_type.StringEncryptedType(
+    JSONTypeOverride,
+    CONFIG.security.app_encryption_key,
+    AesGcmEngine,
+    "pkcs5",
+)
+
+
+def migrate_current_records(
```
pattisdr, line 200: Migrating "current" privacy preferences and last notices served were challenging because the before/after have different sets of unique constraints.

src/fides/api/alembic/migrations/versions/5a8cee9c014c_privacy_preferences_v2_data.py
```diff
@@ -0,0 +1,377 @@
...
+    df["paths"] = df[
+        ["hashed_email", "hashed_phone_number", "hashed_fides_user_device"]
+    ].apply(lambda row: [val for val in row if pd.notna(val)], axis=1)
+
+    network_x_graph: nx.Graph = nx.Graph()
+    # Adds every path to the Graph
+    df["paths"].apply(lambda path: nx.add_path(network_x_graph, path))
+
+    # This is the magic - linking any common records across hashed_email OR hashed_phone OR hashed_device
+    connected_records: List[Set] = list(nx.connected_components(network_x_graph))
```
pattisdr, line 243: This was the best I could come up with (and unfortunately installed another requirement to do it) and ended up doing a lot of computations outside of SQL with pandas.

The main thing, I needed to assign a `group_id` on records that had the same email OR phone OR device id and then collapse preferences on that group_id.

- If Record A has phone 1234 and device id apple
- Record B has device id apple
- Record C has email customer-1@example.com and device id apple, 
- Record D has phone 1234

This treats all of the above as the same person

```python
G = nx.Graph() 
nx.add_path(G, ["1234", "apple"])
nx.add_path(G, ["apple"])
nx.add_path(G, ["apple", "customer-1@example.com"])
nx.add_path(G, ["1234"])
nx.add_path(G, ["123", "customer"])
nx.add_path(G, ["customer"])
nx.add_path(G, ["orange"])

connected_records = list(nx.connected_components(G))
connected_records
```
```bash
[{'1234', 'apple', 'customer-1@example.com'}, {'123', 'customer'}, {'orange'}]
```

pattisdr:
@adamsachs when you get a chance curious to get your thoughts on this migration! This is the last backend PR of the overhaul set.  The "historical" tables are straightforward to migrate but the "current" preferences were tricky -

src/fides/api/alembic/migrations/versions/5a8cee9c014c_privacy_preferences_v2_data.py
adamsachs: thank you for spelling this out here, this is a very clever solution! it's bringing me back to a past life where i spent my days working with graph databases.

i have the same qualms about a new dependency/requirement for "just" a migration, but it does seem like the default dependences for networkx are _relatively light_ (i'm looking [here](https://github.com/networkx/networkx/blob/main/pyproject.toml#L57-L62), i assume that's the right place? it may be good to just confirm these don't bring in a bunch of new transitive dependencies we don't already have...) and this is an elegant way to meet a complex requirement - so i think it's worth it!

i wonder if it'd make sense to just make a note on the dependency itself in `requirements.txt` to indicate that it was only added for a migration - maybe we'd consider removing the migration and the dependency at some point in the future once we're confident everyone's been properly migrated away from this state and if/when we're trying to clean up dependencies for whatever reason.  (i know updating migration history is always a bit dicey, but it feels like it could be acceptable in this case?).

requirements.txt
```diff
@@ -25,6 +25,7 @@ loguru==0.6.0
 multidimensional_urlencode==0.0.4
 okta==2.7.0
 openpyxl==3.0.9
+networkx==3.1
```
adamsachs, line 28: per comment below, i think even something as simple as this could help?

```suggestion
networkx==3.1 # added to help with privacy preference data migration
```

i'm a bit reluctant to add something like "not used in application runtime" just because at some point we _may_ end up using it in the app, and this then creates an extra maintenance step to remove that note... 🤷 probably thinking too much about this, but i think the takeaway is any small note here could help save some potential confusion and uncertainty down the line!

scripts/verify_privacy_preferences_overhaul_migration.py
adamsachs: 👏 ah how nice, you've already created this followup ticket! 🙏

adamsachs (approved):
this looks great @pattisdr! not only well organized and thoroughly documented/tested/scripted as always, but also really clever utilization of `networkx` and the connected components algorithm - it brings me back to a past life :) not sure what it was like implementing this, but you did find a way to make a tricky data migration fun for me to review - which i'd say is a pretty big accomplishment!

given the thorough testing you've already done (you went above and beyond with that verification script!), the only qualm is the added dependency, as you mention. i think maybe just a slight note to "flag" that dependency and mark it for potential removal in the future could be a nice step to mitigate dependency bloat, but open to hearing your thoughts on that too.

PR #4521, opened 2023-12-15
Fix issues with vendor selector behavior

jpople:
Closes #PROD-1450

### Description Of Changes

* After selecting a vendor, the vendor's name can now be edited as a text field
* On clicking or tabbing out of the vendor selector, typed input is not lost
* X button to clear input now only renders if the input is non-empty
* Changed error message when trying to enter a system name that's already taken to match designs
 
[Loom demo](https://www.loom.com/share/935db603c7da45b0b34107af00eba5cd)

AC of "This input is used in all places where a vendor can be selected from compass" has its own ticket [PROD-1355](https://ethyca.atlassian.net/jira/software/projects/PROD/boards/110?assignee=712020%3A21f193da-b0e9-44e4-a953-1cffba0fd533&selectedIssue=PROD-1355) and is not part of this PR.

### Code Changes

* Vendor selector input now stealthily switches between being a select which controls name/vendor ID simultaneously and a text input which controls only the name
* Removed no-longer-required "Separate vendor selector" feature flag
* Updated Yup validation schema for new error message

### Steps to Confirm

With Compass enabled on the manual "Add system" form:

* [ ] Enter a Compass vendor name and accept suggestions
* [ ] System name should be freely editable
* [ ] Clear input by clicking the X
* [ ] Suggestions should disappear and input should return to typeahead state

From a clean form:
* [ ] Enter a new system name (*not* the name of a Compass vendor), then click outside the input
* [ ] Input value should be retained

From a clean form:
* [ ] Enter a new system name, then press Tab
* [ ] Input value should be retained, "Description" field below should be focused

From a clean form:
* [ ] Enter a duplicate system name
* [ ] Error message should read: "You already have a system called $WHAT_YOU_ENTERED. Please specify a unique name for the system".

### Pre-Merge Checklist

* [ ] Update `CHANGELOG.md`

[PROD-1355]: https://ethyca.atlassian.net/browse/PROD-1355?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


clients/admin-ui/cypress/e2e/systems-plus.cy.ts
```diff
@@ -75,11 +77,11 @@ describe("System management with Plus features", () => {
           ).as("getDictSystem");
         });
       });
-      cy.intercept({ method: "PUT", url: "/api/v1/system*" }).as(
-        "putDictSystem"
-      );
+      // cy.intercept({ method: "POST", url: "/api/v1/system*" }).as(
```
allisonking, line 80: remove?

clients/admin-ui/cypress/e2e/systems-plus.cy.ts
```diff
@@ -105,11 +107,11 @@ describe("System management with Plus features", () => {
...
```
allisonking, line 110: remove?

clients/admin-ui/src/features/system/SystemInformationForm.tsx
```diff
@@ -129,12 +129,17 @@ const SystemInformationForm = ({
...
-          .notOneOf(
-            systems
+          .test("is-unique", "", (value, context) => {
+            const takenSystemNames = systems
               .filter((s) => s.name !== initialValues.name)
-              .map((s) => s.name),
-            "System must have a unique name"
-          ),
+              .map((s) => s.name);
+            if (takenSystemNames.some((name) => name === value)) {
```
allisonking, line 136: is there a test for this one? this is a neat validator!

clients/admin-ui/src/features/system/VendorSelector.tsx
```diff
@@ -130,165 +132,194 @@ const VendorSelector = ({
...
+            <span style={{ color: "#824EF2" }}>
+              {suggestions[0].label.substring(searchParam.length)}
+            </span>
+          ) : null}
+        </Text>
+      </VStack>
+    </FormControl>
+  );
+
+  const textInput = (
```
allisonking, line 286: it's a little weird for this to be a constant as opposed to a separate React component with props passed in. not a huge deal, but I think

```js
{isTypeahead ? <TypeaheadSelect isInvalid={isInvalid} ... /> : <TextInput isInvalid={isInvalid} ... />}
```

is a little clearer that we are expecting components. if it feels like we are repeating too many of the same props, can also do

```js
const props = { isInvalid: false }
{isTypeahead ? <TypeaheadSelect {...props} /> : <TextInput {...props} />}
```

allisonking: can this component use the `CustomInput` shared one from `inputs.tsx`? or is it too customized?

clients/admin-ui/src/features/system/VendorSelector.tsx
```diff
@@ -130,165 +132,194 @@ const VendorSelector = ({
...
+            backgroundColor: "transparent",
+            borderColor: "transparent",
+            marginTop: "31.52px",
+            marginLeft: "13px",
+            pointerEvents: "none",
+            fontSize: "14px",
+            zIndex: 1,
+          }}
+        >
+          <span style={{ color: "transparent" }}>{searchParam}</span>
```
allisonking, line 275: I know you just moved this around from existing code, but it is weird to me that we aren't using chakra components here (and for the `<Text/>` that we are using, right above, we are passing `style` as opposed to using the chakra provided classnames

clients/admin-ui/src/features/system/VendorSelector.tsx
```diff
@@ -130,165 +132,194 @@ const VendorSelector = ({
...
+      },
+      // only validate if nothing is typed in the select's search input;
+      // this prevents incorrectly showing a "required field" error after
+      // the input turns into a text input while the name is valid
+      !searchParam
+    );
+  };
+
+  const typeaheadSelect = (
+    <FormControl isInvalid={isInvalid} isRequired width="100%">
```
allisonking, line 192: what would it take for this to use the existing `CustomCreatableSelect` that exists already? it's fine to say this component is just too specialized for that! if it is, it'd be good to have a comment explaining why we aren't reusing the existing component

allisonking (commented):
still need to UAT, but leaving some comments first!

clients/admin-ui/src/features/system/VendorSelector.tsx
jpople: The only real difference with the `CustomInput` is that this one is wrapped in an `InputGroup` with the clear indicator as an `InputRightComponent`-- do you think it's reasonable to add handling for that case into the generic input?

allisonking: I think it could be! maybe if the default input is wrapped in an `InputGroup` (I don't _think_ that actually renders anything but I might be wrong about that), then it could take an optional `InputRightElement`? it's always a balance though—we don't want to add anything too bespoke to the common components, but I think having a right element could be useful for other inputs too so if it doesn't make the code too much more complicated, I think it's worth a shot

clients/admin-ui/src/features/system/VendorSelector.tsx
jpople: I think this one is too specialized, it needs specific handling for:

- when to show the menu
- when to show the clear indicator
- blurring
- tab completion
- the value of the inner input
- its own custom option

I'll add a comment about it.

allisonking:
this is looking great @jpople ! only thing I thought was kind of weird in my testing was that the "Create new system "{word}" option goes away once you start typing more letters? I guess maybe because we've switched to the other input type? I'd expect it to keep saying "Create a new system "{word}" as I keep typing

https://github.com/ethyca/fides/assets/24641006/8c85d534-326e-4e7b-9066-92da5349486f

I don't think it's a big deal, but that's the only thing I saw, everything else seems to meet the AC! 🎉

clients/admin-ui/src/features/system/VendorSelector.tsx
jpople: @allisonking After messing with putting the typeahead into its own component I'm sort of inclined to think that it's not worth it-- both components need access to things from the shared state and from `useFormikContext` and `useField`, so I'd either have to duplicate those hook calls or pass in a whole bunch (like, on the order of 10) of one-off helper methods to the typeahead as props, which I think makes it harder to read.

clients/admin-ui/src/features/system/VendorSelector.tsx
```diff
@@ -130,165 +134,185 @@ const VendorSelector = ({
...
+      handleClear();
+      return;
+    }
     setValue(newValue ? newValue.label : "");
-    setTouched({ ...touched, vendor_id: true, name: !fieldsSeparated });
+    setTouched(
+      { ...touched, vendor_id: true, name: true },
+      // do not validate if a new option was created; this prevents
+      // incorrectly showing a "required field" error while a value is in
+      // the field
```
jpople, line 170: The only scenario I can think of where not validating here (and the similar call below this) would cause problems would be when manually creating a system with a taken name, but validation for taken name seems to work anyway?  I tried also to use Formik's manual `validateField` methods for this but couldn't make them work, but this seems okay instead.

clients/admin-ui/src/features/system/VendorSelector.tsx
```diff
@@ -130,165 +134,185 @@ const VendorSelector = ({
...
+          fieldName="name"
         />
-      </HStack>
-      {fieldsSeparated ? (
-        <DictSuggestionTextInput
+        <Text
+          aria-hidden
+          position="absolute"
+          backgroundColor="transparent"
+          style={{ marginTop: "31.52px", marginLeft: "13px" }}
```
jpople, line 271: For some reason, these margins don't get applied when using the Chakra style props, so I did it this way instead.

clients/admin-ui/src/features/system/VendorSelector.tsx
allisonking: okay, thanks for looking into it!

allisonking (approved):
🎉

PR #4527, opened 2023-12-19
PROD-1485 - rely solely on cookie or fides string for current user preference

eastandwestwind:
Closes https://ethyca.atlassian.net/browse/PROD-1485

### Description Of Changes

Refactor frontend to rely only on cookie / fides string for current user preference. This needs to be tested and merged alongside https://github.com/ethyca/fides/pull/4481

### Code Changes

* [ ] Refactor FE such that we rely only on cookie / fides string for current user preference
* [ ] Adds new FE type `PrivacyNoticeWithPreference` that extends base `PrivacyNotice` with new fields we need to ensure consent is properly handled for GPC and consent resurfacing functionality

### Steps to Confirm

* [ ] Test full-stack with https://github.com/ethyca/fides/pull/4481.
* [ ] Specifically test GPC functionality and consent resurfacing. But overall, we need to ensure no new regressions

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`


clients/fides-js/src/lib/consent-types.ts
```diff
@@ -343,8 +343,14 @@ export type PrivacyNotice = {
   privacy_notice_history_id: string;
   cookies: Array<Cookies>;
   default_preference: UserConsentPreference;
+};
+
+// This type is exclusively used on front-end
+export type PrivacyNoticeExtended = PrivacyNotice & {
```
eastandwestwind, line 349: This is a key part of the changes. We extend base PrivacyNotice with additional helper props for FE

clients/fides-js/src/fides.ts
```diff
@@ -200,6 +200,7 @@ _Fides = {
     customOptionsPath: null,
     preventDismissal: false,
   },
+  shouldResurfaceConsent: false,
```
allisonking, line 203: hm why does this need to be on the Fides object?

clients/fides-js/src/lib/consent-value.ts
```diff
@@ -23,16 +23,16 @@ export const resolveLegacyConsentValue = (
...
-    return transformUserPreferenceToBoolean(notice.current_preference);
-  }
   if (notice.consent_mechanism === ConsentMechanism.NOTICE_ONLY) {
     return true;
   }
   const gpcEnabled =
-    !!notice.has_gpc_flag && context.globalPrivacyControl === true;
+    !!notice.has_gpc_flag &&
+    context.globalPrivacyControl === true &&
+    !previouslyConsented;
```
allisonking, line 35: why would previously consenting affect `gpcEnabled`?

clients/fides-js/src/lib/consent-value.ts
```diff
@@ -23,16 +23,16 @@ export const resolveLegacyConsentValue = (
 
 export const resolveConsentValue = (
   notice: PrivacyNotice,
-  context: ConsentContext
+  context: ConsentContext,
+  previouslyConsented?: boolean
 ): boolean => {
-  if (notice.current_preference) {
-    return transformUserPreferenceToBoolean(notice.current_preference);
-  }
```
allisonking, line 30: this function now needs the cookie to determine this part, instead of looking at `notice.current_preference` right? I think `previouslyConsented` muddies up the intended logic here a bit

clients/fides-js/src/lib/cookie.ts
```diff
@@ -266,7 +267,11 @@ export const buildCookieConsentForExperiences = (
     return cookieConsent;
   }
   experience.privacy_notices.forEach((notice) => {
-    cookieConsent[notice.notice_key] = resolveConsentValue(notice, context);
+    cookieConsent[notice.notice_key] = resolveConsentValue(
```
allisonking, line 270: it feels like we are moving in an odd circle here:

1. this function `buildCookieConsentForExperiences` (called in `initialize`) tries to `resolveConsentValue` based on the privacy notice and assign it to the cookie. however, it needs to know if the notice was `previously_consented`
2. `previous_consented` itself is determined by the cookie in `assignPreviousConsentToNotices` (also called in `initialize`
3.  so if I'm understanding this right, we are assigning `previously_consented` to `notices` based on the `cookie`, then here, we are using the `previously_consented` attr we just added to set the `cookie`? that seems circuitous?

clients/fides-js/src/lib/consent-types.ts
allisonking: I like the name you had in your PR description more: `PrivacyNoticeWithPreference`

clients/fides-js/src/fides.ts
allisonking: I'm not totally following the reasoning behind the `shouldResurfaceConsent` refactor?

allisonking (commented):
I think this may be overcomplicating things (understandably, since this code is already complicated!) but in theory, using only the cookie or fides string as our source of truth should simplify things.

I believe we may not need the `updateCookieAndExperience` function in `initialize` anymore, or if we do, it should be able to be simplified to only update the experience (not the cookie). from what I can tell, we should only be updating the cookie on preference saves now, and not on initialization. Even updating the experience on initialization may not be 100% necessary, (though maybe it's worth it for conveniently having the notice have the preference. I think this could also be accomplished without that though since the components have access to both the cookie and the experience objects, so we could have a separate function that figures out initial notice values based on the cookie)

another thing—I believe the privacy center might also use something like `resolveConsentValue`—do we need to do updates there too? is that a separate PR?

clients/fides-js/src/fides.ts
eastandwestwind: See https://github.com/ethyca/fides/pull/4527/files#diff-efe8c21098d15fb79a27d45475d1158e92c8ee59f04d2b2c6393bcf230f64019R351. We need a way to reliably determine if not all notices have previous consent (thus we know to resurface the banner) before we overwrite the cookie with any default vals. A good example would be for GPC.

To illustrate:

1. Cookie has consent vals "a" = true, "b" = false.
2. User starts a new session on the site, and when fides loads, the experience returns notices `a`, `b`, and a new notice, `c`, which is GPC enabled.
3. In the init of fides, we assign `previously_consented` to `true` for both `a` and `b`, but `c` is `false` since `c` is not found in the cookie. Now know we have to resurface the banner.
4. Further down in the init, we overwrite the cookie based on the experience. See-. When this happens, the cookie will now have "a"=true ,"b"=false, "c"=false (GPC applied). Now we can still use our `shouldResurfaceConsent` var to resurface consent for this user, even when GPC has been applied to the net new notice.

clients/fides-js/src/lib/consent-value.ts
eastandwestwind: `gpcEnabled` is probably not the best var name here as it's diff than `gpcEnabled` on the notice itself. Here it's more like "eligible to be overridden by gpc". If user has previously given consent for this notice, then we don't want to override their previous selection. 

Here's an example:

1. User navs to the site with GPC enabled in browser
2. Fides has a GPC-enabled notice, surfaces the banner with that notice disabled by default and has the label `applied` on the modal. 
3. User decides to consent to this notice, so that the label now says `overridden` (GPC is enabled but consent has been set to override the configured default)
4. Next time the user navs to the site, we ensure we respect the user's choice

clients/fides-js/src/fides.ts
allisonking: what if we didn't overwrite the cookie with the default vals? I can imagine a flow like:

1. Page loads, we have a cookie and an experience with notices. The cookie might not have all the keys of the notice
2. `shouldResurfaceConsent` does its thing as it did before
3. in the `*Overlay` components, when generating the initial values, if the cookie doesn't have a value, then we grab the notice's default preference
4. only once we save, do we write the "default" preference to the cookie

clients/fides-js/src/lib/consent-value.ts
eastandwestwind: yes I think you're right. good catch!

clients/fides-js/src/fides.ts
eastandwestwind: Moving from slack thread to GH for visibility:

Overall we agree that not overwriting the cookie would be simpler. Allison suggested:
* Override vals- we would still overwrite the cookie for this, and doesn't affect approach here
* GPC- shift the “setting” of the notice current preferences to render time as opposed to initializing the cookie with it
* net new consent keys from client-side experience- same approach as GPC, setting current preferences at render time


2 things I pointed out we need to consider:

1. We automatically save user prefs for GPC during init (before render time, see `automaticallyApplyGPCPreferences` fn). If we overwrite the cookie here we'd no longer be able to determine if we should resurface consent during render time. 

2. We also need to ensure this won't have consequences with actual consent compliance. When we dispatch our final FidesInitialized event at the end of init, this triggers consent to be pushed to our datalayer:
```
pushFidesVariableToGTM({
      type: "FidesInitialized",
      detail: {
        consent: window.Fides.consent,
        fides_meta: window.Fides.fides_meta,
        identity: window.Fides.identity,
        tcf_consent: window.Fides.tcf_consent,
      },
    });
```
And we actually use our cookie to set the window.Fides.consent obj, so we need to be sure this cookie obj reflects all the calculated / default consent vals we show on the FE too:
```
// return an object with the updated Fides values
  return {
    consent: cookie.consent,
    fides_meta: cookie.fides_meta,
    identity: cookie.identity,
    fides_string: cookie.fides_string,
    tcf_consent: cookie.tcf_consent,
    experience: effectiveExperience,
    shouldResurfaceConsent: resurfaceConsent,
    geolocation,
    options,
    initialized: true,
  };
```

clients/fides-js/__tests__/lib/gpp/us-notices.ts
```diff
@@ -6,11 +6,11 @@ import {
   setGppOptOutsFromCookie,
 } from "../../../src/lib/gpp/us-notices";
 import {
+  FidesCookie,
   PrivacyExperience,
   PrivacyNotice,
 } from "../../../src/lib/consent-types";
 import { makeStub } from "../../../src/lib/gpp/stub";
-import { FidesCookie } from "../../../src/lib/cookie";
```
allisonking, line 13: if you're having problems with circular dependencies on importing types, you can normally do 

```ts
import type { FidesCookie } from "../../../src/lib/cookie";
```
instead

clients/fides-js/src/components/notices/NoticeOverlay.tsx
```diff
@@ -24,17 +21,29 @@ import { useConsentServed } from "../../lib/hooks";
...
   cookie,
 }) => {
-  const initialEnabledNoticeKeys = useMemo(
-    () => Object.keys(cookie.consent).filter((key) => cookie.consent[key]),
-    [cookie.consent]
-  );
+  const consentContext = useMemo(() => getConsentContext(), []);
+  const initialEnabledNoticeKeys = useMemo(() => {
+    if (experience.privacy_notices) {
+      return experience.privacy_notices?.map((notice) => {
```
allisonking, line 37: no need for the `?` here since you have the `if` above
```suggestion
      return experience.privacy_notices.map((notice) => {
```

clients/fides-js/src/components/notices/NoticeOverlay.tsx
```diff
@@ -24,17 +21,29 @@ import { useConsentServed } from "../../lib/hooks";
...
+        const val = resolveConsentValue(notice, consentContext, cookie);
+        if (val) {
+          return notice.notice_key as PrivacyNotice["notice_key"];
+        }
+        return "";
```
allisonking, line 42: nit: ternary is more concise

```suggestion
        return val ? notice.notice_key as PrivacyNotice["notice_key"] : "";
```

allisonking: is this cast needed because we are going from `PrivacyNoticeWithPreference` to `PrivacyNotice`?

clients/fides-js/src/components/notices/NoticeOverlay.tsx
```diff
@@ -24,17 +21,29 @@ import { useConsentServed } from "../../lib/hooks";
...
   experience,
   options,
   fidesRegionString,
   cookie,
 }) => {
-  const initialEnabledNoticeKeys = useMemo(
-    () => Object.keys(cookie.consent).filter((key) => cookie.consent[key]),
-    [cookie.consent]
-  );
+  const consentContext = useMemo(() => getConsentContext(), []);
```
allisonking, line 34: what about putting the call to `consentContext` inside the `useMemo` for `initialEnabledNoticeKeys`? nowhere else uses this var right?

clients/fides-js/src/integrations/gtm.ts
```diff
@@ -1,5 +1,5 @@
-import { CookieKeyConsent } from "../lib/cookie";
 import { FidesEventDetail } from "../lib/events";
+import { CookieKeyConsent } from "~/lib/consent-types";
```
allisonking, line 2: we'll want a relative import here

clients/fides-js/src/lib/consent-utils.ts
```diff
@@ -242,16 +196,18 @@ export const hasSavedTcfPreferences = (experience: PrivacyExperience) =>
...
+    // Ensure we always resurface consent for TCF if for some reason experience does not have version_hash
+    return true;
   }
+  // If not every notice has previous user consent, we need to resurface consent
   return Boolean(
-    experience?.privacy_notices?.some(
-      (notice) => notice.current_preference == null
+    !experience.privacy_notices?.every((notice) =>
+      noticeHasConsentInCookie(notice, cookie)
     )
```
allisonking, line 211: I think we need to be a little careful here for these cases:
1. `experience.privacy_notices = undefined` returns true when it probably shouldn't (`!undefined = true`)
2. `experience.privacy_notices = []` returns true when it probably shouldn't (`every` on an empty array returns true)

clients/fides-js/src/lib/consent-value.ts
```diff
@@ -22,20 +30,17 @@ export const resolveLegacyConsentValue = (
...
     return true;
   }
-  const gpcEnabled =
-    !!notice.has_gpc_flag && context.globalPrivacyControl === true;
-  if (gpcEnabled) {
-    return false;
+  const preferenceExistsInCookie = noticeHasConsentInCookie(notice, cookie);
+  // Note about GPC - consent has already applied to the cookie at this point, so we can trust preference there
+  if (preferenceExistsInCookie) {
+    return <boolean>cookie.consent[notice.notice_key];
```
allisonking, line 43: I think this should work
```suggestion
    return !!cookie.consent[notice.notice_key];
```

clients/fides-js/src/lib/cookie.ts
```diff
@@ -289,18 +222,17 @@ export const updateExperienceFromCookieConsentNotices = ({
...
-    const defaultPreference = notice.current_preference ?? undefined;
-    const preference = Object.hasOwn(cookie.consent, notice.notice_key)
-      ? transformConsentToFidesUserPreference(
-          Boolean(cookie.consent[notice.notice_key]),
-          notice.consent_mechanism
-        )
-      : defaultPreference;
-    return { ...notice, current_preference: preference };
-  });
+  // todo- instead of updating experience here, push this logic into UI
```
allisonking, line 225: are you still planning on doing this? I think it'd be nice to remove `PrivacyNoticeWithPreference` entirely if possible

clients/fides-js/src/lib/events.ts
```diff
@@ -1,5 +1,5 @@
-import { FidesCookie } from "./cookie";
 import { debugLog } from "./consent-utils";
+import { FidesCookie } from "~/lib/consent-types";
```
allisonking, line 2: also needs relative import

clients/fides-js/src/lib/initialize.ts
```diff
@@ -328,15 +336,23 @@ export const initialize = async ({
...
+      // apply GPC preferences
+      if (shouldInitOverlay && isPrivacyExperience(effectiveExperience)) {
+        automaticallyApplyGPCPreferences({
+          cookie,
+          fidesRegionString,
+          effectiveExperience,
+          fidesOptions: options,
+        });
+      }
       const updated = await updateCookieAndExperience({
```
allisonking, line 356: we can probably rename the `updateCookieAndExperience` function now? it doesn't update the cookie anymore, right? (and arguably doesn't need to update the experience anymore either, but we can punt on that if that makes more sense)

clients/fides-js/src/lib/shared-consent-utils.ts
```diff
@@ -0,0 +1,50 @@
+import {
+  ConsentMechanism,
+  FidesCookie,
+  PrivacyNoticeWithPreference,
+  UserConsentPreference,
+} from "./consent-types";
+
+// PC cannot import consent-utils.ts due to Webpack build issue.
```
allisonking, line 8: weird!

clients/privacy-center/components/consent/NoticeDrivenConsent.tsx
```diff
@@ -39,15 +42,33 @@ import SaveCancel from "./SaveCancel";
...
     return UserConsentPreference.OPT_OUT;
   }
+  const preferenceExistsInCookie = noticeHasConsentInCookie(
+    notice as PrivacyNoticeWithPreference,
+    cookie
+  );
+  if (preferenceExistsInCookie) {
+    return transformConsentToFidesUserPreference(
+      // @ts-ignore
+      cookie.consent[notice.notice_key],
```
allisonking, line 67: is the ts problem that this could be undefined? if so, doing a `!!` would be safer than a `@ts-ignore`
```suggestion
      !!cookie.consent[notice.notice_key],
```

clients/privacy-center/components/consent/NoticeDrivenConsent.tsx
```diff
@@ -71,22 +91,25 @@ const NoticeDrivenConsent = () => {
...
-        const defaultValue = notice
-          ? resolveConsentValue(notice, consentContext)
-          : UserConsentPreference.OPT_OUT;
-        newPreferences[key] = defaultValue;
-      }
-    });
+  const initialDraftPreferences: NoticeHistoryIdToPreference = useMemo(() => {
+    const newPreferences: NoticeHistoryIdToPreference = {};
+    if (experience?.privacy_notices) {
+      experience.privacy_notices?.forEach((notice) => {
```
allisonking, line 97:
```suggestion
      experience.privacy_notices.forEach((notice) => {
```

clients/privacy-center/components/consent/NoticeDrivenConsent.tsx
```diff
@@ -109,13 +132,15 @@ const NoticeDrivenConsent = () => {
             (p) => p.privacy_notice_history_id
           ),
           serving_component: ServingComponent.PRIVACY_CENTER,
+          served_notice_history_id: servedNotice?.served_notice_history_id,
```
allisonking, line 135: is this an unrelated change?

allisonking (commented):
awesome job simplifying things! a few notes:

* doing an `import type {...}` might make a lot of the changes you had to make to avoid circ dependencies go away
* would prefer to not use `PrivacyNoticeWithPreference` at all if possible?
* do we need to update the experience inside the `updateCookieAndExperience` func now? it'd be awesome if we could remove it entirely, but not sure how feasible that is

clients/fides-js/src/lib/cookie.ts
eastandwestwind: Given the timing of the release, I think we defer this if possible to a later refactor. I'll make a ticket for it, with the tech debt discussion in mind from last retro!

clients/fides-js/src/components/notices/NoticeOverlay.tsx
eastandwestwind: yup exactly

clients/fides-js/__tests__/lib/gpp/us-notices.ts
eastandwestwind: ohhh lovely! Thanks so much!

clients/fides-js/src/components/notices/NoticeOverlay.tsx
eastandwestwind: yup you're totally right!

clients/fides-js/src/lib/initialize.ts
eastandwestwind: good point!

clients/privacy-center/components/consent/NoticeDrivenConsent.tsx
eastandwestwind: This is the ts issue:

Screenshot 2024-01-04 at 5 41 23 PM

clients/fides-js/src/lib/cookie.ts
eastandwestwind: this is also related to our ability to get rid of the `PrivacyNoticeWithPreference` type, as this is the only place where it's needed- when we assign it on the experience here. I'll make note of these things in the ticket

eastandwestwind: https://ethyca.atlassian.net/browse/PROD-1568

clients/privacy-center/cypress/e2e/consent-notices.cy.ts
```diff
@@ -313,7 +312,7 @@ describe("Privacy notice driven consent", () => {
         const { preferences } = body;
         expect(
           preferences.map((p: ConsentOptionCreate) => p.preference)
-        ).to.eql(["opt_in", "opt_in", "acknowledge"]);
+        ).to.eql(["opt_in", "opt_out", "acknowledge"]);
```
allisonking, line 315: changed the cookie to make "advertising" opt out so that we have a test case where we are setting a value to `false` as well 👍

clients/privacy-center/components/consent/NoticeDrivenConsent.tsx
```diff
@@ -132,15 +132,13 @@ const NoticeDrivenConsent = () => {
             (p) => p.privacy_notice_history_id
           ),
           serving_component: ServingComponent.PRIVACY_CENTER,
-          served_notice_history_id: servedNotice?.served_notice_history_id,
```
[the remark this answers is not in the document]
eastandwestwind: this is good to remove, thanks!

clients/privacy-center/cypress/e2e/consent-notices.cy.ts
eastandwestwind: thank you!

clients/privacy-center/components/consent/ConfigDrivenConsent.tsx
```diff
@@ -59,6 +64,13 @@ const ConfigDrivenConsent = ({
         consentOption: option,
         consentContext,
       });
+      option.cookieKeys?.forEach((cookieKey) => {
+        const previousConsent = cookieKeyConsent[cookieKey];
+        // For a cookie key to have consent, _all_ data uses that target that cookie key
+        // must have consent.
+        cookieKeyConsent[cookieKey] =
+          previousConsent === undefined ? value : previousConsent && value;
```
allisonking, line 72: hm how does this work if the `previousConsent` was `true` but now they are setting it to `false`? wouldn't the `previousConsent && value` make this still persist as `true`?

clients/privacy-center/cypress/e2e/consent-notices.cy.ts
```diff
@@ -69,9 +69,18 @@ describe("Privacy notice driven consent", () => {
...
+      cy.getCookie(CONSENT_COOKIE_NAME).should("not.exist");
       cy.visit("/consent");
-      cy.getByTestId("consent");
       cy.overrideSettings(SETTINGS);
+      cy.waitUntil(() =>
+        cy
+          .window()
+          // @ts-ignore window should have a store obj
+          .then((win) => win.store.getState().settings.IS_OVERLAY_ENABLED)
+      );
```
allisonking, line 81: we can remove this now

clients/privacy-center/cypress/e2e/consent-notices.cy.ts
```diff
@@ -208,6 +234,13 @@ describe("Privacy notice driven consent", () => {
         cy.getByTestId("save-btn").click();
 
         cy.wait("@patchPrivacyPreference").then(() => {
+          cy.getAllCookies().then((cookies) => {
+            cy.log(
+              `num cookies: ${cookies.length}. names: ${cookies
+                .map((c) => c.name)
+                .join()}`
+            );
+          });
```
allisonking, line 243: can also remove this logging

clients/privacy-center/components/consent/ConfigDrivenConsent.tsx
allisonking: are we not able to call `makeCookieKeyConsent` here? I imagine we are trying to move where we create the cookie from the `useEffect` in `consent.tsx` to here, so are we able to reuse that function?

allisonking: oh doh you're right, I don't know my boolean logic!!!

allisonking (approved):
WOOOOO :shipit:

PR #4539, opened 2023-12-21
Invite users via email

galvana:
Closes https://ethyca.atlassian.net/browse/PROD-1553

### Description Of Changes

Invite users by email! 

Initial design doc here: https://ethyca.atlassian.net/wiki/spaces/EN/pages/2872344650/Hackathon+2023+-+Team+1

#### Frontend

* [x] Adds email field to user creation/edit flow
* [x] Conditionally render setting up a different user's password—only go through this flow if email messaging is not enabled. Otherwise rely on emailing the user to have them set up their own password and activate their own account
* [x] User page and overall user table now have fields for email as well as a badge for if invite is pending
* [x] Refactor login page to use our more standard components (this lets us toggle to see the password!)
* [x] Render a login page with different copy if visiting the url with an invite code- the username field is disabled and prefilled
* [x] Call new accept invite endpoint when "logging in"
* [x] URLs now 'persist' through login, i.e. if you visited `/user-management` while not logged in, got redirected to the login page, then signed in, you'll automatically be redirected to `/user-management` (where in the past, you'd go to the home screen)
* [x] sneaky lil animation for logging in

#### Backend

* [x] Exposes whether email messaging is enabled via the health endpoint
* [x] Email is now a required field on creating new users
* [x] User object has two additional attributes: `disabled` and `disabled_reason`. A user pending an invite would have `disabled=true` and `disabled_reason="pending_invite"`
* [x] Flow for sending an email with an invite code when a user is created
* [x] New endpoint for accepting an invite
* [x] Updated the create user CLI command to require an email

### Steps to Confirm

* [ ] Add `FIDES__ADMIN_UI__URL="http://localhost:3000"` to your `.env` file
* [ ] Configure email messaging via the API, the Fides Postman collection makes this easy to do with these two API calls
  * [ ] Messaging Config - Email > Post Messaging Config
  * [ ] Messaging Config - Email > Messaging Config Secrets
* [ ] In the Admin UI, navigate to **Management > User Management** and click **Add New User**
  * [ ] Fill out the form with an email address you have access to. Verify the password field is disabled, that's a quick way to verify that "invite users via email" is properly configured
 * [ ] Check your email, open the invite link in an incognito window to avoid conflicting with your current Admin UI session
 * [ ] Provide a password and submit, you should be redirected to the Admin UI landing page

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`


src/fides/api/alembic/migrations/versions/13403fee598a_table_updates_for_user_invite.py
```diff
@@ -0,0 +1,71 @@
...
+        ["username"],
+        unique=False,
+    )
+    op.add_column("fidesuser", sa.Column("email_address", CIText(), nullable=True))
+    op.add_column("fidesuser", sa.Column("disabled", sa.Boolean(), nullable=False))
+    op.add_column(
+        "fidesuser",
+        sa.Column(
+            "disabled_reason",
+            sa.Enum("pending_invite", name="disabledreason"),
```
allisonking, line 56: would we prefer this to be a backend only enum?

clients/admin-ui/cypress.config.ts
```diff
@@ -4,6 +4,15 @@ import { defineConfig } from "cypress";
 export default defineConfig({
   e2e: {
     baseUrl: "http://localhost:3000",
+    setupNodeEvents(on) {
+      on("before:browser:launch", (browser, launchOptions) => {
+        if (browser.family === "chromium") {
+          // No need for tests to be slowed down by animations!
```
eastandwestwind, line 10: ⭐ - also does it matter what browser the cypress test uses? Can we just set this across all browsers for e2e testing?

clients/admin-ui/src/pages/login.tsx
```diff
@@ -1,44 +1,136 @@
...
+  );
+};
 
 const useLogin = () => {
-  const [loginRequest, { isLoading }] = useLoginMutation();
+  const [loginRequest] = useLoginMutation();
+  const [acceptInviteRequest] = useAcceptInviteMutation();
+  const [showAnimation, setShowAnimation] = useState(false);
+  // If the user prefers no motion, don't show the animation
+  const reduceMotion = usePrefersReducedMotion();
```
eastandwestwind, line 103: nice touch!

src/fides/api/api/v1/endpoints/user_endpoints.py
```diff
@@ -591,39 +609,64 @@ def user_login(
...
+        )
 
-    return client
+    return user_invite
+
+
+@router.post(
+    urls.USER_ACCEPT_INVITE,
+)
+def accept_user_invite(
```
eastandwestwind, line 647: do we have rate limiting on this endpoint? Even though this isn't a public endpoint, just want to ensure we're being extra careful for security here.

clients/admin-ui/src/features/auth/auth.slice.ts
```diff
@@ -69,12 +69,24 @@ const authApi = baseApi.injectEndpoints({
       query: () => ({ url: `oauth/role` }),
       providesTags: ["Roles"],
     }),
+    acceptInvite: build.mutation<
+      LoginResponse,
+      LoginRequest & { inviteCode: string }
```
eastandwestwind, line 74: how does `inviteCode` get passed in here?

src/fides/api/email_templates/templates/user_invite.html
```diff
@@ -0,0 +1,12 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <title>Welcome to Fides</title>
+  </head>
+  <body>
+    <main>
+      <p>You've been invited to join Fides, click <a href={{admin_ui_url}}/login?invite_code={{invite_code}}&username={{username}}>here</a> to accept the invite and setup your account.</p>
```
eastandwestwind, line 9: Could be useful to have org name here too, in addition to copy that says something about how this link will expire in 24 hours.

src/fides/api/schemas/user.py
```diff
@@ -104,7 +110,14 @@ class UserForcePasswordReset(FidesSchema):
 
 
 class UserUpdate(FidesSchema):
-    """Data required to update a FidesopsUser"""
+    """Data required to update a FidesUser"""
```
eastandwestwind, line 113: 😄

eastandwestwind (dismissed):
Mega feature! Couple small nits and Qs but otherwise looking good

src/fides/api/alembic/migrations/versions/13403fee598a_table_updates_for_user_invite.py
galvana: I tried doing that but I was having issues getting it to work, l'm ok with this for now

src/fides/api/api/v1/endpoints/health.py
```diff
@@ -54,6 +57,20 @@ def get_cache_health() -> str:
         return "unhealthy"
 
 
+def is_email_messaging_enabled(db: Session) -> bool:
```
adamsachs, line 60: per my comment below, may we want this function to actually look at the messaging service that's actively configured to be _used_, i.e. by looking at the `notifications.notification_service_type` config property? and we should be looking at the config proxy (i know that's a bit of a "gotcha"!). 

we've actually built an endpoint that does something very similar to this - check out `get_messaging_status` in `messaging_endpoints.py` - that's evaluating whether a messaging service is fully configured and "ready to go". i think we want a similar check here, but we want to also ensure that this actively configured messaging service is an email messaging service.

if we don't actually verify that the email service is fully configured to be used, then i think there's an edge case here where we set users up for failur e- unless i'm missing something!

src/fides/api/api/v1/endpoints/health.py
```diff
@@ -169,19 +186,23 @@ async def workers_health() -> Dict:
         },
     },
 )
-async def health() -> Dict:
+async def health(db: Session = Depends(get_db)) -> Dict:
```
adamsachs, line 189: this is a no-no! notice that we've explicitly split out the more _involved_ health checks into their own endpoints to keep the core health check as lightweight as possible. we shouldn't introduce a db dependency into this core health check - instead we should either tack this on to the database health check (which already has a db dependency) or make it its own health check endpoint (or some other endpoint).

we've had issues in the past where the core app health check was slow to respond, which can exacerbate deployment debugging and maybe even contribute to general performance issues in deployed environments. happy to talk thru that more if it's not clear (i'm sure @RobertKeyser has some battle wounds he can share!) but the important thing here is keeping the core health check endpoint as lightweight as possible 👍

src/fides/api/models/fides_user_invite.py
```diff
@@ -0,0 +1,88 @@
...
+            salt.encode(encoding),
+        )
+        return hashed_password, salt
+
+    @classmethod
+    def create(
+        cls, db: Session, *, data: dict[str, Any], check_name: bool = False
+    ) -> FidesUserInvite:
+        """
+        Create a FidesUserInvite by hashing the invite code with a generated salt
```
adamsachs, line 44: nice, i like that all the hashing logic is encapsulated in methods on the DB model class!

src/fides/api/models/fides_user_invite.py
```diff
@@ -0,0 +1,88 @@
...
+
+from citext import CIText
+from sqlalchemy import Column, String
+from sqlalchemy.ext.declarative import declared_attr
+from sqlalchemy.orm import Session
+
+from fides.api.cryptography.cryptographic_util import generate_salt, hash_with_salt
+from fides.api.db.base_class import Base
+
+INVITE_CODE_TTL_HOURS = 24
```
adamsachs, line 14: nit: could move this to a config var with a default

src/fides/api/models/fides_user_invite.py
```diff
@@ -0,0 +1,88 @@
...
+
+        current_time_utc = datetime.now(timezone.utc)
+
+        if not self.updated_at:
+            return True
+
+        expiration_datetime = self.updated_at + timedelta(hours=INVITE_CODE_TTL_HOURS)
+        return current_time_utc > expiration_datetime
+
+    def renew_invite(self, db: Session, new_invite_code: str) -> None:
```
adamsachs, line 82: is this used? seems like it could be useful, not saying we need to remove if it's not used but just curious if i'm missing the ref.

src/fides/api/api/v1/endpoints/user_endpoints.py
```diff
@@ -591,39 +609,64 @@ def user_login(
...
 
-def perform_login(
-    db: Session,
-    client_id_byte_length: int,
-    client_secret_btye_length: int,
-    user: FidesUser,
-) -> ClientDetail:
-    """Performs a login by updating the FidesUser instance and creating and returning
-    an associated ClientDetail.
+async def verify_invite_code(
```
adamsachs, line 612: nice to include this via dependency injection! is there a particular reason it's `async`?  just seems a bit odd especially since a non-async db `Session` is used and there aren't any actual async operations being `await`ed in the function.

adamsachs (changes requested):
this is a really awesome feature and it's impressive how quickly you all have put it into place! really excited to have this supported.

the BE looks pretty sound. there are a couple of points that i think need to be addressed before we can merge (pointed out via comments but i'll summarize here):
1. we need to refactor the health check endpoint. we shouldn't be adding a DB dependency to the core health check.
2. i think there are some edge cases around messaging config that i've tried to explain in the comments. the tl;dr is that i think we need to be a bit more consistent in checking the _actively configured_ messaging service and ensuring it's (1) fully configured and (2) configured to send emails. happy to talk thru this one if it's not clear, since it's a bit nuanced and some of the config here is a bit convoluted!

clients/admin-ui/cypress.config.ts
allisonking: I think it's more that the `--force-prefers-reduced-motion` option might be chromium specific. I believe we only use electron (which I think is chromium based) and chrome though, so that should cover our use cases!

clients/admin-ui/src/features/auth/auth.slice.ts
allisonking: ah over here! 

https://github.com/ethyca/fides/blob/2d2ceee61b3c00bb2959f92431a84e06e8faf69a/clients/admin-ui/src/pages/login.tsx#L126-L129

src/fides/api/models/fides_user_invite.py
galvana: Going to cut scope here and just leave it hard-coded for now

src/fides/api/api/v1/endpoints/health.py
galvana: Moved the email check out of the health check

src/fides/config/config_proxy.py
```diff
@@ -163,6 +170,7 @@ class ConfigProxy:
     """
 
     def __init__(self, db: Session) -> None:
+        self.admin_ui = AdminUISettingsProxy(db)
```
adamsachs, line 173: if we're putting these in the config proxy, i'm assuming we'd also want to be able to _read_ the values of these settings through the `GET /config` endpoint? if so, they'll need to be added to the `CONFIG_KEY_ALLOWLIST` in `src/fides/config/utils.py`, and there's also some tests that'll need to be updated.

src/fides/api/service/user/fides_user_service.py
```diff
@@ -0,0 +1,118 @@
...
+)
+from fides.api.schemas.redis_cache import Identity
+from fides.api.service.messaging.message_dispatch_service import dispatch_message
+from fides.config import FidesConfig
+from fides.config.config_proxy import ConfigProxy
+
+
+def invite_user(db: Session, config_proxy: ConfigProxy, user: FidesUser) -> None:
+    """
+    Generates a user invite and sends the invite code to the user via email.
```
adamsachs, line 25: nit: i'd just make a note here in the docstring that this method is a no-op if things aren't set up for user emails - it's a major piece of the functional behavior that i'd want to know without having to parse the impl

src/fides/api/alembic/migrations/versions/31493e48c1d8_table_updates_for_user_invite.py
```diff
@@ -0,0 +1,70 @@
...
+        sa.ForeignKeyConstraint(
+            ["username"], ["fidesuser.username"], ondelete="CASCADE"
+        ),
+        sa.PrimaryKeyConstraint("id", "username"),
+    )
+    op.create_index(
+        op.f("ix_fides_user_invite_id"), "fides_user_invite", ["id"], unique=False
+    )
+    op.add_column("fidesuser", sa.Column("email_address", CIText(), nullable=True))
+    op.add_column("fidesuser", sa.Column("disabled", sa.Boolean(), nullable=False))
```
adamsachs, line 49: doesn't this need a default value and/or data migration to ensure that existing records have the value populated?

src/fides/api/alembic/migrations/versions/31493e48c1d8_table_updates_for_user_invite.py
```diff
@@ -0,0 +1,70 @@
...
+    op.add_column(
+        "fidesuser",
+        sa.Column(
+            "disabled_reason",
+            sa.Enum("pending_invite", name="disabledreason"),
+            nullable=True,
+        ),
+    )
+    op.create_unique_constraint(
+        "fidesuser_email_address", "fidesuser", ["email_address"]
```
adamsachs, line 59: does this play well with null values? i don't _think_ nulls are treated as distinct, are they? this is important, especially since this is a new column and existing records won't have entries

in general, i'd be sure to test this migration pretty well on your local db populated with `fidesuser`s, feels like a few  potential issues lurking here, unless i'm missing something (haven't tested myself!)

adamsachs (commented):
@galvana thanks for reviving this effort and addressing the main issues i'd flagged in the initial review! generally things look good with those. 

i haven't tested this functionality myself manually, but i do want to make sure that we've done some recent manual testing on this, especially since it's been sitting dormant for a while, which makes it a bit more likely that some bugs would come up that slip underneath automated test coverage. have you been able to do some solid manual testing on the functionality?

the most important things i've flagged on this round are with the migration, i think. unless i'm missing something, i think those _need_ to be resolved before we merge - and they generally tell me that we haven't tested the migration on a realistic DB state too much - so can we make sure that's well tested?

my other comments are relatively minor and should either be quick updates or are not must-haves.

let me know if you've got any questions or if i can lend a hand at all on the migration work!

src/fides/api/service/user/fides_user_service.py
galvana: Good callout!

src/fides/api/alembic/migrations/versions/31493e48c1d8_table_updates_for_user_invite.py
galvana: Yes! I updated this to set `disabled` to False by default

src/fides/api/alembic/migrations/versions/31493e48c1d8_table_updates_for_user_invite.py
galvana: That's correct, the unique constraint only applies to non-null values, so the existing rows should be fine with null values for `email_address`

galvana: Ok, here's the local testing I did

1. Checked out `main` and created two users in the database
2. Checked out this branch and ran the migration
3. Migration was successful, neither of the existing users have an `email_address` and both have the default value of False for the `disabled` column

adamsachs: ok great! thanks for adjusting and testing that 👍

src/fides/config/config_proxy.py
adamsachs: nice, thanks for adding 👍

adamsachs (approved):
looking good to me @galvana , thanks for this latest round of adjustments. let's ship it! 🚢

allisonking:
party

Issue #1169, opened 2022-10-01
API routes should use `response_model_exclude_none` instead of sneaking `null` values into JSON


ssangervasi:
### Background
API routes use Pydantic models with `Optional[]` fields. In Python, this is represented by `None` values when the field is not specified (and doesn't have a default). For these models, "optional" and "nullable" are basically the same thing.

However, for the purposes of OpenAPI schemas and JSON response bodies, "optional" means that the key may not be present in the output at all. A key that may be present with value of `null` must indicate that `null` one of of its possible types.

### Issue

Route model fields like `DataCategory.parent_key` show up as optional (non-required) strings, meaning that they will not be included if they are unset (null/None):

<img width="492" alt="Screen Shot 2022-09-30 at 16 59 19" src="https://user-images.githubusercontent.com/2236777/193370633-32622a86-fe4a-444a-a3d9-b902937371fc.png">
(`*` by a field indicates it is required)

However, this route actually returns JSON objects which include `parent_key: null`:
```ts
  {
    description: "Data related to a system account.",
    fides_key: "account",
    name: "Account Data",
    organization_fides_key: "default_organization",
    parent_key: null,
    is_default: true,
  }
```

### Original comment:

Still digging into the OpenApi schema, but the answer may be that we should use [response_model_exclude_none](https://fastapi.tiangolo.com/tutorial/response-model/?h=response_model_exclude_none#use-the-response_model_exclude_unset-parameter) on our routes to exclude optionals instead of passing nulls. Applying that to all crud endpoints would be very nice for consistency.

As-is, I think the TS generator is _technically_ correct because if you look directly at the OpenAPI schema it says the value is optional (not required) and does _not_ say the type can be `null`:
<img width="492" alt="Screen Shot 2022-09-30 at 16 59 19" src="https://user-images.githubusercontent.com/2236777/193370633-32622a86-fe4a-444a-a3d9-b902937371fc.png">
(`*` by a field indicates it is required)

I don't know of a way to get fastapi/pydantic to say the type can be `null` -- [this](https://fastapi.tiangolo.com/it/tutorial/query-params-str-validations/#required-with-none) didn't do it when I tried -- and I would rather exclude them anyway.

_Originally posted by @ssangervasi in https://github.com/ethyca/fides/pull/1110#discussion_r985007442_

PR #4542, opened 2024-01-02
Update fides.js to support multiple descriptions (banner, overlay) and render HTML descriptions

NevilleS:
‼️ contains migration, check downrev before merging! ‼️ 

Closes PROD-1528
Closes PROD-1338
...and adds support for HTML descriptions (unticketed)

### Description Of Changes

As more & more customers are using our consent overlay, we're needing some more flexible options for configuring the major text elements of the banner & modal. This PR adds two major features:
1. Support for separate titles/descriptions for the banner & modal, so you can configure these two UIs separately
2. Support for HTML descriptions in the banner & modal, so you can include inline links which is a common need for legal text (to link out to more detailed policies)

### Code Changes

* [X] Support separate descriptions in Privacy Experience Config API
* [X] Support separate descriptions in fidesjs consent banner & modal
* [X] Support separate titles in Privacy Experience Config API
* [X] Support separate titles in fidesjs consent banner & modal
* [X] Support HTML descriptions in fidesjs consent banner & modal
* [X] Add a new fidesjs option `allowHTMLDescription` to optionally allow HTML
* [x] Add a new privacy center ENV variable to set the `allowHTMLDescription` option
* [x] Add `nh3` for HTML sanitization checks
* [x] Update Admin UI forms to support banner title & description
* [x] Fix Admin UI privacy experience form to not require privacy policy label & URL

### Steps to Confirm

* [ ] Configure banner title & description fields and test their changes in the Cookie House
* [ ] Set `FIDES_PRIVACY_CENTER__ALLOW_HTML_DESCRIPTION=true`, then configure banner & modal description fields with HTML and test their changes in the Cookie House

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`


clients/fides-js/src/components/ConsentBanner.tsx
```diff
@@ -66,7 +66,7 @@ const ConsentBanner: FunctionComponent<BannerProps> = ({
       }
     };
 
-    if (bannerIsOpen && !window.Fides.options.preventDismissal) {
+    if (bannerIsOpen && !window.Fides?.options?.preventDismissal) {
```
NevilleS, line 69: Just being more defensive here.

clients/fides-js/src/components/ExperienceDescription.tsx
```diff
@@ -5,14 +5,29 @@ const TEXT_TO_LINK = "vendors page.";
...
     return null;
   }
 
+  // If allowHTMLDescription is true, render rich HTML content
+  // NOTE: We sanitize these descriptions server-side when configuring the
+  // PrivacyExperience, so it's safe to trust these
+  if (allowHTMLDescription) {
+    return (
+      <div
+        className="fides-html-description"
```
NevilleS, line 24: This class might be more useful in the future; right now I use it to set `white-space: normal` since we're rendering HTML and therefore no longer want to treat whitespace (specifically newline characters!) as semantic in the same way we do for plaintext.

That CSS is in `fides.css` though, you'll see it later 👍

clients/fides-js/src/components/ExperienceDescription.tsx
```diff
@@ -5,14 +5,29 @@ const TEXT_TO_LINK = "vendors page.";
...
+        // eslint-disable-next-line react/no-danger
+        dangerouslySetInnerHTML={{ __html: description }}
```
NevilleS, line 26: You have to be really careful about this. I'm pretty sure I've done this securely, and have even ensured it's disabled by default to be extra cautious, so we need to specific `allowHTMLDescription: true`

clients/privacy-center/cypress/e2e/consent-banner.cy.ts
```diff
@@ -494,81 +636,71 @@ describe("Consent banner", () => {
...
-        // TODO: add tests for top/bottom
-        expect(false).is.eql(true);
-      });
-
-      it.skip("should support styling with CSS variables", () => {
-        // TODO: add tests for CSS
-        expect(false).is.eql(true);
-      });
-
-      describe("cookie enforcement", () => {
```
NevilleS, line 508: Hoisted these tests out into their own top-level `describe` case, that's all 👍

clients/privacy-center/cypress/e2e/consent-banner.cy.ts
```diff
@@ -1059,7 +1191,6 @@ describe("Consent banner", () => {
       });
     });
 
-    // TODO: it should be possible in the future to filter for experience on just country
```
NevilleS, line 1062: Old TODO

clients/privacy-center/cypress/e2e/consent-banner.cy.ts
```diff
@@ -494,81 +636,71 @@ describe("Consent banner", () => {
...
 
-      it.skip("should support option to display at top or bottom of page", () => {
-        // TODO: add tests for top/bottom
-        expect(false).is.eql(true);
-      });
-
-      it.skip("should support styling with CSS variables", () => {
-        // TODO: add tests for CSS
-        expect(false).is.eql(true);
-      });
```
NevilleS, line 506: Deleted these `skip` tests, no need to clutter up

src/fides/api/alembic/migrations/versions/f396c1f84b0f_add_banner_description_to_experience_.py
```diff
@@ -0,0 +1,41 @@
...
+Revises: 4ced99dabebb
+Create Date: 2024-01-02 01:35:41.813279
+
+"""
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "f396c1f84b0f"
+down_revision = "4ced99dabebb"
```
NevilleS, line 13: I think this will conflict with other open PRs so this is dangerous

src/fides/api/custom_types.py
```diff
@@ -28,6 +28,29 @@ def validate(cls, value: str) -> str:
...
+    """
+
+    @classmethod
+    def __get_validators__(cls) -> Generator:  # pragma: no cover
+        yield cls.validate
+
+    @classmethod
+    def validate(cls, value: str) -> str:
+        """Assert text doesn't include any complex/malicious HTML."""
+        # TODO: add a real sanitizer here!
```
NevilleS, line 47: Planning to use `nh3` here to run validations.

NevilleS (commented):
Just some comments to help reviewers...

adamsachs (commented):
BE looks solid to me! i think you've got all the fundamental pieces you need, i know there's at least one TODO so i'll hold off on approving. i like the `HtmlStr` custom type and validator.

i'm still not the most familiar with the use cases, but just want to make sure - should we populate these fields at all in the default "out of the box" experience configs  we define in `src/fides/data/privacy_notices/privacy_experience_config_defaults.yml`?

allisonking (approved):
FE looks great! thanks for working on this!

NevilleS:
@adamsachs 

> i'm still not the most familiar with the use cases, but just want to make sure - should we populate these fields at all in the default "out of the box" experience configs we define in `src/fides/data/privacy_notices/privacy_experience_config_defaults.yml`?

No, I don't think so. In general, the defaults that set the banner & modal descriptions/titles to be identical are reasonable, so I felt like we should leave our defaults unchanged and have these new fields as `null`.

pattisdr:
> i'm still not the most familiar with the use cases, but just want to make sure - should we populate these fields at all in the default "out of the box" experience configs we define in src/fides/data/privacy_notices/privacy_experience_config_defaults.yml?

In the UI, do we drop back to using the existing title/description if banner_title/banner_description don't exist? If not, then yes, we'll need to populate this template.

Further, current behavior of `load_default_experience_configs_on_startup`  only _adds_ new configs from the template if they don't exist, it doesn't update existing configs in the database.  We can talk about how we want to handle these new fields here if needed.

src/fides/api/alembic/migrations/versions/f396c1f84b0f_add_banner_description_to_experience_.py
pattisdr: yep always good to flag thanks!

pattisdr (commented):
looks good so far, backend is fairly straightforward since these new fields are allowed to be nullable - will look again when HtmlStr validation is in -

clients/admin-ui/src/types/api/models/ExperienceConfigCreate.ts
```diff
@@ -33,11 +35,11 @@ export type ExperienceConfigCreate = {
   /**
    * Overlay and Privacy Center 'Privacy policy link label'
    */
-  privacy_policy_link_label?: string;
+  privacy_policy_link_label?: string | null;
```
NevilleS, line 38: NOTE @allisonking: I explicitly added `null` here since there's quite a different in a `PATCH` API between `undefined` and `null`, and I wanted to provide _null_ specifically.

I feel like this lil' tweak will get removed when we regenerate these types automatically...?

clients/admin-ui/src/features/privacy-experience/form/OverlayForm.tsx
```diff
@@ -18,30 +17,26 @@ const BANNER_ENABLED_OPTIONS = enumToOptions(BannerEnabled).map((opt) => ({
   label: BANNER_ENABLED_MAP.get(opt.value) ?? opt.label,
 }));
 
-const INFO_TEXT =
```
NevilleS, line 21: I felt like this warning was confusing. I'll ask for a design review.

clients/admin-ui/src/features/privacy-experience/form/helpers.tsx
```diff
@@ -51,22 +52,66 @@ const privacyCenterValidationSchema = Yup.object()
...
+      .optional()
+      .trim()
+      .url()
+      .label("Privacy policy URL"),
   })
   .concat(buttonGroupValidationSchema);
 
+// Coerce empty strings to "null" values for the update API
+// NOTE: this is identical to the transform for the create API below, but copied
+// into a separate function for Typescript type-safety
```
NevilleS, line 91: nb: this little business-logic helper that coerces `""` to `null` is duplicated for both mutations. I didn't feel up to figuring out the Typescript incantation to support both type signatures in a single function.

clients/admin-ui/src/features/privacy-experience/form/helpers.tsx
```diff
@@ -51,22 +52,66 @@ const privacyCenterValidationSchema = Yup.object()
...
+export const transformFormValuesToExperienceConfigUpdate = (
+  values: ExperienceConfigUpdateParams
+): ExperienceConfigUpdateParams => ({
+  ...values,
+  banner_title: values.banner_title || null,
+  banner_description: values.banner_description || null,
+  privacy_policy_link_label: values.privacy_policy_link_label || null,
```
NevilleS, line 98: This fixes PROD-1338 while we're in here. In Javascript, an empty string `""` is false-y, so it'll be replaced with `null`

NevilleS (commented):
A few more comments for @allisonking

clients/admin-ui/src/features/privacy-experience/form/helpers.tsx
allisonking: I think it'd be something like

```typescript
export const transformFormValuesToExperienceConfig = <T extends {banner_title: string ...}>(values: T) : T => ({
  ...
});

// or
// <T extends Pick<ExperienceConfigUpdateParams, 'banner_title' | 'banner_description'...>>
```

clients/admin-ui/src/types/api/models/ExperienceConfigCreate.ts
allisonking: hmm yes it'd get removed 🤔 so does a `privacy_policy_link_label = null` mean it's cleared, while if `privacy_policy_link_label` isn't provided, it's not touched? if so, I think we need to specify on the BE that this can be null in order for the FE types to be okay

clients/admin-ui/src/features/privacy-experience/form/OverlayForm.tsx
```diff
@@ -70,6 +65,16 @@ const OverlayForm = () => (
         variant="stacked"
         isRequired
       />
+      <CustomTextInput
+        label="Banner title (optional, if different from overlay)"
```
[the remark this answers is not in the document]
NevilleS: Good suggestion 👍

clients/admin-ui/src/types/api/models/ExperienceConfigCreate.ts
NevilleS: Yeah, since the `PATCH` semantics are "don't touch values that aren't included at all", you can do things like this:
```
{
  title: "Just update title"
}
```

...and leave everything else out. That's a subtle mismatch between the type system. 

For example, in Python the field is `privacy_policy_link_label: Optional[str]`, which seems to be converted to `privacy_policy_link_label?: string` in Typescript. Those _look_ the same and function like that in practice but there's a difference!

src/fides/api/custom_types.py
```diff
@@ -28,6 +31,53 @@ def validate(cls, value: str) -> str:
...
+    }
+
+    @classmethod
+    def __get_validators__(cls) -> Generator:  # pragma: no cover
+        yield cls.validate
+
+    @classmethod
+    def validate(cls, value: str) -> str:
+        """Assert text doesn't include an complex/malicious HTML."""
+        cleaned_str = clean(value, tags=HtmlStr.ALLOWED_HTML_TAGS)
```
adamsachs, line 77: two nits:
- still working on my python conventions but pretty sure the convention here is to use the `cls` self-reference arg in the classmethod 👍 (see suggestion)
- you didn't wanna keep the null/false check on `value` that you had in the previous iteration? i know that technically the edge case may be handled by `clean()` but feels like a decent defensive measure and to my eyes, it helps readability/reduces confusion, especially since this is a 3rd party lib. but i can see arguments for why this approach may be _cleaner_ (pun intended 😉)


```suggestion
        cleaned_str = clean(value, tags=cls.ALLOWED_HTML_TAGS)
```

clients/admin-ui/src/types/api/models/ExperienceConfigCreate.ts
allisonking: oh okay, this jogged a very dormant part of my memory and I remember talking about this before: https://github.com/ethyca/fides/issues/1169

adamsachs (commented):
some minor nits on the validation that i wouldn't consider blockers, generally looks good and i think you've plugged the validation into the right places!  i'm not an HTML sanitization expert but the library you're using looks like a good standard, and i trust your judgement there 👍

src/fides/api/custom_types.py
NevilleS: Thanks! Both good suggestions, will update now...

clients/admin-ui/src/types/api/models/ExperienceConfigCreate.ts
NevilleS: Updated to use explicit types in the mutation code that calls the API, instead of the API types. This feels like the best compromise - it uses the autogenerated types for 90% of the way, then hacks them up a bit to reflect the "reality" of how the API is used...

clients/admin-ui/src/features/privacy-experience/form/helpers.tsx
NevilleS: Thanks... I'm going to keep it as is because I find that really tricky Typescript magic can be self-defeating when it comes to maintaining!

NevilleS:
OK, I think I've got everything covered here. Switching to some final polish & UAT.

NevilleS:
OK, @mfbrown and I did some UAT together on the Admin UI form. Here's the main before/after screenshots for that:
| Before | After |
|---|---|
|image|image|

Lots of copy edits in there that @mfbrown and I did together. It's still a bit clumsy and wordy, but we felt that it was best to be "clear" before trying to get clever with wording and didn't have great ideas 😄

UAT complete ✅

src/fides/api/alembic/migrations/versions/f396c1f84b0f_add_banner_description_to_experience_.py
pattisdr: Reconciled with open PR https://github.com/ethyca/fides/pull/4481

PR #4547, opened 2024-01-05
Parameterize Salesforce Refresh Token

SteveDMurphy:
Closes PROD-1535

### Description Of Changes

Parameterizes the token refresh url for Salesforce, as it is different for Sandbox accounts (test.salesforce.com) as noted by many users have found when doing some early testing with sandbox accounts

Production and Dev accounts use the same refresh url (login.salesforce.com)

Additionally, some default data category updates are required that are included here which:
* remove some fields that cannot be updated as they are derived from others
* includes some other minor fields that were not being updated on erasure by default

I'm not sure how much I like having to label the categories as `system.operations` for fields that reflect personal data. It feels as if we are having a collision between identification of where personal data exists and the policy to erase said data. It would be interesting if something like the `skip` attribute of a Collection could be used at the Field level as well when derived or non-updatable on a Collection. That being said, I can sort of argue it both ways

### Code Changes

* [x] parameterize `refresh_token_url`
* [x] update default data categories for Salesforce

### Steps to Confirm

* [x] The salesforce connector is created as needed for both scenarios

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`


SteveDMurphy:
@MarcGEthyca @galvana - adding a reference to the docs changes here as well, curious about your thoughts -> https://github.com/ethyca/fidesdocs/pull/313

galvana (changes requested):
Hey @SteveDMurphy, this looks good! Just make sure you bump the version number on `salesforce_config.yml` so we can auto-update an existing instances on upgrade.

galvana (approved):
Thanks for bumping the version, looks ready to me!

PR #4554, opened 2024-01-10
GPP privacy notice support

adamsachs:
Closes https://ethyca.atlassian.net/browse/PROD-1587, https://ethyca.atlassian.net/browse/PROD-1585

Done along side plus PR: https://github.com/ethyca/fidesplus/pull/1281

Special note: this PR _removes_ the BE constraint/restriction of having privacy notices on the same region with conflicting data uses! the decision was made in conjunction with product. the constraint was not, at this point, providing significant value to end users, since it is primarily meant to avoid a narrow edge case in BE consent propagation, which is not heavily used (currently). on the other hand, enforcing the constraint is a complex and relatively fragile piece of BE code that was proving difficult to maintain, and had tension with some other functionality in this PR. if/when BE consent propagation is more mature, we may come back and look to re-enforce this constraint!

### Description Of Changes

Support for GPP privacy notices. The updates here in OSS:

- Add `gpp_field_mapping`(Optional, `JSONB[]`) and `framework` (Optional, `String`, enum but not enforced at the DB level) columns and fields to the base `PrivacyNotice` DB model and API scheme.
    - `framework` field indicates the framework that a given notice is associated with; currently only `gpp_us_national` and `gpp_us_state` are supported values, but in the future we'll likely have more framework-associated notices. this field helps both the BE and FE determine which notices are associated with a given GPP framework: the BE needs this for filtering the privacy notices based on global GPP settings, to effectively show/hide appropriate GPP notices in `/privacy-notice` and `/privacy-experience` endpoints; the FE needs it in the privacy notice management table to populate a new `framework` column. 
    - `gpp_field_mapping` contains the _details_ of the GPP nature of a given notice, i.e. what GPP fields the notices should "map" to for particular regions. the BE doesn't do anything with this field (at least not yet!) besides validate its basic structure; it's passed back as part of the `PrivacyNotice` response model, specifically within notice records in the `GET /privacy-experience` API. fides.js (our CMP FE) will then use this mapping to determine how to call the GPP API based on user interactions with our notices. This data used to be stored in a [big ol' dictionary](https://github.com/ethyca/fides/blob/e65fce4750a782f8247c933d65ac5e60f4ff5372/clients/fides-js/src/lib/gpp/constants.ts#L35) in the FE code base; it will be removed from there with this unit of work.
- we've added some new `PrivacyNoticeTemplate` records in the `privacy_notice_templates.yml` that represent the GPP privacy notices, per what's spec'd in the associated tickets. these templates have the above new fields populated accordingly. the templates are used to make "out of the box default" `PrivacyNotice` records on application startup.
- `data_uses` is no longer a required field on the base `PrivacyNotice` DB or API models - it should always default to an empty list if no value is provided by a user. One of the AC for this unit of work is to allow for privacy notices without a data use, which previous was not allowed.
    - Privacy notices without data uses still cannot be _enabled_, so there is (API) model validation in place to prevent that from happening, with a clear error message surfaced if that's not the case. There will be separate FE work to detect and correctly represent the "no data use" state of privacy notices so it's clear to end users what needs to change with the notices, and to prevent users from trying to enable notices before data uses are assigned.
- Per the note above, we are _removing_ the "conflicting data use" BE constraint/restriction for privacy notice regions altogether, based on a decision made with product. We judged that, at this point, the technical complexity/cost for maintaining the constraint was not worth the value it was providing.

QUESTIONS: 
 ~- [x] above question about disabled privacy notice template validation and how to best handle GPP notice templates with conflicts~ no longer applicable
- [x] what's the process for generating the notice template `id`s? mine are handcrafted for now, which goes against the seemingly randomized values for other notices, so this likely needs to be addressed
- [ ] product/design need to look over copy for a lot of template text

TODO:
- [x] need to remove privacy notice endpoints as this will be moving over to plus in the associated PR, since we now need the endpoint to reference GPP settings defined in plus.
- [ ] more test coverage!
~- [ ] should consider edge case where GPP privacy notice is _enabled_ but then GPP settings are subsequently changed so as to _hide_ the notice; then a user tries to enable to a conflicting notice and they get an error message that refers to a (now hidden) GPP notice that they can't really see or manage ... 🤔~ no longer applicable, as we are removing the conflict check/constraint on the BE. 

### Code Changes

- Above section covers the updates pretty comprehensively i think!

### Steps to Confirm
_In OSS_
- [x] confirmed new notice templates load well on server startup
- [x] confirmed privacy notice records created from notice templates look good

_Integration testing with alpha tag on plus_
- [x] confirmed new notice templates load well on server startup and create default notices
- [x] confirmed these notices show well in the admin UI
- [x] confirmed `gpp_field_mapping` is properly returned inline on notice records as part of `GET /privacy-experience` response payload for use by fides.js (CMP UI) 
- [x] confirmed `framework` is properly returned inline on notice records as part of `GET /privacy-notice` response payload, for use by admin UI
- [x] confirmed `framework` property can be used by migrated privacy notice endpoint in plus for filtering the notices returned based on global GPP settings
~- [ ] confirm data use validation with API calls to update privacy notices (automated test coverage here in OSS, but would be good to test it in action)~

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`


src/fides/api/util/consent_util.py
```diff
@@ -350,6 +360,11 @@ def create_privacy_notices_util(
                 model=privacy_notice,
                 fields=PRIVACY_NOTICE_ESCAPE_FIELDS,
             )
+        privacy_notice = transform_fields(
```
adamsachs, line 363: this feels a little clunky - but i wasn't finding an obvious way to do this "natively" in the pydantic model. this "manual" transformation doesn't deviate too much from the existing pattern we have for escaping.

src/fides/api/util/consent_util.py
```diff
@@ -458,13 +478,17 @@ def prepare_privacy_notice_patches(
         check_conflicting_data_uses(
             validation_updates,
             existing_notices.values(),
-            ignore_disabled=ignore_disabled,
+            all_data_uses=DataUse.query(db).all(),
+            # ignore_disabled=ignore_disabled, # TODO: figure out if we should re-enable disabled checking of templates, how to properly allow GPP overlaps
```
adamsachs, line 482: one of the major outstanding questions. i'm guessing we have good reason to _not_ ignore disabled on the template updates, so i'll need to figure out a way to let the GPP notice templates bypass this check

pattisdr (commented):
Looking good Adam I think I answered all your questions in-line, thanks as always for annotating your code to speed up review

> should consider edge case where GPP privacy notice is enabled but then GPP settings are subsequently changed so as to hide the notice

sounds fun! 😜  

---
Separately, I do think we should be considering ahead of time what this looks like when regions move to experiences and therefore what additional out of the box ExperienceConfigs will be created for the multilanguage work.  

For example, in the multilanguage work POC, the out of the box Experience Configs were not enough, we shipped with one overlay, one privacy center, and one TCF overlay config, and I had to create a separate overlay config to preserve everything being presented the same way https://github.com/ethyca/fides/compare/main...PROD-1458_multi_language_support#diff-a300585fdad6c844bc10a54b1ff5fbb487fded6bac752379fbd4dcbc27a29618

src/fides/api/util/consent_util.py
adamsachs: (see comment below - different transform, not escaping!)

src/fides/api/util/consent_util.py
adamsachs: ok thank you for finding that context! that's very helpful.

let me take a look and see how tricky or messy it will be to special-case the GPP notices. i suppose we may revisit in a couple of weeks anyway with the overhaul work, so it's probably not something we need to spend a ton of time figuring out?

src/fides/api/schemas/privacy_notice.py
```diff
@@ -28,13 +47,15 @@ class PrivacyNotice(FidesSchema):
...
-    data_uses: Optional[conlist(str, min_items=1)]  # type: ignore
+    data_uses: Optional[List[str]] = []
     enforcement_level: Optional[EnforcementLevel]
     disabled: Optional[bool] = False
     has_gpc_flag: Optional[bool] = False
     displayed_in_privacy_center: Optional[bool] = False
     displayed_in_overlay: Optional[bool] = False
     displayed_in_api: Optional[bool] = False
+    framework: Optional[PrivacyNoticeFramework]
+    gpp_field_mapping: Optional[List[GPPFieldMapping]]
```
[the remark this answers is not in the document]
adamsachs: > optional could set equal to None while we're here, I believe the future pydantic upgrade will require us to be more specific.

yeah, some of the optional/default behavior around this in pydantic was actually tripping me up quite a bit as i was developing! there were a few moving pieces though, so i should be able to add that in now and hopefully it doesn't break anything.

src/fides/api/schemas/privacy_notice.py
```diff
@@ -13,6 +14,24 @@
 from fides.api.schemas.base_class import FidesSchema
 
 
+class GPPMechanismMapping(FidesSchema):
+    field: str
+    not_available: str
+    opt_out: str
+    not_opt_out: str
```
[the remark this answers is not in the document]
adamsachs: hmmm, i may need to call in some GPP help from @allisonking but these felt pretty open-ended - maybe it's safe to say that the fields should be a string that consists only of digits?

src/fides/api/util/consent_util.py
```diff
@@ -286,16 +287,17 @@ def get_or_create_fides_user_device_id_provided_identity(
 
 def validate_notice_data_uses(
     privacy_notices: List[Union[PrivacyNoticeWithId, PrivacyNoticeCreation]],
-    db: Session,
+    valid_data_uses: Iterable[DataUse],
```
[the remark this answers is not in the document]
adamsachs: oh, i think this was basically an auxiliary change - just to DRY up/optimize the code a bit due to other updates. in  `create_privacy_notices_util` we now needed to query for all the data use resources anyway, so i figured i could pass those in to the function here anyway rather than re-querying for them within this function...

maybe that's an over-optimization though and it makes this function less self-contained, perhaps not worth the cost we pay in simplicity/maintainability....

src/fides/data/privacy_notices/privacy_notice_templates.yml
```diff
@@ -1,4 +1,400 @@
...
+  disabled: True
+  has_gpc_flag: False # TODO: not sure?
+  displayed_in_privacy_center: True
+  displayed_in_overlay: True
+  displayed_in_api: False
+  framework: gpp_us_state
+  gpp_field_mapping:
+    - region: us_ca
+      notice:
+        - SaleOptOutNotice
```
allisonking, line 24: I think we need `SharingOptOutNotice` here too
```suggestion
        - SaleOptOutNotice
        - SharingOptOutNotice
```

src/fides/data/privacy_notices/privacy_notice_templates.yml
```diff
@@ -1,4 +1,400 @@
...
+      notice:
+        - SensitiveDataLimitUseNotice
+        - SensitiveDataProcessingOptOutNotice
+      mechanism:
+        - field: SensitiveDataProcessing
+          not_available: 000000000000
+          opt_out: 111111111111
+          not_opt_out: 222222222222
+
+- name: Known Child Sensitive Data Consents (GPP, US National)
```
allisonking, line 368: only this one has the GPP parens in the name?

```suggestion
- name: Known Child Sensitive Data Consents
```

adamsachs:
noting here that in [7d208da](https://github.com/ethyca/fides/pull/4554/commits/7d208dad7a8e1ebd57be541dcdaa78c093bbe0cc) i've added some _questionable_ logic updates to now return privacy notices that have an overlapping _country_ with the privacy experience, rather than the exact matches we were looking for previously. the goal here is to allow us national notices to be returned as part of the experiences served to the experience of any US state. but the change has a bit broader impact, and it may not be the right way to go about this - there are a few different ways we could try to meet this requirement instead.

flagging this for further review...

src/fides/api/models/privacy_experience.py
```diff
@@ -273,8 +279,9 @@ def get_related_privacy_notices(
         """
         if self.component == ComponentType.tcf_overlay:
             return []
+
         privacy_notice_query = get_privacy_notices_by_region_and_component(
-            db, self.region, self.component  # type: ignore[arg-type]
+            db, [self.region.value, self.region_country], self.component  # type: ignore[arg-type]
```
adamsachs, line 284: NOTE: we should be careful here, this is a questionable change - need to review further...

src/fides/data/privacy_notices/privacy_notice_templates.yml
adamsachs: yes good catch, thank yoU!

src/fides/data/privacy_notices/privacy_notice_templates.yml
adamsachs: yup, just a miss in my cleanup i think. thanks for catching!

src/fides/api/schemas/privacy_notice.py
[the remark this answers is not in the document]
adamsachs: added in 0bf0d2d78f21f8dcb0a8261062fcefb95c2a87f9

src/fides/api/schemas/privacy_notice.py
```diff
@@ -13,6 +14,24 @@
...
+
+
+class GPPFieldMapping(FidesSchema):
+    region: str  # TODO: make PrivacyNoticeRegion enum, figure out serialization
```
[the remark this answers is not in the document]
adamsachs: added in 0bf0d2d78f21f8dcb0a8261062fcefb95c2a87f9

src/fides/api/util/consent_util.py
adamsachs: this ended up being pretty simple (i think), added a bypass only for disabled GPP notices in 4bfb89cbb8ee1035b3516094365d83625617c308

src/fides/api/alembic/migrations/versions/956d21f13def_update_privacy_notices_for_gpp.py
```diff
@@ -0,0 +1,79 @@
...
+        "privacynoticehistory",
+        "data_uses",
+        nullable=False,
+    )
+    op.drop_column("privacynoticehistory", "gpp_field_mapping")
+    op.alter_column(
+        "privacynotice",
+        "data_uses",
+        nullable=False,
+    )
```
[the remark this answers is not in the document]
adamsachs: fixed up in 12168fc909f98891f4eb3a3fcb6054b5dbd6a8cb

src/fides/api/schemas/privacy_notice.py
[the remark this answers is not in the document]
allisonking: ah okay so in the GPP spec these are technically either an integer (0,1,2) or an array of integers (i.e. [1,1,1], [2,2,2] etc, where the array can be any size). so for sure we only want digits, but we could even go farther into `int | list[int]` I think. right now I am casting the strings into either int or array of ints depending on the length of the string which works too, but might be worth firming up the type 

https://github.com/ethyca/fides/blob/8d891b380238e123c559ad530e2fae3e27b4ebfb/clients/fides-js/src/lib/gpp/us-notices.ts#L141-L145

src/fides/data/privacy_notices/privacy_notice_templates.yml
```diff
@@ -1,4 +1,400 @@
 privacy_notices:
+### GPP State-by-state ###
+- name: Data sales and sharing
+  notice_key: data_sales_sharing_gpp_us_state
+  id: pri_gpp_1 # TODO: update
```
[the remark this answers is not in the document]
adamsachs: updated mine in 923c532b1af30c12e2aca8ebd8386c7d35fda9b8 to be more similar to existing ids. i put a suffix that's bot a bit more than a few letters from the notice name since that's pretty ambiguous now (we have similar/overlapping names). the suffixes indicate GPP and us nat vs. state. i think that should be fine - let me know if it's problematic or seems odd in any way.

src/fides/data/privacy_notices/privacy_notice_templates.yml
```diff
@@ -1,4 +1,400 @@
...
+  description: We may transfer or share your personal information to third parties in exchange for monetary or other valuable consideration or for the purposes of [cross-context behavioral targeted advertising]. You can learn more about what information is used for this purpose in our privacy notice.
+  internal_description: Data sales and sharing (GPP, California) # TODO: update?
```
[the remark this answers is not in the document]
adamsachs: yeah i think i'll leave these for now, we can circle back with product before releasing to be sure our copy in general looks good for our notice templates, maybe at that point we can decide to bypass this field altogether....

src/fides/api/util/consent_util.py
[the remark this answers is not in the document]
adamsachs: now that we are no longer doing data use conflict validation, the refactor here wasn't providing any utility, so i've reverted in https://github.com/ethyca/fides/pull/4554/commits/4760aa8af379fd63bb6edc81e9107e08d445955c

adamsachs:
marking ready for review for your eyes again @pattisdr! i got more test coverage in place here, i want to have just a bit more coverage around the pydantic model updates but that shouldn't be blocking.

src/fides/api/models/privacy_experience.py
```diff
@@ -432,10 +406,10 @@ def new_experience_needed(
         )
 
         privacy_center_notices: Query = get_privacy_notices_by_region_and_component(
-            db, region, ComponentType.privacy_center
+            db, [region.value], ComponentType.privacy_center
```
pattisdr, line 409: I'm wondering if we need to pass in the region's country here too?  This function is making sure that a `PrivacyExperience` exists if a new region x component that didn't exist before was added to a Notice.  I just want to make sure we have "US" experiences too!

src/fides/api/models/privacy_experience.py
```diff
@@ -432,10 +406,10 @@ def new_experience_needed(
...
         )
         overlay_notices: Query = get_privacy_notices_by_region_and_component(
-            db, region, ComponentType.overlay
+            db, [region.value], ComponentType.overlay
```
pattisdr, line 412: Same thing here, haven't thought through the consequences, but if a PrivacyExperience object does not exist for a given region x component, then  notices will not be shown.

src/fides/api/util/consent_util.py
pattisdr: thanks for clarifying I missed this! 🔍   do any of the gpp field mappings sub fields need to be added to PRIVACY_NOTICE_ESCAPE_FIELDS. For example, gpp_field_mapping > notice.  This would complicate things though not saying we need it.

src/fides/api/util/consent_util.py
```diff
@@ -455,16 +466,14 @@ def prepare_privacy_notice_patches(
 
     # run the validation here on our proposed "dry-run" updates
     try:
-        check_conflicting_data_uses(
-            validation_updates,
-            existing_notices.values(),
-            ignore_disabled=ignore_disabled,
-        )
```
pattisdr, line 462: FYI This same file has the code that prefers opt out preferences in case of conflict for consent propagation https://github.com/ethyca/fides/blob/main/src/fides/api/util/consent_util.py#L97!

src/fides/data/privacy_notices/privacy_notice_templates.yml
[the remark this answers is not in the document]
pattisdr: Yep let's verify, current "out of the box" logic doesn't update these after the fact if there's a diff between notices and the templates, because we assume that their edits take precedence.

src/fides/data/privacy_notices/privacy_notice_templates.yml
```diff
@@ -1,4 +1,401 @@
...
+  regions:
+    - us_ca
+  consent_mechanism: opt_out
+  data_uses:
+    - marketing.advertising.first_party.targeted
+    - marketing.advertising.third_party.targeted
+  enforcement_level: not_applicable
+  disabled: True
+  has_gpc_flag: False # TODO: not sure?
```
pattisdr, line 16: Lots of things to check with product still looks like before release

pattisdr (commented):
OK big picture we need to make sure "US" experiences are being created I think? Currently this happens automatically when privacy notices are created/updated or the privacy experience config is created/updated  -

Also looks like copy is not ready for release -

src/fides/api/models/privacy_experience.py
adamsachs: yup i think you are right, i think this was a miss! not sure i gave it much though on the first pass through, i think i was trying to just touch as little as possible, but looking closer i don't see any reason why we _wouldn't_ also update this accordingly 👍

src/fides/api/util/consent_util.py
adamsachs: ah, i don't think so, i think we're good! the strings there should just be basic uppercase/lowercase alphabet characters. obviously wouldn't hurt but i don't think we need to build in that complexity for now 👍

adamsachs:
> OK big picture we need to make sure "US" experiences are being created I think? Currently this happens automatically when privacy notices are created/updated or the privacy experience config is created/updated -
> 
> Also looks like copy is not ready for release -

👍 thanks for the review @pattisdr ! i updated so that we should now get US experiences, confirmed we are getting them with some integration testing with fidesplus -- see DB records below. things are looking good when hitting the `GET /privacy-experiences` endpoint with GPP national and passing in `us_ca`, `us_ma` or just `us` as a region.
image

mfbrown (changes requested):
Left suggestions and new criteria for splitting out sensitive data and know child are in Jira.

src/fides/data/privacy_notices/privacy_notice_templates.yml
```diff
@@ -1,4 +1,401 @@
...
+  internal_description: Data sales and sharing (GPP, California)
+  regions:
+    - us_ca
+  consent_mechanism: opt_out
+  data_uses:
+    - marketing.advertising.first_party.targeted
+    - marketing.advertising.third_party.targeted
+  enforcement_level: not_applicable
+  disabled: True
+  has_gpc_flag: False # TODO: not sure?
```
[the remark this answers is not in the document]
adamsachs: updated in [c5771ff](https://github.com/ethyca/fides/pull/4554/commits/c5771ffe5b3ca922eb714a98c58c107804f7a596)

src/fides/data/privacy_notices/privacy_notice_templates.yml
```diff
@@ -1,4 +1,401 @@
...
+    - us_co
+    - us_ct
+    - us_ia
+    - us_ut
+    - us_va
+  consent_mechanism: opt_out
+  # data_uses: data uses must be applied manually
+  enforcement_level: not_applicable
+  disabled: True
+  has_gpc_flag: False # TODO: not sure?
```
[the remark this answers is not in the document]
adamsachs: updated in [c5771ff](https://github.com/ethyca/fides/pull/4554/commits/c5771ffe5b3ca922eb714a98c58c107804f7a596)

src/fides/data/privacy_notices/privacy_notice_templates.yml
adamsachs: ok i synced up with Michael and i think we've got it all covered! he'll take another pass but i believe we should be good to go on this front now, unless anything is still sticking out to you.

src/fides/api/schemas/privacy_notice.py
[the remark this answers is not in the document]
adamsachs: added some validation (string of integers) in [e129ad7](https://github.com/ethyca/fides/pull/4554/commits/e129ad7a96137ec177b8cbc8fb57f31ebb7a9fe9) 👍

pattisdr (approved):

mfbrown (approved):

PR #4556, opened 2024-01-11
PROD-1538: Adds ability to use base 64 cookie format

eastandwestwind:
Closes https://ethyca.atlassian.net/browse/PROD-1538

### Description Of Changes

Adds ability to use base64 cookie so that we send less data across all client <---> server packets. This is better for both performance and also decreases chance we'll get flagged by OWASP SQL injection rules.

We give our users the option to configure using base64 in a new env var: `FIDES_PRIVACY_CENTER__BASE_64_COOKIE` which defaults to `false`.

### Code Changes

* [ ] Adds new config var for base 64 cookie usage
* [ ] Updates every read / write cookie to handle base64 encoding. Note that with decoding, we always want to try pure JSON first, and if it fails, try base64 decoding. This is because config may have changed, between when the cookie was written and when the cookie was read.
* [ ] Update tests

### Steps to Confirm

* [ ] With `FIDES_PRIVACY_CENTER__BASE_64_COOKIE` set to `true`, in both Privacy center and consent banner, save consent, see cookie is written in base64 format. 
* [ ] Confirm that preferences persist correctly across page loads
* [ ] Additionally, with `FIDES_PRIVACY_CENTER__BASE_64_COOKIE` set to `false`, confirm no new regressions with other ways to set / save consent in PC

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### DIFF
--- a/clients/privacy-center/app/server-environment.ts
+++ b/clients/privacy-center/app/server-environment.ts
@@ -357,7 +357,7 @@ export const loadPrivacyCenterEnvironment =
         .FIDES_PRIVACY_CENTER__ALLOW_HTML_DESCRIPTION
         ? process.env.FIDES_PRIVACY_CENTER__ALLOW_HTML_DESCRIPTION === "true"
         : null,
-      BASE_64_COOKIE: process.env.BASE_64_COOKIE
+      BASE_64_COOKIE: process.env.FIDES_PRIVACY_CENTER__BASE_64_COOKIE
         ? process.env.FIDES_PRIVACY_CENTER__BASE_64_COOKIE === "true"
         : false,
     };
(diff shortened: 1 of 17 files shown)


clients/privacy-center/app/server-environment.ts
```diff
@@ -355,6 +357,9 @@ export const loadPrivacyCenterEnvironment =
         .FIDES_PRIVACY_CENTER__ALLOW_HTML_DESCRIPTION
         ? process.env.FIDES_PRIVACY_CENTER__ALLOW_HTML_DESCRIPTION === "true"
         : null,
+      BASE_64_COOKIE: process.env.BASE_64_COOKIE
+        ? process.env.BASE_64_COOKIE === "true"
```
allisonking, line 361: I think this should also have the `FIDES_PRIVACY_CENTER__` prefix, right?

allisonking (commented):
looks mostly good, but I think we should rename the env var to have the consistent prefix

clients/privacy-center/app/server-environment.ts
eastandwestwind: ah good catch!

allisonking (approved):
:shipit:

SteveDMurphy (approved):
re-certifying the review from @allisonking :shipit: also verified this live in multiple environments!

PR #4563, opened 2024-01-20
Add automated E2E code coverage for client applications

NevilleS:
### Description Of Changes

Experimenting with using `istanbul`, `@cypress/code-coverage`, and `codecov` together to pull automated coverage reports for the client codebases 👍 

### Code Changes

* [X] Add `istanbul` plugin to `admin-ui` and `privacy-center` projects, but only enable it in "test" builds used for Cypress (ie. `cy:start`)
* [X] Use Cypress' `code-coverage` plugin to publish coverage reports while running specs
* [X] Update GitHub CI (`frontend_checks.yml`) to collect coverage reports and upload to CodeCov
* [X] Modify CodeCov actions to tag `pytest` coverage with a "backend" flag
* [X] Modify CodeCov configuration with separate coverage targets for "backend" vs "admin-ui" vs "privacy-center"

### Steps to Confirm

* [X] View the CodeCov checks attached to this PR:
image

### Pre-Merge Checklist

* Documentation:
* [X] Update `CHANGELOG.md`


allisonking (approved):
Amazing! A few comments:

* Do you know if there's a way to also instrument our unit tests and combine coverage with cypress? We have some unit tests—not a lot, but some that cover nitty gritty logic of helper funcs—which would be nice to have covered as well. Though again, there aren't _that_ many (more so in fides.js) so I don't think this is a high priority
* I wonder if this can work to cover fides.js at all, given that its tests are in a different folder....
* I never added tests to the Regulations page (I know, I know! I'm a monster, but time was short and I'll do it as part of https://ethyca.atlassian.net/browse/PROD-1563 😄 ). it uses many of the same components as the Locations page, but I noticed that `regulations.tsx` doesn't show up in the code coverage report at all, probably because no cypress test visits it. So just flagging that our % may be inflated if there are parts of the app that aren't covered, but are also never visited by cypress, it looks like they just won't be included, as opposed to lowering our %
* overall, great increment! :shipit:

NevilleS:
> Do you know if there's a way to also instrument our unit tests and combine coverage with cypress?

NevilleS:
> Amazing! A few comments:
> 
> * Do you know if there's a way to also instrument our unit tests and combine coverage with cypress? We have some unit tests—not a lot, but some that cover nitty gritty logic of helper funcs—which would be nice to have covered as well. Though again, there aren't _that_ many (more so in fides.js) so I don't think this is a high priority

Yes, that's possible. For the pytest coverage we upload multiple reports to codecov and it aggregates it all together - although I find that kinda hard to "grok". A simpler way would be to run the unit tests & cypress tests from the same job so that the coverage report would aggregate locally on the GitHub runner and then upload together.

> * I wonder if this can work to cover fides.js at all, given that its tests are in a different folder....

I tried really briefly to do this. It was easy enough to add `rollup-plugin-istanbul` to fides.js, but had a couple issues:
* it immediately _explodes_ the bundle size and our build pipeline was very upset with me
* coverage reports didn't work when I ran Cypress tests from `/clients/privacy-center`; I have to assume it was confused by directories saying `/clients/fides-js` or something

Overall I punted on it pretty quickly.

> * I never added tests to the Regulations page (I know, I know! I'm a monster, but time was short and I'll do it as part of https://ethyca.atlassian.net/browse/PROD-1563 😄 ). it uses many of the same components as the Locations page, but I noticed that `regulations.tsx` doesn't show up in the code coverage report at all, probably because no cypress test visits it. So just flagging that our % may be inflated if there are parts of the app that aren't covered, but are also never visited by cypress, it looks like they just won't be included, as opposed to lowering our %

Interesting... I'd have hoped that coverage reports should include all files, especially those that aren't exercised by tests. I wonder if this might be caused less by the lack of a Cypress test and more if `cy:start` doesn't force those files to actually get built.

> * overall, great increment! :shipit:

😄

NevilleS:
Just pushed an update that'll now pull all the source files into the coverage report, and fully untested files show up as semi-transparent entries with 0% coverage:
image

However, it's still misleading - Istanbul doesn't include these untested files in the total coverage statistics, which you can basically see in that screenshot above... and this seems like an open issue with Istanbul in general: https://github.com/istanbuljs/nyc/issues/1415

...overall I hesitate to merge something in that is _actively_ misleading because it doesn't include untested files...

PR #4571, opened 2024-01-24
Do not render nav until we are finished loading

allisonking:
Closes https://ethyca.atlassian.net/browse/PROD-1592

### Description Of Changes

The ticket req was to make the management tab in the nav open by default. The reason why it was starting off closed was because, in general, we have the following accordion groups:

1. Overview
2. Data map
3. Privacy requests
4. Consent
5. Management

However, 4. Consent only shows up in plus environments. To figure out if we are in a plus environment, we have to query for it first. During the query time, the nav only sees 4 options, so it says "open 4 accordion tabs" and by the time the plus query says "yes we have plus, render Consent", only the first 4 tabs open up. 

Relatedly, [Michael also commented a while back](https://ethyca.slack.com/archives/C04S7K0T2V7/p1705085365822779):
> Is there any way to delay showing everything until the page is fully ready? I think menu options for plus features pop in a bit later than the other items. I think it might look cleaner if we determine what to display, then once that determination is made, the fully formed nav is made visible.

So, to feed two birds with one scone, we now render a blank nav until the plus query is ready. This makes sure that the management tab will open up, and also that the plus only menu options don't pop in later 🐦🍰🐦 

https://github.com/ethyca/fides/assets/24641006/68c84b45-96db-47e3-8ae7-931dbb7aeced

### Code Changes

* [x] In `MainSideNav`, query for plus health. While that's loading, render an empty nav box

### Steps to Confirm

* [ ] Load the admin-ui. The Management tab should start off open

### Pre-Merge Checklist

* [x] Update `CHANGELOG.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@ The types of changes are:
 
 ## [Unreleased](https://github.com/ethyca/fides/compare/2.28.0...main)
 
+### Changed
+- Delay rendering the nav until all necessary queries are finished loading [#4571](https://github.com/ethyca/fides/pull/4571)
+
 ## [2.28.0](https://github.com/ethyca/fides/compare/2.27.0...2.28.0)
 
 ### Added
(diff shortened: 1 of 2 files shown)


NevilleS (approved):
Pushed a _minor_ nit, but otherwise LGTM and good idea 👍

PR #4576, opened 2024-01-26
Fides Consent Multi-Language Feature Branch

pattisdr:
❗ Bump downrev before merge 
❗ This is a breaking change. Do not merge without corresponding frontend.

# Consent Multi-Language Feature Branch

### Description Of Changes

Data model changes to support consent multitranslation.

erd_consent_multitranslatino

### Code Changes

* [ ] Adds new columns and tables per ERD above but does not yet remove old columns.  Marks columns that will be deprecated.
* [ ] Adds core Notice and Privacy Experience create/update methods.
    * [ ] PrivacyNotice.create/update also create NoticeTranslations and PrivacyNoticeHistory records
    * [ ] ExperienceConfig.create/update create ExperienceTranslation and PrivacyExperienceConfigHistory records. Additionally, Notices and Regions are linked via these methods.
* [ ] Moves Privacy Experience Config endpoints to Plus
* [ ] Moves Privacy Experience Config and Privacy Notice schemas to Plus
* [ ] Moves out of the box Notices and Privacy Experience Config endpoints to Plus
* [ ] Removes the need to build TCF Privacy Experiences directly.  Now that consent is more Experience-driven than Notice drive, TCF Privacy Experiences will be built in Plus via out of the box Privacy Experience Configs
* [ ] Adds backwards-compatible `Experience.component` and `Experience.show_banner` properties to be surfaced on TCF experiences.
* [ ] Adds new ComponentTypes of `banner_and_modal` and `modal`. Deprecates `overlay` but does not remove.  Leaving for backwards compatibility.

### Steps to Confirm

* [ ] Test this over in Fidesplus https://github.com/ethyca/fidesplus/pull/1299

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`


src/fides/api/models/privacy_experience.py
```diff
@@ -97,36 +189,111 @@ def create(
...
-        PrivacyExperienceConfig record and also create a new PrivacyExperienceConfigHistory entry
+        Updates a PrivacyExperienceConfig which is a large exercise!
+
+        - Updates the PrivacyExperienceConfig details
+        - For each supplied translation, add, update, or delete translations so they match
+        the translations in the update request
+        - For each remaining translation, if the translation or the config has changed,
+        create a historical record for auditing purposes
+        - Link or unlink regions (via Privacy Experiences)
+        - Link or unlink Privacy Notices
```
pattisdr, line 244: Tentative drafts of the PrivacyExperienceConfig update method.  These are monster methods that create/update/delete numerous resources across multiple tables.  Further thought needed into the best way to do this. The create method has similar concerns.

src/fides/api/models/privacy_experience.py
```diff
@@ -327,32 +563,13 @@ def unlink_experience_config(self, db: Session) -> PrivacyExperience:
...
-        )
-        return self.update(  # type: ignore[return-value]
-            db,
-            data={
-                "experience_config_id": default_config.id if default_config else None,
-                "experience_config_history_id": default_config.experience_config_history_id
-                if default_config
-                else None,
-            },
-        )
```
pattisdr, line 354: We previously had a concept of a "default" Privacy Experience Config - one overlay, one privacy experience, and one tcf overlay.   This helped the backend make sure that Privacy Experiences could at least always have a default config linked to it as configuration changes were made in the background.  This concept is going away.

src/fides/api/models/privacy_experience.py
```diff
@@ -363,81 +580,19 @@ def get_privacy_notices_by_region_and_component(
...
-                    PrivacyNotice.displayed_in_overlay,
-                ),
-                and_(
-                    component == ComponentType.privacy_center,
-                    PrivacyNotice.displayed_in_privacy_center,
-                ),
-            )
+            PrivacyExperience.region.in_(regions),
+            PrivacyExperience.component == component,
         )
```
pattisdr, line 592: *Where* a notice is shown is now defined on the experience side of things instead of the notice side of things, so the query here has changed.

src/fides/api/models/privacy_experience.py
```diff
@@ -363,81 +580,19 @@ def get_privacy_notices_by_region_and_component(
...
-def upsert_privacy_experiences_after_notice_update(
-    db: Session, affected_regions: List[PrivacyNoticeRegion]
-) -> List[PrivacyExperience]:
-    """
-    Keeps Privacy Experiences in sync with *PrivacyNotices* changes.
-    Create or update "overlay" or "privacy center" PrivacyExperiences based on the PrivacyNotices in the "affected_regions".
-    To be called whenever PrivacyNotices are created or updated (pass in any regions that were potentially affected)
-
-    PrivacyExperiences should not be deleted.  It's okay if no notices are associated with an Experience.
-    """
```
pattisdr, line 391: "PrivacyExperiences" aren't updated directly by the customer.  Previously, updates to Notices or updates to the PrivacyExperienceConfig could cause updates to the PrivacyExperiences in the background to make sure we always had a PrivacyExperience container for surfacing results to the end user.

Now, updates to PrivacyNotices do not affect which PrivacyExperiences should exist since "where" a notice is displayed is only defined on the PrivacyExperienceConfig.  Any places where notices are being created or updated no longer need to call this method.

src/fides/api/models/privacy_experience.py
```diff
@@ -520,3 +672,28 @@ def region_country(region: str) -> str:
     based on naming convention (i.e. `country_subregion`)
     """
     return region.split("_")[0]
+
+
+def link_notices_to_experience_config(
+    db: Session,
+    notice_keys: List[str],
+    experience_config: PrivacyExperienceConfig,
+) -> Tuple[List[PrivacyNoticeRegion], List[PrivacyNoticeRegion]]:
```
pattisdr, line 681: This helper method lets us add Privacy Notices to a Privacy Experience Config (and unlinks notices not supplied)

src/fides/api/models/privacy_notice.py
```diff
@@ -24,6 +22,20 @@
...
 
+class Language(Enum):
+    """Using BCP 47 Language Tags
+
+    TODO propose adding this list:
+    https://www.techonthenet.com/js/language_tags.php
+    """
+
+    en_us = "en_us"  # US English
+    en_gb = "en_gb"  # British English
```
pattisdr, line 33: Adrian brought up BCP 47 which looked good to me, I was thinking of adding underscored lowercase versions of this format here?

src/fides/api/models/privacy_notice.py
```diff
@@ -332,29 +351,152 @@ def create(
         data: dict[str, Any],
         check_name: bool = False,
     ) -> PrivacyNotice:
+        """
+        Creates a Privacy Notice and then a NoticeTranslation record for each supplied Translation.
+
+        For each Translation a historical record is created which combines details from the Notice and
+        the Translation for auditing purposes.  Privacy preferences are saved against the PrivacyNoticeHistory
+        record which contains the full details."""
+        translations = data.pop("translations", []) or []
```
pattisdr, line 360: PrivacyNotice create method has been updated to also create translations, and then create the PrivacyNoticeHistory record from the combination of the notice and the translation

src/fides/api/models/privacy_notice.py
```diff
@@ -332,29 +351,152 @@ def create(
...
+
         return created
 
     def update(self, db: Session, *, data: dict[str, Any]) -> PrivacyNotice:
         """
-        Overrides the base update method to automatically bump the version of the
-        PrivacyNotice record and also create a new PrivacyNoticeHistory entry
+        Updates the Privacy Notice
+        - Upserts or deletes supplied translations to match translations in the request
+        - For each remaining translation, create a historical record if the base notice or translation changed.
```
pattisdr, line 382: Privacy Notice update method has similar changes.  Multiple resources in other tables are being upserted here, need to revisit with the best way to do this.

src/fides/api/models/privacy_notice.py
```diff
@@ -365,35 +507,27 @@ def check_conflicting_notice_keys(
     ignore_disabled: bool = True,  # For PrivacyNoticeTemplates, set to False
 ) -> None:
     """
-    Checks to see if new notice keys will conflict with any existing notice keys for a specific region
+    Checks to see if new notice keys will conflict with any existing notice keys
```
pattisdr, line 510: Privacy Notices no longer have regions.  We previously had a constraint that didn't allow overlapping notice keys in the same region for enabled notices.  A big question can we enforce that notice keys be unique table wide, like a fides key?  

Alternatively, when adding notices to an experience config, we could add validation there that makes sure no notices added have the same notice key.

src/fides/api/schemas/privacy_notice.py
```diff
@@ -119,40 +130,6 @@ def validate_notice_key(cls, values: Dict[str, Any]) -> Dict[str, Any]:
...
-        cls, values: Dict[str, Any]
-    ) -> Dict[str, Any]:
-        """
-        Add some validation regarding where certain consent mechanisms must be displayed
-        """
-        consent_mechanism: Optional[str] = values.get("consent_mechanism")
-        displayed_in_overlay: Optional[bool] = values.get("displayed_in_overlay")
-        displayed_in_privacy_center: Optional[bool] = values.get(
-            "displayed_in_privacy_center"
-        )
```
pattisdr, line 133: This is logic that was at the notice level but no longer makes sense.  I assume will likely need something similar at the experience level?

src/fides/api/util/consent_util.py
```diff
@@ -575,62 +624,75 @@ def load_default_notices_on_startup(
...
+
+        # Determine which templates don't have corresponding records in Privacy Experience Config
+        new_templates: List[ExperienceConfigTemplate] = [
+            template
+            for template in privacy_experience_config_templates
+            if template.id
+            not in [
+                origin for (origin,) in db.query(PrivacyExperienceConfig.origin).all()
+            ]
+        ]
```
pattisdr, line 658: We previously didn't create Experience Config Templates, only notice-level templates. However, as we are shipping with out-of-the-box translations on the plus side, it would be helpful to have these stored in the database. New logic upserts the templates, and then creates Experience Configs only if they are new.  Data migration needs to backfill these templates, and link the origin.

src/fides/data/privacy_notices/privacy_experience_config_defaults2.yml
```diff
@@ -0,0 +1,160 @@
...
+    - language: en_us
+      accept_button_label: Opt in to all
+      acknowledge_button_label: OK
+      description: We use cookies and similar methods to recognize visitors and remember their preferences. We may also use them to measure ad campaign effectiveness, target ads, and analyze site traffic. Depending on your location, you may opt-in or opt out of the use of these technologies.
+      privacy_policy_link_label: Privacy Policy
+      privacy_policy_url: null
+      privacy_preferences_link_label: Manage preferences
+      reject_button_label: Opt out of all
+      save_button_label: Save
+      title: Manage your consent preferences
```
pattisdr, line 55: Still defined in OSS for the POC, only english translations here.   We can talk further about where this file should live.

src/fides/api/schemas/privacy_experience.py
```diff
@@ -82,30 +97,43 @@ class ExperienceConfigCreate(ExperienceConfigSchema):
...
+    ) -> List[PrivacyNoticeRegion]:
+        """Assert regions aren't duplicated."""
+        if regions and len(regions) != len(set(regions)):
+            raise ValueError("Duplicate regions found.")
+        return regions
 
     @root_validator
     def validate_attributes(cls, values: Dict[str, Any]) -> Dict[str, Any]:
         """Validate minimum set of required fields exist given the type of component"""
         component: Optional[ComponentType] = values.get("component")
```
pattisdr, line 117: We need to figure out what fields are required when adding these new translations -

pattisdr (commented):
This is an early look at the incoming model-level Multi-language support changes @eastandwestwind @adamsachs @jpople 

Main early things I need to get clear on are what the frontend needs from me as far as "component" types and logic changes around whether I return whether a banner should be enabled.

src/fides/api/app_setup.py
```diff
@@ -67,12 +66,12 @@
...
     dirname(__file__),
     "../data/privacy_notices",
-    "privacy_notice_templates.yml",
+    "privacy_notice_templates2.yml",
 )
 PRIVACY_EXPERIENCE_CONFIGS_PATH = join(
     dirname(__file__),
     "../data/privacy_notices",
-    "privacy_experience_config_defaults.yml",
+    "privacy_experience_config_defaults2.yml",
```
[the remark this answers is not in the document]
pattisdr: This may or may not be straightforward as to what this looks like.  For the "generic" notices I ended up having to create a new out of the box Experience Config to preserve the same options that existed when regions + displayed_in* were stored on the notice.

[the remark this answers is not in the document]
pattisdr: Working on a draft of this now want to get ahead of it

.fides/14acee6f5459_translation_data.py
```diff
@@ -0,0 +1,312 @@
...
+def generate_record_id(prefix):
+    return prefix + "_" + str(uuid.uuid4())
+
+
+def determine_needed_experience_configs(bind):
+    """
+    Our data model changes necessitate the creation of new Experience Configs. Now that "where" a Notice is served
+    is being moved from the "Notice" constructs to "Experiences", existing Experience Configs are not sufficient
+    to capture the full range of configuration on current Notices.
+    """
```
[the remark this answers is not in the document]
pattisdr: yep this will definitely need to be revisited in terms of GPP.  There was some very specific logic related to some existing out-of-the-box privacy experience configs, but the addition of GPP notices changes which out-of-the-box configs will need to be created.

src/fides/api/api/v1/endpoints/privacy_experience_config_endpoints.py
```diff
@@ -145,7 +155,7 @@ def experience_config_create(
     *,
     db: Session = Depends(deps.get_db),
     experience_config_data: ExperienceConfigCreate,
-) -> ExperienceConfigCreateOrUpdateResponse:
+) -> PrivacyExperienceConfig:
```
pattisdr, line 158: I'm doing early work on these privacy experience config endpoints help get the schemas more settled.  More testing, validation will need to be added in and I also assume Experience Config Endpoints should be ported over to Plus as well.

src/fides/data/privacy_notices/privacy_experience_config_defaults2.yml
adamsachs: yeah, agreed we can settle on it later, but i might vote for
- separating out translations into their own file, to keep this file more manageable/consumable
- porting both files over to plus now that all of our related endpoints _should_  go to plus, too 👍

src/fides/data/privacy_notices/privacy_notice_templates2.yml
```diff
@@ -0,0 +1,470 @@
...
+  enforcement_level: not_applicable
+  disabled: True
+  has_gpc_flag: False
+  framework: gpp_us_state
+  translations:
+    - language: en_us
+      title: Known Child Sensitive Data Consents
+      description: We may process the personal information of children under 16. You can learn more in our privacy policy.
+  gpp_field_mapping:
+    - region: us_ca
```
adamsachs, line 270: a _little bit_ odd that regions are referenced here in this mapping even though regions are no longer directly on the notice, but i think that's alright!

.fides/14acee6f5459_translation_data.py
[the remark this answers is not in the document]
adamsachs: yes this is certainly a lot to think through...i'm having a bit of trouble grokking everything here, i can continue to look through it and it will probably sink in, maybe we should also just sync up a bit and you can talk me through some of the design/thought here.

i'm confident you've got a solid plan here, just want to make sure i'm tracking what we'll be doing and also perhaps i can help with GPP related adjustments 👍

adamsachs (commented):
@pattisdr this is looking really good, thank you for working through this in a concrete way and providing visibility into you're proposed updates! there are still areas i will need to get a more concrete understanding of, but i'm certainly starting to get a general grasp of what'll be updated, and i left some (rather vague) comments on some particular points -- feel free to take them or leave them at this point!

it'd be good to sync up and talk through:
- the migration approach
- the work you've scoped out for next sprint to get this in place
- what i can best help with, beyond general review and feedback

.fides/14acee6f5459_translation_data.py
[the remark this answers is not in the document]
pattisdr: sorry @adamsachs I didn't mean to imply this was in a "reviewable" state, this was just a messy POC of me trying to prove to myself it is possible!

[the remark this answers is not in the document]
adamsachs: 😄 that's ok! i didn't feel compelled based on anything you'd said, i just wanted to see if i could start wrapping my head around the problems you're trying to wrap your head around :)

.fides/14acee6f5459_translation_data.py
```diff
@@ -0,0 +1,312 @@
...
+    else:
+        next_group["experience_config_id_after"] = pd.Series(dtype=str)
+    return next_group
+
+
+def upgrade():
+    return
+    bind = op.get_bind()
+
+    experience_config_df: DataFrame = determine_needed_experience_configs(bind)
```
pattisdr, line 136: Main point: 

- We need to create a unique set of Experience Configs to surface the same number of notices that we were surfacing before.
- Source of the new Experience Config
- One ExperienceConfig will stay?

.fides/14acee6f5459_translation_data.py
```diff
@@ -0,0 +1,312 @@
...
+
+    # Translate "disabled" concept from the Experience Config -> Privacy Experience.
+    set_experience_disabled = text(
+        """
+        UPDATE privacyexperience
+        SET disabled = privacyexperienceconfig.disabled
+        FROM privacyexperienceconfig
+        WHERE privacyexperience.experience_config_id = privacyexperienceconfig.id;
+        """
+    )
```
pattisdr, line 212: Requirements have changed here.  "Disabled" is staying at the shared ExperienceConfig level instead of moving to the region-specific Experience level.

src/fides/api/models/privacy_experience.py
```diff
@@ -168,22 +408,103 @@ def get_default_config(
...
-    There can only be one component per region.
+    There can only be one component per region.  Most of the details for a given Experience
+    are on the ExperienceConfig (which allows multiple Experiences to share the same
+    configuration options).
     """
 
-    component = Column(EnumColumn(ComponentType), nullable=False)
+    component = Column(
+        EnumColumn(ComponentType), nullable=False
+    )  # I think we should delete this now.
```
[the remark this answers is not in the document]
adamsachs: @pattisdr that makes sense to me! simplifying `PrivacyExperience` to be as close to a "join table" as possible definitely helps make the overall schema easier to work with and make sense of, so if you're not finding any utility for this here, then i'm fully in favor of removing 👍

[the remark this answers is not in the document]
pattisdr: ok perfect - leaving this here for now to be removed in the final schema migration which deletes all the obsolete fields, but in the interim, I am rewriting any related queries to look at the component of the attached PrivacyExperienceConfig instead.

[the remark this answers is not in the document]
pattisdr: Actually I think this is one field that makes sense to remove now

src/fides/api/models/privacy_experience.py
pattisdr: less concerned about performance here, these methods are for setting up privacy experiences, something that will occur by the customer occasionally. 

It's that normally don't like one model to be updating a bunch of other models (maybe hooks are better here), and there's also a chance something happens midway through and only some of these models are updated. 

But the more I've thought about it, these are core updates that need to stay together in one place, so I'm leaning towards leaving as-is

src/fides/data/privacy_notices/privacy_experience_config_defaults2.yml
pattisdr: update here, these files and endpoints are now in plus, still in the same file since we just have the one english translation, but I imagine we will revisit this as we look to supporting more!

src/fides/api/api/v1/endpoints/privacy_experience_config_endpoints.py
pattisdr: Moved!

src/fides/api/api/v1/endpoints/privacy_experience_config_endpoints.py
```diff
@@ -44,6 +46,10 @@
 router = APIRouter(tags=["Privacy Experience Config"], prefix=urls.V1_URL_PREFIX)
 
 
+def get_configured_locations(db: Session = Depends(deps.get_db)) -> Set:
```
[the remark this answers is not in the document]
pattisdr: Currently tabling this for the follow-up locations ticket 👍

src/fides/data/privacy_notices/privacy_notice_templates2.yml
pattisdr: I agree these still feel like notice-specific details

scripts/verify_fideslang_2_data_migration.py
pattisdr, line 1: Schemas used in this file have been ported to plus. I am deleting the migration verification script in response.

scripts/verify_privacy_preferences_overhaul_migration.py
pattisdr, line 1: Similarly, schemas in this file have been ported to plus and I am removing this script that verifies an old migration.

src/fides/api/api/v1/endpoints/privacy_experience_config_endpoints.py
pattisdr, line 1: Privacy Experience Config Endpoints have moved to Plus and have been transformed there.

src/fides/api/app_setup.py
```diff
@@ -185,14 +169,8 @@ async def run_database_startup(app: FastAPI) -> None:
     finally:
         db.close()
 
-    load_default_experience_configs()  # Must occur before loading default privacy notices
-
     if not CONFIG.test_mode:
-        # Default notices subject to change, so preventing these from
-        # loading in test mode to avoid interfering with unit tests.
-        load_default_privacy_notices()
```
pattisdr, line 193: Out of the box configs have moved to Plus.  Previously Experience Configs were loaded first, but now that Notices are attached to Experience Configs, in plus, we flip this order.

src/fides/api/models/privacy_experience.py
```diff
@@ -1,94 +1,252 @@
...
     privacy_center = "privacy_center"
-    tcf_overlay = "tcf_overlay"
+    tcf_overlay = "tcf_overlay"  # TCF Banner + modal combined
+
+
+# Fides JS UX Types - there should only be one of these defined per region
+FidesJSUXTypes: List[ComponentType] = [
+    ComponentType.banner_and_modal,
+    ComponentType.modal,
+]
```
pattisdr, line 82: Plus adds application behavior to prevent a region from being on more than one Fides JS UX Type at a time.  This was previously more of a database constraint.

src/fides/api/models/privacy_experience.py
```diff
@@ -1,94 +1,252 @@
...
 
 
-class BannerEnabled(Enum):
```
pattisdr, line 37: The BannerEnabled concept is gone.  Now whether we load a banner is purely a product of the component type. Previously it was a combination of this field, and other more advanced attributes like notices configured.

src/fides/api/models/privacy_experience.py
pattisdr, line 1: A lot of reorganization here - this file is better viewed as a whole instead of just the diff

src/fides/api/models/privacy_experience.py
```diff
@@ -1,94 +1,252 @@
...
+    dismissable = Column(Boolean, nullable=False, default=True, server_default="t")
+    name = Column(String)
+
+
+class ExperienceConfigTemplate(PrivacyExperienceConfigBase, Base):
+    """Table for out-of-the-box Experience Configurations"""
+
+    privacy_notice_keys = Column(
+        ARRAY(String)
+    )  # A list of notice keys which should correspond to a subset of out-of-the-box notice keys
```
pattisdr, line 117: My current draft of creating Experience Config out of the box templates specifies the notices by key in the template, not by id, to make this more human readable.  However, when creating Privacy Experience Configs directly, the API request should specify notice ids.  There's also a side constraint right now that out of the box Notices need to have unique keys.

src/fides/api/models/privacy_experience.py
```diff
@@ -1,94 +1,252 @@
...
     reject_button_label = Column(String)
     save_button_label = Column(String)
     title = Column(String)
-    version = Column(Float, nullable=False, default=1.0)
 
 
-class PrivacyExperienceConfig(ExperienceConfigBase, Base):
-    """Stores common copy to be shared across multiple regions, e.g.
-    banner titles, descriptions, button labels, etc.
+class DeprecatedConfigFields:
```
pattisdr, line 155: A follow-up commit after we get the data migration in, will remove this class, and other fields marked deprecated.

src/fides/api/models/privacy_experience.py
```diff
@@ -1,94 +1,252 @@
...
+            db.query(ExperienceTranslation)
+            .filter(
+                ExperienceTranslation.language == language,
+                ExperienceTranslation.experience_config_id == self.id,
+            )
+            .first()
+        )
+
     @classmethod
     def create(
```
pattisdr, line 251: This is one of the primary methods of this PR, PrivacyExperienceConfig.create

src/fides/api/models/privacy_experience.py
```diff
@@ -97,38 +255,131 @@ def create(
...
+        Updates a PrivacyExperienceConfig and related resources which is a large exercise!
+
+        - Updates the PrivacyExperienceConfig details
+        - For each supplied translation, add, update, or delete translations so they match
+        the translations in the update request
+        - For each remaining translation, if the translation or the config has changed,
+        create a historical record for auditing purposes
+        - Link or unlink regions (via Privacy Experiences)
+        - Link or unlink Privacy Notices
         """
```
pattisdr, line 312: This is one of the primary methods of this PR, PrivacyExperienceConfig.update

src/fides/api/models/privacy_experience.py
```diff
@@ -141,67 +392,158 @@ def dry_update(self, *, data: dict[str, Any]) -> PrivacyExperienceConfig:
...
-        history: PrivacyExperienceConfigHistory = self.histories.filter_by(  # type: ignore # pylint: disable=no-member
-            version=self.version
-        ).first()
-        return history.id if history else None
-
-    @classmethod
-    def get_default_config(
-        cls, db: Session, component: ComponentType
-    ) -> Optional[PrivacyExperienceConfig]:
-        """Load the first default config of a given component type"""
```
pattisdr, line 162: This concept has been removed, the "default" Experience Config.  We used to ship with one default Config of each Component Type, but now we ship with many ExperienceConfigs, there are no "default" types.

src/fides/api/models/privacy_experience.py
```diff
@@ -141,67 +392,158 @@ def dry_update(self, *, data: dict[str, Any]) -> PrivacyExperienceConfig:
...
+    """Experience Config History table for auditing purposes.
 
-    experience_config_id = Column(
-        String, ForeignKey(PrivacyExperienceConfig.id_field_path), nullable=False
+    When an Experience Config and/or an Experience Translation is modified, a new version is
+    created here that stores a snapshot of these records combined.  Consent reporting can contain
+    an id to this resource which preserves the details of the Experience viewed by the end user.
+    """
+
+    custom_asset_id = Column(String, ForeignKey(CustomAsset.id_field_path))
```
pattisdr, line 485: `custom_asset_id` is not a field that has been explored at all or exposed in the API, we should revisit this concept before merge.

src/fides/api/models/privacy_experience.py
```diff
@@ -141,67 +392,158 @@ def dry_update(self, *, data: dict[str, Any]) -> PrivacyExperienceConfig:
...
         String,
         ForeignKey(PrivacyExperienceConfig.id_field_path),
-        nullable=True,
+        nullable=True,  # Needs an ExperienceConfig to be valid though
         index=True,
     )
 
-    # There can only be one component per region
-    __table_args__ = (UniqueConstraint("region", "component", name="region_component"),)
-
```
pattisdr, line 198: We used to have a unique constraint here on region and component. The component field has been removed, and now PrivacyExperiences are purely 1) how regions are linked to a PrivacyExperienceConfig and 2) the primary artifact queried by the end user because it is region-specific.

src/fides/api/models/privacy_experience.py
```diff
@@ -219,260 +561,41 @@ class PrivacyExperience(Base):
...
-                PrivacyNotice.consent_mechanism.in_(BANNER_CONSENT_MECHANISMS)
-            ).count()
-        )
-
     @staticmethod
-    def create_default_experience_for_region(
-        db: Session, region: PrivacyNoticeRegion, component: ComponentType
-    ) -> PrivacyExperience:
-        """Creates an Experience for a given Component Type and Region and links
-        to default copy (ExperienceConfig)"""
```
pattisdr, line 271: Again, the "default" Experience Config concept no longer exists.

src/fides/api/models/privacy_experience.py
```diff
@@ -219,260 +561,41 @@ class PrivacyExperience(Base):
...
-    def unlink_experience_config(self, db: Session) -> PrivacyExperience:
-        """Removes the experience config
-
-        Note that neither the Experience Config or Experience are deleted; we're only removing
-        a FK if it exists.
-        """
-        return self.update(  # type: ignore[return-value]
-            db,
-            data={"experience_config_id": None, "experience_config_history_id": None},
-        )
```
pattisdr, line 334: We previously created an Experience for a region or a component that could be linked or unlinked from various ExperienceConfigs.  Now we just delete the Experience if a region is removed from a PrivacyExperienceConfig.  As a result, PrivacyPreferences no longer have a FK to this object. By itself, it doesn't add information you don't have from the PrivacyExperienceConfig and user geography.

src/fides/api/models/privacy_experience.py
```diff
@@ -219,260 +561,41 @@ class PrivacyExperience(Base):
...
-
-    @staticmethod
-    def get_overlay_and_privacy_center_experience_by_region(
-        db: Session, region: str
-    ) -> Tuple[Optional[PrivacyExperience], Optional[PrivacyExperience]]:
-        """Load both the overlay and privacy center experience for a given region
-
-        TCF overlays are not returned here.  This method is used in building experiences when Notices
-        are created, which is not applicable for TCF.
-        """
```
pattisdr, line 310: Logic not really relevant anymore.  There are more and different experiences than this now, and the reason for this method doesn't exist anymore either.

src/fides/api/models/privacy_experience.py
```diff
@@ -219,260 +561,41 @@ class PrivacyExperience(Base):
...
-def upsert_privacy_experiences_after_notice_update(
-    db: Session, affected_regions: List[PrivacyNoticeRegion]
-) -> List[PrivacyExperience]:
-    """
-    Keeps Privacy Experiences in sync with *PrivacyNotices* changes.
-    Create or update "overlay" or "privacy center" PrivacyExperiences based on the PrivacyNotices in the "affected_regions".
-    To be called whenever PrivacyNotices are created or updated (pass in any regions that were potentially affected)
-
-    PrivacyExperiences should not be deleted.  It's okay if no notices are associated with an Experience.
-    """
```
pattisdr, line 391: Previously, both edits to a PrivacyNotice and a PrivacyExperienceConfig could cause PrivacyExperiences to be created/updated in the background to make sure we always had a valid PrivacyExperience.

Now, only ExperienceConfigs create PrivacyExperiences, when a region is added to a PrivacyExperienceConfig.

src/fides/api/models/privacy_experience.py
```diff
@@ -219,260 +561,41 @@ class PrivacyExperience(Base):
...
-
-    for experience in experiences_to_unlink:
-        if experience_config.is_default:
-            experience.unlink_experience_config(db)
-        else:
-            experience.link_default_experience_config(db)
-    return [experience.region for experience in experiences_to_unlink]  # type: ignore[misc]
 
 
 def upsert_privacy_experiences_after_config_update(
```
pattisdr, line 588: This method still links PrivacyExperiences to PrivacyExperienceConfigs but a new change is that it also deletes them too, when a region is removed. (Previously we would just unlink the FK).  Now there can be multiple PrivacyExperiences with the same region attached to multiple PrivacyExperienceConfigs as long as there is no overlap in enabled Experience Configs of the same Fides JS UX Type.

src/fides/api/models/privacy_notice.py
pattisdr, line 1: This file is also better viewed as a whole

src/fides/api/models/privacy_notice.py
```diff
@@ -230,24 +251,12 @@ def dry_update(self, *, data: dict[str, Any]) -> FidesBase:
...
         # create a new object with the updated attribute data to keep this
         # ORM object (i.e., `self`) pristine
-        return PrivacyNotice(**cloned_attributes)
-
-    def validate_enabled_has_data_uses(self) -> None:
-        """Validated that enabled privacy notices have data uses"""
-        if not self.disabled and not self.data_uses:
-            raise ValidationError(
-                "A privacy notice must have at least one data use assigned in order to be enabled."
-            )
```
pattisdr, line 244: I made this a Pydantic validator over in plus -

src/fides/api/models/privacy_notice.py
```diff
@@ -324,6 +335,28 @@ def systems_applicable(self) -> bool:
...
+                ExperienceNotices,
+                PrivacyExperience.experience_config_id
+                == ExperienceNotices.experience_config_id,
+            )
+            .filter(ExperienceNotices.notice_id == self.id)
+            .group_by(PrivacyExperience.region)
+            .order_by(PrivacyExperience.region.asc())
+        )
+        return [region[0] for region in configured_regions]
+
```
pattisdr, line 359: This is the property that is used to build the "configured" attribute in the Privacy Notice List view, useful for displaying the number of locations using it.  Previously we could just read the regions straight off the Notice, but it is no longer stored there.

src/fides/api/models/privacy_notice.py
```diff
@@ -324,6 +335,28 @@ def systems_applicable(self) -> bool:
...
     @classmethod
     def create(
```
pattisdr, line 361: This is a core method of this PR, PrivacyNotice.create

src/fides/api/models/privacy_notice.py
```diff
@@ -332,123 +365,249 @@ def create(
...
+            )
 
-        # create the history after the initial object creation succeeds, to avoid
-        # writing history if the creation fails and so that we can get the generated ID
-        data.pop("id", None)
-        history_data = {**data, "privacy_notice_id": created.id}
-        PrivacyNoticeHistory.create(db, data=history_data, check_name=False)
         return created
 
     def update(self, db: Session, *, data: dict[str, Any]) -> PrivacyNotice:
```
pattisdr, line 397: This is a core method of this PR that received major changes, PrivacyNotice.update

src/fides/api/models/privacy_preference.py
```diff
@@ -20,10 +20,19 @@
...
+
+    Not at the db level due to being subject to change.
+    Only add here, do not remove
+    """
+
     privacy_center = "privacy_center"
-    overlay = "overlay"
+    overlay = "overlay"  # DEPRECATED. DO NOT REMOVE.
+    modal = "modal"
+    banner_and_modal = "banner_and_modal"
```
pattisdr, line 35: New modal and banner_and_modal types added. Overlay is deprecated but cannot be removed for backwards compatibility.

src/fides/api/models/privacy_preference.py
```diff
@@ -41,12 +50,19 @@ class ConsentMethod(Enum):
...
+    This differs from component type because we want to record exactly
+    where consent was served.
+
+    Not at the db level due to being subject to change.
+    Only add here, do not remove
+    """
+
+    overlay = "overlay"  # DEPRECATED. DO NOT REMOVE.
     banner = "banner"
+    modal = "modal"
```
pattisdr, line 63: This previous Enum already had more granular options than ComponentType because it was used to communicate more specific detail about where a Notice was served - adding the modal type here.  I'd think going forward, FE would specify banner OR modal OR TCF overlay OR privacy center

src/fides/api/models/privacy_preference_v2.py
```diff
@@ -206,10 +216,10 @@ def privacy_experience_config_history_id(cls) -> Column:
...
 
     # The specific experience under which the user was presented the relevant notice
-    # Minimal information stored here, mostly just region and component type
-    @declared_attr
-    def privacy_experience_id(cls) -> Column:
-        return Column(String, ForeignKey("privacyexperience.id"), index=True)
+    # Minimal information stored here, mostly just region
+    privacy_experience_id = Column(
+        String
+    )  # TODO will be removed.  Preemptively removing the FK to Privacy Experience
```
pattisdr, line 222: This is used on both the Privacy Preference History and Served Notice History columns - it had a FK to privacy_experience. However, now adding and removing a region will add/remove the Privacy Experience.  We will no longer store this for preferences and notices served. The same details are on the Privacy Experience Config History Id and the user geography.

src/fides/api/schemas/privacy_notice.py
```diff
@@ -1,196 +1,27 @@
...
-
-
-class GPPFieldMapping(FidesSchema):
-    region: PrivacyNoticeRegion
-    notice: Optional[List[str]]
-    mechanism: Optional[List[GPPMechanismMapping]]
-
-
-class PrivacyNotice(FidesSchema):
+class PrivacyNoticeHistorySchema(FidesSchema):
```
pattisdr, line 9: All Privacy Notice schemas save this one were moved to Plus. Elements from this schema are used to build a consent email and are left here, including redefining some of the fields here.

src/fides/api/util/consent_util.py
```diff
@@ -284,355 +248,6 @@ def get_or_create_fides_user_device_id_provided_identity(
...
-        if privacy_notice.id not in ids:
-            ids.add(privacy_notice.id)
-        else:
-            raise HTTPException(
-                HTTP_422_UNPROCESSABLE_ENTITY,
-                detail=f"More than one provided PrivacyNotice with ID {privacy_notice.id}.",
-            )
-
-
-def create_privacy_notices_util(
```
pattisdr, line 321: A lot of the consent util methods related to creating out of the box configuration details have moved to plus

src/fides/common/api/v1/urn_registry.py
```diff
@@ -75,11 +66,6 @@
 POLICY_LIST = "/dsr/policy"
 POLICY_DETAIL = "/dsr/policy/{policy_key}"
 
-# Privacy Experience Config URLs
-EXPERIENCE_CONFIG = "/experience-config"
-EXPERIENCE_CONFIG_DETAIL = "/experience-config/{experience_config_id}"
-
-
```
pattisdr, line 82: These endpoints have moved to Plus

src/fides/common/api/v1/urn_registry.py
```diff
@@ -20,15 +20,6 @@
...
 CONSENT_REQUEST_VERIFY = "/consent-request/{consent_request_id}/verify"
 
-# Privacy Preference URLs
-CONSENT_REQUEST_NOTICES_SERVED = "/consent-request/{consent_request_id}/notices-served"
-NOTICES_SERVED = "/notices-served"
-
-# Reporting endpoints - have records for *all* users
-HISTORICAL_PRIVACY_PREFERENCES_REPORT = "/historical-privacy-preferences"
-CURRENT_PRIVACY_PREFERENCES_REPORT = "/current-privacy-preferences"
-
```
pattisdr, line 30: These endpoints are already in Plus, but we were importing their URN's from this file

pattisdr:
- Db scan is failing because I've also updated it with the expected state after the third migration.

src/fides/api/models/privacy_experience.py
```diff
@@ -219,260 +566,41 @@ class PrivacyExperience(Base):
...
-) -> Tuple[List[PrivacyNoticeRegion], List[PrivacyNoticeRegion]]:
+) -> List[PrivacyNoticeRegion]:
     """
-    Keeps Privacy Experiences in sync with ExperienceConfig changes.
-    Create or update PrivacyExperiences for the regions we're attempting to link to the
-    ExperienceConfig.
+    Links regions to a PrivacyExperienceConfig via the PrivacyExperience table.
 
-    Assumes that components on the ExperienceConfig do not change after they're updated.
+    PrivacyExperiences are queried directly for the end-user in a specific region and pull
```
adamsachs, line 601: nit: i may just make a note here in the docstring that this can potentially add or remove `PrivacyExperience` records, to make it clear what "linking" implies!

src/fides/api/models/privacy_notice.py
```diff
@@ -332,123 +365,249 @@ def create(
...
+                PrivacyNoticeHistory.create(
+                    db,
+                    data={
+                        **history_data,
+                        **updated_translation_data,
+                        "translation_id": translation.id,
+                        "version": existing_version + 1.0,
+                    },
+                    check_name=False,
+                )
```
adamsachs, line 449: again, i think there's probably some room for extracting these out into helpers, just to make the top-level method a bit cleaner. but not a major issue as-is!

src/fides/api/models/privacy_notice.py
```diff
@@ -332,123 +365,249 @@ def create(
...
-    for privacy_notice in existing_privacy_notices:
-        if privacy_notice.disabled and ignore_disabled:
-            continue
-        for region in privacy_notice.regions:
-            notice_keys_by_region[PrivacyNoticeRegion(region)].append(
-                (privacy_notice.notice_key, privacy_notice.name)
+        # remove protected fields from the cloned dict
+        cloned_attributes.pop("_sa_instance_state", None)
+        cloned_attributes.pop("id", None)
+        cloned_attributes.pop("translations", [])
```
adamsachs, line 482: _maybe_ some opportunity for extracting into a helper/utility that's shared with the `PrivacyExperienceConfig` dry updates? may not be worth it though!

adamsachs (commented):
@pattisdr i think i've just about gotten through these (mainly model-level) changes here in OSS!

on the whole, i don't see any major issues, this looks really good! i pointed out some notes/comments, but most are minor.

still haven't gotten to the API-level changes over in plus, but it was very helpful to start here and get a good baseline understanding of the model updates -- the mental model is a lot clearer for me now!

thanks for the thorough self-documentation in code comments/docstrings/PR comments, and for writing some well-organized and clean code. for all the substantive changes, this was relatively easy to make sense of! 🚀 

now on to the API updates in plus 😅

src/fides/api/models/privacy_experience.py
```diff
@@ -97,38 +255,131 @@ def create(
...
+        - Link or unlink regions (via Privacy Experiences)
+        - Link or unlink Privacy Notices
         """
-        resource, updated = update_if_modified(self, db=db, data=data)
+        request_translations = data.pop("translations", [])
+        regions = data.pop("regions", [])
+        privacy_notice_ids = data.pop("privacy_notice_ids", [])
+
+        # Do a patch update of the existing privacy experience config if applicable
+        config_updated = update_if_modified(self, db=db, data=data)
```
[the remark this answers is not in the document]
pattisdr: yep, I can move this models.__init__.py! I'll just add code comments too explaining what models they are currently used on.

I had kept it here because it's related to creating historical records in a specific pattern just used by notices and experiences, and experience code imports notice code in other ways too. But I also understand how by themselves they read fairly generic and don't want to cause confusion thinking they are solely for notices.

src/fides/api/models/privacy_experience.py
```diff
@@ -97,38 +255,131 @@ def create(
...
+        ).difference(
+            set(translation.get("language") for translation in request_translations)
+        )
 
-        return resource  # type: ignore[return-value]
+        db.query(ExperienceTranslation).filter(
+            ExperienceTranslation.language.in_(translations_to_remove),
+            ExperienceTranslation.experience_config_id == self.id,
+        ).delete()
+        db.commit()
```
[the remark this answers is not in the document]
pattisdr: Yep I'll do this!  This method is getting long and hard to parse through.  I think your compromise of still leaving the translation creation here (which I think is a core part of this method, an ExperienceConfig needs an ExperienceTranslation to be valid, it's not useful by itself) and separating out the historical record and translation deletion makes sense.  And having good method names here will make it easier to see top-level what is happening 👍

src/fides/api/models/privacy_experience.py
```diff
@@ -141,25 +324,37 @@ def dry_update(self, *, data: dict[str, Any]) -> PrivacyExperienceConfig:
...
+        dry_updated_translations = []
+        for translation_data in data:
+            existing_translation = self.get_translation_by_language(
+                Session.object_session(self), translation_data.get("language")
+            )
+            if existing_translation:
+                cloned_attributes = existing_translation.__dict__.copy()
+                for key, val in translation_data.items():
+                    cloned_attributes[key] = val
+                cloned_attributes.pop("_sa_instance_state")
```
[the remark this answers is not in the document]
pattisdr: Yes I think that's a good idea, a method that clones an object but removes the important fields like the instance state.  I'll do the layering of any updates outside of that method.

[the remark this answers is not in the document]
pattisdr: To be clear, I think `cloned_attributes.pop("_sa_instance_state")` by itself is too small of a utility function, if that's what you meant, this is a common thing I see in codebases working in SQLAlchemy. But an overall cloning utility I think would be very useful.

src/fides/api/models/privacy_notice.py
```diff
@@ -332,29 +351,158 @@ def create(
...
+        This is used to see what an object update would look like, in memory,
+        without actually persisting the update to the db
+        """
+        # Update our attributes with values in data
+        cloned_attributes = self.__dict__.copy()
+        for key, val in data.items():
+            cloned_attributes[key] = val
+
+        # remove protected fields from the cloned dict
+        cloned_attributes.pop("_sa_instance_state", None)
```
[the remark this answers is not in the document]
pattisdr: I have two new methods now, a `clone_object_attributes` and a `dry_update_data` -

src/fides/api/models/privacy_experience.py
```diff
@@ -97,38 +255,131 @@ def create(
...
+                    db,
+                    data={**translation_data, "experience_config_id": self.id},
+                )
 
-        if updated:
-            history_data = create_historical_data_from_record(resource)
-            history_data["experience_config_id"] = self.id
+            if config_updated or translation_updated:
+                existing_version: float = translation.version or 0.0
+                history_data: dict = create_historical_data_from_record(self)
```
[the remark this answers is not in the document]
pattisdr: Correct! When speccing out original requirements, I wondered if we both wanted a historical record of both "privacy notices" and "regions" available.  

But I [checked with product](https://ethyca.atlassian.net/wiki/spaces/PM/pages/2905997363/Multi-language+and+experience+config+implementation+questions+answers#My-first-batch-of-backend-questions-for-consent-multitranslation%3A) and  it sounded like it is fine to leave this off, because it doesn't add additional value when preferences are saved, because we have the "region" the user responded from and the "privacy notice" against which the user is consenting.

src/fides/api/models/privacy_experience.py
```diff
@@ -141,67 +392,163 @@ def dry_update(self, *, data: dict[str, Any]) -> PrivacyExperienceConfig:
...
     """
 
-    component = Column(EnumColumn(ComponentType), nullable=False)
     region = Column(EnumColumn(PrivacyNoticeRegion), nullable=False, index=True)
 
     experience_config_id = Column(
         String,
         ForeignKey(PrivacyExperienceConfig.id_field_path),
-        nullable=True,
+        nullable=True,  # Needs an ExperienceConfig to be valid though
```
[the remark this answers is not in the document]
pattisdr: This was for purely historical reasons for data potentially in a user database. Depending on how your data migration settles out, we could definitely make this non-nullable! I know your current draft wipes PrivacyExperiences altogether, but I think some of our details are still TBD.

 Previously, Privacy Experiences were never deleted, there was exactly one artifact that could be linked and unlinked from various Privacy Experience Configs. It was definitely a _possible_ state that you had a Privacy Experience not linked to a Privacy Experience Config.

I mitigate with existing data by just requiring that a PrivacyExperience has a valid ExperienceConfig attached.

src/fides/api/models/privacy_notice.py
pattisdr: Two methods added like you requested above, one for the historical record, one for deleting the notice translations.  They could be further consolidated with the experience-level methods but I am thinking of keeping them separate mainly for type hinting purposes.

src/fides/api/models/privacy_notice.py
pattisdr: Yes I've been able to share my new `clone_object_attributes` and `dry_update_data` methods in this file 👍

src/fides/api/models/privacy_notice.py
pattisdr: I've tried to mitigate by only exposing this property on the Privacy Notice List and Detail endpoints used in the Admin UI, not for the end user, and also further pulled a lot of the details previously exposed in Privacy Notice API endpoint to help that load.

I agree there's always tradeoffs here.  Model level properties that you then just expose in the API are convenient and readable, but yes defined like this, certainly less performant than it could be.  I'll make a note to revisit this with the locations PR.

pattisdr:
This is ready for another pass Adam! I can tell you took the time to understand what was going on here, thanks for the great comments 👍

src/fides/api/models/privacy_experience.py
```diff
@@ -141,67 +392,163 @@ def dry_update(self, *, data: dict[str, Any]) -> PrivacyExperienceConfig:
...
+        String, ForeignKey(PrivacyExperienceConfig.id_field_path), nullable=False
+    )
+
+    __table_args__ = (
+        UniqueConstraint(
+            "language", "experience_config_id", name="experience_translation"
+        ),
+    )
+
+    histories: RelationshipProperty[AppenderQuery] = relationship(
```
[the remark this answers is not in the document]
pattisdr: OK Michael has said dismissable at least is worth recording for audit purposes?  Is that one field enough to do it this way? I'm also versioning more fields than that as well.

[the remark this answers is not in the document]
adamsachs: i think so! thanks for checking on this, makes sense.

also, it feels easier to walk this back in the direction of making _fewer_ history objects in the future if we want to, whereas the reverse isn't really true :)

NevilleS (approved):
OK - I updated the CHANGELOG here and have a few comments/questions... but this looks ready to merge to `main` so we can focus on the final round of testing and cleanup prior to release 🎉

CHANGELOG.md
```diff
@@ -18,11 +18,19 @@ The types of changes are:
 ## [Unreleased](https://github.com/ethyca/fides/compare/2.31.1...main)
 
 ### Added
+- Ability to link Properties to Privacy Experiences [#4658](https://github.com/ethyca/fides/pull/4658)
+- Added property_id query param to fides.js to filter for specific experiences when prefetch is enabled [#4676](https://github.com/ethyca/fides/pull/4676)
 - Erasure support for Simon Data [#4552](https://github.com/ethyca/fides/pull/4552)
```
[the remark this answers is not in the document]
NevilleS: Done!

clients/admin-ui/cypress/e2e/domains.cy.ts
```diff
@@ -76,10 +75,6 @@ describe("Domains page", () => {
             "have.value",
             "http://localhost"
           );
-          cy.getByTestId("input-config_cors_origin_regex").should(
-            "have.value",
-            "https://.*\\.example\\.com"
-          );
```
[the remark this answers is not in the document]
NevilleS: Moved to our release doc

clients/admin-ui/src/features/privacy-experience/preview/helpers.ts
```diff
@@ -0,0 +1,109 @@
...
+  options: {
+    debug: true,
+    geolocationApiUrl: "",
+    isGeolocationEnabled: false,
+    isOverlayEnabled: true,
+    isPrefetchEnabled: false,
+    overlayParentId: "preview-container",
+    modalLinkId: null,
+    privacyCenterUrl: "http://localhost:3000",
+    fidesApiUrl: "http://localhost:8080/api/v1",
```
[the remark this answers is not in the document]
eastandwestwind: `fidesApiUrl` is required, but looks like `privacyCenterUrl` is not. I'll open a PR to clean these up, as well as removing debug logging from preview mode if we wish?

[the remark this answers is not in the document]
eastandwestwind: https://ethyca.atlassian.net/browse/PROD-1828

PR #4578, opened 2024-01-30
Preventing completion email from being sent for consent requests

galvana:
Closes [PROD-1634](https://ethyca.atlassian.net/browse/PROD-1634)

### Description Of Changes

Prevention privacy request completion emails from being sent for consent requests.

### Code Changes

* [ ] Adding a check for consent requests in `request_runner_service.run_privacy_request` before sending privacy request completion emails

### Steps to Confirm

* [ ] set the following values in `fides/.env`
```
FIDES__NOTIFICATIONS__NOTIFICATION_SERVICE_TYPE=mailgun
FIDES__NOTIFICATIONS__SEND_REQUEST_COMPLETION_NOTIFICATION=true
```
* [ ] Set `"executable": true` for the `marketing.advertising` `consentOption` in _src/fides/data/sample_project/privacy_center/config/config.json_
* [ ] Create a messaging config via the API (Postman)
  * [ ] Post Messaging Config
  * [ ] Messaging Config Secrets
* [ ] Submit a consent request
* [ ] Verify the request in the Admin UI, it should go to complete with no errors

### Pre-Merge Checklist

* [ ] Update `CHANGELOG.md`

[PROD-1634]: https://ethyca.atlassian.net/browse/PROD-1634?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/fides/api/service/privacy_request/request_runner_service.py
+++ b/src/fides/api/service/privacy_request/request_runner_service.py
@@ -490,7 +490,11 @@ async def run_privacy_request(
             if not proceed:
                 return
 
-        if ConfigProxy(session).notifications.send_request_completion_notification:
+        if ConfigProxy(
+            session
+        ).notifications.send_request_completion_notification and not policy.get_rules_for_action(
+            action_type=ActionType.consent
+        ):
             try:
                 initiate_privacy_request_completion_email(
                     session, policy, access_result_urls, identity_data
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ The types of changes are:
 ### Fixed
 
 - Fixed CORS domain update functionality [#4570](https://github.com/ethyca/fides/pull/4570)
+- Completion emails are no longer attempted for consent requests [#4578](https://github.com/ethyca/fides/pull/4578)
 
 ## [2.28.0](https://github.com/ethyca/fides/compare/2.27.0...2.28.0)
 
--- a/tests/ops/service/privacy_request/test_request_runner_service.py
+++ b/tests/ops/service/privacy_request/test_request_runner_service.py
@@ -13,7 +13,6 @@ from pydantic import ValidationError
 from sqlalchemy import column, select, table
 from sqlalchemy.orm import Session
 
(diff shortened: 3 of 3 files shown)


src/fides/api/service/privacy_request/request_runner_service.py
```diff
@@ -490,7 +490,11 @@ async def run_privacy_request(
             if not proceed:
                 return
 
-        if ConfigProxy(session).notifications.send_request_completion_notification:
+        if ConfigProxy(
+            session
+        ).notifications.send_request_completion_notification and not policy.get_rules_for_action(
+            action_type=ActionType.consent
```
adamsachs, line 496: ok, this took a bit of a refresher for me, but i think i'm remembering now - we basically infer an action type for a given `PrivacyRequest` based on the `Policy`, which isn't _required_ to have only rules of a single action type, but practically we expect any `Policy` to have rules with different action types - right?

if that's the case, then this makes sense. if it's not the case, we'd have an edge case here where a hypothetical policy that does access + consent _wouldn't_ trigger a completion email even though it probably should. but that's just totally unrealistic, it's not like we'd actually have a policy with rules that have both of those action types, right!?

anyway, i think this is all good (if my assumptions are correct) - i just wonder when it'll be time for us to update some of our constraints/assumptions in our policy/rule/general DSR framework to be a bit more explicit and consistent about this expectation. i don't think there's a lot of functional risk at the moment, just makes this a bit harder to keep track of when reviewing code 😅 

curious to get your thoughts, this was just a little bit of a rabbit hole i went down before feeling totally good about approving 👍

galvana: I hesitated a bit when I wrote this too but it was more around where this check should live. I thought about moving the check to `initiate_privacy_request_completion_email` and just skip the identity check for consent policies but that still takes us back to your concern of how to reliably determine the action type. As we have it now we have one policy per action type so we should be good for now.

PR #4581, opened 2024-01-30
move pymssql back to an optional requirement

adamsachs:
Closes [PROD-1558](https://ethyca.atlassian.net/browse/PROD-1558)

### Description Of Changes

Pull out the `pymssql` driver dependency into an optional requirements file that is _not_ installed as part of our python package install by default. This is because the driver does not include functionality that's essential for our python package functionality, and it's known to cause problems and require workarounds on M1 macs. The optional dependency can still be installed as part of the python package by including an `[all]` extras keyword in the package dependency reference, e.g. `ethyca-fides[all]==2.28.1a3`.

The pymssql dependency (and any others we decide to put in `optional-requirements.txt`) _will_ still be installed as part of all of our docker image builds. It functions fine on linux containers, and it's important functionality for our deployed application (which is a different use case than our python package!)

### Code Changes

* [x] make an `optional-requirements.txt` to separate out optional requirements not needed for basic python package/CLI usage
* [x] ensure the `optional-requirements` are installed on our docker image builds via the `Dockerfile`, since they're needed for our images and there are no problem with these libraries on linux containers
* [x] allow optional requirements to be specified as an "extra" with the python package install, either via the requirement name `mssql` or via an `all` keyword, e.g. `pip install ethyca-fides[all]==2.28.1a3`

### Steps to Confirm

* [x] confirmed locally that i could install the `ethyca-fides` python package (and therefore run `nox -s "fides_env(test)"` even with `freetds` uninstalled and no cached pymssql packages, i.e. our python package install and the `fides_env` command do _not_ require installing pymssql nor building its wheel on the host machine, while the package is installed and wheel is built on the docker image that's built by `fides deploy up`. the following steps used to produce a build error on an M1 Mac locally, but now no longer do with this update: 
  * [x] uninstall `freetds` locally: `brew uninstall --ignore-dependencies freetds`
  * [x] remove cached wheels from global pip cache: `find ~/Library/Caches/pip/wheels/ | grep pymssql` to find and then remove
  * [x] remove pymssql packages from the repo's local nox caches (`./.nox`), e.g. from the repo root:`rm -rf .nox/fides_env-slim/lib/python3.10/site-packages/pymssql*`
  * [x] test out running `nox -s "fides_env(test)"`. on `main`, this should fail on an M1 mac when it tries to build the `pymssql` wheel. but on this branch, things should succeed 
    
* [x] mssql integration tests still seem to be running well in CI, since that's using the built docker image that has the package installed 👍 https://github.com/ethyca/fides/actions/runs/7727125996/job/21065508406?pr=4581

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

[PROD-1558]: https://ethyca.atlassian.net/browse/PROD-1558?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


README.md
```diff
@@ -27,7 +27,9 @@ In order to get started quickly with Fides, a sample project is bundled within t
 * [Docker](https://www.docker.com/products/docker-desktop) (version 20.10.11 or later)
 * [Python](https://www.python.org/downloads/) (version 3.8 through 3.10)
 
-#### Additional requirements (for ARM Mac Users)
+#### Additional requirements for optional dependencies (for ARM Mac Users)
```
adamsachs, line 30: i considered just removing this section altogether, but then we'd have no trace about what you do need to do if you have this targeted use case. i could still easily be convinced to just remove this though, it feels a bit wordy and confusing as i have it!

optional-requirements.txt
adamsachs, line 1: could also call this `dangerous-requirements` as we had it for a while, or even `driver-requirements` to be very specific 🤷

NevilleS: Optional is OK 😄

README.md
adamsachs: muchhhh better, thank you 🙏

PR #4587, opened 2024-02-01
Improve Snowflake Generate Performance

SteveDMurphy:
Closes PROD-1639

### Description Of Changes

The Snowflake connector has some known out of the box issues, and the suggested workaround has some performance hit. This PR introduces the ability to surface the schema in parallel to reduce the time required. No data or funactionality change, just the option to surface the schema in multiple threads.

This was found as a problem for larger datasets that caused us to hit the FastAPI timeout (60s by default)

### Code Changes

* [x] Include `joblib`
* [x] separate column discovery to new function
* [x] run the discovery in parallel
* [x] use a scale for the number of threads to be used
* [x] ensure test coverage

### Steps to Confirm

* [x] Tested manually with our POCDB in Snowflake, where it took the time down from ~8 seconds to under 3 when using 4 threads

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`


SteveDMurphy:
I think I may need to add some test coverage but was hoping to get an early review to try and get this on an alpha release if one of you has any time @pattisdr or @adamsachs 

It adds a new dependency (`joblib`) but one we already have on `Fidesplus` so I don't see it as a huge trade (but could be wrong!)

src/fides/core/dataset.py
```diff
@@ -400,6 +401,10 @@ def get_snowflake_schemas(
     and being able to fall back to using the connector.
 
     Reference: https://github.com/snowflakedb/snowflake-sqlalchemy/issues/157
+
+    Due to performance issues resulting in FastAPI timeouts, parallelization is
+    no used. The default number of threads to be used is 4 to cover most existing
```
adamsachs, line 406: nit: is this supposed to be `not` or `now`? i think it's now - since it looks like we've switched this to use parallelization1 

```suggestion
    now used. The default number of threads to be used is 4 to cover most existing
```

src/fides/core/dataset.py
```diff
@@ -409,10 +414,23 @@ def get_snowflake_schemas(
             metadata[schema] = {}
             table_cursor = engine.execute(text(f'SHOW TABLES IN "{schema}"'))
             db_tables = [row[1] for row in table_cursor]
-            for table in db_tables:
-                column_cursor = engine.execute(
-                    text(f'SHOW COLUMNS IN "{schema}"."{table}"')
-                )
-                columns = [row[2] for row in column_cursor]
+            number_of_threads = 8 if len(db_tables) > 250 else 4
```
adamsachs, line 417: i think it'd be really nice to make this config-driven in some way, if that's not too crazy of a thing to do. it feels like something we'd potentially want to fine tune "on the fly" - either for pure testing or just depending on various different environmental factors. could end up saving a lot of iteration overhead, where we could simply tweak in a deployed env, rather than having to cut new releases to iterate on the values we want here.

curious to hear what you think though - maybe this would less helpful than i'm imagining!

src/fides/core/dataset.py
```diff
@@ -409,10 +414,23 @@ def get_snowflake_schemas(
...
+            fields = Parallel(n_jobs=number_of_threads, backend="threading")(
```
adamsachs, line 418: not trying to make things more difficult, but could you use the builtin [`multiprocessing`](https://docs.python.org/3/library/multiprocessing.html) package to get the same functionality here? (i think just a `Pool.map` may be able to get you want you need?)

granted - i'm not a python parallelization expert, and i don't know a whole lot about `joblib`. it also seems pretty lightweight, so i don't think it's necessarily a problem. but i do generally think we should have compelling reasons to stray away from the standard library packages -- maybe there are some here, and i just don't know 'em!

adamsachs (commented):
nice, lookin really solid @SteveDMurphy! a couple of comments that you can feel free to take or leave, but seems like a great improvement that looks straightforward enough.

i agree that some test coverage here would be very nice, though it may be a bit hard to get the parallelization functionality itself covered in automated tests. do we at least have the overall functionality of this codepath covered in existing tests, i.e. enough to ensure that this change doesn't cause regressions?

src/fides/core/dataset.py
SteveDMurphy: definitely now! thanks 👍🏽

SteveDMurphy:
> i agree that some test coverage here would be very nice, though it may be a bit hard to get the parallelization functionality itself covered in automated tests. do we at least have the overall functionality of this codepath covered in existing tests, i.e. enough to ensure that this change doesn't cause regressions?

I believe we do, in that the same output should match the results after this change - in the end, I _think_ the only functional change is taking the parallelized output and outputting it correctly so I was going to look at testing that distinctly but it does feel covered... The current testing path should be redirected to hit both of these steps with Snowflake and we should match the same output (it's how I tested it manually as well!)

src/fides/core/dataset.py
SteveDMurphy: It felt like a bit of a pickle, I thought about it and strayed away a bit due to wanting to manage it without adding further config variables. I didn't like the idea of having to restart/re-deploy to generate a database (likely just one time and only for Snowflake). If these values didn't work, then the next step may be to up the timeout of the API or Generate the values in a separate thread entirely (similar to how we Classify today).

These aren't terribly heavy operations really as much as tedious so I did feel hesitant to add much extra, especially as it feels like there may be something better overall in a minor rewrite or other change specific to handling Snowflake

src/fides/core/dataset.py
SteveDMurphy: to be honest, I knew we used this as a fairly simple way to improve performance in Fidesplus today with Classify so considered it relatively low-risk here as well even though it is introducing something "new" (especially taking the required turn-time component into account)

SteveDMurphy: Will take a look now to trial `multiprocessing`! Would be nice to not have to deal with a new addition if we can

src/fides/core/dataset.py
adamsachs: got it, that seems like a totally fair trade off if you don't imagine needing to fine tune the number of threads!

src/fides/core/dataset.py
adamsachs: ah ok totally fine if this is tried and true - doesn't seem like a heavyweight dep and yeah it's already in the plus image.

it's always nice to try and be as slim as possible, especially here in OSS where this gets packaged into the python package. but it seems like minimal impact in this case and not worth a big struggle to switch over!

SteveDMurphy: absolutely, would love to get rid of the requirement in both if possible!

adamsachs (approved):
some potential tweaks/improvements to look into around the edges, but overall this looks like a solid incremental update and i think it's ready for a alpha/beta deployment 👍

PR #4592, opened 2024-02-05
add supported language model + enum + registry YAML

adamsachs:
### Description Of Changes

Defines the `SupportedLanguage` enum (and very-lightweight `Language` pydantic model) for use within the multilanguage epic.

Will be leveraged first by a `GET /languages` endpoint in plus, but also meant to be used more broadly by other notice/experience schemas and models.

### Code Changes

* [x] YAML file enumerating our supported languages (`id` + `name` for each)
* [x] basic API model for a language - just id and name fields, for now
* [x]  dynamically-defined Enum to allow for pydantic model-validation against the supported languages

### Steps to Confirm

* [x] will manually test integration with endpoint functions over in the associated fidesplus PR
* [x] automated test coverage here proves out that the enum is working as we would hope/expect (i think!) 👍 

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### DIFF
--- a/src/fides/api/schemas/language.py
+++ b/src/fides/api/schemas/language.py
@@ -1,0 +1,46 @@
+from enum import Enum
+from os.path import dirname, join
+from typing import Dict
+
+import yaml
+from pydantic import BaseModel
+
+from fides.api.util.text import to_snake_case
+from fides.config.helpers import load_file
+
+LANGUAGES_YAML_FILE_PATH = join(
+    dirname(__file__),
+    "../../data",
+    "language",
+    "languages.yml",
+)
+
+
+class Language(BaseModel):
+    id: str
+    name: str
+
+
+def _load_supported_languages() -> Dict[str, Language]:
+    """Loads language dict based on yml file on disk"""
+    with open(load_file([LANGUAGES_YAML_FILE_PATH]), "r", encoding="utf-8") as file:
+        _languages = yaml.safe_load(file).get("languages", [])
+        language_dict = {}
+        for language in _languages:
+            language_dict[language["id"]] = Language.parse_obj(language)
+        return language_dict
+
+
+supported_languages_by_id: Dict[str, Language] = (
+    _load_supported_languages()
+)  # should only be accessed for read-only access
+
+
+# dynamically create an enum based on definitions loaded from YAML
+SupportedLanguage = Enum(  # type: ignore[misc]
+    "SupportedLanguage",
+    {
+        to_snake_case(language.name): language.id
+        for language in supported_languages_by_id.values()
+    },
+)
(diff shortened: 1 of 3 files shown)


adamsachs:
@pattisdr let me know what you think when you have a chance - the big changes here are the dynamic enum creation (i _think_ this is what you were getting at), and then just moving this over to OSS generally of course.

i'll need to adjust the plus PR accordingly, but i _think_ the automated tests here prove out enough to allow us to merge this into `main` here and then i'll pull it in on the plus side. alternatively i can just start pointing at an alpha tag/commit in plus to do some integration testing before merging, doesn't matter too much to me!

src/fides/api/schemas/language.py
```diff
@@ -0,0 +1,46 @@
...
+supported_languages_by_id: Dict[str, Language] = (
+    _load_supported_languages()
+)  # should only be accessed for read-only access
+
+
+# dynamically create an enum based on definitions loaded from YAML
+SupportedLanguage = Enum(  # type: ignore[misc]
+    "SupportedLanguage",
+    {
+        to_snake_case(language.name): language.id
```
pattisdr, line 43: What language format will we add to the translations then, "portuguese_brazil"?

adamsachs: hmm no the way i imagine this working is that the _value_ of the enum (i.e. the Language `id`) is what would actually be stored on the translation resource -- so it'd be `pt-BR`. the `test_language_enum_in_schema` shows an example.

does this make sense to you? i may be doing something different than what you were imagining!

pattisdr: it can just be a little painful when the keys and values don't match, transforming back and forth from schemas to databases to dictionaries etc, we'll just need to track down all the places that need attending to!

```python
LookupError: 'sr-Cyrl' is not among the defined enum values. Enum name: language. Possible values: en_us, en_gb, portuguese_.., serbi...
```

```python
E           pydantic.error_wrappers.ValidationError: 1 validation error for ExperienceConfigCreateTemplate
E           translations -> 0 -> language
E             value is not a valid enumeration member; permitted: 'en_us', 'en_gb', 'pt-BR', 'sr-Cyrl' (type=type_error.enum; enum_values=[<Language.en_us: 'en_us'>, <Language.en_gb: 'en_gb'>, <Language.portuguese_brazil: 'pt-BR'>, <Language.serbian_cyrillic: 'sr-Cyrl'>])
```

pattisdr (approved):

src/fides/api/schemas/language.py
adamsachs: got it, agreed. we discussed more offline - we _may_ look to eventually align keys and values to match, but we'll merge as-is for now, and @pattisdr will give a try integrating with her ongoing work to see how painful it is. so we may revisit this!

PR #4593, opened 2024-02-05
Ignore 404 - Kustomer and Delighted

SteveDMurphy:
Closes PROD-1646

### Description Of Changes

An edge case found when an erasure for the same user is approved simultaneously, allowing the access portion to succeed (user found) but only one erasure to succeed while the other results in a 404.

As the desired result is to have the user no longer found, ignoring a 404 in this situation feels like the right course of action.

### Code Changes

* [x] Ignore 404 on erasure for both Delighted and Kustomer
* [x] update the Kustomer access ignore rules to be specific to a 404

### Steps to Confirm

* [ ] n/a

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`


galvana (changes requested):
Both of these looks good, just make sure to bump the version number for Delighted and Kustomer so the auto-update takes effect.

SteveDMurphy:
> Both of these looks good, just make sure to bump the version number for Delighted and Kustomer so the auto-update takes effect.

I don't know how this got so lost, but circling back finally @galvana 😅

galvana (approved):

PR #4594, opened 2024-02-06
Structured logging for DSR requests

galvana:
Closes [PROD-1602](https://ethyca.atlassian.net/browse/PROD-1602)

### Description Of Changes

Uses a combination of a `Contextualizable` mixin and a `log_context` decorator. The decorator iterates through the arguments of the decorated function and calls `get_log_context` on every argument that implements `Contextualizable`. The log_context decorator accepts optional keyword arguments to add additional context outside of the contextualizable arguments.

Adding four user-friendly error groups for SaaS connector logs.

- Authentication error
- Client-side error
- Server-side error
- Network/protocol

All HTTP request logs for SaaS connectors will also include the following metadata

- system_key
- connection_key
- action_type
- collection_name
- privacy_request_id
- method
- url
- body (if dev mode is enabled)
- response (if dev mode is enabled)

This logging will apply to access, erasure, and consent requests. We can expand this contextualized logging to database connectors or other parts of the app after we receive initial feedback.

### Code Changes

* [ ] A new `logger_context_utils.py` with a `Contextualizable` base class and `log_context` decorator
* [ ] Addition of `logger.contextualize` in key parts of the access, erasure, and consent request functions

### Steps to Confirm

* [ ] Run the `test_privacy_request_logging.py` test or any of the SaaS connector tests

### Pre-Merge Checklist

* [ ] Update `CHANGELOG.md`

[PROD-1602]: https://ethyca.atlassian.net/browse/PROD-1602?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/fides/api/util/logger_context_utils.py
```diff
@@ -0,0 +1,93 @@
...
+                details["error_group"] = ErrorGroup.client_error.value
+            elif 500 <= response.status_code:
+                details["error_group"] = ErrorGroup.server_error.value
+    return details
+
+
+def exception_details(exception: Exception, url: str) -> Dict[str, Any]:
+    """Maps select connection exceptions to user-friendly error details."""
+
+    details = {"error_group": ErrorGroup.network_error.value}
```
adamsachs, line 80: this confuses me a bit, when considered alongside the function signature. the function seems to apply to any `Exception` generally, but here we effectively assume that it's a `network_error`. that may be the only context in which this helper function is called _currently_, but seems like something that one could easily be tripped up on when trying to extend these utilities to other use cases.

so, either i'm misunderstanding something, or i'd recommend adjusting the function name or logic (or at least docstring)

src/fides/api/util/logger_context_utils.py
```diff
@@ -0,0 +1,93 @@
...
+            connector.configuration.system.id
+            if connector.configuration.system
+            else None
+        ),
+        "connection_key": connector.configuration.key,
+    }
+    if connector.current_collection_name:
+        details["collection"] = connector.current_collection_name
+    if connector.current_privacy_request:
+        details["privacy_request_id"] = connector.current_privacy_request.id
```
adamsachs, line 42: nit: looking ahead to a _broader_ logging capability, can you think of any (simple) ways to make more generic/extensible the pieces that add the generic privacy request details to the log context/dictionary?

i.e., we'll probably want the `privacy_request_id` put on the logger context even when we don't have a `SaaSconnector` available to us (on non-SaaS DSRs). it'd be nice to ensure consistency/DRY code for adding that piece of context in all use cases.

all that being said, that may be an over-abstraction for now, if we're just looking to get the SaaS connector support, so totally fine deferring on this till later. just wanted to mention and keep eyes on what's ahead!

adamsachs: one very rudimentary thing i can think of, at least for now, is just defining the `privacy_request_id` key as a module-level constant rather than a string defined inline. that's at least one measure we can take to help ensure consistency across contexts. we also could do the same for other keys that we anticipate will be reused across different context. 👍 


the reason i _key_ (pun intended) in on this is because consistency across logs is super important for being able to sift through the noise! so the closer we can get to a formal "schema" for the context we're adding (across the entire app), the better!

src/fides/api/util/logger_context_utils.py
adamsachs, line 1: nice to pull these utils out into their own module 👍 

could we go a step further and abstract out the (small amount of) boiler plate that's still required to get the dict that's output by these functions into the logger context? maybe not, but something to consider. basically - how can we make it as simple as possible for someone who wants to come in and add a similar type of logging context from a totally different point in the app, with as consistent a schema as possible to what we've started here. having the functions defined in a single place here helps with that goal, for sure - but i wonder if there's anything else we can do to help guide consistency.

take that with a grain of salt though - i'm aware that we can't over-optimize, so it may be nothing to act on for now!

adamsachs (commented):
@galvana i haven't done UAT testing yet since this is still in draft, but i think the approach here looks great! a couple minor comments for now but, assuming you've confirmed this is giving us good output that's in line with the expectations in the ticket, i  think this is a good path to continue down and i'd be happy to do some proper UAT testing if/when you feel that's ready.

have we confirmed SaaS request execution logging as a sufficient scope for this increment? if so, then this looks pretty much feature-complete, i think. that being said, my comments are mainly around future-proofing and how to ensure consistency as we'll look to build on this. it may not be time to act on those concerns quite yet, but at least wanted to share some of my thoughts as i was going through!

src/fides/api/util/logger_context_utils.py
galvana: Here's an alternate approach https://github.com/ethyca/fides/pull/4609, let me know your thoughts!

tests/ops/integration_tests/test_privacy_request_logging.py
```diff
@@ -0,0 +1,162 @@
...
+        api_client,
+        url,
+        generate_auth_header,
+        consent_policy,
+        provided_identity_and_consent_request,
+        loguru_caplog,
+        provided_identity_value,
+    ):
+        with pytest.raises(ClientUnsuccessfulException):
+            await klaviyo_runner.new_consent_request(
```
galvana, line 140: I need to update this test to be triggered by API calls, the call to `new_consent_request` is just a placeholder for now.

adamsachs (commented):
latest changes look good to me, and i really like the way the changes in https://github.com/ethyca/fides/pull/4609 are looking - that approach feels very extensible while remaining relatively simple!

i think you'll wanna merge that into this PR and finalize things, right? so i'll hold off on a final approval until then, just to see how things shape up here. but everything else looks good! nice work on all of this 👍

adamsachs (approved):
@galvana awesome work here, i really like how this has turned out! a really elegant paradigm you've set up.

i've got one nit that you _could_ look address, but probably not necessary. i haven't done any testing yet, but i'll approve preemptively anyway, and i'll see if i can follow up with some manual testing.

PR #4603, opened 2024-02-08
Add support for STS-issued AWS Credentials

### LINKED ISSUE #4596: Add support for STS-issued AWS credentials

When using either the CLI or the Fides web interface, the AWS resource scanner requests an access key ID and secret. This means that users can only use long-term credentials which goes against [AWS best practices](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_access-keys.html#securing_access-keys). Best practices indicate to use [temporary security credentials](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_use-resources.html)

Adding an additional CLI argument (e.g. `--session_token`) that allows for just-in-time creation and use of temporary keys created by STS.

Alternative would be provisioning a long-term access key.

tariqajyusuf:
Closes #4596

### Description Of Changes

This change adds an option to pass `session_token` to the CLI, UI, or API when using the system discovery feature for AWS. Long-term credentials only require an `access_key_id` and `secret_access_key`. As it stands today, this means that long-term credentials have to be created to use any system discovery. By adding an option to use `session_token`s, we now allow for temporary credentials that can reflect the principle of least access.

### Code Changes

* [ ] Update Python and React data models to include `aws_session_token`
* [ ] Add `--session_token` as an optional argument to `fides generate` and `fides scan`
* [ ] Add optional session token field to AWS discovery.

### Steps to Confirm

After setting up the dev environment

* [ ] Provision temporary credentials (e.g. `aws sts assume-role ...`)
* [ ] Attempt to use just the `access_key_id` and `secret_access_key`, this should fail.
* [ ] Now try to include the `session_token` argument, this should pass.

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### DIFF
--- a/clients/admin-ui/src/features/config-wizard/AuthenticateAwsForm.tsx
+++ b/clients/admin-ui/src/features/config-wizard/AuthenticateAwsForm.tsx
@@ -149,15 +149,17 @@ const AuthenticateAwsForm = () => {
                     label="Secret"
                     // "You can find more about creating access keys and secrets on AWS docs here."
                     tooltip="The secret associated with the Access Key ID used for authentication."
+                    isRequired
                   />
                   <CustomTextInput
                     type="password"
                     name="aws_session_token"
-                    label="Session Token (if applicable)"
+                    label="Session Token"
                     // "You can find more about creating access keys and secrets on AWS docs here."
-                    tooltip="The session token for temporary credentials."
+                    tooltip="The session token when using temporary credentials."
                   />
                   <CustomSelect
+                    isRequired
                     name="region_name"
                     label="AWS Region"
                     // "You can learn more about regions in AWS docs here."
--- a/clients/admin-ui/src/features/config-wizard/AuthenticateAwsForm.tsx
+++ b/clients/admin-ui/src/features/config-wizard/AuthenticateAwsForm.tsx
@@ -33,6 +33,7 @@ import ScannerLoading from "./ScannerLoading";
 const initialValues = {
   aws_access_key_id: "",
   aws_secret_access_key: "",
+  aws_session_token: "",
   region_name: "",
 };
 
@@ -49,6 +50,11 @@ const ValidationSchema = Yup.object().shape({
     .trim()
     .matches(/^[^\s]+$/, "Cannot contain spaces")
     .label("Secret"),
+  aws_session_token: Yup.string()
+    .optional()
+    .trim()
+    .matches(/^[^\s]+$/, "Cannot contain spaces")
+    .label("Session Token (for temporary credentials)"),
   region_name: Yup.string().required().label("Default Region"),
 });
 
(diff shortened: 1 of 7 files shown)


clients/admin-ui/src/features/config-wizard/AuthenticateAwsForm.tsx
```diff
@@ -144,6 +150,13 @@ const AuthenticateAwsForm = () => {
...
+                  <CustomTextInput
+                    type="password"
+                    name="aws_session_token"
+                    label="Session Token (if applicable)"
+                    // "You can find more about creating access keys and secrets on AWS docs here."
+                    tooltip="The session token for temporary credentials."
+                  />
                   <CustomSelect
                     name="region_name"
                     label="AWS Region"
```
adamsachs, line 162: This is functional, really great that you were able to get this updated and working - thank you!

one small cosmetic nit that may make the form a little easier to understand would be to mark the non-optional fields as `isRequired`, which will have them show up with an asterisk in the UI, to indicate that they're required. Then we wouldn't need the `(if applicable)` text on the new field you've added. I'd also probably tweak the tooltip text. Attaching a screenshot of what this would look like, and also providing a code suggestion to hopefully clarify (you'd also need to add `isRequired` to the `aws_acces_key_id` field, github just doesn't let me suggest that far up!):

```suggestion
                    // "You can find more about creating access keys and secrets on AWS docs here."
                    tooltip="The secret associated with the Access Key ID used for authentication."
                    isRequired
                  />
                  <CustomTextInput
                    type="password"
                    name="aws_session_token"
                    label="Session Token"
                    // "You can find more about creating access keys and secrets on AWS docs here."
                    tooltip="The session token when using temporary credentials."
                  />
                  <CustomSelect
                    isRequired
                    name="region_name"
                    label="AWS Region"
```


image


all things considered, this form hasn't been updated in a fairly long time, and it probably needs a more general redesign that's outside of the scope of this unit of work/PR, so if you want to stick with your original implementation, that'd be fine!

src/fides/cli/utils.py
```diff
@@ -330,6 +331,7 @@ def handle_aws_credentials_options(
         aws_config = AWSConfig(
             aws_access_key_id=access_key_id,
             aws_secret_access_key=secret_access_key,
+            aws_session_token=session_token,
```
adamsachs, line 334: interesting, so this seems to still work even when passed through as `None` - seems like the `boto3` client instantiation just ignores any kwargs with `None` values in its constructor, as that's what is going to happen downstream at https://github.com/ethyca/fides/blob/main/src/fides/connectors/aws.py#L23-L33 
```python
service_client = boto3.client(
        service,
        **config_dict,
    )
``` 


nothing to change here since it all works as we want and `boto3` is pretty defensive, just thought it was noteworthy behavior!

adamsachs (approved):
thanks so much for this contribution, this change looks great @tariqajyusuf! appreciate all the diligence in getting this supported throughout the stack, from the CLI to the UI. i was able to test manually with some of our internal AWS accounts and things work great with temporary credentials, as well as with permanent credentials (i.e. no regressions).

i left one UI nitpick that you can feel free to ignore, i'll leave it up to your discretion! the other piece of this is that we'll want to get automated test coverage for this new functionality in place, but that's not really something you're able to do outside of our org (since the GH actions will require our GH secrets, etc.). so i'll go ahead and file a follow up ticket for myself to add that test coverage in - it should be a relatively straightforward update.

thanks again for being proactive and thoughtful in contributing this improvement to our codebase/product, we really appreciate it! 🙏

src/fides/cli/utils.py
tariqajyusuf: Yeah I was watching for this to have some issues as I've experienced that with a few other AWS SDKs but it seems that it does a pretty good job at checking here.

PR #4609, opened 2024-02-12
Context test

galvana:
### Description Of Changes

Alternate proposal for contextualized logging. Uses a combination of a `Contextualizable` mixin and a `log_context` decorator. The decorator iterates through the arguments of the decorated function and calls `get_log_context` on every argument that implements `Contextualizable`. The `log_context` decorator accepts optional keyword arguments to add additional context outside of the contextualizable arguments.

### DIFF
--- a/src/fides/api/models/privacy_request.py
+++ b/src/fides/api/models/privacy_request.py
@@ -8,6 +8,7 @@ from enum import Enum as EnumType
 from typing import Any, Dict, List, Optional, Set, Union
 
 from celery.result import AsyncResult
+from fides.api.util.logger_context_utils import Contextualizable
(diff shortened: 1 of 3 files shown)


src/fides/api/models/privacy_request.py
```diff
@@ -950,6 +951,9 @@ def error_processing(self, db: Session) -> None:
             db=db, data={"message_sent": False, "privacy_request_id": self.id}
         )
 
+    def get_log_context(self):
+        return {"privacy_request_id": self.id}
```
adamsachs, line 955: nit: the key here should be the new enum you've defined (in the other PR), right?

PR #4611, opened 2024-02-12
experience + notice data migration via clearing of experience (config) data

adamsachs:
Closes [PROD-1614](https://ethyca.atlassian.net/browse/PROD-1614)

### Description Of Changes

Trying out a simplified auto-migration approach that deletes all PrivacyExperienceConfig (and PrivacyExperience) records and relies on records being re-created from OOB (out of the box) server logic post-migration. This implies that we'd need execute some sort of manual migration to propagate any "manual" changes to `PrivacyExperienceConfig` records made by existing users on top of the default records.

`PrivacyNotice`s and their data are effectively auto-migrated to the new translation-based data model (a much simpler auto-migration). 

### Code Changes

* [x] some tweaks to existing schema migration to get things working (needs some review)
* [x] rough draft of a simplified data migration with approach described above. 

### Steps to Confirm

* [x]  manual testing with OOB records:
    1. [x] build and run fidesplus `main`: `nox -s "build(slim)" "dev(slim)" -- dev`
    2. [x] after server has successfully booted up, tear it down, but DO NOT wipe the DB/teardown volumes.
    3. [x] checkout (or branch off) `consent_multitranslation_plus` in fidesplus and update `requirements.txt` to point to the commit at the HEAD of the branch PR'd here
    4. [x] re-run fidesplus server with e.g. ` nox -s "build(slim)" "dev(slim)" -- dev`
    5. [x] the migrations on this branch should have run (includes data migration), and the _new_ OOB server loading logic should have also executed. This leaves us with:
        - migrated `PrivacyNotice` (and `NoticeTranslation`) records; updated `PrivacyNoticeHistory` records.
        - net-new `PrivacyExperienceConfig` and `PrivacyExperience` records. 
        - a mix of old and new `PrivacyExperienceConfigHistory` records. The old ones point at the now nonexistent old-style `PrivacyExperienceConfig` records. The new ones point at the new `ExperienceTranslation` records.

here's the privacy experience configs i'm left with after these steps. they look basically as i'd expect, but attaching a screenshot so we can see concretely how these records look: 
image

* [x] manual testing with _edits_ to OOB records, mimicking real-world user edits. perform the same steps above, but between steps 1 and 2, make some manual changes, e.g.:
    * [x] update `PrivacyNotice` fields without changing the record keys or making new records. These should still be pointed to by the new `PrivacyExperienceConfig`s post-migration.
    * [x] try making net-new `PrivacyNotice` records. these will _not_ be pointed to by new `PrivacyExperienceConfig` records automatically, but they should be able to be pointed to via some manual updates post-migration
    * [ ] try deleting `PrivacyNotice` records. What happens here? ❓ 
    * [x] try editing `PrivacyExperienceConfig` records. the edits _will not_ persist across the migration, but ensure migration doesn't fail. establish a path for holding onto the edits (just via history records?) and re-applying them "manually" post-mgiration. 

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

[PROD-1614]: https://ethyca.atlassian.net/browse/PROD-1614?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data_clear_experience_configs.py
+++ b/src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data_clear_experience_configs.py
@@ -1,0 +1,186 @@
+"""translation data
+
+Revision ID: 14acee6f5459
+Revises: a1e23b70f2b2
+Create Date: 2024-01-09 21:17:13.115020
+
+"""
+
+import uuid
+
+import pandas as pd
+from alembic import op
+from sqlalchemy import text
+
+# revision identifiers, used by Alembic.
+
+revision = "14acee6f5459"
+down_revision = "a1e23b70f2b2"
+branch_labels = None
+depends_on = None
+
+
+def generate_record_id(prefix):
+    return prefix + "_" + str(uuid.uuid4())
+
+
+def remove_existing_experience_data(bind):
+    """
+    We remove any existing PrivacyExperience and PrivacyExperienceConfig records,
+    to allow for a "clean-slate" in creation of _new_ PrivacyExperienceConfigs based on the
+    "out of the box" templates.
+
+    This means that any existing user-edited data on PrivacyExperienceConfig records will need to be
+    _manually_ re-added to the post-migration state. It will not be migrated as part of the automatic migration.
+    """
+
+    def drop_fk_constraint(bind):
+        """
+        Drops FK constraint between PrivacyExperienceConfigHistory and PrivacyExperienceConfig,
+        to allow us to delete PrivacyExperienceConfigs without removing their associated history records
+
+        NOTE: should we just remove the column altogether? will we want to keep it around in the "new" model,
+        given that our PrivacyExperienceConfigHistory records will now be pointing to ExperienceTranslation records?
+        On the other hand, it could be useful to help in performing manual reconciliation steps post-migration.
+        """
+        op.drop_constraint(
+            "privacyexperienceconfighistory_experience_config_id_fkey",
+            "privacyexperienceconfighistory",
+            type_="foreignkey",
+        )
+
+    def delete_experiences(bind):
+        """
+        Delete all PrivacyExperience records.
+
+        These records will be effectively re-created as a result of the "out of the box" creation
+        of 'new-style' PrivacyExperienceConfig records post-migration.
+        """
+        bind.execute(
+            text(
+                """
+                DELETE FROM privacyexperience;
+                """
+            )
+        )
+
+    def delete_experience_configs(bind):
+        """
+        Delete all PrivacyExperienceConfig records.
+
+        These records will be effectively re-created via the server's "out of the box" creation
+        of the updated PrivacyExperienceConfig records, post-migration.
+        """
+        bind.execute(
+            text(
+                """
+                DELETE FROM privacyexperienceconfig;
+                """
+            )
+        )
+
+    drop_fk_constraint(bind)
+    delete_experiences(bind)
+    delete_experience_configs(bind)
+
+
+def migrate_notices(bind):
+    """
+    Migrates existing PrivacyNotice records to the new, translation-based data model.
+
+    Notice keys are not impacted. Most existing notice data is moved onto a linked NoticeTranslation record,
+    which is assumed to be english-language as a default. If the existing notice text/content is _not_ english-language,
+    then users will need to *manually* adjust the language on the created translation record.
+
+    The existing PrivacyNoticeHistory records are adjusted to point to the associated NoticeTranslation record,
+    as is expected in the new data model.
+    """
+
+    get_notices_query = text(
+        """SELECT id from privacynotice;
+         """
+    )
+
+    create_notice_translation_query = text(
+        """
+        INSERT INTO noticetranslation (id, privacy_notice_id, language, title, description, created_at, updated_at)
+        SELECT :record_id, :original_notice_id, :language, name, description, created_at, updated_at
+        FROM privacynotice
+        WHERE id = :original_notice_id
+        """
+    )
+    for res in bind.execute(get_notices_query):
+        # Create a Notice Translation for each notice, setting to english as a default
+        bind.execute(
+            create_notice_translation_query,
+            {
+                "record_id": generate_record_id("pri"),
+                "language": "en",
+                "original_notice_id": res["id"],
+            },
+        )
+
+    # Add a FK from the Notice History back to the Notice Translation
+    link_notice_history_to_translation_id = text(
+        """
+        UPDATE privacynoticehistory
+        SET translation_id = noticetranslation.id, title = privacynoticehistory.name
+        FROM noticetranslation
+        WHERE noticetranslation.privacy_notice_id = privacynoticehistory.privacy_notice_id;
+    """
+    )
+
+    bind.execute(link_notice_history_to_translation_id)
+
+
+def downward_migrate_notices(bind):
+    """
+    Back-migration of PrivacyNotice data.
+
+    Moves data in existing NoticeTranslation records back to the associated PrivacyNotice records,
+    to match the old data model.
+    # NOTE: this won't work "well" if multiple translations are present. Should we even attempt it?
+
+    The existing PrivacyNoticeHistory records are adjusted to point back to the associated PrivacyNotice records,
+    rather than the NoticeTranslation records, as is expected in the old data model.
+
+    # TODO: should we attempt to back-migrate region data from the ExperienceConfig records to the PrivacyNotice records? Feels dicey...
+    """
+
+    noticetranslation_data_to_notice_query = text(
+        """
+        UPDATE privacynotice
+        SET name = noticetranslation.title, description = noticetranslation.description,
+        FROM noticetranslation
+        WHERE noticetranslation.privacy_notice_id = privacynotice.id
+        """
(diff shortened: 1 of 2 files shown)


src/fides/api/alembic/migrations/versions/a1e23b70f2b2_add_notice_and_exp_translations.py
```diff
@@ -201,7 +202,9 @@ def upgrade():
     )
     op.add_column(
         "privacyexperienceconfighistory",
-        sa.Column("language", sa.String(), nullable=False),
+        sa.Column(
+            "language", sa.String(), server_default="en", nullable=False
```
adamsachs, line 206: want to review the approach here - could either be a server default or we could relax the nullable constraint, but then i'd assume we still need to make an assumption (in the data migration?) that the value should be english? there may very well be good reasons to avoid assuming a server default, so i'd be open to that!

src/fides/api/alembic/migrations/versions/a1e23b70f2b2_add_notice_and_exp_translations.py
```diff
@@ -263,10 +266,16 @@ def upgrade():
...
-        "privacynoticehistory", sa.Column("language", sa.String(), nullable=False)
+        "privacynoticehistory",
+        sa.Column(
+            "language", sa.String(), server_default="en", nullable=False
+        ),  # set "en" as a server default for existing values ... TODO: maybe make temporarily nullable and set english 'manually' instead ??
     )
     op.add_column(
-        "privacynoticehistory", sa.Column("title", sa.String(), nullable=False)
+        "privacynoticehistory",
+        sa.Column(
```
adamsachs, line 276: i think with this one we can set nullable temporarily here and then set to non-nullable on our post-data-migration schema migration (which will also be removing unneeded columns), right? i.e. the values here should be populated by the data migration, for every existing row, i think...

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data_clear_experience_configs.py
adamsachs, line 1: i think the (upward) data migration is actually relatively straightforward given the assumptions surrounding this. but i may be missing some nuances.

the downward data migration is a bit murkier to me. i'm not sure what we should even be attempting to do...i may be trying to do too much here, or maybe too little!

adamsachs (commented):
some light annotations ... code comments also provide some questions/context!

src/fides/api/alembic/migrations/versions/a1e23b70f2b2_add_notice_and_exp_translations.py
pattisdr: Great point yes, we need to go through every column where I'm saying "nullable=False" and do one of your two options above. I missed a few on the first pass.

 For this, I'd lean towards relaxing the nullable constraint, since there's also a unique constraint on this table of language 'x' experience_config_id.

src/fides/api/alembic/migrations/versions/a1e23b70f2b2_add_notice_and_exp_translations.py
```diff
@@ -263,10 +266,16 @@ def upgrade():
         nullable=True,
     )
     op.add_column(
-        "privacynoticehistory", sa.Column("language", sa.String(), nullable=False)
+        "privacynoticehistory",
+        sa.Column(
+            "language", sa.String(), server_default="en", nullable=False
+        ),  # set "en" as a server default for existing values ... TODO: maybe make temporarily nullable and set english 'manually' instead ??
```
pattisdr, line 272: temporarily nullable would be good 👍 .  Data migration can make them english.

src/fides/api/alembic/migrations/versions/a1e23b70f2b2_add_notice_and_exp_translations.py
pattisdr: I agree, nullable temporarily would be good, and data migration can populate.

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data_clear_experience_configs.py
```diff
@@ -0,0 +1,186 @@
...
+                "language": "en",
+                "original_notice_id": res["id"],
+            },
+        )
+
+    # Add a FK from the Notice History back to the Notice Translation
+    link_notice_history_to_translation_id = text(
+        """
+        UPDATE privacynoticehistory
+        SET translation_id = noticetranslation.id, title = privacynoticehistory.name
```
pattisdr, line 127: Should we populate the language too? knowing it might be wrong

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data_clear_experience_configs.py
```diff
@@ -0,0 +1,186 @@
...
+    Back-migration of PrivacyNotice data.
+
+    Moves data in existing NoticeTranslation records back to the associated PrivacyNotice records,
+    to match the old data model.
+    # NOTE: this won't work "well" if multiple translations are present. Should we even attempt it?
+
+    The existing PrivacyNoticeHistory records are adjusted to point back to the associated PrivacyNotice records,
+    rather than the NoticeTranslation records, as is expected in the old data model.
+
+    # TODO: should we attempt to back-migrate region data from the ExperienceConfig records to the PrivacyNotice records? Feels dicey...
```
pattisdr, line 147: If we're not going to forward migrate them, I don't know if we need to back fill these on the downrev.  I think we need to be taking db snapshots instead for this one.

pattisdr (commented):
haven't tested but looks like you're on the right track. 

I'd make sure we're thinking about what the "manual" path for upgrade too. If we need to go back in and update Experience Config Titles and Descriptions, for example, I'd recommend we run a script or edit through the UI rather than run bulk SQL queries.

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data_clear_experience_configs.py
```diff
@@ -0,0 +1,186 @@
...
+
+        NOTE: should we just remove the column altogether? will we want to keep it around in the "new" model,
+        given that our PrivacyExperienceConfigHistory records will now be pointing to ExperienceTranslation records?
+        On the other hand, it could be useful to help in performing manual reconciliation steps post-migration.
+        """
+        op.drop_constraint(
+            "privacyexperienceconfighistory_experience_config_id_fkey",
+            "privacyexperienceconfighistory",
+            type_="foreignkey",
+        )
```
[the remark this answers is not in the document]
adamsachs: was wondering about this, makes sense !

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data_clear_experience_configs.py
```diff
@@ -0,0 +1,186 @@
...
+            text(
+                """
+                DELETE FROM privacyexperienceconfig;
+                """
+            )
+        )
+
+    drop_fk_constraint(bind)
+    delete_experiences(bind)
+    delete_experience_configs(bind)
```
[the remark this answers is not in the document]
adamsachs: wouldn't that mean we should delete the experiences first? i would think that we first delete the record that is doing the pointing, rather than the one that's _pointed to_ ... but maybe i am wrong or just misunderstanding! :)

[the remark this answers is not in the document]
pattisdr: you're right sorry i've been looking at FK's all day!

pattisdr:
A lingering thing I haven't addressed, we have a component type "overlay", but now that has been split into "banner+modal" or just "modal".  My current iteration just lets the word "overlay" mean "banner+modal" to simplify migration. But if the data migration is just dropping all ExperienceConfigs, maybe we could go ahead and introduce the banner+modal concept then too, and populate those fresh out of the box.

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data_clear_experience_configs.py
adamsachs: agreed, was leaning that way too but wanted to get another opinion 👍 i'll remove the TODO here.

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data_clear_experience_configs.py
adamsachs: oh, yeah i think that would make sense - forgot we had that column on this table!

src/fides/api/alembic/migrations/versions/a1e23b70f2b2_add_notice_and_exp_translations.py
adamsachs: makes sense, thanks for pointing this out on the data migration side too!

adamsachs:
closing as we went a different route for this in https://github.com/ethyca/fides/pull/4636

PR #4615, opened 2024-02-14
Add i18n support to FidesJS to prep for language localization

NevilleS:
Closes PROD-1637

### Description Of Changes

This sets up FidesJS to support localization - the PR itself is fairly large since this is a fairly wide-reaching change, but the core changes are as follows:
1. Add a new `lib/i18n` module that implements some basic i18n features like loading localized message catalogs and translating strings
2. Initialize the `i18n` module during FidesJS initialization with both static strings (from source) and dynamic strings from the `privacy-experience` API

The next step will be to localize all the strings in the UI itself and give the user some tools to select their preferred language, etc. I'm moving that into a follow-up PR: #4631

### Code Changes

- [X] Implement a minimal `i18n` library that supports `load` to load locale-specific message catalogs, `activate` to select the active locale, and `t` to localize strings based on a message id
- [X] Ensure that our minimal `i18n` library is API-compatible with LinguiJS for future improvements: https://lingui.dev/
- [X] Add unit tests for the minimal `i18n` library that pass against both our implementation *and* LinguiJS
- [X] Add `i18n-utils` helpers to integrate the `i18n` library into our codebase
- [X] Load staticallly-defined i18n message catalogs from source (e.g. `locales/en/messages.json`)
- [x] Load dynamically-defined i18n message catalogs from the `privacy-experience` API alongside static messages
- [X] Automatically detect the user locale from the browser `navigator.language`
- [X] Allow overriding the user locale using `?fides_locale={language}` query param
- [X] Automatically match the user's locale to the best available locale from the message catalogs
- [x] Integrate with latest API and update types

### Steps to Confirm

* [X] General regression testing of all experience types (banner, modal, privacy center) - this should be purely additional logic and some refactors under the hood

### Pre-Merge Checklist

* Documentation:
* [X] Update `CHANGELOG.md`


clients/fides-js/src/lib/i18n/index.ts
```diff
@@ -0,0 +1,92 @@
...
+/**
+ * Initialize a global i18n singleton for use in the application.
+ *
+ * DEFER: Swap this implementation for LinguiJS when ready by:
+ * 1) import { i18n } from "@lingui/core"
+ * 2) Delete the call to setupI18n()
+ * 3) Delete our setupI18n.ts file entirely!
+ */
+
+const i18n: I18n = setupI18n();
```
NevilleS, line 87: This shows how we could swap out our i18n implementation (from `setupI18n`) with LinguiJS when ready 👍

clients/fides-js/src/lib/i18n/locales/en/messages.json
```diff
@@ -0,0 +1,3 @@
+{
+  "experience.title": "Title in English"
```
NevilleS, line 2: TODO: update with real message catalog.

Note that I won't actually have `experience.*` strings here, that prefix will be used for the dynamic strings.

clients/fides-js/src/lib/i18n/locales/fr/messages.json
```diff
@@ -0,0 +1,3 @@
+{
+  "experience.title": "Title en Francais"
```
NevilleS, line 2: TODO: update with real message catalog

clients/fides-js/src/lib/i18n/index.ts
```diff
@@ -0,0 +1,92 @@
...
+ *     "experience.accept_button_label": "Accept",
+ *   },
+ *   "fr": {
+ *     "experience.title": "Vos choix en matière de confidentialité",
+ *     "experience.accept_button_label": "Accepter",
+ *   }
+ * }
+ * ```
+ */
+type Messages = Record<string, string>;
```
eastandwestwind, line 28: Is it possible for this to be`Record<string, string | null | undefined>` ? When we `extractMessagesFromExperienceConfig`, for example, can we ever hit a state where a translation with a given key does not exist (like for optional fields on an experience translation).

clients/fides-js/src/lib/i18n/i18n-utils.ts
```diff
@@ -0,0 +1,357 @@
...
+  let currentLocale: Locale = DEFAULT_LOCALE;
+
+  // Messages catalog, which stores i18n-ized messages grouped by locale
+  const allMessages: Record<Locale, Messages> = {};
+
+  // Return a new I18n instance
+  return {
+    activate: (locale: Locale): void => {
+      currentLocale = locale;
+    },
```
eastandwestwind, line 320: When we implement language switching from the UI, all we'll need to do is call this `activate` fn with the chosen locale, right?

clients/fides-js/src/lib/consent-types.ts
```diff
@@ -100,6 +100,9 @@ export type FidesOptions = {
 
   // Encodes cookie as base64 on top of the default JSON string
   base64Cookie: boolean;
+
+  // Allows specifying the preferred locale used for translations
+  fidesLocale?: string;
```
eastandwestwind, line 105: Could you explain what happens if I use geolocation query param like this when running PC: `http://localhost:3000/?geolocation=us-nc`, but also have `fidesLocale` set up to `fr-CA`? 

I think the experience call will still use `us_nc`, and return notices for that region, but I should see the `fr` lang in the UI if we have that lang configured in translations?

eastandwestwind (changes requested):
Overall this is looking solid, and excited to start testing full-stack. I like your descriptive comments and the fact you've anticipated being able to easily swap out with LinguiJS in the future.

I have some minor Qs and asks, otherwise let's get this fully integrated with my work asap and  🚀 🚢

clients/fides-js/src/lib/i18n/i18n-utils.ts
```diff
@@ -0,0 +1,357 @@
...
+  i18n.load("es", messagesEs);
+  i18n.load("fr", messagesFr);
+  return ["en", "es", "fr"];
+}
+
+/**
+ * Parse the provided PrivacyExperience object and load all translated strings
+ * into the message catalog.
+ */
+export function updateMessagesFromExperience(
```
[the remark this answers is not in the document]
NevilleS: 👍 done

clients/fides-js/src/lib/i18n/index.ts
NevilleS: So, the *type* will never have undefined values, but it's definitely possible for a `i18n.t("some.key")` call to not match a string when it _thinks_ one should exist. So for example, something like `i18n.t("exp.accept_button_label")` isn't guaranteed to have a value.

Right now in that case it actually returns the key (e.g. `i18n.t("missing.key") === "missing.key"`) which'll look odd in the UI, so we'll need to check to see if we ever have cases where the components were relying on rendering empty strings... but I don't think so.

clients/fides-js/src/lib/i18n/i18n-utils.ts
NevilleS: Yup 👍

clients/fides-js/src/lib/consent-types.ts
NevilleS: Yes, you're correct. The locale is only used for client-side translations and doesn't affect the experience fetch.

clients/fides-js/src/lib/i18n/index.ts
eastandwestwind: Ah this makes sense 👍

eastandwestwind (approved):

PR #4620, opened 2024-02-16
Fides Location X Privacy Experience Filtering

pattisdr:
Closes #PROD-1636

❗ Needed for plus https://github.com/ethyca/fidesplus/pull/1319
❗ Bump downrev before merge

### Description Of Changes

Builds PrivacyNoticeRegion enum dynamically from locations and location groups. Also saves location groups when locations are saved.

### Code Changes

- Moves locations.yml file to OSS from Plus so it can be used at the SQLAlchemy model level along with some supporting functions
- Replaces existing PrivacyNoticeRegion enum with a new one that pulls locations and location_groups from the `locations.yml` file. 
- Allows storing location_groups in addition to just locations.  This will be useful when checking to see if a region can be added to a Privacy Experience Config - we can check locations and location group columns.
- Adds a method `filter_regions_by_location` to be used in Plus where if locations are configured, we restrict regions to just match configured locations and/or location groups

### Steps to Confirm

* [ ] See Plus PR for testing steps 

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`


src/fides/api/models/privacy_notice.py
```diff
@@ -40,114 +41,6 @@ class UserConsentPreference(Enum):
...
-        ("pt", "pt"),  # portugal
-        ("ro", "ro"),  # romania
-        ("si", "si"),  # slovenia
-        ("sk", "sk"),  # slovakia
-        ("fi", "fi"),  # finland
-        ("se", "se"),  # sweden
-        ("gb_eng", "gb_eng"),  # england
-        ("gb_sct", "gb_sct"),  # scotland
-        ("gb_wls", "gb_wls"),  # wales
-        ("gb_nir", "gb_nir"),  # northern ireland
```
pattisdr, line 129: These are the only PrivacyNoticeRegions that have been removed. The discrepancy will not be a big deal if we take care of it on Experiences and Privacy Experience Configs as part of the migration.  Privacy Preferences are saved against a "string" user location (since the user can be in any region, not necessarily a PrivacyNoticeRegion)

src/fides/api/models/privacy_experience.py
```diff
@@ -227,10 +228,28 @@ class PrivacyExperienceConfig(
...
+
     @property
     def regions(self) -> List[PrivacyNoticeRegion]:
-        """Return the regions using this experience config"""
-        return [exp.region for exp in self.experiences]  # type: ignore[attr-defined]
+        """Filters regions on configured location if applicable"""
+        db = Session.object_session(self)
+        if not db:
+            # For dry update scenarios
+            return []
```
pattisdr, line 250: We create a PrivacyExperienceConfig in memory on the dry_update but don't add to the Session.  Now that I'm accessing the object's session as part of the PrivacyExperienceConfig.regions, this would break

src/fides/api/models/privacy_experience.py
```diff
@@ -227,10 +228,28 @@ class PrivacyExperienceConfig(
...
+        """All regions physically linked to the PrivacyExperienceConfig regardless of which locations
+        have been configured"""
+        db = Session.object_session(self)
+        if not db:
+            # For dry update scenarios, where PrivacyExperienceConfig is not bound to a session
+            return []
+
+        return [
+            exp.region for exp in self.experiences.order_by(PrivacyExperience.region)
+        ]
```
pattisdr, line 242: Use this if you want to know all regions, regardless of location filtering. For example, we may ship with more regions out of the box than the user actually configures.

pattisdr (commented):
More tests are needed, but I'd like to get this up for an early review first since there are some gray areas remaining.

src/fides/api/models/privacy_experience.py
```diff
@@ -574,19 +593,18 @@ def upsert_privacy_experiences_after_config_update(
...
     Links regions to a PrivacyExperienceConfig by adding or removing PrivacyExperience records.
     """
-    current_regions: List[PrivacyNoticeRegion] = experience_config.regions
-    removed_regions: List[
+    regions_to_remove: List[
         PrivacyNoticeRegion
     ] = [  # Regions that were not in the request, but currently attached to the Config
         PrivacyNoticeRegion(reg)
-        for reg in {reg.value for reg in current_regions}.difference(
+        for reg in {reg.value for reg in experience_config.all_regions}.difference(
```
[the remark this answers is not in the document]
adamsachs: thanks for calling this out, i think it makes sense - may be good just to call out the handling of non-selected locations in a code comment, just to reinforce the importance of using the `all_regions` property in a scenario like this 👍

src/fides/api/models/privacy_notice.py
adamsachs: thank you for noting the discrepancies between the "new" locations and the old region enum members!

src/fides/api/schemas/locations.py
```diff
@@ -0,0 +1,168 @@
...
+    filtered_regions: List[PrivacyNoticeRegion] = []
+
+    for region in regions:
+        # Explode any Location Groups to Locations so they can be compared against apples to apples
+        # with locations
+        expanded_regions: Set[str] = location_group_to_location.get(
+            region.value, {region.value}
+        )
+        if expanded_regions.issubset(locations):
+            filtered_regions.append(region)
```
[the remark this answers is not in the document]
adamsachs: ah yes, this is a good edge case - nicely handled! the logic/behavior here seems correct. i think the only tweak i'd consider is whether we want to more properly store the "selected location groups" as a general-purpose method on the LocationRegulationSelections model, so we don't have this logic buried within this helper function.

basically, would it make sense for us to be able to call LocationRegulationSelections.get_selected_location_groups(db) and then compare against that here a bit more directly? that just feels a bit more transparent to me.

the LocationRegulationSelections.get_selected_location_groups method would, underneath the hood, probably do something similar to what's done here, or perhaps we'd take some other measures when actually making any updates to selected locations to ensure that the selected location groups are updated accordingly at that time, i.e. to prevent repeatedly "recalculating" the selected groups, for performance reasons

another middle-ground would be to just pull some of this logic here into its utility function on the LocationRegulationSelections module. i think that's the main thing i'm after - to pull this logic into a function/method that's a bit more visible and easy for other use cases to access as needed.

let me know if that makes sense!

src/fides/api/models/privacy_experience.py
```diff
@@ -609,7 +627,7 @@ def upsert_privacy_experiences_after_config_update(
                 },
             )
 
-    return experience_config.regions
+    return experience_config.all_regions
```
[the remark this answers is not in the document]
adamsachs: ok that makes sense - but in effect, we generally wouldn't expect to have regions that refer to unselected locations at this point, right? they're usually removed as part of the logic above, if i'm following correctly.

i.e. are there scenarios where the `regions` arg that's passed in will actually include non-selected locations? maybe there are scenarios (e.g. on out of the box loads?), which is fine - just curious/trying to firm up my understanding! (i'll also take a look at the plus PR which may provide some insight here)

adamsachs (approved):
@pattisdr this is looking great to me! really just one potential tweak i see, which is to make some of the "selected location group" functionality a bit more visible/centralized so that it can be used in other spots as needed moving forward. let me know what you think of that.

src/fides/api/models/privacy_experience.py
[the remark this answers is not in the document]
pattisdr: I'll still not sure the best way to handle this though! https://github.com/ethyca/fidesplus/pull/1319#discussion_r1496784040

src/fides/api/schemas/locations.py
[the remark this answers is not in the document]
pattisdr: OK progress so far! I've updated to store both locations and location groups from the API endpoints like we talked about.  PATCH locations endpoint looks at locations passed in and combines them into location groups where applicable and saves them into a new column.  Also updated regulation changes to locations can also propagate back to location groups.

That makes this existing funciton a lot simpler, to compare PrivacyNoticeRegions to a union of saved Locations and LocationGroups.

I've left this existing method where it is for circular import regions although it can be moved to a different location if needed.

src/fides/api/models/privacy_experience.py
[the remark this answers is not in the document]
pattisdr: As-is, here, `all_regions` and `regions` should match, I was just trying to reflect that "locations" aren't taken into account here by using "all_regions".

However, I am undecided about expected behavior around a PATCH of a Privacy Experience Config as regards to OOB Regions being removed in the background!  We might lose our magic there.

https://github.com/ethyca/fidesplus/pull/1319/files#r1496784040

src/fides/api/models/privacy_experience.py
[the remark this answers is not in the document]
pattisdr: From comments on that other PR, possible for OOB regions to get removed here that cannot get restored later, but this is sounding like a risk we're fine with.

src/fides/api/models/privacy_experience.py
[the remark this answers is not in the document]
pattisdr: OK all of this has been reverted we're no longer filtering on regions -

src/fides/api/models/location_regulation_selections.py
```diff
@@ -146,6 +176,38 @@ async def get_selected_locations_async(
...
+
+    @classmethod
+    async def get_selected_location_groups_async(
+        cls,
+        async_session: AsyncSession,
+    ) -> Set[str]:
+        """
+        Utility method to get the selected_location_groups, returned as a Set.
+
+        Location Groups aren't saved directly, but as a side-effect of saving locations
```
adamsachs, line 202: i appreciate the diligence in the docstrings here! helps provide context throughout 👍

src/fides/api/models/location_regulation_selections.py
```diff
@@ -195,3 +257,18 @@ async def get_selected_regulations_async(
             if record:
                 return set(record.selected_regulations)
             return set()
+
+
+def group_locations_into_location_groups(saved_locations: Iterable[str]) -> Set[str]:
+    """Combines saved Locations into Location groups if applicable
+    All Locations must be present for a Location Group to be valid.
+    """
+    from fides.api.schemas.locations import location_group_to_location
```
adamsachs, line 266: oof i feel like i've set us up with a bad structure between the db model and the API schema to get us to the point where we need to do this 🤔 

i can see it's not straightforward to resolve this dep conflict so i get why you did this. the best resolution/alternative i see at the moment - i feel we'll probably set ourselves up better moving forward if we move a lot of the enum creation and baseline location "registry" loading out of the API schema and either into the DB model class or into a shared util that both can depend on/point to. we don't need to do this in the current iteration since this should be fine (not trying to make more work that's not strictly necessary!)

i just teased this out a bit in my head, haven't explored thoroughly or worked out fully how it fits together with your updates across both repos. again, we don't need to do this now, especially if it doesn't feel straightforward to you. but whenever we need to do an in-function import like this, it gives me pause and makes me concerned about maintainability...

src/fides/api/models/location_regulation_selections.py
```diff
@@ -195,3 +257,18 @@ async def get_selected_regulations_async(
...
+
+    saved_location_groups: List[str] = []
+
+    for location_group, mapped_locations in location_group_to_location.items():
+        if mapped_locations.issubset(saved_locations):
+            saved_location_groups.append(location_group)
+
+    return set(saved_location_groups)
```
adamsachs, line 274: the bulk of this function is clear and concise though! 👍

adamsachs (approved):
this side looks good @pattisdr! just one comment on the dynamic/in-function import and thinking about how we may want to resolve that moving forward, not strictly necessary for this iteration 👍

src/fides/api/models/location_regulation_selections.py
pattisdr: yep, I was taking a shortcut here - since you flagged as well, I can spend a little time trying to untangle.

pattisdr: For now it's on the model! We can figure out how to extract better later.  If I didn't have location groups being a side effect of saving locations this would be easier but I know it was desired to group this action as much as possible.

PR #4626, opened 2024-02-21
Update privacy experience form (do not merge)

jpople:
Closes PROD-1641

### Description Of Changes

Update the basic privacy experience configuration form.  Previewing, translations, and privacy notices should be merged into this before merging to main.

### Code Changes

* Update types from new API
* Add calls to new endpoints 
* Replace previous PrivacyExperienceForm component and delete the old specific subforms for overlay and privacy center 
* Add new ScrollableList component 

### Steps to Confirm

Running against backend branch `consent_multitranslation_plus`:
* [ ] Navigate to `/consent/privacy-experience`
* [ ] Click on "Create experience" (button is temporary for testing)
* [ ] Fill out form and submit
* [ ] Should submit, and new experience should be shown in experiences table
* [ ] Click on a row in the experiences table
* [ ] Make changes and submit
* [ ] Should submit, and changes should apply

### Pre-Merge Checklist

* [ ] Update `CHANGELOG.md`


eastandwestwind:
Just documenting some weird behavior possibly with `framer-motion` that causes all lists on the form to move around (see vid for reference):

1. Removing items from a list
2. Adding items to a list

https://www.loom.com/share/a1e8d5325bed481ebce4ee1ce3f1157b

eastandwestwind:
Making a separate note that this PR will temporarily break enable / disable experience on the list view, e.g. `http://localhost:3000/consent/privacy-experience`. This is because PATCH `http://localhost:3000/api/v1/experience-config/{{id}}` requires extra data to succeed. 

@jpople could you point to the ticket where we intend on working on this list view?

eastandwestwind (commented):
Awesome work on this @jpople! I have some initial thoughts, mostly around presentation and actual functionality, less around nit-picking code style. See comments above, in addition to these small Qs below.

I think it's more important to get this integrated asap if everything is looking / functioning as we expect.

jpople:
@eastandwestwind

- I also noticed the issue with all of the animated elements scrolling at once on addition or removal; I think it's an issue with Framer, or possibly an issue with the specific version of Framer we're using (I had to go back a few for compatibility with our React version).  I think the problem is caused by the container box for the ScrollableList expanding/shrinking to fit new items instantaneously but the items always animating when they change position-- I think a possible fix would be to add an animation to the expanding/shrinking, but frankly I think it would take me a while to figure out how to do that properly and that seems not worth it.
- Good catch about the enable/disable PATCH; I think that's probably within the scope of this ticket so I can add that on, doesn't seem like too much extra (famous last words)
- Those are both good calls about the business logic and labels.  Some of these I think I took from a slightly older design, so they might be a little off, I'll meet with Michael about getting those things polished up.

PR #4634, opened 2024-02-26
Removing identities from cache

galvana:
Closes [PROD-1605](https://ethyca.atlassian.net/browse/PROD-1605)

### Description Of Changes

Removing provided identities, custom privacy request fields, and encryption keys from the cache. This was done to prevent required values from expiring from the cache if a request was not immediately approved. The new guidance is to store any values provided with the original privacy request creation to the database. Values generated during the request such as masking secrets and derived identities are still kept in the cache only (not to the database).

### Code Changes

* [ ] Added `encryption_key` to the `privacyrequest` table as an encrypted string value
* [ ] Replaced calls to `get_persisted_custom_privacy_request_fields` with `get_custom_privacy_request_field_map` to better reflect the return type (we are returning a map instead of a list of fields)
* [ ] Replaced calls to `get_cached_identity_data` with `get_identity_map`
* [ ] Removed the `cache_data` function and moved `create_masking_secrets` to the `Policy` model
* [ ] Replaced calls to `cache_identity` with `persist_identity`

### Steps to Confirm

* [ ] Modify the `.env` file in the root directory of fides repo and add `FIDES__REDIS__DEFAULT_TTL_SECONDS=30`
* [ ]  Run `nox -s "fides_env(test)"` and wait for the test environment to start up
* [ ] Open the Privacy Center and submit an access request for jane@example.com
* [ ]  Open the Admin UI within 30 seconds and approve the request, and confirm that the access request succeeds
* [ ]  Open the Privacy Center and submit a second access request for jane@example.com
* [ ] Wait 60 seconds!
* [ ] Open the Admin UI and approve the second request, and confirm that the access request processes successfully

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

[PROD-1605]: https://ethyca.atlassian.net/browse/PROD-1605?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/fides/api/service/privacy_request/request_runner_service.py
```diff
@@ -415,6 +417,9 @@ async def run_privacy_request(
             ) and can_run_checkpoint(
                 request_checkpoint=CurrentStep.erasure, from_checkpoint=resume_step
             ):
+                _verify_masking_secrets(policy, privacy_request_id, resume_step)
```
galvana, line 420: Important change! We don't cache the masking secrets until we get to the erasure step. In the event that we resume the erasure, we will error the request if masking secrets are needed for the given policy but have expired. We won't re-generate masking secrets because they need to be consistent for each erasure request.

tests/ops/models/test_privacy_request.py
```diff
@@ -392,7 +392,7 @@ def test_trigger_two_way_policy_webhook_add_derived_identity(
             privacy_request.trigger_policy_webhook(webhook)
             db.refresh(privacy_request)
             assert privacy_request.status == PrivacyRequestStatus.in_processing
-            assert privacy_request.get_cached_identity_data() == {
+            assert privacy_request.get_identity_map() == {
                 "email": "customer-1@example.com",
                 "phone_number": "+5555555555",
```
galvana, line 397: This existing assertion is justification for the order of provided and derived identities in `get_identity_map`

src/fides/api/service/privacy_request/request_runner_service.py
pattisdr: I think there's an edge case here where the failed step expired from the cache - we added awhile back if that was the case to let you re-process from the beginning - so resume_step will be None, but it's possible some data was already masked?

pattisdr (commented):
Does this need a special rollout plan?  

thinking about data that was cached in redis but you're storing somewhere else now - what about privacy requests that are in the queue that don't have this saved or privacy requests that failed -

PR #4636, opened 2024-02-28
multilang data migration

adamsachs:
### Description Of Changes

Working draft of multilang data migration ... 

- Loads new experience config defaults into memory
- Queries and stores existing experience config data
- Reconciles the above two based on some assumptions/mappings of old experience configs/notices to new experience configs, holds this in memory
- Clears existing experiences and experience configs from the db
- Creates experience configs and associated artifacts based on reconciled data in memory
    - experience configs
    - experience config historys
    - experiences
    - experience notices
    - experience translations
- Notices are updated "in place", basically just moving their text fields to a translation record and adding a history record

Still TODO:
- [x] link experience configs to templates?? means the migration will need to create template records ... 
- [ ] better handle down migration??

### Code Changes

### Steps to Confirm

For now, I've just been testing the core migration logic here in OSS fides, since the pieces layered in via OOB loading on fidesplus are relatively minor
* [ ] `git checkout main`, start your fides server (`nox -s dev`). ensure you start with a "fresh" db state.
* [ ] make updates to `PrivacyExperienceConfig` and `PrivacyNotice` records. The script below helps automate some updates to `PrivacyExperienceConfig` records. `PrivacyNotice` records don't have an update API exposed on OSS, so I've just been editing the records' regions field directly on the DB
* [ ] shut down your server, but DO NOT teardown volumes/wipe the DB!
* [ ] `git checkout asachs/multilang-data-migration-v2` (this branch), then `nox -s dev`
* [ ] migration should run successfully on startup. ensure no errors when migration is run on startup, and that DB state looks as expected after startup

[patch_config_updates_migration_test.py.txt](https://github.com/ethyca/fides/files/14440313/patch_config_updates_migration_test.py.txt)

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`


src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data_clear_experience_configs.py
```diff
@@ -0,0 +1,461 @@
...
+
+def load_default_experience_configs():
+    """
+    Loads default experience config definitions from yml file.
+
+    Maps the experience config definitions keyed by the ID enum.
+    """
+    raw_experience_config_map = {}
+    experience_config_map = {}
+    with open("privacy_experience_config_defaults.yml", "r") as f:
```
adamsachs, line 70: argh this doesn't work well if you try to run the migrations manually from the alembic dir, i will either need to embed this data in the file or figure out a way to make alembic be able to read this in either the app startup context or a manual migration context ...

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data_clear_experience_configs.py
```diff
@@ -0,0 +1,565 @@
...
+                    "is_default": True,
+                    "experience_config_id": res["id"],
+                },
+            )
+
+        # Repoint the Experience Config History to the Experience Translation, and set dismissable while we're here.
+        link_config_to_translation_id = text(
+            """
+            UPDATE privacyexperienceconfighistory
+            SET translation_id = experiencetranslation.id, dismissable = dismissable
```
pattisdr, line 405: It seems like there are other new fields that need to get ported back? allow_language_selection, auto detect language, name,

pattisdr (commented):
@adamsachs clever approach - I think this is a solid path using what we learned about existing customer data (no new experience configs created right?) to simplify how we create the new experience configs.  I like how you transfer changes from original experience configs over to potentially multiple new mapped experience configs.  This approach prevents a lot of manual updates needing to be made after the fact.  

I know you're working on this but let's get clear ahead of time what manual changes we'll need to make. It's hopefully a small set now, perhaps some language changes.

My comments are around specific fields I wanted to double check, no action needed specifically now, I will verify further in testing.

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data_clear_experience_configs.py
```diff
@@ -0,0 +1,565 @@
...
+    )
+    has_existing_experience_configs = False
+    for eec in existing_experience_configs:
+        has_existing_experience_configs = True
+        component = ComponentType(eec["component"])
+        configs = []
+        if eea_config_id := EEA_MAPPING.get(component):
+            configs.append(reconciled_experience_config_map.get(eea_config_id.name))
+        if us_config_id := US_MAPPING.get(component):
+            configs.append(reconciled_experience_config_map.get(us_config_id.name))
```
pattisdr, line 219: This is particularly clever, letting original experience config text changes be copied over to potentially multiple new experience configs.  I'd add more code comments here -

Also the decision to not ship with these GPP experiences was helpful!

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data_clear_experience_configs.py
```diff
@@ -0,0 +1,565 @@
...
+class ComponentType(Enum):
+    PrivacyCenter = "privacy_center"
+    Overlay = "overlay"
+    TCFOverlay = "tcf_overlay"
+
+
+EEA_MAPPING = {
+    ComponentType.PrivacyCenter: DefaultExperienceConfigTypes.EEA_PRIVACY_CENTER,
+    ComponentType.Overlay: DefaultExperienceConfigTypes.EEA_MODAL_AND_BANNER,
+    ComponentType.TCFOverlay: DefaultExperienceConfigTypes.EEA_TCF_OVERLAY,
```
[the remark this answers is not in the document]
adamsachs: this is a good question around what's probably the most confusing thing about the script as written. i think you're pointing out a legitimate gap/bug.

basically, i tried to keep a "parallel" path for any given experience migration - if the experience hasn't had any manual updates that apply to it, then fall back to using the default template as-is. i think this should work for cases when there have been no changes to the TCF experience on a client env. but what it doesn't account for is when a client has changed the TCF experience with e.g. an update to a text field. i don't know if this has occurred (i can try to get a survey from the customer team), but it seems plausible, and it would cause a problem.

seems like it may just be worthwhile to build in some special-handling for TCF experience, since it (and only it) cannot actually be tied to notices, i.e. we can't derive its regions based on associated notices 👍

thanks for narrowing in on this!

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data_clear_experience_configs.py
```diff
@@ -0,0 +1,565 @@
...
+    )
+
+    for experience_config in experience_configs:
+        _raw_experience_config_map[
+            DefaultExperienceConfigTypes(experience_config["id"])
+        ] = experience_config.copy()
+        experience_config["regions"] = (
+            set()
+        )  # remove the configs regions, since we will populate based on existing notices
+        experience_config["privacy_notices"] = set()  # initialize the notices field
```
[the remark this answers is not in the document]
adamsachs: yup a good thing to point out. as mentioned in my other comment, the underlying assumption for the TCF experience was that it would have no changes made to it so it'd fall back to the 'default' template logic, but that may not be a good assumption...

src/fides/api/alembic/privacy_experience_config_defaults.yml
```diff
@@ -0,0 +1,130 @@
...
+    - pt
+    - ro
+    - si
+    - sk
+    - fi
+    - se
+    - gb_eng
+    - gb_sct
+    - gb_wls
+    - gb_nir
```
pattisdr, line 71: ah we need the gb reconcilation too!

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data_clear_experience_configs.py
adamsachs: yup, that was a lifesaver 😅 

will add some more code comments for sure 👍

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data_clear_experience_configs.py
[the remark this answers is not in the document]
adamsachs: yup not a good assumption, have confirmed we have at least one customer who has edited the TCF overlay experience config 👍

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data_clear_experience_configs.py
```diff
@@ -0,0 +1,565 @@
...
+    - Then, we remove all existing Experience and ExperienceConfig records to start from a 'blank slate'
+    - Then, based on the reconciled existing ExperienceConfig data, we create new ExperienceConfig and Experience records
+    """
+
+    def create_new_experience_config_templates(experience_configs):
+
+        for experience_config in experience_configs:
+            create_experience_config_query = text(
+                """
+                INSERT INTO experienceconfigtemplate (id, component, name, disabled, privacy_notice_keys, regions)
```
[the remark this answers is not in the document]
adamsachs: whoops, i lazily thought i might be able rely on the remainder being populated by server defaults, didn't verify that assumption. i'll add them in to be populated properly here, thanks for flagging! 👍

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data_clear_experience_configs.py
```diff
@@ -0,0 +1,565 @@
...
+                VALUES (:id, :component, :name, :disabled, :privacy_notice_keys, :regions)
+                """
+            )
+
+            bind.execute(
+                create_experience_config_query,
+                {
+                    "id": experience_config[
+                        "origin"
+                    ],  # origin of the config record is the template ID
```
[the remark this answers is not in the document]
adamsachs: err this is something i added manually to the templates - sorry for not including that earlier! hopefully makes more sense now :)

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data_clear_experience_configs.py
```diff
@@ -0,0 +1,565 @@
...
+            )
+
+    def create_new_experience_config(experience_config):
+        """
+        Creates a new experience config based on the provided definition. Also creates corresponding history record.
+        """
+
+        create_experience_config_query = text(
+            """
+            INSERT INTO privacyexperienceconfig (id, version, origin, component, name, disabled, is_default, accept_button_label, banner_enabled, description, privacy_preferences_link_label, privacy_policy_link_label, privacy_policy_url, save_button_label, title, banner_description, banner_title, reject_button_label, acknowledge_button_label)
```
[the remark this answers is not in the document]
adamsachs: makes sense, i'll make another more comprehensive run through the fields in all places 👍

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data_clear_experience_configs.py
```diff
@@ -0,0 +1,565 @@
...
+            VALUES (:id,  :version, :origin,  :component, :name, :disabled, :is_default, :accept_button_label, :banner_enabled, :description, :privacy_preferences_link_label, :privacy_policy_link_label, :privacy_policy_url, :save_button_label, :title, :banner_description, :banner_title, :reject_button_label, :acknowledge_button_label)
```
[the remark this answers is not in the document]
adamsachs: hopefully the template file that i'd neglected to add makes this clearer now, but in any case i think it's worthy of some code comments for clarification 👍

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data_clear_experience_configs.py
```diff
@@ -0,0 +1,565 @@
...
+            """
+        )
+
+        bind.execute(
+            create_experience_config_query,
+            {
+                **experience_config,
+                "version": 1,
+                "is_default": True,
```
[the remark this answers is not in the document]
adamsachs: whoops, wasn't sure and didn't look very closely! i can remove if it feels confusing.

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data_clear_experience_configs.py
```diff
@@ -0,0 +1,565 @@
...
+            },
+        )
+
+        create_experience_config_history = text(
+            """
+             INSERT INTO privacyexperienceconfighistory (id, experience_config_id, origin, name, disabled, version, component, is_default, accept_button_label, banner_enabled, description, privacy_preferences_link_label, privacy_policy_link_label, privacy_policy_url, save_button_label, title, banner_description, banner_title, reject_button_label, acknowledge_button_label)
+             VALUES (:history_id, :id, :origin, :name, :disabled, :version, :component, :is_default, :accept_button_label, :banner_enabled, :description, :privacy_preferences_link_label, :privacy_policy_link_label, :privacy_policy_url, :save_button_label, :title, :banner_description, :banner_title, :reject_button_label, :acknowledge_button_label)
```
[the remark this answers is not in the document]
adamsachs: also may be worthy of a code comment :)

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data_clear_experience_configs.py
```diff
@@ -0,0 +1,565 @@
...
+            WHERE id = :experience_config_id
+            """
+        )
+        for res in bind.execute(get_experience_config_ids):
+            # Create a translation for each Experience
+            bind.execute(
+                create_translation_query,
+                {
+                    "record_id": generate_record_id("exp"),
+                    "language": "en_us",
```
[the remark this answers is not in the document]
adamsachs: yup i'd noticed that after committing, but forgot to address. thanks for flagging!

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data_clear_experience_configs.py
```diff
@@ -0,0 +1,565 @@
...
+        """
+        )
+
+        bind.execute(link_config_to_translation_id)
+
+    # first extract existing experience (config) data needed for migration
+    experience_configs = determine_needed_experience_configs(bind)
+
+    # wipe the existing experience + experience config data to start from a blank slate
+    remove_existing_experience_data(bind)
```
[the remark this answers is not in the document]
adamsachs: what type of backups? like a new table?

all in favor of making this as safe as possible!

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data_clear_experience_configs.py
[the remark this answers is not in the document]
pattisdr: you're right the templates should be updated by OOB logic I just was wondering why you were only loading some. I guess just the ones you needed.

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data_clear_experience_configs.py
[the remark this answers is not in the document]
pattisdr: a db snapshot I mean!

> ust a csv/json extract to disk? that would work!

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data_clear_experience_configs.py
adamsachs: ah, good catch! i guess when i was testing they were getting a server default? but does feel better to copy the right value from the template back.

src/fides/api/alembic/privacy_experience_config_defaults.yml
adamsachs: err right. i think we can just update the list here, and we shouldn't need to worry about it in terms of data migration?

pattisdr: were you not looking at the regions on the privacy notices though in your migration?

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data_clear_experience_configs.py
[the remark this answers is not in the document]
adamsachs: ah yes, i'm prescribing a db snapshot as part of my upgrade plan, i thought you'd meant something within the migration script itself to be extra safe! which may not be a bad idea...

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data_clear_experience_configs.py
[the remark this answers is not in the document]
adamsachs: yeah, i did want to verify that the OOB logic updates these fields appropriately. but may as well add them here with the 'right' values to be comprehensive.

src/fides/api/alembic/privacy_experience_config_defaults.yml
adamsachs: oh yeah of course, good call 😦 

i guess we can just filter out in the script, shouldn't be too bad 👍

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data_clear_experience_configs.py
pattisdr: most of these would have server defaults, maybe not name though -

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data_clear_experience_configs.py
```diff
@@ -0,0 +1,565 @@
...
+    """
+    _raw_experience_config_map = (
+        {}
+    )  # this will just hold config templates exactly as they are in the 'oob' file
+    _reconciled_experience_config_map = (
+        {}
+    )  # will hold experience configs 'reconciled' between existing data and 'oob' template data
+
+    # this is a bit of a hack! but the path to the file depends on whether we're executing the migration
+    # as part of app startup or as part of a 'manual' migration attempt
```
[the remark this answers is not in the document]
pattisdr: right got it

pattisdr:
@adamsachs heads up the multilanguage branch had a new migration merged to it, the locations one to start saving location groups

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data_clear_experience_configs.py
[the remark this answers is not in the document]
adamsachs: adding `allow_language_selection` and `dismissable` but i don't see `auto_detect_language` in the templates so i'll leave that populated by server defaults 👍

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data_clear_experience_configs.py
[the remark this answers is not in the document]
adamsachs: actually, turns out this is needed, there's no server default!

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data_clear_experience_configs.py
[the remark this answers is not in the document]
pattisdr: auto_detect_language is a new field that just got added last week! I'm waiting on Michael to get the actual default here, templates may be updated, not sure yet.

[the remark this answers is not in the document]
adamsachs: ok cool, yeah if templates are updated over in plus we'll just need to remember to copy over here and update in OSS for the migration 😬

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data_clear_experience_configs.py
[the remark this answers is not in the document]
adamsachs: this should be handled now, confirmed with some manual testing 👍

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data_clear_experience_configs.py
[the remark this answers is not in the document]
adamsachs: added in some taking some csv snapshots on disk for tables that we're touching 👍

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data_clear_experience_configs.py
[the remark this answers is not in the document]
adamsachs: ok, i think i've covered now 👍

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data.py
```diff
@@ -0,0 +1,651 @@
...
+}
+
+
+# as part of this upgrade, we're removing the following as region options across the app
+REGIONS_TO_REMOVE = {
+    "gb_eng",
+    "gb_sct",
+    "gb_wls",
+    "gb_nir",
+}
```
pattisdr, line 82: Also double-checked once more with Michael that it is okay to remove this, there's not specific privacy laws for these regions/countries within the UK 👍

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data.py
```diff
@@ -0,0 +1,651 @@
...
+
+
+# this ties a DB record ID to our logical identifier of the type of OOB experience.
+# the DB record ID is based on what we've defined in the associated config yml.
+class DefaultExperienceConfigTypes(Enum):
+    EEA_PRIVACY_CENTER = 1
+    EEA_MODAL_AND_BANNER = "pri-c8ff-78d6-4a02-850f-2c09dda-over-config"
+    EEA_TCF_OVERLAY = "d7c3ce0a-a3b3-43ff-bf6f-3d8-tcf-over-config"
+    US_MODAL = "pri-76b8-cc52-11ee-9eaa-8ef97a0-over-config"
+    US_PRIVACY_CENTER = "pri-27d4-cc53-11ee-9eaa-8ef97a04-pri-config"
```
pattisdr, line 39: Michael is also saying we should ship with a canadian experience out of the box, I don't know what that means yet, if we need CA regions on our existing experiences, etc.

adamsachs:
just a note: let's not merge this yet, it's failing on the fidesplus image at the point of trying to load the static yml since that gets packaged up with the fides python package. i need to consider what the best alternative is - either loading the static yml from the fidesplus codebase (don't love the idea of referring to a fidesplus artifact here in OSS) or just storing the template/config data as json within the migration script itself, to make it self-sufficient. as i type this out, i'm leaning pretty strongly toward the latter. 

maybe there's another option?

pattisdr:
latter option sounds good, no problem, I'm testing anyway alongside my followup migration # 3!

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data.py
```diff
@@ -0,0 +1,651 @@
...
+    def create_new_experience_config(experience_config):
+        """
+        Creates a new experience config based on the provided definition. Also creates corresponding history record.
+        """
+        # for config record:
+        # - `id` is the `id` grabbed from our config template record
+        # - `origin` is the `origin` grabbed from the config template
+        create_experience_config_query = text(
+            """
+            INSERT INTO privacyexperienceconfig (id, version, origin, component, name, disabled, is_default, allow_language_selection, dismissable, accept_button_label, banner_enabled, description, privacy_preferences_link_label, privacy_policy_link_label, privacy_policy_url, save_button_label, title, banner_description, banner_title, reject_button_label, acknowledge_button_label)
```
pattisdr, line 364: this is trivial, but my feeling is that we should generate the id for this record ourselves using generate_record_id(prefix) instead of the hardcoded one in the template, just to match out of the box behavior. 

We used to use a hardcoded one out of the box before we added experience config templates, but now that we have translations, we've introduced templates, and have stopped with the hardcoded id, only the templates themselves have the hardcoded id.

src/fides/api/alembic/privacy_experience_config_defaults.yml
```diff
@@ -0,0 +1,124 @@
+privacy_experience_configs:
+# US (non-GPP) banner and modal
+- id: pri-76b8-cc52-11ee-9eaa-8ef97a0-over-config
+  origin: pri-76b8-cc52-11ee-9eaa-8ef97a0-over
```
pattisdr, line 4: This would need to first originate on the plus side, but I'm wondering if we should change these autogenerated codes to not have "over" in them, that's a historical holdover

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data.py
```diff
@@ -0,0 +1,651 @@
...
+    """
+
+    def create_new_experience_config_templates(experience_configs):
+
+        for experience_config in experience_configs:
+            # for config template record:
+            # - `id` is the `origin` field from the config template record loaded from disk
+            create_experience_config_template_query = text(
+                """
+                INSERT INTO experienceconfigtemplate (id, component, name, disabled, allow_language_selection, dismissable, privacy_notice_keys, regions)
```
pattisdr, line 340: Let's just double check these translations are adding in plus on startup as we expect!  I'm currently just testing from OSS

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data.py
```diff
@@ -0,0 +1,651 @@
...
+                VALUES (:id, :component, :name, :disabled, :allow_language_selection, :dismissable, :privacy_notice_keys, :regions)
```
pattisdr, line 341: I need to get clear on what allow_language_selection and auto_detect_language should be out of the box on the templates.  I was told "true" for both. They are true at the db level and allow_language_selection is false in the template.  I think they should both be false in the template, because whether these values can be true is also a factor of whether they've paid for translations or the ability to add translations right?

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data.py
```diff
@@ -0,0 +1,651 @@
...
+        # one that is surfaced in the EU (and is associated with the "EU" notices).
+        # because of this, we'll copy over existing experience text/attributes to _both_
+        # of these corresponding new experience configs.
+        if eea_config_id := EEA_MAPPING.get(component):
+            configs.append(reconciled_experience_config_map.get(eea_config_id))
+        if us_config_id := US_MAPPING.get(component):
+            configs.append(reconciled_experience_config_map.get(us_config_id))
+        for config in configs:
+            if config:
+                config["accept_button_label"] = eec["accept_button_label"]
```
pattisdr, line 297: Have we considered automatically enabling/disabling based on their underlying data?

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data.py
pattisdr, line 364: And we're sure no one's created a privacy experience config?

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data.py
```diff
@@ -0,0 +1,651 @@
...
+        """
+        Delete all PrivacyExperience records.
+
+        These records will be effectively re-created as a result of the "out of the box" creation
+        of 'new-style' PrivacyExperienceConfig records post-migration.
+        """
+        bind.execute(
+            text(
+                """
+                DELETE FROM privacyexperience;
```
pattisdr, line 210: The previous schema migration drops the FK constraint on PrivacyPreferenceHistory.privacy_experience because we're not necessarily persisting those resources anymore, and we don't really need it with the combo of the user geography and the privacy experience config history id.  

That also allowed this action to be possible

I think we should probably delete this column now in the followup? Are there slowness risks though both for removing the FK and/or deleting a column from a large table?

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data.py
```diff
@@ -0,0 +1,651 @@
...
+
+# our new OOB US experience configs, keyed by their component type
+US_MAPPING = {
+    ComponentType.PrivacyCenter: DefaultExperienceConfigTypes.US_PRIVACY_CENTER,
+    ComponentType.Overlay: DefaultExperienceConfigTypes.US_MODAL,
+}
+
+# this map contains the biggest _assumption_ about the migration -
+# these are the notices that will be associated with our new OOB experience configs.
+# any existing notices besides these will remain "orphaned" post-migration.
```
pattisdr, line 63: Do we have any customers with other notices? Have we tested this at least doesn't break?

pattisdr (commented):
Looking good Adam, only a couple of actual bugs found, small, I'm mostly just rehashing some of the questions you've probably already answered by looking at the existing data.

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data.py
```diff
@@ -0,0 +1,651 @@
...
+
+def determine_needed_experience_configs(bind):
+    """
+    Our data model changes necessitate the creation of new Experience Configs. Now that "where" a Notice is served
+    is being moved from the "Notice" constructs to "Experiences", existing Experience Configs are not sufficient
+    to capture the full range of configuration on current Notices.
+    """
+
+    notice_query = """select id, name, regions, displayed_in_overlay, displayed_in_privacy_center, notice_key from privacynotice;"""
+    experience_config_query = """SELECT id, component, disabled, accept_button_label, banner_enabled, description, privacy_preferences_link_label, privacy_policy_link_label, privacy_policy_url, save_button_label, title, banner_description, banner_title, reject_button_label, acknowledge_button_label from privacyexperienceconfig;"""
```
pattisdr, line 242: To add robustness here, just _in case_ another privacy experience config exists - we could filter these to ones whose ids match one of the ids in here? For example, if another overlay had been created, it is not repeatable which overlay's details get translated to the new US Overlay and EEA overlay.  This would force us to prioritize the updated details on the OOB experience configs over the custom ones.

https://github.com/ethyca/fides/blob/c91d46c9145891c150e03d39ff0c7f4b5bc37d56/src/fides/data/privacy_notices/privacy_experience_config_defaults.yml#L4

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data.py
pattisdr: EDIT: OK my third follow-up migration now deletes this obsolete column from PrivacyPreferenceHistory and ServedNoticeHistory https://github.com/ethyca/fides/pull/4651/files.

We should still double check if the migration to remove the FK and/or delete the column is okay on these big tables - I think deleting the FK could lock the table. Part of the migration plan could be taking some downtime so this table is not being concurrently accessed.

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data.py
```diff
@@ -0,0 +1,651 @@
...
+        component_mapping = NOTICES_TO_EXPERIENCE_CONFIG.get(notice_key)
+        if component_mapping:
+            regions_to_add = set(existing_notice["regions"]).difference(
+                REGIONS_TO_REMOVE
+            )
+            if existing_notice["displayed_in_overlay"]:
+                experience_config = component_mapping.get(ComponentType.Overlay)
+                reconciled_experience_config_map[experience_config]["regions"].update(
+                    regions_to_add
+                )
```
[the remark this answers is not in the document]
adamsachs: yup great callout! that seems _generally_ true from the data sample we've seen so far, but there does seem to be at least one exception. i'd noted this as a "manual step" in the draft of my [release plan](https://ethyca.atlassian.net/wiki/spaces/EN/pages/2952298561/DRAFT+Multilang+release+plan+2.32.0#Manual-steps) (language is admittedly a bit vague there), i'll clarify more precisely what we need to look for and how it should be addressed.

even though we know this will lead to some necessary manual adjustments in a few cases, i still think the tradeoff we gain in simplicity/predictability is worth it!

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data.py
```diff
@@ -0,0 +1,651 @@
...
+    experience_config_query = """SELECT id, component, disabled, accept_button_label, banner_enabled, description, privacy_preferences_link_label, privacy_policy_link_label, privacy_policy_url, save_button_label, title, banner_description, banner_title, reject_button_label, acknowledge_button_label from privacyexperienceconfig;"""
+
+    existing_notices: ResultProxy = bind.execute(text(notice_query))
+    for existing_notice in existing_notices:
+        notice_key = existing_notice["notice_key"]
+        component_mapping = NOTICES_TO_EXPERIENCE_CONFIG.get(notice_key)
+        if component_mapping:
+            regions_to_add = set(existing_notice["regions"]).difference(
+                REGIONS_TO_REMOVE
+            )
```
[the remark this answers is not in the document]
adamsachs: ah ok good callout! i can fold in a bit more logic here to add `gb` if one of these regions is found 👍

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data.py
```diff
@@ -0,0 +1,651 @@
...
+        experience_config_type = DefaultExperienceConfigTypes(experience_config["id"])
+        _raw_experience_config_map[experience_config_type] = experience_config.copy()
+
+        # the reconciled experience config will have its regions populated by existing notices that are
+        # associated with this experience config. so in most cases, we start "fresh" with an empty set.
+        # the TCF experience is an exception - it is not linked to any notices, and therefore its
+        # regions are not derived from notices, so we hardcode its regions to the config default.
+        reconciled_regions = (
+            experience_config["regions"]
+            if experience_config_type == DefaultExperienceConfigTypes.EEA_TCF_OVERLAY
```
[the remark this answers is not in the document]
adamsachs: oh interesting, good callout. i think we can either try to base this on their existing TCF experience config, or if not then just call it out as a thing to watch out for/adjust post-migration. i'll followup on my end 👍

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data.py
```diff
@@ -0,0 +1,651 @@
...
+
+
+def generate_record_id(prefix):
+    return prefix + "_" + str(uuid.uuid4())
+
+
+# this ties a DB record ID to our logical identifier of the type of OOB experience.
+# the DB record ID is based on what we've defined in the associated config yml.
+class DefaultExperienceConfigTypes(Enum):
+    EEA_PRIVACY_CENTER = 1
```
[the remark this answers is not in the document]
adamsachs: agreed, let's get clarity from product?

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data.py
adamsachs: ok yeah i'm fine with that, i thought maybe the hardcoded id could provide a bit more of a "marker" but if you feel like consistency in how these are generated is better, then i'm good with that!

src/fides/api/alembic/privacy_experience_config_defaults.yml
adamsachs: heh fair point, i hadn't even noticed/payed much attention to that. i'm good with adjusting - want to do that and give me a heads up? i'll need to be regenerating these templates anyway and embedding them as JSON within the migration itself.

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data.py
adamsachs: yup same here! will verify that today 👍

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data.py
adamsachs: yeah, so is this then something we should mark as a manual followup?

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data.py
adamsachs: yup, that was my expectation! thank you for verifying on your end and glad it's holding up under scrutiny 👍 

the only notices that were created in the sample of customer data we got were "test" notices that were not actually in use, so this approach seemed perfectly appropriate to me.

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data.py
adamsachs: yeah that's a good callout, i like it! so basically - ensure we only migrate experience configs we know about. any other experience configs will need to be handled manually.

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data.py
adamsachs: ok got it, either of those approaches seems viable to me - allotting for some downtime for the migration may be prudent regardless of this particular point. let's do some more exploration and plan to finalize 👍

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data.py
pattisdr: a marker to indicate we manually created in a migration? sure that's fine we can leave this as -is

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data.py
pattisdr: OK in https://github.com/ethyca/fidesplus/pull/1334 a new OOB Canadian experience config has been added

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data.py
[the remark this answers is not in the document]
pattisdr: OK confirmed in-person, won't ship with EEA_PRIVACY_CENTER out of the box, but may need to add if customers have checked `displayed_in_privacy_center` manually.

src/fides/api/alembic/privacy_experience_config_defaults.yml
pattisdr: OK i've updated the ids in https://github.com/ethyca/fidesplus/pull/1334

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data.py
pattisdr: OK updates from in-person meeting:

`allow_language_selection` is now false in the db (was already false in the template).  `Auto_detect_language` has been given a true value in the template here https://github.com/ethyca/fidesplus/pull/1334.

So we might add auto_detect_language to the regenerated templates we use here, and add that field to this query?

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data.py
pattisdr: yes I think that would make sense

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data.py
pattisdr: The addition of an OOB canadian config is a little interesting.  I assume updates to the original EEA overlay should be transferred there?

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data.py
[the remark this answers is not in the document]
adamsachs: confirmed we do have at least one instance of this, noting as a manual step 👍

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data.py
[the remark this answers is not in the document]
adamsachs: noting this as a callout for adjustment post-migration, since it seems unlikely that anyone had changed this currently - not an option that was exposed via UI, only could be done directly via API.

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data.py
adamsachs: yup that was the idea! ok i'll leave it in then 👍

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data.py
adamsachs: noted "Determine whether any changes are desired for the accept_button_label on experience configs" as a bullet point in the manual followups

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data.py
adamsachs: yes i think that's right! and we should hardcode its regions i think? 🤷

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data.py
pattisdr: I wasn't talking about the `accept_button_label` specifically here I was just talking about adding `disabling` here as a field we bring over in addition

adamsachs: ohhh sheesh that makes more sense now 😅 

i'd already had that as a followup, so we should be good 👍 
> Enable appropriate experience configs - all will be disabled by default!

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data.py
pattisdr: yes we could hardcode but we'd also want to make sure we didn't put canadian regions on the new EEA banner and modal (even if they are present on the original EEA overlay).  A manual migration step might be to call out which customers need to have the canadian config flipped on, depending on whether they had added canadian regions to their EEA overlay?

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data.py
```diff
@@ -0,0 +1,827 @@
...
+        # the reconciled experience config will have its regions populated by existing notices that are
+        # associated with this experience config. so in most cases, we start "fresh" with an empty set.
+        # the TCF experience and Canada banner modal experience are exceptions:
+        # their regions are not derived from notices (TCF is not linked to notices; Canada experience config is net-new),
+        # so we hardcode their regions to the config defaults.
+        reconciled_regions = (
+            experience_config["regions"]
+            if experience_config_type == DefaultExperienceConfigTypes.EEA_TCF_OVERLAY
+            or experience_config_type
+            == DefaultExperienceConfigTypes.CANADA_BANNER_MODAL
```
pattisdr, line 307: At the end of this we have an EEA Banner and Modal with EEA regions and canadian regions and 4 notices, and a Canada Banner and Modal with canadian regions and no notices.

In my opinion the end state would not have canadian regions on the EEA banner and modal, and the Canada Banner and Modal would have the four notices.

As-is both cannot be enabled at once due to the overlapping regions.

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data.py
```diff
@@ -0,0 +1,827 @@
...
+            },
+        )
+
+        # backfill language column in all experience config history records
+        update_experience_config_history_language = text(
+            """
+            UPDATE privacyexperienceconfighistory
+            SET language = 'en'
+        """
+        )
```
pattisdr, line 571: Note to self, this will be able to be removed after my changes which make language nulllable on the historical records

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data.py
```diff
@@ -0,0 +1,827 @@
...
+                "language": "en",
+                "original_notice_id": res["id"],
+            },
+        )
+
+    # Add a FK from the Notice History back to the Notice Translation
+    link_notice_history_to_translation_id = text(
+        """
+        UPDATE privacynoticehistory
+        SET translation_id = noticetranslation.id, title = privacynoticehistory.name, language = 'en'
```
pattisdr, line 766: So for french customers, they will have to go in and update the language where we've said this is "en".  However, "en" is the identifier for all intents and purposes.  I think they would have to add their french translations, and delete the other translations that are also in french but labeled "english", unless we update this at the db level.

pattisdr: Note this is potentially inaccurate for some customers on historical records.  We don't have to backfill it anymore, but we'd likely want it backfilled for the notice history in use at least, so this is fine.

src/fides/data/location/locations.yml
```diff
@@ -359,16 +359,19 @@ locations:
   name: Ireland
   belongs_to: [eea]
   regulation: [gdpr]
+  default_selected: True
```
adamsachs, line 362: this field will support some new OOB location "seeding" that'll be done over in plus

src/fides/data/location/locations.yml
```diff
@@ -882,227 +913,280 @@ locations:
...
   continent: North America
+  default_selected: True
 - id: us_sc
   name: South Carolina
   belongs_to: [us]
   continent: North America
+  default_selected: True
 - id: us_sd
-  name: San Diego
+  name: South Dakota
```
adamsachs, line 1132: cleaning up an (embarrassing!) typo from earlier 😆

src/fides/data/location/locations.yml
```diff
@@ -882,227 +913,280 @@ locations:
...
   regulation: [vcdpa]
+  default_selected: True
 - id: us_vi
   name: United States Virgin Islands
   belongs_to: [caribbean, us]
   continent: North America
+  default_selected: True
 - id: us_vt
-  name: Vermon
+  name: Vermont
```
adamsachs, line 1166: another typo cleanup

src/fides/api/models/location_regulation_selections.py
```diff
@@ -382,25 +383,28 @@ def load_regulations() -> Dict[str, LocationRegulationBase]:
...
+locations_by_id: Dict[str, Location] = (
+    load_locations()
+)  # should only be accessed for read-only access
+location_groups: Dict[str, LocationGroup] = (
+    load_location_groups()
+)  # should only be accessed for read-only access
+location_group_to_location: Dict[str, Set[str]] = (
+    load_location_group_to_location()
+)  # should only be accessed for read-only access
+default_selected_locations: Set[str] = {
```
adamsachs, line 395: will be used for default location seeding done on startup, over in plus

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data.py
pattisdr: oh sorry yes I was not clear at all.

pattisdr (approved):
looks good to me I'll continue to test this further alongside the 3rd followup migration in https://github.com/ethyca/fides/pull/4651

src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data.py
```diff
@@ -0,0 +1,858 @@
...
+    notice_query = """select id, name, regions, displayed_in_overlay, displayed_in_privacy_center, notice_key from privacynotice;"""
+    experience_config_query = """SELECT id, component, disabled, accept_button_label, banner_enabled, description, privacy_preferences_link_label, privacy_policy_link_label, privacy_policy_url, save_button_label, title, banner_description, banner_title, reject_button_label, acknowledge_button_label from privacyexperienceconfig;"""
+
+    existing_notices: ResultProxy = bind.execute(text(notice_query))
+    for existing_notice in existing_notices:
+        notice_key = existing_notice["notice_key"]
+        component_mappings = NOTICES_TO_EXPERIENCE_CONFIG.get(notice_key, [])
+        if component_mappings:
+            for component_mapping in component_mappings:
+                regions_to_add = set(existing_notice["regions"]).copy()
```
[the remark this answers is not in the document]
adamsachs: whoops - that was an oversight :(

i can think about how to best adjust, unless you've got something on the top of mind. i think it'll just require a bit more hardcoding.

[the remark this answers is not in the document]
pattisdr: I have a draft I'll push up on my PR, you might not like it :)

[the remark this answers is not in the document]
adamsachs: i see it, that's fine by me! :) thanks for cleaning up my mess

PR #4631, opened 2024-02-26
Localize all strings in FidesJS banner & modal based on the user's locale and experience configuration

NevilleS:
### Description Of Changes

This (should!) fully localize the FidesJS banner & modal components. There's some additional follow-up work to do for testing and sending the right history IDs based on the user locale, but this should cover most of the UI work!

### Code Changes

- [X] Update all fides.js components to use `i18n.t(...)` translations instead of static strings or `experience` properties
- [X] Add `messageExists` helper function for cases where we want to have optional labels
- [X] Add E2E tests for optional GPC UI components
- [X] Add E2E tests for optional privacy policy link components
- [X] Add E2E tests for optional banner title/description overrides
- [X] Add static strings for `es` locale
- [X] Disable automatic locale detection if the `privacy-experience` API has it disabled

Remaining items that I'll roll into the next PR for PROD-1597:
- [ ] Add E2E tests that cover the special case where an experience is configured with a translation, but there is not a matching translation for all notices
- [ ] Run Typescript linter on fides-js (there are type errors not being surfaced by eslint)
- [ ] Fix other TODOs for PROD-1597 👍 

Follow-up on (PROD-1744) to fixing saving preferences and notices served:
- [ ] Ensure the correct `privacy_experience_config_history_id` and `privacy_notice_history_id` values are used for reporting based on the selected locale
- [ ] Re-enable disabled tests (marked with `TODO (PROD-1744)`)

Follow-up (PROD-1598):
- [ ] Add ability for user to manually select from available locales in banner & modal

### Pre-Merge Checklist

* [X] All CI Pipelines Succeeded
* Documentation:
  * [ ] documentation complete, [PR opened in fidesdocs](https://github.com/ethyca/fidesdocs/pulls)
  * [ ] documentation [issue created in fidesdocs](https://github.com/ethyca/fidesdocs/issues/new/choose)
* [X] Issue Requirements are Met
* [X] Relevant Follow-Up Issues Created
* [ ] Update `CHANGELOG.md`
* [ ] For API changes, the [Postman collection](https://github.com/ethyca/fides/blob/main/docs/fides/docs/development/postman/Fides.postman_collection.json) has been updated

PR #4639, opened 2024-02-28
Update existing FidesJS E2E tests with multilang translations and add initial i18n tests

NevilleS:
### Description Of Changes

This is the next step in adding localization to all the privacy experiences, and can be merged in ahead of the rest of the effort in https://github.com/ethyca/fides/pull/4631

Pulling this out into an initial PR so that the rest of the UI work is better isolated from some fairly meaty search & replace shenanigans here!

### Code Changes

- [X] Update E2E tests to include multiple translations for both experiences & notices
- [X] Add initial (failing!) E2E tests to cover multiple translations for experiences & notices
- [X] Mark failing tests as 

### Steps to Confirm

* [X] n/a - if this passes all the CI checks, we're golden 👍 

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`


clients/privacy-center/cypress/e2e/consent-i18n.cy.ts
```diff
@@ -0,0 +1,250 @@
...
+      description: "Este sitio web usa cookies y servicios analíticos",
+    },
+    {
+      title: "Essential",
+      description: "Este sitio web utiliza cookies y servicios esenciales",
+    },
+  ];
+
+  // Setup a test case with the given params
+  const visitDemoWithI18n = (
```
eastandwestwind, line 103: 💯 I like how you've made this flexible and we can pass in fixture / options

clients/privacy-center/cypress/e2e/consent-i18n.cy.ts
```diff
@@ -0,0 +1,250 @@
...
+    });
+  };
+
+  describe("when auto_detect_language is true", () => {
+    describe(`when browser language matches default locale (${ENGLISH_LOCALE})`, () => {
+      it("localizes banner_and_modal components in the correct locale", () => {
+        visitDemoWithI18n(ENGLISH_LOCALE, "experience_banner_modal.json");
+        testBannerLocalization(ENGLISH_BANNER);
+        openAndTestModalLocalization(ENGLISH_MODAL);
+        testModalNoticesLocalization(ENGLISH_NOTICES);
```
eastandwestwind, line 183: This reads nicely 👍

clients/privacy-center/cypress/e2e/consent.cy.ts
```diff
@@ -283,7 +283,8 @@ describe("Consent settings", () => {
       });
     });
 
-    it("reflects their choices using fides.js", () => {
+    // TODO (PROD-1597): re-enable this test; it's failing because overlay is not disabled!
+    it.skip("reflects their choices using fides.js", () => {
```
eastandwestwind, line 287: Was this failing before or a new failure?

clients/privacy-center/cypress/fixtures/consent/experience_gpp.json
```diff
@@ -16,27 +16,44 @@
             "language": "en",
             "accept_button_label": "Opt in to all",
             "acknowledge_button_label": "OK",
-            "banner_description": "We use cookies and similar methods to recognize visitors and remember their preferences. We may also use them to measure ad campaign effectiveness, target ads, and analyze site traffic. Depending on your location, you may opt-in or opt out of the use of these technologies.",
-            "banner_title": "Manage your consent preferences",
+            "banner_description": "[banner] We use cookies and similar methods to recognize visitors and remember their preferences. We may also use them to measure ad campaign effectiveness, target ads, and analyze site traffic. Depending on your location, you may opt-in or opt out of the use of these technologies.",
```
eastandwestwind, line 19: what is this [banner] signifier for?

eastandwestwind (approved):
Overall looks great! Thanks for adding all the additional test data / utils with our new multilang structure in mind, as well as ticketing / commenting TODOs where appropriate.

Left a couple comments, nothing blocking.

clients/privacy-center/cypress/e2e/consent.cy.ts
NevilleS: I think it's been there for awhile, I suspect it probably fails in CI every now and then. I noticed more when I was running these tests locally with `it.only` and checking things out - I think there's a race condition here where, if the test runs fast enough, it finishes assertions before `fides.js` finishes initializing and starts calling the privacy experience API, etc.

clients/privacy-center/cypress/fixtures/consent/experience_gpp.json
NevilleS: Just to be sure that it's loading from `banner_title` instead of `title`, since I wanted to keep the strings consistently translated and not using fake strings as much (e.g. "This is a banner title!")

🤷 maybe I should just actually make a more obvious difference in test titles & description and do a bit of Google Translate?

clients/privacy-center/cypress/e2e/consent.cy.ts
eastandwestwind: ah wow, this is good to know. Sounds like we need to be more explicit about waiting for a FidesInitialized event or something before proceeding with test assertions

clients/privacy-center/cypress/fixtures/consent/experience_gpp.json
eastandwestwind: oh I see! My preference is to have the text actually be different between the two, but this is fine for now.

clients/privacy-center/cypress/e2e/consent.cy.ts
NevilleS: These tests are kinda the opposite of that - I think they were written in the era where FidesJS was purely static, so they don't worry about it loading a client-side experience and then changing to new notices 👍 

The gotcha is that the tests _think_ they set `OVERLAY_ENABLED: false` but I think there's a `cy.reload()` in there that causes the state to be lost, etc...

If I can't find a simple fix I'll carve out a follow-up tick

PR #4644, opened 2024-02-29
update drop statement depending on db version

adamsachs:
Closes n/a

### Description Of Changes

a little while back, we [adjusted](https://github.com/ethyca/fides/commit/7bef8b4ea8761049b1148654370e96e0dbe7a7ea) our db migration that creates the `array_accum` aggregate function to differ its syntax depending on db version, due to a change in postgres typing in version 14.

what we forgot to do was update a corresponding `DROP AGGREGATE IF EXISTS` statement to have the appropriate syntax for the version in question! the practical effect here is _very, very_ narrow - the only real scenario this would come up in is when the function actually already exists when we run this migration. that basically only happens if we've downgraded and are re-upgrading, OR if we are invoking the `db/reset` API endpoint/command!

we did hit this on a cloud (internal) env when trying to invoke that API call to reset the app's db. i think the proposed fix here will allow `db/reset` to run successfully on postgres >= 14 🤞 

### Code Changes

* [x] update migration to conditionally change the `DROP AGGREGATE IF EXISTS` command syntax depending on postgres version, to align with the corresponding `CREATE AGGREGATE` commands we run!

### Steps to Confirm

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### DIFF
--- a/src/fides/api/alembic/migrations/versions/fd52d5f08c17_system_dictionary_data_migration.py
+++ b/src/fides/api/alembic/migrations/versions/fd52d5f08c17_system_dictionary_data_migration.py
@@ -5,6 +5,7 @@ Revises: 39397820a0d5
 Create Date: 2023-08-04 14:14:32.421414
 
 """
+
 import json
 from typing import Any, Dict, List, Optional, Tuple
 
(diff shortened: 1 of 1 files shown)


pattisdr (approved):
ah okay! makes sense

PR #4650, opened 2024-03-01
Preview components in admin UI

eastandwestwind:
Closes https://ethyca.atlassian.net/browse/PROD-1596, https://ethyca.atlassian.net/browse/PROD-1460, 
https://ethyca.atlassian.net/browse/PROD-1803

### Description Of Changes

Adds a live preview to the experience config form.

### Steps to Confirm

With admin-ui and privacy-center both running, navigate to `/consent/privacy-experience`.

Edit an existing privacy experience config:
* [ ] Preview should show with existing privacy notices
* [ ] For privacy centers, banner should show; for modal, modal should show; for banner and modal, both should show;
* [ ] When privacy notices are added or removed, preview should update accordingly
* [ ] Edit English translation
* [ ] When text is changed, preview should update accordingly 

Create a new experience config:
* [ ] Before adding a privacy notice, preview area should show alert telling you to add a privacy notice to see the preview
* [ ] Appropriate component preview should show once component is selected and privacy notices are added
* [ ] Before text is added, preview should show with default text

### Pre-Merge Checklist

* [ ] Update `CHANGELOG.md`


clients/admin-ui/src/features/privacy-experience/Preview.tsx
```diff
@@ -0,0 +1,260 @@
...
+        result.error,
+        "A problem occurred while fetching privacy notice data.  Some notices may not display correctly on the preview."
+      );
+      toast({ status: "error", description: errorMsg });
+    }
+    const { data } = await getPrivacyNoticeByIdTrigger(id);
+    return data;
+  };
+
+  useEffect(() => {
```
jpople, line 77: Technically according to best practices, this shouldn't be in a useEffect-- it only depends on a single React state variable, so ideally we'd: 
* track the notices we care about in state in the component above this
* whenever the ids are changed in the form, change it via an event handler
* pass it as a prop into this component
That said, I tried doing it that way and it introduced absolutely horrendous lag, so it lives here.

eastandwestwind: @jpople just so I'm aware, what's the downside of using `useEffect` here when it depends on a single React state variable?

jpople: @eastandwestwind You know what, I think I'm mistaken here-- I was misremembering the principles from [You Might Not Need an Effect](https://react.dev/learn/you-might-not-need-an-effect).  The fact that this is fetching from the API means it's legitimate here, what I was thinking of is that it's better not to use it to *deterministically* calculate state from other state inside the app.

eastandwestwind:
@jpople looking good! Just want to confirm this is a product decision to remove interactivity and to have banner and modal show statically like this?
Screenshot 2024-03-11 at 2 08 59 PM

eastandwestwind:
Additionally, it looks like that bug I originally saw but couldn't reproduce is showing up again with mobile preview mode cutting off the last button:

 Screenshot 2024-03-11 at 2 11 08 PM

eastandwestwind:
One more thing I noticed is that upon editing a translation, the save button still says "Add translation" which may be a bit misleading as we're only editing the current translation:

Screenshot 2024-03-11 at 2 39 45 PM

jpople:
> Just want to confirm this is a product decision to remove interactivity and to have banner and modal show statically like this?

@eastandwestwind Yeah, talked this over with @NevilleS-- this approach let us sidestep both the bug you were having before where the banner would flicker in and out sometimes and also a new bug that popped up with Neville's i18n changes which made it pretty much impossible to view the modal anyway when interaction was working.

PR #4659, opened 2024-03-05
Update location picker to respect selected locations

jpople:
Closes PROD-1707

### Description Of Changes

Changes the location picker in the new privacy experience form so the only options shown are locations and location groups  that are configured.

### Code Changes

* [ ] Change selector on privacy experience config form to filter based on which location options are selected
* [ ] Update `PrivacyNoticeRegion` enum to include new locations and location groups
* [ ] Re-generate API types
* [ ] Small changes to Properties pages to handle changes to schema

### Steps to Confirm

* [ ] Visit privacy experience config form
* [ ] Click "Add location", note options in the select dropdown
* [ ] Navigate to "Locations"
* [ ] Enable one or more new locations
* [ ] Return to experience config form
* [ ] Newly added option(s) should show up in "Locations" dropdown
* [ ] In "Locations", disable a location that isn't currently assigned to an experience
* [ ] In experience config form, that location should no longer show up in "Locations" dropdown

### Pre-Merge Checklist

* [ ] Update `CHANGELOG.md`


clients/admin-ui/src/features/locations/transformations.ts
```diff
@@ -73,21 +71,8 @@ export const groupByBelongsTo = (
...
-  if (isLocation(location)) {
-    return !!location.regulation?.length;
-  }
-  const locationsInGroup = locations.filter((l) =>
-    l.belongs_to?.includes(location.id)
-  );
-  return locationsInGroup.some((l) => l.regulation?.length);
-};
+export const isRegulated = (location: Location) =>
+  !!location.regulation?.length;
```
jpople, line 75: This utility function previously relied on the property differences between Location and LocationGroup to determine whether something was a location or a group, but now the schemas are the same.  That said, it only gets called in one place (`LocationPickerCard.tsx`) and it's on an array of definitely-Locations so we don't actually need to.

clients/admin-ui/src/types/api/models/Property.ts
```diff
@@ -2,11 +2,13 @@
...
+/**
+ * A base template for all other Fides Schemas to inherit from.
+ */
 export type Property = {
-  key: string;
   name: string;
-  type: string;
-  experiences?: Array<string>;
+  type: PropertyType;
+  key?: string;
```
Kelsey-Ethyca, line 13: should this really be optional?

Kelsey-Ethyca (approved):
UAT ✅ 
Uploading Screen Recording 2024-03-07 at 3.12.52 PM.mov…

clients/admin-ui/src/types/api/models/Property.ts
jpople: The key is auto-generated on the backend on creation so it's optional to be able to create with just a name and type-- every property has a key after it's created.

PR #4663, opened 2024-03-05
Update FidesJS to handle "mismatched" translations between notices within an experience

NevilleS:
WIP for PROD-1597, but doesn't finish everything up quite yet...!
Closes PROD-1744 though!

### Description Of Changes

This PR is *mostly* about updating the FidesJS banner & modal components to handle the special case where the notice might not have a translation that the experience does. I also ended up scooping in a few QOL improvements and additional tests along the way since this required a decent amount of refactoring, because I needed to also fix the usage of the "notice served" and "privacy preferences" APIs to include the appropriate history ID from the active translation, which required retooling the UI components a decent amount.

The other notable change here is to abandon my initial idea to extract *all* the notice translations into the global `i18n` message catalog and differentiate them by key, like `i18n.t("exp.notices.pri_1234.title")`. I thought this would be a good idea, but in practice it ended up being clumsy - so instead, I don't use the `i18n` catalog for notices themselves and instead have an alternate set of helpers to select the "best" translation for a notice and then render the title/description inline.

### Steps to Confirm

* [ ] Configure a privacy experience in a language (e.g. Spanish) and then delete the Spanish language translation for one of the notices linked to that experience
* [ ] Test that the CMP UI shows the experience in Spanish but the notice in English
* [ ] Check the API calls to confirm that the history IDs in the "notices served" and "privacy preferences" calls match the translation IDs for Spanish/English/etc.

This mismatch ends up looking like this:
image

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`


clients/fides-js/src/lib/initialize.ts
```diff
@@ -131,22 +131,34 @@ const automaticallyApplyGPCPreferences = ({
         gpcApplied = true;
         return new SaveConsentPreference(
           notice,
-          transformConsentToFidesUserPreference(false, notice.consent_mechanism)
+          transformConsentToFidesUserPreference(
+            false,
+            notice.consent_mechanism
+          ),
+          // TODO (PROD-1597): ...figure out the right translation for these preferences, somehow?
```
NevilleS, line 138: TODO: This is the major outstanding item on this PR; how we handle GPC notices needs to be carved out into a separate PR, so this just needs a bit of tidying to be _stable_ 👍

clients/fides-js/src/lib/consent-types.ts
```diff
@@ -199,9 +199,16 @@ export class SaveConsentPreference {
...
+
+  constructor(
+    notice: PrivacyNotice,
+    consentPreference: UserConsentPreference,
+    noticeHistoryId?: string
+  ) {
     this.notice = notice;
     this.consentPreference = consentPreference;
+    this.noticeHistoryId = noticeHistoryId;
   }
```
Kelsey-Ethyca, line 212: I know noticeHistoryId is optional. You have though updated everywhere SaveConsentPreference is used expect on `clients/fides-js/src/lib/cookie.ts` in `updateCookieFromNoticePreferences` so putting that here for your consideration.

clients/fides-js/src/lib/consent-types.ts
```diff
@@ -700,10 +711,14 @@ export type RecordConsentServedRequest = {
...
   tcf_system_legitimate_interests?: Array<string>;
+  /**
+   * @deprecated has no effect; use privacy_experience_config_history_id instead!
+   */
   privacy_experience_id?: string;
+  privacy_experience_config_history_id?: string;
   user_geography?: string;
   acknowledge_mode?: boolean;
-  serving_component: ServingComponent;
+  serving_component: string; // NOTE: uses a generic string instead of an enum
```
Kelsey-Ethyca, line 721: Why?

clients/fides-js/__tests__/lib/i18n/i18n-utils.test.ts
```diff
@@ -490,14 +603,24 @@ describe("i18n module", () => {
...
+      it("returns the currently active locale", () => {
+        expect(testI18n.locale).toEqual("en");
+        testI18n.load("zz", { "test.greeting": "Zalloz, Jest!" });
+        expect(testI18n.locale).toEqual("en");
+        testI18n.activate("zz");
+        expect(testI18n.locale).toEqual("zz");
+      });
     });
   });
 });
```
Kelsey-Ethyca, line 626: Nice! I will test against all these cases. Let me know if there are others I should test for UAT @NevilleS

clients/fides-js/src/components/notices/NoticeOverlay.tsx
```diff
@@ -1,33 +1,49 @@
...
+} from "../../lib/i18n";
+import { updateConsentPreferences } from "../../lib/preferences";
 import { transformConsentToFidesUserPreference } from "../../lib/shared-consent-utils";
+import ConsentBanner from "../ConsentBanner";
+import { NoticeConsentButtons } from "../ConsentButtons";
+import Overlay from "../Overlay";
+import PrivacyPolicyLink from "../PrivacyPolicyLink";
+import { OverlayProps } from "../types";
+import { NoticeToggleProps, NoticeToggles } from "./NoticeToggles";
+
```
Kelsey-Ethyca, line 37: nit: update the import paths to be absolute `~`

clients/fides-js/src/components/notices/NoticeOverlay.tsx
```diff
@@ -53,37 +71,102 @@ const NoticeOverlay: FunctionComponent<OverlayProps> = ({
...
+      (experience.privacy_notices || []).map((notice) => {
+        const bestTranslation = selectBestNoticeTranslation(i18n, notice);
+        return { notice, bestTranslation };
+      }),
+    [experience.privacy_notices, i18n]
+  );
+
   const [draftEnabledNoticeKeys, setDraftEnabledNoticeKeys] = useState<
-    Array<PrivacyNotice["notice_key"]>
+    Array<string>
```
Kelsey-Ethyca, line 112: is there a "safer" way to type this?

clients/fides-js/src/lib/initialize.ts
NevilleS: Done!

clients/fides-js/src/lib/consent-types.ts
NevilleS: That's a good suggestion, thanks... I was heavily utilizing Typescript for this kind of refactor so I made these IDs optional in some places (e.g. at the final API surface) but was trying to enforce them most of the time.

I just tried making it required but immediately ran into an issue in that I can't really _guarantee_ that a translation will be found by `selectBestNoticeTranslation` so even `NoticeOverlay.tsx:168` can fail to find a valid history ID when creating a new `SaveConsentPreference`... so I need to just let it be optional in the types 👎

clients/fides-js/src/lib/consent-types.ts
NevilleS: Fun fact: Typescript enums aren't actually strings. They compile to Javascript objects at runtime which is surprising and therefore can't be used directly when serializing to an API call, etc. I don't like that at all... we'd be better served with union types for these!

Hard to explain succinctly but happily would dogpile with anyone, I definitely went into the Typescript docs at one point when playing with these recently: https://www.typescriptlang.org/docs/handbook/enums.html

clients/fides-js/__tests__/lib/i18n/i18n-utils.test.ts
NevilleS: Nothing super specific... I think we really just should do a ton of exploratory testing around things and look to see if we can ever get a untranslated string to appear 👍

clients/fides-js/src/components/notices/NoticeOverlay.tsx
NevilleS: I'd definitely like to do that but I don't remember if there's a quirk in the FidesJS codebase for why we don't do this _anywhere_.

clients/fides-js/src/components/notices/NoticeOverlay.tsx
NevilleS: It's definitely a subjective one... in general, these keys are from a JSON API and therefore they are strings. We could do some clever string-typing to make them a real string (e.g. not empty string) and Typescript is pretty good at this!

I didn't like that we were doing the type "algebra" here and fetching the `PrivacyNotice["notice_key"]` type indirectly, I find that kind of thing really increases the cognitive burden of reading the code. The `PrivacyNotice["notice_key"]` type is a `string` after all, so this just showed that indirectly...

If I was being really clever I might compromise with this:
```typescript
export type PrivacyNoticeKey = PrivacyNotice["notice_key"];
```

That's equivalent (`PrivacyNoticeKey` is a `string`) but is a bit more legible, maybe? I dunno.

NevilleS:
OK @Kelsey-Ethyca: updated this with fixes for tests and a bit more polish around GPC. I think this is good to go

Kelsey-Ethyca (approved):

PR #4668, opened 2024-03-07
Update i18n.load() to replace existing messages with newer ones

NevilleS:
### Description Of Changes

Quick update to our `i18n` module to fix an issue that @jpople found while testing it out! Basically we weren't allowing you to update a particular message translation more than once, which would normally be fine, but breaks when we repeatedly initialize FidesJS in preview mode 👍

### Code Changes

- [X] Update `i18n.load` to override existing messages with new ones

### DIFF
--- a/clients/fides-js/src/lib/i18n/i18n-utils.ts
+++ b/clients/fides-js/src/lib/i18n/i18n-utils.ts
@@ -338,7 +338,7 @@ export function setupI18n(): I18n {
     },
 
     load: (locale: Locale, messages: Messages): void => {
-      allMessages[locale] = { ...messages, ...allMessages[locale] };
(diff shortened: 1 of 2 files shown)


clients/fides-js/src/lib/i18n/i18n-utils.ts
```diff
@@ -338,7 +338,7 @@ export function setupI18n(): I18n {
     },
 
     load: (locale: Locale, messages: Messages): void => {
-      allMessages[locale] = { ...messages, ...allMessages[locale] };
+      allMessages[locale] = { ...allMessages[locale], ...messages };
```
NevilleS, line 341: I will *never* be able to remember the precedence of the object spread operator.

jpople (approved):
💯

PR #4673, opened 2024-03-08
UX for experience config form for TCF

jpople:
Closes PROD-1686

### Description Of Changes

* hide "Experience type" field and privacy notice selector when editing TCF experience
* hide preview, instead show placeholder with notice that preview is now shown for TCF

### Steps to Confirm

* [ ] Navigate to privacy experience table
* [ ] Edit TCF experience
* [ ] Should see no experience type field, nor a list for notices
* [ ] Should see no preview
* [ ] Should see "no preview for TCF" notice in preview box

### Pre-Merge Checklist

### DIFF
--- a/clients/admin-ui/src/features/privacy-experience/Preview.tsx
+++ b/clients/admin-ui/src/features/privacy-experience/Preview.tsx
@@ -39,45 +39,31 @@ declare global {
   }
 }
 
-const NoPreviewNotice = () => (
-  <Flex
-    bgColor="white"
-    borderRadius="md"
-    p={6}
-    boxShadow="md"
-    direction="column"
-    align="center"
-    gap="2"
-    maxW="512px"
-  >
-    <Text fontSize="lg" fontWeight="500" align="center">
-      No privacy notices added
-    </Text>
-    <Text color="gray.500" align="center">
-      To view a preview of this experience, add a privacy notice under
-      &quot;Privacy Notices&quot; to the left.
-    </Text>
-  </Flex>
-);
-
-const TCFExperienceNotice = () => (
-  <Flex
-    bgColor="white"
-    borderRadius="md"
-    p={6}
-    boxShadow="md"
-    direction="column"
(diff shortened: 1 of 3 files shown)


clients/admin-ui/src/features/privacy-experience/Preview.tsx
```diff
@@ -60,6 +60,27 @@ const NoPreviewNotice = () => (
...
+    gap="2"
+    maxW="512px"
+  >
+    <Text fontSize="lg" fontWeight="500" align="center">
+      TCF preview not available
+    </Text>
+    <Text color="gray.500" align="center">
+      There is no preview available for TCF. You can edit the available settings
+      and languages to the left.
+    </Text>
```
eastandwestwind, line 80: can we extract / share a component with https://github.com/ethyca/fides/pull/4650/files#diff-7344022362ed6489650e4804f41c1c0cb81687fe5396c525ce66a404a3a12cccR53-R60 for preview not available?

clients/admin-ui/src/features/privacy-experience/preview/helpers.ts
```diff
@@ -71,7 +71,7 @@ export const buildBaseConfig = (
       is_default: true,
       dismissable: experienceConfig.dismissable,
       allow_language_selection: true,
-      auto_detect_language: true,
+      auto_detect_language: false,
```
eastandwestwind, line 74: does this need to be `true` at all for the non-TCF preview?

eastandwestwind (commented):
small nit / clarifying question, otherwise looking good!

clients/admin-ui/src/features/privacy-experience/preview/helpers.ts
jpople: No, we override the language that should be shown in the preview with the `fides_locale` option (in the base preview branch, not here) so this setting isn't relevant-- will revert this change.

PR #4674, opened 2024-03-08
Conditionally Enabling Migrated Experience Configs

pattisdr:
Closes #PROD-1797

### Description Of Changes

In the multitranslation data migration, we originally were disabling the new Experience Configs in the data migration as the safer default and added a manual step to enable them afterwards. However, we've learned it's preferable to ship with a preference towards enabling them if they can be enabled.

### Code Changes

* [ ] Strategy is to enable any Experience Configs if _any_ (not all) of the attached notices are enabled 
* [ ] Since TCF doesn't have notices, copying over the enabled flag of the original Experience Config.  This is likely to be enabled by default.  However, other TCF flags are also required in addition for TCF to work. These TCF Experience Configs are also suppressed entirely without these extra TCF enabled flags too, so for many customers, this won't change anything.

### Steps to Confirm

* [ ] Pre-migration on main, enable the "essential" notice, leaving the data sales and sharing notice disabled
* [ ] Post migration, any Experience Configs w/ the Essential notice (many of them) will be enabled, but the US Modal, for example, will still be disabled because its only notice is disabled
* [ ] The TCF Config will be enabled if it was enabled prior to migration

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### DIFF
--- a/src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data.py
+++ b/src/fides/api/alembic/migrations/versions/14acee6f5459_translation_data.py
@@ -440,7 +440,7 @@ def determine_needed_experience_configs(bind):
     to capture the full range of configuration on current Notices.
     """
 
-    notice_query = """select id, name, regions, displayed_in_overlay, displayed_in_privacy_center, notice_key from privacynotice;"""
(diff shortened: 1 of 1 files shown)


adamsachs (approved):
looks great @pattisdr , thanks for tackling this before heading out!

i'll go ahead and merge it into our integration branch.

PR #4676, opened 2024-03-11
Adding property ID support to fides.js

galvana:
Closes [#PROD-1674](https://ethyca.atlassian.net/browse/PROD-1674)

### Description Of Changes

Updated fides.js to accept a `property_id` query param.

### Steps to Confirm

* [ ] Start fidesplus
* [ ] Set the following env variables in `clients/.env`
```
```
* [ ] Start the Admin UI and Privacy Center
* [ ] In the Admin UI navigate to `Add systems` and add the first page of vendors from Compass
* [ ] Navigate to `Notice` and enable all the notices
* [ ] Navigate to `Properties` and create a new property, associate this property to the `US Modal` experience. Make note of the autogenerated `property ID`
* [ ] Navigate to `Experiences` and enable `US Modal`
* [ ] Verify fides.js by navigating to http://localhost:3001/fides.js?geolocation=US-CA&property_id={property_id}

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`


clients/privacy-center/pages/api/fides-js.ts
```diff
@@ -91,6 +91,11 @@ export default async function handler(
   // Check if a geolocation was provided via headers or query param
   const geolocation = await lookupGeolocation(req);
 
+  // Check if a property_id was provided in the query params
+  const propertyId = Array.isArray(req.query.property_id)
+    ? req.query.property_id[0]
+    : req.query.property_id;
```
NevilleS, line 97: This smells a bit funny, let's follow the standard pattern of `param={value}` instead 👍 
```suggestion
  const { property_id: propertyId } = req.query
```

I'm not 100% sure why the `isArray` call is there?

NevilleS (changes requested):
Some minor nits, yes. However, a larger issue worth flagging... as it stands, technically this will only be supported when FidesJS is bundled server-side (ie "prefetch" is enabled).

I think this might cause some gotchas in the future because that means FidesJS doesn't really "know" it's been initialized with a property ID. In the worst case this could totally happen:
1. Disable prefetch on the Privacy Center
2. Add script `fides.js?property_id=FDS-123` on the target website
3. When page loads, it fetches `fides.js` from the Privacy Center, but since prefetch is disabled no experience is bundled
4. FidesJS initializes and calls `fetchExperience` itself, but without a property ID

The net result is that FidesJS could be added to a site with `fides.js?property_id=...` but it doesn't actually filter the experience to use that property ID... blergh.

clients/privacy-center/pages/api/fides-js.ts
NevilleS: Also - please update the `fides-js` E2E tests for this 👍

clients/privacy-center/pages/api/fides-js.ts
```diff
@@ -91,6 +92,13 @@ export default async function handler(
   // Check if a geolocation was provided via headers or query param
   const geolocation = await lookupGeolocation(req);
 
+  const propertyId = verifyConditionsForPropertyId(
```
NevilleS, line 95: nit: I'd expect a function called `verify***` to not also be a getter and just return a value.

Consider a better version of `safeLookupPropertyId` or `getPropertyIdButAlsoVerifyConditions` etc

clients/privacy-center/common/property-id.ts
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * Verifies that certain conditions are met to be able to fetch experiences by property ID
+ */
+export const verifyConditionsForPropertyId = (
+  req: NextApiRequest,
+  geolocation: UserGeolocation | null,
+  environment: PrivacyCenterEnvironment,
+  fidesString: string | null
+) => {
+  const propertyId = req.query.property_id;
```
NevilleS, line 14: Ooh, I thought of another nit! 😜

The fides-js API route has a detailed jsdoc comment that is used to general OpenAPI docs that document the supported headers/params/etc. Can you update that as well? Thanks 😊

clients/privacy-center/common/property-id.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  if (propertyId !== undefined) {
```
NevilleS, line 16: Super-nit, very stylistic based but-- in general I'd recommend testing for what you *support* and throw errors if it doesn't match. That's a slightly more defensive approach.

Eg. instead of testing that it's not undefined and not an Array, instead test that it is *exactly* a string

NevilleS (approved):
Looks good! I left some extra nitpicks in there for you if you want.

clients/privacy-center/pages/api/fides-js.ts
galvana: Fair point, `safeLookupPropertyId` sounds good to me

clients/privacy-center/common/property-id.ts
galvana:
```
 *       - in: query
 *         name: property_id
 *         required: false
 *         description: Optional identifier used to filter for experiences associated with the given property. If omitted, returns all experiences not associated with any properties.
 *         schema:
 *           type: string
 ```

galvana (commented):
Keep the nits coming! They're all valid things to consider

PR #4689, opened 2024-03-13
Handle a mismatch in translations between experience & GVL translations

NevilleS:
Closes PROD-1811

### Description Of Changes

### Code Changes

* [X] Fix `extractMessagesFromGVLTranslations` to be more defensive when using external API data!
* [X] Narrow available locales to only those that exist in both experience & GVL translations
* [ ] Add "fallback" support to allow special cases like `es-MX` in the experience which should be OK with `es` in the GVL...?

### Steps to Confirm

* [ ] Configure a TCF experience with translations that don't exist in the GVL translations (e.g. `es-MX`) and check that the TCF experience doesn't crash!

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### DIFF
--- a/clients/fides-js/src/lib/i18n/i18n-utils.ts
+++ b/clients/fides-js/src/lib/i18n/i18n-utils.ts
@@ -107,33 +107,37 @@ function extractMessagesFromGVLTranslations(
   // itself is not available in most languages
   const extracted: Record<Locale, Messages> = {};
   locales.forEach((locale) => {
+    // TODO (PROD-1811): prefer a case-insensitive locale match
     const gvlTranslation = gvl_translations[locale] as any;
-    const messages: Messages = {};
+    if (gvlTranslation) {
+      const messages: Messages = {};
 
-    const recordTypes = [
-      "purposes",
-      "specialPurposes",
-      "features",
-      "specialFeatures",
-      "stacks",
-      "dataCategories",
-    ];
-    recordTypes.forEach((type) => {
-      Object.keys(gvlTranslation[type]).forEach((id) => {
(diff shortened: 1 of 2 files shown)


clients/fides-js/src/lib/i18n/i18n-utils.ts
```diff
@@ -107,33 +107,37 @@ function extractMessagesFromGVLTranslations(
   // itself is not available in most languages
   const extracted: Record<Locale, Messages> = {};
   locales.forEach((locale) => {
+    // TODO (PROD-1811): prefer a case-insensitive locale match
```
NevilleS, line 110: While coding this up I realized the right thing to do here is do some reasonably-fuzzy matching of GVL locales to our experience locales 😓

NevilleS (commented):
One TODO for myself...

adamsachs (approved):
this looks good! just making sure i understand the expected chagne in functionality:
- in TCF overlays, we will no longer error if there is an experience config translation language that has no corresponding GVL translation -- this is the required bugfix.
- in TCF overlays, we won't make available any experience config translations that do not have a corresponding GVL translation, to prevent mismatches between the experience-config language and the language used for GVL-provided text within that experience -- this wasn't strictly required, but definitely seems like reasonable/preferable behavior 👍

clients/fides-js/src/lib/i18n/i18n-utils.ts
NevilleS: Moved this to a DEFER since this is a blocker for testing!

PR #4720, opened 2024-03-15
Update FidesJS to support localizing CMP UI with configurable, non-English default locales

NevilleS:
Closes PROD-1831

### Description Of Changes

This fixes a missing gap in the initial implementation of FidesJS localization to support a dynamic, configurable "default locale" set in the Experience configuration. How this works is:
1. During initialization, scan the list of `experience.experience_config.translations` to find which is marked as `is_default`
2. Set the "default locale" in the `i18n` module to use, or fallback to `en` if no API default is provided
3. Use the new "default locale" as the fallback in a few places:
  1. When selecting the active locale for the overall session, the default locale is used when no "better" match exists
  2. When auto-detection is disabled, the default locale will always be used
  3. When the active locale (from the experience) is not available for a notice, the default locale is used if possible

The net result is now what you'd expect: when you set a translation as "default" in the Admin UI, that locale will be used in the FidesJS CMP UI everywhere... unless a better, more specific match is available for the user ;)

### Code Changes

* [X] Update test fixtures to include `is_default: true` and, for good measure, include `fr-CA` as a non-default translation
* [X] Update our `i18n` module to support changing the default locale, so we can rely on `i18n` as the source-of-truth
  * NOTE: Most `i18n` libraries probably won't support this out of the box, so if we do upgrade to something like LinguiJS, react-intl, etc. we'll need to modify them
* [X] Add a new `areLocalesEqual` helper to DRY up existing code for locale comparisons 👍 
* [X] Modify the `initializeI18n` logic to extract the default locale from the experience API
* [X] Modify the `selectBestNoticeTranslation` and `selectBestExperienceConfigTranslation` helpers to fallback to the configurable default locale when no better match exists (instead of English)
* [X] Add unit tests to cover as much as we can
* [X] Add new E2E tests to cover:
  * [X] Specifying an alternate default locale and localizing the CMP to match
  * [X] Specifying an alternate default locale and using that as the fallback for notices when they are missing an exact translation match

### Steps to Confirm

* [ ] Configure a Banner & Modal experience with at least three languages: English, Spanish, and French
* [ ] Disable auto-detect language on the experience
* [ ] Load the CMP UI and check that the experience is rendered in the default language (English)
* [ ] Change the default language to Spanish
* [ ] Load the CMP UI and check that the experience is rendered in the new default language (Spanish)
image
(this screenshot has auto detect disabled and default locale set to Spanish)

Advanced test to check that mismatched notices also fallback to the default:
* [ ] Re-enable auto-detect language on the experience
* [ ] Load the CMP UI and use `?fides_locale=fr` to change user locale to French, check that the experience is rendered in French
* [ ] Delete the French translation for one of the notices
* [ ] Load the CMP UI in French, check that the experience is rendered in French but the modified notice falls back to Spanish
image
(this screenshot has auto detect enabled, browser locale set to French, default locale set to Spanish, but the "Marketing" notice does not have a French translation)

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### DIFF
--- a/clients/fides-js/src/lib/i18n/index.ts
+++ b/clients/fides-js/src/lib/i18n/index.ts
@@ -66,7 +66,7 @@ interface I18n {
 
   /**
    * Set the current default locale for this session.
-   * 
+   *
    * WARN: This does not affect the behaviour of t(), load(), or activate()!
    * To change the active locale and get different translations, use
    * activate(). This method should only be used to determine what a
--- a/clients/fides-js/src/lib/i18n/index.ts
+++ b/clients/fides-js/src/lib/i18n/index.ts
@@ -57,6 +57,28 @@ interface I18n {
   activate(locale: Locale): void;
 
   /**
+   * Get the current default locale for this session.
+   *
+   * WARN: LinguiJS does not support getting/setting the default locale, so
+   * use this sparingly!
+   */
+  getDefaultLocale(): Locale;
+
+  /**
+   * Set the current default locale for this session.
+   * 
+   * WARN: This does not affect the behaviour of t(), load(), or activate()!
+   * To change the active locale and get different translations, use
+   * activate(). This method should only be used to determine what a
+   * "fallback" locale should be for the current user's session if the active
+   * locale cannot be used.
+   *
+   * WARN: LinguiJS does not support getting/setting the default locale, so
+   * use this sparingly!
+   */
+  setDefaultLocale(locale: Locale): void;
+
+  /**
    * Get the currently active locale.
    */
   get locale(): Locale;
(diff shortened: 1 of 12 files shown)


clients/fides-js/src/lib/i18n/index.ts
```diff
@@ -56,6 +56,28 @@ interface I18n {
    */
   activate(locale: Locale): void;
 
+  /**
+   * Get the current default locale for this session.
+   *
+   * WARN: LinguiJS does not support getting/setting the default locale, so
+   * use this sparingly!
+   */
+  getDefaultLocale(): Locale;
```
eastandwestwind, line 65: Do we have a plan for how we’d do this if / when we switch to LinguiJS?

eastandwestwind (approved):
Looks good just 1 question below.

clients/fides-js/src/lib/i18n/index.ts
NevilleS: Yeah, we could figure something reasonable out as needed. This wouldn't be too hard.

PR #4723, opened 2024-03-18
Issues Seeding Data due to Import Issues with Multilang models

pattisdr:
Closes #PROD-1839

### Description Of Changes

Running `nox -s dev -- shell postgres`  fails when the postgres setup script runs due to the way we've defined some of our new multitranslation models:

```shell
sqlalchemy.exc.InvalidRequestError: When initializing mapper mapped class PrivacyNotice->privacynotice, expression 'experiencenotices' failed to locate a name ("name 'experiencenotices' is not defined"). If this is a class name, consider adding this relationship() to the <class 'fides.api.models.privacy_notice.PrivacyNotice'> class after both dependent classes have been defined.
no custom setup logic found for postgres, skipping
```

You can get around this by importing in the ExperienceNotices model early into that script, but instead let's try to define these models more properly. I don't know if anyone else has hit this because `nox - fides_env(test)` works fine for me, if the imports are done in a certain order it would work. 

### Code Changes

* [ ] Move the ExperienceNotices many-to-many model into its own file (ExperienceNotices link Notices to Privacy Experience Configs)
* [ ] Define the `ExperienceConfig.privacy_notices` relationship on one side (ExperienceConfigs) instead of both sides

### Steps to Confirm

* [ ] `nox -s dev -- shell postgres` should work as expected with postgres example database seeded
* [ ] Unit tests should pass here and in fidesplus

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### DIFF
--- a/src/fides/api/models/privacy_notice.py
+++ b/src/fides/api/models/privacy_notice.py
@@ -20,6 +20,7 @@ from fides.api.models import (
     dry_update_data,
     update_if_modified,
 )
+from fides.api.models.experience_notices import ExperienceNotices
 from fides.api.models.location_regulation_selections import (
     DeprecatedNoticeRegion,
     PrivacyNoticeRegion,
@@ -153,15 +154,6 @@ class PrivacyNotice(PrivacyNoticeBase, Base):
         order_by="NoticeTranslation.created_at",
     )
 
-    # Forward declaration to avoid circular import errors
-    PrivacyExperienceConfig = "PrivacyExperienceConfig"
-
-    experience_configs = relationship(
-        PrivacyExperienceConfig,
-        secondary="experiencenotices",
-        back_populates="privacy_notices",
-    )
-
     @hybridproperty
     def default_preference(self) -> UserConsentPreference:
         """Returns the user's default consent preference given the consent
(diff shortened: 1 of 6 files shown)


pattisdr:
Pinned Fidesplus here to look for failing tests https://github.com/ethyca/fidesplus/pull/1356/files, I don't think any imports need to be adjusted on that said, ExperienceNotices as a many-to-many table often isn't called directly

pattisdr:
@galvana can I get you to look at this when you have a monent?

src/fides/api/models/privacy_notice.py
```diff
@@ -153,15 +154,6 @@ class PrivacyNotice(PrivacyNoticeBase, Base):
...
     )
 
-    # Forward declaration to avoid circular import errors
-    PrivacyExperienceConfig = "PrivacyExperienceConfig"
-
-    experience_configs = relationship(
-        PrivacyExperienceConfig,
-        secondary="experiencenotices",
-        back_populates="privacy_notices",
-    )
```
galvana, line 163: Changing `back_populates` to `backref` on `PrivacyExperienceConfig.privacy_notices` creates this relationship automatically on the `PrivacyNotice` right?

galvana (approved):
Straightforward change! I updated the command in the description to `nox -s dev -- shell postgres` to get this to run, you were missing `dev`. But after I got it running everything looked good 👍

pattisdr:
> ! I updated the command in the description to nox -s dev -- shell postgres

Oh sorry! thank you

src/fides/api/models/privacy_notice.py
pattisdr: yes! `backref` is no longer the [preferred way of doing this](https://docs.sqlalchemy.org/en/20/orm/relationship_api.html#sqlalchemy.orm.relationship.params.backref) I think you're supposed to use back_populates now because it's more straightforward, but I think this is a viable option for now for these two tightly linked models

pattisdr:
Thank you for the review @galvana ⭐

PR #4727, opened 2024-03-19
Sets GPP applicable sections appropriately when user is in a state outside of GPP jurisdiction

eastandwestwind:
Closes https://ethyca.atlassian.net/browse/PROD-1618

### Description Of Changes

When a given privacy experience uses `"us_approach": "state",`, we want to ensure we're setting the applicableSections property of the GPP response to [-1]. 

### Code Changes

* [ ] Sets GPP `applicableSections` to -1 whenever state is outside of GPP jurisdiction
* [ ] Added e2e tests

### Steps to Confirm

* [ ] Run e2e tests in `consent-banner-gpp.cy.ts`
* [ ] To test manually, follow steps in https://ethyca.atlassian.net/wiki/spaces/EN/pages/2866774055/2023-12-24+State+of+GPP#Setup, but when you get to choose between “national” and “state-by-state”, choose `state-by-state`
* [ ] Visit the test page from a state not in GPP, e.g. Visit http://localhost:3000/fides-js-demo.html?geolocation=us-nc
* [ ] Confirm that `__gpp('ping', (data) => {console.log({data})})` returns `applicableSections` of `-1` regardless of actions taken on consent banner / modal

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`


clients/privacy-center/cypress/e2e/consent-banner-gpp.cy.ts
```diff
@@ -554,4 +554,202 @@ describe("Fides-js GPP extension", () => {
       });
     });
   });
+
+  describe("with TCF disabled and GPP enabled, with experience region outside of GPP", () => {
```
NevilleS, line 558: This feels like way too much copy-pasta'ing, sorry... can we avoid duplicating _all_ these tests here?

clients/privacy-center/cypress/e2e/consent-banner-gpp.cy.ts
```diff
@@ -554,4 +554,202 @@ describe("Fides-js GPP extension", () => {
...
+    beforeEach(() => {
+      cy.fixture("consent/experience_gpp.json").then((payload) => {
+        stubConfig({
+          options: {
+            isOverlayEnabled: true,
+            tcfEnabled: false,
+          },
+          // override with a region that is outside GPP jurisdictions
+          experience: { ...payload.items[0], region: "us_nc" },
```
NevilleS, line 567: This part of the setup is unique to this scenario, but the rest is all shared right

clients/privacy-center/cypress/e2e/consent-banner-gpp.cy.ts
```diff
@@ -554,4 +554,202 @@ describe("Fides-js GPP extension", () => {
...
+          const [data] = args;
+          expect(data.pingData.applicableSections).to.eql([-1]);
+          // TODO: once locations and regulations are set, this value may change as it is currently hard coded
+          expect(data.pingData.supportedAPIs).to.eql([
+            "8:uscav1",
+            "10:uscov1",
+            "12:usctv1",
+            "11:usutv1",
+            "9:usvav1",
+          ]);
```
NevilleS, line 658: Do we know if this still expected? It feels "wrong" but might not matter.

clients/fides-js/src/lib/gpp/us-notices.ts
```diff
@@ -83,6 +83,10 @@ export const setGppNoticesProvidedFromExperience = ({
   const gppSection = FIDES_REGION_TO_GPP_SECTION[gppRegion];
 
   if (!gppSection) {
+    if (experience?.gpp_settings?.us_approach === GPPUSApproach.STATE) {
+      cmpApi.setApplicableSections([-1]);
```
NevilleS, line 87: Do we even need this `if`?

In other words: should we just _always_ default it to -1 here if later in the code we end up initializing other sections?

NevilleS (changes requested):
This looks pretty straightforward, thanks!

My main ask here is to reduce the duplication in the tests. You've copy-pasta'd ~200 LOC but there's only maybe 10-20 lines different between the the two `describe` blocks and that makes me pretty nervous about maintaining those specs.

clients/privacy-center/cypress/e2e/consent-banner-gpp.cy.ts
NevilleS: Couple suggestions:
1. You could extract the contents of the `beforeEach` block into a `visitDemoWithGPP` helper that provides an option to override just the `experience` part; see the `visitDemoWithI18n` helper I did for i18n tests here: https://github.com/ethyca/fides/blob/79e8c845953ae683647a63874328692ff2a64f63/clients/privacy-center/cypress/e2e/consent-i18n.cy.ts#L355. You could then organize the tests to be something like:
  - describe("with TCF disabled and GPP enabled")
    - describe("when visiting from a state with an applicable section")
      - beforeEach -> setup GPP but from an applicable state
      - (...all existing tests)
    - describe("when visiting from a state that does not have an applicable section")
      - beforeEach -> setup GPP but from outside (e.g. NC)
      - (...just one or two tests for this case without duplicating everything)
2. You could instead refactor this to be a parametrized test runner that sets up cases and expectations and then does the `testCases.forEach(...)` magic to reuse the same describe/it blocks for all the cases

In this case option #1 feels reasonable, it'd allow you to avoid _too much_ duplication and just DRY up some of the shared logic.

clients/privacy-center/cypress/e2e/consent-banner-gpp.cy.ts
eastandwestwind: This is intentional. Because `supported APIs` should include all locations configured with GPP in Admin-UI, we don't currently have a way to query for this, which is why we hard-code all possible GPP states. But we have https://ethyca.atlassian.net/browse/PROD-1595 for update the supportedAPIs specifically for state-by-state approach

clients/fides-js/src/lib/gpp/us-notices.ts
eastandwestwind: So this `if` allows us to exclude TCF. If we remove the `if`, then if both TCF and GPP is enabled, with a user visiting from the EU, this will line will set `-1`, overriding the `2` that was for TCF. 

For GPP with TCF, we do the following if we detect a TC string:

```
const tcSet = setTcString(event, cmpApi);
      if (tcSet) {
        cmpApi.setApplicableSections([TcfEuV2.ID]);
      }
```

So we wouldn't want to override that, if set.

Maybe instead of checking for `experience?.gpp_settings?.us_approach === GPPUSApproach.STATE`, I should check whether `tcSet` is false, then `cmpApi.setApplicableSections([-1])`. Does that make more sense to you?

NevilleS: Oh ok, that comes down to order of initializations then - it looked like this might be running before anything else.

Your existing logic is OK then, it's specific and accurate.

NevilleS (approved):

PR #4731, opened 2024-03-20
PROD-1653 Disable editing of type of data use and name after it has b…

lucanovera:
### Description Of Changes

After creation of a data use declaration, don't allow to edit the data use type or the declaration name.

### Code Changes

* [ ] Pass privacyDeclarationId prop that was missing
* [ ] Disable field is we're editing an already created data use declaration

### Pre-Merge Checklist

* [ ] Update `CHANGELOG.md`

### DIFF
--- a/clients/admin-ui/src/features/system/system-form-declaration-tab/PrivacyDeclarationForm.tsx
+++ b/clients/admin-ui/src/features/system/system-form-declaration-tab/PrivacyDeclarationForm.tsx
@@ -116,6 +116,8 @@ export const PrivacyDeclarationFormComponents = ({
   privacyDeclarationId?: string;
   lockedForGVL?: boolean;
 }) => {
+  const isEditing = !!privacyDeclarationId;
+
   const legalBasisForProcessingOptions = useMemo(
     () =>
       (
@@ -161,7 +163,7 @@ export const PrivacyDeclarationFormComponents = ({
           label="Declaration name (optional)"
           name="name"
           tooltip="Would you like to append anything to the system name?"
-          disabled={lockedForGVL}
+          disabled={isEditing || lockedForGVL}
           variant="stacked"
         />
         <CustomSelect
@@ -175,7 +177,7 @@ export const PrivacyDeclarationFormComponents = ({
           tooltip="For which business purposes is this data processed?"
           variant="stacked"
           isRequired
-          isDisabled={!!privacyDeclarationId || lockedForGVL}
+          isDisabled={isEditing || lockedForGVL}
         />
         <CustomSelect
           name="data_categories"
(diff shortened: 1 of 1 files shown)


lucanovera:
Moving back to draft to address the failing test.

clients/admin-ui/cypress/e2e/systems-plus.cy.ts
```diff
@@ -168,15 +168,26 @@ describe("System management with Plus features", () => {
...
+    it("don't allow editing declaration name after creation", () => {
       cy.getSelectValueContainer("input-name").type("L{enter}");
       cy.getByTestId("save-btn").click();
       cy.wait(["@postSystem", "@getSystem", "@getSystems"]);
       cy.getByTestId("tab-Data uses").click();
       cy.getByTestId("add-btn");
       cy.getByTestId("delete-btn");
       cy.getByTestId("row-functional.service.improve").click();
-      cy.getByTestId("input-name").should("not.be.disabled");
+      cy.getByTestId("input-name").should("be.disabled");
```
jpople, line 179: I think it's better to leave this test relatively in-place so we're still checking to make sure other fields on the data use can be edited in this scenario.  You can add the check that the name and the data use inputs are disabled to the new test you added, then change this line to verify that changes are allowed to one of the other inputs instead.

clients/admin-ui/cypress/e2e/systems-plus.cy.ts
```diff
@@ -168,15 +168,26 @@ describe("System management with Plus features", () => {
...
+    });
+
+    it("don't allow editing data uses after creation", () => {
+      cy.getSelectValueContainer("input-name").type("L{enter}");
+      cy.getByTestId("save-btn").click();
+      cy.wait(["@postSystem", "@getSystem", "@getSystems"]);
+      cy.getByTestId("tab-Data uses").click();
+      cy.getByTestId("add-btn");
+      cy.getByTestId("delete-btn");
```
jpople, line 188: The gets for `add-btn` and `delete-btn` are a little redundant here; those are only in the previous test to make sure that we show the interface for changing when the form isn't locked, which isn't really relevant here.

jpople (changes requested):

clients/admin-ui/cypress/e2e/systems-plus.cy.ts
lucanovera: Okay. So I'll leave the original test, just chaging the final line to check for another input. And then do my new tests below.

lucanovera (commented):
Applied changes to improve, based on @jpople comments.

jpople (approved):
Looks good, nice work!

PR #4743, opened 2024-03-25
PROD-1691 Implement dynamic CMP overrides

eastandwestwind:
Closes https://ethyca.atlassian.net/browse/PROD-1691

### Description Of Changes

CMP overrides consist of 2 groups:

```
{
  // the below are overrides to experience text, which only applies if the user is in the specified locale
  fides_language: "en",
  fides_title: "My title",
  fides_description: "We use cookies and similar tech...",
  fides_privacy_policy_url: "https://example.com/privacy",
  
  // the below overrides primary color, which applies if defined, otherwise we use the default primary color in Fides
 fides_primary_color
}
```

### Code Changes

* [ ] Adds / implements ability to override experience language and primary color
* [ ] Adds `fides_primary_color` as an out-of-the-box config option
* [ ] Adds specific utils to handle generating CSS from the configured primary color
* [ ] Adds e2e tests

### Steps to Confirm

* [ ] Run e2e tests in `privacy-center/cypress/e2e/consent-i18n.cy.ts` and `privacy-center/cypress/e2e/consent-banner.cy.ts`

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`


clients/fides-js/src/lib/i18n/i18n-utils.ts
```diff
@@ -66,11 +68,22 @@ function extractMessagesFromExperienceConfig(
       // For each translation, extract each of the translated fields
       (translation: ExperienceConfigTranslation) => {
         const locale = translation.language;
+        const localeHasOverride =
+          experienceTranslationOverrides?.override_language === locale;
```
eastandwestwind, line 72: this has not yet been tested. We need to ensure the format of override_language matches what we have as enums on the experience translation

clients/fides-js/src/lib/style-utils.ts
```diff
@@ -0,0 +1,97 @@
...
+};
+
+export enum ColorFormat {
+  HEX = "hex",
+  HSL = "hsl",
+}
+
+// takes color, lightens it by grade, returns HSL string or same hex color in case of error
+// grade 1 is one degree lighter, grade 2 is two degrees lighter
+export const generateLighterColor = (
```
eastandwestwind, line 72: also needs unit tests here, in my TODOs

clients/fides-js/src/lib/i18n/i18n-utils.ts
```diff
@@ -66,11 +68,22 @@ function extractMessagesFromExperienceConfig(
...
         const messages: Messages = {};
         EXPERIENCE_TRANSLATION_FIELDS.forEach((key) => {
+          const overrideValue =
+            localeHasOverride &&
+            Object.prototype.hasOwnProperty.call(
+              experienceTranslationOverrides,
+              key
+            )
+              ? // @ts-ignore
```
gilluminate, line 81: This feels dangerous here. Is there a less invasive way to coerce this type to be correct? Is something misconfigured upstream forcing this need (eg convert EXPERIENCE_TRANSLATION_FIELDS to an enum)? Maybe just a comment in addition explaining why?

clients/fides-js/src/lib/initialize.ts
```diff
@@ -198,51 +200,58 @@ const automaticallyApplyGPCPreferences = async ({
...
-      ({ fidesOption, fidesOptionType, fidesOverrideKey, validationRegex }) => {
-        const queryParamOverride: string | null =
-          queryParams.get(fidesOverrideKey);
+    const overrideValidatorMap = getOverrideValidatorMapByType(type);
+    overrideValidatorMap?.forEach(
+      ({ overrideName, overrideType, overrideKey, validationRegex }) => {
+        const queryParamOverride: string | null = queryParams.get(overrideKey);
         const windowObjOverride: string | boolean | undefined = windowObj
-          ? windowObj[fidesOverrideKey]
+          ? // @ts-ignore
```
gilluminate, line 238: same as above

clients/fides-js/src/lib/initialize.ts
```diff
@@ -198,51 +200,58 @@ const automaticallyApplyGPCPreferences = async ({
...
 
-        // Load the override option value, respecting the order of precedence (query params > window object > cookie)
+        // Load the override value, respecting the order of precedence (query params > window object > cookie)
         const value = queryParamOverride || windowObjOverride || cookieOverride;
         if (value && validationRegex.test(value.toString())) {
-          // coerce to expected type in FidesOptions
-          overrideOptions[fidesOption] =
-            fidesOptionType === "string" ? value : JSON.parse(value.toString());
+          // coerce to expected type
+          // @ts-ignore
```
gilluminate, line 247: same as above

gilluminate (commented):
Overall LGTM! just a few places where I called out `@ts-ignore` causing concern.

clients/fides-js/src/lib/consent-constants.ts
```diff
@@ -8,45 +13,83 @@ import { LOCALE_REGEX } from "./i18n/i18n-constants";
...
+    overrideName: "fidesPrimaryColor",
+    overrideType: "string",
+    overrideKey: "fides_primary_color",
+    validationRegex: /(.*)/,
+  },
+];
+
+export const FIDES_OVERRIDE_LANGUAGE_VALIDATOR_MAP: {
+  overrideName: keyof FidesExperienceTranslationOverrides;
+  overrideType: "string" | "boolean";
```
eastandwestwind, line 67: boolean type not needed here

clients/fides-js/src/lib/consent-types.ts
```diff
@@ -615,13 +626,28 @@ export type FidesOptionsOverrides = Pick<
...
 export type FidesOverrides = {
   optionsOverrides: Partial<FidesOptionsOverrides>;
   consentPrefsOverrides: GetPreferencesFnResp | null;
+  experienceTranslationOverrides: Partial<FidesExperienceTranslationOverrides>;
 };
 
+export enum OverrideType {
+  OPTIONS = "options",
+  LANGUAGE = "language",
+  THEME = "theme",
```
eastandwestwind, line 648: theme is not needed

clients/fides-js/src/lib/consent-types.ts
```diff
@@ -615,13 +626,28 @@ export type FidesOptionsOverrides = Pick<
...
+
 export type FidesOverrides = {
   optionsOverrides: Partial<FidesOptionsOverrides>;
   consentPrefsOverrides: GetPreferencesFnResp | null;
+  experienceTranslationOverrides: Partial<FidesExperienceTranslationOverrides>;
 };
 
+export enum OverrideType {
+  OPTIONS = "options",
+  LANGUAGE = "language",
```
eastandwestwind, line 647: rename to `EXPERIENCE_TRANSLATION`

clients/fides-js/src/lib/style-utils.ts
```diff
@@ -0,0 +1,97 @@
...
+  return cssString;
+};
+
+export enum ColorFormat {
+  HEX = "hex",
+  HSL = "hsl",
+}
+
+// takes color, lightens it by grade, returns HSL string or same hex color in case of error
+// grade 1 is one degree lighter, grade 2 is two degrees lighter
```
eastandwestwind, line 71: TODO: write better code comments in this file

clients/fides-js/src/lib/style-utils.ts
```diff
@@ -0,0 +1,97 @@
...
+  color: string | Hsl,
+  format: ColorFormat,
+  grade: 1 | 2
+): string => {
+  const hsl: Hsl | null =
+    format === ColorFormat.HEX ? hexToHSL(color as string) : (color as Hsl);
+
+  if (hsl) {
+    if (hsl.l < 0.25) {
+      // this is the lowest we can go ensuring enough contrast with the original color
```
eastandwestwind, line 82: better explain what this does... `This adds to the lightness parameter in decreasingly smaller increments across pre-determined intervals. The reason for this being that the lighter the original color is, the more impact an increase in lightness has on the resulting color.`

clients/fides-js/src/lib/i18n/i18n-utils.ts
eastandwestwind: 👍 fixed!

clients/privacy-center/cypress/e2e/consent-i18n.cy.ts
```diff
@@ -902,6 +908,256 @@ describe("Consent i18n", () => {
...
+        });
+        describe("when fides_override_language is only part of an experience locale string", () => {
+          beforeEach(() => {
+            visitDemoWithI18n({
+              navigatorLanguage: FRENCH_LOCALE,
+              globalPrivacyControl: true,
+              fixture: "experience_banner_modal.json",
+            });
+          });
+          // TODO (PROD-1885): matchLocale needs to support partial language match
```
eastandwestwind, line 982: ticket here- https://ethyca.atlassian.net/browse/PROD-1885. @NevilleS or @mfbrown can you tell me how urgent support for this is?

eastandwestwind:
Quick demo here of how I've added `fides_primary_color` as not only an override, but also an out-of-the-box config option: https://www.loom.com/share/8f50dba466df4773a02f1e1db51a5b7b

clients/privacy-center/cypress/e2e/consent-i18n.cy.ts
NevilleS: Left some notes on that ticket, but in general I think we should simplify it and not do partial matches at all when overriding translations 👍

clients/fides-js/src/lib/i18n/i18n-utils.ts
```diff
@@ -36,6 +37,63 @@ export function areLocalesEqual(a: Locale, b: Locale): boolean {
   );
 }
 
+function matchLocale(
```
NevilleS, line 40: I don't think you should need to extract this function (see other comments!).

If you _do_ need to extract it though, it still needs some kind of docs comment to explain why - even just a "see `matchAvailableLocales` for details" or similar

clients/fides-js/src/lib/i18n/i18n-utils.ts
```diff
@@ -72,11 +131,28 @@ function extractMessagesFromExperienceConfig(
...
       (translation: ExperienceConfigTranslation) => {
         const locale = translation.language;
+        // We only override experience translations if the override_language matches current locale
+        let localeHasOverride = false;
+        if (experienceTranslationOverrides?.override_language) {
+          localeHasOverride = !!matchLocale(
+            experienceTranslationOverrides?.override_language,
+            [locale]
+          );
+        }
```
NevilleS, line 141: Rather than doing a more complex match here, let's just simplify this with some logic like: 
```suggestion
        if (experienceTranslationOverrides?.override_language) {
          // If translation overrides exist for this language, we will need to apply them below
          localeHasOverride = areLocalesEqual(experienceTranslationOverrides.override_language, locale);
        }
```

This would also allow us to not use the `matchLocale` helper which I think is overly complex for this.

clients/fides-js/src/lib/i18n/i18n-utils.ts
```diff
@@ -72,11 +131,28 @@ function extractMessagesFromExperienceConfig(
...
         const messages: Messages = {};
         EXPERIENCE_TRANSLATION_FIELDS.forEach((key) => {
+          let overrideValue: string | null | undefined = null;
+          if (experienceTranslationOverrides) {
```
NevilleS, line 145:
```suggestion
          if (experienceTranslationOverrides && localeHasOverride) {
```

It's just a bit more legible like this

clients/fides-js/src/lib/consent-types.ts
```diff
@@ -605,6 +608,14 @@ export type OverrideOptions = {
...
   fides_locale: string;
+  fides_primary_color: string;
+};
+
+export type OverrideExperienceTranslations = {
+  fides_title: string;
+  fides_description: string;
+  fides_privacy_policy_url: string;
+  fides_override_language: string;
 };
```
NevilleS, line 619: I don't think I fully grok how these get set. Is it like this?
```
window.fides_overrides = {
  fides_title: "my custom title",
  fides_description: "my custom description",
  fides_privacy_policy_url: "https://example.com/my/custom/url",
  fides_override_language: "en"
}
```

If so I'm finding it confusing that `fides_title` wouldn't just always work, and it needs to be combined with the other `fides_override_language` option. Writing the docs for this might help us find a way that's easier to intuit

clients/fides-js/src/lib/i18n/i18n-utils.ts
NevilleS: Moreover though: let's update the unit tests to cover these cases. Right now test coverage on `i18n-utils` is somewhere around 100% 👍

clients/privacy-center/cypress/e2e/consent-i18n.cy.ts
```diff
@@ -902,6 +908,256 @@ describe("Consent i18n", () => {
...
+                  );
+                  cy.get(".fides-modal-description").contains(
+                    experienceTranslationOverrides.fides_description
+                  );
+                });
+              }
+            );
+          });
+        });
+        describe("when fides_override_language is only part of an experience locale string", () => {
```
NevilleS, line 974:
```suggestion
        });
        
        describe("when fides_override_language is only part of an experience locale string", () => {
```

NevilleS: 😄 

I really wish the formatter added these lines between blocks

clients/privacy-center/cypress/e2e/consent-i18n.cy.ts
```diff
@@ -543,15 +549,15 @@ describe("Consent i18n", () => {
               banner_description: modal.description,
               // Expect privacy policy link to not exist
               privacy_policy_link_label: null,
-              privacy_policy_link_url: null,
+              privacy_policy_url: null,
```
NevilleS, line 552: Hah, thanks! I wish Typescript caught this

NevilleS (changes requested):
Couple comments in here and hopefully some simplification

clients/fides-js/src/lib/i18n/i18n-utils.ts
eastandwestwind: This makes sense to me, thank you!

clients/fides-js/src/lib/consent-types.ts
eastandwestwind: right, it could be set via all 3 different ways we set our other overrides, including the `window.fides_overrides`, but also using cookie or query param.

RE: `I'm finding it confusing that fides_title wouldn't just always work`- I agree, and this was a compromise I had to make with keeping the implementation as generic as possible for overrides while supporting all the ways we set overrides. I'll add this to docs in a separate ticket- https://ethyca.atlassian.net/browse/PROD-1887

clients/fides-js/src/lib/i18n/i18n-utils.ts
eastandwestwind: yup you're right, now that we can simplify the matching 👍

clients/privacy-center/cypress/e2e/consent-banner.cy.ts
```diff
@@ -2339,4 +2340,35 @@ describe("Consent overlay", () => {
         .should("have.length", 3);
     });
   });
+
+  describe("fides overrides", () => {
```
NevilleS, line 2344:
```suggestion
  describe("fides experience translation overrides", () => {
```

nitpicky 😄

clients/privacy-center/cypress/e2e/consent-i18n.cy.ts
```diff
@@ -902,6 +908,259 @@ describe("Consent i18n", () => {
...
     });
+    describe("experience translation overrides", () => {
+      describe("when set via window obj", () => {
+        describe("when fides_override_language exactly matches experience locale", () => {
+          beforeEach(() => {
+            visitDemoWithI18n({
+              navigatorLanguage: ENGLISH_LOCALE,
+              globalPrivacyControl: true,
+              fixture: "experience_banner_modal.json",
+            });
```
NevilleS, line 919: Does this... work? When you call `stubConfig` again later I think you will end up just blowing away the results of this `beforeEach`?

clients/privacy-center/cypress/e2e/consent-i18n.cy.ts
NevilleS: You can nuke this test completely I think

clients/privacy-center/public/fides-js-components-demo.html
```diff
@@ -115,7 +115,7 @@
               notice_key: "advertising",
               description: "Advertising Description Test",
               internal_description: "Advertising Internal Description Test",
-              consent_mechanism: "opt_in",
+              consent_mechanism: "notice_only",
```
NevilleS, line 118: You sure you want to change this one?

NevilleS (approved):
Looking pretty solid to me, couple more nits.

clients/privacy-center/public/fides-js-components-demo.html
eastandwestwind: good catch! didn't mean to commit

clients/privacy-center/cypress/e2e/consent-banner.cy.ts
eastandwestwind: hmm so I want to keep these as generic overrides since here I'm specifically testing `fides_primary_color`. In the `privacy-center/cypress/e2e/consent-i18n.cy.ts` is where I test the experience translation overrides through localization.

clients/privacy-center/cypress/e2e/consent-i18n.cy.ts
eastandwestwind: 🤔

eastandwestwind: Wow, not sure how this was working, but I've fixed it!

gilluminate (approved):

PR #4746, opened 2024-03-26
Fix: move cell styles outside of `article`

gilluminate:
### Description Of Changes

Accidentally moved styles into the `article` styles instead of the root element.

### Code Changes

* [x] move the styles to the root

### Steps to Confirm

* [ ] Check table cell padding

### Screenshot of fix
CleanShot 2024-03-26 at 15 04 13@2x

### Pre-Merge Checklist


galvana (approved):
Verified locally, thanks for the quick fix!
image

PR #4748, opened 2024-03-27
PROD-1714-When-CMP-cannot-be-dismissed-page-content-should-not-scroll-overlay-should-appear

lucanovera:
### Description Of Changes

When a non-dismissable/blocking banner appears (TCF or not) the black overlay should appear on the screen and the scrolling should be blocked. The intent is to indicate to draw attention to the user to the banner as they need to complete that to continue browsing.

### Code Changes

* [ ] Add new isUiBlocking prop to Overlay component
* [ ] Supply new prop on TcfOverlay and NoticeOverlay
* [ ] Add helper functions to block scrolling
* [ ] Implement block scrolling on Overlay component when isUiBlocking is true and the banner is showing (Modal handle it's own blocking/overlay)
* [ ] Make black overlay appear when isUiBlocking is true

### Steps to Confirm

* [ ] Banner that is non-dismissable should have black overlay and block scrolling
* [ ] For testing TCF banner, you may need to make it non-dismissable from the API since you can't do it from the UI currently

### Pre-Merge Checklist

* [ ] Update `CHANGELOG.md`


clients/fides-js/src/lib/ui-utils.ts
```diff
@@ -0,0 +1,6 @@
+export const blockPageScrolling = () => {
+  document.body.style.overflowY = "hidden";
```
eastandwestwind, line 2: it feels a bit dangerous to be setting styles this way. Another option would be to add a class to `body` and then update the styles accordingly in `fides-js/src/components/fides.css`. I'm unsure if one approach is necessarily better than the other, but with setting a class, we can at least have more control over any CSS overrides / changes the customer wishes to implement manually, right?

My main concern is because fides.js loads on so many browsers across the internet, I'm unsure how this might interact with other sites' specific functionality, so I at least want to keep an eye on it as we roll this out to customers.

clients/fides-js/src/lib/ui-utils.ts
```diff
@@ -0,0 +1,6 @@
...
+};
+export const unblockPageScrolling = () => {
+  document.body.style.overflowY = "auto";
```
eastandwestwind, line 5: Do we know that setting "auto" will for sure revert functionality of all kinds of sites back to the original? Or should we be tracking the original `document.body.style.overflowY` just in case that was set to something other than "auto" before we touched it? I'm unsure here, but lemme know your thoughts!

clients/fides-js/src/components/Overlay.tsx
```diff
@@ -65,13 +67,26 @@ const Overlay: FunctionComponent<Props> = ({
...
+  isUiBlocking,
 }) => {
   const delayBannerMilliseconds = 100;
   const delayModalLinkMilliseconds = 200;
   const hasMounted = useHasMounted();
   const [bannerIsOpen, setBannerIsOpen] = useState(false);
   const modalLinkRef = useRef<HTMLElement | null>(null);
 
+  useEffect(() => {
+    if (isUiBlocking && bannerIsOpen) {
```
eastandwestwind, line 79: 👍

eastandwestwind (commented):
Just a couple Qs on approach here!

clients/fides-js/src/lib/ui-utils.ts
lucanovera: Yes, I think the class would be better. I actually coppied this approach from what I found on a11y-dialog.tsx. Let me see if I can change it quickly.

clients/fides-js/src/lib/ui-utils.ts
lucanovera: I will try adding a class with styles. That way it 100% reverts to original

clients/fides-js/src/lib/ui-utils.ts
```diff
@@ -0,0 +1,6 @@
+export const blockPageScrolling = () => {
+  document.body.classList.add("fides-no-scroll");
+};
+export const unblockPageScrolling = () => {
+  document.body.classList.remove("fides-no-scroll");
```
eastandwestwind, line 5: if for some reason this class isn't found, what will happen?

lucanovera: Just google it and tested it on my browser. It does nothing, it doesn't throw an error.

Kelsey-Ethyca (approved):
Thanks for addressing all Catherine's comments

PR #4751, opened 2024-03-27
fix language switcher with queryParams + e2e tests

gilluminate:
Closes [PROD-1888](https://ethyca.atlassian.net/browse/PROD-1888)

### Description Of Changes

i18n context was being initialized prior to the overrides taking effect. Changed to initialize as `null` and then use the override value once it was ready.

### Code Changes

* [x] Initialize i18n context with null
* [x] update i18n context value once overrides were available in the first plausible React component that could access the hook to do so (`NoticeOverlay.tsx`).
* [x] Add e2e tests to Cypress that verify the correct menu item is currently active to help catch this issue earlier.

### Steps to Confirm

* [ ] Add `fides_locale` query string (eg. `?fides_locale=es`) to the url of the fides test page: `/fides-js-components-demo.html?fides_locale=es`
* [ ] use the language picker in the banner to change language to English
* [ ] strings should switch to English

### Pre-Merge Checklist

* [x] All CI Pipelines Succeeded(https://github.com/ethyca/fidesdocs/issues/new/choose)
* [x] Follow up ticket created https://ethyca.atlassian.net/browse/PROD-1892

[PROD-1888]: https://ethyca.atlassian.net/browse/PROD-1888?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/clients/fides-js/src/components/notices/NoticeOverlay.tsx
+++ b/clients/fides-js/src/components/notices/NoticeOverlay.tsx
@@ -1,7 +1,7 @@
 import "../fides.css";
 
 import { Fragment, FunctionComponent, h } from "preact";
-import { useCallback, useMemo, useState } from "preact/hooks";
+import { useCallback, useEffect, useMemo, useState } from "preact/hooks";
 
 import { getConsentContext } from "../../lib/consent-context";
 import {
@@ -72,7 +72,13 @@ const NoticeOverlay: FunctionComponent<OverlayProps> = ({
     return [];
   };
 
-  const { currentLocale } = useI18n();
+  const { currentLocale, setCurrentLocale } = useI18n();
+
+  useEffect(() => {
+    if (!currentLocale && i18n.locale) {
+      setCurrentLocale(i18n.locale);
+    }
+  }, [currentLocale, i18n.locale, setCurrentLocale]);
 
   /**
    * Determine which ExperienceConfig translation is being used based on the
--- a/clients/fides-js/src/lib/i18n/i18n-context.tsx
+++ b/clients/fides-js/src/lib/i18n/i18n-context.tsx
@@ -1,17 +1,15 @@
 import { createContext, h, FunctionComponent } from "preact";
 import { useContext, useState, useMemo, StateUpdater } from "preact/hooks";
-import { i18n } from "./index";
 
 interface I18nContextProps {
-  currentLocale: string;
-  setCurrentLocale: StateUpdater<string>;
+  currentLocale: string | null;
+  setCurrentLocale: StateUpdater<string | null>;
 }
 
 const I18nContext = createContext<I18nContextProps>({} as I18nContextProps);
 
 export const I18nProvider: FunctionComponent = ({ children }) => {
-  const { locale } = i18n;
-  const [currentLocale, setCurrentLocale] = useState(locale);
+  const [currentLocale, setCurrentLocale] = useState<string | null>(null);
 
   const value: I18nContextProps = useMemo(
     () => ({ currentLocale, setCurrentLocale }),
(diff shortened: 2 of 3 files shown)


clients/fides-js/src/lib/i18n/i18n-context.tsx
```diff
@@ -1,17 +1,15 @@
 import { createContext, h, FunctionComponent } from "preact";
 import { useContext, useState, useMemo, StateUpdater } from "preact/hooks";
-import { i18n } from "./index";
 
 interface I18nContextProps {
-  currentLocale: string;
-  setCurrentLocale: StateUpdater<string>;
+  currentLocale: string | null;
+  setCurrentLocale: StateUpdater<string | null>;
```
NevilleS, line 6:
```suggestion
  currentLocale?: string;
  setCurrentLocale: StateUpdater<string | undefined>;
```

...whaddya think?

clients/fides-js/src/components/notices/NoticeOverlay.tsx
```diff
@@ -72,7 +72,13 @@ const NoticeOverlay: FunctionComponent<OverlayProps> = ({
...
   };
 
-  const { currentLocale } = useI18n();
+  const { currentLocale, setCurrentLocale } = useI18n();
+
+  useEffect(() => {
+    if (!currentLocale && i18n.locale) {
+      setCurrentLocale(i18n.locale);
+    }
+  }, [currentLocale, i18n.locale, setCurrentLocale]);
```
NevilleS, line 81: FWIW: seems like the move is to take all the `i18n` props and hoist those into context. I'm not opposed to that at all, it was just very _useful_ to me at the time to use prop-drilling to ensure I found everything 👍

clients/fides-js/src/components/notices/NoticeOverlay.tsx
```diff
@@ -72,7 +72,13 @@ const NoticeOverlay: FunctionComponent<OverlayProps> = ({
     return [];
   };
 
-  const { currentLocale } = useI18n();
+  const { currentLocale, setCurrentLocale } = useI18n();
```
NevilleS, line 75: It occurs to me now... is the language switcher only supported on the NoticeOverlay, but not TcfOverlay yet? If so let's track a follow-up ticket to add to TCF as well.

NevilleS (approved):
Ship it! Nitpick in there for you

clients/fides-js/src/components/notices/NoticeOverlay.tsx
gilluminate: ah, good catch. It is all set up to work, but didn't have this context stuff so is likely not working.

clients/fides-js/src/lib/i18n/i18n-context.tsx
gilluminate: probably more appropriate than initializing with a `null` value

clients/fides-js/src/components/notices/NoticeOverlay.tsx
gilluminate: https://ethyca.atlassian.net/browse/PROD-1892

clients/fides-js/src/components/notices/NoticeOverlay.tsx
gilluminate: Kind of depends how close we want to continue to keep things to the Linugi library. They inject the i18n object into the Provider to use, but then they update it using an event emitter. We would need to write a similar emitter to subscribe to https://github.com/lingui/js-lingui/blob/main/packages/react/src/I18nProvider.tsx#L74

If we are abandoning the idea of being Lingui compatible, I'd just as soon convert the entire i18n-utils to the context.

PR #4760, opened 2024-03-29
DSR 3.0: First Class Tasks + Parallelization

pattisdr:
# DSR 3.0 First Class Tasks and Parallelization 

Closes #PROD-1895 
Closes #PROD-1892

### Description Of Changes

Adds the new DSR 3.0 scheduler which supports executing DSR's in parallel and treats individual tasks in the Access, Erasure, and Consent Graphs as first-class objects.  The graph is built per usual, but then these nodes are persisted to the database as `RequestTasks` that are run as celery tasks in parallel. It uses a custom event-driven scheduler, where each node runs, and upon completion, queues its downstream tasks provided *that* task has all of its upstream tasks completed. The terminator task in each graph also is responsible for queueing the Privacy Request itself from the appropriate checkpoint to wrap up processing.  The individual checkpoints for a privacy request still run sequentially but the tasks within a graph run in parallel.

Maintains support for DSR 2.0 which uses Dask, which runs each of tasks in the Access, Erasure, and Consent graphs sequentially and in-memory.

👉 Key changes to pay attention to are in  `graph_task.py`, `request_runner_service.py`, `create_request_tasks.py`, and `execute_request_tasks.py` for DSR 3.0. 

DSR 2.0 is largely contained in `deprecated_graph_task.py`.

#### Data model

- New encrypted PrivacyRequest column, `filtered_final_upload` that contains encrypted access results we uploaded to the user keyed by policy rule for a limited time
- New encrypted PrivacyRequest column `access_result_urls` to store the access result URL's in case the Privacy Request exits in DSR 3.0 and is requeued, losing access to access result url's in memory (this could happen if a policy had both an access and an erasure rule)
- New `RequestTasks` model. Request tasks are now first class objects in DSR 3.0.  Every single node in the access, consent, and erasure graphs (which correspond roughly to a collection) is saved as a RequestTask to the database immediately after the graph is built.  RequestTasks store the intermediate data they collect on the RequestTasks themselves, whereas with DSR 2.0 stores them in the cache. For backwards compatibility, we continue to write to the cache regardless.

#### Endpoints
- Removes async callback endpoints that supported the deprecated v1 Manual Connector that assumed sequential task execution . ("/privacy-request/{privacy_request_id}/manual_input" and "/privacy-request/{privacy_request_id}/erasure_confirm").  The underlying deprecated connector has never been exposed in the admin UI and that connector is likewise removed here.
- Adds GET `/privacy-request/{privacy_request_id}/tasks`  to get all request tasks associated with a Privacy Request
- Adds GET `/privacy-request/{privacy_request_id}/requeue` to queue a stuck Privacy Request (use with caution)

#### Internal
- Removes sending code to fideslog related to DSR graphs and DSR failures.  We are phasing out fideslog anyway and this avoids having to refactor for DSR 3.0 related pieces. (Removes graph/analytics_events.py and graph/graph_differences.py)
- Adds `Collection.parse_from_task` to support hydrating a serialized collection saved in the database under `RequestTask.collection` into a `Collection` object.
- Add new `ExecutionNode` class to be used for executing a node. The existing `TraversalNode` class is now only used for building the graph, not executing it.  The ExecutionNode largely only has knowledge of itself and incoming and outgoing data edges which we can create from the database record, unlike the TraversalNode which is aware of the whole graph.  DSR 2.0 was previously able to use the TraversalNode because the graph was executed immediately in memory.
- Updates connectors to use ExecutionNodes instead of TraversalNodes
- Adds new `access_runner`, `erasure_runner`, and `consent_runner` which support running an access, erasure, or consent graph on the DSR 2.0 scheduler (in memory with Dask in sequence) or DSR 3.0 scheduler (in parallel, with persistent request tasks)
- GraphTasks are instantiated later, as part of running an individual celery task for a single Node.  They are now created from just TaskResources instead of TraversalNodes since they won't have the context of the whole graph.
- Logic that writes ExecutionLogs now also updates the new RequestTask status and additionally can mark all nodes that can be reached from the current node as errored if the  current node fails.
- Likely breaks the Fides Connector when run with DSR 3.0
- Stops passing in input_data into the request to mask a node.  These are the upstream access results that feed into the current node which were useful for erasure requests on email connectors that did not send access requests directly.  This functionality no longer exists.

#### AP Scheduler Tasks
- Recurring task "poll_for_exited_privacy_request_tasks" that runs on a configurable interval that looks to see if all Request Tasks have had a chance to run, but some have errored, and puts the entire Privacy Request in a status of error so it can be re-processed
- Weekly task `DSR_DATA_REMOVAL` that clears out old RequestTasks and selected columns on the Privacy Request model that are older than a configurable TTL containing encrypted PII

#### Tests
- A lot of the DSR tests modified to run both on the DSR 2.0 and DSR 3.0 schedulers
- Persist privacy requests in the database and add the correct policy to the privacy request for testing, which is needed for DSR 3.0 because each node is run as a celery task and relies on persisted objects.  For testing, we currently do a lot of creating a Privacy Request in memory and the policy we pass into the runner doesn't always match the policy on the privacy request (which is okay for 2.0 but not 3.0)
- Policies also need to be persisted for DSR 3.0 testing, so this means we can't use fake data categories - like "A" and B"

### Steps to Confirm

The best way we can test this is to try to run DSR's under a lot of different conditions under both DSR 3.0 and DSR 2.0! (With and without the worker as well!) I am continuing to test this week too -

* [x] In your toml file, toggle DSR 3.0 to on and task_always_eager to True (to test celery)
```
[execution]
use_dsr_3_0 = true

[celery]
task_always_eager = true
```
* [x] `nox -s dev -- shell postgres mongodb worker`
* [x] In another terminal, `cd /fides/clients`
* [x] `turbo run dev`
* [x] Setup postgres
* [x] http://localhost:3001/add-systems
* [x] Click Add a System
* [x] Add system name: `postgres_example_test_dataset`
* [x] Click Integrations
* [x] Integration type: PostgreSQL
* [x] Add credentials for local postgres example container running in docker (these aren't private secrets, these are for the example postgres database, docker/docker-compose.integration-postgres.yml). Test connection
Screenshot 2024-04-15 at 11 04 22 AM

* [x] Upload `postgres_example_test_dataset.yml` and click Save
* [x] Set up postgres, mongodb, and mailchimp connectors - run access and erasure requests and verify data looks as expected
* [x] Set up mailchimp transactional and run consent requests
* [x] Verify email connector works as expected
* [x]  In your toml file, turn off DSR 3.0 to use DSR 2.0:
```
[execution]
use_dsr_3_0 = false
```
* Repeat, and verify you can use DSR 2.0 as expected
* Verify exception logs in Admin UI look as expected - except DSR 3.0 has evidence items are running in parallel (if you're running in celery with task_always_eager=true and have worker container up).  Here, two tasks are queued at a time:
Screenshot 2024-04-15 at 11 10 16 AM

### Pre-Merge Checklist

* [x] Documentation: [Internal](https://ethyca.atlassian.net/wiki/spaces/EN/pages/3030188035/DSR+3.0+Scheduler)
* [x] Update `CHANGELOG.md`


src/fides/api/task/graph_task.py
```diff
@@ -585,56 +587,74 @@ def erasure_request(
...
         )
-        self.log_end(ActionType.erasure)
+        if self.request_task.id:
+            # DSR 3.0
+            self.request_task.rows_masked = 0  # Saved as part of update_status
+        # TODO Remove when we stop support for DSR 2.0
         self.resources.cache_erasure(
             f"{self.key}", output
         )  # Cache that the erasure was performed in case we need to restart
+        self.log_end(ActionType.erasure)
```
pattisdr, line 657: DSR 3.0 saving rows masked directly on the RequestTask while DSR 2.0 is caching it. This is largely used for testing, and for DSR 2.0 further kept us from re-running an erasure node that had already run.  DSR 3.0 has persistent tasks so can rely on the status of the task itself.

src/fides/api/task/graph_task.py
```diff
@@ -674,55 +697,15 @@ def collect_queries_fn(
         tn: TraversalNode, data: Dict[CollectionAddress, str]
     ) -> None:
         if not tn.is_root_node():
-            data[tn.address] = GraphTask(tn, resources).generate_dry_run_query()  # type: ignore
+            # Mock a RequestTask object in memory
+            resources.privacy_request_task = tn.to_mock_request_task()
+            data[tn.address] = GraphTask(resources).generate_dry_run_query()  # type: ignore
```
pattisdr, line 702: Mocking certain resources to still be able to generate a dry run query -

src/fides/api/task/graph_task.py
```diff
@@ -674,55 +697,15 @@ def collect_queries_fn(
...
-    dsk: Dict[CollectionAddress, Tuple[Any, ...]],
-    resources: TaskResources,
-    start_fn: Callable,
-) -> None:
-    """When resuming a privacy request from a paused or failed state, update the `dsk` dictionary with results we've
-    already obtained from a previous run. Remove upstream dependencies for these nodes, and just return the data we've
-    already retrieved, rather than visiting them again.
-
-    If there's no cached data, the dsk dictionary won't change.
-    """
```
pattisdr, line 694: Moved to deprecated_graph_task along with other DSR 2.0 code

src/fides/api/task/graph_task.py
```diff
@@ -735,69 +718,6 @@ def g() -> List[Dict[str, Any]]:
     return g
 
 
-async def run_access_request(
```
pattisdr, line 738: Moved to deprecated_graph_task along with other DSR 2.0 code

src/fides/api/task/graph_task.py
```diff
@@ -839,172 +759,8 @@ def get_cached_data_for_erasures(
     }
 
 
-def update_erasure_mapping_from_cache(
```
pattisdr, line 842: Moved to deprecated_graph_task along with other DSR 2.0 code

src/fides/api/task/graph_task.py
```diff
@@ -839,172 +759,8 @@ def get_cached_data_for_erasures(
...
-    """
-    cached_erasures: Dict[str, int] = resources.get_all_cached_erasures()
-
-    for collection_name in cached_erasures:
-        dsk[CollectionAddress.from_string(collection_name)] = cached_erasures[
-            collection_name
-        ]
-
-
-async def run_erasure(  # pylint: disable = too-many-arguments
```
pattisdr, line 858: Moved to deprecated_graph_task along with other DSR 2.0 code

src/fides/api/task/graph_task.py
```diff
@@ -839,172 +759,8 @@ def get_cached_data_for_erasures(
...
-        if collection_cycle:
-            raise TraversalError(
-                f"The values for the `erase_after` fields caused a cycle in the following collections {collection_cycle}"
-            )
-
-        v = delayed(get(dsk, TERMINATOR_ADDRESS, num_workers=1))
-        return v.compute()
-
-
-def _evaluate_erasure_dependencies(
```
pattisdr, line 936: Moved to deprecated_graph_task along with other DSR 2.0 code

src/fides/api/task/graph_task.py
```diff
@@ -839,172 +759,8 @@ def get_cached_data_for_erasures(
...
-    for collection in erase_after:
-        if collection in end_nodes:
-            # end_node list is modified in place
-            end_nodes.remove(collection)
-    # this task will execute after the collections in `erase_after` or
-    # execute at the beginning by linking it to the root node
-    return erase_after if len(erase_after) else {ROOT_COLLECTION_ADDRESS}
-
-
-async def run_consent_request(  # pylint: disable = too-many-arguments
```
pattisdr, line 955: Moved to deprecated_graph_task along with other DSR 2.0 code

src/fides/api/task/task_resources.py
```diff
@@ -75,8 +67,6 @@ def build_connector(  # pylint: disable=R0911,R0912
             return BigQueryConnector(connection_config)
         if connection_config.connection_type == ConnectionType.saas:
             return SaaSConnector(connection_config)
-        if connection_config.connection_type == ConnectionType.manual:
-            return ManualConnector(connection_config)
```
pattisdr, line 79: the v1 manual connector which assumed sequential execution is no longer supported, as noted elsewhere

src/fides/api/task/task_resources.py
```diff
@@ -96,39 +86,42 @@ def close(self) -> None:
...
 class TaskResources:
-    """Shared information and environment for all nodes of a given task.
+    """Holds some Database resources for the given task.
     This includes
      - the privacy request
+     - the request task
      - the policy
-     - redis connection
-     -  configurations to any outside resources the task will require to run
+     - configurations to any outside resources the task will require to run
```
pattisdr, line 94: I'm still using this TaskResources object to slim down the number of variables needed to instantiate a GraphTask - some of the attributes have been moved to GraphTask itself.  Largely this was kept however because we are still supporting DSR 2.0. DSR 2.0 queries a bunch of resources up front and saves them here.  Further, it has a lot of the caching methods.

src/fides/common/api/v1/urn_registry.py
```diff
@@ -73,10 +73,14 @@
...
+REQUEST_TASKS = "/privacy-request/{privacy_request_id}/tasks"
+PRIVACY_REQUEST_ACCESS_DATA = "/privacy-request/{privacy_request_id}/access_results"
+PRIVACY_REQUEST_TASK_QUEUE = (
+    "/privacy-request/{privacy_request_id}/tasks/{task_id}/queue"
+)
+
 PRIVACY_REQUEST_VERIFY_IDENTITY = "/privacy-request/{privacy_request_id}/verify"
 PRIVACY_REQUEST_RESUME = "/privacy-request/{privacy_request_id}/resume"
-PRIVACY_REQUEST_MANUAL_INPUT = "/privacy-request/{privacy_request_id}/manual_input"
-PRIVACY_REQUEST_MANUAL_ERASURE = "/privacy-request/{privacy_request_id}/erasure_confirm"
```
pattisdr, line 79: Callback endpoints that supported the deprecated v1 manual connector that assumed sequential execution.

src/fides/config/execution_settings.py
```diff
@@ -48,6 +48,14 @@ class ExecutionSettings(FidesSettings):
...
         description="Allows custom privacy request fields to be used in request execution.",
     )
+    state_polling_interval: int = Field(
+        default=30,
+        description="Seconds between polling for Privacy Requests that should change state",
+    )
+    use_dsr_3_0: bool = Field(
+        default=False,
+        description="Temporary flag to switch to using DSR 3.0 to process your tasks.",
+    )
```
pattisdr, line 58: For a time, supporting both DSR 3.0 and DSR 2.0 schedulers.

src/fides/api/task/deprecated_graph_task.py
```diff
@@ -0,0 +1,314 @@
...
+    dsk: Dict[CollectionAddress, Any] = {
+        k: (
+            t.erasure_request,
+            access_request_data.get(
+                str(k), []
+            ),  # Pass in the results of the access request for this collection
+            *_evaluate_erasure_dependencies(t, erasure_end_nodes),
+        )
+        for k, t in env.items()
+    }
```
pattisdr, line 226: https://github.com/ethyca/fidesops/pull/1168

Previous dsk graph also passed in original input data we used for the access graph, which was helpful for cases like the EmailConnector where the access request doesn't actually retrieve data.  However, this version of the email connector was later replaced and this data was not in use. 

Removing this also prevents us from having to retain this concept in DSR 3.0

```
   dsk: Dict[CollectionAddress, Any] = {
            k: (
                t.erasure_request,
                access_request_data[
                    str(k)
                ],  # Pass in the results of the access request for this collection
                *[
                    access_request_data[
                        str(upstream_key)
                    ]  # Additionally pass in the original input data we used for the access request. It's helpful in
                    # cases like the EmailConnector where the access request doesn't actually retrieve data.
                    for upstream_key in t.input_keys
                ],
            )
            for k, t in env.items()
        }
```

src/fides/api/task/graph_task.py
```diff
@@ -584,57 +586,69 @@ def access_request(self, *inputs: List[Row]) -> List[Row]:
     def erasure_request(
         self,
         retrieved_data: List[Row],
-        inputs: List[List[Row]],
```
pattisdr, line 587: Nothing was actually using these inputs so we've removed it.  It was for a previous version of the Email Connector that was later replaced.  Removing it prevents us from having to manage this same concept in DSR 3.0 

https://github.com/ethyca/fidesops/pull/1168

pattisdr:
@galvana I think this is ready for review! This week I am continuing to work on followup tickets to add things like catch-all schedulers and manage the number of connections but I think the core work is stable. Thank you for offering to review this 🙏

src/fides/api/api/v1/endpoints/privacy_request_endpoints.py
```diff
@@ -1920,3 +1782,133 @@ def _process_privacy_request_restart(
...
+        )
+
+    if pr.status != PrivacyRequestStatus.complete:
+        raise HTTPException(
+            status_code=HTTP_400_BAD_REQUEST,
+            detail=f"Cannot retrieve access data for incomplete request {pr.id}.",
+        )
+
+    logger.info(f"Getting privacy request '{privacy_request_id}' access data")
+    return pr.get_filtered_access_results()
```
pattisdr, line 1834: I am thinking we pull this endpoint out now that I am seeing revised requirements for https://ethyca.atlassian.net/wiki/spaces/PM/pages/3010101257/Self-service+access+package+downloads

src/fides/api/task/execute_request_tasks.py
```diff
@@ -0,0 +1,403 @@
...
+@celery_app.task(base=DatabaseTask, bind=True)
+def run_access_node(
+    self: DatabaseTask,
+    privacy_request_id: str,
+    privacy_request_task_id: str,
+    privacy_request_proceed: bool = True,
+) -> None:
+    """Run an individual task in the access graph for DSR 3.0 and queue downstream nodes
+    upon completion if applicable"""
+    with self.get_new_session() as session:
```
galvana, line 222: Are these the database connections we're worried about? What happens if we exceed the `task_engine_pool_size` limit?

galvana (commented):
First round of comments!

galvana:
I completed the testing checklist and verified the different integration types work as expected for both DSR 2.0 and 3.0 🙌

.fides/db_dataset.yml
```diff
@@ -2077,3 +2081,41 @@ dataset:
...
+    - name: collection_name
+      data_categories: [system]
+    - name: consent_sent
+      data_categories: [system]
+    - name: created_at
+      data_categories: [system]
+    - name: data_for_erasures
+      data_categories: [user]
+    - name: dataset_name
+      data_categories: [user]
```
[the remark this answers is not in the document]
pattisdr: thanks for catching 👍

tests/ops/integration_tests/saas/connector_runner.py
```diff
@@ -93,30 +101,39 @@ async def access_request(
         identities: Dict[str, Any],
         privacy_request_id: Optional[str] = None,
     ) -> Dict[str, List[Row]]:
+        from tests.conftest import access_runner_tester
+
         """Access request for a given access policy and identities"""
```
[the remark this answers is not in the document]
pattisdr: of course, sorry about that

src/fides/api/models/privacy_request.py
```diff
@@ -996,6 +968,173 @@ def error_processing(self, db: Session) -> None:
...
+            .first()
+        )
+        if not terminate:
+            raise Exception(
+                f"Expected {action.value.capitalize()} terminate node cannot be found on privacy request {self.id} "
+            )
+        assert terminate  # for mypy
+        return terminate
+
+    def get_raw_access_results(self) -> Dict[str, Optional[List[Row]]]:
```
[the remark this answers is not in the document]
pattisdr: thank you!

src/fides/api/service/privacy_request/request_service.py
```diff
@@ -137,3 +151,138 @@ async def poll_server_for_completion(
...
+        db.execute(
+            remove_dsr_data,
+            {
+                "ttl": (
+                    datetime.now()
+                    - timedelta(seconds=CONFIG.execution.request_task_ttl)
+                ),
+            },
+        )
+
```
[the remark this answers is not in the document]
pattisdr: solid idea 👍

src/fides/api/graph/data_type.py
```diff
@@ -198,10 +198,11 @@ def is_valid_data_type(type_name: str) -> bool:
...
 
 def get_data_type_converter(type_name: Optional[str]) -> DataTypeConverter:
-    """Return the matching type converter. If an empty string or None is passed in
+    """Return the matching type converter. If an empty string or None or string None is passed in
     will return the No-op converter, so the converter will never be set to 'None'.
-    On an illegal key will raise a KeyError."""
-    if not type_name:
+
+    Only an illegal key will raise a KeyError."""
+    if not type_name or type_name == "None":
```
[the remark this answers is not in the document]
pattisdr: Yes! Well, I created it the scenario, but it's based on the existing name of the NoOpTypeConverter.  I save a version of the collection as JSON in the database on the request task.  In order to serialize the data type converter, I save its name.  When I pull the collection json out of the database and rehydrate to be a proper Collection I convert the name back into a DataTypeConverter.

The official name of the NoOpTypeConverter is string "None".

Nice testing addition 👍

src/fides/api/task/graph_task.py
```diff
@@ -107,17 +106,18 @@ def result(*args: Any, **kwargs: Any) -> Any:
...
                         method_name,
-                        self.traversal_node.address,
+                        self.execution_node.address,
                     )
                     self.log_paused(action_type, ex)
                     # Re-raise to stop privacy request execution on pause.
                     raise
                 except PrivacyRequestErasureEmailSendRequired as exc:
                     traceback.print_exc()
+                    self.request_task.rows_masked = 0
```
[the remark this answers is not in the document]
pattisdr: Yes that's the idea!

src/fides/api/task/graph_task.py
```diff
@@ -154,61 +154,75 @@ def result(*args: Any, **kwargs: Any) -> Any:
...
+            # checkpoints if the graph fails in one of these sections.
+            self.resources.request.cache_failed_checkpoint_details(step=action_type)
             add_errored_system_status_for_consent_reporting(
                 self.resources.session,
                 self.resources.request,
                 self.connector.configuration,
             )
-            # Re-raise to stop privacy request execution on failure.
-            raise raised_ex  # type: ignore
+            if not self.request_task.id:
```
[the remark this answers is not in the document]
pattisdr: correct, I'm checking the request task instead of the config variable because I have some extra logic that switches the scheduler if the privacy request was in progress before DSR 3.0 was added -


for example, say we are running a Privacy Request on DSR 2.0 and it fails.  Then customer upgrades and they start using DSR 3.0.  If this privacy request is reprocessed, it will still run on DSR 2.0 -

src/fides/api/task/execute_request_tasks.py
pattisdr: I'm worried about both fides database connections and external database connections! I do have an API endpoint that lets you manually queue a "pending" task for cases where something has gotten stuck and am looking into more automated catch-all tasks that can re-queue tasks.

Thanks for highlighting this and I will continue to look at ways to recover if we max out database connection limits.

pattisdr:
Thank you for your first-pass review @galvana, I've responded to your changes here.  For database connection limits, I am looking into that under one of my other tickets, "Scheduler Robustness" to have ways to automatically schedule tasks that get into these weird states -

pattisdr:
I introduced some bugs in my logging - fixing

galvana (approved):
Amazing work as usual in such a short amount of time! Thank you for thinking through so many edge cases and for cleanly supporting DSR 2.0 and 3.0 together. Approved!

pattisdr:
Thanks again for all the code review Adrian! 🏆

PR #4764, opened 2024-03-30
Custom identities

galvana:
Closes [PROD-1806](https://ethyca.atlassian.net/browse/PROD-1806)

### Description Of Changes
This change adds the ability to use custom identities in the dataset identity references
```
collections:
  - name: loyalty
    fields:
      - name: id
        fides_meta:
          identity: loyalty_id
 ```
Since the likelihood that a Fides instance will use multiple identities after this change (for example `email` + `customer_id`), I also made a few changes in our graph utils to support multiple identities for SaaS connectors. 

#### Privacy Center
* [ ] Updated the `PrivacyRequestForm.tsx` to be able to render the new custom identities

#### Admin UI
* [ ] Updated the privacy request table and privacy request detail page to support the identity labels that are now returned from the API

#### Identity management
* [ ] Removed the `providedidentitytype` constraint on `providedidentity.field_name` to allow any custom defined field name
* [ ] Updated the `Identity` schema to allow extra fields as long as they have the `LabeledIdentity` type
* [ ] Updated `cache_identity`/`get_cached_identity_data` and `persist_identity`/`get_persisted_identity` functions to support labeled identities

#### Task execution
* [ ] Updated `pre_process_input_data` in `graph_task.py` to return unique output values (see inline comments)
* [ ] Removed single identity constraint from SaaS connectors

### Steps to Confirm

* [ ] Add a custom identity to the access request in `fides/data/sample_project/privacy_center/config/config.json`
```
"identity_inputs": {
  "email": "required",
  "loyalty_id": { "label": "Loyalty ID" }
},
```
* [ ] Run `nox -s fides_env(test)`
* [ ] Navigate to **Systems & Vendors** > **Cookie House Loyalty Program** > **Integrations** and enable the integration
* [ ] Navigate to the Privacy Center and submit the following access request
```
Email: jane@example.com
Loyalty ID: CH-1
```
* [ ] Navigate back to the Admin UI and approve the request
* [ ] Verify the presence of data from the `postgres_example_test_extended_dataset` in the DSR package that is written to `fides_uploads`

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

[PROD-1806]: https://ethyca.atlassian.net/browse/PROD-1806?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### CODE REVIEW

**Comment by adamsachs** [comment]
making my way through a code review, but posting UAT testing results here per instructions on the PR (thank you for providing a great test setup!). things are looking good! happy path results:
image

also tested if an invalid custom value is identity value is provided, request still succeeds but gets no results in the extended dataset, as expected: 
image


noting that the privacy center form validation is a _little_ bit odd in that it allows you to click "continue" but then gives a validation error, as opposed to the other required fields (i.e. the 'standard' identity fields), which actually grey out/block the user from clicking "continue". but that seems acceptable to me, just wanted to point it out (i'm assuming this is expected!) 
image

File: src/fides/api/util/collection_util.py
Comment (adamsachs) [inline src/fides/api/util/collection_util.py:1]: might be nice to get a bit of unit test coverage on these new functions specifically?

File: src/fides/api/task/graph_task.py [hunk 1]
```diff
@@ -376,8 +379,9 @@ def pre_process_input_data(
                             dependent_field_mappings=dependent_field_mappings,
                         )
 
-                    output[FIDESOPS_GROUPED_INPUTS].append(grouped_data)
-        return output
+                    output[FIDESOPS_GROUPED_INPUTS].add(make_immutable(grouped_data))
+
+        return make_mutable(output)
```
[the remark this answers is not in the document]
adamsachs: ok! this is a very helpful explanation. may be good to put a note about this functionality into the method docstring? (it's already a very explanatory docstring :) )

src/fides/api/util/collection_util.py
```diff
@@ -9,6 +12,24 @@
 FIDESOPS_DO_NOT_MASK_INDEX = "FIDESOPS_DO_NOT_MASK"
 
 
+def make_immutable(obj: Any) -> Any:
+    if isinstance(obj, dict):
```
adamsachs, line 16: docstrings here and on `make_mutable` would be nice!

adamsachs (commented):
@galvana really nice work making this comprehensive update, and thank you for anticipating all the impacts this change may have, including with relaxing the constraint to allow >1 identity value!

most of my comments are relatively minor tweaks, and i'm able to generally follow the identity-management updates well - i think your approach for custom identity support there is clever and it seems pretty robust (a few things around the edges that may help to make it a bit more defensive/less prone to accidental error moving forward)!

in terms of the `graph_task` updates, i'm following the concrete functionality you've added, i just generally always have a bit of trouble wrapping my head around what `pre_process_input_data` does generally - it's a pretty weighty method! so i'm feeling a bit less confident in my analysis of those changes. nothing particularly stands out to me as problematic, but it may be good to sync up to understand a bit more concretely how those updates will play in with the overall workflow...

as mentioned above, UAT testing is looking good! so i'm about ready to approve this, perhaps you can just look over my comments and we can align on the graph_task updates, and then we should be good to push this through 👍

src/fides/data/sample_project/postgres_sample.sql
```diff
@@ -189,6 +196,10 @@ INSERT INTO public.type_link_test VALUES
 ('1', 'name1'),
 ('2', 'name2');
 
+INSERT INTO public.loyalty VALUES
+('CH-1', 'Jane Customer', 100, 'Cookie Rookie'),
+('CH-2', 'John Customer', 200, 'Cookie Connoisseur');
```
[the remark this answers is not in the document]
galvana: I just thought it sounded funny 😆 I didn't know about the site

src/fides/api/util/collection_util.py
galvana: Added tests around the mutability functions

galvana (commented):
I addressed most of your comments except the one about adding more tests to the `dict` and `labeled_dict` functions. I'd like to know what test cases you had in mind.

src/fides/data/sample_project/postgres_sample.sql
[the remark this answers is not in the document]
adamsachs: hahaha i know i also thought it sounded great so i looked it up and was so happy with what i found

adamsachs (approved):
looking great after the latest updates! discussed some potential further tweaks offline that may be nice, and also getting some FE code review, but this is looking good to go from my end 👍

clients/privacy-center/components/modals/privacy-request-modal/PrivacyRequestForm.tsx
```diff
@@ -77,32 +70,37 @@ const usePrivacyRequestForm = ({
         return;
       }
 
-      const { email, phone, name, ...customPrivacyRequestFieldValues } = values;
+      // extract identity input values
+      const identityInputValues = Object.fromEntries(
+        Object.entries(action.identity_inputs ?? {}).map(([key, field]) => {
+          const value = values[key] || null;
```
jpople, line 76: Do we care about handling boolean and number values here?  If `values[key]` is falsy (including `false` or `0`), this syntax will overwrite that with null.  If this is just trying to fall back if the value doesn't exist, I would prefer using `??`.

galvana: Good catch, I added explicit checks for `undefined` and `""` before falling back to `null`.

jpople (approved):

PR #4766, opened 2024-04-01
Provide the modal link label function on Fides object

gilluminate:
Closes [#PROD-1710](https://ethyca.atlassian.net/browse/PROD-1710)

### Description Of Changes

Provide the modal link label function to the client based on the current locale unless specified via props to return default locale's language.

Assumes translation string will be returned along with other strings from the server as `modal_link_label` (WIP. Not hooked up to BE yet)

### Code Changes

* [x] Defines a `DEFAULT_MODAL_LINK_LABEL`
* [x] During fides initialization, after i18n has been initialized, sets the function for consumption on `window.Fides` object.
* [x] Updates documentation to include `getModalLinkLabel()` instructions.
* [x] Updates demo pages to use this new function for displaying the modal link when available.

### Steps to Confirm

* [ ] Set the string for each language in the demo
* [ ] visit the demo page with `?fides_locale` set for desired language.
* [ ] Make sure the geolocation used in demo page supports modal link
* [ ] See translated string on demo page.

### Pre-Merge Checklist

* [x] All CI Pipelines Succeeded
* Documentation:
  * [x] documentation complete, [PR opened in fidesdocs](https://github.com/ethyca/fidesdocs/pulls)
  * [ ] documentation [issue created in fidesdocs](https://github.com/ethyca/fidesdocs/issues/new/choose)
* [x] Issue Requirements are Met
* [x] Relevant Follow-Up Issues Created
* [x] Update `CHANGELOG.md`

PR #4767, opened 2024-04-02
PROD-1708: Add modal link label to experience model

eastandwestwind:
This PR (fides) and [this other PR](https://github.com/ethyca/fidesplus/pull/1376) (fides) close https://ethyca.atlassian.net/browse/PROD-1708

### Description Of Changes

Add modal link label to experience model

### Code Changes

* [ ] Add modal link label to experience model
* [ ] Add migration
* [ ] Update BE tests

### Steps to Confirm

* [ ] Run fidesplus, pointing to latest commit on this PR
* [ ] Run fides on this PR using `turbo run dev` in /clients
* [ ] Open up admin-UI, any experience config
* [ ] Update modal link label, save, see that it persists

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`


src/fides/api/models/privacy_experience.py
```diff
@@ -129,6 +129,7 @@ class ExperienceTranslationBase:
     privacy_policy_link_label = Column(String)
     privacy_policy_url = Column(String)
     privacy_preferences_link_label = Column(String)
+    modal_link_label = Column(String)
```
pattisdr, line 132: Do our out of the box experience configs in plus need to ship with this value?

eastandwestwind: nope, we set a default internally in fides-js so no need to have this db-side.

eastandwestwind: https://github.com/ethyca/fides/pull/4766/files#diff-73f3ee3f607d9c4cb47cd9edc7aeda7afc47df7b506ebca6f9f714efcb71b6a1R634

pattisdr: I don't understand. Is this not something we will _ever_ supply out of the box translations for? Why is this field different?

eastandwestwind: Hmm I see, OOB translations might be handy in the future. I'd like to consult with @mfbrown on this one.

To answer your Q, this field is diff from the other fields on experience translation in a couple ways:
1. It's not used automatically by fides internally to render anything in the UI. It's only used when the customer writes js on their site to explicitly grab this value from fides, e.g. `Fides.getModalLinkLabel()`
2. Since this field is used to actually add/replace text on a customers' site, it's possible we might want more predictability with only returning something they've explicitly configured rather than an OOB translation, but I'd like @mfbrown to confirm.

eastandwestwind: plus PR here, to address having OOB translations for this field in the future- https://github.com/ethyca/fidesplus/pull/1376

tests/ops/models/test_privacy_experience.py
```diff
@@ -110,6 +112,7 @@ def test_create_privacy_experience_config(self, db, privacy_notice):
         assert history.name == "My New Experience Config"
         assert history.origin is None
         assert history.privacy_preferences_link_label == "Manage preferences"
+        assert history.modal_link_label == "Manage my consent preferences"
```
pattisdr, line 115: Thanks for adding these assertions on both the translation record and the historical record 👍

pattisdr (approved):

PR #4782, opened 2024-04-10
Update FidesJS `fides_embed` option to support embedding both banner & modal components

guncha:
### Description Of Changes

This turned out to be a bit of an adventure. Initially I wanted to allow embedding the banner and the modal in separate containers and after spending some time going in this direction, it turned out to be not possible without a lot of effort refactoring the code which I didn't feel comfortable doing (also time). The issue is that putting the banner and the modal in two different containers would require separating the components "all the way to the top", so to speak, which would mean they'd have entirely separate state and everything.

The approach here is to embed both the banner and the modal in the same `fides-embed-container`. The `bannerIsOpen` state in `Overlay` is used to control if the banner or the modal is shown.

### Code Changes

* [x] `Overlay` component state management
* [x] CSS overrides when the banner is embedded

### Steps to Confirm

The changes can be tested by using the `fides_embed=true` override. The banner will be shown first and it should be replaced with the Layer 2 "modal" when the "Manage preferences" link is clicked. Using `fides_disable_banner=true` will skip the banner and show the modal immediately, which is the previous behavior.

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### DIFF
--- a/clients/fides-js/docs/interfaces/FidesOptions.md
+++ b/clients/fides-js/docs/interfaces/FidesOptions.md
@@ -76,14 +76,12 @@ ___
 When `true`, require FidesJS to "embed" it's UI into a specific `<div>` on
 the page, instead of as an overlay over the `<body>` itself. This is useful
 for creating a dedicated page to manage consent preferences on your site.
+Both the consent modal and the banner will be embedded into the container.
+To only embed the consent modal, set `fides_disable_banner` to `true`.
 
 To use the `fides_embed` option, ensure that a DOM element with
 `id="fides-embed-container"` exists on the page, which FidesJS will then
 use as the parent element to render within.
-
-Before version 2.34 only the consent modal was embedded while the banner was
-not shown at all, but since that version both will be embedded. To get the
-previous behavior also set `fides_disable_banner` to `true`.
 
 NOTE: If you're using a JavaScript framework (e.g. React), ensure that you
 do not re-render the parent `<div>` element, as this could remove the
--- a/clients/fides-js/docs/interfaces/FidesOptions.md
+++ b/clients/fides-js/docs/interfaces/FidesOptions.md
@@ -81,6 +81,10 @@ To use the `fides_embed` option, ensure that a DOM element with
 `id="fides-embed-container"` exists on the page, which FidesJS will then
 use as the parent element to render within.
 
+Before version 2.34 only the consent modal was embedded while the banner was
+not shown at all, but since that version both will be embedded. To get the
+previous behavior also set `fides_disable_banner` to `true`.
+
 NOTE: If you're using a JavaScript framework (e.g. React), ensure that you
 do not re-render the parent `<div>` element, as this could remove the
 FidesJS UI fully from the page!
--- a/clients/fides-js/src/components/ConsentBanner.tsx
+++ b/clients/fides-js/src/components/ConsentBanner.tsx
@@ -25,6 +25,7 @@ interface BannerProps {
   onVendorPageClick?: () => void;
   renderButtonGroup: (props: ButtonGroupProps) => VNode;
   className?: string;
+  isEmbedded: boolean;
 }
 
 const ConsentBanner: FunctionComponent<BannerProps> = ({
@@ -37,6 +38,7 @@ const ConsentBanner: FunctionComponent<BannerProps> = ({
   onVendorPageClick,
   renderButtonGroup,
   className,
+  isEmbedded,
 }) => {
   const [isMobile, setIsMobile] = useState(window.innerWidth < 768);
 
@@ -71,13 +73,18 @@ const ConsentBanner: FunctionComponent<BannerProps> = ({
     ? i18n.t("exp.banner_description")
     : i18n.t("exp.description");
 
+  const containerClassName = [
+    "fides-banner",
+    "fides-banner-bottom",
+    !bannerIsOpen && "fides-banner-hidden",
+    isEmbedded && "fides-embedded",
+    className,
+  ]
+    .filter((c) => typeof c === "string")
+    .join(" ");
+
   return (
-    <div
-      id="fides-banner-container"
-      className={`fides-banner fides-banner-bottom 
-        ${bannerIsOpen ? "" : "fides-banner-hidden"} 
-        ${className || ""}`}
-    >
+    <div id="fides-banner-container" className={containerClassName}>
       <div id="fides-banner">
         <div id="fides-banner-inner">
           <CloseButton
--- a/clients/fides-js/src/components/ConsentModal.tsx
+++ b/clients/fides-js/src/components/ConsentModal.tsx
@@ -1,4 +1,4 @@
-import { VNode, h } from "preact";
+import { ComponentChildren, VNode, h } from "preact";
 import { HTMLAttributes } from "preact/compat";
 
 import { Attributes } from "../lib/a11y-dialog";
@@ -9,17 +9,17 @@ import ConsentContent from "./ConsentContent";
 
 const ConsentModal = ({
   attributes,
+  children,
   dismissable,
   i18n,
   renderModalFooter,
-  renderModalContent,
 }: {
   attributes: Attributes;
+  children: ComponentChildren;
   dismissable: boolean | undefined;
   i18n: I18n;
   onVendorPageClick?: () => void;
   renderModalFooter: () => VNode;
-  renderModalContent: () => VNode;
 }) => {
   const { container, overlay, dialog, title, closeButton } = attributes;
 
@@ -48,7 +48,7 @@ const ConsentModal = ({
           i18n={i18n}
           renderModalFooter={renderModalFooter}
         >
-          {renderModalContent()}
+          {children}
         </ConsentContent>
       </div>
     </div>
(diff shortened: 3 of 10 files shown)


clients/fides-js/src/components/Overlay.tsx
```diff
@@ -108,12 +111,15 @@ const Overlay: FunctionComponent<Props> = ({
   });
 
   const handleOpenModal = useCallback(() => {
-    if (instance) {
+    if (options.fidesEmbed) {
+      setBannerIsOpen(false);
+      onOpen();
```
guncha, line 116: Now that I think about it, this might fire twice 🤔

clients/fides-js/src/components/Overlay.tsx
```diff
@@ -210,19 +216,20 @@ const Overlay: FunctionComponent<Props> = ({
...
-          renderModalFooter={() =>
-            renderModalFooter({
-              onClose: handleCloseModalAfterSave,
-              isMobile: false,
-            })
-          }
-        >
-          {renderModalContent()}
-        </ConsentContent>
+        bannerIsOpen ? null : (
```
eastandwestwind, line 219: 👍 nice, so this prevents the banner and modal from showing at the same time within the container

clients/fides-js/src/components/Overlay.tsx
eastandwestwind: yeah looks like currently we're firing this `onOpen` twice when fidesEmbed is enabled and we click into the modal. One fires here and one fires in a `useEffect` below- https://github.com/ethyca/fides/pull/4782/files#diff-12e4a1baa5e44f967e542dbb092fa1a6212d8ad7eec90b4b495c32dd707027edR132

We can probably remove the one here, since we need the `onOpen` to be handled in the case where we skip the banner, not only when we explicitly open the modal from the banner.

clients/fides-js/src/components/ConsentModal.tsx
```diff
@@ -48,7 +48,7 @@ const ConsentModal = ({
           i18n={i18n}
           renderModalFooter={renderModalFooter}
         >
-          {renderModalContent()}
+          {children}
```
eastandwestwind, line 51: ah, we were already passing `children`, and duplicating it here I see. Thanks for fixing this!

eastandwestwind (changes requested):
Tested this and looking great! It's almost there, just wanna fix that duplicate event firing.

banner:
Screenshot 2024-04-10 at 5 26 44 PM

modal:
Screenshot 2024-04-10 at 5 27 00 PM

Also confirmed that Admin UI preview mode still looks good:
Screenshot 2024-04-10 at 5 53 00 PM

clients/fides-js/src/components/ConsentBanner.tsx
```diff
@@ -71,13 +73,18 @@ const ConsentBanner: FunctionComponent<BannerProps> = ({
     ? i18n.t("exp.banner_description")
     : i18n.t("exp.description");
 
+  const containerClassName = [
+    "fides-banner",
+    "fides-banner-bottom",
+    !bannerIsOpen && "fides-banner-hidden",
+    isEmbedded && "fides-embedded",
```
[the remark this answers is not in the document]
guncha: Hmm I removed `fides-embed` because it wasn't used in the CSS.. Could it be used by someone else in the wild?

clients/fides-js/src/components/Overlay.tsx
guncha: Makes sense 👍 Yeah.. The state in this component is getting pretty complex (approaching truth table territory), but that's a project for another time.

clients/fides-js/src/components/ConsentBanner.tsx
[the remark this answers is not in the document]
NevilleS: Oh, OK! I didn't notice that. If it's being used, it'd be by us here: https://github.com/ethyca/fidesplus/blob/main/clients/fides-js-extensions/nytimes/src/ext.css

Looking... 👀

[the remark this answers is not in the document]
NevilleS: Nope we use `div#fides-embed-container` in the NYT CSS. Looks good, replace it

NevilleS (approved):
I'm comfy merging this as-is (once we're up to date with main & CHANGELOG) but I do have one more stretch ask...

Can you also update the developer docs here: https://github.com/ethyca/fides/blob/262d2f3b006e2a2045cfde3acf31cb61893b3d7e/clients/fides-js/src/docs/fides-options.ts#L53
(`/clients/fides-js/src/docs/fides-options.ts`)

A quick explanation there will be helpful, and you can reference the `fides_disable_banner` option to explain to a developer how to only embed the modal.

clients/fides-js/docs/interfaces/FidesOptions.md
```diff
@@ -81,6 +81,10 @@ To use the `fides_embed` option, ensure that a DOM element with
 `id="fides-embed-container"` exists on the page, which FidesJS will then
 use as the parent element to render within.
 
+Before version 2.34 only the consent modal was embedded while the banner was
+not shown at all, but since that version both will be embedded. To get the
+previous behavior also set `fides_disable_banner` to `true`.
```
NevilleS, line 86: nit: I'd rather not reference "previous versions" here, just state the fact:
* will embed banner & modal
* if you don't want banner, disable it

does that make sense?

NevilleS (changes requested):
docs nit pick :)

clients/fides-js/src/components/Overlay.tsx
eastandwestwind: agreed. There's a lot of refactoring we can do with fides.js logic overall

clients/privacy-center/cypress/e2e/consent-banner-tcf.cy.ts
```diff
@@ -1134,189 +1134,198 @@ describe("Fides-js TCF", () => {
...
-      cy.fixture("consent/experience_tcf.json").then((experience) => {
-        stubConfig({
-          options: {
-            isOverlayEnabled: true,
-            tcfEnabled: true,
-            fidesEmbed: true,
-            fidesDisableBanner: true,
-          },
-          experience: experience.items[0],
+  describe("when fides_embed is true", () => {
```
NevilleS, line 1137: This diff is hard to read, but I reorganized the specs to be:
* `when fides_embed is true`
  * `when fides_disable_banner is false`
    * (NEW) test the layer 1 banner is shown
  * `when fides_disable_banner is true`
    * (all existing tests go here)

NevilleS, line 1137: This diff is hard to read, but I reorganized the specs to be:
* `when fides_embed is true`
  * `when fides_disable_banner is false`
    * (NEW) test the layer 1 banner is shown
  * `when fides_disable_banner is true`
    * (all existing tests go here)

NevilleS (changes requested):
I think I fixed this, but want to get @guncha to double-check me

clients/fides-js/src/components/Overlay.tsx
```diff
@@ -76,7 +76,7 @@ const Overlay: FunctionComponent<Props> = ({
   const delayBannerMilliseconds = 100;
   const delayModalLinkMilliseconds = 200;
   const hasMounted = useHasMounted();
-  const [bannerIsOpen, setBannerIsOpen] = useState(options.fidesEmbed);
+  const [bannerIsOpen, setBannerIsOpen] = useState(false);
```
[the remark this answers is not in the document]
guncha: The reason I did it was because otherwise, when `fides_disable_banner` is `false`, it would flash the modal and then re-render immediately and show the banner. My change didn't take into account `fides_disable_banner` so it should have had the opposite problem when it was set to true. Seems like it should be `options.fidesEmbed && !options.fidesDisableBanner`. Not quite sure why it would break the tests though so I'll do some investigation.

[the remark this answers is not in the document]
guncha: So yeah, this is the issue with just setting to to `false`:

content-flash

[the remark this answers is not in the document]
guncha: Okay, I think I understand it now. There's this bit of code in `Overlay.tsx`:

https://github.com/ethyca/fides/blob/df46bf6b55544814c341353b2869d3ff9ff270b5/clients/fides-js/src/components/Overlay.tsx#L144-L150

This is needed to first render `ConsentBanner` in the DOM with `bannerIsOpen=false` and then flip it to `true` after a short delay which will result in the removal of the `fides-banner-hidden` class and a nice entry animation.

In the embed case, we are not animating anything so we can jump to the final state, but it has to be the correct one, otherwise there will be a blip of the other component OR nothing being rendered. Still not sure how this worked before, but I'm glad the e2e test caught it.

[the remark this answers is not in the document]
eastandwestwind: I think it worked before because we were not rendering the banner at all with the embed mode. But this is a good catch, I've not used the performance tab in this way before, so I learned something new!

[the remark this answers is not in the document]
NevilleS: Ah, gotcha. Alright, I think overall this still feels like a fragile bit of state to manage (ie if we were being extra careful, that "banner is open" state feels like it's the job of the ConsentBanner component itself to handle it's own animation on mount) but this looks good now.

NevilleS (approved):

eastandwestwind (approved):

PR #4785, opened 2024-04-11
PROD-1748 support automatically translating privacy center

lucanovera:
### Description Of Changes

* [x] Display the best translation available in the consent page of the privacy center
* [x] Use correct history ids for the displayed translation for notices-served and save preferences
* [x] Replace On/Off radio buttons with toggle component

### Code Changes

* [x] Import i18n library
* [x] Create i18n context and useI18n hook. This avoid having to pass i18n instance everywhere.
* [x] Replace default texts with appropiate translation calls
* [x] Replace On/Off radio buttons with toggle component
* [x]  Write tests
### Screenshots
Notice Consent Translations
Screen Shot 2024-04-11 at 10 55 22

Legacy config-based consent works as usual
Screen Shot 2024-04-11 at 10 55 01

### Pre-Merge Checklist

* [x] Update `CHANGELOG.md`


clients/privacy-center/common/i18nContext.tsx
```diff
@@ -0,0 +1,25 @@
+import { createContext, useMemo, useState } from "react";
+import { I18n, setupI18n } from "fides-js";
+
+interface I18nContextProps {
+  currentLocale: string | undefined;
+  setCurrentLocale: (newLocale: string) => void;
+  i18nInstance: I18n;
```
lucanovera, line 7: Storing i18n singleton in the context to avoid passing it down through multiple components

clients/privacy-center/common/hooks/useI18n.ts
```diff
@@ -0,0 +1,51 @@
+import { useCallback, useContext } from "react";
+import {
+  ExperienceConfig,
+  PrivacyNotice,
+  selectBestExperienceConfigTranslation,
+  selectBestNoticeTranslation,
+} from "fides-js";
+import { I18nContext } from "../i18nContext";
+
+const useI18n = () => {
```
lucanovera, line 10: Added hook for quick access to i18n instance and useful functions that already have the necessary context.

clients/privacy-center/components/consent/NoticeDrivenConsent.tsx
```diff
@@ -204,8 +220,10 @@ const NoticeDrivenConsent = ({ base64Cookie }: { base64Cookie: boolean }) => {
     const noticePreferences = Object.entries(draftPreferences).map(
       ([historyKey, preference]) => {
         const notice = notices.find(
-          // TODO (PROD-1748): pass in specific language shown in UI
-          (n) => n.translations[0].privacy_notice_history_id === historyKey
+          (n) =>
+            !!n.translations.find(
+              (t) => t.privacy_notice_history_id === historyKey
```
lucanovera, line 225: To be safe, attempt to find recorded preference from any translation.

clients/cypress-e2e/cypress/e2e/smoke_test.cy.ts
```diff
@@ -125,25 +125,22 @@ describe("Smoke test", () => {
     //  - Product Analytics => true
     cy.getByTestId(`consent-item-marketing.advertising`).within(() => {
       cy.contains("Data Sales or Sharing");
-      cy.getRadio("true").should("be.checked");
-      cy.getRadio("false").should("not.be.checked");
+      cy.getToggle().should("be.checked");
```
lucanovera, line 128: Move from radio buttons to the toggle simplified this kind of checks so much :)

clients/privacy-center/cypress/e2e/consent-i18n.cy.ts
```diff
@@ -1509,7 +1509,149 @@ describe("Consent i18n", () => {
...
+      );
+      cy.getByTestId("consent-item-pri_notice-analytics-000").contains(
+        SPANISH_NOTICES[1].description
+      );
+    });
+
+    /* Commenting tests for now. The problem is that the test runs with config based consent,
+      which don't support translations. Then, we override with cy.overrideSettings({ IS_OVERLAY_ENABLED: true });
+      to use notices-based consent to run the test, but notices-server has already ran with the previous config
+      */
```
jpople, line 1595: May need a second opinion on this, from @NevilleS or @eastandwestwind maybe?  What would be required to make these tests work?

clients/privacy-center/pages/consent.tsx
```diff
@@ -214,16 +221,61 @@ const Consent: NextPage = () => {
     redirectToIndex,
   ]);
 
+  /**
+   * Initialize internalization library.
+   */
+  const experience = useAppSelector(selectPrivacyExperience);
+  const { IS_OVERLAY_ENABLED } = settings;
+  const isConfigDrivenConsent = !IS_OVERLAY_ENABLED;
+  const [isI18Initialized, setIsI18Initialized] = useState(false);
```
jpople, line 230:
```suggestion
  const [isI18nInitialized, setIsI18nInitialized] = useState(false);
```

clients/privacy-center/pages/consent.tsx
```diff
@@ -214,16 +221,61 @@ const Consent: NextPage = () => {
...
+
   return (
     <Stack as="main" align="center" data-testid="consent">
-      <Stack align="center" py={["6", "16"]} spacing={8} maxWidth="720px">
-        <Stack align="center" spacing={3}>
-          <ConsentHeading />
-          <ConsentDescription />
+      {/* Wait until i18n is initalized so we can diplay the correct language and
+       also we can use the correct history ids */}
+      {isI18Initialized && (
```
jpople, line 269: We try to use `condition ? <Component /> : null` rather than `condition && <Component/>` for JSX to prevent confusing behavior with falsy-but-not-false values-- reasoning is explained [here](https://kentcdodds.com/blog/use-ternaries-rather-than-and-and-in-jsx).  It's not totally applicable in this case because `isI18nInitialized` should never be anything other than exactly `true` or `false` but it's easier to just always do it this way by convention.

jpople (changes requested):

clients/privacy-center/common/hooks/useI18n.ts
```diff
@@ -0,0 +1,54 @@
...
+  const i18n = context.i18nInstance;
+
+  if (!context) {
+    throw new Error("useI18n must be used within a I18nProvider");
+  }
+
+  // Useful wrapper for selectBestExperienceConfigTranslation
+  const selectExperienceConfigTranslation = useCallback(
+    (experienceConfig: ExperienceConfig) => {
+      const experienceConfigTransalation =
```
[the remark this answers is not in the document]
lucanovera: Good catch! Thanks

clients/privacy-center/cypress/e2e/consent-i18n.cy.ts
lucanovera: Yeah, I generally don't like that the override is applied after load. I think we should implement it in a way of having that setup from the beginning. But that sounds to me like it could be it's own ticket, as to not keep making this PR bigger.
cc @NevilleS

clients/privacy-center/pages/consent.tsx
lucanovera: Didn't know that. Just applied the change. thanks.

clients/privacy-center/cypress/e2e/consent-i18n.cy.ts
NevilleS: Yeah, I understand and empathize - in fact, I'm quite happy to see that you managed to get the _regular_ UI loading under test! I'll ask a few more questions in Slack about this...

NevilleS: 👍 I'm OK with disabling these tests. That said, personally I like _skipped_ tests more, so can you do something like:
```
// DEFER (PROD-1748): Disable E2E tests for notices served because {xyz}...
describe.skip("when reporting notices served and preferences saved", () => {
...(all these commented out tests here)
});

File: clients/privacy-center/cypress/e2e/consent-i18n.cy.ts [hunk 9]
```diff
@@ -1509,7 +1509,149 @@ describe("Consent i18n", () => {
...
+        fixture: "consent/geolocation.json",
+      }).as("getGeolocation");
+
+      // Experience intercept
+      cy.intercept("GET", `${API_URL}/privacy-experience/*`, {
+        fixture: "consent/experience_privacy_center.json",
+      }).as("getExperience");
+
+      cy.visitWithLanguage("/consent", SPANISH_LOCALE);
+      cy.overrideSettings({ IS_OVERLAY_ENABLED: true });
```
NevilleS, line 1555: One thing you could _try_ to actually get this working as intended would be to change this logic to:
```suggestion
      cy.visitWithLanguage("/", SPANISH_LOCALE);
      cy.overrideSettings({ IS_OVERLAY_ENABLED: true });
      cy.getByTestId({whatever the consent button is}).click();
```
...in other words:
1. Load the home page
2. Override settings to enable overlay
3. Navigate to consent page

That might do the trick?

lucanovera: Might work, yeah, I will try it

clients/privacy-center/components/consent/SaveCancel.tsx
```diff
@@ -1,31 +1,38 @@
...
-);
+  cancelLabel?: string;
+} & StackProps) => {
+  const { i18n } = useI18n();
+
+  return (
+    <Stack direction="row" justifyContent="flex-start" width="full" {...props}>
+      <Button size="sm" variant="outline" onClick={onCancel}>
+        {cancelLabel || "←"}
+      </Button>
```
gilluminate, line 22: The default here is going to be problematic for accessibility; I'm assuming this was meant to avoid translation. We need to translate this one way or another and include it as an `aria-label` if we're just using an icon. Maybe best just to include the translated label similar to the save button?

clients/privacy-center/pages/consent.tsx
```diff
@@ -214,16 +221,61 @@ const Consent: NextPage = () => {
...
+      return;
+    }
+
+    initializeI18n(
+      i18n,
+      window?.navigator,
+      experience as PrivacyExperience,
+      {
+        debug: true,
+      },
```
gilluminate, line 257: should this be hardcoded here?

clients/privacy-center/pages/consent.tsx
```diff
@@ -214,16 +221,61 @@ const Consent: NextPage = () => {
...
+
   return (
     <Stack as="main" align="center" data-testid="consent">
-      <Stack align="center" py={["6", "16"]} spacing={8} maxWidth="720px">
-        <Stack align="center" spacing={3}>
-          <ConsentHeading />
-          <ConsentDescription />
+      {/* Wait until i18n is initalized so we can diplay the correct language and
+       also we can use the correct history ids */}
+      {isI18nInitialized ? (
```
gilluminate, line 269:
```suggestion
      {isI18nInitialized && (
```

clients/privacy-center/pages/consent.tsx
```diff
@@ -214,16 +221,61 @@ const Consent: NextPage = () => {
...
+            <ConsentHeading />
+            <ConsentDescription />
+          </Stack>
+          {consentContext.globalPrivacyControl ? <GpcBanner /> : null}
+          <ConsentToggles storePreferences={storeConsentPreferences} />
         </Stack>
-        {consentContext.globalPrivacyControl ? <GpcBanner /> : null}
-        <ConsentToggles storePreferences={storeConsentPreferences} />
-      </Stack>
+      ) : null}
```
gilluminate, line 278:
```suggestion
      )}
```

gilluminate (changes requested):

clients/privacy-center/components/consent/SaveCancel.tsx
lucanovera: Yes, you're it, it's not accesible. I am adding an aria-label with hardcoded value of "Cancel". We reviewed it during a refinement call and agreed to just use the icon specifically to not have to add new translations strings.

clients/privacy-center/pages/consent.tsx
lucanovera: I'll change it to check for the node environment to be development.

clients/privacy-center/pages/consent.tsx
lucanovera: Yup, I like just using && to. Jeremy mentioned something about false-y values that can create confusion, and that's why they prefer the ? and null: But this is a boolean, so no confusion there. I'll apply the suggestions.

clients/privacy-center/pages/consent.tsx
lucanovera: cc @gilluminate

gilluminate: @jpople that's easily solved by always forcing boolean on non-boolean types.
for example, in the article you referenced here:
```
!!contacts.length && ...
```
Personally I find it to be a much cleaner approach and the way that React documentation suggests it should be written.

gilluminate (approved):
Thanks for those changes. Looking good! 🍎

jpople (approved):

PR #4786, opened 2024-04-11
PROD-1941 Fix bug that prevented adding new privacy center translations

lucanovera:
### Description Of Changes

Fix bug that prevented adding new translation for privacy center

### Code Changes
Investigated the problem a bit, and notices that there was a form validator error about some field being null.
Captura de pantalla 2024-04-11 a la(s) 11 42 45 a  m

The problem is that the privacy center experience don't have a need for this field, but the validator stills checks for it. Added nullable option to the validator to account for this use case. 

### Pre-Merge Checklist

* [x] Update `CHANGELOG.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,10 @@ The types of changes are:
 
 ## [Unreleased](https://github.com/ethyca/fides/compare/2.34.0...main)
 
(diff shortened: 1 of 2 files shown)


jpople (approved):
💯

PR #4795, opened 2024-04-15
PROD-1869 - Support config for pre-response webhooks

eastandwestwind:
Closes https://ethyca.atlassian.net/browse/PROD-1869

### Description Of Changes

Backend infra to support future pre-response hook functionality. 

This PR:
*  allows configuration of pre-response hooks
*  creates the necessary DB tables for storage of pre-response hook configs and replies from pre-response hooks

### Code Changes

* [ ] Adds 2 new db tables
* [ ] Adds CRUD operations for config of pre-response hooks
* [ ] Adds unit tests for CRUD ops

### Steps to Confirm

* [x] Full unit test coverage has been created for the new `pre_approval_webhook_endpoints.py`, but if you wish to manually test, confirm all the new endpoints are successful.
* [x] First you'll need to create a valid ConnectionConfig. PATCH `/api/v1/connection` with:
```
[
   {
      "name":"my_connection_config",
      "key":"my_connection_config",
      "connection_type":"https",
      "access":"read",
      "secrets":{
         "url":"http://example.com",
         "authorization":"test_authorization"
      }
   }
]
```
* [x] Now you can create any number of pre-approval webhook that is linked to this connection config. PUT `/api/v1/dsr/webhook/pre_approval` with:
```
[
   {
      "connection_config_key":"my_connection_config",
      "name":"My webhook",
      "key":"my_webhook"
   }
]
```
* [x] GET `/api/v1/dsr/webhook/pre_approval` should return this item
* [x] GET `/api/v1/dsr/webhook/pre_approval/my_webhook` should also return this item
* [x] You should be able to edit your pre approval webhook. PATCH `/api/v1/dsr/webhook/pre_approval/my_webhook` with, for example,  `{"name": "Some other name"}`
* [x] GET `/api/v1/dsr/webhook/pre_approval/my_webhook` should reflect this change
* [x] You should be able to delete your webhook. DELETE `/api/v1/dsr/webhook/pre_approval/my_webhook` should be successful, and GET `/api/v1/dsr/webhook/pre_approval` should not return that webook at all.

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`


src/fides/api/schemas/pre_approval_webhooks.py
```diff
@@ -0,0 +1,52 @@
...
+class PreApprovalWebhookResponse(WebhookBase):
+    """Response schema after creating/updating/getting a PreApprovalWebhook"""
+
+    connection_config: Optional[ConnectionConfigurationResponse]
+
+
+class Config:
+    """Set orm_mode to True"""
+
+    orm_mode = True
```
pattisdr, line 38: Assume you mean this to be nested under PreApprovalWebhookResponse!

pattisdr:
Starting review!

src/fides/api/api/v1/endpoints/pre_approval_webhook_endpoints.py
```diff
@@ -0,0 +1,258 @@
...
+)
+from fides.api.models.pre_approval_webhook import PreApprovalWebhook
+from fides.api.oauth.utils import verify_oauth_client
+from fides.api.schemas import pre_approval_webhooks as schemas
+from fides.api.util.api_router import APIRouter
+from fides.api.util.connection_util import get_connection_config_or_error
+from fides.common.api import scope_registry as scopes
+from fides.common.api.v1 import urn_registry as urls
+
+router = APIRouter(tags=["Pre Approval Webhooks"], prefix=urls.V1_URL_PREFIX)
```
pattisdr, line 28: This router needs to be imported into src/fides/api/api/v1/api.py to actually register the endpoint!

tests/ops/api/v1/endpoints/test_pre_approval_webhook_endpoints.py
pattisdr, line 1: All tests failing -

src/fides/api/models/pre_approval_webhook.py
```diff
@@ -0,0 +1,32 @@
...
+
+class PreApprovalWebhook(Base):
+    name = Column(String, unique=True, nullable=False)
+    key = Column(String, index=True, unique=True, nullable=False)
+    connection_config_id = Column(
+        String, ForeignKey(ConnectionConfig.id_field_path), nullable=False
+    )
+    connection_config = relationship(
+        ConnectionConfig, back_populates="pre_approval_webhook", uselist=False
+    )
```
pattisdr, line 19: Getting this error, you likely need to set up the corresponding relationship on the ConnectionConfig since you're using `back_populates` - 
```
Error occurred writing config settings to database: One or more mappers failed to initialize - can't proceed with initialization of other mappers. Triggering mapper: 'mapped class PreApprovalWebhook->preapprovalwebhook'. Original exception was: Mapper 'mapped class ConnectionConfig->connectionconfig' has no property 'pre_approval_webhook'
```

pattisdr (changes requested):
@eastandwestwind before I dig into your endpoints, passing it back early to get the relationships set up and the endpoints registered -

src/fides/api/models/pre_approval_webhook.py
eastandwestwind: yes good catch! I've added this now!

pattisdr: there's often some tricky import issues to figure out around these!

pattisdr: Maybe experiment with putting ConnectionConfig in quotes where the relationship is defined on the PreApprovalWebhook side -

eastandwestwind: hm, that doesn't work either (locally when I run `nox -s collect_tests`). Very weird as the pattern is similar to how we've previously defined the `access_manual_webhook` relationship on `ConnectionConfig`

pattisdr: this can be tricky! There's not a one-size-fits-all answer, sometimes I end up having to collocate models in the same file, sometimes I drop back into using "backref" (the old way) instead of "back_populates" which defines the relationship on one size instead of both.

pattisdr:
Starting review!

src/fides/api/api/v1/endpoints/pre_approval_webhook_endpoints.py
```diff
@@ -0,0 +1,259 @@
...
+) -> PreApprovalWebhook:
+    """Helper method to load PreApprovalWebhook or throw a 404"""
+    logger.info("Finding PreApprovalWebhook with key '{}'", webhook_key)
+    pre_approval_webhook = PreApprovalWebhook.get_by(
+        db=db, field="key", value=webhook_key
+    )
+    if not pre_approval_webhook:
+        raise HTTPException(
+            status_code=HTTP_404_NOT_FOUND,
+            detail=f"No Policy found for key {webhook_key}.",
```
pattisdr, line 61: copy/paste issue in exception -

src/fides/api/api/v1/endpoints/pre_approval_webhook_endpoints.py
```diff
@@ -0,0 +1,259 @@
...
+    urls.WEBHOOK_PRE_APPROVAL_DETAIL,
+    status_code=HTTP_200_OK,
+    response_model=schemas.PreApprovalWebhookResponse,
+    dependencies=[Security(verify_oauth_client, scopes=[scopes.WEBHOOK_READ])],
+)
+def get_pre_approval_webhook_detail(
+    *,
+    db: Session = Depends(deps.get_db),
+    webhook_key: FidesKey,
+) -> PolicyPreWebhook:
```
pattisdr, line 77: Copy/paste issue, this return type is a PreApprovalWebhook - fix this and you can remove the `#ignore` on L81.

src/fides/api/api/v1/endpoints/pre_approval_webhook_endpoints.py
```diff
@@ -0,0 +1,259 @@
...
+    """
+    pre_approval_webhooks = PreApprovalWebhook.query(db=db)
+
+    keys = [
+        get_key_from_data(webhook.dict(), PreApprovalWebhook.__name__)
+        for webhook in webhooks
+    ]
+    names = [webhook.name for webhook in webhooks]
+    # Because resources are dependent on each other for order, we want to make sure that we don't have multiple
+    # resources in the request that actually point to the same object.
```
pattisdr, line 103: This entire copy/pasted comment doesn't feel relevant here - Policy Webhooks had a specific order in which they ran, these don't seem to.

src/fides/api/api/v1/endpoints/pre_approval_webhook_endpoints.py
```diff
@@ -0,0 +1,259 @@
...
+    Loads the given Pre-Approval Webhook by key
+    """
+    return get_pre_approval_webhook_or_error(db, webhook_key)  # type: ignore
+
+
+def put_webhooks(
+    db: Session = Depends(deps.get_db),
+    webhooks: List[schemas.PreApprovalWebhookCreate] = Body(...),
+) -> List[PreApprovalWebhook]:
+    """
```
pattisdr, line 88: Ah okay I was wondering why you'd extracted these functions out since they were only being called in one place, but I see you've patterned it off of policy_webhook_endpoints where the corresponding function was shared between multiple endpoints.

src/fides/api/schemas/pre_approval_webhooks.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+class WebhookBase(FidesSchema):
+    """Base schema for Webhooks"""
+
+    key: Optional[FidesKey]
+    name: Optional[str]
+
+
+class PreApprovalWebhookCreate(WebhookBase):
```
pattisdr, line 18: Should we make the name required on create? I see this is patterned off of the Policy webhooks but we get 500 errors without it.

src/fides/api/models/connectionconfig.py
```diff
@@ -148,6 +148,13 @@ class ConnectionConfig(Base):
         uselist=False,
     )
 
+    pre_approval_webhooks = relationship(  # type: ignore[misc]
+        "PreApprovalWebhook",
+        back_populates="connection_config",
+        cascade="delete",
+        uselist=False,
```
pattisdr, line 155: I think we should remove `uselist=False` because a connection config can have multiple pre approval webhooks right? If uselist=False, this will just return one object.

pattisdr (changes requested):
@eastandwestwind endpoints looking good, mostly just some trivial copy/paste issues. 

The main thing I see missing are model-level tests (like tests/ops/models/test_connectionconfig.py).  They don't need to be large, because you're testing some of this through your endpoints, but I'm particularly looking for tests that create a PreApprovalWebhook and PreApprovalWebhookReply and assert their relationships on both sides.  For example, assert PreApprovalWebhook.connection_config and ConnectionConfig.pre_approval_webhooks are the types you expect, same for PreApprovalWebhookReplay and the privacy request relationship.

src/fides/api/models/connectionconfig.py
pattisdr: Model-level tests are good for flushing these out -

src/fides/api/api/v1/endpoints/pre_approval_webhook_endpoints.py
eastandwestwind: yeah, I can inline actually!

pattisdr: not a big deal either way -

src/fides/api/schemas/pre_approval_webhooks.py
eastandwestwind: hm, yes, that makes sense, as the table requires it and we don't generate a "name" if empty 👍

src/fides/api/models/connectionconfig.py
eastandwestwind: Ahhh I see, that makes sense! I'll remove this

eastandwestwind: This means I likely should add `uselist=False` to the `privacy_request` field on my `PreApprovalWebhookReply` api model: 
```
privacy_request = relationship(
        "PrivacyRequest", back_populates="pre_approval_webhook_reply", uselist=False  # type: ignore
    )
```

tests/ops/models/test_pre_approval_webhook.py
```diff
@@ -0,0 +1,133 @@
...
+        assert loaded_reply.is_eligible is True
+        assert loaded_reply.created_at is not None
+
+        # assert relationship with PrivacyRequest
+        privacy_request = (
+            db.query(PrivacyRequest)
+            .filter_by(id=privacy_request_status_pending.id)
+            .first()
+        )
+        assert loaded_reply.privacy_request == privacy_request
```
pattisdr, line 131: i like to also test the reverse, "privacy_request.pre_approval_webhook_reply".  There can be multiple replies right?  so shouldn't it be pre_approval_webhook_replies on the privacy request model side?

pattisdr (commented):
great @eastandwestwind thank you for taking the time to write these nice model tests.  My last question is about the naming of the privacy request <> pre approval webhook reply relationship because it maybe should be plural and represented in the test

tests/ops/models/test_pre_approval_webhook.py
eastandwestwind: that's a good idea and makes sense, as we can have multiple replies against the same privacy request 👍 

updating this now...

pattisdr (approved):
Great work @eastandwestwind

pattisdr:
Yes, I don't always aim for 100%, depending on the circumstance, you can look at the failures and see if there are any lines you think should be covered -

PR #4780, opened 2024-04-08
pre-bundle GPP extension

gilluminate:
Closes #<issue>

### Description Of Changes

Changed GPP extension to be pre-bundled in appropriate circumstances, as opposed to another fetch.

### Code Changes

* [x] Included GPP extension as part of the `fides.js` response using the exact same logic as the fetch was using.
* [x] Fix: Remove an unnecessary `export` from the GPP extension, which was causing errors once pre-bundled instead of fetched.
* [x] Removed `extensions.ts` and references, since loading GPP was it's only purpose (so far).
* [x] Used `setTimeout` to avoid race condition with GPC which was exposed when extension setup functions were removed.

### Misc. Clean up

* [x] Fixed consent settings cypress test (was giving false positive on `cy.getByTestId("section-MSPA").should("not.exist");` which was only true because *nothing* was on the page. By checking for something to exist _first_, it ensures the negative is accurate). Noticed while running the GPP tests in cypress to get my bearings on the topic.
* [x] Updated debug log from "Fides overlay is now showing!" to "Fides overlay is now in the DOM!" since Modal only experiences don't actually show anything at that point until the button is clicked.
* [x] Cleaned up the `updateExperience` function in fides.ts for easier to read code and consistency with fides-tcf.ts

### Steps to Confirm

* [ ] Enable GPP in PC
* [ ] check for existence of `__gpp` on the window object.

### Pre-Merge Checklist

* [x] All CI Pipelines Succeeded
* Documentation:
  * [ ] documentation complete, [PR opened in fidesdocs](https://github.com/ethyca/fidesdocs/pulls)
  * [ ] documentation [issue created in fidesdocs](https://github.com/ethyca/fidesdocs/issues/new/choose)
* [x] Issue Requirements are Met
* [x] Relevant Follow-Up Issues Created
* [x] Update `CHANGELOG.md`

PR #4799, opened 2024-04-15
Developer option to force GPP

gilluminate:
Closes [#PROD-1946](https://ethyca.atlassian.net/browse/PROD-1946)

### Description Of Changes

- [x] depends on PR #4780 to be merged first!

Updates fides.js to accept a new query string option that ensures the GPP API is always included.

When the GPP API is included, it sets `applicableSections: [-1]` whenever a user visits a page from a location that is not one of the supported sections.

### Code Changes

* adds support for new `gpp=true` query string option
* includes gpp in fides.js response regardless of location when query string is present
* Cypress tests to ensure inclusion as well as expected `applicableSections: [-1]` result

### Steps to Confirm

* visit demo page with query string included and from a location that is not supported (eg. http://localhost:3000/fides-js-demo.html?gpp=true&geolocation=us-id)
* inspect the GPP API and by calling the following from the browser console:
```
__gpp('ping', (data) => {console.log(data)})
```
* note that `applicableSections` property in that response is set to `[-1]`
```
__gpp('ping', (data) => {console.log(data.applicableSections)})
```

### Pre-Merge Checklist

* [x] All CI Pipelines Succeeded(https://github.com/ethyca/fidesdocs/issues/new/choose)
* Documentation:
* [x] Update `CHANGELOG.md`


clients/fides-js/src/fides-ext-gpp.ts
```diff
@@ -224,6 +230,7 @@ export const initializeGppCmpApi = () => {
         cmpApi,
         cookie: event.detail,
         experience: window.Fides.experience,
+        forcedGpp: window.Fides.options.forceGpp,
```
eastandwestwind, line 233: can we be consistent with naming, e.g. making every reference either `forcedGpp` or `forceGpp`?

clients/fides-js/src/lib/gpp/us-notices.ts
```diff
@@ -83,7 +85,10 @@ export const setGppNoticesProvidedFromExperience = ({
   const gppSection = FIDES_REGION_TO_GPP_SECTION[gppRegion];
 
   if (!gppSection) {
-    if (experience?.gpp_settings?.us_approach === GPPUSApproach.STATE) {
+    if (
+      (forcedGpp && !experience?.gpp_settings) ||
```
eastandwestwind, line 89: 👍

clients/fides-js/src/lib/consent-types.ts
```diff
@@ -117,6 +117,9 @@ export interface FidesInitOptions {
 
   // Defines default primary color for consent components, but can still be overridden with overrides or custom CSS
   fidesPrimaryColor: string | null;
+
+  // Whether the developer forced the inclusion of the GPP extension
+  forceGpp: boolean;
```
eastandwestwind, line 122: Is this the only var in  `FidesInitOptions` that cannot be set via an .env var? If so, can we comment this?

I think we should also comment that it can only be set via an explicit query param.

eastandwestwind (changes requested):
Looking good, just some small asks.

Additionally, do we have either a ticket for or any existing plans for adding this new interface to docs?

clients/fides-js/src/fides-ext-gpp.ts
gilluminate: 🤔 one is more like `isForced` where we're checking for the property and the other is more like `doForce` where we're making the assignment so the naming difference was intentional like that, but I can see how that's maybe a little confusing.

clients/fides-js/src/lib/consent-types.ts
gilluminate: technically, it's the only option that can be set as a query param on the `script` tag itself. I discussed this with Michael and going to include it as a section on the FidesOptions documentation. Great callout.

gilluminate:
@eastandwestwind Thanks for this feedback! Great callouts. I added docs and comments for this, as well as recommended adjustments.

clients/fides-js/src/lib/consent-types.ts
eastandwestwind: ah gotcha, thanks for including in the docs!

clients/fides-js/docs/interfaces/FidesOptions.md
```diff
@@ -162,3 +162,13 @@ overriden at the page-level as needed. Only applicable to a TCF experience.
 For more details, see the [TCF CMP API technical specification](https://github.com/InteractiveAdvertisingBureau/GDPR-Transparency-and-Consent-Framework/blob/master/TCFv2/IAB%20Tech%20Lab%20-%20CMP%20API%20v2.md#what-does-the-gdprapplies-value-mean)  * 
 
 Defaults to `true`.
+
+## Forcing GPP Inclusion
```
eastandwestwind, line 166: thanks for adding this!

eastandwestwind (approved):

PR #4812, opened 2024-04-17
Added a `reinitialize` method to Fides.js

guncha:
Closes https://ethyca.atlassian.net/browse/PROD-1967

### Description Of Changes

Adds a new `reinit()` method to FideJS global as well as a `prevConfig` property that stores the configuration that was used to call `init`. There are also minor changes to the `init` functions to treat them like instance methods, which they actually are, instead of programming with globals.

Opening this now to get some early feedback. This is being used in my WT branch to "move" the embed to the slideout and back. It works fine so I don't expect any big changes though perhaps we want to consider supporting multiple embeds per page (this already works basically, but would need to be more carefully considered). We can discuss in the forthcoming PR.

### Pre-Merge Checklist

* [x] Update `CHANGELOG.md`


clients/fides-js/src/fides-tcf.ts
```diff
@@ -104,7 +100,8 @@ const updateExperience = ({
 /**
  * Initialize the global Fides object with the given configuration values
  */
-const init = async (config: FidesConfig) => {
+async function init(this: FidesGlobal, config: FidesConfig) {
```
NevilleS, line 103: Hah, wow - TIL about Typescript `this` parameters: https://www.typescriptlang.org/docs/handbook/2/functions.html#declaring-this-in-a-function

Interesting

clients/fides-js/src/fides-tcf.ts
```diff
@@ -172,7 +172,7 @@ const init = async (config: FidesConfig) => {
   // Initialize the CMP API early so that listeners are established
   initializeTcfCmpApi();
   if (initialFides) {
-    Object.assign(_Fides, initialFides);
+    Object.assign(this, initialFides);
```
NevilleS, line 175: Explain why you think this is preferable - and are we sure it's safe to refactor like this?

clients/fides-js/src/fides.ts
```diff
@@ -47,20 +47,11 @@ declare global {
   }
 }
 
-// The global Fides object; this is bound to window.Fides if available
-// eslint-disable-next-line no-underscore-dangle,@typescript-eslint/naming-convention
-let _Fides: FidesGlobal;
-
-const updateExperience = ({
+const updateExperience: UpdateExperienceFn = ({
```
NevilleS, line 50: 🙏 bless

clients/fides-js/src/lib/consent-types.ts
```diff
@@ -137,7 +137,9 @@ export interface FidesGlobal extends Fides {
   saved_consent: NoticeConsent;
   gtm: typeof gtm;
   identity: FidesJSIdentity;
+  prevConfig: FidesConfig | undefined;
   init: (config: FidesConfig) => Promise<void>;
+  reinit: () => Promise<void>;
```
NevilleS, line 142: two nits:
1. Can we use `.reinitialize` instead? I did some polling of function names and I think that's the winner and I'm not in love with the overly-abbreviated `reinit` (mostly since `init()` isn't really expected to be used by a developer: https://ethyca.com/docs/dev-docs/js/reference/interfaces/Fides#init)
2. Can you update the `Fides` type imported from the `docs/` module to include this function and document there 👍

NevilleS (commented):
This seems pretty reasonable so far - left some questions and nits.

clients/fides-js/src/fides-tcf.ts
```diff
@@ -185,14 +185,15 @@ const init = async (config: FidesConfig) => {
     updateExperience,
     overrides,
   });
-  Object.assign(_Fides, updatedFides);
+  Object.assign(this, updatedFides);
```
gilluminate, line 188: aside from the `this` conversation...it also seems odd to me that we're mixing both spread operators and Object.assign in the same codebase. Couldn't we just use spread like config does, or am I missing something?
```suggestion
  this = {...this, ...updatedFides};
```

Maybe `Object.assign` was originally used to avoid the need for `// eslint-disable-next-line no-param-reassign`?

Consistency is my main concern.

clients/fides-js/src/fides-tcf.ts
guncha: This is preferable because it's closer to functional style which is preferable because it makes code easier to reason about and maintain. By functional I mean functions that take all the data they operate on as arguments and just return values, without modifying any external state (including the arguments). Pure idempotent functions are the best, which `init()` certainly isn't, but it's a matter of degree. In JS as well as C++ and any other OOP language I know, `this` is just another argument to the function so `Fides.init()` is implemented as `init(Fides)`.

I also makes it easier to unit test when you're not relying on some captured state. We could import the `init` function in a test file and call it repeatedly on different objects to test its behavior.

As for safety, I think that's a very relevant question. If you're working with global state, you have to consider any other code that might be using and changing it. The more visible some state is, the more code you have to audit, with the global state being the most visible of all. Luckily, `_Fides` is only visible to this module so it's not too bad. The only possible unsafety I can think of with this change is if someone is doing something like this:

```ts
const { init } = Fides;
init();
```

But they really shouldn't be doing that in the first place 😄

clients/fides-js/src/fides-tcf.ts
guncha: Ah, spreading wouldn't work with the current code because you cannot replace `this` with a new value. I used spread with the `config` to avoid modifying the argument passed to the function because it's what gets stored in `prevConfig`. Otherwise, using `reinit()` would call `init` with slightly different arguments than it was initially called with, which, while it worked fine locally, is not something I'd want other developers to discover and have to keep in mind when working on this.

clients/fides-js/src/lib/consent-types.ts
guncha: Agreed, done and done.

clients/fides-js/src/fides-tcf.ts
gilluminate: Thanks for that explanation. So if I understand correctly, spreading would have been ok with the original code, but the `this` usage makes it inappropriate.

clients/fides-js/src/fides.ts
```diff
@@ -175,6 +172,13 @@ _Fides = {
   saved_consent: {},
   gtm,
   init,
+  prevConfig: undefined,
+  reinitialize() {
+    if (!this.prevConfig || !this.initialized) {
+      throw new Error("Fides must be initialized before reinitializing");
+    }
+    return this.init(this.prevConfig);
```
NevilleS, line 180: I know this is maybe really Hard™️, but I think we need to ensure that `Fides.reinitialize` cleans up our existing rendered DOM elements automaticaly... 😄 

I wrote some quick Cypress specs to exercise this and toggled `fides_embed` off and on. Since `reinitialize` doesn't wipe previously rendered elements, I could get into a fun state where I had both an embedded and overlay banner running simultaneously:
image

NevilleS (changes requested):

clients/fides-js/src/fides.ts
guncha: What do you mean, that's a FEATURE!

NevilleS: OK, I added some code to `initOverlay` to unmount previously rendered instances, which makes this work swimmingly.

NevilleS:
OK @guncha, this is ready for your re-review. I think it looks pretty good...

NevilleS (approved):
This looks solid to me... if @guncha is OK with the edits I made, I think it's ready to ship

clients/fides-js/src/lib/consent-types.ts
```diff
@@ -130,21 +130,21 @@ export interface FidesInitOptions {
...
 export interface FidesGlobal extends Fides {
+  config?: FidesConfig;
   consent: NoticeConsent;
   experience?: PrivacyExperience | EmptyExperience;
-  geolocation?: UserGeolocation;
+  fides_meta: FidesJSMeta;
   fides_string?: string | undefined;
+  geolocation?: UserGeolocation;
+  identity: FidesJSIdentity;
+  initialized: boolean;
```
guncha, line 140: Not sure what the intended order here is, it's alphabetish right now.

clients/fides-js/src/lib/initOverlay.ts
```diff
@@ -9,6 +9,13 @@ import { ColorFormat, generateLighterColor } from "./style-utils";
 const FIDES_EMBED_CONTAINER_ID = "fides-embed-container";
 const FIDES_OVERLAY_DEFAULT_ID = "fides-overlay";
 
+/**
+ * Save a reference to the parent element used to render the overlay. This
+ * allows us to detect re-renders and defensively unmount previous versions of
+ * the overlay from the DOM if this occurs.
+ */
+let renderedParentElem: ContainerNode | undefined;
```
guncha, line 17: This works! If we ever wanted to allow multiple instances of the widget on the same page, this would have to be per-instance, but baby steps 👍

guncha (commented):
LGTM!

clients/fides-js/src/lib/consent-types.ts
NevilleS: Yeah, I know. I'm doing small tidy ups here and there - personally I think the right order should be:
* properties (alphabetized)
* functions (alphabetized)

clients/fides-js/src/lib/initOverlay.ts
NevilleS: Yeah - to do that I'd prefer to have a better, more reliable overall `FidesContext` or something like that to actually store state in a clean way...

PR #4817, opened 2024-04-21
DSR Scheduler Robustness

pattisdr:
Closes #PROD-1750 

### Description Of Changes

Adds some scheduling improvements for DSR 3.0 as a follow-up to the DSR 3.0 core https://github.com/ethyca/fides/pull/4760

### Code Changes

* [ ] Bugfix in logging:  We no longer cache the collection when a Privacy Request fails, we only care about the checkpoint.  Post DSR 3.0, we only cache the failed checkpoint, not the failed collection itself, since with parallelization there could be multiple failed collections. Second, the collection was never actually used in resuming a failed privacy request.
* [ ] Expand the number of places where we cache the failed checkpoint to make it easier to resume a failed Privacy Request from a checkpoint even if we fail in non-standard or surprising ways.  In particular, cache failed checkpoints as we enter every checkpoint in the Request Runner Service, instead of just in a few failure locations.
* [ ] Start caching the latest celery task id for a Request Task in the same location we cache them for Privacy Requests
* [ ] When looking at whether or not a RequestTask can be queued, also take into account whether a celery task for that same Request Task is already in-flight.  This is not super common, but with parallelization it is sometimes possible to have a single node wanting to be queued by multiple upstream nodes around the same time, when the upstream nodes complete around the same time.  This just reduces some potential over-scheduling.  Plus, since I am only caching one celery task id, this better keeps them closer to 1:1 (note that it is okay if a node is queued multiple times, however)
* [ ] The model method to cancel a Privacy Request already could revoke the celery task associated.  Expand this to revoke celery tasks created for request tasks too if applicable.
    * [ ]  Further, add additional logic in the prerequisite checks for running a Request Task that fails early if the overall Privacy Request has been canceled
* [ ] Add an API endpoint to manually re-queue stuck Privacy Requests. This should be used with caution - and is to recover from unexpected failure states.   For now, holding off on adding a process to try to figure out what is "stuck" automatically. I did some experimentation with this, but I think because DSR 3.0 hasn't been used yet, we should monitor to see how frequently if this is happening, as this should be a rare occurrence if you have enough DB connections, resources, etc.  If we start to see this, we should address preventatively first over a task that automatically fixes it after it occurs.

### Steps to Confirm

- In your toml file, toggle DSR 3.0 to on and task_always_eager to True (which means we're not running Celery workers, it's easier to bog down the system if you do this). Also artificially constrain the pool size for the application db.  This should make privacy requests bog down really quickly:
 ```bash
[execution]
use_dsr_3_0 = true
```
```bash
[celery]
task_always_eager = true
```
```
[database]
task_engine_pool_size=1
task_engine_max_overflow=1
```

-  `nox -s dev -- shell postgres`
- In another terminal, `cd /fides/clients`
 - `turbo run dev`
- Setup postgres connector in Admin UI
- http://localhost:3001/add-systems
- Click Add a System
- Add system name: `postgres_example_test_dataset` > Click Save
- Click Integrations tab
- Select Integration type: PostgreSQL
- Add credentials for local postgres example container running in docker and click save(these aren't private secrets, these are for the example postgres database, docker/docker-compose.integration-postgres.yml). 
image
- Click Test integration and verify connection test was successful
- Under Dataset row click on notepad and pencil and upload the contents of `postgres_example_test_dataset.yml` minus the dataset key at the top and click Save
- Submit a couple of access requests with:
```bash
curl -X 'POST' \
  'http://localhost:8080/api/v1/privacy-request' \
  -d '[
  {
    "identity": {
      "email": "customer-1@example.com"
    },
    "policy_key": "default_access_policy"
  }
]'
```
- Depending on your defaults, you should have a couple of Privacy Requests in Approved or in Progress that are stuck - the Request Tasks timed out in the background
- Fix your settings, remove the pool size/max overflow constraints.  Leave `use_dsr_3_0=True`, but add celery task_always_eager = false (so we run workers)
- Bring down the server, rerun with `nox -s dev -- shell postgres worker`
- Requeue the privacy request with the endpoint to fix stuck requests:
Screenshot 2024-04-22 at 9 59 06 AM

```
curl -X 'POST' \
  'http://localhost:8080/api/v1/privacy-request/pri_7b55a69a-294b-4657-97c6-1ef6a711faab/requeue' \
  -H 'Authorization: Bearer XXXX \
  -d ''
```
- Privacy Request should complete
- Verify access package in `fides_uploads` and confirm data is what is expected
- Do other spot-checking of DSR 3.0 and DSR 2.0 - 

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`


src/fides/api/api/v1/endpoints/privacy_request_endpoints.py
```diff
@@ -1805,80 +1790,65 @@ def get_individual_privacy_request_tasks(
...
-    PRIVACY_REQUEST_TASK_QUEUE,
+    PRIVACY_REQUEST_REQUEUE,
     dependencies=[
         Security(verify_oauth_client, scopes=[PRIVACY_REQUEST_CALLBACK_RESUME])
     ],
-    response_model=Dict,
+    response_model=PrivacyRequestResponse,
 )
-def queue_task_manually(
+def requeue_privacy_request(
```
pattisdr, line 1799: DSR 3.0 Core (the feature branch PR) has an endpoint to manually queue an individual Request Task if it was stuck. This was a little limited based on the Privacy Request Status, plus it meant a customer might have to determine multiple tasks that needed to be requeued, a little too much manual effort.

To simplify, I am replacing with another endpoint from which to restart a Privacy Request that is assumed to be stuck.  This should only be used for now if you believe the Privacy Request to be in a bad state, for example, it is stuck in In Processing or Approved.  You should be careful to just use this if you think it is stuck, as we generally only want the Privacy Request in the queue once which is why we are careful about all this Privacy Request Status checking. However, it's possible the Privacy Request Status is incorrect with unexpected failure states.

src/fides/api/api/v1/endpoints/privacy_request_endpoints.py
```diff
@@ -1805,80 +1790,65 @@ def get_individual_privacy_request_tasks(
...
-        raise HTTPException(
-            status_code=HTTP_400_BAD_REQUEST,
-            detail=f"Request failed. Cannot run {request_task.action_type.value} task {request_task.id} when access tasks haven't completed.",
-        )
-    # TODO prevent re-running access task after erasure section has started?
+    # Both DSR 2.0 and 3.0 cache checkpoint details
+    checkpoint_details: Optional[
+        CheckpointActionRequired
+    ] = pr.get_failed_checkpoint_details()
+    resume_step = checkpoint_details.step if checkpoint_details else None
```
pattisdr, line 1825: Both DSR 2.0 and DSR 3.0 use this cached checkpoint if the Privacy Request fails early so we can resume from a specific checkpoint instead of running from the beginning.  

Additionally DSR 3.0 stores Request Tasks that also allow you to deduce at what checkpoint the Privacy Request failed, so that is also added here as a backup.  

An example of why we currently try to resume from checkpoints instead of the beginning: if an erasure request fails inside of the erasure graph, it is possible we've already erased some of the data that would make it difficult to re-run the access section.

src/fides/api/models/privacy_request.py
```diff
@@ -523,14 +524,6 @@ def verify_identity(self, db: Session, provided_code: str) -> "PrivacyRequest":
...
         return self
 
-    def cache_task_id(self, task_id: str) -> None:
-        """Sets a task_id for this privacy request's asynchronous execution."""
-        cache: FidesopsRedis = get_cache()
-        cache.set(
-            get_async_task_tracking_cache_key(self.id),
-            task_id,
-        )
-
```
pattisdr, line 533: This was an unused method, instead we cached this id without having the privacy request in memory from the request runner service.

src/fides/api/models/privacy_request.py
```diff
@@ -680,28 +673,24 @@ def get_paused_collection_details(
     def cache_failed_checkpoint_details(
         self,
         step: Optional[CurrentStep] = None,
-        collection: Optional[CollectionAddress] = None,
```
pattisdr, line 683: We're not actually using the collection to cache failed checkpoint details anywhere so this is being removed altogether

src/fides/api/models/privacy_request.py
```diff
@@ -937,19 +926,46 @@ def pause_processing_for_email_send(self, db: Session) -> None:
...
+            parent_task_id = (
+                self.get_cached_task_id()
+            )  # Celery task for current Privacy Request
+            if parent_task_id:
+                task_ids.append(parent_task_id)
+
+            for celery_task_id in task_ids:
+                logger.info("Revoking task {} for request {}", celery_task_id, self.id)
+                # Only revokes if execution is not already in progress.
+                celery_app.control.revoke(celery_task_id, terminate=False)
```
pattisdr, line 968: The model-level method to cancel privacy requests also tries to revoke the celery task - I've added trying to revoke the associated celery tasks for the Request Task as well. Further, Request Task execution has a new check early on that fails early if the overall Privacy Request was moved to "canceled".

This isn't super likely to get hit as we currently only allow pending Privacy Requests to be cancelled, perhaps because it was tricky to interrupt the Privacy Request running in memory in a reliable way.  We'd probably have more flexibility now to interrupt this with DSR 3.0 if desired.

src/fides/api/models/privacy_request.py
```diff
@@ -1750,15 +1772,79 @@ def get_pending_downstream_tasks(self, db: Session) -> Query:
...
+            logger.debug(
+                "Celery Task {} already processing for {} task {}. Privacy Request: {}, Request Task {}.",
+                celery_task_id,
+                self.action_type.value,
+                self.collection_address,
+                self.privacy_request_id,
+                self.id,
+            )
+
+        return task_in_flight
```
pattisdr, line 1850: I added a light check to reduce some of the over-scheduling if two nodes running in parallel finish about the same time, realize that the other is finished, and go to queue their shared downstream node about the same time.  It is hard to hit this state in testing with workers, but possible.

src/fides/api/service/privacy_request/request_runner_service.py
```diff
@@ -288,15 +282,7 @@ def queue_privacy_request(
...
-    try:
-        cache.set(
-            get_async_task_tracking_cache_key(privacy_request_id),
-            task.task_id,
-        )
-    except DataError:
-        logger.debug(
-            "Error tracking task_id for request with id {}", privacy_request_id
-        )
+    cache_task_tracking_key(privacy_request_id, task.task_id)
```
pattisdr, line 285: Moved this to a method I can share when caching the celery task id for the Request Task

src/fides/api/service/privacy_request/request_runner_service.py
```diff
@@ -416,6 +402,9 @@ def run_privacy_request(
                 request_checkpoint=CurrentStep.upload_access,
                 from_checkpoint=resume_step,
             ):
+                privacy_request.cache_failed_checkpoint_details(
+                    CurrentStep.upload_access
+                )
```
pattisdr, line 407: Now I am more using "cache_failed_checkpoint_details" to update every time a checkpoint is hit, rather than just some selected locations when a Privacy Request fails to have a better understanding of the latest checkpoint even when a Privacy Request fails in unexpected ways.

src/fides/api/task/create_request_tasks.py
```diff
@@ -482,8 +467,8 @@ def run_access_request(
         )
 
     for task in ready_tasks:
-        log_task_queued(task)
-        run_access_node.delay(privacy_request.id, task.id, privacy_request_proceed)
+        log_task_queued(task, "main runner")
+        queue_request_task(task, privacy_request_proceed)
```
pattisdr, line 471: I created a new method, queue_request_task to figure out the particular celery task to run based on the Request Task action type so this could be shared, and also use this method to cache the celery task id in one location.

src/fides/api/task/execute_request_tasks.py
```diff
@@ -51,26 +54,29 @@ def run_prerequisite_task_checks(
             f"Privacy request with id {privacy_request_id} not found"
         )
 
+    if privacy_request.status == PrivacyRequestStatus.canceled:
+        raise PrivacyRequestStatusCanceled(
+            f"Cannot execute request task {privacy_request_task_id} of privacy request {privacy_request_id}: status is {privacy_request.status.value}"
+        )
+
```
pattisdr, line 61: New check to exit a Request Task early if we realize the overall Privacy Request was cancelled.

src/fides/api/api/v1/endpoints/privacy_request_endpoints.py
```diff
@@ -1805,80 +1790,65 @@ def get_individual_privacy_request_tasks(
...
-            status_code=HTTP_400_BAD_REQUEST,
-            detail=f"Request failed. Cannot run {request_task.action_type.value} task {request_task.id} for privacy request {privacy_request.id} of status {privacy_request.status.value}",
-        )
+    Don't use this unless the Privacy Request is stuck.
+    """
+    pr: PrivacyRequest = get_privacy_request_or_error(db, privacy_request_id)
 
-    if request_task.status in COMPLETED_EXECUTION_LOG_STATUSES:
+    if pr.status not in [
+        PrivacyRequestStatus.pending,
```
pattisdr, line 1812: Ah, I should remove this status, looks to me like this can mean the request is awaiting approval.

galvana: Ah, you're right!

src/fides/api/api/v1/endpoints/privacy_request_endpoints.py
```diff
@@ -1805,80 +1790,64 @@ def get_individual_privacy_request_tasks(
...
-        request_task.id,
+        "Manually requeuing Privacy Request {} from step {}",
+        pr,
+        resume_step.value if resume_step else None,
     )
 
-    # Requeuing the particular task now that the callback is received
-    log_task_queued(request_task)
-    task_function(request_task).delay(
-        privacy_request_id=privacy_request.id, privacy_request_task_id=request_task.id
```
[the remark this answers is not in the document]
pattisdr: I just thought it was too time consuming to have the user figure out which task needed to run!  If queued from the right checkpoint, both DSR 2.0 and DSR 3.0 have some checks to try to only re-run what hasn't been run.  I generally don't think people should use this though unless they truly think something is stuck.

galvana:
Ran the manual test steps locally and confirmed the requeue endpoint is working as expected 🙌 
image

src/fides/api/task/execute_request_tasks.py
```diff
@@ -51,26 +54,29 @@ def run_prerequisite_task_checks(
...
-            RequestTask.collection_address.in_(request_task.upstream_tasks or []),
-        )
-
-        if not all(
-            upstream_task.status in COMPLETED_EXECUTION_LOG_STATUSES
-            for upstream_task in upstream_results
-        ) or not upstream_results.count() == len(request_task.upstream_tasks or []):
+        # Only running the upstream check instead of RequestTask.can_queue_request_task since
+        # the node is already queued.
+        if not request_task.upstream_tasks_complete(session, should_log=False):
```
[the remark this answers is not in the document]
pattisdr: I felt like we should be double checking here, especially because this is laying the groundwork to be able to execute individual tasks later in isolation if needed.  For example, the asynchronous task work would likely want to be queuing a specific task, and I just heard another customer use case today related to wanting to do operations on one task. I just want multiple safeguards around this -

src/fides/api/api/v1/endpoints/privacy_request_endpoints.py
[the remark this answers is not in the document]
pattisdr: I do think we'll need to add a different version of being able to queue a specific task soon based on use cases I'm hearing about but I'd rather add that separately alongside specific requirements.

src/fides/api/util/cache.py
```diff
@@ -240,3 +242,56 @@ def get_all_cache_keys_for_privacy_request(privacy_request_id: str) -> List[Any]
...
+    for _, task_details in queried_tasks.items():
+        for _, state_array in task_details.items():
+            state: str = state_array[0]
+            if state in [
+                "active",
+                "received",
+                "reserved",
+                "retry",
+                "scheduled",
+                "started",
```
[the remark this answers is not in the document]
pattisdr: I had a hard time deciding which states should go in this list, it was from a mixture of manual testing and looking at lists of celery states, and methods on celery's Inspect class. I am not positive all of these would get thrown in this location, but if they did, I wanted them to count as being "in flight".  Let me do some more research on this one - 

https://github.com/celery/celery/blob/main/celery/states.py
https://github.com/celery/celery/blob/905ce2e923bb80cbbe28180d91342760c8c53bd3/celery/app/control.py#L71

src/fides/api/models/privacy_request.py
pattisdr: Thank you Adrian!

galvana (approved):
Thanks for answering my questions, this looks good to go!

PR #4837, opened 2024-04-26
DSR 3.0 Bugfix: Approved -> Failed Privacy Requests Don't Change Status

pattisdr:
Closes #PROD-2005

### Description Of Changes

Specifically, **if manual request approval is required** and the **privacy request fails**, the Privacy Request can get stuck in an approved state.  While I tested the "manual request approval" flag in development, the large majority of the "error" testing was done with no approval required.  Good find by Roger!

### Code Changes

* [ ]  Update the existing polling task that checks for Privacy Requests whose sub-tasks have all had a chance to run, but some are errored, and transitions the Privacy Request to "Error" to also look for the "Approved" status. 
    👉  Privacy Requests that are in-flight are expected to be in a "In-Processing" or "Approved" status, and I was only covering the first case. 

### Steps to Confirm
- Setup DSR 3.0 and require manual request approval. The worker doesn't specifically need to be running here, you don't need parallelization to hit this case.
```
[execution]
require_manual_request_approval = true
use_dsr_3_0 = true
```
- `nox -s dev -- shell postgres `
- `cd fides/clients`
- `turbo run dev`
- Add a postgres integration in the frontend with a bad secret.  
- Submit a Privacy Request through the Privacy Center. 
- Previously, this would hang indefinitely in Approved.  Now, this will hang briefly, but at the next poll, (check your terminal), you should see something like:
>  “Polling for privacy requests awaiting errored status change”
> Marking access step of pri_f1433833-31ed-4d6f-b5d9-e415f1e18b1a as error
- This will mark the Privacy Request as errored.
- Fix the secret
- Reprocess the privacy request 
- It should complete

### Pre-Merge Checklist

* Documentation (not needed)
* [x] Update `CHANGELOG.md`


src/fides/api/service/privacy_request/request_service.py
```diff
@@ -187,10 +187,19 @@ def poll_for_exited_privacy_request_tasks(self: DatabaseTask) -> Set[str]:
...
+        # Privacy Requests in these states should be examined to see if all of its Request Tasks have had a chance
+        # to complete.
         in_progress_privacy_requests = (
             db.query(PrivacyRequest)
-            .filter(PrivacyRequest.status == PrivacyRequestStatus.in_processing)
+            .filter(
+                PrivacyRequest.status.in_(
+                    [PrivacyRequestStatus.in_processing, PrivacyRequestStatus.approved]
+                )
+            )
```
pattisdr, line 202: This is the fix: an important detail I missed is that a Privacy Request that is actively in-flight, can have a status of "In Processing" or "Approved".

tests/conftest.py
```diff
@@ -35,6 +31,8 @@
 from fides.api.main import app
 from fides.api.models.privacy_request import (
     EXITED_EXECUTION_LOG_STATUSES,
+    generate_request_callback_pre_approval_jwe,
+    generate_request_callback_resume_jwe,
```
pattisdr, line 35: Just applying the results of `nox -s static_checks`

tests/ops/integration_tests/test_execution.py
```diff
@@ -446,11 +446,8 @@ async def test_skip_collection_new_request(
         assert mongo_logs.filter_by(status="skipped").count() == 9
 
     @mock.patch("fides.api.task.graph_task.GraphTask.log_start")
+    @pytest.mark.usefixtures("use_dsr_2_0")
     @pytest.mark.asyncio
-    @pytest.mark.parametrize(
-        "dsr_version",
-        ["use_dsr_3_0", "use_dsr_2_0"],
-    )
     async def test_run_disabled_collections_in_progress(
```
pattisdr, line 451: I am removing running this flaky test on DSR 3.0. This test was written for DSR 2.0 but it fails occasionally on CI. It is hard to reproduce when you run unit tests locally, you have to be running Python 3.8.x, and it will maybe fail 1/20 of the time.   I just don't think this is a good test here.

src/fides/api/service/privacy_request/request_service.py
galvana: This makes sense!

galvana:
Verified the privacy request status is set to `Approved` and that the job detects the errored request
image

```
2024-04-29 08:51:42 2024-04-29 15:51:42.991 | DEBUG    | fides.api.service.privacy_request.request_service:poll_for_exited_privacy_request_tasks:190 - Polling for privacy requests awaiting errored status change
2024-04-29 08:51:43 2024-04-29 15:51:43.021 | INFO     | fides.api.service.privacy_request.request_service:poll_for_exited_privacy_request_tasks:233 - Marking access step of pri_f32ab96e-05a7-420f-abe9-0cd745d84014 as error
```
image

Reprocessed successfully!
image

I also verified that the request goes to error if the integration isn't fixed but the request is reprocessed.

pattisdr:
Thank you for your review @galvana ⭐

PR #4851, opened 2024-05-01
Search for both email and user_id to get a segment_id

SteveDMurphy:
Closes INTS-294

### Description Of Changes

Includes the option to also collect the segment ID via `email` instead of the `user_id`

I'm unclear if there is any issue with running dual read requests, assuming the returned values are consolidated if (as in our test case) the segment_id is returned for both

### Code Changes

* [ ] add a read request that also queries the email attribute of a segment user profile

### Steps to Confirm

* [ ] validate the test postman request also returns when querying by email

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,9 @@ The types of changes are:
 ### Added
 - Added multiple language translations support for privacy center consent page [#4785](https://github.com/ethyca/fides/pull/4785)
 
+### Changed
+- Included searching by `email` for the Segment integration [#4851](https://github.com/ethyca/fides/pull/4851)
+
 ## [2.35.0](https://github.com/ethyca/fides/compare/2.34.0...2.35.0)
 
 ### Added
--- a/data/saas/config/segment_config.yml
+++ b/data/saas/config/segment_config.yml
@@ -44,35 +44,34 @@ saas_config:
     - name: segment_user
       requests:
         read:
-          method: GET
-          path: /v1/spaces/<namespace_id>/collections/users/profiles/user_id:<user_id>/metadata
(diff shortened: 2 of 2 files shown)


MarcGEthyca:
Steve and I talked about this a bit this am, and I wanted to capture some of that discussion here!

It does look like it may be a better bet to use the 'email' as our identity rather than setting up two read requests that return the same value (the 'segment id') by email or by user_id. One of the concerns is that other customers could end up needing to use the user_id rather than email owing to their own use of Segment, otoh, using email would be super consistent with our overall flow =). We could also add this flexibility later if it is needed.

We set up a local saas_config.toml for Steve and tried running the tests for segment but while we managed to get past a few of the errors, we may need to look further to isolate and resolve these errors.

galvana (approved):
Hey Steve, thanks for adding the new endpoint! I just updated the formatting for how you specify a list of `read` requests, take a look. As for your question about deduplicating results, yes, if a user if found by both `user_id` and `email` lookups, we will still only send one `segment_id` to the downstream collections. This is a recent introduction so I'm glad we got it just in time for this 😄

SteveDMurphy:
> Hey Steve, thanks for adding the new endpoint! I just updated the formatting for how you specify a list of `read` requests, take a look. As for your question about deduplicating results, yes, if a user if found by both `user_id` and `email` lookups, we will still only send one `segment_id` to the downstream collections. This is a recent introduction so I'm glad we got it just in time for this 😄 

Thank you for this @galvana !! Your change makes perfect sense for the list formatting - thanks for pushing it 💙

PR #4853, opened 2024-05-01
Persist the state of the reporting table so that viewers can return to their previous view easily

gilluminate:
Closes [PROD-1907](https://ethyca.atlassian.net/browse/PROD-1907)

### Description Of Changes

This modifies the existing Data Map Reporting page to keep track of the various states in Local Storage in addition to the react state. This will allow the user to reload the page, or close the page and come back to it, and not need to adjust these settings again.

This includes storing

- The overall "Group by" status
- Each column’s “Group/Display” status
- Each column’s width
- Column visibility
- Column order

### Code Changes

* created a new hook `useLocalStorage` in AdminUI client that is a modified version of the one in FidesJS.
* converted or created state management over to local storage for the following:
```
export enum DATAMAP_LOCAL_STORAGE_KEYS {
  GROUP_BY = "datamap-group-by",
  COLUMN_ORDER = "datamap-column-order",
  TABLE_GROUPING = "datamap-table-grouping",
  TABLE_STATE = "datamap-report-table-state",
  DISPLAY_ALL_COLUMNS = "datamap-display-all-columns",
}
```

Some things needed to be updated to accept initialized values from the storage, since many were just using hard coded defaults:

* Column ordering modal now respects column ordering when initializing
* The `tableInstance` is now initialized using the stored state if there is one.

### Steps to Confirm

1. Visit Data map report page in Admin UI at `/reporting/datamap`
2. Resize a column or two
3. Change the "Group by..." to something different than the default (eg. "Data Use")
4. Open the "Edit Columns" modal and hide a few columns as well as reording a few columns
5. Change one or two of the columns to "Display All" from the heading dropdown menu.
6. Now reload the page and ensure that all of the above changes have been remembered 

### Pre-Merge Checklist

* [x] Update `CHANGELOG.md`

[PROD-1907]: https://ethyca.atlassian.net/browse/PROD-1907?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ The types of changes are:
 - Added access and erasure support for Marigold Engage by Sailthru integration [#4826](https://github.com/ethyca/fides/pull/4826)
 - Added multiple language translations support for privacy center consent page [#4785](https://github.com/ethyca/fides/pull/4785)
 - Added ability to export the contents of datamap report [#1545](https://ethyca.atlassian.net/browse/PROD-1545)
+- Added state persistence across sessions to the datamap report table [#4853](https://github.com/ethyca/fides/pull/4853)
 
 ### Fixed
 - Remove the extra 'white-space: normal' CSS for FidesJS HTML descriptions [#4850](https://github.com/ethyca/fides/pull/4850)
--- a/clients/admin-ui/cypress/e2e/datamap-report.cy.ts
+++ b/clients/admin-ui/cypress/e2e/datamap-report.cy.ts
@@ -82,6 +82,19 @@ describe("Minimal datamap report table", () => {
     cy.getByTestId("datamap-report-heading").should("be.visible");
   });
 
+  it("can group by data use", () => {
+    cy.getByTestId("group-by-menu").should("contain.text", "Group by system");
+    cy.getByTestId("group-by-menu").click();
+    cy.getByTestId("group-by-menu-list").within(() => {
+      cy.getByTestId("group-by-data-use-system").click();
+    });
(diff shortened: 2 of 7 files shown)


jpople (commented):
Code looks good!  UAT'd:

* Changes to column visibility, width, and order are working as expected
* The "group by" state for the whole table is stored, but not "group/display" for individual columns (e.g. the data categories column), which I think is meant to be part of the AC:

> Components to include in the state
> 
> Each column’s “Group/Display” status

Also, is this something we can add a Cypress test for?

gilluminate:
> * The "group by" state for the whole table is stored, but not "group/display" for individual columns (e.g. the data categories column), which I think is meant to be part of the AC:

Ah, good catch! Yes, I misunderstood the AC there. I'll add or update to include this.

gilluminate:
> Also, is this something we can add a Cypress test for?

I can look into that. Might need to figure out how to persist the Cypress environment's local storage between page loads.

gilluminate:
@jpople updated to address your comments. Thanks for those!

I verified that we can go ahead and keep my accidental inclusion of the "Group by" state, in addition to the rest. I've update the AC to reflect that.

I added some Cypress tests for verifying persistence on page reload. I didn't include column sizing or column reordering because a) they weren't included in the original tests, b) I think they were left off originally because it would be difficult to simulate dragging items to specific spots and check that they are in the correct spot using Cypress alone. Neither task can be accomplished without a mouse at the moment. We should probably fix that for accessibility in the future, which will in turn enable better testing.

jpople (approved):

PR #4855, opened 2024-05-02
Update the get_cached_identity_data function to support non-JSON-encoded value

galvana:
Closes [PROD-2009](https://ethyca.atlassian.net/browse/PROD-2009)

### Description Of Changes

This is a fix for a defect that was introduced with the custom identities feature, specifically [this line](https://github.com/ethyca/fides/pull/4764/files#diff-ef7d3e10be42d5a81721075db1647f7a59ad17ab20b273bfcda1edbcbc8ba391R356). Moving forward, we want to make sure that any non-JSON-encoded values in the cache can still be read. There is no data lost between the two caching methods since we're just formatting the values differently. This means the change can be targeted and everything can continue to work as expected as long as `get_cached_identity_data` returns the expected response.

### Code Changes

* [ ] Updated `get_cached_identity_data` and added a test

### Steps to Confirm

* [ ] This one is tricky to verify since it involves queuing a privacy request pre-2.34, upgrading to 2.34 then approving the request. I just wrote a test to make sure we could reproduce the issue reported in prod and verify the fix.

### Pre-Merge Checklist

* [ ] Update `CHANGELOG.md`

[PROD-2009]: https://ethyca.atlassian.net/browse/PROD-2009?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ The types of changes are:
 
 ### Fixed
 - Remove the extra 'white-space: normal' CSS for FidesJS HTML descriptions [#4850](https://github.com/ethyca/fides/pull/4850)
+- Fixing issue where privacy requests not approved before upgrading to 2.34 couldn't be processed [#4855](https://github.com/ethyca/fides/pull/4855)
 
 
 ## [2.35.0](https://github.com/ethyca/fides/compare/2.34.0...2.35.0)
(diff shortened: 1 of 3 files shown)


adamsachs (approved):
this makes sense to me @galvana! just one minor nit on potentially adding some more context around the code comment, to be extra sure that this isn't confusing for anyone looking at this moving forward.

also, as i mentioned separately - it was a bit surprising to me that we're keeping the cache in tact across upgrades, but i guess that makes sense. in any case, this is a good, defensive measure 👍

PR #4858, opened 2024-05-02
Fix handling of some ISO-3166 geolocation edge cases in Privacy Center /fides.js endpoint

NevilleS:
Closes PROD-2015

### Description Of Changes

This fixes a subtle error where the Privacy Center would consider some valid ISO-3166 codes as invalid if they had single-character regions; for example, `SE-O` would be rejected. The fix here is trivial (accept 1-3 characters!) but this PR also adds additional test coverage for confidence, and also adds an extra defensive check to fallback to just the country header if an invalid CloudFront region header is detected.

### Code Changes

* [X] Update geolocation unit tests
* [X] Fix ISO-3166 regex in `geolocation.ts`
* [X] Discard invalid CloudFront region headers and fallback to country-only

### Steps to Confirm

* [X] Run test suite 👍 

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### DIFF
--- a/clients/privacy-center/common/geolocation.ts
+++ b/clients/privacy-center/common/geolocation.ts
@@ -1,11 +1,15 @@
 import type { NextApiRequest } from "next";
 import { UserGeolocation } from "fides-js";
 
-// Regex to validate a location string, which must:
+// Regex to validate an ISO-3166 location string, which must:
 // 1) Start with a 2-3 character country code (e.g. "US")
-// 2) Optionally end with a 2-3 character region code (e.g. "CA")
+// 2) Optionally end with a 1-3 character region code (e.g. "CA")
 // 3) Separated by a dash (e.g. "US-CA")
-const VALID_ISO_3166_LOCATION_REGEX = /^\w{2,3}(-\w{2,3})?$/;
+const VALID_ISO_3166_LOCATION_REGEX = /^\w{2,3}(-\w{1,3})?$/;
+
+// Regex to validate a standalone ISO-3166-2 region code, which must be a 1-3
+// character code (e.g. "CA")
+const VALID_ISO_3166_2_REGION_REGEX = /^\w{1,3}?$/;
 
 // Constants for the supported CloudFront geolocation headers
 // (see https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/adding-cloudfront-headers.html#cloudfront-headers-viewer-location)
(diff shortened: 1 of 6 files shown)


gilluminate:
@NevilleS this all looks good, but we also do a validation on the `fides-js-demo.html` and `fides-js-components-demo.html` pages. Is that still necessary? If so, do those need to be updated to reflect these same changes?

clients/privacy-center/common/geolocation.ts
```diff
@@ -1,11 +1,15 @@
...
 import { UserGeolocation } from "fides-js";
 
-// Regex to validate a location string, which must:
+// Regex to validate an ISO-3166 location string, which must:
 // 1) Start with a 2-3 character country code (e.g. "US")
-// 2) Optionally end with a 2-3 character region code (e.g. "CA")
+// 2) Optionally end with a 1-3 character region code (e.g. "CA")
 // 3) Separated by a dash (e.g. "US-CA")
-const VALID_ISO_3166_LOCATION_REGEX = /^\w{2,3}(-\w{2,3})?$/;
+const VALID_ISO_3166_LOCATION_REGEX = /^\w{2,3}(-\w{1,3})?$/;
```
RobertKeyser, line 8: nit: _technically_ the `\w` metacharacter in regex would match some values that are invalid, e.g. `_` and also country codes that starts with a digit. 

This is how the regex was already built, so it doesn't make anything _worse_ as written. It's also of minimal risk IMO, but I just wanted to call it out.

The fix would be to replace your `\w` metacharacters with a set: `[A-Za-z]`

clients/privacy-center/common/geolocation.ts
```diff
@@ -1,11 +1,15 @@
...
+
+// Regex to validate a standalone ISO-3166-2 region code, which must be a 1-3
+// character code (e.g. "CA")
+const VALID_ISO_3166_2_REGION_REGEX = /^\w{1,3}?$/;
```
RobertKeyser, line 12: Same here re: `\w` matching.

clients/privacy-center/common/geolocation.ts
RobertKeyser: Might also worth adding it to the unit tests 😄

NevilleS: Er my last comment is a bit backwards, low reading comprehension there - but yes, I do think I should reject numeric country codes, that's just another subtle footgun!

NevilleS: OK, updated! I think it's preferable to be defensive here and not overly permissive - the change to discard invalid  region headers is there as a safety hatch, but I think it's good to be strict about the `?geolocation=` param since that's a user-facing API

NevilleS:
> @NevilleS this all looks good, but we also do a validation on the `fides-js-demo.html` and `fides-js-components-demo.html` pages. Is that still necessary? If so, do those need to be updated to reflect these same changes?

I don't think it was ever really needed on those pages... removing for consistency, so that our tests don't silently discard that and hide potential issues...!

Pushed up and re-running.

gilluminate (approved):
Approved without looking too closely at the validation specifics since @RobertKeyser is the expert there.

PR #4864, opened 2024-05-04
Use CheckboxTree for data map filtering

gilluminate:
Closes [PROD-1916](https://ethyca.atlassian.net/browse/PROD-1916)

### Description Of Changes

Update the filters on the table to apply a modified version of the hierarchy component that we are using on the dataset page. 

### Code Changes

* New `DatamapReportFilterModal` which reduces a lot of over-engineered complexity and implements the `CheckboxTree` component
* Update old `getQueryParamsFromList` as new `getQueryParamsFromArray` utility
* Update checkbox tree to have animated caret similar to accordion.
* Unit test and e2e test

### Steps to Confirm

* visit Data Map Report `/reporting/datamap`
* click the "Filter" button in the top bar
* expand one of the 3 categories
* Note the new style for checking boxes uses our existing tree component rather than just a flat list of check boxes within each of the 3 main sections
* Expand/collapse and click around.
* When desired filter is selected, click the "Done" button. This will apply the filter to the table.
* Open the filter modal again using the "Filter" button.
* Click the "Reset filters" button which will clear all filters.

### Pre-Merge Checklist

* [x] Update `CHANGELOG.md`

[PROD-1916]: https://ethyca.atlassian.net/browse/PROD-1916?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/clients/admin-ui/src/features/common/utils.ts
+++ b/clients/admin-ui/src/features/common/utils.ts
@@ -48,12 +48,24 @@ export const getFileNameFromContentDisposition = (
   return match ? match[1] : defaultName;
 };
 
+/**
+ * Constructs a query string from an array of values.
+ *
+ * @param valueList - An array of string values to be included in the query string.
+ * @param queryParam - The name of the query parameter.
+ * @returns A query string where each value from the array is assigned to the query parameter.
+ * If the array is empty, the function returns undefined.
+ *
+ * @example
+ * getQueryParamsFromArray(['1', '2', '3'], 'id');
+ * // returns 'id=1&id=2&id=3'
+ */
 export const getQueryParamsFromArray = (
-  strings: string[],
+  valueList: string[],
   queryParam: string
 ) => {
-  if (strings.length > 0) {
-    return `${queryParam}=${strings.join(`&${queryParam}=`)}`;
+  if (valueList.length > 0) {
+    return `${queryParam}=${valueList.join(`&${queryParam}=`)}`;
   }
   return undefined;
 };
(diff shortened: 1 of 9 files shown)


clients/admin-ui/src/features/common/utils.ts
```diff
@@ -47,3 +47,13 @@ export const getFileNameFromContentDisposition = (
   const match = contentDisposition.match(/filename=(.+)/);
   return match ? match[1] : defaultName;
 };
+
+export const getQueryParamsFromArray = (
+  strings: string[],
```
eastandwestwind, line 52: can we re-name `strings` to something more specific, and also add a docstring to this fn?

eastandwestwind (changes requested):
Tested this, and it works great! Couple small requests below.

Also here's some things I've noticed that I'm unsure if caused by this PR, but would like to double check:

1. The vertical spacing between items within the accordions seems uneven:

Screenshot 2024-05-07 at 5 45 20 PM

2. The `group_by` query param seems to have some encoding issues in the resulting API call:
Screenshot 2024-05-07 at 5 49 43 PM

Let me know and let's re-ticket items as needed!

gilluminate:
> 1. The vertical spacing between items within the accordions seems uneven:

I think that was existing, but it's an easy fix. Great catch.

gilluminate:
> 2. The `group_by` query param seems to have some encoding issues in the resulting API call:

That's also existing behavior, and I'm worried about breaking it for the backend. Might be best to re-ticket that one as you suggested.

eastandwestwind (approved):
Thanks for making those changes! 

Mind creating a new ticket for the query param formatting?

PR #4878, opened 2024-05-09
Remove response limit from fides.js

andres-torres-marroquin:
Closes #PROD-2031

### Description Of Changes

By default Next.JS throws "API Routes Response Size Limited to 4MB" when the response is larger than 4MB, disabling this will not throw data in the console. But this can allow us to have very big responses without noticing anymore. From the Next.JS docs: API Routes in Next.js are designed to deliver quick responses and are not built to support the transmission of large amounts of data.

### Code Changes

* [ ] Added a `config` variable to `fides-js.ts` to change the `api.responseLimit` to `false`

### Steps to Confirm

* [ ] Change the `script` variable on `fides-js.ts` file to include a lot of comments, just 4MB of it, and you'll get a message on the dev web server shell. After this change gets into main, the message will be gone.

### Pre-Merge Checklist

* [ ] Update `CHANGELOG.md`


galvana (approved):
This works as expected! I added a long description to one of the banner descriptions and I was able to create a payload over 4MB in size from `fides.js`. The error doesn't actually prevent the download but removing this limit eliminates this false alarm that might cause concern for some users.

PR #4879, opened 2024-05-09
Update property model

galvana:
Closes [PROD-2025](https://ethyca.atlassian.net/browse/PROD-2025)

### Description Of Changes

Adds `privacy_center_config`, `stylesheet`, and `paths` to the `Property` model.

### Code Changes

* [ ] Expanded the `PrivacyCenterConfig` schema to fully implement the schema as it is defined in the `privacy_center`  client
* [ ] Moved the CSS validator from fidesplus into fides and added a new `CssStr` custom type with CSS validation
* [ ] Added `privacy_center_config` and `stylesheet` fields to the `plus_property` table
* [ ] Added `plus_property_path` table to track the Property paths

### Steps to Confirm

* [ ] See testing steps in https://github.com/ethyca/fidesplus/pull/1423

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

[PROD-2025]: https://ethyca.atlassian.net/browse/PROD-2025?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/fides/api/schemas/privacy_center_config.py
```diff
@@ -28,16 +51,85 @@ def validate_default_value(cls, values: Dict[str, Any]) -> Dict[str, Any]:
...
+
+class ConsentConfigPage(FidesSchema):
+    consentOptions: List[ConfigConsentOption]
+    description: str
+    description_subtext: Optional[List[str]]
+    policy_key: Optional[str]
+    title: str
+
+    @root_validator(pre=True)
+    def validate_consent_options(cls, values: Dict[str, Any]) -> Dict[str, Any]:
```
adamsachs, line 100: again, i think this could be a bit more defensive given that it's a `pre` validator. i realize that the values here may be constrained by the FE client, but better to make as few assumptions as possible about the input values to keep the API robust.

tests/ops/models/test_property.py
```diff
@@ -29,24 +42,62 @@ def minimal_experience(
...
+                    name="Second Property",
+                    type=PropertyType.website,
+                    experiences=[minimal_experience],
+                    privacy_center_config=privacy_center_config,
+                    stylesheet=":root:root { --chakra-colors-gray-50: #fff9ea; }",
+                    paths=["test"],
+                ).dict(),
+            )
+
+        first_prop.delete(db)
```
adamsachs, line 99: haven't verified this test specifically, but from what i can see with testing on the API, i think the `Second Property` actually gets created here _before_ the `IntegrityError` is raised, so i think this test could/should be adjusted to ensure that the `Second Property` isn't actually created!

adamsachs (changes requested):
@galvana this generally looks on the right track, but i've got some points that i think would be good to address to make these models and the API a bit more robust. let me know your thoughts or any questions!

src/fides/api/models/property.py
```diff
@@ -60,21 +73,62 @@ def create(
...
         experiences = data.pop("experiences", [])
+        paths = data.pop("paths", None) or []
+
         prop: Property = super().create(db=db, data=data, check_name=check_name)
         link_experience_configs_to_property(
             db, experience_configs=experiences, prop=prop
         )
-        return prop
+
+        property_paths = [
```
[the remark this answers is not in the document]
galvana: Checking existing `PropertyPaths` and raises an exception if the paths are already associated with another Property. Updated tests.

src/fides/api/models/property.py
```diff
@@ -60,21 +73,62 @@ def create(
...
+
+        # delete PropertyPath instances that are not in the updated paths
+        for path in set(existing_paths.keys()) - updated_paths:
+            db.delete(existing_paths[path])
+
+        # create new PropertyPath instances for paths that don't exist
+        for path in updated_paths - set(existing_paths.keys()):
+            self._property_paths.append(PropertyPath(property_id=self.id, path=path))
+
+        self.save(db)
```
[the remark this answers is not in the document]
galvana: Similar fix as above

tests/ops/models/test_property.py
galvana: Updated test

src/fides/api/models/property.py
```diff
@@ -60,21 +74,82 @@ def create(
...
+        paths = data.pop("paths", None) or []
+
+        matching_paths = PropertyPath.filter(
+            db=db, conditions=(PropertyPath.path.in_(paths))
+        ).all()
+        if matching_paths:
+            raise KeyOrNameAlreadyExists(
+                f'The path(s) \'{", ".join([matching_path.path for matching_path in matching_paths])}\' are already associated with another property.'
+            )
+
```
adamsachs, line 86: very nice! as a nit, i might move this into a helper function to share between create and update in case the logic ever needs to be adjusted, but i'll let you decide whether or not to add that last bit of polish :)

src/fides/api/schemas/property.py
```diff
@@ -29,3 +33,10 @@ class Property(FidesSchema):
     type: PropertyType
     id: Optional[str] = None
     experiences: List[MinimalPrivacyExperience]
+    privacy_center_config: Optional[PrivacyCenterConfig]
+    stylesheet: Optional[CssStr]
+    paths: List[str]
```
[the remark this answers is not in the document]
adamsachs: ah ok! makes sense, thanks for clarifying.

tests/ops/models/test_property.py
```diff
@@ -29,43 +45,92 @@ def minimal_experience(
...
+                    privacy_center_config=privacy_center_config,
+                    stylesheet=":root:root { --chakra-colors-gray-50: #fff9ea; }",
+                    paths=["test"],
+                ).dict(),
+            )
+
+        second_prop = Property.filter(
+            db=db, conditions=(Property.name == "Second Property")
+        ).first()
+        assert second_prop is None
```
adamsachs, line 105: 🎊

tests/ops/schemas/test_property_schema.py
adamsachs, line 1: these are great! nice job covering these edge cases 👍

adamsachs (approved):
this is looking good to me @galvana! thanks for addressing the feedback 👍

PR #4881, opened 2024-05-11
2.27.a14 Special Image Prep

pattisdr:
Closes #PROD-1590

### Description Of Changes

This PR contains experimental work into a 2.27.0+ upgrade where we separate out operations from the privacy preferences schema and data migration that are touching every row, and extracts them into a script that could be run while the application was live to improve migration speeds.  A corresponding plus version would need to be created that didn't assume certain fields in consent reporting existed (likely Pydantic changes) until we could get the consent reporting fields back-filled. 

### Code Changes

### Steps to Confirm

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`


src/fides/api/alembic/migrations/versions/f9b28f36b53e_privacy_preference_v2.py
```diff
@@ -214,12 +233,44 @@ def upgrade():
...
+    op.drop_index(
+        op.f("ix_privacypreferencehistory_preference"),
+        table_name="privacypreferencehistory",
+    )
+    op.create_index(
+        op.f("ix_privacypreferencehistory_preference_old_column"),
+        "privacypreferencehistory",
+        ["preference"],
+        unique=False,
+    )
```
adamsachs, line 247: so is this simply to change the name of the index and free up that namespace for the index to be created on the new column (i.e. in L263-L268)? just wanna make sure i'm following!

src/fides/api/alembic/migrations/versions/f9b28f36b53e_privacy_preference_v2.py
```diff
@@ -214,12 +233,44 @@ def upgrade():
...
+        op.f("ix_privacypreferencehistory_preference"),
+        "privacypreferencehistory",
+        ["preference"],
+        unique=False,
+    )
+
+    logger.info(
+        "Added new empty preference column of new varchar type. Existing preferences stored under 'preference_old_column'."
     )
+
```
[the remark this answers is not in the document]
adamsachs: thank you for spelling this out! super clear here 👍

scripts/manual_privacy_preference_v2_data_migration.py
```diff
@@ -0,0 +1,152 @@
...
+import time
+
+from loguru import logger
+
+from fides.api.db.session import get_db_session
+from fides.config import CONFIG
+
+"""
+Run in shell with:
+python3 scripts/manual_privacy_preference_v2_data_migration.py -table privacypreferencehistory
```
adamsachs, line 11: nit: not sure exactly how this will be executed to go against the customer DBs, but it may be nice to allow config options to be passed in via CLI arg? or at least without requiring `fides` to be installed in the python env on which they're executed? 

but may not be a big deal if we know we've got a shell that can run this with the required setup.

src/fides/api/alembic/migrations/versions/f9b28f36b53e_privacy_preference_v2.py
adamsachs: OK so to recap - this operation is now needed because we now have 2 columns (old + new) in place of one, and we want to make sure the new column index is named as we expect - so we need to delete and recreate the old index, since we still need that index, but need to allow the index on the _new_ column to be named appropriately. right?

my first instinct was to just create the new index on the new column with a different name, to save us these drop + re-create index operations here on the old column, but then we'd still need to adjust that new index name on the new column at _some_ point so that it's consistent with our mainline codebase ... so may as well just do that now.

i think that all makes sense, just talking myself through it out loud to verify! :)

src/fides/api/alembic/migrations/versions/f9b28f36b53e_privacy_preference_v2.py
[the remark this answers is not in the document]
adamsachs: nit: why not switch around the order a bit (there may be a good reason that i'm just not realizing):

- I'm renaming the old column
- I'm adding a new column with the desired type (and just name it `preference`)
~- I'm renaming the new column to "preference"~

adamsachs (commented):
some minor notes/questions/clarifications for my understanding, but generally this makes sense to me! 👍

src/fides/api/alembic/migrations/versions/f9b28f36b53e_privacy_preference_v2.py
pattisdr: yes this what I was thinking @adamsachs, the goal to get this index name updated as part of this migration while we were taking downtime

src/fides/api/alembic/migrations/versions/f9b28f36b53e_privacy_preference_v2.py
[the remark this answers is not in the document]
pattisdr: I'll look into this thank you! that sounds faster, I can't remember if I did this intentionally

scripts/manual_privacy_preference_v2_data_migration.py
pattisdr: Can you give me an example of what you mean here? What would that look like.

adamsachs: oh just thinking something like 

```
python3 scripts/manual_privacy_preference_v2_data_migration.py -table privacypreferencehistory -database_uri [DATABSE_URI_HERE]
```

to make this script a bit more "standalone" (i.e. fewer dependencies), but i suppose we'd still probably want to be able to use the `get_db_session` helper from `fides` anyway, so my suggestion may not be a good one, i.e. may just be better to ensure we can execute the script from a python env with `fides` installed and with the right config env vars set, that shouldn't be too hard :)

pattisdr: thank you for elaborating!

PR #4884, opened 2024-05-13
Improve location and property_id error response to return 400 status instead of 500 server error

andres-torres-marroquin:
Closes [PROD-2035](https://ethyca.atlassian.net/browse/PROD-2035)

### Description Of Changes

Improved validation error messages for `fides.js`.

### Code Changes

* [ ] Improved error message when `geolocation` param is provided but is invalid, e.g. `Invalid request: geolocation "foo" is invalid, must supply an ISO 3166-2 location code
* [ ] 4xx-class errors are returned for other types of invalid requests for `property_id` param, and for `IS_OVERLAY_ENABLED`, `IS_PREFETCH_ENABLED` and `FIDES_STRING` environment variables.

### Steps to Confirm

* [ ] Try using a `geolocation` param with wrong location text in `fides.js`, so it checks the format.
* [ ] Try using `property_id` param without `geolocation` param.

### Pre-Merge Checklist

* [ ] Update `CHANGELOG.md`

[PROD-2035]: https://ethyca.atlassian.net/browse/PROD-2035?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


clients/privacy-center/pages/api/fides-js.ts
```diff
@@ -109,15 +109,28 @@ export default async function handler(
...
+
+    propertyId = safeLookupPropertyId(
+      req,
+      geolocation,
+      environment,
+      fidesString
+    );
+  } catch (error) {
+    if (error instanceof Error) {
+      console.error(error);
```
galvana, line 127: Add this line before the console.error to disable the lint error
```suggestion
      // eslint-disable-next-line no-console
      console.error(error);
```

clients/privacy-center/pages/api/fides-js.ts
```diff
@@ -109,15 +109,28 @@ export default async function handler(
...
+      res
+        .status(400) // 400 Bad Request. Malformed request.
+        .send(error.message);
```
galvana, line 130: You can do this instead of checking if `error` is an instance of `Error` and then you could fallback to still return an error message instead of just the empty `return`
```
res
  .status(400) // 400 Bad Request. Malformed request.
  .send(
    error instanceof Error ? error.message : "An unknown error occurred."
  );
```

clients/privacy-center/__tests__/common/geolocation.test.ts
```diff
@@ -178,47 +178,101 @@ describe("getGeolocation", () => {
...
+        await lookupGeolocation(req);
+      } catch (error) {
+        if (error instanceof Error) {
+          expect(error.message).toMatch(
+            "Provided location (US-NewYork) query parameter is not in ISO 3166 format."
+          );
+        } else {
+          expect(typeof error).toBe(Error);
+        }
+      }
```
galvana, line 275: You can simplify these checks using Jest's `toThrow` syntax
```
await expect(lookupGeolocation(req)).rejects.toThrow(
  "Provided location (US-NewYork) query parameter is not in ISO 3166 format."
);
```

galvana (commented):
This worked as expected! I just made some small recommendations to clean up the tests a bit and a way to handle errors in the `lookupGeolcation` call

clients/privacy-center/__tests__/common/geolocation.test.ts
andres-torres-marroquin: I think for this one it throws an `Error` object and not a string. So I think this may not work, will check it now and comment back.

andres-torres-marroquin: I was wrong, this works perfectly, I like it, thanks!

galvana (approved):

PR #4885, opened 2024-05-13
Fix middleware logging to show only the information that is correct

andres-torres-marroquin:
Closes #PROD-2033
### Description Of Changes

Fix middleware logging to show only the information that is correct. The middleware tools provided by Next.JS are not enough for measuring the time it takes to process a request, and it is also not possible to capture the returned status code to the client.

### Code Changes

* [ ] Removed code that was logging deceptive information.

### Steps to Confirm

* [ ] Check that the logging output has been reduced to `path` and `method`.

### Pre-Merge Checklist

* [ ] Update `CHANGELOG.md`


clients/privacy-center/middleware.ts
```diff
@@ -1,21 +1,12 @@
...
   const logDict = {
     method: request.method,
-    status_code: response.status,
-    handler_time: `${handlerTime}ms`,
     path: request.nextUrl.pathname,
   };
 
   /* eslint-disable no-console */
   console.info(JSON.stringify(logDict));
   /* eslint-enable no-console */
```
galvana, line 11: We can get rid of this line since it doesn't refer to anything afterwards.

galvana, line 11: We can get rid of this line since it doesn't refer to anything afterwards.

andres-torres-marroquin, line 11:
```suggestion
```

galvana (approved):

PR #4886, opened 2024-05-13
Adding property ID to privacy preference history

galvana:
Closes [PROD-2037](https://ethyca.atlassian.net/browse/PROD-2037)

### Code Changes

* [ ] Adding optional `property_id` to `PrivacyPreferenceHistory` model

### Steps to Confirm

* [ ] See https://github.com/ethyca/fidesplus/pull/1425

### Pre-Merge Checklist

* [ ] Update `CHANGELOG.md`

[PROD-2037]: https://ethyca.atlassian.net/browse/PROD-2037?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/fides/api/models/privacy_preference.py
```diff
@@ -388,6 +388,8 @@ def __tablename__(self) -> str:
         MutableDict.as_mutable(JSONB)
     )  # Dict of TCF attributes saved, for a TCF notice
 
+    property_id = Column(String, index=True, nullable=True)
```
pattisdr, line 391: hm, this is the same as `plus_property.id` right? Just trying to think through if this should be a FK. We are enforcing other resources like the privacy notice, privacy notice history, etc. be proper FK's.  

Cons are that this is a large table for some customers, adding a FK is tricky.  It could be added with something like a "NOT VALID" at first and then issuing a separate query to validate the constraint separately.

It might not be worth that though, a string could be fine too, just wanted to surface

galvana: Yes! Same thing, that was my intention but I forgot to add the foreign key constraint 😵‍💫. I set it to be nullable since it's not guaranteed that the associated privacy experience will be retrieved by property ID, which is the main bit of info we're tracking with this new addition. I also set `ondelete="SET NULL"` since we want to be able to delete properties freely and still keep these historical records around.

I didn't know about "NOT VALID" in Postgres so thanks for sharing! How would that apply here? From my understanding it allows us to delay the constraint validation and create the constraint more quickly. Would this optimization still be needed if every row has a null `property_id`?

pattisdr: discussed in person, we'll revert back to a string to preserve the history in case the property is deleted in the future.

src/fides/api/alembic/migrations/versions/57479150efd6_adding_property_id_to_historical_records.py
```diff
@@ -0,0 +1,49 @@
...
+def upgrade():
+    op.add_column(
+        "privacypreferencehistory", sa.Column("property_id", sa.String(), nullable=True)
+    )
+    op.create_index(
+        op.f("ix_privacypreferencehistory_property_id"),
+        "privacypreferencehistory",
+        ["property_id"],
+        unique=False,
+    )
```
pattisdr, line 28: Comment on ticket: https://ethyca.atlassian.net/browse/PROD-2037?focusedCommentId=40988

src/fides/api/models/privacy_preference.py
```diff
@@ -327,6 +327,12 @@ def __tablename__(self) -> str:
         MutableDict.as_mutable(JSONB)
     )  # Dict of all the TCF attributes served if applicable
 
+    property_id = Column(
+        String,
+        index=True,
+        nullable=True,
+    )
```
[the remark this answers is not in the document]
pattisdr: Also, we could define this field on the `ConsentReportingMixinV2` in one place I believe, it's shared between PrivacyPreferenceHistory and ServedNoticeHistory

pattisdr (approved):
Approved pending a couple of requested changes to add a model-level test, and to consolidate the field definitions on the ConsentReportingMixinV2

PR #4893, opened 2024-05-16
Add option to `downgrade` DB as part of `/admin/db` endpoint 

adamsachs:
Closes [PROD-2087](https://ethyca.atlassian.net/browse/PROD-2087)

### Description Of Changes

Adds a downgrade option to our `/admin/db` endpoint. Allows specifying a migration revision as a `revision` query param. `revision` query param can also be used for upgrade command now, if desired - though in practice, that probably won't be needed much.

Also adds the current revision number/name as a field in the `/health/database` endpoint response, e.g.
```
{
  "database": "healthy",
  "database_revision": "fc2b2c06e595"
}
```

See issue description for more context.

### Code Changes

* [x] add `downgrade` action to path param enum. executes an alembic downgrade against the db
* [x] add `revision` optional query param to allow for migration revision name/number to be specified on the endpoint
* [x] add the current revision number/name as a field in the `/health/database` endpoint. 
* [x] fix a downgrade migration that was invalid and would break/stop the downgrade if executed 

### Steps to Confirm

* [x] run fides server: `nox -s dev`
* [x] hit the `/api/v1/admin/db/downgrade` endpoint (e.g. thru swagger UI) and specify a `revision` that's lower than the `head` on main, e.g. `9e83545ed9b6`
* [x] confirm DB is altered as expected and its `alembic_version.version_num` value is the `revision` that was specified in the step above  

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

[PROD-2087]: https://ethyca.atlassian.net/browse/PROD-2087?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


CHANGELOG.md
```diff
@@ -42,6 +41,8 @@ The types of changes are:
 - Added custom JSON (de)serialization to shared DB engines to handle non-standard data types in JSONB columns [#4818](https://github.com/ethyca/fides/pull/4818)
 - Added state persistence across sessions to the datamap report table [#4853](https://github.com/ethyca/fides/pull/4853)
 - Removed currentprivacypreference and lastservednotice tables [#4846](https://github.com/ethyca/fides/pull/4846)
+- Added initial version for Helios: Data Discovery and Detection [#4839](https://github.com/ethyca/fides/pull/4839)
+- Adds new var to track fides js overlay types [#4869](https://github.com/ethyca/fides/pull/4869)
```
adamsachs, line 45: some unrelated changelog maintenance while i was here: these were released in 2.36.0 and showed up under 2.36.0 on that release branch/tag, so just adjusting in `main` to keep that correct going forward

src/fides/api/alembic/migrations/versions/fc2b2c06e595_add_detection_discovery_tables.py
```diff
@@ -131,7 +131,7 @@ def downgrade():
 
     # Staged resource table
     op.drop_index(op.f("ix_stagedresource_resource_type"), table_name="stagedresource")
-    op.drop_index(op.f("ix_stagedresource_key"), table_name="stagedresource")
+    op.drop_index(op.f("ix_stagedresource_urn"), table_name="stagedresource")
```
adamsachs, line 134: unrelated to main point of the PR, but i noticed this when trying to downgrade, i'd been sloppy in keeping this downgrade consistent with the actual table definition, so it broke the downgrade command. fixing this up.

src/fides/api/api/v1/endpoints/admin.py
```diff
@@ -27,23 +28,28 @@ class DBActions(str, Enum):
...
+                raise HTTPException(
+                    status_code=status.HTTP_501_NOT_IMPLEMENTED,
+                    detail="Resetting the application database outside of dev_mode is not supported.",
+                )
 
-    await configure_db(CONFIG.database.sync_database_uri)
+            reset_db(CONFIG.database.sync_database_uri)
+            action_text = "reset"
+
+        await configure_db(CONFIG.database.sync_database_uri, revision=revision)
```
pattisdr, line 52: If existing upgrade fails here, it will fail in the logs but the API will return that it performed successfully

src/fides/api/api/v1/endpoints/admin.py
```diff
@@ -27,23 +28,28 @@ class DBActions(str, Enum):
...
-    action_text = "upgrade"
 
-    if action == DBActions.reset:
-        if not CONFIG.dev_mode:
-            raise HTTPException(
-                status_code=status.HTTP_501_NOT_IMPLEMENTED,
-                detail="Resetting the application database outside of dev_mode is not supported.",
-            )
+    if action == DBActions.downgrade:
+        downgrade_db(CONFIG.database.sync_database_uri, revision=revision)  # type: ignore[arg-type]
```
pattisdr, line 37: If downgrade fails, say the downrev is wrong, or it is a downgrade we cannot complete (there are many downgrades that don't actually work) - there will be an uncaught 500 here

src/fides/api/api/v1/endpoints/admin.py
```diff
@@ -27,23 +28,28 @@ class DBActions(str, Enum):
...
+        if action == DBActions.reset:
+            if not CONFIG.dev_mode:
+                raise HTTPException(
+                    status_code=status.HTTP_501_NOT_IMPLEMENTED,
+                    detail="Resetting the application database outside of dev_mode is not supported.",
+                )
 
-    await configure_db(CONFIG.database.sync_database_uri)
+            reset_db(CONFIG.database.sync_database_uri)
+            action_text = "reset"
```
pattisdr, line 50: - If action is "reset" and revision is also in request body (say it's a lower downrev)
- reset_db is called, and then configure_db is called, this will result in a downgrade.

Looks like this was already the existing behavior.  If we are separating out "downgrading" should we not allow a revision to be supplied on reset?

src/fides/api/api/v1/endpoints/health.py
```diff
@@ -32,6 +32,7 @@ class DatabaseHealthCheck(BaseModel):
     """Database Healthcheck Schema"""
 
     database: str
+    database_revision: str
```
pattisdr, line 35: Looks like this is potentially Optional

src/fides/api/db/database.py
```diff
@@ -41,15 +41,24 @@ def upgrade_db(alembic_config: Config, revision: str = "head") -> None:
     command.upgrade(alembic_config, revision)
 
 
-async def migrate_db(database_url: str, samples: bool = False) -> None:
+def downgrade_db(database_url: str, revision: str = "head") -> None:
```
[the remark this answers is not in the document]
pattisdr: are there any steps in the corresponding upgrade or reset step that also need to be here? I don't think so but want to double check.

pattisdr (commented):
i really like exposing the current revision here.

Further comments: https://ethyca.atlassian.net/browse/PROD-2087?focusedCommentId=41060

src/fides/api/db/database.py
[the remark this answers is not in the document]
adamsachs: > I understand why this was organized like this (namely you probably don't want to be getting the get_alembic_config in the endpoint), but it feels like "downgrade_db" and "upgrade_db" should have the same structure

yeah i felt that way too, let me try to restructure it a bit, will also make it a bit DRYer.

> are there any steps in the corresponding upgrade or reset step that also need to be here? I don't think so but want to double check.

i don't think so either, not from my testing! the alembic command to downgrade seemed sufficient. the reset step does do some more alembic cleanup, but i don't think that's appropriate for a downgrade. i could certainly be missing something though if something catches your eye!

pattisdr:
Reviewing -

src/fides/api/db/database.py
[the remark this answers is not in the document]
pattisdr: nice, the restructure looks good

src/fides/api/api/v1/endpoints/admin.py
pattisdr: nice, new try/catch looking good:

Error: Internal Server Error
--



```json
{
  "detail": "Database downgrade failed: PostgreSQL ENUM type requires a name.. Check server logs for more details"
}
```

src/fides/api/api/v1/endpoints/admin.py
pattisdr: nice, new try/catch looking good:

Error: Internal Server Error
--



```json
{
  "detail": "Database downgrade failed: PostgreSQL ENUM type requires a name.. Check server logs for more details"
}
```

.github/pull_request_template.md
```diff
@@ -23,3 +23,7 @@ _Write some things here about the changes and any potential caveats_
 * [ ] Relevant Follow-Up Issues Created
 * [ ] Update `CHANGELOG.md`
 * [ ] For API changes, the [Postman collection](https://github.com/ethyca/fides/blob/main/docs/fides/docs/development/postman/Fides.postman_collection.json) has been updated
+* If there are any database migrations:
+  * [ ] Ensure that your downrev is up to date with the latest revision on `main`
+  * [ ] Ensure that your `downgrade()` migration is correct and works
+    * [ ] If a downgrade migration is not possible for this change, please call this out in the PR description!
```
pattisdr, line 29: nice changelog additions 👍

src/fides/api/api/v1/endpoints/admin.py
pattisdr: Not addressed, but also not part of this ticket, so we can defer

pattisdr (approved):
Looks good to me, I especially like the re-org. 

There's an outstanding existing issue if we upgrade and it fails, the API returns that it succeeded, but the logs show the failure.  I think we should reticket -

adamsachs:
> Looks good to me, I especially like the re-org.
> 
> There's an outstanding existing issue if we upgrade and it fails, the API returns that it succeeded, but the logs show the failure. I think we should reticket -

ah yes sorry i ignored that because it was already there and i was the slightest bit unsure about how to best handle/got a bit worried about breaking things. i'll create a follow up ticket - thanks for the speedy and helpful review on this!

src/fides/api/api/v1/endpoints/admin.py
adamsachs: thank you for the diligence! https://ethyca.atlassian.net/browse/PROD-2118

adamsachs:
ticketed https://ethyca.atlassian.net/browse/PROD-2118 as a follow up for the upgrade issue you uncovered 👍

PR #4894, opened 2024-05-16
Conditional index creation

galvana:
Closes [PROD-2037](https://ethyca.atlassian.net/browse/PROD-2037)

### Description Of Changes

This is subset of the work done for https://github.com/ethyca/fides/pull/4886. @pattisdr raised concerns over the time spent adding the `propery_id` indices to the `privacypreferencehistory` and `servednoticehistory` tables so this is a proposal to minimize prod interruptions during deployment.

The approach is to only run the standard index creation process for databases that have less than 1 million rows for the `privacypreferencehistory` and `servednoticehistory` tables. The Alembic migration will conditionally skip the index creation if the tables have more than 1 million rows. We can use the `verify_index_creation.py` script to report on which instances don't have the indices and create them manually.

### Code Changes

* [ ] Splitting out the addition of the `property_id` column and index creation
* [ ] Adding a standalone script to verify the existence and status of the two indices

### Steps to Confirm

* [ ] Start the server using `nox -s dev`
* [ ] Shell into the `fides` container using `nox -s shell` and run the following command
```
python scripts/verify_concurrent_index_creation.py
```
you should see
```
{"ix_privacypreferencehistory_property_id": "complete", "ix_servednoticehistory_property_id": "complete"}
```

### Pre-Merge Checklist

[PROD-2037]: https://ethyca.atlassian.net/browse/PROD-2037?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/fides/api/alembic/migrations/versions/57479150efd6_add_property_id_to_historical_records.py
```diff
@@ -0,0 +1,60 @@
...
+down_revision = "fc2b2c06e595"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column(
+        "privacypreferencehistory", sa.Column("property_id", sa.String(), nullable=True)
+    )
+    with op.get_context().autocommit_block():
```
galvana, line 23: Concurrent index creation cannot happen inside a transaction so this needs to auto-commit immediately.

scripts/verify_concurrent_index_creation.py
```diff
@@ -0,0 +1,65 @@
...
+if __name__ == "__main__":
+    SessionLocal = get_db_session(CONFIG)
+    db: Session = SessionLocal()
+
+    table_index_map: Dict[str, List[str]] = {
+        "privacypreferencehistory": ["ix_privacypreferencehistory_property_id"],
+        "servednoticehistory": ["ix_servednoticehistory_property_id"],
+    }
+
+    index_info: Dict[str, str] = get_index_info(db, table_index_map)
```
galvana, line 64: The status can be `not found`, `complete`, or any of the CREATE INDEX statuses https://www.postgresql.org/docs/12/progress-reporting.html#CREATE-INDEX-PROGRESS-REPORTING

src/fides/api/alembic/migrations/versions/57479150efd6_add_property_id_to_historical_records.py
pattisdr: I missed discussion on this one - My gut opinion is that a migration should not auto-commit.   

Sometimes we are upgrading customers across multiple fides versions and multiple migrations, and if something goes wrong with the upgrade, it is fairly straightforward to bail early and roll them back, because the migrations themselves did not commit (doesn't happen until they're all done).

In my opinion, I'd have the standard image add the index per usual, not concurrently, but then create special images for large customers that omits the index creation.  Then we separately in a postgres shell create the index by hand.

This is just one thought, but I worry the auto-commit here is dangerous.

src/fides/api/alembic/migrations/versions/4fba24045fec_add_property_id_indices.py
```diff
@@ -0,0 +1,53 @@
...
+    privacypreferencehistory_count = connection.execute(
+        sa.text("SELECT COUNT(*) FROM privacypreferencehistory")
+    ).scalar()
+    if privacypreferencehistory_count < 1000000:
+        op.create_index(
+            op.f("ix_privacypreferencehistory_property_id"),
+            "privacypreferencehistory",
+            ["property_id"],
+            unique=False,
+        )
```
pattisdr, line 32: I'd add logs in both places if this doesn't run

galvana: Done!

pattisdr (approved):
This seems reasonable.   We're only creating an index as part of the migration for tenants with low row counts, otherwise it has to be run manually.   This alone won't guarantee that the migration goes smoothly, a tenant could be dealing with higher traffic levels as well. Upgrading during lower traffic times would add more confidence.

src/fides/api/alembic/migrations/versions/4fba24045fec_add_property_id_indices.py
```diff
@@ -0,0 +1,66 @@
...
+branch_labels = None
+depends_on = None
+
+def upgrade():
+    connection = op.get_bind()
+
+    # only run the index creation if the tables have less than 1 million rows
+    privacypreferencehistory_count = connection.execute(
+        sa.text("SELECT COUNT(*) FROM privacypreferencehistory")
+    ).scalar()
```
pattisdr, line 25: Note this count can take a little while to fetch on larger tenants -

galvana: I was looking into more efficient row count techniques but I wasn't sure if that was overkill, is there an approach that you recommend from your experience?

pattisdr: there are ways to estimate a count https://stackoverflow.com/questions/7943233/fast-way-to-discover-the-row-count-of-a-table-in-postgresql, I also am not sure if this is overkill.

PR #4895, opened 2024-05-17
Add shouldShowExperience to the Fides global and FidesInitialized events

guncha:
### Description Of Changes

This PR tweaks the Fides API and moves some stuff around to expose the logic used to determine if the banner should be shown to the user. Previously this would only be available deep in the render tree, but now it's accessible on the `FidesInitialized` event and on the global too.

The goal here was to keep the changes as simple as possible. I certainly could have refactored more, but went back and forth on it and ultimately erred on less is more.

Oh, and for 2061, Fides will still throw an error when it's configured to embed the content, and there's no embed container on the page. My recommendation is for the customer to always have a hidden container that gets shown or moved around based on the new `shouldShowExperience` field.

Closes #PROD-1980
Closes #PROD-2061

### Code Changes

1. Add a `cookie` property to Fides
2. Add a `shouldShowExperience` method that depends on the `experience`, `cookie` and `saved_consent`, all of which are saved on the Fides global.
3. Refactor some initialization code to access and update these properties in the initialization process
4. Update the FidesInitialized events to add the `shouldShowExperience` value

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`


clients/fides-js/src/fides.ts
```diff
@@ -104,42 +108,44 @@ async function init(this: FidesGlobal, config: FidesConfig) {
...
+    cookie: this.cookie,
+    savedConsent: this.saved_consent,
     updateExperienceFromCookieConsent: updateExperienceFromCookieConsentNotices,
   });
   if (initialFides) {
     Object.assign(this, initialFides);
     updateWindowFides(this);
-    dispatchFidesEvent("FidesInitialized", cookie, config.options.debug);
+    dispatchFidesEvent("FidesInitialized", this.cookie, config.options.debug, {
+      shouldShowExperience: this.shouldShowExperience(),
```
eastandwestwind, line 132: something to be aware of is that this `shouldShowExperience` could be different than the `shouldShowExperience` on L 147 of this file. So, if a customer is listening / responding to the 1st call, it could be incorrect.  For example, the banner will surface even though the 1st `shouldShowExperience` is false, in the case where the experience returns a new notice with a key that does not yet exist in the cookie.

clients/fides-js/src/lib/initialize.ts
```diff
@@ -363,10 +363,10 @@ export const initialize = async ({
         `User location could not be obtained. Skipping overlay initialization.`
       );
       shouldInitOverlay = false;
-    } else if (!isPrivacyExperience(effectiveExperience)) {
+    } else if (!isPrivacyExperience(fides.experience)) {
       fetchedClientSideExperience = true;
       // If no effective PrivacyExperience was pre-fetched, fetch one using the current region string
-      effectiveExperience = await fetchExperience(
+      fides.experience = await fetchExperience(
```
eastandwestwind, line 369: I like that we're now using the global fides to assign updated experience 👍

clients/fides-js/src/lib/initialize.ts
```diff
@@ -413,23 +413,23 @@ export const initialize = async ({
...
+        cookie: fides.cookie,
+        experience: fides.experience,
       });
       debugLog(
         options.debug,
         "Updated current cookie state from experience",
         updatedCookie
       );
-      Object.assign(cookie, updatedCookie);
+      fides.cookie = updatedCookie;
```
eastandwestwind, line 424: 💯

clients/fides-js/src/lib/initialize.ts
```diff
@@ -463,29 +463,31 @@ export const initialize = async ({
          * don't block the rest of the code from executing, we use setTimeout with
          * no delay which simply moves it to the end of the JavaScript event queue.
          */
-        setTimeout(() => {
-          automaticallyApplyGPCPreferences({
-            savedConsent,
-            effectiveExperience: effectiveExperience as PrivacyExperience,
-            cookie,
+        setTimeout(
+          automaticallyApplyGPCPreferences.bind(null, {
```
eastandwestwind, line 467: This method has the potential to alter whether or not a banner/modal will show in the UI. 

E.g. a new user visits the site with GPC enabled in their browser. The only notice that Fides returns in the experience is GPC enabled. First, Fides calls `FidesInitialized` twice, where `shouldShowExperience` is true. Then Fides hits this line and automatically saves this preference in the API and cookie, and does not surface the banner/modal. 

So, in this case, Fides will send the `FidesUpdated` event automatically, without the user explicitly selecting a consent option. We _could_ pass in `shouldShowExperience` in this event, too, but it really depends on how this var will be queried by the customer

eastandwestwind (commented):
Nice work @guncha ! Overall I know we're moving in the right direction with consolidating how we pass around state for consent throughout Fides, and you're right that it's a bit risky to do a larger refactor, but this is on our radar. 

Could caveats and things to point out here, curious to hear your thoughts.

eastandwestwind:
Something I've been thinking about in the background, but not required here is the need for Fides to have another event, something like `FidesReady`, which signifies a) we have everything need to calculate the "final" decision of whether consent is going to be re-surfaced via banner/modal, and b) thus, nothing on the Fides global / cookie will change in the given init lifecycle so the customer can rely on the vals to do whatever they need.

clients/fides-js/src/lib/initialize.ts
eastandwestwind: OK so I just confirmed with @NevilleS that GPC shouldn't affect the customer they are specifically using TCF, but this is something we'll want to update in the future. So the approach you have here is fine for now 👍

clients/fides-js/src/fides.ts
eastandwestwind: update here- for TCF, this should remain the same across `FidesInitialized` calls, as we rely on the tcf_version_hash to form this calculation. This shouldn't change across these calls, so I think we're good to go! We just want to be aware in case we have non-TCF customers wanting to use this. We have a follow-up ticket to track a larger init events refactor so that this won't be as misleading

clients/fides-js/src/lib/initialize.ts
```diff
@@ -463,29 +463,31 @@ export const initialize = async ({
...
-          });
-        });
+          })
+        );
       }
     }
   }
 
+  const { consent, fides_meta, identity, fides_string, tcf_consent } =
+    fides.cookie;
```
gilluminate, line 481: love this simplification

clients/fides-js/src/fides.ts
guncha: Yeah, this could cause some issues.. Having `FidesInitialized` event handlers be called twice, potentially with different data, could easily confuse some code that's not expecting it. My preference would be to use a different event name for the first one and perhaps not even pass `shouldShowExperience` to it.

clients/fides-js/src/lib/initialize.ts
guncha: Interesting.. Thanks for explaining this! That's another situation where `shouldShowExperience` is not going to match the actual behavior. Another one that I'm aware of is setting `fides_disable_banner` to true. This is not an exact property. Perhaps in the future we could just run the view logic (i.e., just attempt to render the banner and use that to determine the value of `shouldShowExperience`).

eastandwestwind (approved):
I've tested this out and confirmed that the following worked:

```
window.addEventListener("FidesInitialized", (evt) => {
          console.log(evt.detail.extraDetails.shouldShowExperience);
        });
```

and that `Fides.shouldShowExperience()` also returns the appropriate value

clients/fides-js/src/docs/fides-event.ts
```diff
@@ -104,6 +104,11 @@ export interface FidesEvent extends CustomEvent {
        */
       servingComponent?: "banner" | "modal" | "tcf_banner" | "tcf_overlay";
 
+      /**
+       * Whether the user should be shown the consent experience. Only available on FidesInitialized events.
```
NevilleS, line 108: nit: Any reason why we can't include this on all events? It'd be easier to understand...

clients/fides-js/src/fides-tcf.ts
```diff
@@ -236,6 +243,16 @@ const _Fides: FidesGlobal = {
     }
     return this.init(this.config);
   },
+  shouldShowExperience() {
+    if (!isPrivacyExperience(this.experience) || !this.cookie) {
+      throw new Error("Should have an experience and a cookie");
+    }
```
NevilleS, line 249: Won't this cause `FidesInitialized` to throw errors and never initialize when no experience applies? That's a common use case

NevilleS (changes requested):
Couple nitpicks and some questions

clients/fides-js/src/fides-tcf.ts
NevilleS: e.g. https://cookiehouse-plus-rc.fides-staging.ethyca.com/?geolocation=RU

clients/fides-js/src/docs/fides-event.ts
guncha: Well, I think this is only really useful for the initialized event and to get this on the other events, I'd have to pass FidesGlobal down to all of the code that emits them. So there's a reason - I didn't say it was a good one :)

clients/fides-js/src/fides-tcf.ts
guncha: It worked okay in my testing because `isPrivacyExperience({}) == true` and `{}` is what the server responds with when there's no specific experience and also what the FE code falls back to if there was an error fetching it. Initially though `this.experience` starts out as `undefined` so if this is called too early, it will result in a (programmer) error, hence this check.

clients/fides-js/src/docs/fides-event.ts
NevilleS: Heh, I'm not buying it! The type signature for dispatchFidesEvent should just require the FidesGlobal as an arg then, I think it has only like 5 unique call sites? I suspect all of those are already just plucking from the FidesGlobal to provide args to that function today, so the code is telling us it should be zhushed slightly...

clients/fides-js/src/fides-tcf.ts
NevilleS: Ah OK! Good - I didn't test your branch specifically, I'm just on the hunt for potential breaking changes. Ship it

clients/fides-js/src/docs/fides-event.ts
gilluminate: I'm in agreement that this is only really useful for the initialized event. Correct me if I'm wrong @NevilleS, but all other events happen as a result of the user being shown the experience already so it's too late to be useful.

NevilleS: Here's my final defense:
1. API Consistency: It's _easier_ to understand `FidesEvent` if the properties are the same across all events. As a developer, I don't have to learn about the "special cases", I can just look at `detail.extraDetails.shouldShowExperience` and trust that, whether my code is subscribed to `FidesUpdated` or `FidesInitialized` or `FidesUpdating`, etc.
2. Utility: It's also useful to have `shouldShowExperience: false` be triggered by `FidesUpdated` once someone's consent preferences are saved. If you're building your own show/hide logic for a custom banner, you can use that as a state variable... 👍

I can't *guarantee* that we need it though, so we could ship this as-is and add support to other events later if it becomes important; but we'd also need to have a good explanation for when to "hide" a custom experience after it's been shown...

NevilleS: Anyways I should be clear!

We can merge this as-is (well, once CI checks are passing). But I think developer using this API for their own apps will find it a bit confusing. Happy to be wrong there though 👍

NevilleS (approved):
Although I still have some nitpicks in there I won't hold this up 👍 

The CI check failures seem like they might be spurious false alarms so I'll re-run them now to see...

(

NevilleS:
> Although I still have some nitpicks in there I won't hold this up 👍
> 
> The CI check failures seem like they might be spurious false alarms so I'll re-run them now to see...

Still failing on admin-ui tests, which should be unrelated to these changes. Just merged in `main` to see.

clients/fides-js/src/docs/fides-event.ts
guncha: I gave this a solid attempt and it started to turn into a much larger refactor. Basically, everywhere we pass down options, cookie, experience and saved_consent, which is throughout the UI components, the FidesGlobal can be used instead. Some of those code paths emit events while others don't, but it would be strange to keep using different patterns for the same state variables.

NevilleS (changes requested):
OK, I confirmed that those CI failures are a real issue - the Admin UI "Preview" is failing, at least in the Cypress tests 👍 

We'll need to get those passing before merging this

clients/fides-js/src/fides.ts
```diff
@@ -112,42 +115,44 @@ async function init(this: FidesGlobal, providedConfig?: FidesConfig) {
...
+    });
   }
-  const experience = initialFides?.experience ?? config.experience;
+
   const updatedFides = await initialize({
     ...config,
-    cookie,
-    savedConsent,
-    experience,
+    fides: this,
```
NevilleS, line 145: OK, I think this is where this is failing in more E2E testing. I added some logging just before this call to `initialize(...)` like this:

```
  console.log(
    "DEBUG: calling initialize() with config, this, overrides...",
    config,
    this,
    overrides
  );

  const updatedFides = await initialize({
    ...config,
    fides: this,
    renderOverlay,
    updateExperience,
    overrides,
  });
```

In my test that outputs this (sorry, copy-pasting from console is hard here and not super helpful):
image

The issue is:
* at this point in the code, `this.config.experience = { ...configured experience object... }` but `this.experience` is still undefined
* when we do `{ ...config, fides }` and pass that down, you previously were received `experience` as an argument to `initialize()`, but now it's trying to look at `fides.experience`
* but nothing initializes `this.experience` to the value from the config object, so it's ignored

NevilleS (changes requested):
Alright, I think I've traced the root cause here and we can do some more testing to finish this up 👍

clients/fides-js/src/fides.ts
eastandwestwind: Right, previously we had this line `const experience = initialFides?.experience ?? config.experience;` that handled when experience was passed on the config, so I think we'll just need to account for this still.

eastandwestwind: Having issues running cypress admin-ui locally due to the following err, so I'm having to run this through github to test:
```
admin-ui:cy:open: We detected that the Chromium Renderer process just crashed.
admin-ui:cy:open:
admin-ui:cy:open: This can happen for a number of different reasons.
admin-ui:cy:open:
admin-ui:cy:open: If you're running lots of tests on a memory intense application.
admin-ui:cy:open:   - Try increasing the CPU/memory on the machine you're running on.
admin-ui:cy:open:   - Try enabling experimentalMemoryManagement in your config file.
admin-ui:cy:open:   - Try lowering numTestsKeptInMemory in your config file during 'cypress open'.
```

gilluminate: I've been experiencing the same in the Cypress GUI 
CleanShot 2024-05-23 at 09 12 21@2x

clients/fides-js/src/docs/fides.ts
```diff
@@ -246,6 +255,11 @@ export interface Fides {
    */
   config?: any;
 
+  /**
+   * @internal
+   */ 
+    cookie?: any;
```
NevilleS, line 261:
```suggestion
  /**
   * @internal
   */ 
  cookie?: any;
```
  
```

File: clients/fides-js/src/fides-tcf.ts [hunk 10]
```diff
@@ -175,30 +187,34 @@ async function init(this: FidesGlobal, providedConfig?: FidesConfig) {
...
     updateExperienceFromCookieConsent: updateExperienceFromCookieConsentTcf,
   });
   // Initialize the CMP API early so that listeners are established
   initializeTcfCmpApi();
   if (initialFides) {
     Object.assign(this, initialFides);
-    dispatchFidesEvent("FidesInitialized", cookie, config.options.debug);
+    updateWindowFides(this);
+    dispatchFidesEvent("FidesInitialized", this.cookie, config.options.debug, {
+      shouldShowExperience: this.shouldShowExperience(),
```
NevilleS, line 200: We established below that the `this.experience = ...` assignment on line 203 was needed because otherwise `this.experience` was `undefined` in some cases when calling `await initialize({ ... })` on line 204.

So the corollary: we're *sure* that `this.experience` will be populated by line 200 in the special case when `initialFides` is truthy here, right? Looking at the implementation of `getInitialFides()`, I think that's safe, but... just triple-checking 😉

NevilleS (approved):
OK final** nitpick/question but I think we're good to go

PR #4900, opened 2024-05-20
support new query string `initialize=false`

gilluminate:
Closes [PROD-2086](https://ethyca.atlassian.net/browse/PROD-2086)

### Description Of Changes

Add a `/fides.js?initialize=false` option to the Privacy Center that disables the initialization but still stores the server-side “config” in the `window.Fides` object that FidesJS can access later via `init()`.

Takes advantage of the `window.Fides.config` property that `reinitialize` is using. When that property exists, we use it for the initialization (which in turn simplifies reinitialize since it doesn't need to be passed in as a property every time). 

### Code Changes

* Update `fides.ts` and `fides-tcf.ts` to take advantage of `window.Fides.config` (`this.config`), if it exists, when initializing fides.js
* update `reinitialize()` to not pass it in to `init()` now that it's already self aware of that property
* Update Next's route for fides-js.ts to not automatically trigger the `init()` function if query string `initialize=false` is present.

### Steps to Confirm

* Visit demo page of Privacy Center with initialization set to false `/fides-js-demo.html?initialize=false`
* Notice in the devTools console log that initialization has been skipped and nothing else.
* In devTools console, update some config settings at this point (eg `window.Fides.config.options.fidesPrimaryColor = "#008000"`)
* In devTools console, execute the `window.Fides.init()` command.
* Everything will initialize as it normally would have if the query string wasn't present (feel free to compare the 2)
* The config setting you injected should take effect (eg. color should be green when modal gets opened).

### Pre-Merge Checklist

* Documentation:
  * [x] added to developer docs
* [x] Update `CHANGELOG.md`

[PROD-2086]: https://ethyca.atlassian.net/browse/PROD-2086?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


clients/privacy-center/pages/api/fides-js.ts
```diff
@@ -277,12 +285,22 @@ export default async function handler(
...
-    window.Fides.init(fidesConfig);`
+    ${
+      initializeQuery === "false"
+        ? `// Capture the base config for later initialization
+        window.Fides.config = fidesConfig;`
+        : `// Initialize fides.js with custom config
+        window.Fides.init(fidesConfig);`
+    }
+    ${
+      environment.settings.DEBUG && initializeQuery === "false"
```
NevilleS, line 300: This DEBUG setting is misleading you - it's meant to be for FidesJS (it's passed along as a global setting to the initialize config) so we shouldn't also use it for Privacy Center. It's poorly named and documented ☺️

For other debug logging I use NODE_ENV === "development". I noticed some inconsistencies with this in my WIP branch adding some more debug logging...

clients/privacy-center/pages/api/fides-js.ts
```diff
@@ -277,12 +285,22 @@ export default async function handler(
...
-    e2eQuery === "true"
+    e2eQuery === "true" // let Cypress set the config and initialize fides.js
       ? ""
       : `
-        // Initialize fides.js with custom config
     var fidesConfig = ${fidesConfigJSON};
-    window.Fides.init(fidesConfig);`
+    ${
+      initializeQuery === "false"
+        ? `// Capture the base config for later initialization
```
NevilleS, line 294: I think this would be easier to understand if you flipped this around a bit and did the following:
1. Always set `window.Fides.config`
2. Call `window.Fides.init` unless initialize query is false

It's a single difference, but as is the Fides.config variable is only set in one branch, which feels like a foot gun to me

clients/privacy-center/pages/api/fides-js.ts
```diff
@@ -257,6 +263,8 @@ export default async function handler(
   /* eslint-disable @typescript-eslint/no-use-before-define */
   const customFidesCss = await fetchCustomFidesCss(req);
 
+  const { initialize: initializeQuery } = req.query;
```
NevilleS, line 266: nit: Id prefer to colocated the little bit of logic that checks for the value "false" here and comment why this is being done, instead of putting that online below with the script string building. It feels easier to come back to this later.

NevilleS (approved):
Minor nits, and one suggestion to make it easier to reason about the Fides.config internal variable for our future selves...

clients/privacy-center/pages/api/fides-js.ts
gilluminate: oh! much appreciated. I'll swap that out.

clients/fides-js/src/fides.ts
```diff
@@ -78,8 +78,18 @@ const updateExperience: UpdateExperienceFn = ({
 /**
  * Initialize the global Fides object with the given configuration values
  */
-async function init(this: FidesGlobal, config: FidesConfig) {
-  this.config = config;
+async function init(this: FidesGlobal, config?: FidesConfig) {
+  // confused by the "this"? see https://www.typescriptlang.org/docs/handbook/2/functions.html#declaring-this-in-a-function
+
+  if (!config && !this.config) {
```
guncha, line 84: You're checking for the same conditions twice here which suggests that this could be simplified further. I think this is equivalent:

```ts
async function init(this: FidesGlobal, providedConfig?: FidesConfig) {
  const config = (this.config ??= providedConfig) ?? raise("Fides must be initialized with a configuration object");
  // ...
}
```

Where `raise` is a tiny helper to allow throwing errors in an expression context. `throw` is a statement so we can't use it like`foo ?? throw "Trouble!"`

guncha (commented):
Looks great! Do we still need the `reinitialize` function? Seems like it's no longer needed as it's the same as `init` being called without any arguments.

gilluminate:
> Do we still need the `reinitialize` function?

@guncha I don't mind keeping it as a more explicit action that can be taken, especially since it's so tiny. I could also see it doing more things down the road. @NevilleS thoughts?

clients/fides-js/src/fides.ts
gilluminate: Thanks! Updated to implement. It's high time we added a common util file also!

NevilleS:
> > Do we still need the `reinitialize` function?
> 
> @guncha I don't mind keeping it as a more explicit action that can be taken, especially since it's so tiny. I could also see it doing more things down the road. @NevilleS thoughts?

Since it's part of our documented API, we should still maintain it. If it's equivalent to just calling `Fides.init()` though, we could add _that_ API to the docs and deprecate `reinitialize` 😄

gilluminate:
> If it's equivalent to just calling `Fides.init()` though, we could add _that_ API to the docs and deprecate `reinitialize` 😄

@NevilleS @guncha I've updated the docs to reflect that change! https://github.com/ethyca/fides/pull/4900/commits/ded903e71aae0225026708c931393bce2b2e0d22

PR #4901, opened 2024-05-21
PROD-2040 Migration and CRUD support for property-specific messaging templates

eastandwestwind:
Closes https://ethyca.atlassian.net/browse/PROD-2040

### Description Of Changes

This PR ensures that all customers have 1 default property, and that their existing messaging templates are associated with that property.

This PR also provides full CRUD support for all future-anticipated endpoints needed to complete BE for property-specific messaging.

Highest priority for review is ensuring that no regressions occurs due to this PR. The riskiest part is the migration, and ensuring this will be successful across all scenarios.

### Code Changes

* [ ] Add migration to support property-specific messaging
* [ ] Adds CRUD logic in anticipation of endpoints
* [ ] Adds CRUD tests with full coverage

### TODO (dev notes)
* [ ] Manually test migration- all paths
* [ ] Manually test Admin-UI for backwards compatibility
* [x] Add validation: ensure we always have a default property on property created / update / delete
* [x] Add validation: Property cannot be deleted if its being used on an email config, add tests for this
* [x] Update DB/validation: we can now have empty properties on a message template, allow overlapping type/properties as long as it's "disabled". Add check that if "enabled" we don't have overlapping type/properties
* [x] Update CRUD to support partial update and/or enable/disable of a specific message template

### Steps to Confirm

* [ ] Confirm no regressions in Admin-UI properties / messaging templates for both plus and non-plus users

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`


src/fides/api/alembic/migrations/versions/3b0a209ec772_property_specific_messaging.py
```diff
@@ -0,0 +1,276 @@
...
+        postgresql_where=sa.text("is_default"),
+    )
+
+    bind: Connection = op.get_bind()
+
+    # First, ensure we have exactly 1 default property in the DB
+    existing_properties: ResultProxy = bind.execute(
+        text("select id from plus_property;")
+    )
+    # todo- what's the right way to get count of rows that's performant?
```
eastandwestwind, line 163: Looking for pointers here

src/fides/api/models/property.py
```diff
@@ -51,6 +56,24 @@ def __tablename__(self) -> str:
         lazy="selectin",
     )
 
+    # todo- do I need to manually insert messaging_templates here or will DB handle it?
```
eastandwestwind, line 59: need guidance here, I'm unsure how these relationships work under the hood

src/fides/api/models/privacy_experience.py
```diff
@@ -48,7 +48,7 @@ class ComponentType(Enum):
 FidesJSOverlayTypes: List[ComponentType] = [
     ComponentType.banner_and_modal,
     ComponentType.modal,
-    ComponentType.tcf_overlay
+    ComponentType.tcf_overlay,
```
eastandwestwind, line 51: My local `black` formatter is probably configured differently than it's supposed to be

src/fides/api/models/property.py
```diff
@@ -41,6 +43,9 @@ def __tablename__(self) -> str:
         primary_key=True,
         default=generate_id,
     )
+    # Right now, we use server default to write this val.
+    # In the future we may allow ability to configure which property is the default
+    is_default = Column(Boolean, server_default="f", default=False)
```
eastandwestwind, line 48: todo- Add logic to adding a property, where if it's the only property, it's the default property. We need something that will ensure there's always a default property in advance of being able to provide this toggle in the UI

src/fides/api/models/messaging_template.py
```diff
@@ -56,13 +60,71 @@
...
+def _link_properties_to_messaging_template(
+    db: Session,
+    properties: List[Dict[str, Any]],
+    messaging_template: MessagingTemplate,
+) -> Optional[List[Property]]:
+    """
+    Link supplied properties to MessagingTemplate and unlink any properties not supplied.
+    """
+    if len(properties) == 0:
+        return []
```
pattisdr, line 72: Is your intent to unlink properties not supplied, per the docstring?

There's a case If a MessagingTemplate has some number of existing properties and you update them to have no properties, it will exit here, without removing the properties

src/fides/api/models/property.py
```diff
@@ -111,6 +146,11 @@ def update(self, db: Session, *, data: Dict[str, Any]) -> Property:
                 f'The path(s) \'{", ".join([matching_path.path for matching_path in matching_paths])}\' are already associated with another property.'
             )
 
+        # Ensure that there is at least 1 default. Relevent if we somehow reach a state where there is no defualt property
```
pattisdr, line 149:
```suggestion
        # Ensure that there is at least 1 default. Relevent if we somehow reach a state where there is no default property
```

src/fides/api/models/property.py
```diff
@@ -111,6 +146,11 @@ def update(self, db: Session, *, data: Dict[str, Any]) -> Property:
...
+        has_default_property = Property.get_by(db=db, field="is_default", value=True)
+        if not has_default_property:
+            data["is_default"] = True
```
pattisdr, line 152: There's probably an edge case here where someone is trying to update their sole existing property to have a value of is_default = False which would be allowed -

src/fides/api/models/property.py
```diff
@@ -214,3 +254,26 @@ def link_experience_configs_to_property(
 
     prop.save(db)
     return prop.experiences
+
+
+class MessagingTemplateToProperty(Base):
+    @declared_attr
+    def __tablename__(self) -> str:
+        return "messaging_template_to_property"
```
pattisdr, line 262: Do we need another unique constraint on this table for `messaging_template_id` and `property_id`? In other words, restrict so each pair just shows up once

src/fides/api/service/messaging/message_dispatch_service.py
```diff
@@ -341,6 +343,7 @@ def _build_email(  # pylint: disable=too-many-return-statements
             body=_render(messaging_template.content["body"]),  # type: ignore
         )
     if action_type == MessagingActionType.PRIVACY_REQUEST_ERROR_NOTIFICATION:
+        # get email template doesn't use defualt templates in messaging_template.py
```
pattisdr, line 346:
```suggestion
        # get email template doesn't use default templates in messaging_template.py
```

pattisdr: Can you further explain this code comment, I don't understand what you mean here

src/fides/api/schemas/messaging/messaging.py
```diff
@@ -431,3 +432,25 @@ class MessagingTemplateResponse(MessagingTemplateBase):
 class BulkPutMessagingTemplateResponse(BulkResponse):
     succeeded: List[MessagingTemplateResponse]
     failed: List[BulkUpdateFailed]
+
+
+class MessagingTemplateWithPropertiesBase(BaseModel):
+    id: Optional[str]  # Since summary returns db or defaults, this can be null
```
pattisdr, line 438: Wait why can this be null?

src/fides/api/models/property.py
```diff
@@ -45,6 +55,9 @@ def __tablename__(self) -> str:
         primary_key=True,
         default=generate_id,
     )
+    # Right now, we use server default to write this val.
+    # In the future we may allow ability to configure which property is the default
+    is_default = Column(Boolean, server_default="f", default=False)
```
pattisdr, line 60: Refresh my memory please, is it the intent that you can have multiple MessagingTemplates of the same "type" as long as only one is enabled?

src/fides/api/service/messaging/messaging_crud_service.py
```diff
@@ -81,31 +87,261 @@ def get_messaging_config_by_key(db: Session, key: FidesKey) -> MessagingConfigRe
...
+    if update_id and len(possible_overlapping_templates) == 1:
+        if update_id == possible_overlapping_templates[0].id:
+            return
+
+    # Otherwise, we need to check whether properties will overlap
+    for db_template in possible_overlapping_templates:
+        for db_property in db_template.properties:
+            if db_property.id in new_property_ids:
+                raise MessagingConfigValidationException(
+                    f"There is already an enabled messaging template with template type {template_type} and property {db_property.id}"
```
pattisdr, line 173: What if we are updating Template A.  Say there are two possible overlapping templates, Template A and Template B. Seems like Template A might fail here, because it is not excluded in the overlapping templates. Do we need to exclude `update_id` from possible_overlapping_templates?

src/fides/api/service/messaging/messaging_crud_service.py
```diff
@@ -81,31 +87,261 @@ def get_messaging_config_by_key(db: Session, key: FidesKey) -> MessagingConfigRe
...
+
+def get_template_by_id(db: Session, template_id: str) -> MessagingTemplate:
+    messaging_template: Optional[MessagingTemplate] = MessagingTemplate.get(
+        db, object_id=template_id
+    )
+    if not messaging_template:
+        raise MessagingConfigNotFoundException(
+            f"No messaging template found with id {template_id}"
+        )
+    return messaging_template
```
pattisdr, line 270: You might be able to make use of this new method in other locations on this page -

src/fides/api/service/messaging/messaging_crud_service.py
```diff
@@ -81,31 +87,261 @@ def get_messaging_config_by_key(db: Session, key: FidesKey) -> MessagingConfigRe
...
+            )
+        else:
+            db_templates[template.type] = [
+                {
+                    "id": template.id,
+                    "type": template.type,
+                    "is_enabled": template.is_enabled,
+                    "properties": template.properties,
+                }
+            ]
```
pattisdr, line 315: It looks to me like you're adding the same thing to the array, that the main difference in one conditional, you need to create a key if it doesn't exist.   This is a good use case for a `defaultdict`, something like this:

```python
from collections import defaultdict

db_templates: Dict[str, List[Dict[str, Any]]] = defaultdict(list)
for template in MessagingTemplate.all(db):
    db_templates[template.type].append(
        {
            "id": template.id,
            "type": template.type,
            "is_enabled": template.is_enabled,
            "properties": template.properties,
        }
    )
```

pattisdr (changes requested):
pass complete - it seems there are several bugs in the migration and model-level methods and I am unsure about the data model decision to put is_enabled on the MessagingTemplate.

src/fides/api/models/messaging_template.py
```diff
@@ -56,13 +60,71 @@
...
-    key = Column(String, index=True, unique=True, nullable=False)
+    type = Column(String, index=True, nullable=False)
     content = Column(MutableDict.as_mutable(JSONB), nullable=False)
+    properties: RelationshipProperty[List[Property]] = relationship(
+        "Property",
+        secondary="messaging_template_to_property",
+        back_populates="messaging_templates",
+        lazy="selectin",
+    )
+    is_enabled = Column(Boolean, default=False, nullable=False)
```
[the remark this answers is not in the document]
eastandwestwind: Right now, this is what product wants for the UI, where we enable / disable at the template level, not at the property and template level. But I agree, we should probably plan for something more flexible at the db level, just in case. 

I was struggling before with figuring out how to write data to the many-to-many table when we had `is_enabled` there, but I'll see what I can do. 

Maybe some examples you can point me to of other many-to-many tables where we do something similar? I didn't see anything in the SQLAlchemy docs.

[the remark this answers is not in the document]
pattisdr: If that's what product wants, that's fine, enabling/disabling at the template level. I was just double checking.

src/fides/api/service/messaging/message_dispatch_service.py
eastandwestwind: Oh, some action types, like `MessagingActionType.PRIVACY_REQUEST_ERROR_NOTIFICATION` for example, don't use the default templates that are configurable in the Admin-UI. They are instead hard-coded in `fides`, which we retrieve using `get_email_template(action_type)`

src/fides/api/models/property.py
eastandwestwind: Ah this makes sense 👍

src/fides/api/alembic/migrations/versions/5f96f13be8ec_property_specific_messaging_migration_.py
```diff
@@ -0,0 +1,189 @@
...
+        )
+    else:
+        # Create new property, label as default
+        insert_into_property_query: TextClause = text(
+            "INSERT INTO plus_property (id, name, is_default, type)"
+            "VALUES (:id, :name, :is_default, :type)"
+        )
+        characters = string.ascii_uppercase + string.digits
+        # fixme- does this need the plu_ prefix?
+        new_property_id: str = "plu_FDS-" + "".join(random.choices(characters, k=6))
```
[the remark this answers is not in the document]
eastandwestwind: In our `Properties` model, when we generate a unique ID, we prefix with `FDS`, so this is to keep this convention.

```
def generate_id(self) -> str:
        """
        Generate a unique ID in the format 'FDS-XXXXXX' using uppercase alphanumeric characters.
        """
        characters = string.ascii_uppercase + string.digits
        return "FDS-" + "".join(random.choices(characters, k=6))
```

If we use this, we probably won't need the `plu_` too.

src/fides/api/alembic/migrations/versions/5f96f13be8ec_property_specific_messaging_migration_.py
```diff
@@ -0,0 +1,189 @@
...
+        }
+        try:
+            bind.execute(insert_into_property_query, new_property)
+        except IntegrityError as exc:
+            raise Exception(
+                f"Fides attempted to create a new default property but got error: {exc}. "
+            )
+
+    # STEP 2: Ensure all default templates are saved to the DB
+    existing_templates: ResultProxy = bind.execute(
```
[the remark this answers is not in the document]
eastandwestwind: hmm how are you seeing this? I'm seeing it returns a `ResultProxy` in the lib annotations

src/fides/api/alembic/migrations/versions/5f96f13be8ec_property_specific_messaging_migration_.py
```diff
@@ -0,0 +1,189 @@
...
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    # Reverse data migration: remove templates that were automatically created by the forward migration
+    bind = op.get_bind()
+    bind.execute(text("DELETE * FROM messaging_template_to_property;"))
+    bind.execute(text("DELETE FROM messaging_template WHERE is_enabled = FALSE;"))
+    # ### end Alembic commands ###
```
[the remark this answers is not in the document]
eastandwestwind: Yes, I'll ensure we delete the items we created in this downgrade. We can alter "Default Property" name to identify these defaults in a more reliable way, i.e "Default Property Autogenerated"

src/fides/api/alembic/migrations/versions/5f96f13be8ec_property_specific_messaging_migration_.py
```diff
@@ -0,0 +1,189 @@
...
+        update_property_query: TextClause = text(
+            "UPDATE plus_property SET is_default= TRUE WHERE id= :id"
+        )
+        bind.execute(
+            update_property_query,
+            {"id": existing_properties.first()["id"]},
+        )
+    else:
+        # Create new property, label as default
+        insert_into_property_query: TextClause = text(
```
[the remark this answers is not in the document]
eastandwestwind: Correct 👍

src/fides/api/alembic/migrations/versions/2736c942faa2_property_specific_messaging_db_models.py
```diff
@@ -0,0 +1,118 @@
...
+        unique=False,
+    )
+    op.create_index(
+        op.f("ix_messaging_template_to_property_property_id"),
+        "messaging_template_to_property",
+        ["property_id"],
+        unique=False,
+    )
+    op.add_column(
+        "messaging_template", sa.Column("is_enabled", sa.Boolean(), nullable=False)
```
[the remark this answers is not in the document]
eastandwestwind: yes good catch! I'll update the migration accordingly, thanks 👍

src/fides/api/alembic/migrations/versions/5f96f13be8ec_property_specific_messaging_migration_.py
```diff
@@ -0,0 +1,189 @@
...
+    )
+    for template in updated_templates:
+        insert_into_messaging_template_to_property_query: TextClause = text(
+            "INSERT INTO messaging_template_to_property (id, messaging_template_id, property_id)"
+            "VALUES (:id, :messaging_template_id, :property_id)"
+        )
+        new_messaging_template_to_property: Dict[str, Any] = {
+            "id": generate_record_id("mes"),
+            "messaging_template_id": template["id"],
+            "property_id": default_property_id,
```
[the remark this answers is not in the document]
eastandwestwind: Litte more context: A customer must explicitly turn on the property specific messaging as a feature, via a new ENV var (will be done in this ticket- https://ethyca.atlassian.net/browse/PROD-2049). 

If that ENV var is `False`, when we send out emails from Fides, we get the 1st template by type, basically ignoring the property id. 

Before this ENV is enabled, the customer is expected to have to manually configure and explicitly enable their property-specific messaging templates in Admin-UI. 

That said, here in the migration, we can probably get away with not assigning a property at all to the default templates to keep things simpler here. Let me confirm with @mfbrown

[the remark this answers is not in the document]
eastandwestwind: Actually, just remembered, there's a reason for assigning default property to all templates- in the edge case where: 

1. We do not assign default property to templates
2. Customer adds a new template with same type as an existing template, say `subject identity verification`, with property A. So now we have 1 `subject identity verification` template with no properties, and another `subject identity verification` template with property A. Both templates are enabled.
3. Customer turns OFF the ENV variable, so now Fides is in non-property-specific messaging mode.
4. At runtime, Fides needs to send out an email, but there are 2 templates with type `subject identity verification`
5. Which one do we use?

 We need to guarantee that we have a reliable way to send emails for the non-property-specific use case. If we enforce the fact that there is always 1 default to fall back on, this becomes simpler imo. Thoughts?

src/fides/api/schemas/messaging/messaging.py
eastandwestwind: Items from the DB will have an id associated with it, but "defaults" in this context are the static strings we have set within Fides for each messaging type. 

Basically this is covers the case where not all messaging templates have been saved to the DB, for example, if an open source customer converts to paid, and has new access to the property-specific messaging Admin-UI page. 

This allows all templates to show up in the initial template list table.

src/fides/api/models/property.py
eastandwestwind: you're right- right now the user has no way to configure a property as default / not default, but I'll prevent them from configuring this and raise an error.

I'll add a test case for this too.

src/fides/api/service/messaging/messaging_crud_service.py
eastandwestwind: great idea 💯

src/fides/api/service/messaging/messaging_crud_service.py
eastandwestwind: yes we actually check for the `update_id` matching one of the `possible_overlapping_templates` just above this code block. 

I'll add a unit test for this to be sure

eastandwestwind:
Back over to you @pattisdr , I appreciate all the time you're spending on this 🙏

src/fides/api/service/messaging/messaging_crud_service.py
pattisdr: I thought this was failing for me on "update" maybe, not "create" but I'll double check on the next round

eastandwestwind: oh hm, let me know what you see!

src/fides/api/alembic/migrations/versions/5f96f13be8ec_property_specific_messaging_migration_.py
```diff
@@ -0,0 +1,187 @@
...
```
pattisdr, line 167: You don't define `default_property_id` now in the case where there was already exactly one property, which would cause a failure on this line:    ` "property_id": default_property_id,`

src/fides/api/alembic/migrations/versions/2736c942faa2_property_specific_messaging_db_models.py
```diff
@@ -0,0 +1,126 @@
...
+    op.create_index(
+        op.f("ix_messaging_template_to_property_property_id"),
+        "messaging_template_to_property",
+        ["property_id"],
+        unique=False,
+    )
+    op.add_column(
+        "messaging_template", sa.Column("is_enabled", sa.Boolean(), nullable=True)
+    )
+    op.execute("UPDATE messaging_template SET is_enabled = FALSE;")
```
pattisdr, line 75: Wait why do existing messaging templates get set to disabled? Wouldn't they already be in use?

src/fides/api/alembic/migrations/versions/5f96f13be8ec_property_specific_messaging_migration_.py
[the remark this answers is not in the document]
pattisdr: good name, that works 👍

src/fides/api/alembic/migrations/versions/5f96f13be8ec_property_specific_messaging_migration_.py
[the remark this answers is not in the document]
pattisdr: I still feel like the particular use case I brought up reveals inconsistent behavior, where if they had one property, their property became linked to the templates, but if they had more than one, it did not.

Current behavior?
Customer has 0 properties -> Fides generated default property created and linked to the templates
Customer has 1 property  -> Use-created property is labeled as the default, and linked to the templates
Customer has 2 properties -> Fides generated default property is created and linked to the templates.

In the case of two properties, if you always need at least one default property (and we can't decide which one they would want to be a default), I'd maybe create a default property and link it to the templates, but also link their existing properties to the templates as well?

Open for discussion of course

src/fides/api/alembic/migrations/versions/5f96f13be8ec_property_specific_messaging_migration_.py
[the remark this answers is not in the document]
pattisdr: I was doing type() or these docs https://docs.sqlalchemy.org/en/20/core/connections.html#sqlalchemy.engine.Connection.execute

src/fides/api/alembic/migrations/versions/2736c942faa2_property_specific_messaging_db_models.py
pattisdr: Further, your data migration downgrade deletes templates which are disabled, which deletes everything!

src/fides/api/api/v1/endpoints/messaging_endpoints.py
```diff
@@ -529,26 +529,27 @@ def update_messaging_templates(
...
             content["subject"] = (
                 content["subject"] or default_template["content"]["subject"]
             )
             content["body"] = content["body"] or default_template["content"]["body"]
 
-            MessagingTemplate.create_or_update(
-                db, data={"key": key, "content": content}
+            create_or_update_basic_templates(
+                db,
+                data={"type": template_type, "content": content, "is_enabled": False},
```
pattisdr, line 547: If you have existing enabled templates this will update them to be disabled -

src/fides/api/service/messaging/messaging_crud_service.py
```diff
@@ -80,32 +88,283 @@ def get_messaging_config_by_key(db: Session, key: FidesKey) -> MessagingConfigRe
...
+    """
+    template = None
+    templates = (
+        MessagingTemplate.query(db=db)
+        .filter(MessagingTemplate.type == data["type"])
+        .all()
+    )
+
+    if len(templates) > 1:
+        default_property = Property.get_by(db=db, field="is_default", value=True)
```
pattisdr, line 144: Edge case here, if we have multiple templates defined, and none of our properties are a default, then none of the templates will be updated, instead one will be created

src/fides/api/service/messaging/messaging_crud_service.py
```diff
@@ -80,32 +88,283 @@ def get_messaging_config_by_key(db: Session, key: FidesKey) -> MessagingConfigRe
...
+        if default_property:
+            template = MessagingTemplate.filter(
+                db=db,
+                conditions=(
+                    (MessagingTemplate.type == data["type"])
+                    & (Property.id == default_property.id)
```
pattisdr, line 150: This doesn't look like it actually filters Messaging Templates that match a certain property, the second statement doesn't add anything -it's like this `& ('red' == 'red')`

src/fides/api/service/messaging/messaging_crud_service.py
```diff
@@ -80,32 +88,283 @@ def get_messaging_config_by_key(db: Session, key: FidesKey) -> MessagingConfigRe
...
+                ),
+            ).first()
+    elif len(templates) == 1:
+        template = templates[0]
```
pattisdr, line 154: For readability, I also might swap these conditional, put the 1 case first, and the more than 1 second, just my opinion

src/fides/api/api/v1/endpoints/messaging_endpoints.py
```diff
@@ -529,26 +529,27 @@ def update_messaging_templates(
...
 
             content["subject"] = (
                 content["subject"] or default_template["content"]["subject"]
             )
             content["body"] = content["body"] or default_template["content"]["body"]
 
             MessagingTemplate.create_or_update(
-                db, data={"key": key, "content": content}
+                db,
+                data={"type": template_type, "content": content, "is_enabled": False},
```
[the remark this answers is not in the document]
pattisdr: ok a couple of comments added to that logic -

src/fides/api/models/messaging_template.py
pattisdr: nice changes here -

src/fides/api/service/messaging/message_dispatch_service.py
pattisdr: thank you I didn't realize that distinction! do you mind adding some of this detail to the code comment, and fixing the typo while you're there

src/fides/api/schemas/messaging/messaging.py
pattisdr: thank you for the explanation!

src/fides/api/service/messaging/messaging_crud_service.py
```diff
@@ -80,32 +88,283 @@ def get_messaging_config_by_key(db: Session, key: FidesKey) -> MessagingConfigRe
...
+def get_all_basic_messaging_templates(db: Session) -> List[MessagingTemplate]:
+    """
+    Retrieve all templates from the database, filling in default values if any default template type
+    is not found in the database.
+    """
+    #
     templates_from_db = {
-        template.key: template.content for template in MessagingTemplate.all(db)
+        template.type: template.content for template in MessagingTemplate.all(db)
     }
```
pattisdr, line 99: The results of this can be confusing if there are multiple templates saved of the same type - for each template I think it will show the last template of that type -

src/fides/api/service/messaging/messaging_crud_service.py
pattisdr: ah okay yes, I think the bug I described above is still here - you have: `if update_id and len(possible_overlapping_templates) == 1` but what if there were two overlapping_templates?

src/fides/api/service/messaging/messaging_crud_service.py
pattisdr: you might be able to use in update_messaging_template too?

pattisdr (changes requested):
Nice round of changes @eastandwestwind, I'd like to see all the cases tested on the migration and there are some remaining bugs.

src/fides/api/alembic/migrations/versions/2736c942faa2_property_specific_messaging_db_models.py
eastandwestwind: We're applying the concept of enable/disable by template only to property-specific templates. 

This means we ignore this enable/disable flag for our Basic Templates avail to OSS users, or to Paid users who have not yet enabled the property-specific messaging features.

We set all to disabled so that the user needs to explicitly "enable" (and thus configure) the templates they wish to use for property-specific messaging.

src/fides/api/alembic/migrations/versions/5f96f13be8ec_property_specific_messaging_migration_.py
eastandwestwind: Ah you're right, thanks!

src/fides/api/alembic/migrations/versions/2736c942faa2_property_specific_messaging_db_models.py
pattisdr: still, your downgrade deletes all disabled templates!

src/fides/api/api/v1/endpoints/messaging_endpoints.py
eastandwestwind: I wanted to ensure that if the Basic template UI is being used, then we're not risking the customer enabling the ENV var, and inadvertently sending out messaging templates that were intended only for Basic template mode. I don't HAVE to do this, but it will align with our migration to use "is_enabled=False" across the board. This way, a customer has to explicitly pick / choose which templates they wish to use for property-specific messaging.

src/fides/api/service/messaging/messaging_crud_service.py
eastandwestwind: You're right- this is a state we shouldn't hit as we migrate to ensure there is always a default property, but this is something we should still handle. We should probably just fall back to the 1st item in the list of templates, unless you have better ideas.

src/fides/api/service/messaging/messaging_crud_service.py
eastandwestwind: Ah yes, I was struggling with the syntax for this. Something like:

```
conditions=(
                    (MessagingTemplate.type == data["type"])
                    & (default_property.id in MessagingTemplate.properties)
                ),
```

src/fides/api/service/messaging/messaging_crud_service.py
eastandwestwind: Agreed, thanks!

src/fides/api/api/v1/endpoints/messaging_endpoints.py
pattisdr: okay thanks for explaining your thought process, can you add this as a code comment, so it's clear that disabling every template that gets saved here is intentional?  Especially the bit from the migration comment where you said the "disabled" field is ignored in certain cases.

src/fides/api/service/messaging/messaging_crud_service.py
eastandwestwind: Good thinking. Perhaps we should use the first template instead to align with https://github.com/ethyca/fides/pull/4901#discussion_r1627904594. 

More context:

This is a "basic" template method, and the template-specific messaging feat is only avail to plus users.

This edge case would only be hit if a user enables/configures template-specific messaging, then disables it again. They would have artifact templates configured. This is problematic especially if they downgrade to OSS and cannot change the enabling/disabling of each template.

Because of this, I think we should be consistent with how we render templates in the UI and how we allow users to [create/update templates here](https://github.com/ethyca/fides/pull/4901#discussion_r1627904594).

I don't think we should use enabled/disabled since enabling/disabling by template is only avail for Plus with the ENV var enabled.

pattisdr: > I don't think we should use enabled/disabled since enabling/disabling by template is only avail for Plus with the ENV var enabled.

Yep thanks for clarifying, I didn't know about this requirement!  Again any code comments around the basic templates to help future folks parse non-standard behavior would be useful.

src/fides/api/service/messaging/message_dispatch_service.py
eastandwestwind: absolutely!

src/fides/api/alembic/migrations/versions/2736c942faa2_property_specific_messaging_db_models.py
eastandwestwind: oh I see!

eastandwestwind: adding a unique prefix to messaging templates I've created and deleting based on this

src/fides/api/service/messaging/messaging_crud_service.py
eastandwestwind: I misread that, thanks!

src/fides/api/service/messaging/messaging_crud_service.py
eastandwestwind: I've actually extracted a core "decider" method to handle this in a more consistent way across create/ get/update use cases for our Basic templates 👍

src/fides/api/service/messaging/messaging_crud_service.py
eastandwestwind: not sure if this will work as I get a mypy error `"RelationshipProperty[List[Property]]" has no attribute "__iter__" (not iterable)`

pattisdr: I can switch over and look at the syntax in a little bit when I'm done with Adam's CR

eastandwestwind: Closer but still not working:

```
template = MessagingTemplate.filter(
            db=db,
            conditions=(
                (
                    MessagingTemplate.type
                    == MessagingActionType.SUBJECT_IDENTITY_VERIFICATION.value
                )
                & (MessagingTemplate.properties.any(name=default_property.id))
            ),
        ).first()
```

eastandwestwind: Tried something like [this example](https://stackoverflow.com/a/6512603), which is very similar to this many-to-many relationship I've set up in the model.

```
template = MessagingTemplate.filter(
                db=db,
                conditions=(
                    (MessagingTemplate.type == template_type) &
                    (MessagingTemplate.properties.any(Property.id == default_property.id))
                ),
            ).first()
```

eastandwestwind: This actually worked! But I discovered a very small edge case where we were inadvertently removing properties when we updated basic template if properties previously existed on it.

pattisdr: nice Catherine!

src/fides/api/alembic/migrations/versions/5f96f13be8ec_property_specific_messaging_migration_.py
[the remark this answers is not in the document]
eastandwestwind: For now let's not link all existing properties to templates since the user can always configure whichever properties they need through the Admin-UI. We are additionally discussing some functionality where the templates linked to the default property might act as a catch-all at run-time for email requests that come in from un-configured properties. See #7 here- https://ethyca.atlassian.net/wiki/spaces/EN/pages/3055681577/2024-05-09+Property+Specific+Messaging#Open-questions%3A

We can follow-up as needed.

eastandwestwind:
Making a note here that I've manually tested all migration use cases:

# Test case 1 property

insert into plus_property (id, name, type) values ('plu_FDS-X88ZWUV', 'Test Property 1', 'website');

select * from plus_property;
select * from messaging_template;
select * from messaging_template_to_property;

alembic downgrade 52a5f1a957bc

confirm all tables are back to before migration


# Test case no property

select * from plus_property;
select * from messaging_template;
select * from messaging_template_to_property;

alembic downgrade 52a5f1a957bc

confirm all tables are back to before migration


# Test case 2 properties and 1 existing messaging template type

insert into plus_property (id, name, type) values ('plu_FDS-X88ZWUV', 'Test Property 1', 'website');
insert into plus_property (id, name, type) values ('plu_FDS-X88ZWUW', 'Test Property 2', 'website');
INSERT INTO messaging_template (id, key, content) values ('mes-lksjdfh', 'subject_identity_verification'
, '{}');

select * from plus_property;
select * from messaging_template;
select * from messaging_template_to_property;

alembic downgrade 52a5f1a957bc

confirm all tables are back to before migration

eastandwestwind:
OK @pattisdr I've manually tested the migration, see notes above. I think this is looking good from my end now, let me know your thoughts!

src/fides/api/service/messaging/messaging_crud_service.py
```diff
@@ -80,32 +88,321 @@ def get_messaging_config_by_key(db: Session, key: FidesKey) -> MessagingConfigRe
...
+                db=db,
+                conditions=(
+                    (MessagingTemplate.type == template_type)
+                    & (
+                        MessagingTemplate.properties.any(
+                            Property.id == default_property.id
+                        )
+                    )
+                ),
+            ).first()
```
pattisdr, line 154: A couple things, this should also be ordered - what is "first" can change.  

Second, good job figuring out this query.  This is a little less efficient than it could be, SQLAlchemy is joining in your Manytomany table behind the scenes  "MessagingTemplateToProperty" and then doing a "Exists Select 1)...

I think it's a little clearer to join in the many to many table directly, something like:

```python
db.query(MessagingTemplate).join(MessagingTemplateToProperty).filter(MessagingTemplate.type==template_type, MessagingTemplateToProperty.property_id==default_property_id).order_by(MessagingTemplate.created_at.first()),
```

or however you choose to order.

In both of these queries, there's some interesting SQLAlchemy magic happening around where you defined the relationships to join the right columns together!

pattisdr (commented):
OK @eastandwestwind almost there, just a couple of things around non-repeatable ordering, and efficiency/readability around one of the queries.

Thanks for the diligence in responding to the comments here. I like how you abstracted out some potentially confusing logic into `_basic_messaging_template_by_type` with good docstrings to help future users in the latest iteration, same with added code comments.

Your internal migration testing docs will be very useful for others. Testing all migration paths before sending to code review will reduce back and forth as well 👍

src/fides/api/service/messaging/messaging_crud_service.py
```diff
@@ -80,32 +88,321 @@ def get_messaging_config_by_key(db: Session, key: FidesKey) -> MessagingConfigRe
...
+
+    Scenario C: Multiple templates configured with type, one with default property
+    Result: Return the template with default property
+    """
+    template = None
+    templates = (
+        MessagingTemplate.query(db=db)
+        .filter(MessagingTemplate.type == template_type)
+        .all()
+    )
```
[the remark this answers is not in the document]
eastandwestwind: Ah good idea, thank you! I think getting the last updated item makes sense in this context 👍

src/fides/api/service/messaging/messaging_crud_service.py
eastandwestwind: oh interesting, I didn't know that SQLalchemy would use the `MessagingTemplateToProperty` table here. Thanks for pointing this out! Do you have any additional resources for learning how SQLAlchemy emits queries behind the scenes? I have a lot to learn about SQLAlchemy overall.

pattisdr: you can wrap a query object that's ready to be executed but not actually executing in a print statement! for example(db.query().filter()...) drop the .first() or .all() -

pattisdr (approved):

src/fides/api/service/messaging/messaging_crud_service.py
eastandwestwind: ohhh thank you!

PR #4907, opened 2024-05-22
WIP: PROD-1983 Privacy Center custom paths for properties

lucanovera:
### Description Of Changes

Update the privacy center to accept the new custom paths for properties. ie. mypravacycenter.com/mynewproperty. If a custom path was used, it will call the new "get property by path" plus endpoint and display the privacy center. It will also load the privacy center config & stylesheet from that specific property.

### Code Changes

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`


clients/privacy-center/components/modals/consent-request-modal/ConsentRequestModal.tsx
```diff
@@ -19,7 +19,11 @@
   );
 
   const successHandler = useCallback(() => {
-    router.push("consent");
+    let consentRoute = "/consent";
+    if (router.query.customPropertyPath) {
+      consentRoute = `/${router.query.customPropertyPath}/consent`;
+    }
+    router.push(consentRoute);
```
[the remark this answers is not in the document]
lucanovera: Hmmm I see why this may cause a warning. However, if that customPropertyPath were to have a malicious content, the page would render a 404 page. It will only render if we get a result back from the _get properties by path_ endpoint. Also, the resulting history push starts with a '/', so it should stay on the domain.

clients/privacy-center/features/common/property.slice.ts
```diff
@@ -0,0 +1,37 @@
...
+  initialState,
+  reducers: {
+    /**
+     * Load property, replacing the current state entirely.
+     */
+    loadProperty(draftState, { payload }: PayloadAction<Property | undefined>) {
+      if (process.env.NODE_ENV === "development") {
+        // eslint-disable-next-line no-console
+        console.log("Loading Privacy Center property into Redux store...");
+      }
```
gilluminate, line 23: It might be worth creating a `debugLog` util for privacy center like fides-js has. Not necessary before merge, just a suggestion for future reference.

clients/privacy-center/app/server-environment.ts
```diff
@@ -287,156 +255,99 @@ export const loadStylesFromFile = async (
...
-    }
-    // DEFER: Log a version number here (see https://github.com/ethyca/fides/issues/3171)
-    if (process.env.NODE_ENV === "development") {
-      console.log("Load Privacy Center environment for session...");
-    }
+export const loadPrivacyCenterEnvironment = async ({
+  customPropertyPath,
+}: {
+  customPropertyPath?: string;
+}): Promise<PrivacyCenterEnvironment> => {
```
gilluminate, line 262: can we update the interface here to have the prop be optional instead of requiring it and forcing the use of an empty `{}` in existing code?
```suggestion
export const loadPrivacyCenterEnvironment = async ({
  customPropertyPath,
}: {
  customPropertyPath?: string;
} | undefined): Promise<PrivacyCenterEnvironment> => {
```

clients/privacy-center/pages/api/fides-js.ts
```diff
@@ -102,7 +102,7 @@ export default async function handler(
   res: NextApiResponse
 ) {
   // Load the configured consent options (data uses, defaults, etc.) from environment
-  const environment = await loadPrivacyCenterEnvironment();
+  const environment = await loadPrivacyCenterEnvironment({});
```
gilluminate, line 105: Follow up from comment above about making this property optional...
The complexity is being surfaced to the interface here, whereas it's always best to keep complexity down inside the method.

gilluminate (changes requested):
Looking good overall, just one update requested to reduce complexity

clients/privacy-center/app/server-environment.ts
lucanovera: Yess, that would be better. Let me change that.

PR #4920, opened 2024-05-23
Adding request overrides for consent requests

galvana:
Closes [#PROD-1526](https://ethyca.atlassian.net/browse/PROD-1526)

### Description Of Changes

Adding the ability to specify SaaS request overrides of type `OPT_IN` and `OPT_OUT`

### Code Changes

* [ ] Added `OPT_IN` and `OPT_OUT` request override types and related tests
* [ ] Misc cleanup of existing request override fixtures and tests

### Steps to Confirm

* [ ] Refer to new tests for consent request overrides

### Pre-Merge Checklist

* [ ] Update `CHANGELOG.md`


src/fides/api/service/saas_request/override_implementations/mailchimp_request_overrides.py
galvana, line 1: Moving this to `tests` so it's not mixed in with the overrides used for actual connectors

tests/ops/integration_tests/saas/connector_runner.py
```diff
@@ -462,9 +466,10 @@ def dataset_config(
     return dataset
 
 
-def _privacy_preference_history(
+def create_privacy_preference_history(
```
galvana, line 469: Making this public since I used it outside of the `connector_runner`

tests/ops/integration_tests/saas/request_override/test_consent_request_override_task.py
```diff
@@ -0,0 +1,147 @@
...
+
+        saas_config = saas_consent_request_override_connection_config.get_saas_config()
+        dataset_name = saas_config.fides_key
+
+        privacy_request.consent_preferences = [
+            {"data_use": "marketing.advertising", "opt_in": opt_in}
+        ]
+        privacy_request.save(db)
+
+        spy = mocker.spy(SaaSRequestOverrideFactory, "get_override")
```
galvana, line 79: Verifying that the test calls the right request override function by spying on the `get_override` method of `SaaSRequestOverrideFactory`. I wasn't able to figure out how to spy on the actual request override functions, something about the way the functions are dynamically invoked that doesn't work in a straightforward way with the pytest spy

src/fides/api/service/saas_request/override_implementations/mailchimp_request_overrides.py
adamsachs: oh nice, great call! that was confusing before.

src/fides/api/service/saas_request/saas_request_override_factory.py
```diff
@@ -42,8 +44,8 @@ class SaaSRequestOverrideFactory:
 
     @classmethod
     def register(cls, name: str, request_types: List[SaaSRequestType]) -> Callable[
-        [Callable[..., Union[List[Row], int, None]]],
-        Callable[..., Union[List[Row], int, None]],
+        [Callable[..., Union[List[Row], int, bool, None]]],
+        Callable[..., Union[List[Row], int, bool, None]],
```
adamsachs, line 48: may be time for us to alias this type annotation 😛 

don't need to adjust here, but throwing it out here in case you want to!

src/fides/api/service/connectors/saas_connector.py
```diff
@@ -786,6 +797,41 @@ def _invoke_masking_request_override(
...
+        client: AuthenticatedClient,
+        policy: Policy,
+        privacy_request: PrivacyRequest,
+        query_config: SaaSQueryConfig,
+        secrets: Any,
+    ) -> bool:
+        """
+        Invokes the appropriate user-defined SaaS request override for consent requests.
+
+        Includes the necessary data preparations for override input
```
adamsachs, line 812: nit: this looks like a c+p error, i don't think there's any data prep done here!

adamsachs (approved):
@galvana this looks solid to me, nice work! i like that this override framework is continuing to expand in what feels like a pretty natural way :)

i left a few nits that you can feel free to take or leave. i assume there's going to be a _real_ use case coming up to leverage these overrides, in which case we'll put it to the test! that's usually the best confirmation that the override has what it needs and there aren't any unexpected issues - but i think it makes sense to put the (proposed) framework updates you have here as their own increment, and then follow up with the first implementation after 👍

PR #4927, opened 2024-05-28
Updating privacy preference tables

galvana:
Closes [PROD-2109](https://ethyca.atlassian.net/browse/PROD-2109)

### Description Of Changes

Table and model changes to support the API driven consent feature. Most of the changes are caused by the addition of the `external_id` default identity.

### Code Changes

* [ ] Migration to add `external_id` and `hashed_external_id` columns (where applicable) and conditional logic to skip creating indices for large tables
* [ ] Updating the `POST /consent-request` endpoint to support the new `external_id` default identity
* [ ] Relevant tests + minor fixture cleanup

### Steps to Confirm

* [ ] Follow testing steps here https://github.com/ethyca/fidesplus/pull/1443

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

[PROD-2109]: https://ethyca.atlassian.net/browse/PROD-2109?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pattisdr (commented):
Early comment

src/fides/api/alembic/migrations/versions/5fe01e730171_add_external_id_to_privacy_preference.py
```diff
@@ -0,0 +1,219 @@
...
+    op.create_unique_constraint(
+        "last_saved_for_fides_user_device_per_property_id",
+        "currentprivacypreferencev2",
+        ["fides_user_device", "property_id"],
+    )
+    op.create_unique_constraint(
+        "last_saved_for_phone_number_per_property_id",
+        "currentprivacypreferencev2",
+        ["phone_number", "property_id"],
+    )
```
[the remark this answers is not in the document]
galvana: Thanks for this nice write up! I didn't know about Squawk, I use this moving forward. I was thinking I had to schedule downtime for this, I even asked Neville about our downtime allotment. I'll try the recommendation here to create the index concurrently and create the constraint using the index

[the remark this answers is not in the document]
galvana: @pattisdr I updated the constraint creation commands to use `CREATE USING INDEX` so that we could run them without downtime. I also moved the constraint creation inside the < 1 million row condition. The post-upgrade scripts in Fidesplus have the new `ix_currentprivacypreferencev2_*` composite indices and the new constraint creation commands.

pattisdr:
Starting review!

src/fides/api/api/v1/endpoints/consent_request_endpoints.py
```diff
@@ -590,6 +618,8 @@ def infer_target_identity_type(
         target_identity_type = ProvidedIdentityType.email.value
     elif identity_data.phone_number:
         target_identity_type = ProvidedIdentityType.phone_number.value
+    elif identity_data.external_id:
+        target_identity_type = ProvidedIdentityType.external_id.value
```
pattisdr, line 622: I was a little confused in testing but I think this behavior is okay? I feel like it is what it is, trying to maintain multiple workflows for saving consent in the privacy center.

Consent requests from the privacy center have traditionally been saved against one identity type. It was originally just an email or phone, but then we had a customer request to allow this privacy center consent request to work with only the fides user device id. We prioritized email or phone over fides user device id if they had been provided/verified, but the fides user device id was a catch-all
It seems possible both the external id and fides user device id are supplied. I think it makes sense to prioritize the external id if it exists over the fides user device id as you've done here. But it's important to keep in mind that only one ProvidedIdentity is created. Only the external id would be saved against the consent request.
This one provided identity takes priority if subsequent requests are sent to save privacy requests or notices served with other identities.
So potentially confusing scenario:

Create a consent request with both an external id and a fides user device. Causes a ProvidedIdentity to be created for just external id
Save notices served under the same consent id:
If just fides user device id is in the request body, external id will be magically added in from the "consent request"
If identities are empty, external id alone will show up against the preferences, even though the fides user device was in the original consent request
If external id is different now in request body, it will be overridden by the external id in the original consent request
What do you think?

src/fides/api/models/privacy_preference.py
```diff
@@ -163,15 +182,28 @@ def __tablename__(self) -> str:
...
+        ),
+        UniqueConstraint(
+            "external_id",
+            "property_id",
+            name="last_saved_for_external_id_per_property_id",
+        ),
+        Index(
+            "idx_preferences_gin",
+            text("(preferences->'preferences') jsonb_path_ops"),
+            postgresql_using="gin",
```
pattisdr, line 206: Have you confirmed this index performs as expected?

src/fides/api/models/privacy_request.py
```diff
@@ -1307,6 +1307,7 @@ class ProvidedIdentityType(EnumType):
     ga_client_id = "ga_client_id"
     ljt_readerID = "ljt_readerID"
     fides_user_device_id = "fides_user_device_id"
+    external_id = "external_id"
```
pattisdr, line 1310: I don't think this enum is used in the database but can you double check if you haven't?

src/fides/api/alembic/migrations/versions/5fe01e730171_add_external_id_to_privacy_preference.py
```diff
@@ -0,0 +1,288 @@
...
+    ).scalar()
+    if currentprivacypreferencev2_count < 1000000:
+        op.create_index(
+            op.f("ix_currentprivacypreferencev2_hashed_external_id"),
+            "currentprivacypreferencev2",
+            ["hashed_external_id"],
+            unique=False,
+        )
+        op.create_index(
+            op.f("ix_currentprivacypreferencev2_property_id"),
```
pattisdr, line 85: this one isn't referenced in the instructions you print to the user

pattisdr:
Starting review!

src/fides/api/api/v1/endpoints/consent_request_endpoints.py
pattisdr: good code comment -

src/fides/api/models/privacy_request.py
pattisdr: excellent

src/fides/api/models/privacy_preference.py
```diff
@@ -153,6 +166,12 @@ def __tablename__(self) -> str:
 
     fides_string = Column(String)
 
+    property_id = Column(
+        String,
+        index=True,
```
pattisdr, line 171: I'd remove `index=True` - I don't think you're adding this index in the migration, and we probably don't want want/need the overhead anyway.

src/fides/api/models/privacy_preference.py
```diff
@@ -153,6 +166,12 @@ def __tablename__(self) -> str:
 
     fides_string = Column(String)
 
+    property_id = Column(
```
pattisdr, line 169: Was this just added to get the privacy center version of the Get preferences endpoint working? (Do you think this change makes sense, outside of my specific request for just that one endpoint? Just wanted to understand your POV here, not necessarily disagreeing).

src/fides/api/alembic/migrations/versions/5fe01e730171_add_external_id_to_privacy_preference.py
```diff
@@ -0,0 +1,269 @@
...
+            "- 'ALTER TABLE currentprivacypreferencev2 "
+            "ADD CONSTRAINT last_saved_for_external_id_per_property_id "
+            "UNIQUE USING INDEX ix_currentprivacypreferencev2_external_id_property_id'\n"
+            "- 'ALTER TABLE currentprivacypreferencev2 "
+            "ADD CONSTRAINT last_saved_for_fides_user_device_per_property_id "
+            "UNIQUE USING INDEX ix_currentprivacypreferencev2_fides_user_device_property_id'\n"
+            "- 'ALTER TABLE currentprivacypreferencev2 "
+            "ADD CONSTRAINT last_saved_for_phone_number_per_property_id "
+            "UNIQUE USING INDEX ix_currentprivacypreferencev2_phone_number_property_id'"
+        )
```
pattisdr, line 174: Because this is non-standard, we haven't had an instance of this in our codebase, can you add code comments to the migration particularly around how you're creating the unique index concurrently, and then using that index that is already unique to create a constraint and why to help future devs?

pattisdr (approved):
Good work - just a few small changes requested -

I think the operations spelled out to take place separately without downtime should work without locking the db (creating a unique index concurrently first and then using that to create the unique constraint). Has this been tested?

src/fides/api/alembic/migrations/versions/5fe01e730171_add_external_id_to_privacy_preference.py
galvana: Added this comment to the top of the migration
```
WARNING - Conditional migration based on table size

This migration script checks the size of specific tables and performs the following actions:

1. If the tables have less than 1 million rows:
   - Creates the necessary indices
   - Adds the required constraints

2. If the tables are too large (1 million rows or more):
   - Logs non-blocking SQL commands to create the indices and constraints
   - The non-blocking commands utilize:
     - `CREATE INDEX CONCURRENTLY` for creating indices
     - `ADD CONSTRAINT USING INDEX` for adding constraints
   - These commands add the indices and constraints without blocking any reads on the tables
```

pattisdr: nice this looks great -

PR #4928, opened 2024-05-28
FIx privacy center paths

lucanovera:
### Description Of Changes

Fixprivacy center custom paths not working in production because it's using the wrong url to connect within docker.


galvana (approved):
👍

PR #4929, opened 2024-05-29
To address issue raised in PROD-2074 fix failing test connection

MarcGEthyca:
Closes #PROD-2074

### Description Of Changes

The prior test connection was a call out to an endpoint with a test user specified. I found the endpoint /api/v2/users/count which will just return an approximate count of users. It requires no other params than just those required for auth so it is a solid choice for a test request

### Code Changes

Modified the config.yml, editing the  test_request to point to the new endpoint and removed params that aren't needed for this new endpoint

### Steps to Confirm

* [ ] Add/configure the connector in UI
* [ ] Verify config and icon UI looks good
* [ ] Ensure the test integration button works
Note: Lack of current Zendesk account could complicate testing

### Pre-Merge Checklist

* Documentation:
  no changes needed
* [x] Update `CHANGELOG.md`


Kelsey-Ethyca (approved):

MarcGEthyca:
While we can not test this fix directly, we are confident it will address the issue and in this case would like to ask our reporting customer to verify this is remedied once they can test it (or if we find another means of testing it first).

PR #4930, opened 2024-05-29
s3 connector (for data detection & discovery) POC - fides

adamsachs:
Closes [PROD-2151](https://ethyca.atlassian.net/browse/PROD-2151)

‼️ includes a DB migration, bump downrev as needed!

Needed for https://github.com/ethyca/fidesplus/pull/1456. See loom there for a demo of the connector functionality that's been added here..

### Description Of Changes

OSS-side changes to support s3 as (initially) a d&d data source. Eventually we'll want to support S3 as a DSR data source, but that's not supported for now.

This PR creates an "S3" connection type that provides a space for AWS authentication credentials to be stored. We implement a new (and improved!) AWS auth paradigm that can leverage an optional "assume role ARN" to assume a specific role for S3 communications. This paradigm is not yet leveraged for our other AWS connections (e.g. DynamoDB), but the underlying utility can easily be leveraged, and we should look to support this paradigm for those connection types soon.

The S3 connection type implements a basic connection test to list buckets (as that's a key operation for our d&d monitor), but no support for the methods that need to be implemented for DSR processing.

### Code Changes

* [x] add `S3Connector`, `ConnectionType.s3` enum, and `S3Schema` for secrets/credentials. add logo for the FE
* [x] refactor, rename and improve an `aws_util..py` that can be used generically to get an AWS session with provided credentials that optionally assumes a provided role ARN

### Steps to Confirm

* [ ] connection type can be selected in the UI and tested successfully - with and without a provided assume role ARN
* [ ] further manual testing done while integrating with the associated fidesplus PR

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

[PROD-2151]: https://ethyca.atlassian.net/browse/PROD-2151?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/fides/api/util/aws_util.py
adamsachs, line 1: changed name of  `storage_authenticator.py` to make this more generic. also updated some functionality too (see below)

src/fides/api/util/aws_util.py
```diff
@@ -0,0 +1,67 @@
...
+        sts_client.get_caller_identity()
+    elif auth_method == AWSAuthMethod.AUTOMATIC.value:
+        session = Session()
+        logger.info("Successfully created automatic session")
+    else:
+        logger.error("Auth method not supported for S3: {}", auth_method)
+        raise ValueError(f"Auth method not supported for S3: {auth_method}")
+
+    if assume_role_arn:
+        if sts_client is None:
```
adamsachs, line 46: i wasn't sure if it was safe to call the 
```
sts_client = session.client("sts")
sts_client.get_caller_identity()
```
block for all use cases with the "automatic"  auth method, since we hadn't been doing that before. so rather than risk a regression, i'm only creating the sts client when secret keys are provided (as it had been done before), OR unless it's explicitly needed to assume role...the code is a bit harder to follow/less DRY, but i'd rather be safe for now...

src/fides/api/util/aws_util.py
```diff
@@ -0,0 +1,67 @@
...
+
+from boto3 import Session
+from botocore.exceptions import ClientError
+from loguru import logger
+
+from fides.api.common_exceptions import StorageUploadError
+from fides.api.schemas.storage.storage import AWSAuthMethod, StorageSecrets
+
+
+def get_aws_session(
```
adamsachs, line 11: more appropriately/generically named function now. the function here can and should be used by some of our other AWS connectors/codepaths, since this seems like a pretty solid authentication paradigm to promote across AWS resource types.

src/fides/api/service/connectors/s3_connector.py
```diff
@@ -0,0 +1,75 @@
...
+    """
+    AWS S3 Connector - this is currently used just to test connections to S3.
+
+    NOTE: No DSR processing is yet supported for S3.
+    """
+
+    def create_client(self) -> Any:  # type: ignore
+        """Returns a client for s3"""
+        config = S3Schema(**self.configuration.secrets or {})
+        return get_aws_session(
```
adamsachs, line 27: this will raise an exception if authentication fails - that'll be caught by the wrapping `test_connection` function and handled appropriately there, which is fine for now.

src/fides/api/schemas/connection_configuration/connection_secrets_dynamodb.py
```diff
@@ -25,6 +25,8 @@ class DynamoDBSchema(ConnectionConfigSecretsSchema):
         sensitive=True,
     )
 
+    # TODO: include an aws_assume_role_arn and more closely follow the pattern in `connection_secrets_s3`
```
adamsachs, line 28: i don't want to make our changes too wide-ranging quite yet, so i've just put in a note that we should look to switch dynamoDB over to this authentication paradigm soon. dynamoDB does require a region (I think?), so i'm not sure how that changes things. but I still think that provided an option to assume a role will be the right thing to support here moving forward 👍

adamsachs (commented):
some self-review

src/fides/api/schemas/connection_configuration/connection_secrets_s3.py
```diff
@@ -0,0 +1,54 @@
...
+    aws_secret_access_key: Optional[str] = Field(
+        title="Secret Access Key",
+        description="Part of the credentials that provide access to your AWS account. This is required if using secret key authentication.",
+        sensitive=True,
+    )
+
+    aws_assume_role_arn: Optional[str] = Field(
+        title="Assume Role ARN",
+        description="If provided, the ARN of the role that should be assumed to connect to s3.",
+    )
```
daveqnet, line 33: Schema looks good! From my perspective:

- Access Key ID: not particularly sensitive, a bit like a username
- Secret Access Key: sensitive, treat it like a pssword
- Assume Role ARN: not particularly sensitive, again a bit like a username

src/fides/api/util/aws_util.py
```diff
@@ -0,0 +1,67 @@
+from typing import Any, Dict, Optional
+
+from boto3 import Session
+from botocore.exceptions import ClientError
+from loguru import logger
+
+from fides.api.common_exceptions import StorageUploadError
+from fides.api.schemas.storage.storage import AWSAuthMethod, StorageSecrets
+
```
daveqnet, line 9: Just a general comment on the utility here rather than any specific line of code: great idea to make a generic utility than can be reused, AWS auth won't change.

As I understand it, the utility gives three authn options:

1. Use an IAM user's long-term access key to authenticate for the entire session.
2. Rely on boto3's automatic method of checking the local environment for different ways AWS credentials can be provided by the env, and authenticate using them for the session. There are a few of these, ranging from env vars to using the EC2 instance metadata service, all documented by AWS [here].(https://boto3.amazonaws.com/v1/documentation/api/latest/guide/credentials.html) 
3. Use an IAM user's long-term access key to get temporary security credentials (a.k.a. assume an IAM role) and use these temporary creds for the session.

adamsachs: that's _just about_ spot on! the only slight difference is that i've also allowed the AWS credentials derived by boto3 (i.e. option #2) to be used in step #3, i.e. to assume an IAM role (if a user has specified `automatic` authentication method _while also_ providing an assume role ARN). 

is that even a valid use case? i was a bit unsure whether this is something that can (or should) be supported. if not, i can easily clean up the code and make this a bit more straightforward.

adamsachs:
@pattisdr i wanted to get this OSS-side PR on your queue for review, no big rush though. i'm still working through the final touches on the plus-side to get it ready for review, but i think this can be reviewed in isolation.

pattisdr:
Starting review - apologies for the merge conflicts 😓

src/fides/api/schemas/connection_configuration/connection_secrets_s3.py
```diff
@@ -0,0 +1,54 @@
...
+from fides.api.schemas.connection_configuration.connection_secrets import (
+    ConnectionConfigSecretsSchema,
+)
+from fides.api.schemas.storage.storage import AWSAuthMethod
+
+
+class S3Schema(ConnectionConfigSecretsSchema):
+    """Schema to validate the secrets needed to connect to Amazon S3"""
+
+    auth_method: AWSAuthMethod = Field(
```
pattisdr, line 15: I got an unhandled 500 error when attempting to add an invalid authentication method for S3:

```
2024-06-13 20:23:45.752 | INFO     | fides.api.util.connection_util:validate_secrets:112 - Validating secrets on connection config with key 's3'
2024-06-13 20:23:45.752 | ERROR    | fides.api.util.logger:_log_exception:32 - Object of type AWSAuthMethod is not JSON serializable
```

pattisdr: Is there an existing way to prompt to the end user what their options are here?  Seems like an improvement we might ticket

src/fides/api/alembic/migrations/versions/cb344673f633_add_s3_connection_type.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+def upgrade():
+    # Add 's3' to ConnectionType enum
+    op.execute("alter type connectiontype rename to connectiontype_old")
+    op.execute(
+        "create type connectiontype as enum('mongodb', 'mysql', 'https', 'snowflake', "
+        "'redshift', 'mssql', 'mariadb', 'bigquery', 'saas', 'manual', 'manual_webhook', "
+        "'timescale', 'fides', 'sovrn', 'attentive', 'dynamodb', 'postgres',"
+        "'generic_consent_email', 'generic_erasure_email', 's3')"
```
pattisdr, line 26: this is the most important place to make sure we add scylla, others will get flagged with merge conflicts

src/fides/api/service/connectors/s3_connector.py
```diff
@@ -0,0 +1,75 @@
...
+    def retrieve_data(
+        self,
+        node: ExecutionNode,
+        policy: Policy,
+        privacy_request: PrivacyRequest,
+        request_task: RequestTask,
+        input_data: Dict[str, List[Any]],
+    ) -> List[Row]:
+        """DSR execution not yet supported for S3"""
+        raise NotImplementedError()
```
pattisdr, line 60: As discussed in my PR this can break the DSR workflow if an S3 integration exists that has a dataset.

src/fides/api/service/connectors/s3_connector.py
```diff
@@ -0,0 +1,75 @@
...
+class S3Connector(BaseConnector):
+    """
+    AWS S3 Connector - this is currently used just to test connections to S3.
+
+    NOTE: No DSR processing is yet supported for S3.
+    """
+
+    def create_client(self) -> Any:  # type: ignore
+        """Returns a client for s3"""
+        config = S3Schema(**self.configuration.secrets or {})
```
pattisdr, line 26: if secrets only have an authentication method, nothing else, the Connection Test will be successful -

src/fides/api/schemas/storage/storage.py
```diff
@@ -48,15 +48,15 @@ class Config:
         extra = Extra.forbid
 
 
-class S3AuthMethod(Enum):
+class AWSAuthMethod(Enum):
```
pattisdr, line 51: Nice making this more generic

src/fides/api/service/connectors/s3_connector.py
```diff
@@ -0,0 +1,75 @@
...
+        """
+        Connects to AWS S3 and lists buckets to validate credentials.
+        """
+        logger.info("Starting test connection to {}", self.configuration.key)
+        try:
+            session = self.client()
+            s3 = session.resource("s3")
+            s3.buckets.all()
+        except Exception as error:
+            raise ConnectorFailureException(str(error))
```
pattisdr, line 47: I'd raise a ConnectionException instead - the endpoint that tests connections is looking for this exception. Otherwise, you'll have unhandled failures that will return a 500.  (Also looks like the UI shows green in that case), separate issue.

src/fides/api/schemas/connection_configuration/connection_secrets_s3.py
```diff
@@ -0,0 +1,54 @@
...
+        title="Access Key ID",
+        description="Part of the credentials that provide access to your AWS account. This is required if using secret key authentication.",
+    )
+    aws_secret_access_key: Optional[str] = Field(
+        title="Secret Access Key",
+        description="Part of the credentials that provide access to your AWS account. This is required if using secret key authentication.",
+        sensitive=True,
+    )
+
+    aws_assume_role_arn: Optional[str] = Field(
```
pattisdr, line 30: hm I never did get this to work in testing, just using the user urn. in the sandbox creds

src/fides/api/service/connectors/s3_connector.py
pattisdr, line 47: This is logging sensitive credentials to the terminal

src/fides/api/util/aws_util.py
```diff
@@ -0,0 +1,67 @@
...
+        logger.info("Successfully created automatic session")
+    else:
+        logger.error("Auth method not supported for S3: {}", auth_method)
+        raise ValueError(f"Auth method not supported for S3: {auth_method}")
+
+    if assume_role_arn:
+        if sts_client is None:
+            sts_client = session.client("sts")
+        try:
+            response = sts_client.assume_role(
```
pattisdr, line 49: hm i haven't gotten this operation to work yet with the creds referenced in the plus PR - 

An error occurred (AccessDenied) when calling the AssumeRole operation: <redacted> is not authorized to perform: sts:AssumeRole on resource: <redacted>

src/fides/api/schemas/connection_configuration/connection_secrets_s3.py
pattisdr: I don't know much about ARN's, reading here: https://docs.aws.amazon.com/IAM/latest/UserGuide/reference-arns.html, can you explain/point to good docs about what this particular param is for?

src/fides/api/schemas/connection_configuration/connection_secrets_s3.py
adamsachs: oof that error isn't good! let me follow up on that.

in terms of providing more explicit guidance for the end users, i completely agree - we're building out the new "integrations" UI for S3 as we speak, so hopefully that can be accommodated there - we are typing this pretty strongly/explicitly on the BE so it feels like that _should_ be do-able...

src/fides/api/service/connectors/s3_connector.py
adamsachs: yup sounds good, i'll adjust to just no-op 👍

src/fides/api/service/connectors/s3_connector.py
adamsachs: OK! clearly i need to treat this function a bit better, i rushed it a bit - i'll dig into it further, thanks for highlighting.

src/fides/api/service/connectors/s3_connector.py
adamsachs: thanks for flagging that!

adamsachs: 😱 

clearly need to address that.

src/fides/api/schemas/connection_configuration/connection_secrets_s3.py
adamsachs: err yeah so the use case here is to use a _different_ ARN whose role you want to assume, using the provided credentials as a "starting" user. 

honestly i'm not sure the best reading for this - it's a concept i learned about a little while ago, and @daveqnet refreshed my memory nicely [here](https://ethyca.slack.com/archives/C01TUEAGCR1/p1718027044938379?thread_ts=1718025829.889269&cid=C01TUEAGCR1), which i think helps explain the use case? 

i can pass along in slack the role ARN that i'm assuming, as a test! it's basically a standard "s3 full access" role.

src/fides/api/util/aws_util.py
adamsachs: yes sorry i wasn't clear about this functionality! passed along a role ARN on slack

src/fides/api/alembic/migrations/versions/cb344673f633_add_s3_connection_type.py
adamsachs: thanks for calling out, i reconciled in https://github.com/ethyca/fides/pull/4930/commits/11d9b63e63c467c1edad4e318e8463ca6051adec

src/fides/api/service/connectors/s3_connector.py
adamsachs: updated in https://github.com/ethyca/fides/pull/4930/commits/1cfd96f4d75e57a8d28263c0bf4a49fdb1080022

src/fides/api/service/connectors/s3_connector.py
adamsachs: adjusted in https://github.com/ethyca/fides/pull/4930/commits/1cfd96f4d75e57a8d28263c0bf4a49fdb1080022

adamsachs: so i'm not seeing this now, with my most recent changes. frankly i hadn't seen it before (just because i hadn't been looking closely) - but could you let me know specifically what you were doing to cause the sensitive credentials to be logged, so that we can be sure it's no longer a problem?

src/fides/api/service/connectors/s3_connector.py
adamsachs: thanks for calling this out - my connection test was not a valid connection test! 😮 the command to retrieve buckets that i'd been calling just seemed to return an empty list with an invalid client, not actually fail.

i've adjusted the code to perform a better test - to get caller identity with sts. as the AWS documentatino notes, no permissions are required for this operation, and as far as i can tell, it's the generally recommended way to confirm valid credentials.

for what it's worth - if you have a valid `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` set in your `.env` (or just generally set as env vars in your container), then it's actually _expected_ that you'll be able to have a successful connection test using `automatic` authentication method with no credentials. i know the UX for that isn't the clearest, but i think it's acceptable for now. let me know if that doesn't make sense!

src/fides/api/schemas/connection_configuration/connection_secrets_s3.py
pattisdr: thanks for the info 👍

src/fides/api/service/connectors/s3_connector.py
pattisdr: ah okay thanks for explaining the `automatic` use case 👍

src/fides/api/service/connectors/s3_connector.py
pattisdr: I believe I was just using the user arn from your original credentials which was causing that to get logged, but i wasn't seeing the access ikd/key.  I'm not seeing this anymore, it looks good.

pattisdr (approved):
this looks good to me @adamsachs nice round of changes

adamsachs:
test is failing in `main` and is unrelated - will go ahead and merge 👍

PR #4933, opened 2024-05-29
PROD-2117 frontend generate served notice history guid and pass to be

lucanovera:
### Description Of Changes
- For notices-served and privacy-preferences, we want to generate the served_notice_history_id (a uuid v4) in the client instead of in the backend (that way we also don't have to wait for the response in order to use the same id for both requests).

### Code Changes

* [ ] Split hooks.ts into multiple files
* [ ] Remove await from privacy-preferences called to optimize speed
* [ ] Generate a starting uuidv4 and regenerate each time notices-served is called (same behavior as before)
* [ ] Use the same uuiv4 in notices-served request as for saved-preferences
* [ ] Update tests

### Steps to Confirm
* [x] Open development mode in the browsers, go to network tab
* [x] Trigger a fides modal
* [x] In the dev tools, find the served-notices request and check that the body of the request should include served_notice_history_id with a valid uuidv4
* [x] Click save
* [x] In the dev tools, find the preferences-saved request and check that the body of the request should include served_notice_history_id with the same uuidv4 as the served-notices request

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`


clients/fides-js/src/lib/hooks/useUUID4.ts
```diff
@@ -0,0 +1,14 @@
...
+ * Custom hook that generates a UUIDv4.
+ * The returned value stays the same for the lifetime of the component.
+ * @returns The generated UUIDv4.
+ */
+const useUUID4 = () => {
+  const [uuid] = useState<string>(uuidv4());
+
+  return uuid;
+};
+export default useUUID4;
```
jpople, line 14: This doesn't appear to me to get imported anywhere, is it still needed?

lucanovera: Good catch, it's not needed. It was used in an earlier approach, but then had to change it. I left it as it might be useful for someone,  but I can easly remove it too.

jpople: If you think it'll still be useful in the future I don't have a problem leaving it in, just wanted to make sure it wasn't a total oversight!

jpople (approved):
✅ Working as expected!

PR #4935, opened 2024-05-30
Fix: get the top-level hostname

gilluminate:
Closes [PROD-2027](https://ethyca.atlassian.net/browse/PROD-2027)

### Description Of Changes

IP addresses and top level hostnames that included multiple TLDs (like `co.uk`) were not working for setting a Fides Cookie.

This fix update cookie setter to guarantee highest level domain allowed to set cookie. It relies on browser's own methods to determine viable domain.

### Code Changes

* Loop through parts of hostname backwards until cookie gets set.
* Validate the cookie gets set and stop looping

### Steps to Confirm

* Update `/etc/hosts` on your local machine to include new line:
  ```127.0.0.1 privacy.example.co.uk```
* Fire up privacy center on localhost
* Instead of visiting localhost, visit `http://privacy.example.co.uk:3001/fides-js-demo.html`
* In the banner/modal make a selection that will set the cookie
* Verify that the cookie was set with domain `example.co.uk`
* repeat for the IP address `http://127.0.0.1:3001/fides-js-demo.html` and ensure that works.

(domains like `www.fides.com` and `localhost` should _continue_ working with this enhancement in place)

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

[PROD-2027]: https://ethyca.atlassian.net/browse/PROD-2027?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/clients/fides-js/__tests__/lib/cookie.test.ts
+++ b/clients/fides-js/__tests__/lib/cookie.test.ts
@@ -45,8 +45,9 @@ const mockGetCookie = jest.fn((): string | undefined => "mockGetCookie return");
 const mockSetCookie = jest.fn(
   /* eslint-disable-next-line @typescript-eslint/no-unused-vars */
   (name: string, value: string, attributes: object, encoding: object) => {
+    // Simulate that browsers will not write cookies to known top-level public domains like "com" or "co.uk"
     if (
-      ["com", "ca", "org", "uk", "co.uk"].indexOf(
+      ["com", "ca", "org", "uk", "co.uk", "in", "co.in", "jp", "co.jp"].indexOf(
         (attributes as { domain: string }).domain
       ) > -1
     ) {
@@ -260,6 +261,14 @@ describe("saveFidesCookie", () => {
     {
       url: "https://privacy.subdomain.example.co.uk",
       expected: "example.co.uk",
+    },
+    {
+      url: "https://example.co.in",
+      expected: "example.com.in",
+    },
+    {
+      url: "https://example.co.jp",
+      expected: "example.co.jp",
     },
   ])(
     "calculates the root domain from the hostname ($url)",
--- a/clients/fides-js/__tests__/lib/cookie.test.ts
+++ b/clients/fides-js/__tests__/lib/cookie.test.ts
@@ -44,8 +44,16 @@ mockUuid.v4.mockReturnValue(MOCK_UUID);
 const mockGetCookie = jest.fn((): string | undefined => "mockGetCookie return");
 const mockSetCookie = jest.fn(
   /* eslint-disable-next-line @typescript-eslint/no-unused-vars */
-  (name: string, value: string, attributes: object, encoding: object) =>
-    `mock setCookie return (value=${value})`
+  (name: string, value: string, attributes: object, encoding: object) => {
+    if (
+      ["com", "ca", "org", "uk", "co.uk"].indexOf(
+        (attributes as { domain: string }).domain
+      ) > -1
+    ) {
+      return undefined;
+    }
+    return `mock setCookie return (value=${value})`;
+  }
 );
 const mockRemoveCookie = jest.fn(
   /* eslint-disable-next-line @typescript-eslint/no-unused-vars */
@@ -249,6 +257,10 @@ describe("saveFidesCookie", () => {
     { url: "https://www.another.com", expected: "another.com" },
     { url: "https://privacy.bigco.ca", expected: "bigco.ca" },
     { url: "https://privacy.subdomain.example.org", expected: "example.org" },
+    {
+      url: "https://privacy.subdomain.example.co.uk",
+      expected: "example.co.uk",
+    },
   ])(
     "calculates the root domain from the hostname ($url)",
     ({ url, expected }) => {
@@ -259,8 +271,12 @@ describe("saveFidesCookie", () => {
       });
       const cookie: FidesCookie = getOrMakeFidesCookie();
       saveFidesCookie(cookie);
-      expect(mockSetCookie.mock.calls).toHaveLength(1);
-      expect(mockSetCookie.mock.calls[0][2]).toHaveProperty("domain", expected);
+      const numCalls = expected.split(".").length;
+      expect(mockSetCookie.mock.calls).toHaveLength(numCalls);
+      expect(mockSetCookie.mock.calls[numCalls - 1][2]).toHaveProperty(
+        "domain",
+        expected
+      );
     }
   );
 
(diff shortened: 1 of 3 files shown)


clients/fides-js/src/lib/cookie.ts
```diff
@@ -192,14 +192,10 @@ export const getOrMakeFidesCookie = (
  * Save the given Fides cookie to the browser using the current root domain.
  *
  * This calculates the root domain by using the last two parts of the hostname:
```
NevilleS, line 194:
```suggestion
 * This calculates the root domain by using the last parts of the hostname:
```

clients/fides-js/src/lib/cookie.ts
```diff
@@ -215,26 +211,33 @@ export const saveFidesCookie = (
...
-      path: "/",
-      // An explicit domain allows subdomains to access the cookie.
-      domain: rootDomain,
-      expires: CONSENT_COOKIE_MAX_AGE_DAYS,
-    },
-    CODEC
-  );
+  const hostname = window.location.hostname.split(".");
+  let topViableDomain = "";
+  for (let i = 1; i < hostname.length + 1; i += 1) {
```
NevilleS, line 221:
```suggestion
  for (let i = 1; i <= hostname.length; i += 1) {
```
I'm sorry this is so nitpicky but I had to

clients/fides-js/src/lib/cookie.ts
```diff
@@ -215,26 +211,33 @@ export const saveFidesCookie = (
...
+    // This loop guarantees to get the top-level hostname because that's the smallest one browsers will let you set cookies in. We test a given suffix for whether we are able to set cookies, if not we try the next suffix until we find the one that works.
+    topViableDomain = hostname.slice(-i).join(".");
```
NevilleS, line 223: So that I understand, for something like `app.privacy.example.com` this loop will do:
1. `com`
2. `example.com`
3. `privacy.example.com`
4. `app.privacy.example.com`

?

clients/fides-js/src/lib/cookie.ts
```diff
@@ -215,26 +211,33 @@ export const saveFidesCookie = (
...
-    {
-      // An explicit path ensures this is always set to the entire domain.
-      path: "/",
-      // An explicit domain allows subdomains to access the cookie.
-      domain: rootDomain,
-      expires: CONSENT_COOKIE_MAX_AGE_DAYS,
-    },
-    CODEC
-  );
+  const hostname = window.location.hostname.split(".");
```
NevilleS, line 219: nit:
```suggestion
  const hostnameLabels = window.location.hostname.split(".");
```

Since I'd reasonably assume a variable called `hostname` is a string `privacy.example.com` and not an array `["privacy", "example", "com"]`

NevilleS (approved):
nitpicky style/questions in there, but looks reasonable!

clients/fides-js/src/lib/cookie.ts
gilluminate: no, once the loop is able to set the cookie the loops stops with the break. In your example, the loop will do:

1. `com`
2. `example.com`

And now the cookie is set and will `break`

clients/fides-js/src/lib/cookie.ts
gilluminate: this was bugging me as well. I'll go with a more conventional `hostnameParts`

clients/fides-js/__tests__/lib/cookie.test.ts
```diff
@@ -44,8 +44,16 @@ mockUuid.v4.mockReturnValue(MOCK_UUID);
...
-  (name: string, value: string, attributes: object, encoding: object) =>
-    `mock setCookie return (value=${value})`
+  (name: string, value: string, attributes: object, encoding: object) => {
+    if (
+      ["com", "ca", "org", "uk", "co.uk"].indexOf(
+        (attributes as { domain: string }).domain
+      ) > -1
+    ) {
+      return undefined;
+    }
```
NevilleS, line 54: Aha, that's pretty reasonable and builds some decent confidence 👍!

In that case, my classic nit is just to provide some breadcrumbs here:
```suggestion
    // Simulate that browsers will not write cookies to known top-level public domains like "com" or "co.uk"
    if (
      ["com", "ca", "org", "uk", "co.uk"].indexOf(
        (attributes as { domain: string }).domain
      ) > -1
    ) {
      return undefined;
    }
```

clients/fides-js/src/lib/cookie.ts
```diff
@@ -215,26 +211,35 @@ export const saveFidesCookie = (
...
+      },
+      CODEC
+    );
+    if (c) {
+      const cookieString = getCookieByName(CONSENT_COOKIE_NAME);
+      if (cookieString) {
+        break;
+      }
+    }
+  }
```
NevilleS, line 242: If this exits and fails to save a cookie, what should we do?

clients/fides-js/__tests__/lib/cookie.test.ts
```diff
@@ -249,6 +257,10 @@ describe("saveFidesCookie", () => {
     { url: "https://www.another.com", expected: "another.com" },
     { url: "https://privacy.bigco.ca", expected: "bigco.ca" },
     { url: "https://privacy.subdomain.example.org", expected: "example.org" },
+    {
+      url: "https://privacy.subdomain.example.co.uk",
+      expected: "example.co.uk",
+    },
```
NevilleS, line 263: Just for completeness, I'd add one or two other examples like:
```suggestion
    {
      url: "https://privacy.subdomain.example.co.uk",
      expected: "example.co.uk",
    },
    {
      url: "https://example.co.in",
      expected: "example.com.in",
    },
    {
      url: "https://example.co.jp",
      expected: "example.co.jp",
    },
```

It's pretty duplicative in the code today, yes, but I think it feels right

NevilleS (commented):
All just nits! 👍

clients/fides-js/src/lib/cookie.ts
gilluminate: I wondered about throwing an error, but I thought the only real world scenario where this happens is if the browser is blocking 1st party cookies. Should we do something a little more invasive, like a browser `alert` to let the user know it's not going to work?

gilluminate: Actually, I propose we solve that separately. fides.js should probably check to see if cookies can be written prior to showing a banner at all. We should be thoughtful about how we want to handle it in that initialization phase.

PR #4946, opened 2024-06-03
ScyllaDB Connector

pattisdr:
Closes #PROD-2148

❗ Contains connection type migration. Bump downrev before merge.
👉  Supports https://github.com/ethyca/fidesplus/pull/1447

### Description Of Changes

Adds the infrastructure to run a local scylla db instance in Docker, and a simple integration to connect to it. Access and erasure requests are not yet supported for scylla.  This first step is just supporting scylla detection and discovery in Fidesplus.

### Code Changes

* [ ] Installs `scylla-driver`
* [ ] Adds a new scylla ConnectionType
* [ ] Adds support for bringing up a test scylla db in Docker: Adds a new docker-compose integration file for scylla db which uses the bitnami image, which allows you to put initialization scripts in the docker entrypoint directory, which doesn't seem to be supported in the regular cassandra/scylla db images.
* [ ] Adds sample cql file to populate database with fake data - creates two keyspaces, a couple of tables, and populates
* [ ] Adds the start of a Scylla Connector.  Retrieve data and mask data are not currently supported, just testing a connection.

### Steps to Confirm

* [ ] Run `nox -s dev -- shell scylladb`
* [ ] In fides/clients, `turbo run dev`
* [ ] Go to System > Add system, add name: Scylla, click save
* [ ] Click on integration tab, search for scylla:
Screenshot 2024-06-11 at 12 03 00 PM
* [ ] Add creds (these are not secret, these are test credentials to the scylla instance you're running in your own docker container locally) - `scylla_user`,`scylla_pass`
* [ ] Click test connection, verify this is green
Screenshot 2024-06-11 at 12 04 16 PM

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`


docker/docker-compose.integration-scylladb.yml
```diff
@@ -0,0 +1,15 @@
+services:
+  scylladb_example:
+    image: bitnami/scylladb:6.0.0
```
pattisdr, line 3: See comments on bitnami image in corresponding plus PR! https://github.com/ethyca/fidesplus/pull/1447#discussion_r1636464528

src/fides/api/alembic/migrations/versions/3304082a6cee_scylla_connection_type.py
```diff
@@ -0,0 +1,46 @@
...
+    op.execute(
+        "create type connectiontype as enum('mongodb', 'mysql', 'https', 'snowflake', 'redshift', 'mssql', 'mariadb', 'bigquery', 'saas', 'manual', 'manual_webhook', 'timescale', 'fides', 'sovrn', 'attentive', 'dynamodb', 'postgres', 'generic_consent_email', 'generic_erasure_email', 'scylla')"
+    )
+    op.execute(
+        (
+            "alter table connectionconfig alter column connection_type type connectiontype using "
+            "connection_type::text::connectiontype"
+        )
+    )
+    op.execute("drop type connectiontype_old")
```
pattisdr, line 31: Currently following the pattern of renaming the old enum, creating a new enum with the new connection type, switching to the new enum and deleting the old enum.  Is there any continued risk with this type of pattern? I think this sort of migration would be quick given the likely table sizes.

We discussed getting rid of the enum altogether but I'm not tackling that here for time -

pattisdr: Note to self to double check these enum values before merge -

src/fides/api/schemas/connection_configuration/connection_secrets_scylla.py
```diff
@@ -0,0 +1,46 @@
...
+    password: str = Field(
+        title="Password",
+        description="The password used to authenticate and access the database.",
+        sensitive=True,
+    )
+    keyspace: Optional[str] = Field(
+        title="Keyspace",
+        description="The keyspace used.",
+        sensitive=True,
+    )
```
pattisdr, line 36: Adding an optional keyspace here - might be useful to specify a specific one for DSR's, but this is not required for detection and discovery -

noxfiles/run_infrastructure.py
```diff
@@ -84,6 +85,8 @@ def run_infrastructure(
         if datastore in DOCKERFILE_DATASTORES
     ]
 
+    _run_cmd_or_err(f'echo "Docker datastores {docker_datastores}"')
```
adamsachs, line 88: nice touch for info!

src/fides/api/alembic/migrations/versions/3304082a6cee_scylla_connection_type.py
adamsachs: i don't think there's any risk beyond the extra bit of manual work required, especially for coordination when we've got multiple changes to this in-flight at the same time (as you mention - thanks for calling that out!)

but totally fine by me that you're not tackling getting rid of the enum here 👍

src/fides/api/schemas/connection_configuration/connection_secrets_scylla.py
adamsachs: interesting, i'm not sure how we'll want to handle this on the d&d side - ultimately i think we'll want to be able to leverage the same connection config for both DSRs and d&d monitors - so maybe it's just something that users will have "to know" that this won't impact the d&d workflow.

not worth any changes for now, i think how you've handled it is fine - just pondering...

src/fides/api/service/connectors/scylla_connector.py
```diff
@@ -0,0 +1,97 @@
...
+        logger.info("Starting test connection to {}", self.configuration.key)
+        config = ScyllaSchema(**self.configuration.secrets or {})
+        cluster = self.client()
+
+        try:
+            with cluster.connect(
+                config.keyspace if config.keyspace else None
+            ) as client:
+                client.execute("select now() from system.local")
+        except cassandra.cluster.NoHostAvailable as exc:
```
adamsachs, line 56: very thorough exception handling here! taking some notes on the test/error cases to improve my own handling in the s3 connector - seems worth the effort given that this is the primary diagnostic hook/tool for connection debugging 👍

src/fides/api/service/connectors/scylla_connector.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def retrieve_data(
+        self,
+        node: ExecutionNode,
+        policy: Policy,
+        privacy_request: PrivacyRequest,
+        request_task: RequestTask,
+        input_data: Dict[str, List[Any]],
+    ) -> List[Row]:
+        """Retrieve scylla data - not yet implemented"""
```
adamsachs, line 84: nit: worth raising a `NotImplementedError` here? that always feels like the most proper thing, just to make it transparent to anyone who somehow hits the codepath at runtime without reading the codebase, but maybe i'm overthinking it :)

adamsachs: now i'm wondering whether raising the error may unexpectedly impact other DSR workflows and i shouldn't be doing that... 🤔

tests/ops/api/v1/endpoints/test_connection_template_endpoints.py
```diff
@@ -339,7 +339,7 @@ def test_search_system_type(self, api_client, generate_auth_header, url):
         resp = api_client.get(url + "system_type=database", headers=auth_header)
         assert resp.status_code == 200
         data = resp.json()["items"]
-        assert len(data) == 10
+        assert len(data) == 11
```
adamsachs, line 342: i'll also need to update on my PR 😄

adamsachs (approved):
@pattisdr this side all looks solid to me! very thorough, nice job with everything.

as you said, let's just keep an eye on merge conflicts since we've got similar codepaths being updated from a few spots - thanks for calling that out 👍

src/fides/api/service/connectors/scylla_connector.py
pattisdr: If there are datasets for the connection config this could get picked up in a DSR I believe!

pattisdr: yep confirmed a NotImplementedError will break DSR's if a dataset is also attached to the ConnectionConfig .

PR #4960, opened 2024-06-06
set up tests for validating various domains

gilluminate:
### Description Of Changes

Set up tests and instructions to run against various real world TLDs and validate that our cookie logic is sound. Can only be run locally and manually, but handy to keep around for validating any future changes.

CleanShot 2024-06-06 at 10 46 01@2x

### DIFF
--- a/clients/privacy-center/cypress/e2e/consent-banner-cookies.cy.ts
+++ b/clients/privacy-center/cypress/e2e/consent-banner-cookies.cy.ts
@@ -1,0 +1,71 @@
+import { CONSENT_COOKIE_NAME } from "fides-js";
+import { stubConfig } from "~/cypress/support/stubs";
+
+const domains: {
+  domain: string;
+  expected: string;
+}[] = require("../fixtures/domains.json");
+
+/**
+ * This test is for validating our cookie domain logic. We want to ensure that cookies are able to be set on the topmost domain without needing to be the full domain.
+ *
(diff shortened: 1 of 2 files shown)


NevilleS (approved):
This is smart!

One minor suggestion is that, sometimes when I write little "leave-behind" tests like this, I like to write something to sanity-check the test preconditions where possible? Basically something that would fail really fast and obviously if you missed the comment and never configured `/etc/hosts`, because I suspect the test failures that will happen here if you run this _without_ that setup would be a bit cryptic.

Something like...
```
it("has a locally configured /etc/hosts file containing the domains.json for testing", () => {
  # Visit first domain and check that it loads! If this fails, see instructions above.
  cy.request(`http://${domains[0]}:3001`).should(...<check it resolves to localhost?>)
});
```

the idea is that test gives a really clear failure

gilluminate:
@NevilleS while I love that idea, unfortunately there's not a convenient way to stop the rest from running if that first test fails. Since the easiest way to run that test is by simply using `cy.visit`, at the end of the day it's already happening on the first test that runs (and all the rest). If there were a way to kill the rest of the tests on first failure I would do it, but it's going to be redundant without that feature.

NevilleS:
> @NevilleS while I love that idea, unfortunately there's not a convenient way to stop the rest from running if that first test fails. Since the easiest way to run that test is by simply using `cy.visit`, at the end of the day it's already happening on the first test that runs (and all the rest). If there were a way to kill the rest of the tests on first failure I would do it, but it's going to be redundant without that feature.

That's fair. The closest thing you have in Cypress is `beforeEach`, but that can be unnecessarily duplicative. Ship it

PR #4970, opened 2024-06-10
Add time choice to discovery monitor config form

jpople:
Closes PROD-2173

### Description Of Changes

Adds time selection to the discovery monitor config form.   

Screenshot 2024-06-10 at 16 51 00

Also some tweaks to form: 

* change label from "Execution start date" to "Automatic execution start time"  
* show datetime field when editing as well as on creation.
* remove validation requirement that datetime be in the future

### Code Changes

* [ ] Change date input to a datetime-local input

### Steps to Confirm

* [ ] Create a new monitor
* [ ] Edit the monitor
* [ ] Time entered for "Automatic execution start time" field should save and be populated

### Pre-Merge Checklist

* [ ] Update `CHANGELOG.md`


clients/admin-ui/src/features/common/form/inputs.tsx
```diff
@@ -1291,6 +1291,7 @@ export const CustomDatePicker = ({
             <Label htmlFor={props.id || name} fontSize="xs" my={0} mr={1}>
               {label}
             </Label>
+            {!!tooltip && <QuestionTooltip label={tooltip} />}
```
gilluminate, line 1294: :heart:

clients/admin-ui/src/features/common/form/inputs.tsx
```diff
@@ -1308,6 +1309,57 @@ export const CustomDatePicker = ({
...
+  const isInvalid = !!(meta.touched && meta.error);
+
+  const { validateField } = useFormikContext();
+
+  return (
+    <FormControl isRequired={isRequired} isInvalid={isInvalid}>
+      <VStack align="start">
+        {!!label && (
+          <Flex align="center">
+            <Label htmlFor={props.id || name} fontSize="xs" my={0} mr={1}>
```
gilluminate, line 1342: This `htmlFor` isn't actually doing anything, as far as I can tell. To be valid HTML, this `for` attribute needs to be the same as the `id` attribute value on the form field. The `<Input>` below isn't defining an `id` at all. The `for` attribute does not work with the `name` attribute.

https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/for

Unfortunately, it looks like this bad pattern exists in a lot of places in our app so we'll need to clean those up eventually.

gilluminate (changes requested):

clients/admin-ui/src/features/common/form/inputs.tsx
```diff
@@ -1308,6 +1309,60 @@ export const CustomDatePicker = ({
...
+  disabled,
+  isRequired,
+  ...props
+}: CustomInputProps & FieldHookConfig<string>) => {
+  const [field, meta, { setValue, setTouched }] = useField(name);
+  const isInvalid = !!(meta.touched && meta.error);
+
+  const { validateField } = useFormikContext();
+
+  const fieldId = props.id ?? name;
```
gilluminate, line 1337: Is there ever a chance that props.id could be an empty string `""`? If so the `||` operator is likely safer to use here.
```suggestion
  const fieldId = props.id || name;
```

gilluminate (approved):
Approved with one nit pick to look into.

PR #4972, opened 2024-06-10
disable the 'connect to a database' button if discovery + detection feature flag is on

adamsachs:
Partially closes [PROD-2138](https://ethyca.atlassian.net/browse/PROD-2138?focusedCommentId=41424)

Related to https://github.com/ethyca/fidesplus/pull/1455

### Description Of Changes

Disables the 'connect to a database' button in the UI if the discovery + detection feature flag is on.

This reduces confusion for end users while we still in the process of phasing out the 'legacy' scanning and classification workflows: if a user is using detection + discovery, then they should not be using the old "connect to a database" functionality.

### Code Changes

* [x] disable the 'connect to a database' button in the UI if the discovery + detection feature flag is on

### Steps to Confirm

* [x] tested both positive and negative cases locally, see screenshot below for when the feature flag is enabled (and button is disabled):

image

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

[PROD-2138]: https://ethyca.atlassian.net/browse/PROD-2138?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


clients/admin-ui/src/features/dataset/DatabaseConnectForm.tsx
```diff
@@ -245,7 +245,11 @@ const DatabaseConnectForm = () => {
                 colorScheme="primary"
                 type="submit"
                 isLoading={isSubmitting || isLoading}
-                isDisabled={isSubmitting || isLoading}
+                isDisabled={
+                  isSubmitting ||
+                  isLoading ||
+                  features.flags.dataDiscoveryAndDetection
+                }
```
NevilleS, line 252: This looks technically correct, but I think it'd be less confusing to go further up in the workflow and completely hide the "Connect to a database" option when this flag is enabled.

i.e this screen:
image

NevilleS (changes requested):

NevilleS:
Also, re: this comment:
> not sure if there's a way to disable a feature flag for particular cypress tests, or if i should just remove these tests since we want to deprecate this button altogether soon...

To be very thorough in my answer, since you're a thorough guy...
1. It's definitely *possible* to disable a flag for a single test. You could navigate to `/management/about` and disable the flag in the test itself and navigate back... or it'd also be possible to trigger the Redux action directly 😄 
2. It's very reasonable to also disable these the legacy tests with `it.skip` and `describe.skip` though. You're right, the go-forward is the new monitoring UI, so regressions in the legacy fashion isn't our primary concern. You have my blessing 👍

clients/admin-ui/src/features/dataset/DatabaseConnectForm.tsx
adamsachs: adjusted this in https://github.com/ethyca/fides/pull/4972/commits/d74c03f9bb597d4d2c0fca29981f57340b00a730

adamsachs:
thanks for your comments @NevilleS! i made adjustments per your suggestions and i think this is good to go now 👍

NevilleS (approved):
LGTM. I pushed a minor nit to just add a tooltip as a hint (mostly for ourselves!) since it's non-obvious why that button is disabled:
image

adamsachs:
> LGTM. I pushed a minor nit to just add a tooltip as a hint (mostly for ourselves!) since it's non-obvious why that button is disabled: image


that's great, thanks for that polish! i also was thinking something like that would be nice, just wasn't confident enough in my FE abilities to venture there 😆

PR #4983, opened 2024-06-13
PROD-2058 implement endpoints for property specific messaging

eastandwestwind:
Closes https://ethyca.atlassian.net/browse/PROD-2058

### Description Of Changes

Implements API endpoints for property-specific messaging. The following new endpoints were added:

- GET `/messaging/templates/summary`- get paginated list of all templates
- GET `/messaging/templates/default/{template_type}`- get the default messaging template by type. This is not dependent on the DB at all, rather hard-coded in the BE.
- POST `/messaging/templates/{template_type}`- create a new messaging template . Use req body formatted like:

```
content: Dict[str, Any]
properties: Optional[List[str]]
is_enabled: bool
```

- PUT `/messaging/templates/{id}`- update an existing messaging template. Use req body formatted like:

```
content: Dict[str, Any]
properties: Optional[List[str]]
is_enabled: bool
```

- PATCH  `/messaging/templates/{id}`- update an existing messaging template. Use req body formatted like:

```
content: Optional[Dict[str, Any]]
properties: Optional[List[str]]
is_enabled: Optional[bool]
```

- GET `/messaging/templates/{id}`- get an existing messaging template by id
- DELETE `/messaging/templates/{id}` - delete a messaging template by id

### Code Changes

* [ ] Adds new endpoints to support property-specific messaging
* [ ] Adds endpoint tests

### Steps to Confirm

* [ ] Attempt to add / edit / create /delete messaging templates via endpoints. See endpoint tests in `ops/api/v1/endpoints/test_messaging_endpoints.py` for details of how the endpoints should function.

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`


eastandwestwind:
Pipeline failures dependent on https://github.com/ethyca/fides/pull/4993 being merged

src/fides/api/api/v1/endpoints/messaging_endpoints.py
```diff
@@ -565,4 +588,190 @@ def update_basic_messaging_templates(
...
+    )
+
+
+@router.get(
+    MESSAGING_TEMPLATE_DEFAULT_BY_TEMPLATE_TYPE,
+    dependencies=[Security(verify_oauth_client, scopes=[MESSAGING_TEMPLATE_UPDATE])],
+    response_model=MessagingTemplateDefault,
+)
+def get_default_messaging_template(
+    template_type: MessagingActionType,
```
galvana, line 623: Using `MessagingActionType` for this input is too open since it's possible for the user to select a message type that doesn't have a default value and this would return a 500 error. We should either:
- Return a message saying that specific message types don't have a default value
- Use a more restrictive type and update the docstring to reflect that this endpoint will only return defaults for the configurable messaging templates

The constant you defined [here](https://github.com/ethyca/fides/pull/4986/files#diff-591f5709c6b30f40e4b4f1e97bad0be6e40bb3f4e89bca18d5715c57e7278480R69) would be perfect for this use case if you converted it to an enum

src/fides/api/api/v1/endpoints/messaging_endpoints.py
```diff
@@ -565,4 +588,190 @@ def update_basic_messaging_templates(
...
+    return get_default_template_by_type(template_type)
+
+
+@router.post(
+    MESSAGING_TEMPLATES_BY_TEMPLATE_TYPE,
+    dependencies=[Security(verify_oauth_client, scopes=[MESSAGING_TEMPLATE_UPDATE])],
+    response_model=Optional[MessagingTemplateWithPropertiesDetail],
+)
+def create_property_specific_messaging_template(
+    template_type: MessagingActionType,
```
galvana, line 640: Same comment as above

src/fides/api/schemas/messaging/messaging.py
```diff
@@ -416,41 +416,64 @@ class MessagingConfigStatusMessage(BaseModel):
...
     content: Dict[str, Any]
 
+    class Config:
+        orm_mode = True
+        use_enum_values = True
+
 
 class MessagingTemplateWithPropertiesBodyParams(BaseModel):
 
     content: Dict[str, Any]
```
galvana, line 470: I just learned this during my last feature! Pydantic's `Field` class let's your specify an `example` that is picked up nicely by Swagger
```suggestion
    content: Dict[str, Any] = Field(
        example={
            "subject": "Message subject",
            "body": "Custom message body",
        }
    )
```
image

Or we could create a new schema for the content since the keys will always be `subject` and `body`

tests/ops/api/v1/endpoints/test_messaging_endpoints.py
galvana, line 1: Very nice test coverage! 🙌

galvana (approved):
Very straightforward set of endpoints to manage all of these templates! Everything worked as expected so feel free to address my comments only if time permits. The two things that stuck out to me were:

- Having these endpoints in fides, in my opinion, since they are for plus functionality, the endpoints should be defined in fidesplus
- Using a more constrained enum instead of `MessagingActionType` for the `GET/POST /api/v1/messaging/templates/{template_type}` endpoints

src/fides/api/api/v1/endpoints/messaging_endpoints.py
eastandwestwind: Ah yes, our CRUD layer raises a `MessagingTemplateValidationException` and I've forgotten to handle this at the endpoint layer, good catch!

eastandwestwind: I think a 400 makes sense since it's technically an invalid / bad request

src/fides/api/api/v1/endpoints/messaging_endpoints.py
eastandwestwind: Here we already will raise a 400 on `MessagingTemplateValidationException` so we're covered when an invalid template_type is provided 👍

eastandwestwind:
Ticket to move endpoints to plus as a follow-up: https://ethyca.atlassian.net/browse/PROD-2211

PR #4987, opened 2024-06-14
Updating privacy request search endpoint

galvana:
Closes [PROD-2124](https://ethyca.atlassian.net/browse/PROD-2124)

### Description Of Changes

Marked the existing `GET /privacy-request` as deprecated in favor of a new `POST /privacy-request/search` endpoint. We needed to specify custom identities and custom privacy request fields but these inputs are a combination of identity/custom field name or and value. I first thought of supporting dynamic query params but that didn't work out since we need to make a distinction if the new query param is an identity or a custom field lookup.

I also made the decision to include the identity or custom field name for the search to be more specific. The current `GET /privacy-request` endpoint just matches on the identity value without considering which identity type the result matched on.

The new endpoint also removes the single string query param of `identity` in favor of an `identities` value in the POST body.
```
{
  "identities": {
    "email": "user@example.com"
  }
}
```

The `external_id` field is still supported as a standalone field, for backwards compatibility, even though we support `external_id` as a default identity type as well
```
{
  "identities": {
    "external_id": "ext-123"
  }
}
```

### Code Changes

* [ ] New `POST /privacy-request/search` endpoint and `PrivacyRequestFilter` schema to validate the payload.

### Steps to Confirm
* [x] Set `FIDES__EXECUTION__ALLOW_CUSTOM_PRIVACY_REQUEST_FIELD_COLLECTION=true` in your `.env` file to enable custom privacy request field collection
* [x] Create a new privacy request with custom identities and custom privacy request fields
```
curl -X 'POST' \
  'http://localhost:8080/api/v1/privacy-request' \
  -d '[
  {
    "identity": {
      "email": "user@example.com",
      "loyalty_id": {
        "label": "Loyalty ID",
        "value": "CH-1"
      }
    },
    "custom_privacy_request_fields": {
      "site_id": {
        "label": "Site ID",
        "value": "site-123"
      }
    }
  }
]'
```
* [x] Use the new endpoint to search by the new fields
```
curl -X 'POST' \
  'http://localhost:8080/api/v1/privacy-request/search?page=1&size=50' \
  -H 'Authorization: Bearer REDACTED' \
  -d '{
  "identities": {
    "loyalty_id": "CH-1"
  }
}'
```
```
curl -X 'POST' \
  'http://localhost:8080/api/v1/privacy-request/search?page=1&size=50' \
  -H 'Authorization: Bearer REDACTED' \
  -d '{
  "custom_privacy_request_fields": {
    "site_id": "site-123"
  }
}'
```

### Pre-Merge Checklist

* [ ] Update `CHANGELOG.md`

[PROD-2124]: https://ethyca.atlassian.net/browse/PROD-2124?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/fides/api/api/v1/endpoints/privacy_request_endpoints.py
```diff
@@ -437,8 +443,40 @@ def _filter_privacy_request_queryset(
...
+            for field_name, value in custom_privacy_request_fields.items()
+        ]
+
+        custom_fields_query = select(
+            [CustomPrivacyRequestField.privacy_request_id]
+        ).where(
+            or_(*custom_field_conditions)
+            & (CustomPrivacyRequestField.privacy_request_id.isnot(None))
+        )
+        query = query.filter(PrivacyRequest.id.in_(custom_fields_query))
```
galvana, line 477: This is the key part of this change 🔑

src/fides/api/api/v1/endpoints/privacy_request_endpoints.py
```diff
@@ -634,6 +667,128 @@ def get_request_status(
...
+@router.get(
+    PRIVACY_REQUESTS,
+    dependencies=[Security(verify_oauth_client, scopes=[PRIVACY_REQUEST_READ])],
+    response_model=Page[
+        Union[
+            PrivacyRequestVerboseResponse,
+            PrivacyRequestResponse,
+        ]
+    ],
+    deprecated=True,
```
galvana, line 679: Flagging as `deprecated` so it's reflected in Swagger
image

pattisdr:
@galvana can you make sure unit tests are running over here before I start a review?

pattisdr:
Thanks for helping with the numpy changes and getting tests passing on this branch @galvana 👍 starting review now -

src/fides/api/api/v1/endpoints/privacy_request_endpoints.py
pattisdr: Could we maybe add to the docstring that identities and and custom fields are "or" filters?

src/fides/api/api/v1/endpoints/privacy_request_endpoints.py
```diff
@@ -634,6 +667,132 @@ def get_request_status(
...
+    To see individual execution logs, use the verbose query param `?verbose=True`.
+    """
+
+    return _shared_privacy_request_search(
+        db=db,
+        params=params,
+        request_id=request_id,
+        identity=identity,
+        identities=None,
+        custom_privacy_request_fields=None,
```
pattisdr, line 723: Could you call this out, why these two fields are null, just to make it more easily parseable how you're sharing this method between the old endpoint and the new endpoint -

src/fides/api/schemas/privacy_request.py
```diff
@@ -339,3 +340,42 @@ class RequestTaskCallbackRequest(FidesSchema):
     rows_masked: Optional[int] = Field(
         default=None, description="Number of records masked, as an integer"
     )
+
+
+class PrivacyRequestFilter(FidesSchema):
```
pattisdr, line 345: Should we consider forbidding extra fields to be passed in here? they won't have an effect, given how you're explicitly defining them by name when calling `_shared_privacy_request_search`, I was just surprised that I could pass in the old `identity` value in the request body with the new endpoint (which is just ignored)

src/fides/api/schemas/privacy_request.py
```diff
@@ -339,3 +340,42 @@ class RequestTaskCallbackRequest(FidesSchema):
...
+    started_gt: Optional[datetime] = None
+    completed_lt: Optional[datetime] = None
+    completed_gt: Optional[datetime] = None
+    errored_lt: Optional[datetime] = None
+    errored_gt: Optional[datetime] = None
+    external_id: Optional[str] = None
+    action_type: Optional[ActionType] = None
+    verbose: Optional[bool] = False
+    include_identities: Optional[bool] = False
+    include_custom_privacy_request_fields: Optional[bool] = False
```
pattisdr, line 366: Some of these fields felt strange to me at first being in the request body instead of being query params, but with the introduction of these more complex searches like custom_privacy_request_fields, which probably do make sense here, I agree with the call to just keep them all together in one place -

pattisdr (approved):
Just some additional documentation requested if you agree with me there - really nice PR

src/fides/api/schemas/privacy_request.py
galvana: Sounds good, less surprises that way

src/fides/api/api/v1/endpoints/privacy_request_endpoints.py
galvana: > Both the old and new versions of the privacy request search endpoints use this shared function.
    The `identities` and `custom_privacy_request_fields` parameters are only supported in the new version
    so they are both set to None here.

PR #4990, opened 2024-06-15
Upgrade Privacy Request table to use FidesTable V2

gilluminate:
Closes [PROD-1786](https://ethyca.atlassian.net/browse/PROD-1786)

### Description Of Changes

Update the privacy request screen to utilize the new table

### Code Changes

* Replace old, hard-coded table with FidesTable v2
* move filters to a modal
* create new `MultiSelectTag` component for filtering by multiple values per column
* enable server-side sorting
* clean up header actions for consistency and better UI

### Steps to Confirm

* visit Privacy Requests page in Admin UI `/privacy-requests`
* When I click each row I am taken to the details view of the privacy request
* Filter button opens the filter modal 
* All existing filters are modeled in the modal.
* All data currently represented in the table is migrated to the new table.

### Pre-Merge Checklist

* [x] Update `CHANGELOG.md`

[PROD-1786]: https://ethyca.atlassian.net/browse/PROD-1786?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


clients/admin-ui/src/features/common/table/v2/FidesTable.tsx
```diff
@@ -74,19 +75,20 @@ const tableHeaderButtonStyles = {
   },
 };
 
+interface HeaderContentProps<T> {
```
jpople, line 78: Question for you, is there a significant difference between TS interfaces and types for component props?

jpople (approved):
Screenshot 2024-06-17 at 17 35 46

This is all looking good to me-- one small suggestion, I feel like there should be a fallback of some kind for the "Days left" and "Actions" cells when those aren't applicable?  From this I feel like it's not completely clear whether these are supposed to be empty or not.

clients/admin-ui/src/features/common/table/v2/FidesTable.tsx
gilluminate: Great question.

Yes and No. The general rule is "use `interface` until you need `type`". Not a very big difference, especially if you aren't exporting it. If exporting, you run the risk of 2 interfaces with the same name conflicting (extending) one another accidentally, where that's not possible with type.

For example:
```
interface Props {
  prop1: boolean;
}
interface Props {
  prop2: string;
} 
```
creates the equivalent of
```
interface Props {
  prop1: boolean;
  prop2: string;
}
```
if those were both using `type` then only `prop2` would apply.

I think the main reason our app is inundated with `type` is because there is no naming convention and just plain old `Props` is being used everywhere, which was likely creating the scenario where they were accidentally extending each other.

I prefer `interface` for components because a) if you are consistent with naming conventions (eg. `${componentName}Props`) and never export them unless absolutely necessary (rare) then the general rule "use `interface` until you need `type`" applies. b) this is what most component libraries are using (see Chakra's components, eg. `BoxProps`) so extending those with like definitions feels cleaner. c) nitpicky: it's more clear to use `extends` than `&` IMO.

PR #4998, opened 2024-06-18
Include AWS Tags in Resource Generation

### LINKED ISSUE #4997: Include AWS tags as Fides tags in resource generation

We use AWS tags for data governance management; we express a lot of details on compliance requirements using this system. Right now, we don't populate any details about AWS tags when calling any of the boto3 APIs.

Adding tags as a meta field. Something like `awstags`.

I considered adding these tags directly into AWS but tags are based on single text which makes it hard to adapt this directly.

tariqajyusuf:
Closes #4997

### Description Of Changes

Includes AWS tags as meta fields in resource generation.

### Code Changes

* [x] Fix typing error in storage.py
* [x] Change `get_tagging_resources` to return a full list of resources including tags. Also change downstream flows.
* [x] Add AWS tags to Fides meta

### Steps to Confirm

* [x] Start dev environment.
* [x] Attempt to scan an AWS environment with tagged resources.

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`


rbart:
yes please!

src/fides/connectors/aws.py
```diff
@@ -244,6 +244,7 @@ def create_redshift_systems(
             fides_key=cluster["ClusterIdentifier"],
             name=cluster["ClusterIdentifier"],
             description=f"Fides Generated Description for Redshift Cluster: {cluster['ClusterIdentifier']}",
+            meta={(pair["Key"], pair["Value"]) for pair in cluster["Tags"]},
```
SteveDMurphy, line 247: It looks like we may need to account and handle the scenario where `Tags` doesn't exist (You can see the [checks here](https://github.com/ethyca/fides/actions/runs/9914563221/job/27394125529?pr=4998#step:7:872) )

src/fides/connectors/aws.py
```diff
@@ -295,6 +297,7 @@ def create_rds_systems(
             fides_key=instance["DBInstanceIdentifier"],
             name=instance["DBInstanceIdentifier"],
             description=f"Fides Generated Description for RDS Instance: {instance['DBInstanceIdentifier']}",
+            meta={(pair["Key"], pair["Value"]) for pair in instance["TagList"]},
```
SteveDMurphy, line 300: Similar here, account and handle the scenario where `TagList` doesn't exist ([checks](https://github.com/ethyca/fides/actions/runs/9914563221/job/27394125529?pr=4998#step:7:918))

SteveDMurphy (changes requested):
Thanks for making this happen @tariqajyusuf as well as for the patience with the review time!! 

There are a couple of items that may need a closer look that I've commented on around handling, (`Tags` and `TagList` not existing) but I was also curious if we wanted to also look for tags that could be found on Dynamo Datasets. Would it be value-added to do those now in your eyes as well?

Outside of those changes, can you also make a Changelog entry under `unreleased`? Happy to jump in and help out as well if you'd like! Apologies I was unable to try and catch up with you synchronously this evening.

tariqajyusuf:
I added some checks and made sure that they don't break the tests. I'm still getting a strange issue with an unrelated test `tests/ctl/core/test_api.py::TestSystemUpdate::test_system_update_system_cookies` and I'm stumped as to why.

On the DynamoDB dataset comment, I think that could make sense to include that but I think that will require some propagating down since the `describe_table` call from boto3 doesn't return resource tags. That will require a little more bandwidth than I currently have right now.

Updated the changelog!

SteveDMurphy (approved):
Looks great @tariqajyusuf - thanks for making those changes and getting this added in!

PR #5004, opened 2024-06-19
Segment Integration: user_id over email

SteveDMurphy:
Closes [FIDES-746](https://ethyca.atlassian.net/browse/FIDES-746)

### Description Of Changes

It's been determined that the email cannot be used as the identity unless it is also the user_id. For this we need to infer the user_id from either traits or external_ids

### Code Changes

* [ ] Update the data_protection_Request to use the user_id instead of the identity email

### Steps to Confirm

* [ ] Validate the erasure completes with our test data

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### DIFF
--- a/data/saas/config/segment_config.yml
+++ b/data/saas/config/segment_config.yml
@@ -166,7 +166,10 @@ saas_config:
       - name: workspace_name
         connector_param: workspace
       - name: user_id
-        identity: email
+        references:
+          - dataset: <instance_fides_key>
+            field: traits.user_id
+            direction: from
     body: |
       {
         "regulation_type": "Suppress_With_Delete",
--- a/data/saas/dataset/segment_dataset.yml
+++ b/data/saas/dataset/segment_dataset.yml
@@ -7,8 +7,6 @@ dataset:
         fields:
           - name: segment_id
             data_categories: [user.unique_id]
-            fidesops_meta:
-              primary_key: True
       - name: track_events
         fields:
           - name: external_ids
@@ -25,7 +23,7 @@ dataset:
             fields:
               - name: name
                 data_categories: [user.name]
-                fidesops_meta:
+                fides_meta:
                   data_type: string
               - name: version
                 data_categories: [system.operations]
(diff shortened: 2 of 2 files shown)


data/saas/dataset/segment_dataset.yml
```diff
@@ -81,41 +79,53 @@ dataset:
...
             data_categories: [user.contact.email]
-            fidesops_meta:
+            fides_meta:
               data_type: string
           - name: firstName
             data_categories: [user.name]
-            fidesops_meta:
+            fides_meta:
+              data_type: string
+          - name: first_name
```
SteveDMurphy, line 88: additional fields found in testing (these may be configurable by client?)

data/saas/config/segment_config.yml
```diff
@@ -166,7 +166,10 @@ saas_config:
       - name: workspace_name
         connector_param: workspace
       - name: user_id
-        identity: email
+        references:
+          - dataset: <instance_fides_key>
+            field: traits.user_id
```
SteveDMurphy, line 171: I chose to use the traits, but it could also be that we use `external_ids` however it will end up dropping extra requests so decided against it for the time being

data/saas/dataset/segment_dataset.yml
```diff
@@ -7,8 +7,6 @@ dataset:
         fields:
           - name: segment_id
             data_categories: [user.unique_id]
-            fidesops_meta:
-              primary_key: True
```
galvana, line 11: Did we test that erasure was called after removing `primary_key` from this field? I don't think the `data_protection_request` will work if none of the fields are flagged as `primary_key`

SteveDMurphy: For some reason, it needed to be moved to the `traits` collection from what I could tell, so did it here:
https://github.com/ethyca/fides/pull/5004/files#diff-eb9b94a980a356ef500f8905ac01f617398e36f2e0123fe92d7b441b4a6894ceR98-R99

data/saas/dataset/segment_dataset.yml
```diff
@@ -81,41 +79,55 @@ dataset:
...
               data_type: string
           - name: gender
             data_categories: [user.demographic.gender]
-            fidesops_meta:
+            fides_meta:
               data_type: string
           - name: id
             data_categories: [user.unique_id]
+            fides_meta:
+              primary_key: true
```
SteveDMurphy, line 99: This is the spot @galvana - happy to try and move it elsewhere if a different space makes sense? I think I was also looking at having this be where the erase happened instead of in the data protection request

galvana: Ah, I completely missed it, this makes sense though 👍

galvana (approved):

PR #5019, opened 2024-06-24
Update pymongo to 4.7.3

erosselli:
Closes #<issue>

### Description Of Changes

### Code Changes

### Steps to Confirm

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`


src/fides/api/service/connectors/mongodb_connector.py
```diff
@@ -71,7 +71,7 @@ def test_connection(self) -> Optional[ConnectionTestStatus]:
             )
             if default_auth_db:
                 db = client[default_auth_db]
-                db.collection_names()
+                db.list_collection_names()
```
erosselli, line 74: As per the [pymongo 4.0 migration docs](https://pymongo.readthedocs.io/en/stable/migrate-to-pymongo4.html#database-collection-names-is-removed), `collection_names` is removed and we need to use `list_collection_names` instead.

tests/ops/integration_tests/setup_scripts/postgres_setup.py
```diff
@@ -12,6 +12,10 @@
     ConnectionConfig,
     ConnectionType,
 )
+
+# Need to manually import this model because it's used in src/fides/api/models/property.py
+# but that file only imports it conditionally if TYPE_CHECKING is true
+from fides.api.models.privacy_experience import PrivacyExperienceConfig
```
erosselli, line 18: this change was needed so the `postgres_setup` script would run correctly, but is technically unrelated to the pymongo upgrade

tests/ops/integration_tests/test_integration_mongodb_example.py
```diff
@@ -18,9 +18,9 @@ def test_mongo_example_data(integration_mongodb_connector):
...
     }.difference(collection_names) == set()
 
-    assert db.customer.count() == 3
-    assert db.payment_card.count() == 2
-    assert db.orders.count() == 4
-    assert db.employee.count() == 2
-    assert db.product.count() == 3
-    assert db.reports.count() == 4
+    assert db.customer.count_documents({}) == 3
+    assert db.payment_card.count_documents({}) == 2
```
erosselli, line 22: As per the [pymongo migration docs](https://pymongo.readthedocs.io/en/stable/migrate-to-pymongo4.html#collection-count-and-cursor-count-is-removed), `count` was removed and we need to use `count_documents` or `estimated_document_count` instead.

src/fides/api/service/connectors/mongodb_connector.py
```diff
@@ -44,7 +44,7 @@ def create_client(self) -> MongoClient:
         """Returns a client for a MongoDB instance"""
         uri = (self.configuration.secrets or {}).get("url") or self.build_uri()
         try:
-            return MongoClient(uri, serverSelectionTimeoutMS=5000)
+            return MongoClient(uri, serverSelectionTimeoutMS=5000, uuidRepresentation='pythonLegacy')
```
erosselli, line 47: In pymongo version 4, [default UUID representation changes](https://pymongo.readthedocs.io/en/stable/migrate-to-pymongo4.html#encoding-a-uuid-raises-an-error-by-default) from 'pythonLegacy' to 'unspecified', which raises an error when encoding UUIDs.

I've added the 'pythonLegacy' string to keep the existing behavior, I'd love an extra pair of eyes on the [relevant mongo docs ](https://pymongo.readthedocs.io/en/stable/examples/uuid.html#supported-uuid-representations) so that we can determine if this is the best solution. TBH I didn't understand all of the doc. 

Another things is that according to the docs, we should be able to pass in a constant to `uuidRepresentation`, like
```py
from bson.binary import UuidRepresentation
client = MongoClient(uuidRepresentation=UuidRepresentation.PYTHON_LEGACY)
```
however when I tried it, it turns out the field expects a string but `UuidRepresentation.PYTHON_LEGACY` is a number so I'm not sure what I should do there 🤔

pattisdr: Thanks for flagging! What's tricky here is that we don't know which uuid subtypes our customers are using.  Different customers could use different uuid representations. We could make this configurable, but since we don't officially support uuid types yet, I don't think this is worth pursuing at this time unless it is a customer request.  Instead, I think we should pick one for now, as you've done here, but I'd lean towards "standard" instead, as I think this would be more common.  I'd also like to see test coverage against a UUID type since we don't have this data type in our sample mongo data.  

So next steps, something like:

1) Update to `UuidRepresentation.standard` here
2) Add a `customer_uuid` field or similar as a standard UUID type to `customer_details` records in `mongo_sample.js`, to automatically add this data on startup in our example database, for example:
```js
db.customer_details.insert([
    {
        "customer_id": 1,
        "customer_uuid": new UUID('3b241101-e2bb-4255-8caf-4136c566a962'),
        "gender": "male"
        ...
   }
])
```
3) Update `data/dataset/mongo_example_test_dataset.yml` to add this `customer_uuid` field to the customer_details table, and also give it a "user." data type so Fides will know to target it in DSR's:
```yaml
  - name: customer_details
    fields:
    ...
    - name: customer_uuid
      data_categories:
        - user.unique_id
```
4) Update one of our end-to-end tests for mongo access requests to verify this customer uuid is serialized properly.   I think `tests/ops/integration_tests/test_mongo_task.py > test_object_querying_mongo` would be a good one because it uses real datasets.  `target_categories` would need to be updated to also look for `user.unique_id`, and then you could assert the value of `customer_uuid` is what you expect.
5) Also complete end-to-end DSR testing similar to your "Steps to confirm" above, running a DSR on a table with a UUID

pattisdr (commented):
looking good @erosselli just a suggestion on next steps for uuid representation -

pattisdr (approved):
great @erosselli just a quick consolidation of the changelog. thanks for taking the time to add test coverage on a UUID type.

PR #5020, opened 2024-06-25
Fides Pydantic V2 Upgrade

pattisdr:
Closes #PROD-2263

❗ Dependent on https://github.com/ethyca/fideslang/pull/11
ℹ️ Replaces https://github.com/ethyca/fides/pull/4442

### Description Of Changes

Pydantic V2/Fast API Upgrade

### Code Changes

- Removes support for Python 3.8-
-  Upgrades Pydantic to V2, upgrades FastAPI, and other supporting dependencies
- class Config -> `model_config=ConfigDict(...)`
- `dict()` -> `model_dump()`, often with mode="json" since model_dump() is not always serializable as json, depending on model contents
- `parse_obj()` -> `model_validate()`
- `from_orm()` -> `model_validate()`
- conlist -> Annotated[List[], Field(max_length=50)]
- `root_validators` -> `model_validators`. mode="after" is often the better way to go because before validators can be called in surprising places, where sometimes the incoming data is a dict, or other times, the schema itself
- `__init_subclass__` -> `__pydantic_init_subclass__`
- `__fields__` -> `model_fields()`
- `customise_sources` -> `customise_setting_sources`
- `update_forward_refs` -> `model_rebuild`
-  Always=True -> `validate_default=True` on the field directly
- `ConstrainedStr` now need to be defined using StringConstraints
- Sometimes our str(errors) needed to be wrapped with jsonable_encoder.  Errors are not always json serializable
- Lots of default values of `None` added to keep Optional fields optional
- `@app.on_event("startup")` is deprecated, using [Lifespan](https://fastapi.tiangolo.com/advanced/events/) instead
- Updated definitions of custom types: SafeStr, HtmlStr, PhoneNumber, GPPMechanismConsentValue, URLOrigin, and CssStr
- Pydantic V2 can't serialize unknown types. Adding FieldSerializers on some of our Python classes that we're using as types for DSR purposes since Pydantic V2 can't serialize them otherwise.
- Using duck typing to serialize deeping nested collections to match original behavior for DSR processing.  A customer's collection might not be fully defined with a Pydantic schema which was causing some of the collection representation to get dropped in V2.
- Updates to `NoValidationSchema` so it can still be used as a Pydantic schema that skips validation allowing a schema to be used for docs-only
- You can't just stash random things on schemas, use `json_schema_extra`
- We return a lot of Pydantic error messages in failed API calls, and the text has changed
- Models are no longer equal to the dicts containing their data, resulting in a lot of test changes
- Middleware can no longer be updated after the application has started - hack to get this working since we allow cors origins to be updated after the fact
- Exposes admin-ui settings in the Config API 

### Steps to Confirm

This is something that really needs a full regression.  These are features that needed more attention:

* [x] Checking adding multiple types of integrations, postgres, bigquery, attentive, mailchimp
* [x] Verify connection secret schemas show up in docs
* [ ] Running DSR's with Postgres and Mongo and verifying deeply nested results
* [x] Verifying saving messaging secrets
* [x] Verifying saving storage secrets
* [x] General verification that secrets aren't leaked in error messages/logs.  Pydantic error messages are more verbose.
* [x] CORS origins, updating CORS origins after the application is running
* [x] Setting env vars in the toml and as env vars
* [x] CLI commands

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`


ThomasLaPiana:
@pattisdr You're a rockstar! :laughing:

pattisdr:
Your work has been so helpful @ThomasLaPiana especially seeing how you resolved some of the more subjective pieces. More to do but making progress!

ThomasLaPiana:
> Your work has been so helpful @ThomasLaPiana especially seeing how you resolved some of the more subjective pieces. More to do but making progress!

I'm working on moving one of our services from Python 3.7 to 3.11 and including the Pydantic upgrade...I will also be referring to this work! :smile:

tests/ops/models/test_application_config.py
```diff
@@ -221,7 +221,7 @@ def test_update_config_set(self, db):
         ApplicationConfig.update_config_set(db, CONFIG)
         # check a few specific config properties of different types on the database record
         db_config = ApplicationConfig.get_config_set(db)
-        assert dumps(db_config) == CONFIG.json()
+        assert dumps(db_config, separators=(",", ":")) == CONFIG.model_dump_json()
```
pattisdr, line 224: `model_dump_json` stores in a more condensed format, to save space, so I'm adjusting the test

src/fides/api/task/create_request_tasks.py
```diff
@@ -188,7 +188,7 @@ def base_task_data(
         # when executing the node, so we don't have to recalculate incoming
         # and outgoing edges.
         collection_representation = json.loads(
-            dataset_graph.nodes[node].collection.json()
+            dataset_graph.nodes[node].collection.model_dump_json(serialize_as_any=True)
```
pattisdr, line 191: Another important place for DSR 3.0 - to save the collection on the RequestTask with duck typing so all the nested sub fields are persisted

src/fides/api/custom_types.py
```diff
@@ -70,89 +66,71 @@ class HtmlStr(str):
...
+    return str(value).rstrip("/")
 
-    @classmethod
-    def validate(cls, value: Any, field: ModelField, config: BaseConfig) -> AnyUrl:
-        value = super().validate(value, field, config)
-        if value.path:
-            raise ValueError("URL origin values cannot contain a path.")
 
-        return value
+URLOrigin = Annotated[AnyUrl, AfterValidator(validate_path_of_url)]
```
pattisdr, line 127: AnyURL's add a trailing slash to everything now, and are used for cors_origins, we can't have trailing slashes there.  Updated the validator to strip it off.

pattisdr:
This gets out of date quickly, will reconcile within the next couple of days, focused on Plus at the moment -

src/fides/config/utils.py
```diff
@@ -66,5 +66,5 @@ def get_dev_mode() -> bool:
         "active_default_storage_type",
     ],
     "consent": ["override_vendor_purposes"],
-    "admin-ui": ["enabled", "url"],
+    "admin_ui": ["enabled", "url"],
```
pattisdr, line 69: @galvana I think we meant to expose this here, but double checking that it is safe to show

@daveqnet this is adding admin ui settings to the `CONFIG_KEY_ALLOWLIST`

galvana: Yes! This is safe, it's just the Admin UI URL

src/fides/api/schemas/connection_configuration/connection_secrets_bigquery.py
```diff
@@ -15,14 +15,16 @@ class KeyfileCreds(BaseModel):
...
-    private_key: Optional[str] = Field(None, sensitive=True)
+    private_key_id: Optional[str] = Field(default=None, title="Private Key ID")
+    private_key: Optional[str] = Field(
+        default=None, json_schema_extra={"sensitive": True}
+    )
     client_email: Optional[EmailStr] = None
-    client_id: Optional[str] = Field(None, title="Client ID")
-    auth_uri: Optional[str] = Field(None, title="Auth URI")
-    token_uri: Optional[str] = Field(None, title="Token URI")
+    client_id: Optional[str] = Field(default=None, title="Client ID")
```
galvana, line 23: Why do we need to add the `default` keyword here? Is this something we should change for all usages of `Field`?

pattisdr: we don't have to add it, it's not required but doesn't hurt -

src/fides/config/utils.py
pattisdr: thanks for confirming 👍

src/fides/api/schemas/privacy_request.py
```diff
@@ -115,53 +111,39 @@ class FieldsAffectedResponse(FidesSchema):
...
+    def serialize_status(self, status: ExecutionLogStatus) -> str:
+        """For statuses, we want to use the name instead of the value
+        This is for backwards compatibility where we are repurposing the "paused" status
+        to read "awaiting processing"
         """
-        data = super().dict(*args, **kwargs)
-        if isinstance(data.get("status"), ExecutionLogStatus):
-            data["status"] = data["status"].name
-        return data
+        return status.name
```
galvana, line 129: So clean! ✨

src/fides/connectors/aws.py
```diff
@@ -244,7 +244,7 @@ def create_redshift_systems(
             fides_key=cluster["ClusterIdentifier"],
             name=cluster["ClusterIdentifier"],
             description=f"Fides Generated Description for Redshift Cluster: {cluster['ClusterIdentifier']}",
-            meta={(pair["Key"], pair["Value"]) for pair in cluster.get("Tags", {})},
+            meta={pair["Key"]: pair["Value"] for pair in cluster.get("Tags", {})},
```
galvana, line 247: Was this just something we missed in the original implementation? Returning a set of tuples instead of a dictionary?

src/fides/api/db/samples.py
```diff
@@ -122,7 +120,7 @@ def load_sample_connections_from_project() -> List[SampleConnection]:
     return valid_connections
 
 
-def load_sample_yaml_file(file: TextIO, expand_vars: bool = True) -> Dict:
+def load_sample_yaml_file(file: IO, expand_vars: bool = True) -> Dict:
```
galvana, line 123: What was the reason for this change?

src/fides/api/schemas/connection_configuration/connection_secrets_bigquery.py
galvana: Got it! Thanks

src/fides/connectors/aws.py
pattisdr: This was something I updated after a merge with main yesterday, it's brand new, I think it might have been getting coerced with pydantic v1, and not coerced in Pydantic v2.

src/fides/api/db/samples.py
pattisdr: mypy I believe -

pattisdr: upgrading mypy flagged hundreds of places -

requirements.txt
```diff
@@ -4,6 +4,7 @@ anyio==3.7.1
 APScheduler==3.9.1.post1
 asyncpg==0.27.0
 boto3==1.26.1
+bump-pydantic==0.8.0
```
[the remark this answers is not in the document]
pattisdr: `bump-pydantic` has been removed

galvana (approved):
Thanks for tracking the serialization warnings! I'm going to approve to unblock this PR but let me know if there is any other testing I should do before we finally merge.

clients/admin-ui/src/types/api/models/ClassifyParams.ts
```diff
@@ -8,7 +8,7 @@
  * and can rely on the defaults
  */
 export type ClassifyParams = {
-  possible_targets?: Array<any>;
+  possible_targets?: null;
```
eastandwestwind, line 11: why does this return only null?

eastandwestwind: Hmm, ok, I'll update back to using `Array<any>` for now.

clients/admin-ui/src/types/api/models/ConnectionConfigurationResponse.ts
```diff
@@ -11,18 +11,18 @@ import type { SaaSConfigBase } from "./SaaSConfigBase";
...
-  saas_config?: SaaSConfigBase;
-  secrets?: any;
-  authorized?: boolean;
-  enabled_actions?: Array<ActionType>;
+  updated_at?: string | null;
+  disabled?: boolean | null;
+  last_test_timestamp?: string | null;
+  last_test_succeeded?: boolean | null;
+  saas_config?: SaaSConfigBase | null;
+  secrets?: null;
```
eastandwestwind, line 25: here's another instance of pydantic using `null`. I'm changing this back to `any` for now

clients/admin-ui/src/types/api/models/ClassifyParams.ts
pattisdr: @eastandwestwind did I accidentally override this? Should I put back to Array<any>?

eastandwestwind: If the tests pass, then this is fine as `null`. I don't want to run into issues in the future where every time we autogenerate types, we have to do manual changes like this.

I think the real question is how we get pydantic to better transform types like `Optional[Iterable]` from the BE.

clients/admin-ui/src/types/api/models/ConnectionConfigurationResponse.ts
eastandwestwind: keep as `null` if tests pass

clients/admin-ui/src/types/api/models/ClassifyParams.ts
pattisdr: Thank you 🙏

clients/admin-ui/src/types/api/models/ConnectionConfigurationResponse.ts
pattisdr: got it 👍

ThomasLaPiana:
@pattisdr bravo!

PR #5022, opened 2024-06-25
Replaces typescript-cookie with js-cookie

eastandwestwind:
Closes https://ethyca.atlassian.net/browse/PROD-2192

### Description Of Changes

Replaces typescript-cookie with js-cookie. Preserves custom CODEC for encoding / decoding cookie names / values.

### Code Changes

* [ ] Replaces typescript-cookie with js-cookie
* [ ] Update tests

### Steps to Confirm

* [ ] Please regression test TCF and non-TCF, saving preferences through various ways. This change basically touches everywhere we read / write / remove the cookie. Test 2 key things as follows:
* [ ] Test that preferences are preserved across page loads in both the cookie itself and that it's reflected in the UI. 
* [ ] Test that cookies are deleted from the browser when a specific notice has cookies linked to it in the Admin-UI, and the user opts out of that notice.

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`


clients/fides-js/__tests__/lib/cookie.test.ts
```diff
@@ -60,17 +57,17 @@ const mockRemoveCookie = jest.fn(
...
-    value: string,
-    attributes: object,
-    encoding: object
-  ) => mockSetCookie(name, value, attributes, encoding),
-  removeCookie: (name: string, attributes?: CookieAttributes) =>
-    mockRemoveCookie(name, attributes),
-}));
+
+// @ts-ignore
+jest.spyOn(Cookies, "withConvertor").mockReturnValue(
```
eastandwestwind, line 62: Here is where I'm struggling- trying to mock the "withConvertor" method which, in the actual lib, returns an instance of `CookiesStatic`, providing all of our `get`, `set`, etc methods.

clients/fides-js/src/lib/cookie.ts
```diff
@@ -28,18 +28,21 @@ export const CONSENT_COOKIE_NAME = "fides_consent";
...
+ * See: https://www.npmjs.com/package/js-cookie#converters
  */
-const CODEC: Types.CookieCodecConfig<string, string> = {
-  decodeName: decodeURIComponent,
-  decodeValue: decodeURIComponent,
-  encodeName: encodeURIComponent,
-  encodeValue: encodeURIComponent,
-};
+
+const cookies: CookiesStatic = Cookies.withConverter({
```
eastandwestwind, line 38: This `withConvertor` method allows us to define a custom CODEC for encoding/decoding the name/val of cookies.

clients/fides-js/__tests__/lib/cookie.test.ts
eastandwestwind: I've tried mocking the lib directly, too, e.g. 
```
jest.mock('js-cookie', () => ({
  __esModule: true,
  withConvertor: jest.fn(() => ({
    get: () => mockGetCookie(),
    set: (name: string, value: string, attributes: object) =>
        mockSetCookie(name, value, attributes),
    remove: (name: string, attributes?: CookieAttributes) =>
        mockRemoveCookie(name, attributes),
  }))
}));
```

clients/fides-js/src/lib/cookie.ts
NevilleS: Since you're doing this statically when the `cookie.ts` file is loaded, this might be why your mock isn't working. I have a feeling this code runs against the real (non-mocked) version, and then afterwards your code registers the spy.

You could try adding some logging or breakpoints to check for that.

clients/fides-js/__tests__/lib/cookie.test.ts
eastandwestwind: ahh turns out I just had to remove that `__esModule: true,` bit, thanks so much @gilluminate for solving this!

gilluminate: so very close!

clients/fides-js/src/lib/cookie.ts
eastandwestwind: Yes this was the tricky bit, but @gilluminate was able to figure out the correct way to mock this!

clients/fides-js/src/lib/cookie.ts
```diff
@@ -340,9 +338,9 @@ export const makeConsentDefaultsLegacy = (
 /**
  * Given a list of cookies, deletes them from the browser
  */
-export const removeCookiesFromBrowser = (cookies: Cookies[]) => {
-  cookies.forEach((cookie) => {
-    removeCookie(cookie.name, {
+export const removeCookiesFromBrowser = (cookiesToRemove: CookiesType[]) => {
```
gilluminate, line 341: love the clarity added by renaming this

clients/fidesui/src/index.ts
```diff
@@ -53,7 +53,7 @@ export * from "@chakra-ui/visually-hidden";
  */
 /* eslint-disable import/export */
 export * from "./icons";
-export { AddIcon, QuestionIcon, WarningIcon, LinkIcon } from "./icons";
+export { AddIcon, LinkIcon,QuestionIcon, WarningIcon } from "./icons";
```
gilluminate, line 56: as nit picky as it gets, but looks like this could use an extra space after the `,`

gilluminate (approved):

clients/fidesui/src/index.ts
eastandwestwind: ah nice!

PR #5005, opened 2024-06-19
PROD-2039 Messaging / Property Specific Mail Configurations

lucanovera:
### Description Of Changes

- Support having custom email templates for each property. 

### Code Changes
* [ ] Nav: Rename managment to settings
* [ ] Nav: Move Properties under settings
* [ ] Add new messaging option under settings
* [ ] Add messaging page. A table that list messaging templates with toggle to enable/disable
* [ ] Add new message, select message type modal
* [ ] Add new message page
* [ ] Edit message page

### Steps to Confirm

* [ ] Go to Settings > Messaging
* [ ] Edit the messages for the Default property
* [ ] Submit DSR request through privacy center
* [ ] The email received should match the customized messages

With a custom property:
* [ ] Create a new property
* [ ] Use custom privacy center paths for properties feature (https://github.com/ethyca/fides/pull/4907)
* [ ] Go to Settings > Messaging
* [ ] Add the messages for the custom property you've created
* [ ] Navigate to your custom path for privacy center (eg. /mycustomproperty)
* [ ] Submit a DSR request
* [ ] The email received should match the customized messages for that property

### Screenshots
Messaging Page:
Captura de pantalla 2024-06-22 a la(s) 10 39 20 a  m

Add new message / select type modal:
Captura de pantalla 2024-06-22 a la(s) 10 39 41 a  m

Add new / edit message page form:
Captura de pantalla 2024-06-22 a la(s) 10 39 58 a  m

Error messages:
Captura de pantalla 2024-06-22 a la(s) 10 41 00 a  m

New navigation:
Captura de pantalla 2024-06-22 a la(s) 10 39 10 a  m

### Pre-Merge Checklist

* [x] All CI Pipelines Succeeded
* [x] Issue Requirements are Met
* [x] Update `CHANGELOG.md`

PR #5027, opened 2024-06-27
PROD-2053-Update-navigation-Management-and-Properties

lucanovera:
### Description Of Changes

- Rename "/management" routes to "/settings" to match the new nav name.

### Code Changes

* [ ] Update routes
* [ ] Move files from pages/management to pages/settings using git mv to keep history
* [ ] Update hardcoded routes in cypress tests
* [ ] No need for changelog update, the changelog was already updated in #5005

### Steps to Confirm

* [ ] Log in admin-ui
* [ ] Navigate all options under Settings
* [ ] The routes should say "settings/" instead of "/management" (some routes don't say either)
* [ ] All pages under Setting should load correctly

### Pre-Merge Checklist


clients/admin-ui/cypress/e2e/organization.cy.ts
```diff
@@ -67,7 +67,7 @@ describe("Organization page", () => {
   describe("can edit organization configuration", () => {
     beforeEach(() => {
       stubOrganizationCrud();
-      cy.visit("/management/organization");
+      cy.visit("/settings/organization");
```
jpople, line 70: While we're in here, can we replace these hardcoded strings with references to the appropriate constants from `routes.ts` instead?

jpople (approved):
Looks good!  Confirmed all pages load and have the correct URLs.

clients/admin-ui/cypress/e2e/organization.cy.ts
lucanovera: Sure, let me change those

PR #5037, opened 2024-06-28
PROD-2254: User must have delete scope to delete systems

eastandwestwind:
Closes https://ethyca.atlassian.net/browse/PROD-2254

### Description Of Changes

Remove "OR" condition in the security of Delete Systems endpoint such that we only check for users with the "SYSTEM_DELETE" scope explicitly.

### Code Changes

* [ ] Restrict Delete systems API endpoint such that user must have "SYSTEM_DELETE" scope

### Steps to Confirm

* [ ] Log into Admin-UI with viewer scope
* [ ] Confirm that you cannot delete systems

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### DIFF
--- a/src/fides/api/api/v1/endpoints/system.py
+++ b/src/fides/api/api/v1/endpoints/system.py
@@ -37,7 +37,6 @@ from fides.api.oauth.system_manager_oauth_util import (
 )
 from fides.api.oauth.utils import (
     get_current_user,
-    verify_oauth_client,
     verify_oauth_client_prod,
 )
 from fides.api.schemas.connection_configuration import connection_secrets_schemas
--- a/src/fides/api/api/v1/endpoints/system.py
+++ b/src/fides/api/api/v1/endpoints/system.py
@@ -33,7 +33,6 @@ from fides.api.models.fides_user import FidesUser
 from fides.api.models.sql_models import System  # type:ignore[attr-defined]
 from fides.api.oauth.system_manager_oauth_util import (
     verify_oauth_client_for_system_from_fides_key,
-    verify_oauth_client_for_system_from_fides_key_cli,
     verify_oauth_client_for_system_from_request_body_cli,
 )
 from fides.api.oauth.utils import get_current_user, verify_oauth_client_prod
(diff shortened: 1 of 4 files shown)


src/fides/api/api/v1/endpoints/system.py
```diff
@@ -312,7 +312,6 @@ async def upsert(
 )
 async def delete(
     fides_key: str = Security(
-        verify_oauth_client_for_system_from_fides_key_cli,
```
pattisdr, line 315: great, I think you found the right spot! Would like to see a test verifying a data steward cannot delete a system

pattisdr: You will probably have to swap this though for a different function, probably `verify_oauth_client`.

eastandwestwind: Absolutely. This is what I was mentioning in standup where if we want to create a release branch in an hr, I don't have time as I have an interview now. I'll create a separate ticket though

eastandwestwind: https://ethyca.atlassian.net/browse/PROD-2265

pattisdr: This comment needs to be addressed: https://github.com/ethyca/fides/pull/5037#discussion_r1658907279, it needs to be swapped for the right function, not just removed

pattisdr: Oh right, and there are more adjustments needed because verify_oauth_client_for_system_from_fides_key_cli returns the fides key but verify_oauth_client does not. I'll look into it.

tests/ctl/core/test_api.py
```diff
@@ -2542,9 +2542,7 @@ def test_system_delete_as_system_manager(
             resource_id=system.fides_key,
             headers=auth_header,
         )
-        assert result.status_code == HTTP_200_OK
-        assert result.json()["message"] == "resource deleted"
-        assert result.json()["resource"]["fides_key"] == system.fides_key
+        assert result.status_code == HTTP_403_FORBIDDEN
```
pattisdr, line 2545: We have to tests when we're making permissions updates, but luckily this already existed!

eastandwestwind: Ah, I missed this, thanks for updating!

src/fides/api/api/v1/endpoints/system.py
```diff
@@ -293,6 +297,12 @@ async def upsert(
 
 @SYSTEM_ROUTER.delete(
     "/{fides_key}",
+    dependencies=[
```
eastandwestwind, line 299: 💯

pattisdr (approved):

PR #5040, opened 2024-06-28
fix various Cypress race conditions in Privacy Center

gilluminate:
### Description Of Changes

Running Cypress locally has caused some level of developer frustration. Some tests require messing with environment variables to get them to behave like the CI tests, others fail randomly due to race conditions. These fixes help make everything more consistent, and thereby help give more confidence in the tests as we work.

### Code Changes

* Move the definition of the `patchNoticesServed` intercept for `consent-i18n.cy.ts` in to the `beforeEach` section to make sure it's ready prior to the test running at all. This was one that was randomly failing.
* Force `isOverlayEnabled` to be false on a particular consent test that was previously requiring developers to set to `false` manually before running. No longer need to update `.env` to run tests!
* Force `isOverlayEnabled` to be true on a test that would then fail when trying to update `.env` to make the previous test work. This one would also randomly fail without any `.env` adjustments due to the runner picking up env variable *after* the override sometimes.
* Address some *timezone* issues that recently crept in.

### Steps to Confirm

* Privacy Center and Admin UI Cypress tests should continue to pass in the CI environment.
* Running the Cypress tests locally should pass consistently (run it a couple times to be sure)

### Pre-Merge Checklist

* [x] Update `CHANGELOG.md`

### DIFF
--- a/clients/privacy-center/cypress/e2e/consent-i18n.cy.ts
+++ b/clients/privacy-center/cypress/e2e/consent-i18n.cy.ts
@@ -1632,6 +1632,13 @@ describe("Consent i18n", () => {
       beforeEach(() => {
         beforeAll();
 
+        // Consent reporting intercept
+        cy.intercept(
+          "PATCH",
+          `${API_URL}/consent-request/consent-request-id/notices-served`,
+          { fixture: "consent/notices_served.json" }
+        ).as("patchNoticesServed");
+
         cy.visitWithLanguage("/", SPANISH_LOCALE);
         cy.wait("@getVerificationConfig");
         cy.overrideSettings(SETTINGS);
@@ -1651,13 +1658,6 @@ describe("Consent i18n", () => {
       });
 
       it("calls notices served with the correct history id for the notices", () => {
-        // Consent reporting intercept
-        cy.intercept(
-          "PATCH",
-          `${API_URL}/consent-request/consent-request-id/notices-served`,
-          { fixture: "consent/notices_served.json" }
-        ).as("patchNoticesServed");
-
         cy.wait("@patchNoticesServed").then((interception) => {
           expect(interception.request.body.privacy_notice_history_ids).to.eql(
             EXPECTED_NOTICE_HISTORY_IDS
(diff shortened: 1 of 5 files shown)


clients/privacy-center/cypress/e2e/home.cy.ts
```diff
@@ -35,26 +35,35 @@ describe("Home", () => {
...
-    }).as("getExperience");
-    cy.visit("/");
-    cy.overrideSettings(settings);
+  it(
+    "should show an empty state when notice-driven but there are no notices",
+    {
+      env: {
+        // race condition sometimes picks up env variable *after* the override, so let's be double sure
+        FIDES_PRIVACY_CENTER__IS_OVERLAY_ENABLED: true,
+      },
```
NevilleS, line 44: What does this do? I can't really find any references for the `it()` function...

clients/privacy-center/cypress/e2e/consent.cy.ts
```diff
@@ -403,24 +396,28 @@ describe("Consent settings", () => {
...
-      cy.visit("/fides-js-demo.html");
+      cy.visit("/fides-js-demo.html?initialize=false");
       cy.get("#consent-json");
-      // eslint-disable-next-line cypress/no-unnecessary-waiting
-      cy.wait(2000);
-      cy.waitUntilFidesInitialized().then(() => {
-        cy.window({ timeout: 500 }).should("have.property", "dataLayer");
-        cy.window().then((win) => {
+      cy.window();
+      cy.window().then((win) => {
```
NevilleS, line 404: Is the double `.window()` intentional to make this less flakey?

clients/privacy-center/cypress/e2e/consent-i18n.cy.ts
```diff
@@ -1632,6 +1632,13 @@ describe("Consent i18n", () => {
       beforeEach(() => {
         beforeAll();
 
+        // Consent reporting intercept
+        cy.intercept(
+          "PATCH",
+          `${API_URL}/consent-request/consent-request-id/notices-served`,
+          { fixture: "consent/notices_served.json" }
+        ).as("patchNoticesServed");
```
NevilleS, line 1640: 👍

NevilleS (approved):
Overall looks good - just some clarifying questions

clients/privacy-center/cypress/e2e/consent.cy.ts
gilluminate: I think this might be a copy/pasta issue. I'll double check 😅

clients/privacy-center/cypress/e2e/home.cy.ts
gilluminate: if `it` has 3 parameters, the 2nd one is for options. `env` is one of the options allowed for setting an environment variable for this test only.
see https://docs.cypress.io/guides/guides/environment-variables#Option-5-Test-Configuration

clients/privacy-center/cypress/e2e/consent.cy.ts
gilluminate: removed! Good catch.

PR #5048, opened 2024-07-01
PROD-2245 Use pip-compile to freeze requirements

andres-torres-marroquin:
Closes [PROD-2245]

### Description Of Changes

Use pip-compile to freeze requirements.

### Code Changes

* [ ] Renamed each of requirements files to `*.in` files: `requirements.txt`, `dev-requirements.txt` and `optional-requirments.txt`.
* [ ] Now when updating requirements, the `*.in` files are the ones to be updated, and after that a new nox command has to be ran: `nox -s upgrade_packages`.

### Steps to Confirm

* [ ] Add a requirement to any of the `*.in` files.
* [ ] Run `nox -s upgrade_packages`.
* [ ] Verify corresponding requirement.txt file with new requirements.

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

[PROD-2245]: https://ethyca.atlassian.net/browse/PROD-2245?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


adamsachs:
@andres-torres-marroquin just taking a quick look at this, it looks promising - but it looks like we're failing to build our docker image when `RUN pip install --no-cache-dir install -r requirements.txt` is run from the `Dockerfile`. are you aware of why that's failing and do you have a path forward on how to fix that step?

pattisdr:
this is neat; I'd also like to ask if we can hold off on merging this until the pydantic v2 work is in to not slow down that effort -

adamsachs:
> this is neat; I'd also like to ask if we can hold off on merging this until the pydantic v2 work is in to not slow down that effort -

makes sense, thanks for calling out @pattisdr - i'll throw a `don't merge` tag on here and @andres-torres-marroquin i'll also just move this into draft so we avoid an accidental merge before @pattisdr is ready with the pydantic upgrade. the pydantic upgrade is time-sensitive (need to be done by july 30) and also a large-scale update, so let's make sure that gets across the line and then we'll follow up with this change 👍

pattisdr:
@andres-torres-marroquin reached out to me directly and said the steps to reconcile would be:

```
1. copy all the requirements.txt to [requirements.in](http://requirements.in/) 
2. run nox -s upgrade_packages
3. and it should regenerate requirements.txt  from [requirements.in](http://requirements.in/)
```
so that shouldn't be too bad

CHANGELOG.md
```diff
@@ -17,6 +17,9 @@ The types of changes are:
 
 ## [Unreleased](https://github.com/ethyca/fides/compare/2.40.0...main)
 
+### Changed
+ - Removed support for Python 3.8
+
```
pattisdr, line 22: Drive-by comment: this doesn't remove support for Python 3.8, it just removes the testing of it 

Slack thread: https://ethyca.slack.com/archives/C04S7K0T2V7/p1720035732012629?thread_ts=1720029377.583919&cid=C04S7K0T2V7

Fideslang specifically restricts Python versions I believe, so my guess is that we'd want Fideslang to go out as well. I am already doing this for Pydantic V2 -

pattisdr: https://github.com/ethyca/fideslang

adamsachs:
thanks for resurfacing this @Andrés. 

in general, i think the main 'risk' in adopting this approach is if we're not 'refreshing' the frozen requirements frequently and then risk our dependencies growing stale and behind the curve (and not automatically picking up useful patches).

to mitigate that a bit, perhaps we can find a way to either automate or encourage/prescribe that we refresh the frozen requirements on some cadence? as part of every release? open to other ideas/suggestions, just wanted to throw this out there.

PR #5051, opened 2024-07-02
Move .env var to before server runs

gilluminate:
This environment variable still isn't working correctly in tests because it's read by the Next server, not the client. Setting it in Cypress is never going to work because Cypress relies on the server to already be running.

### Description Of Changes

Force this env variable for all tests, so that developer's machine settings don't cause unexpected failures.

### Code Changes

* move `FIDES_PRIVACY_CENTER__IS_OVERLAY_ENABLED=true` from Cypress config to `package.json` script that runs the cypress start.

### Steps to Confirm

* set `FIDES_PRIVACY_CENTER__IS_OVERLAY_ENABLED=false` in local `privacy-center/.env`
* run `npm run cy:start` and then run `cy:run`
* All tests should pass

### DIFF
--- a/clients/privacy-center/cypress/e2e/home.cy.ts
+++ b/clients/privacy-center/cypress/e2e/home.cy.ts
@@ -35,35 +35,26 @@ describe("Home", () => {
     cy.get("body").should("have.css", "background-color", "rgb(255, 99, 71)");
   });
 
-  it(
-    "should show an empty state when notice-driven but there are no notices",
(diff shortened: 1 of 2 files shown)


NevilleS (approved):
Makes sense, I had a feeling about that `.env`...

gilluminate:
> Makes sense, I had a feeling about that `.env`...

@NevilleS yes you did!

PR #5054, opened 2024-07-02
Adding output template to read requests

galvana:
Closes [PROD-2209](https://ethyca.atlassian.net/browse/PROD-2209)

### Description Of Changes

Introduces the concept of "output templates" for `read` `SaaSRequests`. This allows the user to specify a template that will be used to generate each row of a collection response. This was done to allow param values (connector params, identity values, dataset reference values) to be aggregated in a `read` request and be used as input data for `update`/`delete` requests.

To demonstrate the usefulness of this new concept, I refactored the following integrations to use output templates instead of request overrides:

- Adyen
- AppsFlyer
- Statsig

### Code Changes

* [ ] Added new `ReadSaaSRequest` schema to only allow the `output` field for `read` requests
* [ ] Updated the base `SaaSConnector` class to support output templates
* [ ] Minor refactor of `SaaSQueryConfig`
* [ ] New tests for the validation functions of the `SaaSRequest` and `ReadSaaSRequest` schemas
* [ ] Output template tests focused on `SaaSConnector.retrieve_data()`

### Steps to Confirm

* [ ] Integration tests

### Pre-Merge Checklist

* [ ] Update `CHANGELOG.md`

[PROD-2209]: https://ethyca.atlassian.net/browse/PROD-2209?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/fides/api/schemas/saas/saas_config.py
```diff
@@ -194,8 +194,8 @@ def validate_grouped_inputs(cls, values: Dict[str, Any]) -> Dict[str, Any]:
 
         return values
 
-    @root_validator
-    def validate_override(cls, values: Dict[str, Any]) -> Dict[str, Any]:
+    @root_validator(allow_reuse=True)
```
adamsachs, line 197: nit: does this really need the `allow_reuse=True`? i thought this was used if you actually wanted to reuse the same function/method for multiple validators; in your case here, you're defining a new `validate_request` validator function in the `ReadSaaSRequest` now below, right?

src/fides/api/service/connectors/saas_connector.py
```diff
@@ -275,29 +282,75 @@ def retrieve_data(
...
+                                [param_value_map],
+                                read_request.output,
+                                processed_rows,
+                            )
+                        )
+
+            # This allows us to build an output object even if we didn't generate and execute
+            # any HTTP requests. This is useful if we just want to select specific input_data
+            # values to provide as row data to the mask_data function
+            if not read_request.path and read_request.output:
```
adamsachs, line 312: nit: this may read a bit simpler and i think is equivalent? 
```suggestion
            elif read_request.output:
```

src/fides/api/service/connectors/saas_connector.py
```diff
@@ -275,29 +282,75 @@ def retrieve_data(
...
+        if not output_template:
+            return processed_rows or []
+
+        result = []
+        for processed_row in processed_rows or [None]:  # type: ignore
+            for param_value_map in param_value_maps:
+                if processed_row:
+                    param_value_map[ALL_OBJECT_FIELDS] = processed_row
+                row = assign_placeholders(output_template, param_value_map)
+                result.append(json.loads(row))  # type: ignore
```
adamsachs, line 350: is this fully insured that it's valid, de-serializable JSON here?  do we perhaps need some validation on the `output` field for the read request? (i'll leave a comment up there too)

src/fides/api/schemas/saas/saas_config.py
```diff
@@ -223,10 +223,47 @@ def validate_override(cls, values: Dict[str, Any]) -> Dict[str, Any]:
         return values
 
 
+class ReadSaaSRequest(SaaSRequest):
+    """
+    An extension of the base SaaSRequest that allows the inclusion of an output template
+    that is used to format each collection result.
+    """
+
+    output: Optional[str]
```
adamsachs, line 232: per comment below, just wondering if we should have some sort of validation here that this is a valid JSON template?

adamsachs (approved):
i like this new feature a lot @galvana! i think the implementation looks solid, just a few nits and then a question about whether it would be wise to have some more formal guard rails around the output template value.

i'll approve tentatively though and leave that to your discretion 👍

PR #5057, opened 2024-07-03
ensure format and sort checks are run on CI

adamsachs:
Closes [PROD-2226](https://ethyca.atlassian.net/browse/PROD-2226)

### Description Of Changes

Previously, our CI's BE checks for PRs was running the "fix" version of `isort` and `black` formatting commands. The "fix" version of these commands will adjust/fix the formatting issues in the files, but it won't actually error out. This is the opposite of what we want in the CI job: the file adjustments are only made locally in the CI runner, and the job itself doesn't error to alert us that there are formatting issues present in the codebase!

This PR changes this CI behavior to run the "check" versions of these commands, which _will_ error out if errors are found. (An example is the `isort` [run](https://github.com/ethyca/fides/actions/runs/9777236095/job/26991466667) on the first commit to this PR/branch!) The PR ensures that the "fix" versions of the commands are still run when executing the `static_checks` nox session, so that users can locally run these commands to make formatting fixes to their files and then push them up, with a bit of discertion.

This has always been the intended behavior, i think; it mirrors what is done (correctly) over in `fidesplus` currently. i think this was just a longstanding missed bug in our CI pipeline here in OSS `fides`.

### Code Changes

* [x] switch the default parameterization of `isort` and `black` sessions to be `check` rather than `fix`
* [x] ensure that `static_checks` runs the `fix` parameterization of each session, since that's the intention when running `static_checks` (used for local dev, not for CI runs)

### Steps to Confirm

* [x] the `isort` [run](https://github.com/ethyca/fides/actions/runs/9777236095/job/26991466667) on the first commit to this PR/branch shows that we now successfully flag a CI error if files aren't sorted correctly; same behavior goes for `black`
* [x] i was able to run `nox -s static_checks` locally and it fixed up the files that were sorted incorrectly in the first commit

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

[PROD-2226]: https://ethyca.atlassian.net/browse/PROD-2226?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/noxfiles/ci_nox.py
+++ b/noxfiles/ci_nox.py
@@ -3,9 +3,9 @@
 from functools import partial
 from typing import Callable, Dict
 
+import nox
 from nox.command import CommandFailed
 
-import nox
 from constants_nox import (
     CONTAINER_NAME,
     IMAGE_NAME,
--- a/noxfiles/ci_nox.py
+++ b/noxfiles/ci_nox.py
@@ -3,9 +3,9 @@
 from functools import partial
 from typing import Callable, Dict
 
-import nox
 from nox.command import CommandFailed
 
+import nox
 from constants_nox import (
     CONTAINER_NAME,
     IMAGE_NAME,
@@ -59,8 +59,6 @@ def black(session: nox.Session, mode: str) -> None:
 def isort(session: nox.Session, mode: str) -> None:
     """Run the 'isort' import linter."""
     install_requirements(session)
-    command = ("isort", "--show-config")
-    session.run(*command)
     command = ("isort", "src", "tests", "noxfiles", "scripts", "noxfile.py")
     if mode == "check":
         command = (*command, "--check")
--- a/noxfiles/ci_nox.py
+++ b/noxfiles/ci_nox.py
@@ -59,6 +59,8 @@ def black(session: nox.Session, mode: str) -> None:
 def isort(session: nox.Session, mode: str) -> None:
     """Run the 'isort' import linter."""
     install_requirements(session)
+    command = ("isort", "--show-config")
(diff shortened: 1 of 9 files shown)


noxfiles/ci_nox.py
```diff
@@ -35,8 +35,8 @@ def static_checks(session: nox.Session) -> None:
 @nox.parametrize(
     "mode",
     [
-        nox.param("fix", id="fix"),
```
adamsachs, line 38: this change in order turns out to not technically be needed but i figure i'd keep it here to make it consistent with the order in `fidesplus`

noxfiles/setup_tests_nox.py
```diff
@@ -1,5 +1,7 @@
 from typing import Optional
 
+from nox import Session
```
adamsachs, line 3: an isort/formatting error that had been flying under the radar, exposed now because we're properly checking in CI

pyproject.toml
```diff
@@ -103,6 +103,7 @@ exclude = '''
 ###########
 [tool.isort]
 known_first_party = ["versioneer"]
+known_third_party = ["nox"]
```
adamsachs, line 106: i was having issues locally with `isort` picking up `nox` as a "first party" library (this was _not_ happening on CI). this forces `isort` to recognize `nox` as a _third party_ library in all contexts.

adamsachs (commented):
some notes -

PR #5060, opened 2024-07-03
Fix issue with monitor results pagination

jpople:
Closes #PROD-2233

### Description Of Changes

Fixes a bug where, when switching between pages using the result tables' pagination menu or searching, results from previous table views were remaining on the table.  

[Loom demonstrating bugged behavior](https://www.loom.com/share/acb25aef6b6349f6ae28208e1dab795e)

[Loom demonstrating fixed behavior](https://www.loom.com/share/7ea80b830cfa41acb35750009e3d6034)

The issue was that the `key` prop, which is used by React to determine which rows to clear when rerendering, wasn't unique enough-- if a result on page A had the same name as a result on page B, its row would stay on the table instead of being removed.  I fixed this by making all discovery/detection tables also include their monitor config ID in their row ID (from which the key is derived).

### Code Changes

* [ ] Unify `getRowId` across all D&D tables
* [ ] Change `getRowId` to include monitor config ID
* [ ] Update (a lot of) tests to use changed `data-testid`s

### Steps to Confirm

* [ ] Create several discovery monitors for integrations that have data (I used S3)
* [ ] Execute all monitors until there are >25 results
* [ ] View results in "Activity" and "Detection" tables
* [ ] Should be able to switch pages and search
* [ ] Table should show only the 25 results per page that are supposed to be populated

### Pre-Merge Checklist

* [ ] Update `CHANGELOG.md`


lucanovera (approved):
Good fix! Appreciate the loom videos too. Approve

PR #5070, opened 2024-07-09
fix mysql integration: pin 9, remove command

thingscouldbeworse:
prereq for PROD-2083

### Description Of Changes
removes command calling naive plugin, changes image to mysql:9 instead of 8

### Code Changes

### Steps to Confirm

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### DIFF
--- a/docker-compose.integration-tests.yml
+++ b/docker-compose.integration-tests.yml
@@ -23,9 +23,8 @@ services:
       - "0.0.0.0:6432:5432"
 
   mysql-test:
-    image: mysql:8
+    image: mysql:9
     platform: linux/amd64
-    command: --mysql-native-password=ON
     restart: always
     environment:
       - MYSQL_HOST=mysql_example
--- a/docker-compose.integration-tests.yml
+++ b/docker-compose.integration-tests.yml
@@ -22,21 +22,6 @@ services:
     ports:
       - "0.0.0.0:6432:5432"
 
-  mysql-test:
-    image: mysql:9
-    platform: linux/amd64
-    restart: always
-    environment:
-      - MYSQL_HOST=mysql_example
-      - MYSQL_DATABASE=mysql_example
-      - MYSQL_ROOT_PASSWORD=example
-      - MYSQL_USER=mysql_user
-      - MYSQL_PASSWORD=mysql_pw
-    expose:
-      - 3306
-    ports:
-      - "3306:3306"
-
   sqlserver-test:
     image: mcr.microsoft.com/azure-sql-edge:latest # Equivalent to SQL Server 2016
     environment:
@@ -46,3 +31,5 @@ services:
       - 1433
     ports:
       - "0.0.0.0:1433:1433"
+
+  # mysql integration test config lives at `docker/docker-compose.integration-mysql.yml`
--- a/docker/docker-compose.integration-mysql.yml
+++ b/docker/docker-compose.integration-mysql.yml
@@ -1,8 +1,7 @@
 services:
   mysql_example:
-    image: mysql
+    image: mysql:9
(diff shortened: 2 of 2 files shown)


docker/docker-compose.integration-mysql.yml
```diff
@@ -1,8 +1,7 @@
 services:
   mysql_example:
-    image: mysql
+    image: mysql:9
```
adamsachs, line 3: i'd actually stray away from pinning this, i think (assuming that doesn't break CI...if it did, that'd surprise me!)

the reason is that it's helpful to actually pull in the latest and greatest, it's a way to flag to us if our current integration support is not working -- for whatever reason -- with the latest version of the db. it's obviously inconvenient when CI breaks unexpectedly from underneath us, but in my view, that's a helpful defense mechanism against us accidentally making assumptions that don't hold with new versions of the software we're integrating with.

what do you think?

docker-compose.integration-tests.yml
```diff
@@ -23,9 +23,8 @@ services:
       - "0.0.0.0:6432:5432"
 
   mysql-test:
-    image: mysql:8
```
adamsachs, line 26: i see here that this one _had_ been pinned. again, i think it'd just be better to remove.

as we discussed offline, i'm just generally not sure what this compose YML is being used for anyway, so we could perhaps remove this entry all together - but we also don't _need_ to in this increment, if we want to play it safe and keep it as as simple a change as possible :)

adamsachs (commented):
@thingscouldbeworse this looks good, and thank you for taking the time to dig on this!

i just wonder whether we should keep the image unpinned - curious to see what you think of that. besides that, i think this is good to go and we can get CI fixed up on `main` 🎊

docker/docker-compose.integration-mysql.yml
thingscouldbeworse: I tend to lie on the side of pinning dependencies in tests just so we don't have something like exactly this where something fails (not silently, but quietly maybe 😄) over a holiday week and isn't caught. But it looks like integration tests don't fail builds/block releases so maybe the failure is a better indicator in a case like this, that we need to update tests. I'll defer to you, no strong feelings.

docker-compose.integration-tests.yml
thingscouldbeworse: Let me try removing it and see if it causes problems, I definitely pinned my `9` after seeing that `8` but a pinned and un-run `8` isn't useful

docker/docker-compose.integration-mysql.yml
thingscouldbeworse: Actually I revise what I said before, I think that we should be pinning to _something_ as a version, but it doesn't perturb me to pin to `latest` or something else that would allow updates. Do you know what we're getting when we just leave off a version? I'll look that up, I don't know if it switched to v9 last week as a `latest` release or `stable` or something else

docker-compose.integration-tests.yml
thingscouldbeworse: removed here: https://github.com/ethyca/fides/pull/5070/commits/5b10b02bbd03b8d02610c7aaf7b4198c3b0847c1

docker/docker-compose.integration-mysql.yml
thingscouldbeworse: pinned to `latest` here: https://github.com/ethyca/fides/pull/5070/commits/8d2f89f7e9d0ab102332dab7168cf12f20b41ae6

adamsachs (approved):
looks great, thanks for the diligence and follow up!

PR #5071, opened 2024-07-09
PROD-1898 Provide API support for (optional) pagination of the `/system` endpoint and /datasets

erosselli:
Closes [PROD-1898](https://ethyca.atlassian.net/browse/PROD-1898)

### Description Of Changes

Added optional pagination and filtering in the list endpoints for systems and datasets

### Code Changes

* Created the `apply_filters_to_query` reusable function. (In the future we'll use it in the dataset endpoints) 
* Added logic to be able to override default routes that were created with a generic route factory, when custom implementation logic is needed. In this case, this allowed adding the pagination logic to the datasets endpoint
* Added custom pagination & filtering logic to the systems endpoint

### Steps to Confirm

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

[PROD-1898]: https://ethyca.atlassian.net/browse/PROD-1898?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/fides/api/api/v1/endpoints/generic_overrides.py
```diff
@@ -0,0 +1,64 @@
...
+    data_categories: Optional[List[str]] = Query(None),
+) -> Union[Page[Dataset], List[Dataset]]:
+    """
+    Get a list of all of the Datasets.
+    If pagination parameters (size, page) are provided, then the response will be paginated & provided
+    filters (search, data_categories) will be applied.
+    Otherwise all Datasets will be returned (this may be a slow operation if there are many datasets,
+    so using the pagination parameters is recommended).
+    """
+    if page and size:
```
galvana, line 48: For paginated endpoints, it's acceptable to just pass in either `page` or `size` params and use a default value for the missing value. I think in our case, we should enable pagination if either `page` or `size` are provided
```suggestion
    if page or size:
```
Then do something like this on like 57 (uses the default values from `Params`)
```
pagination_params = Params(page=page or 1, size=size or 50)
 ```

File: src/fides/api/api/v1/endpoints/system.py [hunk 2]
```diff
@@ -366,13 +373,47 @@ async def create(
...
 ) -> List:
-    """Get a list of all of the resources of this type."""
+    """Get a list of all of the Systems.
+    If pagination parameters (size, page) are provided, then the response will be paginated & provided
+    filters (search, taxonomy fields) will be applied.
+    Otherwise all Systems will be returned (this may be a slow operation if there are many systems,
+    so using the pagination parameters is recommended).
+    """
+    if size and page:
+        pagination_params = Params(size=size, page=page)
```
galvana, line 395: Same comment as above about enabling pagination on either `size` or `page`

src/fides/api/api/v1/endpoints/system.py
```diff
@@ -366,13 +373,47 @@ async def create(
             scopes=[SYSTEM_READ],
         )
     ],
-    response_model=List[BasicSystemResponse],
+    response_model=Union[List[BasicSystemResponse], Page[BasicSystemResponse]],
     name="List",
```
galvana, line 377: I think it'd be nice to include that this is optionally paginated like you did for the dataset endpoint
```suggestion
    name="List systems (optionally paginated)",
```

galvana (commented):
@erosselli good work! I left some minor suggestions but this should be good to go once we figure out the failing tests. We can talk on Monday about why you made some of the fixture changes so we can hopefully fix the tests.

src/fides/core/api.py
```diff
@@ -11,12 +11,27 @@ def generate_resource_url(
...
+
+    processed_query_params = []
+    for key, value in query_params.items():
+        if isinstance(value, list):
+            processed_query_params.extend([f"{key}={v}" for v in value])
+        else:
+            processed_query_params.append(f"{key}={value}")
+
+    query_string = "&".join(processed_query_params)
+    print("AAA ", query_string)
```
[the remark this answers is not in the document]
erosselli: oops! yeah, I'll remove it 🙈

galvana (approved):
Thanks for making the recommended changes and for adding additional test coverage for the filter! I'm glad we rolled back the fixture changes for now and avoided having to update unrelated tests

PR #5079, opened 2024-07-11
Add support for "Key-pair" authentication to Snowflake integration

andres-torres-marroquin:
Closes #PROD-2140

### Description Of Changes

Add support for Key-Pair authentication to the SnowflakeConnector, so it now allows a couple fields to the `SnowflakeSchema`: `private_key` and `private_key_passphrase`.

### Code Changes

* [ ] Added a couple fields to the `SnowflakeSchema`: `private_key` and `private_key_passphrase`.
* [ ] Added some logic to the `SQLConnector` so now `connect_args` can be passed down to `SQLConnector.create_client` through the new `SQLConnector.get_connect_args()` method.
* [ ] Added `SnowflakeSchema.get_connect_args()` so it now takes the `private_key` and decrypts it if `private_key_passphrase` was defined as well.
* [ ] Updated tests so they are ran in `password` mode and `private_key` mode.

### Steps to Confirm

* [ ] Add a Snowflake Integration with `private_key`.

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### DIFF
--- a/src/fides/api/schemas/connection_configuration/connection_secrets_snowflake.py
+++ b/src/fides/api/schemas/connection_configuration/connection_secrets_snowflake.py
@@ -81,17 +81,13 @@ class SnowflakeSchema(ConnectionConfigSecretsSchema):
             )
 
         if not any([values.get("password"), private_key]):
-            raise ValueError(
-                "Must provide either a password or a private key."
-            )
+            raise ValueError("Must provide either a password or a private key.")
 
         if private_key:
             try:
                 values["private_key"] = format_private_key(private_key)
-            except IndexError as e:
-                raise ValueError(
-                    f"Invalid private key format"
-                )
+            except IndexError:
+                raise ValueError("Invalid private key format")
 
         return values
 


src/fides/api/schemas/connection_configuration/connection_secrets_snowflake.py
```diff
@@ -19,9 +19,19 @@ class SnowflakeSchema(ConnectionConfigSecretsSchema):
...
+        sensitive=True,
+    )
+    private_key: Optional[str] = Field(
+        title="Private key",
+        description="The private key used to authenticate and access the database. It could be encrypted or unencrypted, if encrypted, provide the passphrase.",
+        sensitive=True,
+    )
+    private_key_passphrase: Optional[str] = Field(
+        title="Passphrase",
+        description="The passphrase used for the encrypted private key.",
```
pattisdr, line 34: Drive by comment - for Pydantic V2, "optional" fields by itself will still be required, it just means that the default is allowed to be none. To keep this optional, can you add a `default=None` field here for both of these new fields.

adamsachs (commented):
thanks for getting a solid starting approach in place! here's a few initial comments, haven't had a chance to fully review yet.

src/fides/api/service/connectors/sql_connector.py
```diff
@@ -557,6 +564,27 @@ def build_uri(self) -> str:
...
+                    config.private_key.encode(),
+                    password=config.private_key_passphrase.encode(),
+                    backend=default_backend(),
+                )
+                private_key = private_key_encoded.private_bytes(
+                    encoding=serialization.Encoding.DER,
+                    format=serialization.PrivateFormat.PKCS8,
+                    encryption_algorithm=serialization.NoEncryption(),
+                )
+                connect_args["private_key"] = str(private_key)
```
adamsachs, line 585: Are we sure this works with the string? based on the [docs](https://docs.snowflake.com/en/developer-guide/python-connector/sqlalchemy#key-pair-authentication-support), it looks like it's expecting the raw bytes? just wanted to check that we've tested this.

src/fides/api/schemas/connection_configuration/connection_secrets_snowflake.py
```diff
@@ -37,20 +58,33 @@ class SnowflakeSchema(ConnectionConfigSecretsSchema):
...
         "warehouse_name",
         "database_name",
         "schema_name",
     ]
 
+    @root_validator()
+    def validate_private_key_and_password(cls, values: dict) -> dict:
+        private_key: str = values.get("private_key", "")
+
+        if values.get("password") and values.get("private_key"):
```
adamsachs, line 78: nit to use the local var 
```suggestion
        if values.get("password") and private_key:
```

src/fides/api/service/connectors/sql_connector.py
```diff
@@ -202,8 +204,13 @@ def create_client(self) -> Engine:
             uri,
             hide_parameters=self.hide_parameters,
             echo=not self.hide_parameters,
+            connect_args=self.get_connect_args(),
         )
 
+    def get_connect_args(self) -> Dict[str, Any]:
```
adamsachs, line 210: 👍

adamsachs (commented):
@andres-torres-marroquin generally looking good to me, I'd just like to get the key parsing bit sorted and removed ideally - it feels like it shouldn't be necessary, but maybe there's a nuance there I'm not aware of.

also - have you done any manual testing? we'd like to be extra sure here that this is working as we expect end to end. if you have, pasting some steps and/or screenshots would be helpful - if the screenshots are at all sensitive, feel free to paste it to the jira ticket.

tests/fixtures/snowflake_fixtures.py
```diff
@@ -94,12 +94,86 @@ def snowflake_connection_config(
...
+    yield connection_config
+    connection_config.delete(db)
+
+
+@pytest.fixture(
+    params=[
+        "snowflake_connection_config",
+        "snowflake_connection_config_with_keypair",
+    ]
+)
```
galvana, line 163: Clever!

galvana (approved):
@andres-torres-marroquin and I verified both authentication methods from the Admin UI. I just called out the need for error handling around missing passphrases for encrypted private keys but we're following up with that in a separate ticket.

PR #5099, opened 2024-07-18
Adding support for special characters in SaaS request body

galvana:
Closes [PROD-2377](https://ethyca.atlassian.net/browse/PROD-2377)

### Description Of Changes

We're seeing this error when sending non-Latin-1 characters (`1234 Peterson’s Farm Rd.`)
```
'latin-1' codec can't encode character '\u2019' in position 73: Body ('’') is not valid Latin-1. Use body.encode('utf-8') if you want to send it encoded in UTF-8.
```

This is because the `Requests` library uses `Latin-1` as the default encoding for string bodies. The fix is to call `encode("utf-8")` on the request body.

### Code Changes

* [ ] Updated `AuthenticatedClient` and `logger_context_utils.py`

### Pre-Merge Checklist

* [ ] Update `CHANGELOG.md`

[PROD-2377]: https://ethyca.atlassian.net/browse/PROD-2377?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/fides/api/service/connectors/saas/authenticated_client.py
```diff
@@ -219,6 +219,10 @@ def send(
         # extract the hostname from the complete URL and verify its safety
         deny_unsafe_hosts(urlparse(prepared_request.url).netloc)
 
+        # utf-8 encode the body before sending
+        if isinstance(prepared_request.body, str):
+            prepared_request.body = prepared_request.body.encode("utf-8")
```
galvana, line 224: The main change

src/fides/api/util/logger_context_utils.py
```diff
@@ -103,7 +103,12 @@ def request_details(
         LoggerContextKeys.url.value: prepared_request.url,
     }
     if CONFIG.dev_mode and prepared_request.body is not None:
-        details[LoggerContextKeys.body.value] = prepared_request.body
+        if isinstance(prepared_request.body, bytes):
+            details[LoggerContextKeys.body.value] = prepared_request.body.decode(
+                "utf-8"
+            )
```
galvana, line 109: Now that the body can be `bytes`, we need to decode it so we can output it to the logs.

tests/ops/integration_tests/saas/test_onesignal_task.py
```diff
@@ -5,6 +5,7 @@
 from tests.ops.test_helpers.saas_test_utils import poll_for_existence
 
 
+@pytest.mark.skip(reason="Temporarily disabled test")
```
galvana, line 8: Disabling flaky test

tests/ops/service/connectors/saas/test_authenticated_client.py
```diff
@@ -145,6 +176,19 @@ def test_client_ignores_errors(
...
+    ):
+        request_params = SaaSRequestParams(
+            method=HTTPMethod.POST,
+            path="/",
+            body='{"addr": "1234 Peterson’s Farm Rd."}',
+            headers={"Content-Type": "application/json"},
+        )
+
+        test_authenticated_client.uri = test_http_server
+        test_authenticated_client.send(request_params)
```
galvana, line 190: I'm spinning up a minimal test server so that the `send()` can get past the connection errors and actually get to the part of its code where it attempts to parse the body. I wasn't able to reproduce the error when using mocks.

tests/ops/integration_tests/saas/test_onesignal_task.py
galvana: @gilluminate brought this up, but we're most likely going to talk about it during the next standup. We might end up disabling the SaaS integration tests altogether

tests/ops/integration_tests/saas/test_onesignal_task.py
```diff
@@ -45,7 +46,7 @@ async def test_non_strict_erasure_request(
 
         player_id = onesignal_erasure_data
         (
-            access_results,
+            _,
```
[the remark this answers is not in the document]
galvana: It's just used as a "throwaway" or "don't care" variable. We need to unpack the tuple returned from `non_strict_erasure_request` but since we don't use the `access_results` we can just use `_` to be able to unpack the tuple but indicate that we don't need to use that variable.

PR #5105, opened 2024-07-18
mark events for performance monitoring

gilluminate:
### Description Of Changes

mark all FidesJS events for performance monitoring

### DIFF
--- a/clients/fides-js/src/lib/events.ts
+++ b/clients/fides-js/src/lib/events.ts
@@ -65,7 +65,7 @@ export const dispatchFidesEvent = (
     const event = new CustomEvent(type, {
       detail: { ...cookie, debug, extraDetails: constructedExtraDetails },
     });
-    performance.mark(type);
+    performance?.mark(type);
     debugLog(
       debug,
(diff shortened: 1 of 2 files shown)


clients/fides-js/src/lib/events.ts
```diff
@@ -65,6 +65,7 @@ export const dispatchFidesEvent = (
     const event = new CustomEvent(type, {
       detail: { ...cookie, debug, extraDetails: constructedExtraDetails },
     });
+    performance.mark(type);
```
NevilleS, line 68: nit:
```suggestion
    performance && performance.mark(type);
```

I know caniuse says `performance` is everywhere, but it's probably something that could be disabled by a useragent, etc?

NevilleS (changes requested):
One nitpick here!

clients/fides-js/src/lib/events.ts
gilluminate: gooooood call!

NevilleS (approved):
ship it!

PR #5110, opened 2024-07-19
backwards compatibility for TCF optimization work

gilluminate:
### Description Of Changes

Include query param when calling the privacy experience endpoint so that it's not excluded by default, thereby making it backwards compatible.

### Code Changes

* add new `exclude_gvl_languages` query param to `/api/v1/privacy-experience` api call

### Steps to Confirm

* This change won't have any real effect until the `fidesplus` backend is updated to accommodate.

### Pre-Merge Checklist

### DIFF
--- a/clients/fides-js/src/services/api.ts
+++ b/clients/fides-js/src/services/api.ts
@@ -67,6 +67,7 @@ export const fetchExperience = async (
     has_config: "true",
     systems_applicable: "true",
     include_gvl: "true",
+    exclude_gvl_languages: "true", // backwards compatibility for TCF optimization work
     include_meta: "true",
     ...(propertyId && { property_id: propertyId }),
   };
--- a/clients/privacy-center/cypress/e2e/consent-i18n.cy.ts
+++ b/clients/privacy-center/cypress/e2e/consent-i18n.cy.ts
@@ -1525,7 +1525,7 @@ describe("Consent i18n", () => {
       });
       it("falls back to default locale", () => {
         visitDemoWithI18n({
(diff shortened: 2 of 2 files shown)


clients/privacy-center/cypress/e2e/consent-i18n.cy.ts
```diff
@@ -1525,7 +1525,7 @@ describe("Consent i18n", () => {
       });
       it("falls back to default locale", () => {
         visitDemoWithI18n({
-          navigatorLanguage: ENGLISH_LOCALE,
```
gilluminate, line 1528: updated this test as an aside to help catch issues with the previous hotfix.

clients/fides-js/src/services/api.ts
```diff
@@ -67,6 +67,7 @@ export const fetchExperience = async (
     has_config: "true",
     systems_applicable: "true",
     include_gvl: "true",
+    exclude_gvl_languages: "true", // backwards compatibility for TCF optimization work
```
pattisdr, line 70: You said `exclude_gvl_translations` in slack! 

I'll update to be `exclude_gvl_languages` though to match here, I'm not opinionated since I'm viewing this as a temporary query param.

gilluminate: oh! you are correct. should have been `exclude_gvl_translations`

pattisdr: Updated! Backend is here: https://github.com/ethyca/fidesplus/pull/1521

pattisdr (approved):

PR #5114, opened 2024-07-22
Fix bug with changing monitor frequency from scheduled to "not scheduled"

jpople:
Closes PROD-2305

### Description Of Changes

Fixes a bug where discovery monitors that were scheduled for execution couldn't have their execution frequency changed to "not scheduled".

The issue was that the backend was treating a value of "undefined" for `execution_frequency` as "not scheduled", but the frontend isn't able to explicitly send "undefined" in its API request's payload.  To fix, added a `NOT_SCHEDULED` option to the MonitorFrequency enum and updated form on frontend and validation on backend to handle it as an explicit value.

### Code Changes

* [ ] Add `NOT_SCHEDULED` to MonitorFrequency enum
* [ ] Update UI to send "Not scheduled" instead of `undefined` when "Not scheduled" is selected in the form

### Steps to Confirm

Must run Plus backend on branch `jpople/prod-2305/not-scheduled-edit-fix` (see [here](https://github.com/ethyca/fidesplus/pull/1522))
* [ ] Create a discovery monitor with execution frequency set
* [ ] Edit that monitor, changing its frequency to "Not scheduled"
* [ ] Change should persist to DB

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`


src/fides/api/models/detection_discovery.py
```diff
@@ -205,6 +206,9 @@ def derive_execution_trigger_dict(data: Dict[str, Any]) -> None:
         """
         execution_frequency = data.pop("execution_frequency", None)
         execution_start_date = data.pop("execution_start_date", None)
+        if execution_frequency == MonitorFrequency.NOT_SCHEDULED:
+            data["monitor_execution_trigger"] = None
+            return
```
erosselli, line 211: Maybe we can add a case to the parameters of the test `test_create_monitor_config_execution_trigger_logic` so that we cover this line , what do you think?

erosselli: actually I see you did this already, unsure why codecov's reporting that these lines aren't covered

erosselli (approved):
Nice work, specially adding tests 😎

PR #5125, opened 2024-07-23
Allow banner to show notice information

gilluminate:
Closes [PROD-2427](https://ethyca.atlassian.net/browse/PROD-2427) and [PROD-2386](https://ethyca.atlassian.net/browse/PROD-2386)

### Description Of Changes

Update experience form with option for including notices in the banner & update banner to accommodate.

### Code Changes

* Added support for new experience config props `show_layer1_notices` and `layer1_button_options`
* Added notices toggle in UI of Privacy Experience form and dropdown to select Acknowledge button.
* Added list of notices in FidesJS if toggle is on, as well as display/save correct button if Acknowledge button is selected.
* Update rules for calling notices served API
--
* Fixed unrelated UI issue where the delete button for Notices wasn't able to be clicked because of layout issues
* Fixed unrelated failing FF Cypress test due to `realClick` being unavailable for that browser

---

CleanShot 2024-07-30 at 12 16 20@2x

---

CleanShot 2024-07-30 at 12 17 16@2x

---

### Steps to Confirm

* Visit Privacy Experiences in Admin UI
* Click an experience that has banners
* Add Notices if there aren't any on the experience
* toggle the new switch "Allow banner to show notice information"
* Notice the Preview window shows the notices in a list on the banner
* Click to open Banner Options menu
* Select Acknowledge option
* Notice the Preview window shows the acknowledge button in place of Opt In/Opt Out buttons

- [x] NOTE: saving will not work until related BE is done.

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

[PROD-2427]: https://ethyca.atlassian.net/browse/PROD-2427?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[PROD-2386]: https://ethyca.atlassian.net/browse/PROD-2386?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/clients/fides-js/src/components/fides.css
+++ b/clients/fides-js/src/components/fides.css
@@ -257,17 +257,6 @@ div#fides-button-group {
   z-index: 5;
 }
 
-div.fides-acknowledge-button-container {
-  margin-inline: var(--fides-overlay-padding);
-  margin-bottom: var(--fides-overlay-padding);
-  display: flex;
-  justify-content: end;
-}
-
-div.fides-banner-acknowledge .fides-banner-button {
-  max-width: 168px;
-}
-
 button.fides-banner-button {
   font-size: var(--fides-overlay-font-size-buttons);
   display: inline-block;
@@ -321,6 +310,10 @@ button.fides-banner-button.fides-banner-button-tertiary {
   font-weight: 500;
   font-size: var(--fides-overlay-font-size-body);
   line-height: 1.25em;
+}
+
+button.fides-banner-button.fides-acknowledge-button {
+  min-width: 160px;
 }
 
 /** Modal */
--- a/clients/fides-js/src/components/fides.css
+++ b/clients/fides-js/src/components/fides.css
@@ -244,6 +244,10 @@ div#fides-banner-description {
   flex: 1;
 }
 
+div#fides-banner-notices {
+  margin-top: 16px;
+}
+
 div#fides-button-group {
   margin-top: 8px;
   margin-bottom: var(--fides-overlay-padding);
--- a/clients/fides-js/src/components/fides.css
+++ b/clients/fides-js/src/components/fides.css
@@ -481,7 +481,7 @@ div#fides-consent-content .fides-modal-description {
 }
 
 /* Responsive overlay */
-@media (max-width: 48em) {
+@media (max-width: 768px) {
   div.fides-modal-content,
   div#fides-consent-content {
     width: 100% !important;
@@ -1099,6 +1099,10 @@ div#fides-overlay-wrapper .fides-toggle .fides-toggle-display {
     div#fides-button-group
     .fides-banner-button-group {
     padding-left: 0;
+  }
+  .fides-banner-secondary-actions {
+    flex-direction: column-reverse;
+    gap: 12px;
   }
 }
 /* TCF should always be full width and not floating */
--- a/clients/fides-js/src/components/notices/NoticeOverlay.tsx
+++ b/clients/fides-js/src/components/notices/NoticeOverlay.tsx
@@ -247,6 +247,8 @@ const NoticeOverlay: FunctionComponent<OverlayProps> = ({
 
   const isDismissable = !!experience.experience_config?.dismissable;
 
+  const isSaveOnly = privacyNoticeItems.length === 1;
+
   return (
     <Overlay
       options={options}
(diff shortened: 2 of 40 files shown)


clients/fides-js/src/components/notices/NoticeOverlay.tsx
```diff
@@ -263,37 +266,43 @@ const NoticeOverlay: FunctionComponent<OverlayProps> = ({
...
+                enabledKeys={draftEnabledNoticeKeys}
+                onSave={(
+                  consentMethod: ConsentMethod,
+                  keys: Array<PrivacyNotice["notice_key"]>
+                ) => {
+                  handleUpdatePreferences(consentMethod, keys);
+                  onSave();
+                }}
+                isAcknowledge={isAcknowledge}
+                hideOptInOut={isAcknowledge}
```
eastandwestwind, line 299: do we also need to hide opt in/out when `isSaveOnly==true` here?

clients/fides-js/src/lib/hooks/useConsentServed.ts
```diff
@@ -39,13 +39,17 @@ export const useConsentServed = ({
       }
 
       // Disable the notices-served API if the serving component is a regular
-      // banner (or unknown!). This means we trigger the API for:
+      // banner (or unknown!) unless the option for displaying notices in the
+      // banner has been set. This means we trigger the API for:
       // 1) MODAL
       // 2) TCF_OVERLAY
       // 3) TCF_BANNER
+      // 4) BANNER when show_layer1_notices is true
```
eastandwestwind, line 47: thanks for updating!

clients/fides-js/src/components/fides.css
```diff
@@ -253,17 +257,6 @@ div#fides-button-group {
...
-  margin-inline: var(--fides-overlay-padding);
-  margin-bottom: var(--fides-overlay-padding);
-  display: flex;
-  justify-content: end;
-}
-
-div.fides-banner-acknowledge .fides-banner-button {
-  max-width: 168px;
-}
-
```
eastandwestwind, line 266: CSS changes always make me uneasy as we can't confirm 100% the changes  we make don't affect our customers. Do we have to remove existing CSS to get the new acknowledge button working correctly? Just curious if we can minimize changes at all to the classes that already exist

clients/fides-js/src/components/fides.css
```diff
@@ -319,6 +312,10 @@ button.fides-banner-button.fides-banner-button-tertiary {
   line-height: 1.25em;
 }
 
+button.fides-banner-button.fides-acknowledge-button {
```
eastandwestwind, line 315: This class existed before- looks like we were using `168px` before. Is there a strong reason to change this?

clients/fides-js/src/components/fides.css
```diff
@@ -1096,6 +1093,10 @@ div#fides-overlay-wrapper .fides-toggle .fides-toggle-display {
     .fides-banner-button-group {
     padding-left: 0;
   }
+  .fides-banner-secondary-actions {
```
eastandwestwind, line 1096: Did this class exist before? If so, are we certain it won't break current UI?

clients/fides-js/src/components/fides.css
```diff
@@ -244,6 +244,10 @@ div#fides-banner-description {
   flex: 1;
 }
 
+div#fides-banner-notices {
+  margin-top: 16px;
+}
+
```
eastandwestwind, line 250: This class is new, so this looks good as it's a low risk CSS change 👍

eastandwestwind (commented):
First pass code-only review- just some small things.

The most risky thing I want to make sure of is the CSS changes- let's just eval that what we change with existing classes / DOM elements won't break current customers, or is at least low-risk

clients/fides-js/src/components/notices/NoticeOverlay.tsx
gilluminate: I didn't change the "Save Only" scenario at all. It was only a factor for modal before, so I kept it that way here.

clients/fides-js/src/components/fides.css
gilluminate: From what I could tell, `fides-acknowledge-button-container` didn't exist and `.fides-banner-button` was never a child of `.fides-banner-acknowledge` so neither of these styles were ever applied. This wasn't so much of a change as it was a fix with the new styles added below along with piggy-backing on the "first button" logic/styles for the acknowledge button.

clients/fides-js/src/components/fides.css
gilluminate: the class existed but looked broken in mobile view with the "Manage Preferences" button appearing _below_ the language switcher. This update puts the language switcher at the bottom as expected. Happy to remove this change from this current PR and create a "bug" if we think that's a safer bet.

clients/fides-js/src/components/fides.css
gilluminate: as mentioned above, it was never actually being applied. The old was also a 'max-width' before which didn't make much sense. I'm using a 'min-width' to make the short "Ok" string not look to small, and matching Jack's designs in Figma which used 160px.

clients/fides-js/src/components/fides.css
gilluminate: ignore my last comment. I just double checked this stuff...

The reason for this change is that the old/existing Acknowledge button did not include the language switcher, or the manage and privacy links. It was just an "Ok" button by itself.

CleanShot 2024-07-30 at 15 54 25@2x

When I noticed that, I checked with Jack who confirmed that the only change should be the Opt In/Out buttons should be replaced with the Acknowledge button.

These styles were supporting the standalone button only.

To fix it, I moved the Acknowledge button as part of the "first button" (ie. Save Button) logic instead, which is integrated with the actions to the left already. The "first button" styles apply now, and these are obsolete.

Any custom styles would have already been applied to "first button" so I feel confident in this update, especially since it now correctly includes the fix.

clients/fides-js/src/components/fides.css
gilluminate: Jack's latest design had 160 so when I made the change mentioned above, I also went with his size.

clients/fides-js/src/components/fides.css
eastandwestwind: this is fine to keep here, just wanted to make sure we keep any changes to a min. But this fixes an existing issue, so this is fine!

clients/fides-js/src/components/fides.css
eastandwestwind: Ohh ok this makes sense, thanks for explaining!

clients/fides-js/src/lib/hooks/useMediaQuery.ts
```diff
@@ -1,5 +1,11 @@
 import { useEffect, useState } from "preact/hooks";
 
+/**
+ * This React hook listens for matches to a CSS media query. It allows the rendering of components based on whether the query matches or not. The media query string can be any valid CSS media query, for example '(prefers-color-scheme: dark)'.
+ * @param query - The CSS media query string to match against the viewport.
+ * @example const isMobile = useMediaQuery('(max-width: 768px)');
+ * @returns A boolean value indicating whether the media query matches the current viewport.
```
eastandwestwind, line 7: 💯

eastandwestwind (approved):
Ran through manual testing matrix via screenshare, everything is working as expected!

PR #5126, opened 2024-07-24
Pydantic v2 migration (misc fixes)

galvana:
### Description Of Changes

A branch to track any misc changes I need to make as part of the Pydantic v2 migration review

### Pre-Merge Checklist

### DIFF
--- a/src/fides/api/schemas/redis_cache.py
+++ b/src/fides/api/schemas/redis_cache.py
@@ -30,12 +30,14 @@ class Identity(IdentityBase):
     """Some PII grouping pertaining to a human"""
 
     # These are repeated so we can continue to forbid extra fields
-    phone_number: Optional[PhoneNumber] = Field(None, title="Phone number")
-    email: Optional[EmailStr] = Field(None, title="Email")
-    ga_client_id: Optional[str] = Field(None, title="GA client ID")
-    ljt_readerID: Optional[str] = Field(None, title="LJT reader ID")
-    fides_user_device_id: Optional[str] = Field(None, title="Fides user device ID")
-    external_id: Optional[str] = Field(None, title="External ID")
+    phone_number: Optional[PhoneNumber] = Field(default=None, title="Phone number")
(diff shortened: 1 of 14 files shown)


src/fides/api/schemas/redis_cache.py
```diff
@@ -95,7 +97,7 @@ def labeled_dict(
         self, include_default_labels: Optional[bool] = False
     ) -> Dict[str, Any]:
         """Returns a dictionary that preserves the labels for all custom/labeled identities."""
-        d = super().dict()
+        d = super().model_dump()
```
pattisdr, line 100: I think this one was intentional because `model_dump()` is overridden here too?

pattisdr: I need to probably figure out how to do this better

pattisdr:
Ah I had been deprioritizing this, but in fidesplus testing D&D I am seeing warnings getting turned around and then flagged as an error

```bash
fidesplus  | pydantic.warnings.PydanticDeprecatedSince20: The `dict` method is deprecated; use `model_dump` instead. Deprecated in Pydantic V2.0 to be removed in V3.0. See Pydantic V2 Migration Guide at https://errors.pydantic.dev/2.7/migration/
fidesplus  | 2024-07-25 23:11:24.353 | ERROR    |  - Failed to label staged resource | {'resource': 'REDACTED', 'error': 'The `dict` method is deprecated; use `model_dump` instead. Deprecated in Pydantic V2.0 to be removed in V3.0. See Pydantic V2 Migration Guide at https://errors.pydantic.dev/2.7/migration/'}
```

pattisdr:
I'm going to go ahead and merge this one Adrian!

src/fides/api/schemas/redis_cache.py
pattisdr: Or maybe it works because I later changed it to loop through `for field in self.__pydantic_extra__ or {}` in this function body

PR #5132, opened 2024-07-26
PROD-2473 Update `GET /dataset` endpoint to support `exclude_saas_datasets` and `only_unlinked_datasets` params

erosselli:
Closes [PROD-2473](https://ethyca.atlassian.net/browse/PROD-2473)

### Description Of Changes
Added two query params `exclude_saas_datasets` and `only_unlinked_datasets` to the `GET /datasets` endpoint to match those in `GET /filters/datasets`. 

### Code Changes

* Updated `list_dataset_paginated` to add the two new filter options,  `exclude_saas_datasets` and `only_unlinked_datasets`
* Added relevant unit tests
* Added comment to docstring for `GET /filter/datasets` endpoint

### Steps to Confirm

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

[PROD-2473]: https://ethyca.atlassian.net/browse/PROD-2473?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/fides/api/api/v1/endpoints/dataset_endpoints.py
+++ b/src/fides/api/api/v1/endpoints/dataset_endpoints.py
@@ -591,6 +591,7 @@ def get_ctl_datasets(
     only_unlinked_datasets: bool = False,
 ) -> List[Dataset]:
     """
+    Deprecated. Use `GET /datasets` instead.
     Returns all CTL datasets .
     """
 
--- a/src/fides/api/api/v1/endpoints/dataset_endpoints.py
+++ b/src/fides/api/api/v1/endpoints/dataset_endpoints.py
@@ -584,6 +584,7 @@ def delete_dataset(
     f"/filter{DATASETS}",
     dependencies=[Security(verify_oauth_client, scopes=[DATASET_READ])],
     response_model=List[Dataset],
+    deprecated=True,
 )
 def get_ctl_datasets(
     db: Session = Depends(deps.get_db),
--- a/src/fides/api/api/v1/endpoints/generic_overrides.py
+++ b/src/fides/api/api/v1/endpoints/generic_overrides.py
@@ -46,7 +46,7 @@ async def list_dataset_paginated(
     """
     Get a list of all of the Datasets.
     If any pagination parameters (size or page) are provided, then the response will be paginated
-    & provided filters (search, data_categories) will be applied.
+    & provided filters (search, data_categories, exclude_saas_datasets, only_unlinked_datasets) will be applied.
     Otherwise all Datasets will be returned (this may be a slow operation if there are many datasets,
     so using the pagination parameters is recommended).
     """
--- a/src/fides/api/api/v1/endpoints/generic_overrides.py
+++ b/src/fides/api/api/v1/endpoints/generic_overrides.py
@@ -4,11 +4,13 @@ from fastapi import APIRouter, Depends, Query, Security
 from fastapi_pagination import Page, Params
 from fastapi_pagination.ext.async_sqlalchemy import paginate as async_paginate
 from fideslang.models import Dataset
+from sqlalchemy import and_, not_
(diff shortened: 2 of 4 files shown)


src/fides/api/api/v1/endpoints/dataset_endpoints.py
```diff
@@ -591,6 +591,7 @@ def get_ctl_datasets(
     only_unlinked_datasets: bool = False,
 ) -> List[Dataset]:
     """
+    Deprecated. Use `GET /datasets` instead.
```
galvana, line 594: You can add a `deprecated` flag to the router info, here's an [example](https://github.com/ethyca/fides/blob/main/src/fides/api/api/v1/endpoints/privacy_request_endpoints.py#L689):

```
@router.get(
    PRIVACY_REQUESTS,
    dependencies=[Security(verify_oauth_client, scopes=[PRIVACY_REQUEST_READ])],
    response_model=Page[
        Union[
            PrivacyRequestVerboseResponse,
            PrivacyRequestResponse,
        ]
    ],
    deprecated=True,
)
def get_request_status(
```

Then it will be nice and obvious in the Swagger docs
image

src/fides/api/api/v1/endpoints/generic_overrides.py
```diff
@@ -37,27 +40,47 @@ async def list_dataset_paginated(
...
-    if page or size:
-        query = select(CtlDataset)
-        filter_params = FilterParams(search=search, data_categories=data_categories)
-        filtered_query = apply_filters_to_query(
-            query=query,
-            search_model=CtlDataset,
-            taxonomy_model=CtlDataset,
-            filter_params=filter_params,
+    if not page and not size:
+        return await list_resource(CtlDataset, db)
```
galvana, line 54: We would need to apply the `exclude_saas_datasets` and `only_unlinked_datasets` filters to the non-paginated response if we wanted to keep backwards compatibility. Did we make a decision to omit this?

tests/ops/api/v1/endpoints/test_dataset_endpoints.py
```diff
@@ -2001,7 +2001,7 @@ def test_list_dataset_no_pagination(
         api_client: TestClient,
         generate_auth_header,
         ctl_dataset,
-        saas_ctl_dataset,
+        secondary_sendgrid_instance,
```
galvana, line 2004: Nice reuse!

galvana (commented):
Looks good! I just have one question about backwards compatibility and a suggestion for Swagger

src/fides/api/api/v1/endpoints/dataset_endpoints.py
erosselli: ah cool! didn't know this 😎

src/fides/api/api/v1/endpoints/generic_overrides.py
erosselli: hmm that's a good point. I can definitely add them in -- however I think the idea was to at some point in the future deprecate the non-paginated response entirely since it has performance issues when there's a big number of elements, and the existing non-paginated endpoint with the filters (`/filters/dataset`) still exists so it _can_ be used if that's still needed. 

What do you think? I think ideally if anyone wants to move from the deprecated endpoint to the new one, they should probably add in the pagination support while they're it?

erosselli: added it in https://github.com/ethyca/fides/pull/5132/commits/071258d5d423d453aecb6decaca8d9e424566f57

galvana (approved):
Thanks for making the changes, I'm glad it didn't end up being a big change!

PR #5138, opened 2024-07-30
Prod 2435: bq connection test

thingscouldbeworse:
Closes #PROD-2435

### Description Of Changes
A more reliable connection test for bigquery

### Code Changes

* Override the base `test_connection` function of Bigquery connectors to run `list_all_projects` and return failure on an empty return value.

### Steps to Confirm

* Create any integration for a Bigquery data source and click "test integration"
* On a dev instance manually force an exception (`raise`, `i =1/0`, etc) inside the try/except
* Manually set `all_projects` to `[]` and confirm the test fails

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`


src/fides/api/util/connection_util.py
```diff
@@ -323,6 +324,26 @@ def connection_status(
 ) -> TestStatusMessage:
     """Connect, verify with a trivial query or API request, and report the status."""
 
+    # test if this is a Bigquery connection
+    if connection_config.connection_type == ConnectionType.bigquery:
```
adamsachs, line 328: hmm. this will probably work, but i think there may be a cleaner way to inject the "custom" bigquery logic.

what i'd lean toward is implementing an overridden `test_connection` method in the `BigQueryConnector` class that's a subclass of the `SQLConnector` (this is the bigquery "DSR connector"): https://github.com/ethyca/fides/blob/main/src/fides/api/service/connectors/sql_connector.py#L476

that allows us to keep all the generic wrapping logic in tact, and it puts in the "custom" bigquery logic into a class specifically scoped toward bigquery. this will pave the road nicely for when we move the `BigQueryConnector` to leverage the python bigquery client, as is currently used in the d&d workflow.

src/fides/api/service/connectors/sql_connector.py
```diff
@@ -510,6 +511,19 @@ def query_config(self, node: ExecutionNode) -> BigQueryQueryConfig:
...
+    # Overrides SQLConnector.test_connection
+    def test_connection(self) -> Optional[ConnectionTestStatus]:
+        try:
+            bq_schema = BigQuerySchema(**self.configuration.secrets)
+            client = bq_schema.get_client()
+            all_projects = [project for project in client.list_projects()]
+            if all_projects:
+                return ConnectionTestStatus.succeeded
+            else:
+                raise ConnectionException("No datasets found with the provided credentials.")
```
adamsachs, line 523: this is nice handling! just a tweak on the wording:
```suggestion
                raise ConnectionException("No bigquery projects found with the provided credentials.")
```

src/fides/api/service/connectors/sql_connector.py
```diff
@@ -510,6 +511,19 @@ def query_config(self, node: ExecutionNode) -> BigQueryQueryConfig:
...
+        except Exception as e:
+            raise ConnectionException(f"Connection error: {e}")
```
adamsachs, line 525: this is solid for the exception that's raised up (and hopefully gets to the UI?) in addition, it might be nice to _log_ the exception here too, ideally with a full stack trace - since this endpoint is specifically for troubleshooting, giving as much error information as possible can be helpful.

you can use `logger.exception` - i think that may give you a stack trace, or you may need to still specify some args/options to get a full stack trace.

src/fides/api/service/connectors/sql_connector.py
```diff
@@ -510,6 +511,19 @@ def query_config(self, node: ExecutionNode) -> BigQueryQueryConfig:
         """Query wrapper corresponding to the input execution_node."""
         return BigQueryQueryConfig(node)
 
+    # Overrides SQLConnector.test_connection
+    def test_connection(self) -> Optional[ConnectionTestStatus]:
```
adamsachs, line 515: may be a question of preference, but i tend to put these sort of comments as proper method docstrings, and that plays nicely with many IDEs. i would note here that we're specifically testing with the native bigquery python client that's used for the d&d workflows, even though it is not _yet_ used for the DSR workflows. something like this?

```suggestion
    def test_connection(self) -> Optional[ConnectionTestStatus]:
    """
    Overrides SQLConnector.test_connection with a BigQuery-specific connection test.
    
    The connection is tested using the native python client for BigQuery, since that is what's used
    by the detection and discovery workflows/codepaths. We will eventually look to migrate the
    rest of this class, used for DSR execution, to also make use of the native bigquery client. 
    """
```

adamsachs (commented):
@thingscouldbeworse this looks good to me! just a few things that i think we could polish up here to get it as clean as possible, but generally this is how i was imagining this would work - nice job!

also, i assume you've done some manual testing? i didn't get to do any yet. if you've done that, could you please list out the manual steps you've taken to confirm in the PR description? i think it would be good for at least one of us to manually test out some of the error cases locally and make sure this is working as we expect - if you've done that, then great

src/fides/api/service/connectors/sql_connector.py
thingscouldbeworse: docstring here: https://github.com/ethyca/fides/pull/5138/commits/16d65db4c965c945833221f7d03c3fb705a31073#diff-15895ed412c26a8f5c097a7eddfa11273da29a8d999e7c342bdeb475224fff8aR516

src/fides/api/service/connectors/sql_connector.py
thingscouldbeworse: logger exception here: https://github.com/ethyca/fides/pull/5138/commits/16d65db4c965c945833221f7d03c3fb705a31073#diff-15895ed412c26a8f5c097a7eddfa11273da29a8d999e7c342bdeb475224fff8aR537 I do see the exception in logs if I induce one

src/fides/api/service/connectors/sql_connector.py
thingscouldbeworse: fixed here https://github.com/ethyca/fides/pull/5138/commits/16d65db4c965c945833221f7d03c3fb705a31073#diff-15895ed412c26a8f5c097a7eddfa11273da29a8d999e7c342bdeb475224fff8aR534

adamsachs (approved):
looks good @thingscouldbeworse, thanks for making the requested updates! 👍

PR #5140, opened 2024-07-30
PROD-2350 DSR Support for ScyllaDB

erosselli:
Closes [PROD-2350](https://ethyca.atlassian.net/browse/PROD-2350)

### Description Of Changes

This PR adds DSR support for ScyllaDB. Scylla connector already existed because we support ScyllaDB for D & D , so this PR just needed to add some extra methods to the existing connector and create a `ScyllaDBQueryConfig` class.  Scylla uses Cassandra Query Language (CQL) which is very SQL-like in its syntax, so I tried to reuse as much of the SQL code as possible while making the CQL-specific changes. 

---
**_PERFORMANCE CONCERNS_**

I have some concerns about the performance of DSR queries against Scylla DB that I want to call out. 

Scylla DB [encourages a query-first approach to data modelling](https://docs.scylladb.com/stable/get-started/data-modeling/index.html#data-modeling), where you design your data model around the queries you'll need to execute. In our case, we don't have control over how the data is modeled, since the DB is not ours; we're running a very specific set of queries for the DSR feature but if the data model was not built with these types of queries in mind, queries may have big performance issues, especially when dealing with a lot of data. See [my comment on the `ALLOW FILTERING` clause](https://github.com/ethyca/fides/pull/5140#discussion_r1707022261) in the PR for more detail.

---

### Code Changes

* Implemented the `retrieve_data` and `mask_data` methods in `ScyllaConnector`
* Created the `ScyllaDBQueryConfig` class (`src/fides/api/service/connectors/scylla_query_config.py`)
    * In order to be able to reuse some code, I moved some of the methods from the existing `SQLQueryConfig` class into a more generic, abstract `SQLLikeQueryConfig` .
    * Both the existing `SQLQueryConfig` and the new `ScyllaDBQueryConfig` classes inherit from this abstract class
* Added integration tests in `tests/ops/integration_tests/test_sql_task.py` and 
* Added an example scylla dataset `data/dataset/scylladb_example_test_dataset.yml`

#### Configuring the integration
* Run fides with the scylla integration: `nox -s dev -- shell scylladb`
* From the fides admin (or the fides API) , create/edit a system. Under the system's `integrations` tab, add a Scylla DB integration (credentials and details on how to configure the integration can be found in [this confluence page](https://ethyca.atlassian.net/wiki/spaces/EN/pages/3098345474/2024-06-03+ScyllaDB+Data+Detection+and+Discovery+Design+Doc))
* Click Test Connection and make sure it works
* Go to the Datasets page, and add a dataset using the `data/dataset/scylladb_example_test_dataset.yml` file
* Go back to the system you created/edited , go to the integrations tab under this system, and in the Dataset dropdown of the integration choose the dataset you just added (Example ScyllaDB dataset)
* Save the integration

#### Access Requests
* Run the privacy experience, and click "Access your data". Input the email `customer-1@example.com` 
     * Not sure if there's any way to see the results cause when I run this I see this message `Message send is disabled for action type privacy_request_complete_access` , but I was just printing the retrieved rows in `ScyllaConnector.result_to_rows` in order to check that they're correct
     * Printing the results should show one row from the `users` table, three rows from the `user_activity` table, two rows from the `payment_methods` table and two rows from the `orders` table
     
#### Erasure Requests
 * Go back to the privacy experience and click "Erase your Data". Input the same email as before `customer-1@example.com`. 
 * You should not see any errors in the fides terminal output
 * Go to the `scylla-example` container in Docker Desktop , and click on the `Exec` tab (or exec into the docker container without  Docker Desktop if you prefer) 
 * Open a CQL shell, running `cqlsh -u cassandra` (it'll prompt you for the password) 
 * Run `USE app_keyspace;` 
 * Query the users, orders, user_activity, and payment_methods to check that data was masked appropriately 
     * As of today, we only mask string data types I think
     * To query, you can just `SELECT` like in SQL. If you want to add a WHERE clause, make sure you add ALLOW FILTERING at the end, e.g `SELECT name, user_id FROM users WHERE email = 'customer-1@example.com' ALLOW FILTERING;`

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

[PROD-2350]: https://ethyca.atlassian.net/browse/PROD-2350?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.vscode/settings.json
```diff
@@ -14,5 +14,8 @@
   "typescript.enablePromptUseWorkspaceTsdk": true,
   "typescript.tsdk": "./clients/node_modules/typescript/lib",
   "editor.defaultFormatter": "esbenp.prettier-vscode",
-  "typescript.preferences.importModuleSpecifier": "non-relative"
+  "typescript.preferences.importModuleSpecifier": "non-relative",
+  "[python]": {
+    "editor.defaultFormatter": "ms-python.black-formatter"
+  }
```
erosselli, line 20: I just wanted to be able to auto-format from vscode -- if this is annoying for others I can remove

data/dataset/scylladb_example_test_dataset.yml
```diff
@@ -0,0 +1,78 @@
+dataset:
+  - fides_key: scylladb_example_test_dataset
+    data_categories: []
+    description: ScyllaDB dataset containing a users table and user_activity table.
+    name: Example ScyllaDB dataset
```
erosselli, line 5: example scylla dataset -- some of the data categories are. bit wonky, I auto-generated the dataset using the D & D feature so categories were added by the classifier 😅

docker/sample_data/scylla/scylla_example.cql
```diff
@@ -5,16 +5,45 @@ CREATE TABLE if not exists users (
...
+INSERT INTO users (user_id, name, age, email, alternative_contacts, do_not_contact, last_contacted, states_lived, logins) VALUES (8, 'Jasmine', 34, 'customer-8@example.com', { 'phone': '+1 (831) 214-12359',  'work_email': 'customer-8@example.com' }, false, '2024-01-23', {'CA', 'NY'}, ['08:12:54.123456', '08:12:54.123456']) IF NOT EXISTS;
 
 
+CREATE TABLE if not exists user_activity (
+    user_id INT,
+    timestamp TIMESTAMP,
+    user_agent TEXT,
+    activity_type TEXT,
+    PRIMARY KEY (user_id, timestamp)
+);
```
erosselli, line 39: example of a table that relates to another table, as well as of a composite primary key

src/fides/api/service/connectors/scylla_connector.py
```diff
@@ -97,7 +147,21 @@ def mask_data(
...
+        for row in rows:
+            generated_update_stmt = query_config.generate_update_stmt(
+                row, policy, privacy_request
+            )
+            if generated_update_stmt is not None:
+                statement, params = generated_update_stmt
+                with client.connect() as connection:
+                    self.set_schema(connection)
+                    connection.execute(statement, parameters=params)
+                    # FIXME: scylla doesn't return the number of updated rows -- unsure if this is important?
```
erosselli, line 163: We use the number of updated rows in `src/fides/api/task/graph_task.py` , line 650, in the call to 
```py
self.resources.cache_erasure(
            self.key.value, output
 )
```

not sure if there are any implications to returning None / 0 / some other meaningless value

src/fides/api/service/connectors/scylla_query_config.py
```diff
@@ -0,0 +1,74 @@
...
+class ScyllaDBQueryConfig(SQLLikeQueryConfig[ScyllaDBStatement]):
+    def dry_run_query(self) -> Optional[str]:
+        """dry run query for display"""
+        return None
+
+    def query_to_str(
+        self, t: ScyllaDBStatement, input_data: Dict[str, List[Any]]
+    ) -> Optional[str]:
+        """Convert query to string"""
+        return None
```
erosselli, line 26: I wasn't sure if I needed to implement these, I saw that some connectors implement them and some don't , not sure how to tell if it's needed

tests/ops/integration_tests/test_sql_task.py
```diff
@@ -776,6 +778,146 @@ async def test_mariadb_access_request_task(
...
+        policy,
+        integration_scylladb_config,
+        scylladb_integration_no_keyspace,
+        privacy_request,
+        dsr_version,
+        request,
+    ) -> None:
+        request.getfixturevalue(dsr_version)  # REQUIRED to test both DSR 3.0 and 2.0
+
+        with pytest.raises(ScyllaConnectorMissingKeyspace) as err:
```
erosselli, line 801: so this fails for DSR 3.0 with a `did not raise` error -- I think this is because the error is raised in a celery task and so the test doesn't catch it? I do see the exception in the logs. Not too sure how to fix this 😓

tests/ops/integration_tests/test_sql_task.py
```diff
@@ -776,6 +778,146 @@ async def test_mariadb_access_request_task(
...
+            [integration_scylladb_config_with_keyspace],
+            {"email": seed_email},
+            get_cached_data_for_erasures(privacy_request.id),
+            db,
+        )
+        # FIXME: this isn't really testing anything, but scylla doesn't return the number of updated rows :(
+        assert results == {
+            "scylla_example_with_keyspace:user_activity": 0,
+            "scylla_example_with_keyspace:users": 0,
+        }
```
erosselli, line 918: I'm trying to come up with a better way to test the erasure request since this is a useless assertion

src/fides/api/service/connectors/scylla_query_config.py
```diff
@@ -0,0 +1,92 @@
...
+            return f"{string_path} IN ({operand})"
+        return f"{string_path} {operator} %({operand})s"
+
+    def get_formatted_query_string(
+        self,
+        field_list: str,
+        clauses: List[str],
+    ) -> str:
+        """Returns an SQL query string."""
+        return f"SELECT {field_list} FROM {self.node.collection.name} WHERE {' OR '.join(clauses)} ALLOW FILTERING;"
```
erosselli, line 60: CQL does not encourage you to filter on any column in a `WHERE` clause -- rather it only wants you to filter on clustering columns (i.e columns that belong to a primary key). In order to bypass this restriction, we need to add `ALLOW FILTERING` to the query. This could have a potentially large performance impact: if the columns on the `WHERE` clause are not clustering columns and don't have any secondary indexes created, the query may be slow, particularly if the table has a lot of rows.

The following is quoted from [the ScyllaDB docs](https://opensource.docs.scylladb.com/stable/cql/dml/select.html#allowing-filtering): 
> Please note that a query using ALLOW FILTERING may thus have unpredictable performance (for the definition above), i.e. even a query that selects a handful of records may exhibit performance that depends on the total amount of data stored in the cluster.

src/fides/api/service/connectors/scylla_connector.py
```diff
@@ -38,41 +49,65 @@ def create_client(self) -> Cluster:
...
+    def query_config(self, node: ExecutionNode) -> ScyllaDBQueryConfig:
+        return ScyllaDBQueryConfig(node)
+
+    def set_schema(self, connection: Session) -> None:
+        """Optionally override to set the schema for a given cluster that
+        persists through the entire session"""
+        config = self.get_config()
+        if not config.keyspace:
+            raise ScyllaConnectorMissingKeyspace(
+                f"No keyspace provided in the ScyllaDB configuration for connector {self.configuration.name}"
```
galvana, line 66: We don't prompt the user for a name when configuring an integration so the `name` won't be available
image
It'd be more reliable to use `configuration.key`

galvana (commented):
Nice work figuring out all the ScyllaDB quirks! I would like to go over some of the base class refactors with you just to make sure I understand them completely. I left a handful of inline comments but here's some more general thoughts:

- I know this was added in a separate ticket but I think the `keyspace` param shouldn't be flagged as sensitive. If it's not needed for authenticating, I would argue that it'd be better from a supportability perspective if we we could see the current keyspace value when navigating to the integration tab.
- Thank you for the thorough testing notes! I was able to run the verification steps locally and everything worked as expected. You commented that you weren't sure how to verify the access results, you can view the DSR package (.zip file) if you navigate to the `fides_uploads` directory under your local `fides` repo.
image
image

src/fides/api/schemas/connection_configuration/connection_secrets_scylla.py
```diff
@@ -31,7 +31,7 @@ class ScyllaSchema(ConnectionConfigSecretsSchema):
     )
     keyspace: Optional[str] = Field(
         title="Keyspace",
-        description="The keyspace used.",
+        description="The keyspace used. If not provided, DSRs for this integration will error. ",
```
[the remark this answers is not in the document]
erosselli: Yeah, this was a bit annoying.  

So Scylla doesn't seem to have a "default" keyspace, unlike our SQL integrations (e.g in Postgres we have the `public` schema). So the following query 
`SELECT user_id FROM users WHERE email = 'test@test.com';` will work in a SQL integration but not in Scylla, cause there's no keyspace specified. The way to do this in Scylla is to either call `USE the_keyspace_name;` beforehand, or use the full name of the table in the query `SELECT user_id FROM the_keyspace_name.users WHERE email = 'test@test.com';`  So we NEED the user to tell us what the keyspace is.

Your guess is correct, for D & D we don't need to set a specific keyspace since it will just discover them. 

I think that ideally we wouldn't need to set the keyspace at the integration level (we had quite a long slack thread on this the other day, I'll slack it to you), since this is effectively preventing us from linking multiple schemas (or keyspaces in Scylla) to a single system integration. 

I'm happy to discuss more if you think there's a better way to solve this , and I agree about adding the D & D scenario in the description so it's clearer for users.

src/fides/api/service/connectors/query_config.py
```diff
@@ -352,86 +354,123 @@ def format_fields_for_query(
...
+    @abstractmethod
+    def format_clause_for_query(
+        self, string_path: str, operator: str, operand: str
     ) -> str:
-        """Returns a formatted SQL UPDATE statement to fit the Snowflake syntax."""
-        return f"UPDATE {self.node.address.collection} SET {','.join(update_clauses)} WHERE {' AND '.join(pk_clauses)}"
+        """Returs clause formatted in the specific SQL dialect for the query"""
 
-    def generate_query(
+    def generate_query_without_tuples(  # pylint: disable=R0914
```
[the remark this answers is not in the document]
erosselli: yes, I needed to reuse it but didn't want to mess with the existing inheritance of the classes, and I didn't want to inherit from `QueryStringWithoutTuplesOverrideQueryConfig` since that inherits directly from `SQLQueryConfig` which has things that don't make sense for Scylla

src/fides/api/service/connectors/scylla_query_config.py
```diff
@@ -0,0 +1,92 @@
...
+        self, t: ScyllaDBStatement, input_data: Dict[str, List[Any]]
+    ) -> str:
+        """string representation of a query for logging/dry-run"""
+
+        def transform_param(p: Any) -> str:
+            if isinstance(p, str):
+                return f"'{p}'"
+            return str(p)
+
+        query_str = str(t[0])
```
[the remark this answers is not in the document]
erosselli: oof yeah, this code's looking a bit rough haha tbh I just copied a lot of it from the SQL implementation and made the Scylla-related changes . I'll improve it

> Do we need to cast this to a string, since the type hint for ScyllaDBStatement already specifies this as a string?

You're right, I don't think we need the cast

src/fides/api/service/connectors/query_config.py
```diff
@@ -337,8 +337,10 @@ def generate_update_stmt(
         return None
 
 
-class SQLQueryConfig(QueryConfig[Executable]):
-    """Query config that translates parameters into SQL statements."""
+class SQLLikeQueryConfig(QueryConfig[T], ABC):
+    """
+    Abstract query config for SQL-like languages (that may not be strictly SQL).
+    """
```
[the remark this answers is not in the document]
erosselli: Yes, and also `generate_update_stmt` and `format_fields_for_query`. 
For clarification: 

- Both `format_fields_for_query` and `get_update_stmt` keep their implementations with no changes, they were just moved up from `SQLQueryConfig` into `SQLLikeQueryConfig`. 

- Both `generate_query_without_tuples` and `generate_update_stmt`  were also part of `SQLQueryConfig` and I kept almost the entire implementation, only changing a few lines that were integration-specific, replacing them with calls to some of these other abstract methods that each class implements. 

- These integration-specific methods are `format_query_data_name`, `get_formatted_query_string`, `get_update_clauses`, `format_query_stmt`, `format_clause_for_query` and  `format_key_map_for_update_stmt`.  
    - Most of these are needed because Scylla uses a different syntax for expressing parameters in a query: SQL databases, that we connect to with SQLAlchemy , use `:` to denote a parameter , e.g `SELECT user_id FROM users WHERE email = :email`, while Scylla uses `%()s` , e.g `SELECT user_id FROM users WHERE email = %(email)s`.
    - Another difference is in the fact that for Scylla queries, we need to pass in the query parameters to the `execute` method (called in the connector itself), so we return the query and the query params separately. Meanwhile for SQL queries , we call the SQLAlchemy `text` function and pass in the parameters to it, so the query config returns the query with the parameters already replaced

galvana (approved):
Thanks for researching these last few issues and making the changes!

PR #5145, opened 2024-07-31
Sailthru Integration: ignore valid 400 for user not existing

SteveDMurphy:
Closes FIDES-1082

### Description Of Changes

* Ignores 400 returned when a user cannot be found

### Code Changes

* [ ] Ignore the 400 for a user not found

### Steps to Confirm

* [ ] Tested with client in deployment

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`


tests/ops/integration_tests/saas/test_marigold_engage_task.py
```diff
@@ -21,6 +21,18 @@ async def test_access_request(
...
+        self,
+        marigold_engage_runner: ConnectorRunner,
+        policy,
+    ):
+        access_results = await marigold_engage_runner.access_request(
+            access_policy=policy,
+            identities={"email": "notfound@example.com"},
+            skip_collection_verification=True,
+        )
+        assert access_results == {"marigold_engage_instance:user": []}
```
galvana, line 34: Verifying the access request is successful even if the user is not found

tests/ops/integration_tests/saas/connector_runner.py
```diff
@@ -142,11 +143,13 @@ async def access_request(
             identities,
             self.db,
         )
-        # verify we returned at least one row for each collection in the dataset
-        for collection in self.dataset["collections"]:
-            assert len(
-                access_results[f"{fides_key}:{collection['name']}"]
-            ), f"No rows returned for collection '{collection['name']}'"
+
+        if not skip_collection_verification:
```
galvana, line 147: Small change to our connector runner to skip collection verification. In our case, the user collection will be empty but that should be allowed

src/fides/api/service/saas_request/override_implementations/marigold_engage_request_overrides.py
```diff
@@ -67,10 +67,12 @@ def marigold_engage_user_read(
                 method=HTTPMethod.GET,
                 path="/user",
                 query_params=signed_payload(secrets, payload),
-            )
+            ),
+            ignore_errors=[400],
```
galvana, line 71: Because we're using a `request_override` for the user read, we can't define `ignore_errors` on the SaaS config, it needs to be defined on the `client.send` inside the request override. It's possible for a request override function to make multiple `client.send` calls so if we define `ignore_errors` at the SaaS config level, we couldn't programmatically determine which `client.send` to apply that to.

src/fides/api/service/saas_request/override_implementations/marigold_engage_request_overrides.py
```diff
@@ -67,10 +67,12 @@ def marigold_engage_user_read(
...
         )
-        user = response.json()
-        output.append(user)
+        if response.ok:
+            user = response.json()
+            output.append(user)
```
Vagoasdf, line 76: Shouldn´t we add a comment or specific message in case the user is not found?

src/fides/api/service/saas_request/override_implementations/marigold_engage_request_overrides.py
Vagoasdf: We could define a variable that makes explicit for future programmers that we are ignoring the 400 error because the user might not exist, and also remove a magic number/list, by defining something like 

` error400ForUserNotFound = [400]`

Vagoasdf: ~~Also, Do we have an explanation from their docs on why an UserNotFound is a `400` error instead of a `404` ? Usually  NotFound errors are `404`~~. 
Nevermind, [found the reference on the Docs](https://getstarted.meetmarigold.com/engagebysailthru/Content/developers/api/user.html#ErrorCodes).   I'ts odd that they override the usual 404 error. 
Maybe we should check for the specific error message "User not found with {KEY}", just so we dont blank accept any 400 error

Vagoasdf (changes requested):
I have some doubts after double checking the documentation on [Error Codes for Marigold ](https://getstarted.meetmarigold.com/engagebysailthru/Content/developers/api/user.html#ErrorCodes). This implementation allows for all 400 errors to pass, but there might be a few of them, like *Missing required parameter* or *account not enabled for lookup* that we wouldn't want to ignore. Maybe we can check directly the error message, and allow/deny success acordding to that?

src/fides/api/service/saas_request/override_implementations/marigold_engage_request_overrides.py
```diff
@@ -62,15 +68,36 @@ def marigold_engage_user_read(
...
             )
         )
-        user = response.json()
-        output.append(user)
+
+        if response.ok and not error_msg:
+            output.append(response_json)
+        elif error_msg and error_msg.startswith("User not found with"):
+            context_logger.info(
+                f"Connector request successful. Ignoring {error_msg.lower()}"
```
Vagoasdf, line 94: This is better, logging the ignored message so  we can check it on the future

src/fides/api/service/saas_request/override_implementations/marigold_engage_request_overrides.py
Vagoasdf: Its better now. But i still think we should rename the variable, just so we have more clarity on what type of "400" error are we ignoring

Vagoasdf (changes requested):
Looks a lot better! Just small improvements in order to avoid 400 as a magic number

Vagoasdf (approved):

PR #5147, opened 2024-07-31
Improve API error messages when using `is_default` field on taxonomy resources

NevilleS:
### Description Of Changes

Today we throw a generic 403 error for a variety of edits which make it _look_ like the user is lacking sufficient permissions, but in reality it is that they are forbidden from editing the `is_default` field on a resource (e.g. `DataUse`, `DataCategory`).

#### Current Error Message
For example, we had a user report this unexpected error: they called the `/api/v1/data_category/upsert` API with this body:
```
[
    {
        "version_added": "2.0.0",
        "is_default": true,
        "fides_key": "user.unique_id.my_custom_category",
        "organization_fides_key": "default_organization",
        "name": "custom data category",
        "description": "description for custom data category",
        "parent_key": "user.unique_id",
        "active": true
    }
]
```

Which returns this error:
```
{ 'detail': { 'error': 'user does not have permission to modify',
              'fides_key': 'user.unique_id.my_custom_category',
              'resource_type': 'DataCategory'}}
```

#### New Error Message
As you can see above, the message is confusing - the error actually isn't that the user is lacking permission, it's that the user is trying to create a resource with `is_default: true`, which we never allow.

With this change, the error looks like this instead:
```
{ 'detail': { 'error': "cannot create a resource where 'is_default' is true",
              'fides_key': 'user.unique_id.my_custom_category',
              'resource_type': 'DataCategory'}}
```

There are some slight variations on those messages for deleting/editing/etc.

### Code Changes

* [X] Added a `ForbiddenIsDefaultTaxonomyError` error type with more user-friendly error messages
* [X] Updated `test_api.py` integration tests to assert on better error messages
* [X] Use the `ForbiddenIsDefaultTaxonomyError` error in the various code paths that were throwing generic 403s due to `is_default` rejections

### Steps to Confirm

* [X] Run integration tests

### Pre-Merge Checklist

* Documentation:
  * ~[ ] documentation complete, [PR opened in fidesdocs](https://github.com/ethyca/fidesdocs/pulls)~
  * ~[ ] documentation [issue created in fidesdocs](https://github.com/ethyca/fidesdocs/issues/new/choose)~
  * ~[ ] if there are any new client scopes created as part of the pull request, remember to update public-facing documentation that references our scope registry~
* [x] Update `CHANGELOG.md`
* ~[ ] For API changes, the [Postman collection](https://github.com/ethyca/fides/blob/main/docs/fides/docs/development/postman/Fides.postman_collection.json) has been updated~
  * ~[ ] Ensure that your downrev is up to date with the latest revision on `main`~
  * ~[ ] Ensure that your `downgrade()` migration is correct and works~
    * ~[ ] If a downgrade migration is not possible for this change, please call this out in the PR description!~

### DIFF
--- a/src/fides/api/util/errors.py
+++ b/src/fides/api/util/errors.py
@@ -98,11 +98,11 @@ class ForbiddenIsDefaultTaxonomyError(ForbiddenError):
         action: str = "modify",
         error_message: str = None,
     ) -> None:
-        error = error_message or f"cannot {action} a resource where 'is_default' is true"
+        error = (
+            error_message or f"cannot {action} a resource where 'is_default' is true"
+        )
         super().__init__(
-            resource_type=resource_type,
-            fides_key=fides_key,
-            error_message=error
+            resource_type=resource_type, fides_key=fides_key, error_message=error
         )
 
 
--- a/src/fides/api/util/errors.py
+++ b/src/fides/api/util/errors.py
@@ -72,13 +72,38 @@ class ForbiddenError(HTTPException):
     To be raised when a user cannot modify an entity.
     """
 
-    def __init__(self, resource_type: str, fides_key: str) -> None:
+    def __init__(
+        self,
+        resource_type: str,
+        fides_key: str,
+        error_message: str = "user does not have permission to modify",
+    ) -> None:
         detail = {
-            "error": "user does not have permission to modify",
+            "error": error_message,
             "resource_type": resource_type,
             "fides_key": fides_key,
         }
         super().__init__(status.HTTP_403_FORBIDDEN, detail=detail)
+
+
+class ForbiddenIsDefaultTaxonomyError(ForbiddenError):
+    """
+    To be raised when a user cannot modify a resource from the default Fideslang taxonomy (`is_default` is True)
+    """
+
+    def __init__(
+        self,
+        resource_type: str,
+        fides_key: str,
+        action: str = "modify",
+        error_message: str = None,
+    ) -> None:
+        error = error_message or f"cannot {action} a resource where 'is_default' is true"
+        super().__init__(
+            resource_type=resource_type,
+            fides_key=fides_key,
+            error_message=error
+        )
 
 
 def get_full_exception_name(exception: Exception) -> str:
--- a/tests/ctl/core/test_api.py
+++ b/tests/ctl/core/test_api.py
@@ -2938,8 +2938,6 @@ class TestDefaultTaxonomyCrud:
             resource_id=resource.fides_key,
             headers=test_config.user.auth_header,
         )
-        print(f"result object: {result}")
-        print(f"result json: {result.json()}")
         assert result.status_code == 403
         assert (
             "cannot modify a resource where 'is_default' is true"
--- a/tests/ctl/core/test_api.py
+++ b/tests/ctl/core/test_api.py
@@ -2941,7 +2941,10 @@ class TestDefaultTaxonomyCrud:
         print(f"result object: {result}")
         print(f"result json: {result.json()}")
         assert result.status_code == 403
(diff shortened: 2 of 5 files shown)


src/fides/api/util/errors.py
```diff
@@ -72,15 +72,40 @@ class ForbiddenError(HTTPException):
...
 
-    def __init__(self, resource_type: str, fides_key: str) -> None:
+    def __init__(
+        self,
+        resource_type: str,
+        fides_key: str,
+        error_message: str = "user does not have permission to modify",
+    ) -> None:
         detail = {
-            "error": "user does not have permission to modify",
```
NevilleS, line 77: This was the main issue - this error message couldn't be overridden with anything more relevant to more specific types of 403 errors.

src/fides/api/util/errors.py
```diff
@@ -72,15 +72,40 @@ class ForbiddenError(HTTPException):
...
+            "error": error_message,
             "resource_type": resource_type,
             "fides_key": fides_key,
         }
         super().__init__(status.HTTP_403_FORBIDDEN, detail=detail)
 
 
+class ForbiddenIsDefaultTaxonomyError(ForbiddenError):
```
NevilleS, line 89: Added this as an explicit way to encapsulate these types of "is default" logic errors and provide a more useful message.

src/fides/api/util/errors.py
```diff
@@ -72,15 +72,40 @@ class ForbiddenError(HTTPException):
...
+        self,
+        resource_type: str,
+        fides_key: str,
+        action: str = "modify",
+        error_message: str = None,
+    ) -> None:
+        error = (
+            error_message or f"cannot {action} a resource where 'is_default' is true"
+        )
+        super().__init__(
```
adamsachs, line 104: nice OOP :)

src/fides/api/util/errors.py
```diff
@@ -72,15 +72,40 @@ class ForbiddenError(HTTPException):
...
+class ForbiddenIsDefaultTaxonomyError(ForbiddenError):
+    """
+    To be raised when a user cannot modify a resource from the default Fideslang taxonomy (`is_default` is True)
+    """
+
+    def __init__(
+        self,
+        resource_type: str,
+        fides_key: str,
+        action: str = "modify",
```
NevilleS, line 98: This overcomplicates it a bit so that the error message says either "cannot modify", "cannot create", "cannot delete", etc.. I only did this as a bit of Python practice, to be honest.

src/fides/api/util/errors.py
```diff
@@ -72,15 +72,40 @@ class ForbiddenError(HTTPException):
...
+        error_message: str = None,
+    ) -> None:
+        error = (
+            error_message or f"cannot {action} a resource where 'is_default' is true"
```
NevilleS, line 102: The `error_message or <template string>` allows providing an explicit `error_message` argument here. I use that in one place where the error is actually more subtle- we prevent the ability to modify `is_default` on a resource that already exists.

tests/ctl/core/test_api.py
```diff
@@ -2938,7 +2938,13 @@ def test_api_cannot_delete_default(
             resource_id=resource.fides_key,
             headers=test_config.user.auth_header,
         )
+        print(f"result object: {result}")
+        print(f"result json: {result.json()}")
```
NevilleS, line 2942: Doh, should have removed these. Going to wait for the tests to finish running in CI, then will delete.

NevilleS (commented):
Quick self-review

adamsachs (approved):
lgtm! clean impl of a nice, targeted improvement 👍

src/fides/api/util/errors.py
NevilleS: OK good, thanks. This was a bit of Python practice 👍

PR #5153, opened 2024-08-04
Refactoring ConsentableItem to simplify Pydantic v2 migration

galvana:
### Description Of Changes

Updating the `ConsentableItem` schema to use `external_id` instead of `id` to remove the need for a custom `from_orm` function. Using `external_id` allows us to just map the SQLAlchemy model's `external_id` to the Pydantic model's `external_id` without needing to map `external_id` to `id`.

### Code Changes

* [ ] Updating `ConsentableItem` and removing the custom `from_orm`

### Steps to Confirm

* [ ] Tests should pass

### Pre-Merge Checklist

### DIFF
--- a/src/fides/api/schemas/consentable_item.py
+++ b/src/fides/api/schemas/consentable_item.py
@@ -1,37 +1,21 @@
-from typing import List, Optional, Type
+from typing import List, Optional
 
-from pydantic import BaseModel, Field
+from pydantic import Field
 
-from fides.api.models.consent_automation import ConsentableItem as ConsentableItemModel
+from fides.api.schemas.base_class import FidesSchema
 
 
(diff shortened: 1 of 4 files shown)


src/fides/api/schemas/consentable_item.py
```diff
@@ -1,38 +1,22 @@
...
 
-    id: str
+    external_id: str
     type: str
     name: str
     notice_id: Optional[str] = None
     children: List["ConsentableItem"] = Field(default_factory=list)
     unmapped: Optional[bool] = False
 
-    @classmethod
```
eastandwestwind, line 20: nice that we can get rid of this!

eastandwestwind (approved):

PR #5164, opened 2024-08-07
PROD-2345 - Implement notice-based consent for SaaS connector

eastandwestwind:
Closes https://ethyca.atlassian.net/browse/PROD-2345

### Description Of Changes

Implements the ability for SaaS connectors to handle notice-based consent. 

When we detect that a specific SaaS Connector has a registered `update_consent`fn, we follow this notice-based flow, where we send a map of notice -> user preference to the connected service. 

Before, we only supported opt-in / opt-out consent for SaaS connectors, in which we only were able to send one user preference to each configured fn in the yml.

The result of this PR is the internal structure to support this notice-based consent flow, but we require at least 1 SaaS implementation of the `update_consent` fn to fully e2e test. For now we rely on unit tests, but we'll want to test e2e once https://ethyca.atlassian.net/browse/PROD-2484 is complete.

### Code Changes

* [ ] Implement notice-based consent for base SaaS connector
* [ ] Adds tests

### Steps to Confirm

* [ ] Confirm tests pass

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`


src/fides/api/schemas/consentable_item.py
```diff
@@ -60,3 +61,17 @@ def merge_consentable_items_recursive(
         merge_consentable_items_recursive(api_item, target_item)
 
     return api_items
+
+
+def build_consent_item_hierarchy(
```
eastandwestwind, line 66: this fn lives in Fidesplus, but we need it in Fides

src/fides/api/service/connectors/saas_connector.py
```diff
@@ -615,86 +633,155 @@ def run_consent_request(
...
+            SaaSRequestOverrideFactory.get_override(
+                saas_config.type, SaaSRequestType.UPDATE_CONSENT
+            )
         )
 
-        if should_opt_in is None:
-            logger.info(
-                "Skipping consent requests on node {}: No actionable consent preferences to propagate",
-                node.address.value,
+        if notice_based_override_fn:
```
eastandwestwind, line 653: here, we're saying if we have registered a notice_based override, we follow the notice-based flow

src/fides/api/util/consent_util.py
```diff
@@ -54,9 +58,15 @@ def filter_privacy_preferences_for_propagation(
     return filtered_on_use
 
 
-def should_opt_in_to_service(
```
eastandwestwind, line 57: renamed this existing fn to better reflect what it returns, which is a Tuple of:
1. Some kind of consent preference(s), for opt-in/out this is just True or False, for notice-based this is a mapping
2. The complete list of preferences that apply for sake of reporting

src/fides/api/util/consent_util.py
```diff
@@ -85,15 +95,39 @@ def should_opt_in_to_service(
...
         return None, []  # We should do nothing here
 
-    # Collapse relevant preferences into whether we should opt-in or opt-out
+    filtered_preferences: List[PrivacyPreferenceHistory] = []
+
+    # 1. NOTICE-BASED WORKFLOW
+    if is_notice_based:
+        notice_id_to_preference_map: Dict[str, UserConsentPreference] = {}
+        # retrieve notices from the DB if they are associated with a relevant preference
+        notices_with_preference: Query = session.query(PrivacyNotice).filter(
```
eastandwestwind, line 114: This chunk of work attempts to build our notice map and preferences in the most efficient way possible, in which we only need to make one DB query instead of multiple.

tests/ops/service/connectors/test_saas_connector.py
```diff
@@ -878,6 +904,44 @@ def test_preferences_executable(
...
+        name = iterable_connection_config_no_secrets.saas_config["type"]
+        register(name, SaaSRequestType.UPDATE_CONSENT)(valid_consent_update_override)
+        assert valid_consent_update_override == SaaSRequestOverrideFactory.get_override(
+            name, SaaSRequestType.UPDATE_CONSENT
+        )
+        privacy_preference_history.privacy_request_id = (
+            privacy_request_with_consent_policy.id
+        )
+        # todo- implement manually getting this key, remove this placeholder in test
+        privacy_preference_history.notice_key = "example_privacy_notice_1"
```
eastandwestwind, line 924: Todo- need to confirm notice_key should not be auto-generated on the PrivacyPreferenceNoticeHistory model when a row on `PrivacyPreferenceNoticeHistory` is initially created.

eastandwestwind: If we cannot expect it to be auto-generated, then we'll need to implement additional code to get notice_key based on the privacy preference history id. 

To do this:

* `privacy_notice_history_id` on `PrivacyPreferenceHistory` matches `privacy_notice_history_id` on one (or more?) of `privacy_notice.translations`

tests/ops/service/connectors/test_saas_connector.py
```diff
@@ -878,6 +904,44 @@ def test_preferences_executable(
...
+        execution_node = traversal_node.to_mock_execution_node()
+        connector.run_consent_request(
+            node=execution_node,
+            policy=consent_policy,
+            privacy_request=privacy_request_with_consent_policy,
+            identity_data={"ljt_readerID": "abcde"},
+            request_task=request_task,
+            session=db,
+        )
+        # todo- how to assert override fn was called?
```
eastandwestwind, line 939: Open question here

src/fides/api/task/graph_task.py
```diff
@@ -815,17 +815,18 @@ def build_consent_dataset_graph(datasets: List[DatasetConfig]) -> DatasetGraph:
...
+        connection_config: ConnectionConfig = dataset_config.connection_config
+
+        if connection_config.connection_type != ConnectionType.saas:
+            continue
+
+        saas_config = connection_config.get_saas_config()
+        if not saas_config:
+            continue
+
+        if ActionType.consent in saas_config.supported_actions:
```
eastandwestwind, line 827: Thanks for updating this @galvana ! From what I understand, our check that `saas_config.get("consent_requests")` exists was too restrictive, because for bi-directional / notice-based consent, we won't have any consent requests defined in the yml, and instead we rely on the `supported_actions` which is based on whether or not we have that `CONSENT_UPDATE` override fn defined. Is that correct?

tests/ops/service/connectors/test_saas_connector.py
eastandwestwind: update- from @galvana: 

> here's where the notice_key is set, it's in Fidesplus https://github.com/ethyca/fidesplus/blob/main/src/fidesplus/api/preferences/preferences_util.py#L444 but because this is a Fidesplus workflow, we can assume it will be there for us

eastandwestwind: We should add a test in Fidesplus to assert that the `privacy_notice_history_id` val is set when notices served endpoint is hit, comment above it's for bi-directional consent

src/fides/api/service/connectors/saas_connector.py
```diff
@@ -861,23 +972,31 @@ def _invoke_masking_request_override(
 
     @staticmethod
     def _invoke_consent_request_override(
+        override_function: RequestOverrideFunction,
         override_function_name: str,
```
galvana, line 976: In my previous implementation, I passed in `override_function_name` because I performed the function lookup inside `_invoke_consent_request_override`. Now, we look up the function outside because we need it to determine which flow to use. I think in our new use case we can remove this param, and just call `override_function.__name__` if we want to use the function name in the logs. Better to pull the name directly from the function.

src/fides/api/service/connectors/saas_connector.py
```diff
@@ -621,80 +665,147 @@ def run_consent_request(
...
+            fired = self._invoke_consent_request_override(
+                notice_based_override_fn,
+                saas_config.type,
+                self.create_client(),
+                policy,
+                privacy_request,
+                self.secrets,
+                identity_data,
+                notice_id_to_preference_map,  # type: ignore[arg-type]
+                self.build_notice_based_consentable_item_hierarchy(
```
galvana, line 711: We should move this check earlier and avoid calling the `consent_update` altogether if the consentable items are `None`. I would treat this the same as not having any consent preferences.

src/fides/api/service/connectors/saas_connector.py
```diff
@@ -621,80 +665,147 @@ def run_consent_request(
...
+                )
+                raise SkippingConsentPropagation(
+                    f"Skipping consent propagation for node {node.address.value} - no actionable consent preferences to propagate"
+                )
+            cache_initial_status_and_identities_for_consent_reporting(
+                db=session,
+                privacy_request=privacy_request,
+                connection_config=self.configuration,
+                relevant_preferences=filtered_preferences,
+                relevant_user_identities=identity_data,
```
galvana, line 700: The old flow used the consent requests defined in the SaaS config to determine the "relevant user identities". We should look into the importance of using a filtered list versus all of the identities. If we need to filter this list down, maybe we can add another function so that each integration can return which identities are relevant.

galvana (commented):
Thanks for taking on such a core part of bidirectional consent 💪! I left some minor recommendations, let me know that you think. Also, a more general nit, would you be open to changing instances of `fn` with `function`?

src/fides/api/service/connectors/saas_connector.py
```diff
@@ -621,80 +665,147 @@ def run_consent_request(
...
+                    fired = self._invoke_consent_request_override(
+                        override_function,
+                        consent_request.request_override,
+                        self.create_client(),
+                        policy,
+                        privacy_request,
+                        self.secrets,
+                        None,
+                        None,
+                        None,
```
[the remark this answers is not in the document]
eastandwestwind: so much better!

src/fides/api/service/connectors/saas_connector.py
eastandwestwind: good call!

src/fides/api/service/connectors/saas_connector.py
eastandwestwind: Right, we'd determined that for now we needed all identities. But yes, how do we determine on the integration side which specific identity we need for any given request? I think we can definitely follow-up with work to expose a "relevant_identities" fn per integration if that's needed...

galvana (approved):
Thanks for making the changes! This looks good to go 👍

PR #5173, opened 2024-08-09
add git tag version to window.Fides

gilluminate:
Closes [PROD-1235](https://ethyca.atlassian.net/browse/PROD-1235)

### Description Of Changes

To aid in troubleshooting, debugging, and all-around code hygiene, we'll now include a version number in the built FidesJS package like so:
```window.Fides.version = "2.22.0";```

### Code Changes

* add new `git-tag-version` npm package
* during rollup.js build, replace `__FIDES_JS_VERSION_NUMBER__` in the scripts with the currently tagged version.

### Steps to Confirm

* Run the demo site (http://localhost:3001/fides-js-demo.html)
* In the browser console, type `Fides.version` and press return
* validate that the current tagged version is returned

### Pre-Merge Checklist

* [ ] All CI Pipelines Succeeded
* [ ] Issue Requirements are Met
* [ ] Update `CHANGELOG.md`

[PROD-1235]: https://ethyca.atlassian.net/browse/PROD-1235?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #5185, opened 2024-08-13
Revert "add git tag version to window.Fides (#5173)"

gilluminate:
This reverts commit 57be06ee007505de17d6700c1adb8395db261482.


NevilleS (approved):
We'll find a new way!

PR #5186, opened 2024-08-13
Fix various issues with Consentable Items form

gilluminate:
Closes [PROD-2344](https://ethyca.atlassian.net/browse/PROD-2344)

### Description Of Changes

* There was an extra forward-slash in the endpoint URL causing problems
* tie the save button on the Consentable Items form to the `isLoading` flag so we can get the spinner to show up
* remove "default notices" text since we ended up not adding that feature


galvana (approved):
Verified locally ✅

PR #5212, opened 2024-08-19
Classifier category badge updates

jpople:
Closes PROD-2546

### Description Of Changes

Visual update to data category badges and status badges in discovery field tables and tweaks to behavior.

Old:

Screenshot 2024-08-19 at 11 51 22

New:

Screenshot 2024-08-19 at 11 51 32

Behavior changes:

* Status now shows "reviewed" instead of "pending review" when user has assigned any data categories
* Classifier-assigned data categories now show a sparkle icon in their badge
* Add category button now appears when category is "None", except on nested fields (the server 400s when trying to update a category on a field with subfields, as only terminal-level fields are supposed to have data categories)

### Code Changes

* Refactor result status badge and classification badge components

### Pre-Merge Checklist

* [ ] Update `CHANGELOG.md`


lucanovera (approved):
Looks much cleaner now. I found no bugs with the UI. I think it's a bit odd that when you delete a category it goes back to the classification result, but I guess that's helpful if you accidentally change the category and want to go back. I wasn't able to test the None state because of that, but I see that you've added a test to cover that.

Everything looks good. Just needs the changelog update. I will leave the approve now.

jpople:
> I think it's a bit odd that when you delete a category it goes back to the classification result, but I guess that's helpful if you accidentally change the category and want to go back. I wasn't able to test the None state because of that, but I see that you've added a test to cover that.

Yeah, this behavior is a bit strange but I talked it over with Adam, Kirk and Jack last week and it seems like there's no way around it for now without extra backend changes, and we don't ever want to fully get rid of the classifier data categories.  The None state where the field doesn't have subfields _shouldn't_ ever happen TMK, just wanted it covered in case I'm failing to think of something.

PR #5218, opened 2024-08-20
Add "data type" column to D&D tables

jpople:
Closes PROD-2379

### Description Of Changes

Adds a "data type" column to fields in D&D tables.

Screenshot 2024-08-20 at 11 10 50

### Code Changes

* [ ] Regenerate types
* [ ] Refactor D&D table cells; collect them all in one directory and change some names for consistency

### Steps to Confirm

Running fidesplus backend on branch `https://github.com/ethyca/fidesplus/pull/1567/files`, "Data type" column should appear on fields in D&D result tables.

### Pre-Merge Checklist

* [ ] Update `CHANGELOG.md`


lucanovera (commented):
Nice job organizing the files. I see the data type column correctly in the Data Detection portion. 
Captura de pantalla 2024-08-20 a la(s) 3 23 30 p  m

But no values appear in the Data Discovery column, looking at the response from the API I see that the source_data_type is null but the data_type says string. So I wonder if this is a backend issue or is it expected to get it as data_type and that should be the fallback value to source_data type in the FE.

Also, I saw you removed the "--" fallback for _description_ in the useDetectionResultColumns on line 124 but it is still there in other parts of the file and in useDiscoveryResultColumns too, we'll probably want to keep a consistency there.

jpople:
@lucanovera 

> I saw you removed the "--" fallback for description in the useDetectionResultColumns on line 124 but it is still there in other parts of the file and in useDiscoveryResultColumns too, we'll probably want to keep a consistency there.

Thanks for catching this, fixed.

> But no values appear in the Data Discovery column, looking at the response from the API I see that the source_data_type is null but the data_type says string. So I wonder if this is a backend issue or is it expected to get it as data_type and that should be the fallback value to source_data type in the FE.

This was a backend and frontend issue-- I was querying the wrong field and the backend was returning null for the right field (`source_data_type` is correct).  It should be resolved, try pulling the latest and you should see `source_data_type` populated in both detection and discovery tables.

lucanovera:
> @lucanovera
> 
> > I saw you removed the "--" fallback for description in the useDetectionResultColumns on line 124 but it is still there in other parts of the file and in useDiscoveryResultColumns too, we'll probably want to keep a consistency there.
> 
> Thanks for catching this, fixed.
> 
> > But no values appear in the Data Discovery column, looking at the response from the API I see that the source_data_type is null but the data_type says string. So I wonder if this is a backend issue or is it expected to get it as data_type and that should be the fallback value to source_data type in the FE.
> 
> This was a backend and frontend issue-- I was querying the wrong field and the backend was returning null for the right field (`source_data_type` is correct). It should be resolved, try pulling the latest and you should see `source_data_type` populated in both detection and discovery tables.

Thanks for the updates. I just tested with the latest from BE and FE branches and I see the data types correctly.

lucanovera (approved):
Approved!

PR #5226, opened 2024-08-22
PROD-2486 Add Dynamic Erasure Email Integration

erosselli:
Closes [PROD-2486](https://ethyca.atlassian.net/browse/PROD-2486)

#### Main idea

This introduces a new type of integration called Dynamic Erasure Email integration. This integration is very similar to our existing Generic Erasure Email integration, with the key difference being that the recipient email address is not set as part of the connection config, but rather calculated at runtime using the privacy request's custom request fields. The connection config has an `recipient_email_address` field that is a _reference_  to the dataset field that contains the email address. 

For example, let's suppose we have the following table `tenants` :

| tenant_id    | email |
| -------- | ------- |
| 123  | tenant@example.com   |
| 456 | a-different-tenant@example.com    |

The idea behind this integration is that we set the `recipient_email_address` field to point to the `email` field of the `tenants` collection, e.g `tenants_dataset.tenants.email` , and we provide a `tenant_id` as part of the erasure request, which will then be used to look up the corresponding email address. 

The `third_party_vendor_name` is also a dataset reference field, since the vendor name will depend on the email address.

#### Relevant changes

- We added a new field to the `fides_meta` of the field, called `custom_request_field`. This depends on https://github.com/ethyca/fideslang/pull/13
- We relaxed the graph traversability requirement so that unreachable node that have at least one field with the `custom_request_field` attribute are ignored by the reachability check 
- Created the new connector type 

#### Next Steps
Ideally we want to fully support custom request fields as part of the DSR graph. Once we implement that, a lot of the hacky workarounds of this PR won't be necessary. The idea here was to release this within 1 sprint and have it targeted to a customer request, but we do plan to fully support this feature in the future.

### Code Changes

* Add a new connection type `dynamic_email_erasure` and create a new `DynamicErasureEmailConnector`  
* Move some logic from `GenericErasureEmailConnector` into a new base class `BaseErasureEmailConnector`, that has logic common to both `GenericErasureEmailConnector` and `DynamicErasureEmailConnector`. They both inherit from this class.
* Make some changes in the graph building logic to ignore unreachable nodes if they have custom request fields. 
* Update `SQLConnector` and `SnowflakeConnector` with a new `execute_standalone_retrieval_query`  that allows us to execute a query without caring about the incoming/outgoing edges of the node. This is used to execute the query to retrieve the email address using the provided custom fields

Related Docs PR: https://github.com/ethyca/fidesdocs/pull/413/files

#### Environment setup
1. Make sure the following is set in the sample_project `fides.toml` file in the `[execution]` section:
```
allow_custom_privacy_request_field_collection = true
allow_custom_privacy_request_fields_in_request_execution = true
```
2. Edit the `initiate_scheduled_batch_email_send` function in `src/fides/api/service/privacy_request/email_batch_service.py` so the job runs more than once a week, e.g every 30s:
```
scheduler.add_job(
        func=send_email_batch,
        kwargs={},
        id=BATCH_EMAIL_SEND,
        coalesce=False,
        replace_existing=True,
        trigger="cron",
        second="30",
    )
```
3. Run the sample project, `nox -s "fides_env(test)"`
4. Open the `postgres_example` database in the SQL client of your choice (or open a psql shell in the container) and edit the second row in the `dynamic_email_address_config` table so that the `email_address` column contains your email address. This is so you'll actually receive the emails. 
5. Open the Swagger API (or your API Postman collection if you prefer), and call the `PUT /api/v1/messaging/default` [endpoint](http://localhost:8080/docs#/Messaging/put_default_config_api_v1_messaging_default_put) with the following payload:
```
{
  "service_type": "mailgun",
  "details": {
    "is_eu_domain": false,
    "api_version": "v3",
    "domain": <the-domain>
  }
}
```
You can get the domain from our 1password mailgun test credentials.
6. From the Swagger API (or Postman) , call the `PUT /api/v1/messaging/default/{service_type}/secret` [endpoint](http://localhost:8080/docs#/Messaging/put_default_config_secrets_api_v1_messaging_default__service_type__secret_put), with `service_type` set to `mailgun`. You can get the api key from the same 1password credentials as before.

#### Testing the integration
1. Create an erasure privacy request through the Swagger API (or Postman) using the `POST /api/v1/privacy-request` [endpoint](http://localhost:8080/docs#/Privacy%20Requests/create_privacy_request_api_v1_privacy_request_post) with the following payload:
```
[
  {
    "identity": {
      "email": "jane@example.com"
    },
    "custom_privacy_request_fields": {
      "tenant_id": {
        "label": "Tenant Id",
        "value": "site-id-2"
      }
    },
    "policy_key": "default_erasure_policy"
  }
]
```
2. Go to the Admin UI and approve the privacy request. In about 1-2 min, you should receive the erasure request email from mailgun in your inbox
    2.1  The setup for this to work is done in the sample project environment, but you can go to the Systems list in the Admin UI and see for yourself. There's a system for the new Dynamic Erasure Email Integration, which has the dataset reference that points to the Postgres Integration for the `dynamic_email_address_config` collection.  
    
#### Testing error cases

**Creating the integration from scratch**
1. Create a new system and go to its `Integrations` tab. Click `Dynamic Erasure Email`. 
2. When filling out the integration config, try to write invalid dataset references. Some cases to test, and their expected outcomes:
    2.1 Reference is not dot-delimited, e.g `asingleword` => frontend form shows an error specifying correct format
    2.2 Reference is dot-delimited, but only has 2 parts, e.g `collection.field` => frontend form shows an error
    2.3 Reference field references a fake dataset, e.g `fake_dataset.collection.field` => backend returns error which is displayed on the frontend
    2.4 `recipient_email_address` and `third_party_vendor_name` reference different datasets => backend returns error which is displayed on the frontend
    2.5 `recipient_email_address` and `third_party_vendor_name` reference different collections from the same dataset => backend returns error which is displayed on the frontend

**Running a misconfigured integration**
1. To simulate a misconfigured integration, open the `datasetconfig` table and edit the `fides_key` of the `postgres_example_custom_request_field_dataset` row, e.g change it to `test`. 
2. Create the privacy request as before, and approve it for the admin.
3. Wait for the task to run, and you should see the privacy request with an error status, as well as an `ExecutionLog` in the `Activity Timeline` section that shows the error detail

**Privacy request with no email lookup results**
Creating and approving a privacy request with the following payload
```
[
  {
    "identity": {
      "email": "jane@example.com"
    },
    "custom_privacy_request_fields": {
      "tenant_id": {
        "label": "Tenant Id",
        "value": "site-id-5"
      }
    },
    "policy_key": "default_erasure_policy"
  }
]
```
should cause the privacy request to error, and an ExecutionLog with the error details added. 

**Privacy request with multiple results for email lookup**
Creating and approving a privacy request with the following payload
```
[
  {
    "identity": {
      "email": "jane@example.com"
    },
    "custom_privacy_request_fields": {
      "tenant_id": {
        "label": "Tenant Id",
        "value": "site-id-multiple-emails"
      }
    },
    "policy_key": "default_erasure_policy"
  }
]
```
should cause the privacy request to error, and an ExecutionLog with the error details added. 

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

[PROD-2486]: https://ethyca.atlassian.net/browse/PROD-2486?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.fides/fides.toml
```diff
@@ -51,6 +51,8 @@ subject_identity_verification_required = false
 task_retry_count = 0
 task_retry_delay = 1
 use_dsr_3_0 = false
+allow_custom_privacy_request_field_collection = true
+allow_custom_privacy_request_fields_in_request_execution = true
```
erosselli, line 55: temporary change so we enable custom request field collection -- I'll undo before merging

clients/admin-ui/src/features/datastore-connections/system_portal_config/forms/ConnectorParametersForm.tsx
```diff
@@ -235,18 +235,23 @@ export const ConnectorParametersForm = ({
         connectionConfig.enabled_actions || []
       ).map((action) => action.toString());
 
-      // @ts-ignore
-      initialValues.secrets = connectionConfig.secrets ?? {};
+      initialValues.secrets = connectionConfig.secrets
+        ? { ...connectionConfig.secrets }
+        : {};
```
erosselli, line 240: the frontend was exploding when trying to process the fides dataset reference. I suspect the underlying issue was there from before and we simply weren't using this? anyways this is a quick & dirty fix (same with the fix below) to get it to work for now, I'm happy to create a ticket to address this more thoroughly later

clients/admin-ui/src/features/datastore-connections/system_portal_config/forms/ConnectorParametersForm.tsx
```diff
@@ -256,7 +261,7 @@ export const ConnectorParametersForm = ({
 
     if (_.isEmpty(initialValues.enabled_actions)) {
       initialValues.enabled_actions = connectionOption.supported_actions.map(
-        (action) => action.toString(),
+        (action) => action.toString()
```
erosselli, line 264: woops, it seems my frontend setup is not quite working ok. TODO:  undo these changs

clients/admin-ui/src/types/api/models/ConnectionConfigurationResponse.ts
```diff
@@ -22,7 +22,7 @@ export type ConnectionConfigurationResponse = {
   last_test_timestamp?: string | null;
   last_test_succeeded?: boolean | null;
   saas_config?: SaaSConfigBase | null;
-  secrets?: null;
+  secrets?: any;
```
erosselli, line 25: same here, this was causing some issues so I reverted it back to any as a quick workaround

data/dataset/postgres_example_custom_request_field_dataset.yml
```diff
@@ -0,0 +1,22 @@
...
+              primary_key: True
+          - name: email_address
+            data_categories: [system.operations]
+            fides_meta:
+              data_type: string
+          - name: custom_field
+            data_categories: [system.operations]
+            fides_meta:
+              data_type: string
+              custom_request_field: custom_field
```
erosselli, line 22: I wonder if it makes more sense for `custom_request_field` to be a boolean? Right now I'm not using its value. I think we have two options here: 
1. Make custom_request_field a boolean. In this case the name of the DSR field is expected to match the name of the field in the dataset
2. Keep the string, which will represent the name of the custom field in the DSR (potentially different than the name of the field in the dataset) , and then map the DSR value to the field value

src/fides/api/graph/config.py
```diff
@@ -493,6 +497,13 @@ def identities(self) -> Dict[FieldPath, str]:
             if field.identity
         }
 
+    def custom_request_fields(self) -> Dict[FieldPath, str]:
```
erosselli, line 500: utility to know whether a collection has custom_request_fields. I followed the same format as the `identities` function to try to keep it consistent

src/fides/api/graph/traversal.py
```diff
@@ -380,6 +380,18 @@ def edge_ends_with_collection(_edge: Edge) -> bool:
...
+        # if these are reachable or not. They will be used independently
+        # by the Dynamic Email Erasure Connector.
+        # TODO: ideally we'll update the Traversal code to include these
+        # "custom request field datasets" as part of the main graph. This is
+        # a targeted workaround for now.
+        remaining_node_keys = [
+            key
+            for key in remaining_node_keys
+            if not self.traversal_node_dict[key].node.collection.custom_request_fields()
+        ]
```
erosselli, line 393: We take the `remaining_node_keys` and filter out the nodes that have custom_request_fields , so we don't error because they're unreachable. This was a workaround proposed by @galvana so that we could deliver this feature in time, but in the future we'd want to incorporate custom request fields into the graph building logic

src/fides/api/models/privacy_request.py
```diff
@@ -1743,6 +1743,16 @@ class TraversalDetails(FidesSchema):
     input_keys: List[str]
 
 
+    @classmethod
+    def create_empty_traversal(cls, connection_key: str):
```
erosselli, line 1747: because the custom_request_fields datasets aren't necessarily reachable, they don't have a traversal node, so we need this "empty traversal" as a placeholder in some places

src/fides/api/schemas/connection_configuration/connection_secrets_dynamic_erasure_email.py
```diff
@@ -0,0 +1,27 @@
...
+from fides.api.schemas.connection_configuration.connection_secrets_email import (
+    EmailSchema,
+)
+
+
+class DynamicErasureEmailSchema(EmailSchema):
+    """Schema to validate the secrets needed for a dynamic erasure email connector"""
+
+    third_party_vendor_name: str
+    recipient_email_address: FidesDatasetReference = FieldInfo(
```
erosselli, line 17: maybe we should rename this to `recipient_email_addres_field` or something, cause it's not an email address anymore

src/fides/api/service/connectors/base_connector.py
```diff
@@ -115,3 +115,20 @@ def dry_run_query(self, node: ExecutionNode) -> Optional[str]:
     @abstractmethod
     def close(self) -> None:
         """Close any held resources"""
+
+    def execute_standalone_retrieval_query(
+        self, node: ExecutionNode, fields: List[str], filters: Dict[str, List[Any]]
+    ) -> List[Row]:
```
erosselli, line 121: In order to be able to execute a connector query that doesn't check the nodes or incoming or outgoing edges, we define this new method to execute a "standalone" query. For now we're only implementing it for SQLConnector and Snowflake (client that requested this feature uses Snowflake and SQL connector seems to cover many cases).  This was to limit the scope of this PR

src/fides/api/service/connectors/dynamic_erasure_email_connector.py
```diff
@@ -0,0 +1,258 @@
...
+
+        Returns the found email address, or None if none were found. Expects that the query using
+        the provided custom request fields returns 1 single row, otherwise an error is logged and
+        None is returned.
+        """
+        # The following code is a bit hacky, but we need to instantiate a RequestTask
+        # and an Execution node in order to execute a standalone retrieval query on the connector.
+        # The only fields we really care about in the RequestTask are those related to the collection
+        # and to the dataset itself.
+
```
erosselli, line 66: note to self: I think I probably need to add some exception handling in this function

src/fides/api/service/connectors/dynamic_erasure_email_connector.py
```diff
@@ -0,0 +1,258 @@
...
+class DynamicErasureEmailConnector(GenericErasureEmailConnector):
+    """Email Erasure Connector that performs a lookup for the email to use."""
+
+    config_schema = DynamicErasureEmailSchema
+
+    def __init__(self, configuration: ConnectionConfig):
+        super().__init__(configuration)
+
+    def get_email_address_from_custom_request_fields(
```
erosselli, line 45: hopefully the docstrings and comments make this function self-explanatory, but this is where the main logic of this feature lives. We use the graph_dataset generated from the dataset reference to get the relevant connector and execute a standalone query against it in order to retireve the email address using the custom fields of the privacy request

src/fides/api/service/connectors/dynamic_erasure_email_connector.py
```diff
@@ -0,0 +1,258 @@
...
+        )
+
+        # Instatiate the ConnectionConfig's connector so that we can execute the query
+        # to retrieve the email addresses based on the custom request fields
+        connector = get_connector(connection_config)
+
+        skipped_privacy_requests: List[str] = []
+        # We'll batch identities for each email address
+        # so we'll only send 1 email to each email address
+        batched_identities: Dict[str, List[str]] = {}
```
erosselli, line 191: in the generic erasure email connector, batched_identities is a list. Here we have a dictionary that maps the recipient email address to its identities

src/fides/api/service/connectors/dynamic_erasure_email_connector.py
```diff
@@ -0,0 +1,258 @@
...
+
+        for email_address, identities in batched_identities.items():
+            try:
+                send_single_erasure_email(
+                    db=db,
+                    subject_email=email_address,
+                    subject_name=self.config.third_party_vendor_name,
+                    batch_identities=identities,
+                    test_mode=False,
+                )
```
erosselli, line 239: Instead of sending one single erasure email, we send one email per recipient email address

src/fides/api/service/connectors/query_config.py
```diff
@@ -519,27 +547,30 @@ def generate_query(
...
+            #             self.format_clause_for_query(string_path, "=", string_path)
+            #         )
+            #         query_data[string_path] = (data.pop(),)
+            #     elif len(data) > 1:
+            #         clauses.append(
+            #             self.format_clause_for_query(string_path, "IN", string_path)
+            #         )
+            #         query_data[string_path] = tuple(data)
+            # if len(clauses) > 0:
+            #     query_str = self.get_formatted_query_string(field_list, clauses)
```
erosselli, line 572: I wanted to reuse this, will remove commented code once I test that the regular `generate_query` method is still working as usual

src/fides/api/service/connectors/query_config.py
```diff
@@ -627,6 +658,48 @@ class QueryStringWithoutTuplesOverrideQueryConfig(SQLQueryConfig):
     Generates SQL valid for connectors that require the query string to be built without tuples.
     """
 
+    # Overrides SQLQueryConfig.generate_raw_query
+    def generate_raw_query(
```
erosselli, line 662: TODO: reuse code from `generate_query_without_tuples`

src/fides/api/service/connectors/sql_connector.py
```diff
@@ -146,6 +146,27 @@ def test_connection(self) -> Optional[ConnectionTestStatus]:
...
 
+    def execute_standalone_retrieval_query(
+        self, node: ExecutionNode, fields: List[str], filters: Dict[str, List[Any]]
+    ):
+        if not node.collection.custom_request_fields():
+            logger.error(
+                "Cannot call execute_standalone_retrieval_query on a collection without custom request fields"
+            )
+            return None
+
```
erosselli, line 157: not sure if we want to keep this check or if we want to allow other cases in which the standalone query could be useful

src/fides/api/service/privacy_request/email_batch_service.py
```diff
@@ -141,8 +141,9 @@ def initiate_scheduled_batch_email_send() -> None:
...
         trigger="cron",
-        minute="0",
-        hour="12",
-        day_of_week="mon",
-        timezone="US/Eastern",
+        second="30",
+        # minute="0",
+        # hour="12",
+        # day_of_week="mon",
+        # timezone="US/Eastern",
```
erosselli, line 148: temporary change so the email erasure task runs every 30s instead of weekly

src/fides/api/task/create_request_tasks.py
```diff
@@ -191,9 +193,21 @@ def base_task_data(
...
+                node
+            ].format_traversal_details_for_save()
+        else:
+            # If node is not in traversal_nodes, then it is a node added for
+            # custom request field processing. We manually build the traversal details,
+            # with no incoming or outgoing edges and no input keys.
+            graph_node = dataset_graph.nodes[node]
+            traversal_details = TraversalDetails.create_empty_traversal(
+                graph_node.dataset.connection_key
+            ).model_dump(mode="json")
```
erosselli, line 210: this was needed so the unreachable nodes wouldn't break the request tasks... not the cleanest solution but I couldn't find a better one so suggestions are welcome

src/fides/data/sample_project/fides.toml
```diff
@@ -23,6 +23,11 @@ root_password = "Testpassword1!"
 
 [execution]
 require_manual_request_approval = true
+use_dsr_3_0 = true
+state_polling_interval = 30
+request_task_ttl = 604800
+allow_custom_privacy_request_field_collection = true
+allow_custom_privacy_request_fields_in_request_execution = true
```
erosselli, line 30: same here, changed just for testing

clients/admin-ui/src/types/api/models/ConnectionConfigurationResponse.ts
pattisdr: @erosselli can you explain in more detail what issues you were running into? I'm wondering if this is pydantic related

clients/admin-ui/src/features/datastore-connections/system_portal_config/forms/ConnectorParametersForm.tsx
pattisdr: Can you be more specific about what was exploding, I'm wondering if it is pydantic related?

src/fides/api/service/connectors/dynamic_erasure_email_connector.py
```diff
@@ -0,0 +1,258 @@
...
+                ExecutionLog.create(
+                    db=db,
+                    data={
+                        "connection_key": self.configuration.key,
+                        "dataset_name": self.configuration.name,
+                        "collection_name": self.configuration.name,
+                        "privacy_request_id": privacy_request.id,
+                        "action_type": ActionType.erasure,
+                        "status": ExecutionLogStatus.complete,
+                        "message": f"Erasure email instructions dispatched for '{self.configuration.name}'",
```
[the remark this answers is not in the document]
pattisdr: These logs help inform what shows up here:

Screenshot 2024-08-25 at 2 18 49 PM

src/fides/api/task/create_request_tasks.py
pattisdr: The big thing here would be making these nodes reachable in the first place unfortunately

src/fides/api/service/connectors/dynamic_erasure_email_connector.py
pattisdr: If we have the nodes hooked up properly, then we wouldn't need a "standalone" query - the custom fields connector could have already retrieved the relevant row and stored it on the request task before we got to the dynamic erasure  email connector

pattisdr (commented):
You did a nice job tracking down all the individual pieces to get something working end-to-end @erosselli - and your self-review of this PR is great. But big picture, I have several concerns with how we're slotting in this new connector type and would consider refactoring so this doesn't require so much special handling.  This might have to be a follow-up due to time constraints.  Generally I think the dynamic erasure email connector is doing too much.

My primary issues:

- Separation of concerns: Dynamic Email Connector is doing all the heavy lifting.  this includes querying the Postgres Custom Field Collection, when that is typically taken care of as part of the Postgres Custom Field Collection upstream node. Each node is typically responsible for querying itself.
- Lots of special casing given that Postgres Custom Field Collection is not reachable via custom field
- Dynamic Email Connector is being treated as a bulk email connector. Bulk emails are sent once weekly, combining the last week's users that need their info deleted into a single email.  However, with dynamic emails, I don't think we get a lot of benefit out of postponing these to fire once weekly. Have we considered firing these as part of the erasure graph itself?

Here's how I might approach this connector instead:

- Make collections like the Postgres Custom Field Collection reachable by custom field, not just identity data.  That way the Postgres Custom Field Collection could be a proper part of the graph, with upstream nodes (the root collection).  I definitely get that this is easier said than done, but avoids a lot of the special casing.
- Also add the Dynamic Email Collection to the graph. We typically need a DatasetConfig/Dataset for this.  One way to achieve this would be when setting up the integration, a dataset is automatically created with just a single recipient_email_address field that adds fides_meta.references to point to the postgres collection with the custom field. - Then our graph would look like: Root -> Postgres Custom Field Collection -> Dynamic Email Collection -> Terminator. 
- Unlike the other email connectors, the dynamic email connector could look more similar to a saas connector/database connector and have access/erasure methods.  The erasure method itself would use the access data already retrieved by upstream Postgres Custom Field Collection to fire off a single erasure email instead of being one that is run at the end in the "bulk email" section.  I know I'm hand waving over some of the complexity here, since there is no "access" step here -

data/dataset/postgres_example_custom_request_field_dataset.yml
erosselli: it definitely adds more flexibility, and we can keep it if we see it being useful. In that case I need to add the corresponding logic to map the DSR value to the field value

pattisdr:
@erosselli thanks for discussing next steps - as part of this increment, can you add code comments to the special-cased locations, where it is temporary until we get custom fields treated in more of a first class way?

src/fides/api/service/connectors/dynamic_erasure_email_connector.py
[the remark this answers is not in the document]
erosselli: So that's copied from the existing email connector , which uses the configuration name. Do you still think we should change it? maybe we can use the name , and fallback to the key if the name doesn't exist?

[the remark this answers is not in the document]
pattisdr: yes fallback to key if name doesn't exist is a good idea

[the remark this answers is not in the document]
erosselli: just pushed the change

src/fides/api/service/connectors/dynamic_erasure_email_connector.py
```diff
@@ -0,0 +1,342 @@
...
+                filtered_user_identities: Dict[str, Any] = (
+                    filter_user_identities_for_connector(self.config, user_identities)
+                )
+
+                # If we couldn't get the email or the user identities, we skip the privacy request
+                if not email or not filtered_user_identities:
+                    skipped_privacy_requests.append(privacy_request.id)
+                    self.add_skipped_log(db, privacy_request)
+                    continue
+
```
erosselli, line 233: same question here, do we skip these privacy requests or do we want it to error? The existing generic erasure email connector skips the request if `filtered_user_identities` is None

src/fides/api/service/connectors/dynamic_erasure_email_connector.py
```diff
@@ -0,0 +1,342 @@
...
+        try:
+            if not self.config.test_email_address:
+                raise MessageDispatchException(
+                    f"Cannot test connection. No test email defined for {self.configuration.name}"
+                )
+            # synchronous for now since failure to send is considered a connection test failure
+            send_single_erasure_email(
+                db=db,
+                subject_email=self.config.test_email_address,
+                subject_name=self.config.third_party_vendor_name,
```
erosselli, line 335: subject_name should probably also be a dataset reference...

src/fides/api/service/connectors/base_erasure_email_connector.py
```diff
@@ -0,0 +1,137 @@
...
+CONFIG = get_config()
+
+ERASURE_EMAIL_CONNECTOR_TYPES = [
+    ConnectionType.generic_erasure_email,
+    ConnectionType.attentive,
+    ConnectionType.dynamic_erasure_email,
+]
+
+
+class BaseErasureEmailConnector(BaseEmailConnector):
```
erosselli, line 40: Most of the logic here was directly taken from the GenericEmailErasureConnector, just made the typing a bit more flexible so we can inherit from it in both cases

src/fides/api/service/connectors/dynamic_erasure_email_connector.py
pattisdr: maybe we break this up into two? It seems like a legitimate use case where the user identity doesn't exist so I'd say we skip it there.  But if the email doesn't exist, that seems like a problem, maybe it should fail. what do you think @galvana

src/fides/api/service/connectors/dynamic_erasure_email_connector.py
```diff
@@ -0,0 +1,342 @@
...
+            db, field="fides_key", value=dataset_key
+        )
+
+        if not dataset_config:
+            logger.error(
+                "DatasetConfig with key '{}' not found. Skipping erasure email send for connector: '{}'.",
+                dataset_key,
+                self.configuration.key,
+            )
+            return
```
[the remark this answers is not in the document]
pattisdr: I would raise an exception here just in case so the customer sees the failed privacy requests and we can work on addressing the connector instead of having this issue go undetected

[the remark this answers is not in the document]
erosselli: I don't think raising the exception is enough for the error to be visible , I think I also need to add an ExecutionLog for each privacy request. I'll do that as well to ensure the privacy request shows as failed. Also if I raise the exception it seems like the task keeps retrying which is probably not what we want?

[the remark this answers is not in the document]
pattisdr: ah right we're outside of that typical flow

src/fides/api/service/connectors/dynamic_erasure_email_connector.py
```diff
@@ -0,0 +1,258 @@
...
+
+        # Get the graph dataset from the dataset config
+        graph_dataset: GraphDataset = dataset_config.get_graph()
+
+        # Build the CollectionAddress using the dataset name and the collection name
+        split_field = dataset_reference.field.split(".")
+        collection_name = split_field[0]
+        collection_address = CollectionAddress(graph_dataset.name, collection_name)
+        # And get the field name from the datasetreference
+        field_name = ".".join(split_field[1:])
```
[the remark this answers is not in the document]
erosselli: Added validation to check the formatting of the reference fields and corresponding error message :)

src/fides/api/service/connectors/dynamic_erasure_email_connector.py
```diff
@@ -0,0 +1,502 @@
...
+    connector: BaseConnector
+    collection_address: CollectionAddress
+    collection_data: Any
+    email_field: str
+    vendor_field: str
+    dsr_field_to_collection_field: Dict[str, str]
+
+
+class BatchedIdentitiesData(NamedTuple):
+    email_address: str
```
erosselli, line 51: these are just some helper types I added to make the code easier to read

src/fides/api/service/connectors/dynamic_erasure_email_connector.py
```diff
@@ -0,0 +1,502 @@
...
+                send_single_erasure_email(
+                    db=db,
+                    subject_email=batched_identity_data.email_address,
+                    subject_name=batched_identity_data.vendor_name,
+                    batch_identities=batched_identity_data.identities,
+                    test_mode=False,
+                )
+            except MessageDispatchException as exc:
+                logger.info("Erasure email failed with exception {}", exc)
+                raise
```
pattisdr, line 153: We talked about this I believe...but to check my understanding, here the email didn't send (say we had bad mailgund credentials), and all the associated privacy requests are left in a "awaiting email send" status to be picked up on the next email send. 

Would it be useful to add an execution log for every privacy request in this batch that the email send failed?

tests/ops/api/v1/endpoints/test_connection_config_endpoints.py
```diff
@@ -1558,6 +1558,236 @@ def test_put_sovrn_connection_config_secrets_missing(
...
+        assert (
+            dynamic_erasure_email_connection_config_no_secrets.last_test_timestamp
+            is None
+        )
+        assert (
+            dynamic_erasure_email_connection_config_no_secrets.last_test_succeeded
+            is None
+        )
+
+    def test_put_dynamic_erasure_email_connection_config_mismtached_datasets(
```
pattisdr, line 1684: typos mismatched*

src/fides/api/service/connectors/dynamic_erasure_email_connector.py
```diff
@@ -0,0 +1,502 @@
...
+                        email, vendor_name, []
+                    )
+
+                batched_identities[key].identities.extend(
+                    filtered_user_identities.values()
+                )
+
+            except Exception as exc:
+                logger.error(
+                    "An error occurred when retrieving email from custom request fields for connector {}. Skipping email send for privacy request with id {}. Error: {}",
```
pattisdr, line 121: Similar here, we're not "skipping", we're "failing"

src/fides/api/service/connectors/dynamic_erasure_email_connector.py
```diff
@@ -0,0 +1,502 @@
...
+                "Connector configuration references different collections for email and vendor fields.",
+            )
+            raise DynamicErasureEmailConnectorException(error_log)
+
+        # We get the DatasetConfig instance using the dataset key provided in the configuration
+        dataset_config: Optional[DatasetConfig] = DatasetConfig.get_by(
+            db, field="fides_key", value=email_dataset_key
+        )
+
+        if not dataset_config:
```
pattisdr, line 266: Didn't we talk about the possibility of multiple matches and raising an error there as well?

src/fides/api/service/connectors/dynamic_erasure_email_connector.py
```diff
@@ -0,0 +1,502 @@
...
+            error_log = f"DatasetConfig with key {email_dataset_key} not found. Skipping erasure email send for connector: {self.configuration.key}."
+            logger.error(error_log)
+            self.error_all_privacy_requests(
+                db,
+                privacy_requests,
+                f"Connector configuration references DatasetConfig with key {email_dataset_key}, but such DatasetConfig was not found.",
```
pattisdr, line 272: Error message maybe: "but no such DatasetConfig was found"

src/fides/api/service/connectors/dynamic_erasure_email_connector.py
```diff
@@ -0,0 +1,502 @@
...
+        # Get the corresponding ConnectionConfig instance
+        connection_config: ConnectionConfig = (
+            db.execute(
+                db.query(ConnectionConfig).filter(
+                    ConnectionConfig.id == dataset_config.connection_config_id
+                )
+            )
+            .scalars()
+            .first()
+        )
```
pattisdr, line 289: Do you need the db.execute wrapper? Can you just do:

```
 db.query(ConnectionConfig).filter(ConnectionConfig.id == dataset_config.connection_config_id).first()
```

src/fides/api/service/connectors/dynamic_erasure_email_connector.py
```diff
@@ -0,0 +1,502 @@
...
+        try:
+            if not self.config.test_email_address:
+                raise MessageDispatchException(
+                    f"Cannot test connection. No test email defined for {self.configuration.name}"
+                )
+            # synchronous for now since failure to send is considered a connection test failure
+            send_single_erasure_email(
+                db=db,
+                subject_email=self.config.test_email_address,
+                subject_name="Test Vendor",  # Needs to be a string since vendor depends con custom request field values
```
pattisdr, line 495: typo: depends on

src/fides/api/service/connectors/dynamic_erasure_email_connector.py
```diff
@@ -0,0 +1,502 @@
...
+
+        # Get collection name and field name
+        collection_name = split_field[0]
+        field_name = ".".join(split_field[1:])
+
+        return collection_name, field_name
+
+    def process_connector_config(
+        self, db: Session, privacy_requests: Query
+    ) -> ProcessedConfig:
```
pattisdr, line 205: The function name/ProcessedConfig is throwing me here - maybe it's the word "config". which has other usages in our codebase..

src/fides/api/schemas/connection_configuration/connection_secrets_email.py
```diff
@@ -14,11 +14,17 @@ class AdvancedSettings(BaseModel):
...
+    """
+    Base Email Schema that contains the fields common to all email schemas.
+    Used to validate secrets for all email connectors
+    """
 
-    third_party_vendor_name: str
-    recipient_email_address: EmailStr
     test_email_address: Optional[EmailStr] = (
         None  # Email to send a connection test email
     )
```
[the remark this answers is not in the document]
erosselli: could you clarify a bit what you mean here? Are you saying the type shouldn't be `EmailStr` but `str`? 🤔

src/fides/api/schemas/connection_configuration/connection_secrets_dynamic_erasure_email.py
```diff
@@ -0,0 +1,72 @@
...
+        json_schema_extra={
+            "external_reference": True
+        },  # metadata added so we can identify these secret schema fields as external references
+    )
+
+    @staticmethod
+    def validate_dataset_references(
+        db: Session,
+        connection_secrets: "DynamicErasureEmailSchema",
+    ) -> Any:
```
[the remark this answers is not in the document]
erosselli: that makes sense 😄

src/fides/api/service/connectors/dynamic_erasure_email_connector.py
```diff
@@ -0,0 +1,502 @@
...
+    filter_user_identities_for_connector,
+    send_single_erasure_email,
+)
+from fides.config import get_config
+
+CONFIG = get_config()
+
+
+class DynamicErasureEmailConnectorException(Exception):
+    pass
```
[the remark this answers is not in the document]
erosselli: Are all exceptions meant to be in common exceptions? This one doesn't seem "common", and it'd be weird for another class to raise it

src/fides/api/service/connectors/dynamic_erasure_email_connector.py
```diff
@@ -0,0 +1,502 @@
...
+    def get_config(self, configuration: ConnectionConfig) -> DynamicErasureEmailSchema:
+        return DynamicErasureEmailSchema(**configuration.secrets or {})
+
+    def batch_email_send(self, privacy_requests: Query) -> None:
+        logger.debug(
+            "Starting batch_email_send for connector: {} ...", self.configuration.name
+        )
+
+        db: Session = Session.object_session(self.configuration)
+        proccessed_config = self.process_connector_config(db, privacy_requests)
```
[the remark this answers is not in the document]
erosselli: oof , you made me realize I never installed my go-to spell checker vscode extension 😄

src/fides/api/service/connectors/dynamic_erasure_email_connector.py
erosselli: That sounds reasonable. This part is copied from the Generic Erasure Email connector, which also doesn't add any execution logs for this case. But I can add them here if we don't find having slightly different behavior between both connectors

src/fides/api/service/connectors/dynamic_erasure_email_connector.py
erosselli: ah yes but that turned out not to be the case, I was just very confused because the fides_key for the connection configs usually match those in the dataset. I double and triple checked, also validated with Adrian, that the dataset reference field points to a single DatasetConfig, not the Dataset itself. This is assuming the fides_key of a DatasetConfig is unique , otherwise we could still have the issue I think

src/fides/api/service/connectors/dynamic_erasure_email_connector.py
```diff
@@ -0,0 +1,502 @@
...
+        Sends an email to the "test_email" configured, just to establish that the email workflow is working.
+        """
+        logger.info("Starting test connection to {}", self.configuration.key)
+
+        db = Session.object_session(self.configuration)
+
+        try:
+            if not self.config.test_email_address:
+                raise MessageDispatchException(
+                    f"Cannot test connection. No test email defined for {self.configuration.name}"
```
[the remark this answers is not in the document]
erosselli: Created https://ethyca.atlassian.net/browse/PROD-2709

src/fides/api/schemas/connection_configuration/connection_secrets_email.py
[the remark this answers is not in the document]
pattisdr: Maybe, I just experimented briefly, I'm just hoping you could investigate this a bit more! I don't think the "Test Connection" button works for your connector via the UI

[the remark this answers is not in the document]
erosselli: I'll check it out 😄

src/fides/api/models/privacy_request.py
```diff
@@ -1775,6 +1775,23 @@ class TraversalDetails(FidesSchema):
...
 
+    # TODO: remove this method once we support custom request fields in DSR graph.
+    @classmethod
+    def create_empty_traversal(cls, connection_key: str) -> TraversalDetails:
+        """
+        Creates an "empty" TraversalDetails object that only has the dataset connection key set.
+        This is a bit of a hacky workaround needed to implement the Dynamic Erasure Emails feature,
+        and should be needed only until we support custom request fields as entry points to the DSR graph.
+        This is needed because custom request field nodes aren't currently reachable, so they don't have
+        a real TraversalNode associated to them.
```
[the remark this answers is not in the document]
erosselli: I agree it's kind of confusing, but I'm not too sure on how I can make this more clear for the reader. Do you have any suggestions for a better wording?

src/fides/api/service/connectors/dynamic_erasure_email_connector.py
```diff
@@ -0,0 +1,502 @@
...
+        # We execute a standalone retrieval query on the connector to get the email address
+        # based on the custom request fields on the privacy request and the field_name
+        # provided as part of the connection config.
+        # Following the previous example, this query would be something like
+        # SELECT email_address FROM site_info WHERE site_id = '1234'
+        retrieved_data = processed_config.connector.execute_standalone_retrieval_query(
+            node=execution_node,
+            fields=[processed_config.email_field, processed_config.vendor_field],
+            filters=query_filters,
+        )
```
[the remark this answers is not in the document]
pattisdr: I think that's a great workaround @erosselli! good callout

src/fides/api/service/connectors/dynamic_erasure_email_connector.py
```diff
@@ -0,0 +1,502 @@
...
+                if key not in batched_identities:
+                    batched_identities[key] = BatchedIdentitiesData(
+                        email, vendor_name, []
+                    )
+
+                batched_identities[key].identities.extend(
+                    filtered_user_identities.values()
+                )
+
+            except Exception as exc:
```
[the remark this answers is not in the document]
erosselli: that's a good callout. The only caveat I see is that since this could run against any connector, we'd need to catch all possible connection errors (i.e from sqlalchemy, mongo, dynamo, etc etc). The good thing is right now we're only supporting this for SQL connectors (and Snowflake which we actually query with the SQLQueryConfig) so I can add the ones from that connector now. If we ever add support for other connectors we'd also need to make sure to catch their exceptions

src/fides/api/graph/config.py
pattisdr: If I didn't say this elsewhere, great docstring here now 👍

src/fides/api/service/connectors/dynamic_erasure_email_connector.py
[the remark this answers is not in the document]
pattisdr: That's true.. there's also been some back and forth about the tradeoffs with this pattern elsewhere because we make it harder to roubleshoot - I expect us to continue to focus here -

src/fides/api/service/connectors/dynamic_erasure_email_connector.py
pattisdr: Thanks @erosselli

src/fides/api/service/connectors/dynamic_erasure_email_connector.py
pattisdr: OK thanks for clarifying this was confusing to me too

src/fides/api/service/connectors/dynamic_erasure_email_connector.py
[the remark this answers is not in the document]
pattisdr: For posterity, not a lot of privacy requests expected for this iteration so we'll postpone performance related things here.

src/fides/api/service/connectors/dynamic_erasure_email_connector.py
pattisdr: we'll just go with this, the excellent docstrings you've added to this file help explain some of the non-standard operations you're doing

pattisdr (approved):
Looks good @erosselli, thanks for all the extra docstrings and clarification with this nonstandard connector added on this iteration- for a PR this size, I'd do one more skim through to make sure there's no code you're accidentally committing, check your downrev one last time, etc.

PR #5232, opened 2024-08-26
PROD-2597: Implements privacy request fuzzy search

eastandwestwind:
Closes https://ethyca.atlassian.net/browse/PROD-2597 and  https://ethyca.atlassian.net/browse/PROD-2592

### Description Of Changes

Implements fuzzy search on request id or identity value. 

* To keep this fast and lightweight as possible, we do a prefix-forward fuzzy match, so this requires the beginning character(s) to match e.g. ("me" will return "meow@email.com" but not "ame@email.com"). 
* And yes, we can use either privacy request ids OR any identity value on the FE input, including phone numbers and all other identity types!

### Code Changes

* [ ] Adds ahocorasick module to aid with fast fuzzy search
* [ ] Adds fuzzy search utils as a central hub to manage building the search tree for fuzzy search, in addition to setting the cache signal that determines whether or not to utilize the cache (vs the DB) for fuzzy search
* [ ] Adds a new `DecryptedIdentityCacheMixin` that has been added to `PrivacyRequest` base model so that we can easily store and retrieve decrypted identities by privacy request from a new Redis cache.
* [ ] Adds tests for privacy request search endpoint, including creating bulk privacy requests for testing search

### Steps to Confirm

* [ ] Create privacy requests in privacy center
* [ ] Using admin UI privacy request manager, search for the identity value used, confirm results are fast and show expected items. Follow endpoint tests as a guidance here.
* [ ] Using admin UI privacy request manager, search for privacy request id, confirm results are fast and show expected items
* [ ] Confirm that no "Toggle PII" button exists on Admin-UI privacy request manager. Confirm that identities are no longer hidden.

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`


src/fides/api/util/fuzzy_search_utils.py
```diff
@@ -0,0 +1,92 @@
...
+            if not all_identities_for_request:
+                # Safeguard if specific privacy requests do not have identities stored in the cache
+                logger.info(
+                    "Decrypted identities could not be found in cache for privacy request. Proceeding to look up in DB instead of using cache."
+                )
+                _add_decrypted_identities_to_automaton(request.get_persisted_identity.__dict__, request.id, automaton)  # type: ignore
+                # Write to cache for later
+                request.cache_decrypted_identities_by_privacy_request()
+            _add_decrypted_identities_to_automaton(all_identities_for_request, request.id, automaton)  # type: ignore
+    return automaton
```
galvana, line 92: I was thinking you would follow a singleton approach like we do for our Redis cache https://github.com/ethyca/fides/blob/main/src/fides/api/util/cache.py#L124, it would be a huge speed-up if we can build this up once on system startup and just refresh it periodically or add to it as new requests come in

galvana (changes requested):
@eastandwestwind I have some concerns about performance that I left some comments on but let's discuss it tomorrow

docker-compose.yml
```diff
@@ -91,6 +91,7 @@ services:
       POSTGRES_USER: "postgres"
       POSTGRES_PASSWORD: "fides"
       POSTGRES_DB: "fides"
+    command: ["postgres", "-c", "log_statement=all"]
```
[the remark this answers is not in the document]
eastandwestwind: ah yes, good catch!

src/fides/api/util/fuzzy_search_utils.py
eastandwestwind: Ah so you mean build the Automaton only once as well. Lemme think about how best to do this...

clients/admin-ui/src/features/privacy-requests/RequestTable.tsx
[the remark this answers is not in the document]
eastandwestwind: this was existing, but I found a way to fix this!

clients/admin-ui/cypress/e2e/privacy-requests.cy.ts
```diff
@@ -88,12 +88,11 @@ describe("Privacy Requests", () => {
       cy.wait("@getPrivacyRequest");
     });
 
-    it("shows the request details", () => {
+    it.only("shows the request details", () => {
```
galvana, line 91:
```suggestion
    it.only("shows the request details", () => {
```

src/fides/api/models/privacy_request.py
```diff
@@ -529,6 +531,13 @@ def persist_identity(self, db: Session, identity: Identity) -> None:
                     data=provided_identity_data,
                 )
 
+        # Simultaneously add identities to automaton for fuzzy search
+        try:
+            self.add_identities_to_automaton()
+        except Exception as exc:
+            # This should never affect the ability to create privacy requests
```
galvana, line 538: Nice!

galvana (approved):
Thanks for making the changes! This looks good, I just made some small comments but none of them are blockers so I'm approving the PR. The only other thing I would check with @mfbrown if we haven't already is if we're ok with the fuzzy search requiring the identity value to match at the start of the string.

PR #5238, opened 2024-08-27
include autofix for VS code users

gilluminate:
Uses filetype specific settings to avoid potential conflicts with the Python side of things.


.vscode/settings.json
```diff
@@ -17,7 +17,14 @@
   "typescript.enablePromptUseWorkspaceTsdk": true,
   "typescript.tsdk": "./clients/node_modules/typescript/lib",
   "editor.defaultFormatter": "esbenp.prettier-vscode",
-  "typescript.preferences.importModuleSpecifier": "non-relative",
```
gilluminate, line 20: removed this also because it has never behaved the way we expected it to.

lucanovera (approved):
Thanks for making this change

PR #5253, opened 2024-09-03
PROD-2655: validate removing path

thingscouldbeworse:
Closes #PROD-2655

### Description Of Changes

Change cli server_host validator to remove any path given in config file and test to confirm

### Code Changes

* [ ] if there's a forward slash in the `server_host` variable split the string and take everything before the forward slash

### Steps to Confirm

* [ ] run the tests
* [ ] change the `./fides/fides.toml` env var of `server_host` to `localhost/api/v1` instead of just `localhost` and try to run `fides status`

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### DIFF
--- a/src/fides/cli/utils.py
+++ b/src/fides/cli/utils.py
@@ -87,7 +87,7 @@ def check_server(cli_version: str, server_url: str, quiet: bool = False) -> None
         # The server is ratelimiting us
         echo_red("Server ratelimit reached. Please wait one minute and try again.")
         raise SystemExit(1)
-    elif health_response.status_code != 200:
+    if health_response.status_code != 200:
         echo_red(
             f"Server response: HTTP {health_response.status_code} - {health_response.text}"
         )


src/fides/config/cli_settings.py
```diff
@@ -45,8 +45,14 @@ class CLISettings(FidesSettings):
     @field_validator("server_url")
     @classmethod
     def get_server_url(cls, value: str, info: ValidationInfo) -> str:
-        """Create the server_url."""
+        """Create the server_url from the server configuration values. Strips path if present."""
         host = info.data.get("server_host")
+        # remove path if present from host
+        if "/" in host:
+            print(
+                f"server_host variable {host} has a forward slash, removing it. No path or porotocol should be present in the server_host variable."
```
adamsachs, line 53: nit on the wording - we're not just removing the forward slash, right? we're removing everything that follows it.

src/fides/config/cli_settings.py
```diff
@@ -45,8 +45,14 @@ class CLISettings(FidesSettings):
...
+            )
+            host = host.split("/")[0]
```
adamsachs, line 55: this will lead to some pretty unexpected results (which may be a bit hard to grok, for the end user) if the forward slash is part of a protocol prefix, or if it's some other unexpected input that isn't a path specification. i know the path specification is the primary use case we're trying to solve here, but maybe it'd be worth using [urllib.parse](https://docs.python.org/3/library/urllib.parse.html) to be a bit more precise/surefire in our handling? it's part of stdlib, so we wouldn't have to introduce a new dependency or anything.

let me know what you think!

adamsachs (commented):
looks generally solid, but a few things on edge cases that i think we could handle better, since this is all about making the UX as seamless as possible! let me know what you think.

src/fides/config/cli_settings.py
thingscouldbeworse: good point, here it does it with `urlparse` https://github.com/ethyca/fides/pull/5253/commits/579dd5b7f0610a5ad9e43816b8334a9c3d18ac9d#diff-5d19bcc00ba5a896a25504e3fd1154fa415dae45e11e9a4fa764e3a305916110R56

src/fides/config/cli_settings.py
thingscouldbeworse: true, fixed along with the `urlparse` change https://github.com/ethyca/fides/pull/5253/commits/579dd5b7f0610a5ad9e43816b8334a9c3d18ac9d#diff-5d19bcc00ba5a896a25504e3fd1154fa415dae45e11e9a4fa764e3a305916110R59

src/fides/cli/utils.py
```diff
@@ -85,6 +87,15 @@ def check_server(cli_version: str, server_url: str, quiet: bool = False) -> None
         # The server is ratelimiting us
         echo_red("Server ratelimit reached. Please wait one minute and try again.")
         raise SystemExit(1)
+    elif health_response.status_code != 200:
+        echo_red(f"Server response: {health_response.text}")
+        raise SystemExit(1)
+
+    if not health_response.json().get("version", False):
+        echo_red(
+            f"Server returned malformed response: {health_response.text}\nPlease check the server and config and try again."
```
adamsachs, line 96: nice handling here 👍

src/fides/cli/utils.py
```diff
@@ -85,6 +87,15 @@ def check_server(cli_version: str, server_url: str, quiet: bool = False) -> None
         # The server is ratelimiting us
         echo_red("Server ratelimit reached. Please wait one minute and try again.")
         raise SystemExit(1)
+    elif health_response.status_code != 200:
+        echo_red(f"Server response: {health_response.text}")
```
adamsachs, line 91: nit: might be nice to also include the status code here?

src/fides/config/cli_settings.py
```diff
@@ -45,16 +46,20 @@ class CLISettings(FidesSettings):
...
+        port_str: str = f":{port}" if port else ""
+        server_url = f"{protocol}://{host}{port_str}"
+        # check if the host var had a path included
+        parsed = urlparse(server_url)
+        if parsed.path:
+            print(
+                f"Warning: The server_host value '{host}' includes a path. This will be stripped. Please update the server_host value to '{parsed.netloc}'."
+            )
+            server_url = f"{parsed.scheme}://{parsed.netloc}{port_str}"
+            print(f"The server_url: `{server_url}` will be used.")
```
adamsachs, line 62: nice! this looks pretty solid now 👍

adamsachs (approved):
looks good @thingscouldbeworse, thanks for the extra care!

one more nit i saw and also looks like a lint failure - otherwise good to go imo!

src/fides/cli/utils.py
thingscouldbeworse: Good call, added here: https://github.com/ethyca/fides/pull/5253/commits/222832d894e899ef01a6b25fd09dcae554fc5cfa

adamsachs (approved):
👍

PR #5256, opened 2024-09-04
Migrate identity hashes

galvana:
Closes [PROD-2633](https://ethyca.atlassian.net/browse/PROD-2633)

⚠️ This PR contains a database migration
### Description Of Changes

This PR migrates usages of the `bcrypt` hash to `SHA-256 + salt` for **non-credential data**. The `bcrypt` hash function is a computationally intensive function that should only be used for passwords, client secrets, and other credentials. The use of `bcrypt` to hash identity data like emails and phone numbers was causing performance issues for any endpoint that hashed incoming identities either for saving or for searching.

At a high level the migration approach is:

- Update any searches that rely on identity hashes to search using a `bcrypt` hash **and** a `SHA-256` hash
- Immediately start saving identities using `SHA-256`
- Migrate select tables with `bcrypt` hashes to `SHA-256` hashes
- Once the migration is complete for a table, only search using the new `SHA-256` hashes

### Code Changes

* [ ] Added an `is_hash_migrated` column to select tables to track the migration
* [ ] Added an `identity_table` with a single encrypted value to store the system-wide identity salt
  * [ ] Accessed with a `@cache` decorated `get_identity_salt` function that also initializes the salt value on server startup if it doesn't exist
* [ ] Updated filters in the consent and privacy request endpoints to use both the bcrypt and SHA-256 hashes so we can cut over as soon as the migrations are complete (see inline comments for details)
* [ ] Migrated the existing `hash_with_salt` to `hash_credential_with_salt` (bcrypt) for passwords and `hash_value_with_salt` (SHA-256) for identity values
* [ ] `HashMigrationMixin` with shared migration logic
* [ ] `bcrypt_migration_task` job

### Steps to Confirm

* [ ] Checkout main and create some privacy requests
* [ ] Checkout this branch and check the logs for `Completed hash migration for...`
* [ ] Check the `privacyrequest` table and make sure every row has `is_hash_migrated` set to true

This is the easiest check but this migration applies to
- CurrentPrivacyPreference
- ProvidedIdentity
- CustomPrivacyRequestField
- PrivacyPreferenceHistory
- ServedNoticeHistory

### Pre-Merge Checklist

* [ ] Update `CHANGELOG.md`

[PROD-2633]: https://ethyca.atlassian.net/browse/PROD-2633?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/fides/api/api/v1/endpoints/consent_request_endpoints.py
```diff
@@ -101,12 +101,12 @@ def _filter_consent(
...
 
     if identity:
-        hashed_identity = ProvidedIdentity.hash_value(value=identity)
+        identity_hashes = ProvidedIdentity.hash_value_for_search(identity)
         identities: Set[str] = {
             identity[0]
             for identity in ProvidedIdentity.filter(
                 db=db,
-                conditions=(ProvidedIdentity.hashed_value == hashed_identity),
+                conditions=(ProvidedIdentity.hashed_value.in_(identity_hashes)),
```
galvana, line 109: This is the main change to be able support the "on the fly" cutover to the new SHA-256 hashes. Instead of checking that the `hashed_value` is equal to a `hashed_identity` we check that it matches either the bcrypt hash or the SHA-256 hash. Each record can be found by one of those two hashes as the migration is in progress but `hash_value_for_search` will only compute the SHA-256 hash after the migration is complete and no longer compute the expensive bcrypt hash.

src/fides/api/cryptography/cryptographic_util.py
```diff
@@ -16,12 +17,26 @@ def decode_password(password: str) -> str:
...
 
 
+def hash_value_with_salt(text: bytes, salt: bytes) -> str:
+    """
+    Hashes the text using SHA-256 with the provided salt and returns the hex string representation.
+
+    **This is a quick hashing function only suitable for general data, use `hash_credential_with_salt` to hash
+    passwords or other credentials.**
+    """
+    return hashlib.sha256(text + salt).hexdigest()
```
galvana, line 37: Splitting this out into two hashing functions, bcrypt will still be used for credentials, but we will use SHA-256 for non-credential values like identities.

src/fides/api/migrations/hash_migration_mixin.py
```diff
@@ -0,0 +1,82 @@
...
+    """
+    A mixin that includes common logic for models that have hashed fields that need be be migrated from using bcrypt to SHA-256.
+    """
+
+    # set existing rows to false but set to true on create since we'll
+    # be using a new hashing function moving forward
+    is_hash_migrated = Column(Boolean, nullable=False, server_default="f", default=True)
+
+    @declared_attr
+    def __table_args__(cls: Any) -> Tuple:
```
galvana, line 21: I was struggling with the type for `cls` so I just went with `Any`. Open to suggestions here!

src/fides/api/migrations/hash_migration_mixin.py
```diff
@@ -0,0 +1,82 @@
...
+        """Hash value using bcrypt."""
+
+    @classmethod
+    @abstractmethod
+    def hash_value(
+        cls,
+        value: MultiValue,
+        encoding: str = "UTF-8",
+    ) -> Optional[str]:
+        """Hash value using SHA-256."""
```
galvana, line 50: I set these as abstract because I expect any class that implements the `HashMigrationMixin` to have these two functions. Not sure if there is a better way to do this

src/fides/api/migrations/hash_migration_mixin.py
```diff
@@ -0,0 +1,82 @@
...
+        Once the migration is complete for this model, the function will only return the faster,
+        SHA-256 and avoid the computationally expensive bcrypt hash.
+        """
+
+        hashed_values = []
+        if hashed_value := cls.hash_value(value):
+            hashed_values.append(hashed_value)
+        if not HashMigrationTracker.is_migrated(cls.__name__):
+            if hashed_value := cls.bcrypt_hash_value(value):
+                hashed_values.append(hashed_value)
```
galvana, line 68: This is where the cutover happens, once the table is migrated we stop using bcrypt because all rows will be using the SHA-256 hash.

src/fides/api/models/identity_salt.py
```diff
@@ -0,0 +1,42 @@
...
+)
+
+from fides.api.db.base_class import Base  # type: ignore[attr-defined]
+from fides.api.db.base_class import JSONTypeOverride
+from fides.config import CONFIG
+
+
+class IdentitySalt(Base):
+    """
+    A single-row table used to store the encrypted salt value used for identity hashing (SHA-256).
```
galvana, line 16: I decided to go with this automatic approach to prevent the user from having to deal with new Fides config settings. I also know the salt isn't strictly a sensitive value, but anything extra we do to hide it won't hurt.

src/fides/api/models/identity_salt.py
```diff
@@ -0,0 +1,42 @@
...
+    )  # Type bytea in the db
+    single_row = Column(
+        Boolean,
+        default=True,
+        nullable=False,
+        unique=True,
+    )  # used to constrain table to one row
+
+    CheckConstraint("single_row", name="identity_salt_single_row_check")
+    UniqueConstraint("single_row", name="identity_salt_single_row_unique")
```
galvana, line 42: Just going with a `single_row` constraint like we do for other tables. Ideally we would make this read only but I didn't look further into this yet

src/fides/api/migrations/hash_migration_mixin.py
adamsachs: 🤷 i wouldn't lose too much sleep over it

src/fides/api/migrations/hash_migration_mixin.py
```diff
@@ -0,0 +1,82 @@
...
+        Creates a dynamically named partial index on the table to track unmigrated rows.
+        The size of the index will shrink as more rows are migrated.
+        """
+        return (
+            Index(
+                f"idx_{cls.__tablename__}_unmigrated",  # type: ignore[attr-defined]
+                "is_hash_migrated",
+                postgresql_where=cls.is_hash_migrated.is_(False),  # type: ignore[attr-defined]
+            ),
+        )
```
adamsachs, line 32: this seems fancy, and i'm not entirely sure how it works, so i'll defer to you on it if you're sure this is the right approach.

generally though, i just wanna double check that creating an index for this column is _worth_ it. i understand that it'll help improve the performance/efficiency of the migration task, but at what cost? will it slow down writes to this table more generally, moving forward?

maybe that negative impact is minimal, maybe i'm overthinking it or thinking about it incorrectly. if you've considered and investigated this, then i'll trust your judgement on it - i just wanted to flag it in case it hadn't been considered yet!

src/fides/api/migrations/hash_migration_tracker.py
adamsachs, line 1: again, may just wanna flag some of this for removal once we're comfortable that everyone's completed the migration. some nice modules here that we'll probably want to get rid of just to reduce complexity over time, what at shame 😅

adamsachs (commented):
@galvana this is looking quite good - a really nice implementation and an elegant way of handling the migration. very well documented throughout, and nicely designed - it's really easy to follow.

i've got some minor comments throughout, let me know what you think. i didn't test myself, but assume you've given it, or will give it, some rigorous manual testing, given how involved of a change it is.

i tried to leave some comments on places where we may want to mark code for eventual deprecation/removal since some of this is only needed temporarily by nature...but i certainly missed some spots. so that's just a general comment - it may be good taking a pass through and clearly marking which modules/classes/functions are just needed in an "interim" while we ensure migrations are completed....

galvana (commented):
Thanks for the review @adamsachs. I addressed most of your concerns except for these two:

- Still using a Celery task for the migration, I'm just sticking to the convention for our other jobs
- I ran into conflicts adding `ABC` to `HashMigrationMixin` and decided it wasn't worth troubleshooting further at this point in the sprint

src/fides/api/migrations/hash_migration_job.py
```diff
@@ -0,0 +1,102 @@
...
+    ), "Scheduler is not running! Cannot migrate tables with bcrypt hashes."
+
+    logger.info("Initiating scheduler for hash migration")
+    scheduler.add_job(
+        func=bcrypt_migration_task,
+        id=HASH_MIGRATION,
+    )
+
+
+@celery_app.task(base=DatabaseTask, bind=True)
```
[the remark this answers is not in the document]
adamsachs: heh, i do appreciate looking for consistency, but i'm _pretty_ sure that this is rarely/never intended if the function  isn't actually  going to be run as a celery task.

as i look a bit closer into it, i suppose it's _possible_ that this type of usage could be intentional, since we actually leverage a different db engine (our 'task engine') within the `DatabaseTask.get_new_session`...so maybe there's a use case where we'd want to use that engine, even if the functions aren't running as celery tasks? it feels a bit contrived - but i'd like to check with @pattisdr if that's what was intended when adding the decorator to e.g. [this function](https://github.com/ethyca/fides/pull/4760/files#diff-e9b609a6bd089ccdf2444e0784f0f81e6fbcc0380601a583c854ae7a0e420557R179-R180) in the DSR 3.0 refactor - or if that's also just accidental usage!

i do wanna get clear on this, because as your comment makes clear, if we don't start clearing this up, it's only going to cause more confusion as we add more scheduled tasks!

src/fides/api/migrations/hash_migration_mixin.py
adamsachs: perfect, thank you @galvana !

src/fides/api/migrations/hash_migration_tracker.py
adamsachs: beautiful! an original work?

src/fides/api/alembic/migrations/versions/a249a089f23b_add_identity_salt_table.py
```diff
@@ -0,0 +1,50 @@
...
+            nullable=True,
+        ),
+        sa.Column(
+            "encrypted_value",
+            sqlalchemy_utils.types.encrypted.encrypted_type.StringEncryptedType(),
+            nullable=True,
+        ),
+        sa.Column("single_row", sa.Boolean(), nullable=False),
+        sa.PrimaryKeyConstraint("id"),
+        sa.UniqueConstraint("single_row"),
```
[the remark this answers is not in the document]
adamsachs: interesting! thanks for adding 👍

src/fides/api/migrations/hash_migration_job.py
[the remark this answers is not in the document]
pattisdr: Too long ago to recall but I doubt it was intentional, just an easy way to get the session.  I agree we're creating unneccesary confusion here by doing it like this.

adamsachs (approved):
looks great @galvana , thanks for the extra care in testing some of my questions out! 👍 

only lingering question for me is the celery decorator, don't need to hold up approval on that but i would like to get that resolved one way or another before we merge

src/fides/api/migrations/hash_migration_job.py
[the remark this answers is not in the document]
adamsachs: thanks @pattisdr  - 

in that case @galvana, i'd suggest we remove this. it should be pretty straightforward for us to get a well-handled db session - see an analogous example [here](https://github.com/ethyca/fidesplus/blob/main/src/fidesplus/api/service/discovery/discovery_monitor_executor.py#L101) in the discovery monitor executor (this function is also run as a scheduled task!). that's using a function defined in `fidesplus` to retrieve the db session but we can/should add a `@contextmanager` decorator to our `deps.get_db()` in fides OSS, i think, so it can be used similarly...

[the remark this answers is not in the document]
galvana: Thank you two for the input! I've promoted `get_db` to a context manager, let's just make sure all the tests pass

PR #5260, opened 2024-09-05
PROD-2663: specify dataset on fides pull

thingscouldbeworse:
Closes #PROD-2663

### Description Of Changes
 option for resource and key on push/pull

### Code Changes
* [ ] `fides pull` now has subcommands for each resource type and those subcommands take a `fides_key` argument to specify which one
* [ ] if a manifest file already exists for that resource type the file is rewritten with the server version of the resource. Otherwise a new file is created

### Steps to Confirm
All these commands should be valid
* [ ] install editable `pip install -e ./`
* [ ] `fides pull`
* [ ] `fides pull dataset fides_db`
* [ ] `fides pull .fides/`
* [ ] `fides pull .fides/ -a .fides/test_resources.yml`
* [ ] `fides pull dataset fides_db`

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`


src/fides/cli/options.py
```diff
@@ -38,6 +38,18 @@ def resource_type_argument(command: Callable) -> Callable:
     return command
 
 
+def resource_type_option(command: Callable) -> Callable:
+    "Add the resource_type option."
+    command = click.option(
+        "--resource-type",
```
NevilleS, line 44: I _really_ prefer this as an argument (e.g. `fides pull dataset {key}`) than as an option (`--resource-type`). It just feels out of place to me?

Is there a particular reason you didn't use the argument type?

thingscouldbeworse: In my mind (and the default behavior for click) an "argument" like that is a requirement, non optional, whereas an option is optional. It looks like we could [override the behavior](https://stackoverflow.com/questions/44056153/optional-argument-in-command-with-click) and write a custom class to make optional click arguments if you feel strongly enough

thingscouldbeworse: This is all a consideration because `fides push` by default pushes everything. We could also split out this functionality into a different command

NevilleS: I'd avoid trying to hack the Click internals at all for this, they are well considered to make parsing a terminal command unambiguously.

That said... can you do this without achieve this by just having the "push" command group have a few subcommands with names like:
- name=system
- name=dataset
- name=data_category
- name='' 

The empty string would effectively be the default sub-command. I think it might let you do that?

If it feels janky and weird, don't do it and follow the standard way for Click apps 👍

src/fides/core/pull.py
```diff
@@ -66,6 +68,40 @@ def pull_existing_resources(
...
+                f"Manifest file {manifest_path} does not already exist and will be created"
+            )
+            manifest = {}
+        print("Writing out resource to file...", end=" ")
+        manifest[resource_type] = [server_resource]
+        write_manifest_file(manifest_path, manifest)
+        print("done.")
+
+    else:
+        echo_red(f"Resource with fides_key: {fides_key} not found on the server.")
```
pattisdr, line 102: This error message can be confusing if "resource_type" didn't match what was expected with the fides key

src/fides/core/pull.py
```diff
@@ -66,6 +68,40 @@ def pull_existing_resources(
...
+    fides_key: str,
+    resource_type: str,
+) -> None:
+    """
+    Pull a resource from the server by its fides_key and update the local manifest file.
+    """
+    if manifests_dir[-1] == "/":
+        manifests_dir = manifests_dir[:-1]
+    manifest_path = f"{manifests_dir}/{resource_type}.yaml"
+    print(f"Pulling resource with fides_key: {fides_key}...", end=" ")
```
pattisdr, line 84: Maybe add the resource type to this error message too  and add error handling if the resource_type is unexpected

src/fides/cli/commands/ungrouped.py
```diff
@@ -205,12 +207,35 @@ def worker(ctx: click.Context) -> None:
...
     """
     Parse local manifest files and upload them to the server.
     """
+    if fides_key and not resource_type:
+        echo_red("Must specify a resource type when using `fides_key`.")
+        raise SystemExit(1)
+
     config = ctx.obj["CONFIG"]
     taxonomy = _parse.parse(manifests_dir)
+    # if the user has specified a specific fides_key, push only that dataset from taxonomy
```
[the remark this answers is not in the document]
thingscouldbeworse: In the ticket it specifies that we should be able to push any resource (if I'm reading that correctly) https://ethyca.atlassian.net/browse/PROD-2663 " argument accepts top level resources via keys. (datasets, systems, categories, uses, subjects) "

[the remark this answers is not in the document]
pattisdr: I guess I interpreted it as only that resource would be pushed or pulled when the argument was supplied.  If that's not expected that's fine

src/fides/core/pull.py
thingscouldbeworse: I can see that, I changed here to always say the resource type, does that seem like it will cut it? https://github.com/ethyca/fides/pull/5260/commits/828cbe872ae6a3ff6ade8bb366a9175bc1df8b39

src/fides/cli/commands/ungrouped.py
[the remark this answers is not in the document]
thingscouldbeworse: hmm, the argument being the `fides_key`? That is my intention here, I might be missing something. Does it seem like I'm pushing the wrong thing under the old "default" behavior?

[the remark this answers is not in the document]
pattisdr: It does limit just the datasets - I have two datasets, `fides_db` and `fides_db_2`, and if I do `fides push dataset fides_db_2`, just `fides_db_2` is pushed.  I wasn't expecting the policies, data_categories, and subjects, uses, etc. to be also pushed, but I'm understanding your intent is to push all other types and only limit the one if a fides key is supplied.

```shell
fidesuser@3b88299a132b:/fides$  fides push dataset fides_db_2

> Loaded config from: /fides/.fides/fides.toml
Loading resource manifests from: .fides/
Taxonomy successfully created.
Orphan Dataset Warning: The following datasets are not found referenced on a System
fides_db_2
----------
Processing data_category resource(s)...
PUSHED 85 data_category resource(s).
----------
Processing policy resource(s)...
PUSHED 1 policy resource(s).
----------
Processing organization resource(s)...
PUSHED 1 organization resource(s).
----------
Processing dataset resource(s)...
PUSHED 1 dataset resource(s).
----------
Processing data_use resource(s)...
PUSHED 56 data_use resource(s).
----------
Processing data_subject resource(s)...
PUSHED 15 data_subject resource(s).
----------
Processing system resource(s)...
PUSHED 3 system resource(s).

```

src/fides/core/pull.py
pattisdr: helpful thanks

pattisdr (commented):
pausing here, I want to switch to main and make sure I understand current `fides push`/`fides pull` behavior, and compare it to here -

Also there's no additional test coverage included here looks like - our cli tests could be more thorough but there's some tests that could be appended to

src/fides/cli/commands/pull.py
```diff
@@ -0,0 +1,183 @@
...
+    config = ctx.obj["CONFIG"]
+    _pull.pull(
+        url=config.cli.server_url,
+        manifests_dir=manifests_dir,
+        headers=config.user.auth_header,
+        fides_key=fides_key,
+        resource_type="dataset",
+        all_resources_file=None,
+    )
+    echo_green(f"Successfully pulled {fides_key} resource from the server.")
```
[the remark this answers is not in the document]
thingscouldbeworse: Yes and it feels redundant with the other. Removed here: https://github.com/ethyca/fides/pull/5260/commits/190114519eea8c15c85dc6e483e12580e3b0a806#diff-aa790a7a4419e75ad08bcd578d54f03be70c59691b10c20c3857fcf5a9bdb783L83

CHANGELOG.md
```diff
@@ -47,6 +47,7 @@ The types of changes are:
 - Adding erasure support for Microsoft Advertising [#5197](https://github.com/ethyca/fides/pull/5197)
 - Implements fuzzy search for identities in Admin-UI Request Manager [#5232](https://github.com/ethyca/fides/pull/5232)
 - New purpose header field for TCF banner [#5246](https://github.com/ethyca/fides/pull/5246)
+- `fides` subcommands `push` and `pull` accept a `--fides-key` option and `--resource-type` allowing you to push/pull only one resource by name and type [#5260](https://github.com/ethyca/fides/pull/5260)
```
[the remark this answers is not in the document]
thingscouldbeworse: yes, updated here: https://github.com/ethyca/fides/pull/5260/commits/6c0f5462d15ace4fa8db798eca90e4a8bbe6ac2b

pattisdr (approved):
good call to do a more targeted change here.

To capture what I understand is the behavior here:
- A `fides pull dataset test_dataset` command will write the `test_dataset` definition to a brand new  `dataset.yaml` file even if there is already a dataset definition within the current directory in a separate pre-existing file of a different name.

I think a little more clarity in the docstrings would be useful

PR #5263, opened 2024-09-06
PROD-2644 Improve logging and error visibility for TraversalErrors

erosselli:
Closes [PROD-2644](https://ethyca.atlassian.net/browse/PROD-2644)

### Description Of Changes

When a `TraversalError` is raised as part of running an access or erasure request the logged error does not include the privacy request id, making it hard to find, and the UI shows the request in an error state but provides no information on what the error actually was. This PR adds a log that includes the privacy request id, and also creates an `ExecutionLog` instance so the error can be surfaced in the UI.

### Code Changes

* Created two specific error classes `UnreachableNodeError` and `UnreachableEdgeError` that inherit from `TraversalError`, and throw those instead when we have an unreachable node or edge, respectively. 
* Catch all `TraversalErrors` in both `run_access_request` and `run_access_request_deprecated` , log the error _including the id of the privacy request_ , create the necessary `ExecutionLog` instances, and raise the error again.
* Unrelated to the ticket but had to remove the `with_server_health_check` decorator from both the `worker` and `status` commands, they were mistakenly added in a recent PR.

#### Setup
* Start up Fides with the postgres integration, i.e `nox -s dev -- postgres`
* Create two datasets that have no identities and no references, such as
```yaml
    fides_key: prod-2644-example
    name: Example PROD-2644
    description: Example to reproduce PROD-2644
    collections:
      - name: address
        fields:
          - name: city
            data_categories: [user.contact.address.city]
          - name: house
            data_categories: [user.contact.address.street]
```
and
```yaml
    fides_key: prod-2644-example-2
    name: Example PROD-2644 2
    description: Second example to reproduce PROD-2644
    collections:
      - name: login
        fields:
          - name: id
          - name: customer_id
      - name: report
        fields:
          - name: id
          - name: email
```
* Add two different Postgres integrations to connect to the postgres DB running on port 6432
*  For each integration, go to the integration details, and choose one of the datasets you just created (e.g `prod-2644-example`) from the `Dataset` dropdown. You should have two integrations, each linked to a different dataset

#### DSR 2.0
* From the Privacy Center or the Swagger API , create a Privacy Request. 
* In the Admin UI, open the created privacy request: it should be in `error` status, and there should be an Error log in the `Activity Timeline` section (see screenshots below)

Screenshot 2024-09-11 at 11 19 14 AM
Screenshot 2024-09-11 at 11 19 22 AM
Screenshot 2024-09-11 at 11 19 29 AM

You should also see the log in the fides container, e.g 
`2024-09-06 13:57:08 2024-09-06 16:57:08.400 | ERROR    | fides.api.task.deprecated_graph_task:run_access_request_deprecated:116 - TraversalError encountered for privacy request pri_83ace5f2-e4f7-4c2b-9ce3-fb989cddb214. Error: Some nodes were not reachable:  prod-2644-example:address`

#### DSR 3.0
* Stop your fides server
* Open the `fides.toml` file, and replace the following two lines:
   * in the `[execution]` section, change `use_dsr_3_0=false` to `use_dsr_3_0=true`
   * in the `[celery]` section, change `task_always_eager = true` to `task_always_eager = false`
* Run the server with `nox -s dev -- shell postgres worker`
* Make sure the integrations and datasets you created in the Setup section still exist. 
* Repeat the same steps as for DSR 2.0 -- you may need to manually approve the request from the Admin UI, and wait a bit for the task to run in the worker. 

#### Other error cases
When the error is not for unreachable nodes, but rather a more generic `TraversalError` (e.g there's a cycle) then the error's shown a bit differently in the UI

Screenshot 2024-09-12 at 2 30 02 PM
Screenshot 2024-09-12 at 2 30 13 PM

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

[PROD-2644]: https://ethyca.atlassian.net/browse/PROD-2644?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/fides/api/common_exceptions.py
+++ b/src/fides/api/common_exceptions.py
@@ -25,6 +25,14 @@ class TraversalError(FidesopsException):
     """Fidesops error with the names of all nodes that could not be reached."""
 
 
+class UnreachableNodesError(TraversalError):
+    """Fidesops error with the names of all nodes that could not be reached."""
+
+
+class UnreachableEdgesError(TraversalError):
+    """Fidesops error with the names of all edges that could not be reached."""
+
+
 class ValidationError(FidesopsException):
     """Data does not pass validation."""
 
--- a/src/fides/api/common_exceptions.py
+++ b/src/fides/api/common_exceptions.py
@@ -22,15 +22,15 @@ class FidesopsException(Exception):
 
 
 class TraversalError(FidesopsException):
-    """Fidesops error with the names of all nodes that could not be reached."""
+    """Error with the names of all nodes that could not be reached."""
 
 
 class UnreachableNodesError(TraversalError):
-    """Fidesops error with the names of all nodes that could not be reached."""
+    """Error with the names of all nodes that could not be reached, inherits from TraversalError."""
 
 
 class UnreachableEdgesError(TraversalError):
-    """Fidesops error with the names of all edges that could not be reached."""
+    """Error with the names of all edges that could not be reached, inherits from TraversalError."""
 
 
 class ValidationError(FidesopsException):
--- a/src/fides/api/graph/traversal.py
+++ b/src/fides/api/graph/traversal.py
@@ -7,7 +7,11 @@ import pydash.collections
 from fideslang.validation import FidesKey
 from loguru import logger
 
-from fides.api.common_exceptions import TraversalError
+from fides.api.common_exceptions import (
+    TraversalError,
+    UnreachableEdgesError,
+    UnreachableNodesError,
+)
 from fides.api.graph.config import (
     ROOT_COLLECTION_ADDRESS,
     TERMINATOR_ADDRESS,
@@ -378,7 +382,6 @@ class Traversal:
                 raise TraversalError(
                     f"""Node could not be reached given the specified ordering:
                     [{','.join([str(tn.address) for tn in running_node_queue.data])}]""",
-                    [str(tn.address) for tn in running_node_queue.data],
                 )
 
         # Remove nodes that have custom request fields, since we don't care if these are reachable or not.
@@ -397,7 +400,7 @@ class Traversal:
                 "Some nodes were not reachable: {}",
                 ",".join([str(x) for x in remaining_node_keys]),
             )
-            raise TraversalError(
+            raise UnreachableNodesError(
                 f"Some nodes were not reachable: {','.join([str(x) for x in remaining_node_keys])}",
                 [key.value for key in remaining_node_keys],
             )
@@ -408,7 +411,7 @@ class Traversal:
                 "Some edges were not reachable: {}",
                 ",".join([str(x) for x in remaining_edges]),
             )
-            raise TraversalError(
+            raise UnreachableEdgesError(
                 f"Some edges were not reachable: {','.join([str(x) for x in remaining_edges])}",
                 [f"{edge}" for edge in remaining_edges],
             )
--- a/src/fides/api/graph/traversal.py
+++ b/src/fides/api/graph/traversal.py
@@ -377,7 +377,8 @@ class Traversal:
                 )
                 raise TraversalError(
                     f"""Node could not be reached given the specified ordering:
-                    [{','.join([str(tn.address) for tn in running_node_queue.data])}]"""
+                    [{','.join([str(tn.address) for tn in running_node_queue.data])}]""",
+                    [str(tn.address) for tn in running_node_queue.data],
                 )
 
         # Remove nodes that have custom request fields, since we don't care if these are reachable or not.
--- a/src/fides/api/graph/traversal.py
+++ b/src/fides/api/graph/traversal.py
@@ -397,7 +397,8 @@ class Traversal:
                 ",".join([str(x) for x in remaining_node_keys]),
             )
             raise TraversalError(
-                f"Some nodes were not reachable: {','.join([str(x) for x in remaining_node_keys])}"
+                f"Some nodes were not reachable: {','.join([str(x) for x in remaining_node_keys])}",
+                [key.value for key in remaining_node_keys],
             )
 
         # error if there are edges that have not been visited
@@ -407,7 +408,8 @@ class Traversal:
                 ",".join([str(x) for x in remaining_edges]),
             )
             raise TraversalError(
-                f"Some edges were not reachable: {','.join([str(x) for x in remaining_edges])}"
+                f"Some edges were not reachable: {','.join([str(x) for x in remaining_edges])}",
+                [f"{edge}" for edge in remaining_edges],
             )
 
         end_nodes = [
--- a/src/fides/api/graph/traversal.py
+++ b/src/fides/api/graph/traversal.py
@@ -6,6 +6,7 @@ from typing import Any, Callable, Dict, List, Set, Tuple, cast
 import pydash.collections
 from fideslang.validation import FidesKey
 from loguru import logger
+from sqlalchemy.orm import Session
 
 from fides.api.common_exceptions import (
     TraversalError,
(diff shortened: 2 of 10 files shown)


src/fides/api/task/deprecated_graph_task.py
```diff
@@ -108,7 +110,26 @@ def run_access_request_deprecated(
     session: Session,
 ) -> Dict[str, List[Row]]:
     """Deprecated: Run the access request sequentially in-memory using Dask"""
-    traversal: Traversal = Traversal(graph, identity)
+    try:
+        traversal: Traversal = Traversal(graph, identity)
+    except TraversalError as err:
```
pattisdr, line 115: Nice tracking this down for DSR 2.0 👍, I have similar concerns with the creation of the execution log

src/fides/api/task/create_request_tasks.py
```diff
@@ -429,41 +429,59 @@ def run_access_request(
...
+            privacy_request.add_error_execution_log(
+                session,
+                connection_key=errors,
+                dataset_name=errors,
+                collection_name=errors,
+                message=f"{err}",
+                action_type=ActionType.access,
+            )
+            privacy_request.error_processing(session)
+            raise err
```
pattisdr, line 484: Did you test this for DSR 3.0 as well? Your instructions are for DSR 2.0 -

This involves setting the following env vars:
```
[execution]
use_dsr_3_0 = true
```

```
[celery]
task_always_eager = false
```

and then running:
```
nox -s dev -- shell postgres worker
```

I keep running into a separate issue on the worker container, I want to make sure this isn't just me:

```
./../data/language/languages.yml from /fides/.fides/fides.toml
> Loaded config from: /fides/.fides/fides.toml
Connection failed, webserver is unreachable at URL:
http://localhost:8080/health
> Loaded config from: /fides/.fides/fides.toml
Connection failed, webserver is unreachable at URL:
http://localhost:8080/health
```

src/fides/api/graph/traversal.py
```diff
@@ -387,7 +387,8 @@ def edge_ends_with_collection(_edge: Edge) -> bool:
                 ",".join([str(x) for x in remaining_node_keys]),
             )
             raise TraversalError(
-                f"Some nodes were not reachable: {','.join([str(x) for x in remaining_node_keys])}"
+                f"Some nodes were not reachable: {','.join([str(x) for x in remaining_node_keys])}",
+                [key.value for key in remaining_node_keys],
             )
```
pattisdr, line 392: Why is this formatting and the one below changed?

src/fides/api/task/create_request_tasks.py
```diff
@@ -429,41 +429,59 @@ def run_access_request(
...
+                "TraversalError encountered for privacy request {}. Error: {}",
+                privacy_request.id,
+                err,
+            )
+            errors = ", ".join(err.errors)
+            privacy_request.add_error_execution_log(
+                session,
+                connection_key=errors,
+                dataset_name=errors,
+                collection_name=errors,
```
pattisdr, line 479: Your dataset name and collection name are incorrect here, same for the other example. I understand that this error is being thrown outside of a typical node, and there could be many dataset names and collection names, but I don't think we should try to cram everything in here.

In this case, instead of creating an "execution log", I might suggest an "audit log" which we've used in the past for errors on privacy requests that pertain to multiple nodes.



```
fides=# select dataset_name, collection_name from executionlog;
       dataset_name        |      collection_name      
---------------------------+---------------------------
 prod-2644-example:address | prod-2644-example:address
 prod-2644-example:address | prod-2644-example:address

```

pattisdr (changes requested):
@erosselli good start, also could you add some tests please?

src/fides/api/graph/traversal.py
erosselli: I don't think it's just the formatting, I added the failing node keys as the list of errors so I could get them when catching the exception

src/fides/api/task/create_request_tasks.py
erosselli: ah I hadn't seen these audit logs , I'll look into that! 😄

pattisdr: i'm not sure if it's exactly what you need my concern is that a list of strings under dataset_name or collection _name wasn't the intent of this column, but if audit log is also not a good fit, we can pick the closest one

erosselli: yeah I can take a look. I also thought of adding one ExecutionLog per failing node, but was afraid that it would crowd the UI a bit. Do you think that'd be a better approach?

pattisdr: That's a good idea as well, I think that would generally work unless that had hundreds+ unconnected nodes. I guess my preference might be a 1) single audit log 2) execution log per failing node 3) single execution log with just the first unconnected node 3) single execution log with all the unconnected nodes.

src/fides/api/common_exceptions.py
```diff
@@ -25,6 +25,14 @@ class TraversalError(FidesopsException):
     """Fidesops error with the names of all nodes that could not be reached."""
 
 
+class UnreachableNodesError(TraversalError):
+    """Fidesops error with the names of all nodes that could not be reached."""
+
+
+class UnreachableEdgesError(TraversalError):
+    """Fidesops error with the names of all edges that could not be reached."""
+
```
erosselli, line 34: These both inherit from `TraversalError` so any code that's currently catching the `TraversalError` exception should continue to work as before. I added these more specific ones just to be able to know the specific reason for the traversal error, in order to show a more accurate error message in the `ExecutionLog`

src/fides/cli/commands/ungrouped.py
```diff
@@ -183,7 +182,6 @@ def webserver(ctx: click.Context, port: int = 8080) -> None:
 @click.command()
 @click.pass_context
 @with_analytics
-@with_server_health_check
```
pattisdr, line 186: Great troubleshooting @erosselli!

src/fides/api/graph/traversal.py
pattisdr: ah! I missed the fidesops exception takes in a string as well as a list of errors!

src/fides/api/common_exceptions.py
```diff
@@ -25,6 +25,14 @@ class TraversalError(FidesopsException):
     """Fidesops error with the names of all nodes that could not be reached."""
 
 
+class UnreachableNodesError(TraversalError):
+    """Fidesops error with the names of all nodes that could not be reached."""
```
pattisdr, line 29: Minor "Fidesops" was the repo that used to process DSR's versus Fidesctl that did more database scanning and D&D. Now they're all unified under one "Fides" but there's still fidesops/fidesctl references still lingering

src/fides/api/task/deprecated_graph_task.py
```diff
@@ -108,7 +114,49 @@ def run_access_request_deprecated(
...
+                session,
+                connection_key=None,
+                dataset_name=dataset,
+                collection_name=collection,
+                message=message,
+                action_type=ActionType.access,
+            )
+        privacy_request.error_processing(session)
+        raise err
+
```
pattisdr, line 159: Can we dry this up into a function that is shared between DSR 2.0 and DSR 3.0?

pattisdr (commented):
great improvement, traversal errors have been something that has been buried for a long time - I think test coverage is good enough here.  The "edges" case is not going to be hit anyway.  Just a quick request to dry some of the code up -

pattisdr (approved):
Looks good, approved with passing tests :shipit:

PR #5271, opened 2024-09-09
Attentive suggestions

galvana:
### Description Of Changes

Example of how to handle optional identities + misc cleanup suggestions

### DIFF
--- a/data/saas/config/attentive_config.yml
+++ b/data/saas/config/attentive_config.yml
@@ -23,15 +23,15 @@ saas_config:
     path: /v1/me
 
   endpoints:
-    - name: delete_request
+    - name: user
       requests:
         delete:
           method: POST
           path: /v1/privacy/delete-request
           body: |
             {
-             "subjectEmail": "<email>",
-             "subjectPhone": "<phone_number>",
+             "subjectEmail": "<email?>",
+             "subjectPhone": "<phone_number?>",
              "requestMsg": "Delete Request managed by Fides API"
             }
           param_values:
--- a/data/saas/dataset/attentive_dataset.yml
+++ b/data/saas/dataset/attentive_dataset.yml
@@ -2,34 +2,4 @@ dataset:
   - fides_key: <instance_fides_key>
     name: attentive
     description: A sample dataset representing the Attentive connector for Fides
-    collections:
-      - name: delete_request
-        fields:
-          - name: id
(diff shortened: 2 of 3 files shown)


data/saas/config/attentive_config.yml
```diff
@@ -23,15 +23,15 @@ saas_config:
     path: /v1/me
 
   endpoints:
-    - name: delete_request
+    - name: user
```
galvana, line 26: Same as the feedback on PowerReviews, we want the endpoint name to reflect the model type that this represents, in this case `user`.

data/saas/config/attentive_config.yml
```diff
@@ -23,15 +23,15 @@ saas_config:
...
       requests:
         delete:
           method: POST
           path: /v1/privacy/delete-request
           body: |
             {
-             "subjectEmail": "<email>",
-             "subjectPhone": "<phone_number>",
+             "subjectEmail": "<email?>",
+             "subjectPhone": "<phone_number?>",
```
galvana, line 34: We have this `?` notation in our placeholders to indicate that this value might be `null` but it's ok! I added this here so there aren't any errors if either `email` or `phone_number` are missing

data/saas/dataset/attentive_dataset.yml
```diff
@@ -2,34 +2,4 @@ dataset:
...
-              data_type: string
-          - name: processingStartDateTime
-            data_categories: [system.operations]
-            fidesops_meta:
-              data_type: string
-          - name: processingEndDateTime
-            data_categories: [system.operations]
-            fidesops_meta:
-              data_type: string
+    collections: []
```
galvana, line 5: We can just set `collections` to `[]` if we don't have a `read` request

tests/ops/integration_tests/test_attentive_task.py
```diff
@@ -25,8 +24,25 @@ async def test_non_strict_erasure_request(
             erasure_policy=erasure_policy_string_rewrite,
             identities={
                 "email": attentive_erasure_identity_email,
+            },
+        )
+        assert erasure_results == {"attentive_instance:user": 1}
+
+    async def test_non_strict_erasure_request_with_phone_number(
```
galvana, line 31: I added two separate tests to verify that the delete will work with either email or phone number

Vagoasdf (approved):

PR #5275, opened 2024-09-10
PROD-2654 - MySQL on RDS as a detection/discovery source

andres-torres-marroquin:
Closes #PROD-2654

### Description Of Changes

Added model for MySQL on RDS.

### Code Changes

* [ ]  Added connection_secrets_rds_mysql.py
* [ ]  Added migration for the new type on connection

### Steps to Confirm

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### DIFF
--- a/tests/ops/api/v1/endpoints/test_connection_template_endpoints.py
+++ b/tests/ops/api/v1/endpoints/test_connection_template_endpoints.py
@@ -1350,14 +1350,6 @@ class TestGetConnectionSecretSchema:
                     "title": "Secret Access Key",
                     "type": "string",
                 },
-                "ca_cert_url": {
-                    "default": "https://truststore.pki.rds.amazonaws.com/global/global-bundle.pem",
-                    "description": "The URL to the CA certificate "
-                    "used to authenticate the RDS "
-                    "instances.",
-                    "title": "CA Certificate URL",
-                    "type": "string",
-                },
                 "region": {
                     "description": "The AWS region where the RDS "
                     "instances are located.",


src/fides/api/schemas/connection_configuration/connection_secrets_rds_mysql.py
```diff
@@ -0,0 +1,35 @@
...
+)
+
+
+class RDSMySQLSchema(BaseAWSSchema):
+    """
+    Schema to validate the secrets needed to connect to a RDS MySQL Database
+    """
+
+    username: str = Field(
+        default="fides_explorer",
```
erosselli, line 17: Looking at connection secrets for other integrations, we don't set defaults for `username` fields... in what way is this connection different?

src/fides/api/schemas/connection_configuration/connection_secrets_rds_mysql.py
```diff
@@ -0,0 +1,35 @@
...
+        title="Region",
+        description="The AWS region where the RDS instances are located.",
+    )
+    ca_cert_url: str = Field(
+        default="https://truststore.pki.rds.amazonaws.com/global/global-bundle.pem",
+        title="CA Certificate URL",
+        description="The URL to the CA certificate used to authenticate the RDS instances.",
+    )
+
+    _required_components: ClassVar[List[str]] = ["auth_method"]
```
erosselli, line 31: Can you explain a bit what this is used for?

erosselli (commented):
Also maybe we can add a test on `tests/ops/api/v1/endpoints/test_connection_template_endpoints.py::TestGetConnectionSecretSchema`  to test for the added schema

src/fides/api/schemas/connection_configuration/connection_secrets_rds_mysql.py
erosselli: so I'm still a bit confused (perhaps just from the lack of context I have on the feature 😅 ) -- if the username is not named fides_explorer then this won't work? if so why is it configurable?

erosselli (approved):
LGTM!

adamsachs (changes requested):
@andres-torres-marroquin i think we at least need a bit of a stronger type on the cert URL property here, and may need to think a bit more about the security implications - see my comment below

src/fides/api/schemas/connection_configuration/connection_secrets_rds_mysql.py
```diff
@@ -0,0 +1,31 @@
...
+    username: str = Field(
+        default="fides_explorer",
+        title="Username",
+        description="The user account used to authenticate and access the databases.",
+    )
+    region: str = Field(
+        title="Region",
+        description="The AWS region where the RDS instances are located.",
+    )
+    ca_cert_url: str = Field(
```
adamsachs, line 23: knowing what i know about how this field is used downstream in the monitor, and the fact that it's user-input data - i think we need some more validation on this field. at a minimum, we should be using the `AnyHttpUrl` pydantic class (or even better, using @pattisdr 's `AnyHttpUrlString` in `fideslang) to put some bounds on this field.

additionally, we should probably perform some more checks on the URL before we fetch it and save it to the filesystem downstream. i'm not exactly sure the best way to check that, but it feels like it'd be wise...

if you feel like it makes more sense to just hardcode the default value here rather than accept user config for it, that may be OK. i know i'd vouched for a configurable field earlier, since i thought the flexibility would be a good fallback in case a customer isn't using this global AWS cert that we expect, but i hadn't really considered the security implications. sorry to go back and forth on this! i still think the flexibility would be nice, but we need to be able to do it in a reasonably secure way, so let's weigh that accordingly.

if you do decide to remove the flexibility, i think it would be good for us to confirm with our prospective customers of this feature whether the global AWS cert we're defaulting to is one that would work for their use case 👍

src/fides/api/schemas/connection_configuration/connection_secrets_rds_mysql.py
```diff
@@ -0,0 +1,26 @@
...
+)
+
+
+class RDSMySQLSchema(BaseAWSSchema):
+    """
+    Schema to validate the secrets needed to connect to a RDS MySQL Database
+    """
+
+    username: str = Field(
+        default="fides_explorer",
```
adamsachs, line 15: nit: `fides_explorer` feels a bit arbitrary/random to me as a default, we don't use "explorer" anywhere else in the product and it's not really a standard. I'd probably just stick with `fides` or `fides_service_user` or `fides_db_user` or something like that

src/fides/api/schemas/connection_configuration/connection_secrets_rds_mysql.py
```diff
@@ -0,0 +1,26 @@
...
+    BaseAWSSchema,
+)
+
+
+class RDSMySQLSchema(BaseAWSSchema):
+    """
+    Schema to validate the secrets needed to connect to a RDS MySQL Database
+    """
+
+    username: str = Field(
```
adamsachs, line 14: nit: I might rename this `db_username` to clarify it's not an AWS IAM username

adamsachs (approved):
👍

src/fides/api/schemas/connection_configuration/connection_secrets_rds_mysql.py
```diff
@@ -0,0 +1,26 @@
...
+    BaseAWSSchema,
+)
+
+
+class RDSMySQLSchema(BaseAWSSchema):
+    """
+    Schema to validate the secrets needed to connect to a RDS MySQL Database
+    """
+
+    db_sername: str = Field(
```
adamsachs, line 14: @andres-torres-marroquin looks like a typo here. should read `db_username`

PR #5278, opened 2024-09-11
PROD-2722, PROD-2761

adamsachs:
Closes [PROD-2722](https://ethyca.atlassian.net/browse/PROD-2722), [PROD-2761](https://ethyca.atlassian.net/browse/PROD-2761)

### Description Of Changes

- fix PROD-2722: remove default value for the `auth_method` field on the base AWS connection secret type, so it's now correctly populated as a `required` field
    - implement a fallback so that UI doesn't crash if there aren't any `required` fields inferred by the pydantic secrets schema
- fix PROD-2761: set the `region_name` arg in all of our AWS sessions that we instantiate within  our base `get_aws_session` utility function

### Code Changes

* [x] remove default value for the `auth_method` field on the base AWS connection secret type so that it shows up as `required`
* [x] implement a fallback so that UI doesn't crash if there aren't any `required` fields inferred by the pydantic secrets schema
* [x] set the `region_name` arg in all of our AWS sessions that we instantiate within  our base `get_aws_session` utility function 

### Steps to Confirm

* [x] spin up fidesplus pointed at this commit and ensure that the s3 integration page loads successfully
image

* [x] test assume role workflow with dynamodb monitor - connection test works! (trust me that secret keys + an assume role ARN are configured here! 😄)
image

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

[PROD-2722]: https://ethyca.atlassian.net/browse/PROD-2722?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[PROD-2761]: https://ethyca.atlassian.net/browse/PROD-2761?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/fides/api/schemas/connection_configuration/connection_secrets_base_aws.py
+++ b/src/fides/api/schemas/connection_configuration/connection_secrets_base_aws.py
@@ -14,7 +14,6 @@ class BaseAWSSchema(ConnectionConfigSecretsSchema):
     auth_method: AWSAuthMethod = Field(
         title="Authentication Method",
         description="Determines which type of authentication method to use for connecting to Amazon Web Services. Currently accepted values are: `secret_keys` or `automatic`.",
-        default=AWSAuthMethod.SECRET_KEYS,
     )
 
     aws_access_key_id: Optional[str] = Field(
--- a/src/fides/api/util/aws_util.py
+++ b/src/fides/api/util/aws_util.py
@@ -34,11 +34,13 @@ def get_aws_session(
             region_name=storage_secrets.get("region_name"),  # type: ignore
         )
     elif auth_method == AWSAuthMethod.AUTOMATIC.value:
-        session = Session()
+        session = Session(
+            region_name=storage_secrets.get("region_name"),  # type: ignore
+        )
         logger.info("Successfully created automatic session")
     else:
-        logger.error("Auth method not supported for S3: {}", auth_method)
-        raise ValueError(f"Auth method not supported for S3: {auth_method}")
+        logger.error("AWS auth method not supported: {}", auth_method)
+        raise ValueError(f"AWS auth method not supported: {auth_method}")
 
     # Check that credentials are valid
     sts_client = session.client("sts")
@@ -57,6 +59,7 @@ def get_aws_session(
                 aws_access_key_id=temp_credentials["AccessKeyId"],
                 aws_secret_access_key=temp_credentials["SecretAccessKey"],
                 aws_session_token=temp_credentials["SessionToken"],
+                region_name=storage_secrets.get("region_name"),  # type: ignore
             )
         except ClientError as error:
             logger.exception(
--- a/src/fides/api/util/aws_util.py
+++ b/src/fides/api/util/aws_util.py
@@ -19,7 +19,6 @@ def get_aws_session(
     If an `assume_role_arn` is provided, the secrets will be used to
     assume that role and return a Session instantiated with that role.
     """
-    sts_client = None
     if auth_method == AWSAuthMethod.SECRET_KEYS.value:
         if storage_secrets is None:
             err_msg = "Storage secrets not found for S3 storage."
--- a/src/fides/api/util/connection_type.py
+++ b/src/fides/api/util/connection_type.py
@@ -181,6 +181,11 @@ def get_connection_type_secret_schema(*, connection_type: str) -> dict[str, Any]
     }
     if schema.get("additionalProperties"):
         schema.pop("additionalProperties")
+
+    # set an empty array for the 'required' key, if there are no required fields on the schema
+    # this helps the FE, which is expecting _some_ value here, even if it's an empty list
+    if "required" not in schema:
+        schema["required"] = []
     return schema
 
 
--- a/src/fides/api/schemas/connection_configuration/connection_secrets_dynamodb.py
+++ b/src/fides/api/schemas/connection_configuration/connection_secrets_dynamodb.py
@@ -16,7 +16,9 @@ class DynamoDBSchema(BaseAWSSchema):
         description="The AWS region where your DynamoDB table is located (ex. us-west-2).",
     )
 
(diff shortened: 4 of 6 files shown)


src/fides/api/schemas/connection_configuration/connection_secrets_base_aws.py
```diff
@@ -14,7 +14,6 @@ class BaseAWSSchema(ConnectionConfigSecretsSchema):
     auth_method: AWSAuthMethod = Field(
         title="Authentication Method",
         description="Determines which type of authentication method to use for connecting to Amazon Web Services. Currently accepted values are: `secret_keys` or `automatic`.",
-        default=AWSAuthMethod.SECRET_KEYS,
```
adamsachs, line 17: the default value here was making it so that it didn't show up in the `required` array, even though it's a required field. as far as i can tell, there's no two ways about this with pydantic - you can't really declare a field with a default value _and_ consider it required, or you have to consider _all_ default fields as required (which we don't want to do).

the default value here is only a nice to have - we're just guessing on what people prefer, anyway...we don't really know. on the other hand, i think showing this field as `required` is more important, since it's required.

src/fides/api/util/connection_type.py
```diff
@@ -181,6 +181,11 @@ def get_connection_type_secret_schema(*, connection_type: str) -> dict[str, Any]
     }
     if schema.get("additionalProperties"):
         schema.pop("additionalProperties")
+
+    # set an empty array for the 'required' key, if there are no required fields on the schema
+    # this helps the FE, which is expecting _some_ value here, even if it's an empty list
+    if "required" not in schema:
+        schema["required"] = []
```
adamsachs, line 188: as the comment says, this is basically a fallback, since having _no_ `required` field on our response seems to cause problems for the FE...

src/fides/api/util/aws_util.py
```diff
@@ -57,6 +59,7 @@ def get_aws_session(
                 aws_access_key_id=temp_credentials["AccessKeyId"],
                 aws_secret_access_key=temp_credentials["SecretAccessKey"],
                 aws_session_token=temp_credentials["SessionToken"],
+                region_name=storage_secrets.get("region_name"),  # type: ignore
```
adamsachs, line 62: provides a `region_name` on the AWS session in the "assume role" use case

src/fides/api/util/aws_util.py
```diff
@@ -34,11 +34,13 @@ def get_aws_session(
             region_name=storage_secrets.get("region_name"),  # type: ignore
         )
     elif auth_method == AWSAuthMethod.AUTOMATIC.value:
-        session = Session()
+        session = Session(
+            region_name=storage_secrets.get("region_name"),  # type: ignore
```
adamsachs, line 38: provides a region_name on the AWS session in the "automatic" auth use case

src/fides/api/util/aws_util.py
```diff
@@ -34,11 +34,13 @@ def get_aws_session(
...
+        )
         logger.info("Successfully created automatic session")
     else:
-        logger.error("Auth method not supported for S3: {}", auth_method)
-        raise ValueError(f"Auth method not supported for S3: {auth_method}")
+        logger.error("AWS auth method not supported: {}", auth_method)
+        raise ValueError(f"AWS auth method not supported: {auth_method}")
```
adamsachs, line 43: just cleaning up some logging

adamsachs (commented):
some self-review

src/fides/api/util/aws_util.py
pattisdr: great

src/fides/api/util/aws_util.py
```diff
@@ -34,11 +34,13 @@ def get_aws_session(
             region_name=storage_secrets.get("region_name"),  # type: ignore
         )
     elif auth_method == AWSAuthMethod.AUTOMATIC.value:
-        session = Session()
```
pattisdr, line 37: unused sts_client definition above ^

src/fides/api/util/aws_util.py
```diff
@@ -34,11 +34,13 @@ def get_aws_session(
             region_name=storage_secrets.get("region_name"),  # type: ignore
```
pattisdr, line 34: Ah right I'd seen this region_name on a quick skim - different use case!

src/fides/api/util/connection_type.py
pattisdr: when did you run into this?

pattisdr (approved):
Tested against the various use cases with automatic and the assume role case, as well as testing connections. All worked as expected, thank you for prioritizing this!

src/fides/api/util/connection_type.py
pattisdr: ah it's the other ticket! https://ethyca.atlassian.net/browse/PROD-2722 I'm caught up

pattisdr: This is that not-so-great endpoint where we were returning the pydantic v1 schema and now there's some changes to try to backport the pydantic v2 schema to return the pydantic v1 version so the UI doesn't have to change. Ideally we'd rework the backend and frontend together at some point.

src/fides/api/util/aws_util.py
adamsachs: good catch, thanks! fixed 👍

src/fides/api/util/connection_type.py
adamsachs: makes sense, agreed that it'd be nice to refactor!

PR #5294, opened 2024-09-17
Adding namespace fides_meta support for BigQuery datasets

galvana:
Closes [PROD-2735](https://ethyca.atlassian.net/browse/PROD-2735)

### Description Of Changes

Updates the `BigQueryConnector` to use a namespaced query config. This allows queries to specify the correct project and dataset IDs without relying on the dataset value from the `connectionconfig`.

### Code Changes

* [ ] Updated the dataset label for the BigQuery connector form to read `Default BigQuery Dataset`
* [ ] Added the `BigQueryNamespaceMeta` schema (`project_id` and `dataset_id`)
* [ ] Updated `BigQueryConnector` and `BigQueryQueryConfig`
* [ ] Updated the BigQuery tests to run access and erasures with
  * [ ] Connection-level dataset ID (previous functionality)
  * [ ] Dataset-level project and dataset IDs (new functionality)

### Steps to Confirm

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

[PROD-2735]: https://ethyca.atlassian.net/browse/PROD-2735?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/fides/api/schemas/namespace_meta/bigquery_namespace_meta.py
```diff
@@ -0,0 +1,18 @@
+from typing import Optional
+
+from fides.api.schemas.base_class import FidesSchema
+
+
+class BigQueryNamespaceMeta(FidesSchema):
```
galvana, line 6: I'm only using this in a limited scope. It'd be nice to use this to validate datasets when we link them to a specific connection config but I'm removing this from scope for now.

src/fides/api/service/connectors/query_config.py
```diff
@@ -810,14 +813,41 @@ class BigQueryQueryConfig(QueryStringWithoutTuplesOverrideQueryConfig):
     Generates SQL valid for BigQuery
     """
 
+    def __init__(
+        self,
+        node: ExecutionNode,
+        namespace_meta: Optional[BigQueryNamespaceMeta] = None,
```
galvana, line 819: Updating the init to include an optional `namespace_meta` object

src/fides/api/service/connectors/query_config.py
```diff
@@ -810,14 +813,41 @@ class BigQueryQueryConfig(QueryStringWithoutTuplesOverrideQueryConfig):
...
     ) -> str:
-        """Returns a query string with backtick formatting for tables that have the same names as
-        BigQuery reserved words."""
-        return f'SELECT {field_list} FROM `{self.node.collection.name}` WHERE {" OR ".join(clauses)}'
+        """
+        Returns a query string with backtick formatting for tables that have the same names as
+        BigQuery reserved words.
+        """
+
+        return f'SELECT {field_list} FROM `{self._generate_table_name()}` WHERE {" OR ".join(clauses)}'
```
galvana, line 850: Here it is 😄! 

(Mel was over my shoulder as I typed here it is and suggested I add a smiley face)

src/fides/api/service/connectors/sql_connector.py
```diff
@@ -115,6 +124,18 @@ def default_cursor_result_to_rows(results: LegacyCursorResult) -> List[Row]:
             rows.append({col[0]: row_tuple[count] for count, col in enumerate(columns)})
         return rows
 
+    @staticmethod
+    def get_namespace_meta(db: Session, dataset: str) -> Optional[Dict[str, Any]]:
```
galvana, line 128: This is the way we pull in the namespace info into the connector. I thought about caching this for a slight performance improvement but decided against a premature optimization, keeping it simple for now

src/fides/api/service/connectors/sql_connector.py
```diff
@@ -115,6 +124,18 @@ def default_cursor_result_to_rows(results: LegacyCursorResult) -> List[Row]:
...
+        """
+        Util function to return the namespace meta for a given ctl_dataset.
+        """
+
+        return db.scalar(
+            select(CtlDataset.fides_meta["namespace"].cast(JSONB)).where(
```
galvana, line 134: Just queries for the `fides_meta` column so we don't return a huge payload from the database for large datasets

src/fides/api/service/connectors/sql_connector.py
```diff
@@ -529,7 +550,14 @@ def create_client(self) -> Engine:
     # Overrides SQLConnector.query_config
     def query_config(self, node: ExecutionNode) -> BigQueryQueryConfig:
         """Query wrapper corresponding to the input execution_node."""
-        return BigQueryQueryConfig(node)
+
+        db: Session = Session.object_session(self.configuration)
+        namespace_meta: Optional[BigQueryNamespaceMeta] = None
+
+        if raw_meta := SQLConnector.get_namespace_meta(db, node.address.dataset):
+            namespace_meta = BigQueryNamespaceMeta(**raw_meta)
```
galvana, line 558: Borrowing the `self.configuration` so we can get a database session

tests/fixtures/bigquery_fixtures.py
galvana, line 1: Updating the fixtures to have:
- a connection config **with** the dataset defined and **no** namespace metadata on the dataset
- a connection config **without** the dataset defined and namespace metadata on the dataset

requirements.txt
```diff
@@ -16,7 +16,7 @@ types-defusedxml==0.7.0.20240218
 expandvars==0.9.0
 fastapi[all]==0.111.0
 fastapi-pagination[sqlalchemy]==0.12.25
-fideslang==3.0.4
+fideslang @ git+https://github.com/ethyca/fideslang.git@0d8c203295d6d427b9274db5d9b8815065bdf75b
```
adamsachs, line 19: explicitly marking this as a TODO so we don't forget to update!

src/fides/api/schemas/namespace_meta/bigquery_namespace_meta.py
adamsachs: nice, that's an interesting idea. somewhat related to [my comment on the plus PR here](https://github.com/ethyca/fidesplus/pull/1621#discussion_r1769068314)

src/fides/api/service/connectors/sql_connector.py
adamsachs: no this is really solid, i like it 👍

src/fides/api/service/connectors/sql_connector.py
adamsachs, line 558: right so what i'm thinking is that instead of "casting" the raw_meta here, we'd cast it _within_ the scope of the `BigQueryQueryConfig`, such that we could initialize a generic `namespace_meta` attribute _generically_ on the `SQLConnector` base class

adamsachs (commented):
@galvana nice clean implementation here, this looks quite good to me! i do have a comment trying to push the envelope a bit on making this a bit more generic such that we can more easily start building the support for this namespace meta for other datasources, which we know we'll have to do.

it's not a must-have if you'd prefer to get this increment in place as you have it, but it's a tweak that i don't think should be too difficult to make right now, and it'll ensure our subsequent implementations for this feature follow a similar pattern. let me know what you think!

src/fides/api/service/connectors/query_config.py
galvana: This is valid, I went ahead and made this change. Let me know what you think of my implementation

src/fides/api/service/connectors/sql_connector.py
galvana: I updated it to this
```
db: Session = Session.object_session(self.configuration)
        return BigQueryQueryConfig(
            node, SQLConnector.get_namespace_meta(db, node.address.dataset)
        )
```
The `SQLLikeQueryConfig` super class validates the meta dict based on the `namespace_meta_schema` defined at the `BigQueryQueryConfig`
```
class SQLLikeQueryConfig(QueryConfig[T], ABC):
    """
    Abstract query config for SQL-like languages (that may not be strictly SQL).
    """

    namespace_meta_schema: Optional[Type[NamespaceMeta]] = None

    def __init__(self, node: ExecutionNode, namespace_meta: Optional[Dict] = None):
        super().__init__(node)
        self.namespace_meta: Optional[NamespaceMeta] = None

        if namespace_meta is not None:
            if self.namespace_meta_schema is None:
                raise MissingNamespaceSchemaException(
                    f"{self.__class__.__name__} must define a namespace_meta_schema when namespace_meta is provided."
                )
            try:
                self.namespace_meta = self.namespace_meta_schema.model_validate(
                    namespace_meta
                )
            except ValidationError as exc:
                raise ValueError(f"Invalid namespace_meta: {exc}")
```

galvana (commented):
Thanks for the review @adamsachs. Here's my attempt at making some of the namespace meta logic a bit more generic.

src/fides/api/service/connectors/query_config.py
```diff
@@ -832,8 +844,9 @@ def _generate_table_name(self) -> str:
 
         table_name = self.node.collection.name
         if self.namespace_meta:
-            table_name = f"{self.namespace_meta.dataset_id}.{table_name}"
-            if project_id := self.namespace_meta.project_id:
+            bigquery_namespace_meta = cast(BigQueryNamespaceMeta, self.namespace_meta)
```
adamsachs, line 847: nit: this is to satisfy mypy, yeah? i feel like there _should_ be a way to get the `self.namespace_meta` more properly typed using generics, which feels a bit cleaner. i.e., you've already actually cast this to the right strongly-typed class in the constructor, so it just seems odd you need to 'cast' it again here. 

but i also realize that generics can get a bit verbose and convoluted, especially in python...so all good keeping it as you've got it here. probably just my java-trained instincts getting the best of me anyway 😅

adamsachs (approved):
nice touch up @galvana! this looks good to go from my POV 👍

PR #5297, opened 2024-09-18
Add onFidesEvent method for an alternative way to subscribe to Fides events

guncha:
### Description Of Changes

This PR adds an `onFidesEvent` method to the global `Fides` object that subscribes the callback to Fides events. This is equivalent to calling `window.addEventListener(...)` yourself, except that the callback receives the `FidesEventDetail` object instead of a `CustomEvent`. This is needed for the Google Tag Manager template integration as it doesn't allow usage of APIs like `addEventListener` and also doesn't support `CustomEvent` or, really, any other non-primitive type.

See this [Slack thread](https://ethyca.slack.com/archives/C076JELF2HL/p1726258977000399) for the rationale.

### Code Changes

* [x] Add `onFidesEvent` method to `fides.ts` and `fides-tcf.ts`

### Steps to Confirm

* Load the new `fides.js` version on a page
* Make sure `Fides.onFidesEvent` is available
* Call it with a handler (e.g. `console.log`), trigger the event and verify that the handler was called

### Pre-Merge Checklist

* [x] Update `CHANGELOG.md`

### DIFF
--- a/clients/fides-js/src/lib/events.ts
+++ b/clients/fides-js/src/lib/events.ts
@@ -81,3 +81,21 @@ export const dispatchFidesEvent = (
     window.dispatchEvent(event);
   }
 };
+
+/**
+ * An alternative way to subscribe to Fides events. The same events are supported, except the callback
+ * receives the event details directly. This is useful in restricted environments where you can't
+ * directly access `window.addEventListener`.
+ *
+ * Returns an unsubscribe function that can be called to remove the event listener.
+ */
+export const onFidesEvent = (
+  type: FidesEventType,
+  callback: (evt: FidesEventDetail) => void,
+): (() => void) => {
+  const listener = (evt: FidesEvent) => callback(evt.detail);
+  window.addEventListener(type, listener);
+  return () => {
+    window.removeEventListener(type, listener);
+  };
+};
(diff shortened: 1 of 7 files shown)


clients/fides-js/src/lib/events.ts
```diff
@@ -81,3 +81,21 @@ export const dispatchFidesEvent = (
...
   }
 };
+
+/**
+ * An alternative way to subscribe to Fides events. The same events are supported, except the callback
+ * receives the event details directly. This is useful in restricted environments where you can't
+ * directly access `window.addEventListener`.
+ *
+ * Returns an unsubscribe function that can be called to remove the event listener.
+ */
```
NevilleS, line 91: This is great, but I'd like it on `src/docs/fides.ts` instead so it's added to the official FidesJS SDK docs 👍

guncha:
Btw, there's an issue with `typedoc` where adding a `@returns` tag will remove the description so I opted to keep it instead.

clients/fides-js/src/lib/events.ts
```diff
@@ -81,3 +81,21 @@ export const dispatchFidesEvent = (
     window.dispatchEvent(event);
   }
 };
+
+/**
+ * An alternative way to subscribe to Fides events. The same events are supported, except the callback
+ * receives the event details directly. This is useful in restricted environments where you can't
+ * directly access `window.addEventListener`.
+ *
+ * Returns an unsubscribe function that can be called to remove the event listener.
```
eastandwestwind, line 90: can we add a usage example here, e.g `Fides.onFidesEvent("FidesUpdated", (data) => console.log(data))`

 Were you able to get this added in the `docs/fides.ts` file, too?

eastandwestwind (approved):
Small change request but nothing blocking. Manually tested this and was able to get the event details surfaced using the new `onFidesEvent` method.

Thanks @guncha !

clients/fides-js/src/lib/events.ts
guncha: Usage example with docs added 👍

eastandwestwind (approved):

PR #5298, opened 2024-09-18
Support Accept/Reject All with minimal TCF

gilluminate:
Closes [PROD-2777](https://ethyca.atlassian.net/browse/PROD-2777)

### Description Of Changes

Today, the “minimal” TCF banner can display the banner but the “Accept All” and “Reject All” actions are not enabled until the “full” experience is loaded. This can take several seconds to do on a low-speed connection, which creates a window where users can see a banner but are unable to opt-in.

These changes are set up for handling a new scenario in which a minimal version of `experience.gvl` is provided as part of the minimal TCF experience response.

If no GVL is included, we'll at least show some spinners during the loading of the full experience to not appear "broken" during that time.

### Code Changes

* [x] New Spinner component (very small, css only)
* [x] Show spinner on the Modal button (Manage Preferences) while the full experience finishes loading
* [x] Show spinner on Accept/Reject buttons if the gvl object is not included with the minimal experience (for backwards compatibility mostly)
* [x] Update/create several of the utilites that support "Accept All" from the full experience to also support it from the minimal experience, which only supplies IDs
* [x] Remove any blockers from letting users Accept/Reject all from the minimal experience

### Steps to Confirm

1. visit the Fides.JS demo site with TCF enabled (eg. http://localhost:3001/fides-js-demo.html?geolocation=eea)
2. In browser's dev tools, enable throttling in the Network tab to mimic a slow network
3. Clear any existing cookie and reload the page
  * If the response is slow enough with the throttling, you should see a spinner on the Manage Preferences button until the full experience has loaded
  * If the backend task for including a minimal gvl isn't finished yet, you should see a spinner on all 3 buttons while the full experience loads
  * If you attempt to click Accept All or Reject all before the full experience has loaded (the spinner is still spinning), it should work as expected and close the banner. This behavior should be exactly the same as clicking the respective buttons after the full experience loads. (Cookie set with same fides_string, payload for saving the preference api should be the same, CMP Validator should report the same)

### Pre-Merge Checklist

* [x] Update `CHANGELOG.md`

[PROD-2777]: https://ethyca.atlassian.net/browse/PROD-2777?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ The types of changes are:
 - Add ability to edit dataset YAML from dataset view [#5262](https://github.com/ethyca/fides/pull/5262)
 - Added support for "in progress" status in classification [#5248](https://github.com/ethyca/fides/pull/5248)
 - Clarify GCP service account permissions when setting up Google Cloud SQL for Postgres in Admin-UI [#5245](https://github.com/ethyca/fides/pull/5266)
+- Support minimal GVL in minimal TCF response allowing Accept/Reject from banner before full GVL is loaded [#5298](https://github.com/ethyca/fides/pull/5298)
 
 ### Changed
 - Validate no path in `server_host` var for CLI config; if there is one then take only up until the first forward slash
--- a/clients/fides-js/rollup.config.mjs
+++ b/clients/fides-js/rollup.config.mjs
@@ -15,7 +15,7 @@ const GZIP_SIZE_ERROR_KB = 45; // fail build if bundle size exceeds this
 const GZIP_SIZE_WARN_KB = 35; // log a warning if bundle size exceeds this
 
 // TCF
-const GZIP_SIZE_TCF_ERROR_KB = 85;
+const GZIP_SIZE_TCF_ERROR_KB = 86;
 const GZIP_SIZE_TCF_WARN_KB = 75;
 
 const preactAliases = {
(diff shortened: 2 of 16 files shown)


Kelsey-Ethyca:
> If the response is slow enough with the throttling, you should see a spinner on the Manage Preferences button until the full experience has loaded

This is working, I see the spinner. We don't want to draw attention to the Manage Preferences button though since we want them opting in. I suggest we only show the spinner if the button is clicked before layer two is loaded. 

> If the backend task for including a minimal gvl isn't finished yet, you should see a spinner on all 3 buttons while the full experience loads

I was never able to see a spinner on all three?

> If you attempt to click Accept All or Reject all before the full experience has loaded (the spinner is still spinning), it should work as expected and close the banner. This behavior should be exactly the same as clicking the respective buttons after the full experience loads. 

The spinner was not showing but I could see the network calls had not completed when I clicked accept all, the banner closed, and the ticket was created!

gilluminate:
@Kelsey-Ethyca 

> I was never able to see a spinner on all three?

Are you using the backend with Dawn's latest changes or not yet?

Kelsey-Ethyca:
> @Kelsey-Ethyca
> 
> > I was never able to see a spinner on all three?
> 
> Are you using the backend with Dawn's latest changes or not yet?

I am!

gilluminate:
> > @Kelsey-Ethyca
> > > I was never able to see a spinner on all three?
> > 
> > 
> > Are you using the backend with Dawn's latest changes or not yet?
> 
> I am!

Cool. Then you shouldn't see all 3 spinners. That was for backwards compatibility :)

Kelsey-Ethyca (approved):
Beautiful!

PR #5306, opened 2024-09-20
Fix cached translation issues with Minimal TCF experience

gilluminate:
Closes [PROD-2764](https://ethyca.atlassian.net/browse/PROD-2764)

### Description Of Changes

When the fides.js script gets cached, but a user changes their browser's language, we get into a weird state with multiple languages (especially after the GVL languages load correctly).

### Code Changes

* Update fides.js response headers to include `Accept_Language` in the `Vary` to instruct caches to reset when language changes
* Address a few inconsistencies left over from when we switched to Minimal TCF
* Determine if Minimal TCF Experience translation matches the user's locale (in the case of caching or `window.fides_overrides.fides_locale`) and load the available translation until full TCF Experience is ready with the correct language.

### Steps to Confirm

1. Throttle network in browser dev tools to help see when things are loading slower
2. Load demo page with TCF enabled (eg. http://localhost:3001/fides-js-demo.html?geolocation=eea)
3. Everything should load normally. Do not open the modal or save anything. We don't want to set a cookie yet.
4. Change browser's preferred language in browser settings to something other than English that is available on your configured TCF experience (eg. French)
5. Make sure "Disable Cache" is not active in your browser's network tab while throttling
6. reload the page (not a hard reload, we want to test the browser cache)
7. Page should load with the selected language in-spite of fides.js having been cached already (it should ignore the cache with the browser setting change).
8. Revert browser to english. You can clear cookies and cache now.
9. Update `fides-js-demo.html`:
Replace:
```
// window.fides_overrides = {
//   fides_embed: true,
// };
```
With:
```
window.fides_overrides = {
  fides_locale: "fr",
};
```
10. Reload the page now
11. Since the server isn't aware of the page's `window` setting, it will first load the English banner and the language picker's globe icon should immediately start spinning to indicate a language is loading. Once the full experience loads, it should replace English with the configured language of French. This is just an edge case we need to handle gracefully.

### Pre-Merge Checklist

* [ ] Update `CHANGELOG.md`

[PROD-2764]: https://ethyca.atlassian.net/browse/PROD-2764?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/clients/fides-js/src/components/tcf/TcfOverlay.tsx
+++ b/clients/fides-js/src/components/tcf/TcfOverlay.tsx
@@ -107,33 +107,29 @@ export const TcfOverlay = ({
   const [experience, setExperience] = useState<PrivacyExperience>();
 
   useEffect(() => {
-    // We need to separately track if the experience and GVL translations are loading
-    // since they are loaded asynchronously and we need to know when *both* are done
-    // in order to set the i18n loading state.
-    let isExperienceLoading = false;
     let isGVLLangLoading = false;
 
-    if (!!userlocale && userlocale !== minExperienceLocale) {
+    if (!!userlocale && bestLocale !== minExperienceLocale) {
       // The minimal experience translation is different from the user's language.
       // This occurs when the customer has set their overrides on the window object
       // which isn't available to us until the experience is fetched or when the
       // browser has cached the experience from a previous userLocale. In these cases,
       // we'll get the translations for the banner from the full experience.
+      debugLog(
+        options.debug,
+        `Best locale does not match minimal experience locale (${minExperienceLocale})\nLoading translations from full experience = ${bestLocale}`,
+      );
       setIsI18nLoading(true);
     }
-    if (!!userlocale && userlocale !== DEFAULT_LOCALE) {
+    if (!!userlocale && bestLocale !== DEFAULT_LOCALE) {
       // We can only get English GVL translations from the experience.
       // If the user's locale is not English, we need to load them from the api.
-      setIsI18nLoading(true);
+      // This only affects the modal.
       isGVLLangLoading = true;
       loadGVLTranslations(bestLocale).then(() => {
         isGVLLangLoading = false;
-        if (!isExperienceLoading) {
-          setIsI18nLoading(false);
-        }
       });
     }
-    isExperienceLoading = true;
     fetchExperience({
       userLocationString: fidesRegionString,
       fidesApiUrl: options.fidesApiUrl,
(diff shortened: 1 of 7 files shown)


clients/fides-js/src/components/notices/NoticeOverlay.tsx
```diff
@@ -75,12 +75,6 @@ const NoticeOverlay: FunctionComponent<OverlayProps> = ({
     return [];
   };
 
-  useEffect(() => {
-    if (!currentLocale && i18n.locale) {
-      setCurrentLocale(i18n.locale);
-    }
-  }, [currentLocale, i18n.locale, setCurrentLocale]);
```
gilluminate, line 82: i18n-context was updated which made this unnecessary

clients/fides-js/src/components/tcf/TcfOverlay.tsx
```diff
@@ -55,28 +57,21 @@ export const TcfOverlay = ({
...
     i18n.getDefaultLocale(),
   );
 
-  useEffect(() => {
-    if (!currentLocale) {
-      // initialize the i18n locale using the minimal experience
-      setCurrentLocale(minExperienceLocale);
-    }
-  }, [currentLocale, minExperienceLocale, setCurrentLocale]);
-
```
gilluminate, line 79: i18n-context was updated which made this unnecessary

clients/fides-js/src/components/tcf/TcfOverlay.tsx
```diff
@@ -100,33 +95,29 @@ export const TcfOverlay = ({
   const [experience, setExperience] = useState<PrivacyExperience>();
 
   useEffect(() => {
-    // We need to track if the experience and GVL translations are loading separately
-    // because they are loaded asynchronously and we need to know when both are done
-    // in order to set the i18n loading state.
-    let isExperienceLoading = false;
```
gilluminate, line 106: This was causing the globe to spin unnecessarily while GVL translations loaded in the background. GVL is only needed for Modal, so this is overkill on the banner.

clients/fides-js/src/lib/i18n/i18n-context.tsx
```diff
@@ -28,17 +27,8 @@ interface I18nProviderProps {
...
   const [isLoading, setIsLoading] = useState<boolean>(false);
 
-  const onSetCurrentLocale = useCallback(
-    (locale: string) => {
-      i18nInstance.activate(locale);
-      setCurrentLocale(locale);
-    },
-    [i18nInstance, setCurrentLocale],
-  );
-
```
gilluminate, line 41: This was all overkill and was causing some problems. Just use the ol' `i18n.locale` and `i18n.activate` as intended

PR #5230, opened 2024-08-23
Support the “minimal” TCF experience API and dynamically fetch “full” API response after banner is shown

gilluminate:
Closes [PROD-2608](https://ethyca.atlassian.net/browse/PROD-2608) and [PROD-2611](https://ethyca.atlassian.net/browse/PROD-2611)

### Description Of Changes

* fetch “minimal” TCF experience API in privacy center when bundling FidesJS
* modify TCF banner to render only using the minimal API response fields
* remove display of “stacks” entirely, and render purposes & special features from the minimal API

### Code Changes

* [ ] _list your code changes here_

### Steps to Confirm

* Full regression test on FidesJS is recommended.
* No real change to functionality other than the speed in which the banner appears.
* Be sure to check TCF banner UI in both desktop and mobile views.

### Pre-Merge Checklist

* [ ] All CI Pipelines Succeeded
* Documentation:
  * [ ] documentation complete, [PR opened in fidesdocs](https://github.com/ethyca/fidesdocs/pulls)
  * [ ] documentation [issue created in fidesdocs](https://github.com/ethyca/fidesdocs/issues/new/choose)
  * [ ] if there are any new client scopes created as part of the pull request, remember to update public-facing documentation that references our scope registry
* [ ] Issue Requirements are Met
* [ ] Relevant Follow-Up Issues Created
* [ ] Update `CHANGELOG.md`
* [ ] For API changes, the [Postman collection](https://github.com/ethyca/fides/blob/main/docs/fides/docs/development/postman/Fides.postman_collection.json) has been updated
* If there are any database migrations:
  * [ ] Ensure that your downrev is up to date with the latest revision on `main`
  * [ ] Ensure that your `downgrade()` migration is correct and works
    * [ ] If a downgrade migration is not possible for this change, please call this out in the PR description!

[PROD-2608]: https://ethyca.atlassian.net/browse/PROD-2608?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PROD-2611]: https://ethyca.atlassian.net/browse/PROD-2611?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #5309, opened 2024-09-20
revert to undefined instead of empty object

gilluminate:
Closes #<issue>

### Description Of Changes

In version 2.44 of the Privacy Center, the value of the default experience was set to `{}` (see https://github.com/ethyca/fides/pull/5230/files#diff-ba7bb0c80dbf370b49c211bf0e888dc34cf0f7a713e2b4b685d98b0b46b25014R155-R158).

This causes problems specifically for clients not using prefetch and CDNs as now Fides.js believe that the user received a valid, albeit empty experience and then does not call the Fides API to check for experiences.

Reverting back to `undefined` for that reason.

### Code Changes

* initialize `experience` as `undefined`

### Pre-Merge Checklist

* [x] Update `CHANGELOG.md`

### DIFF
--- a/clients/privacy-center/pages/api/fides-js.ts
+++ b/clients/privacy-center/pages/api/fides-js.ts
@@ -149,13 +149,21 @@ export default async function handler(
     return;
   }
 
-  // If a geolocation can be determined, "prefetch" the experience from the Fides API immediately.
-  // This allows the bundle to be fully configured server-side, so that the Fides.js bundle can initialize instantly!
-
+  /**
+   * NOTE: initializing `experienc` as an empty object `{}` causes problems specifically
+   * for clients not using prefetch and CDNs as now Fides.js believe that the user
+   * received a valid, albeit empty experience and then does not call the Fides API to
+   * check for experiences.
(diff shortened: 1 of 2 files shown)


clients/privacy-center/pages/api/fides-js.ts
```diff
@@ -149,13 +149,20 @@ export default async function handler(
...
+   * for clients not using prefetch and CDNs as Fides.js interprets that as a valid, albeit
+   * empty, experience and then does not make a follow up call to the `privacy-experience` API.
+   * This is why we initialize `experience` as `undefined`.
+   */
   let experience:
     | PrivacyExperience
     | PrivacyExperienceMinimal
-    | EmptyExperience = {};
+    | EmptyExperience
+    | undefined;
```
tvandort, line 162: It could be worthwhile to define `type ExperienceNotFound = undefined;` and use that here like so:

```ts
  let experience:
    | PrivacyExperience
    | PrivacyExperienceMinimal
    | EmptyExperience
    | ExperienceNotFound;
```

I think this increases readability due to making the types more self documenting.

tvandort (approved):

PR #5318, opened 2024-09-23
Updating custom integrations to support multiple identities

galvana:
Closes [PROD-2794](https://ethyca.atlassian.net/browse/PROD-2794)

### Description Of Changes

Updating the Responsys and Firebase Auth integrations to support multiple identities. These two integrations are custom and were relying on some code that is no longer in use.

### Code Changes

* [ ] Updated Responsys and Firebase Auth

### Steps to Confirm

* [ ] These two integrations should not fail if a privacy request is submitted with two identities

### Pre-Merge Checklist

* [ ] Update `CHANGELOG.md`

[PROD-2794]: https://ethyca.atlassian.net/browse/PROD-2794?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/fides/api/service/saas_request/override_implementations/oracle_responsys_request_overrides.py
+++ b/src/fides/api/service/saas_request/override_implementations/oracle_responsys_request_overrides.py
@@ -15,7 +15,7 @@ from fides.api.service.saas_request.saas_request_override_factory import (
     register,
 )
 from fides.api.util.collection_util import Row
-from fides.api.util.saas_util import get_identity
+from fides.api.util.saas_util import get_identities
 
 
 def oracle_responsys_config_parse_profile_lists(list_restrictions: str) -> List[str]:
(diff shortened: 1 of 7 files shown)


src/fides/api/util/saas_util.py
```diff
@@ -433,24 +433,6 @@ def map_param_values(
...
-
-    if not privacy_request:
-        return None
-
-    identity_data: Dict[str, Any] = privacy_request.get_cached_identity_data()
-    # filters out keys where associated value is None or empty str
-    identities = list({k for k, v in identity_data.items() if v})
-    if len(identities) > 1:
-        raise FidesopsException(
-            "Only one identity can be specified for SaaS connector traversal"
```
galvana, line 449: This is the error we're currently seeing, but this isn't accurate since we support multiple identities now

tests/ops/integration_tests/saas/connector_runner.py
```diff
@@ -340,15 +347,16 @@ async def _base_erasure_request(
...
-        if (
-            ActionType.access
-            in SaaSConfig(**self.connection_config.saas_config).supported_actions
-        ):
-            # verify we returned at least one row for each collection in the dataset
-            for collection in self.dataset["collections"]:
-                assert len(
-                    access_results[f"{fides_key}:{collection['name']}"]
-                ), f"No rows returned for collection '{collection['name']}'"
+        if not skip_access_results_check:
```
galvana, line 350: Quick fix to avoid checking for access data. This is necessary since our test account for Responsys isn't going to return access data for every collection.

src/fides/api/service/saas_request/override_implementations/oracle_responsys_request_overrides.py
```diff
@@ -136,42 +136,46 @@ def oracle_responsys_profile_list_recipients_read(
...
-        # Oracle Responsys doesn't support the initial + in phone numbers
-        for idx, query_id in enumerate(query_ids):
-            query_ids[idx] = query_id[1:] if query_id.startswith("+") else query_id
-        query_attribute = "m"
-    else:
-        raise FidesopsException(
-            "Unsupported identity type for Oracle Responsys connector. Currently only `email` and `phone_number` are supported"
-        )
+    for identity in get_identities(privacy_request):
+        if identity == "email":
```
Vagoasdf, line 140: We could extract the code fragment for retrieving the query_ids and query_attributes into its own method for readability, now that we are one level deeper with a for loop.   It would help readability

Vagoasdf (approved):
Looks pretty good.  I just have some suggestions on using functions and splitting, but nothing crucial

PR #5319, opened 2024-09-23
Attentive fix

Vagoasdf:
### Description Of Changes

Currently we have the Attentive email integration as a high level connection type category, and that was causing conflict with lower level connection types, such as the new SaaS Attentive configuration. Since we are transitioning into using SaaS we've updated the reference to attentive_email, in order to avoid name collisions

 Ideally, the Attentive Email integration should be a generic email integration. That should be the next step on the transition

***Note: We Should check our strategy of communication with our clients before updating this, to avoid any breakage***

### Code Changes

* [x] Updated connectionType and connectionConfig attentive references
* [x] Migration for connectionType Attentive

### Steps to Confirm

* [ ] Check that the Attentive Email Integration Works by creating a system for Attentive Email Integration
* [ ] Check that the Attentive Saas Integration Works by creating a system for Attentive Saas Integration. See 1Password for the API KEY

### Pre-Merge Checklist

* [x] Update `CHANGELOG.md`


galvana (changes requested):
It looks like this change has some unwanted change (74 files changed 😬). Can you try fixing or recreating the feature branch to only include the relevant changes?

Vagoasdf:
Looks like something happened with git and it decided to include everything on the main branch that was changed on the previous weeks. Rebased and cleaned it up, it should be ok now

galvana (approved):
The change looks good! I tested it locally and the upgrade/downgrade worked as expected. I just updated the down revision to be up to date with main and made some small formatting/label changes in the attentive_config.yml but the rest looks solid

galvana (changes requested):
Redacting my approval, it looks like we have some failing tests. Let's fix those before we can merge this in

galvana (approved):
Thanks for fixing the tests, this looks good to me!

PR #5320, opened 2024-09-23
Shipstation Integration

Linker44:
Closes #<issue>

### Description Of Changes

This is a PR for the ShipStation Integration

### Code Changes

* [ ] added config and dataset yml files
* [ ] added tests

### Steps to Confirm

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### DIFF
--- a/data/saas/config/shipstation_config.yml
+++ b/data/saas/config/shipstation_config.yml
@@ -1,0 +1,66 @@
+saas_config:
+  fides_key: <instance_fides_key>
+  name: Shipstation
+  type: shipstation
+  description: A sample schema representing the Shipstation connector for Fides
+  version: 0.1.0
+
+  connector_params:
+    - name: api_key
+    - name: api_secret
+
+  external_references:
+    - name: customer_name
+      label: Customer Full Name
+      description: Dataset reference to the location of the customer name
+
+  client_config:
+    protocol: https
+    host: ssapi.shipstation.com
+    authentication:
+      strategy: basic
+      configuration:
+        username: <api_key>
+        password: <api_secret>
+
+  test_request:
+    method: GET
+    path: /customers
+
+  endpoints:
+    - name: get_customer
+      requests:
+        read:
+          method: GET
+          path: /customers/<customer_id>/
+          param_values:
+            - name: customer_id
+              references:
+                - dataset: <instance_fides_key>
+                  field: list_orders.customerId
+                  direction: from
+      after: [<instance_fides_key>.list_orders]
+    - name: list_orders
+      requests:
+        read:
+          method: GET
+          path: /orders
+          headers:
+            - name: Accept
+              value: application/json
+          data_path: orders
+          query_params:
+            - name: customerName
+              value: <customer_name_value>
+            - name: pageSize
+              value: 500
+          param_values:
+            - name: customer_name_value
+              references:
+                - customer_name
+          postprocessors:
+            - strategy: filter
+              configuration:
+                field: customerEmail
+                value:
+                  identity: email
--- a/data/saas/dataset/shipstation_dataset.yml
+++ b/data/saas/dataset/shipstation_dataset.yml
@@ -1,0 +1,375 @@
+dataset:
+  - fides_key: <instance_fides_key>
+    name: shipstation
+    description: A sample dataset representing the Shipstation connector for Fides
+    collections:
+      - name: get_customer
+        fields:
+          - name: customerId
+            data_categories: [system.operations]
+            fidesops_meta:
+              data_type: integer
+          - name: createDate
+            data_categories: [system.operations]
+            fidesops_meta:
+              data_type: string
+          - name: modifyDate
+            data_categories: [system.operations]
+            fidesops_meta:
+              data_type: string
+          - name: name
+            data_categories: [system.operations]
+            fidesops_meta:
+              data_type: string
+          - name: company
+            data_categories: [system.operations]
+            fidesops_meta:
+              data_type: string
+          - name: street1
+            data_categories: [system.operations]
+            fidesops_meta:
+              data_type: string
+          - name: street2
+            data_categories: [system.operations]
+            fidesops_meta:
+              data_type: string
+          - name: city
+            data_categories: [system.operations]
+            fidesops_meta:
+              data_type: string
+          - name: state
+            data_categories: [system.operations]
+            fidesops_meta:
+              data_type: string
+          - name: postalCode
+            data_categories: [system.operations]
+            fidesops_meta:
+              data_type: string
+          - name: countryCode
+            data_categories: [system.operations]
+            fidesops_meta:
+              data_type: string
+          - name: phone
+            data_categories: [system.operations]
+            fidesops_meta:
+              data_type: string
+          - name: email
+            data_categories: [system.operations]
+            fidesops_meta:
+              data_type: string
+          - name: addressVerified
+            data_categories: [system.operations]
+            fidesops_meta:
+              data_type: string
+          - name: marketplaceUsernames
+            fidesops_meta:
+              data_type: "object[]"
+            fields:
+              - name: customerUserId
+                data_categories: [system.operations]
+                fidesops_meta:
+                  data_type: integer
+              - name: customerId
+                data_categories: [system.operations]
+                fidesops_meta:
+                  data_type: integer
+              - name: createDate
+                data_categories: [system.operations]
+                fidesops_meta:
+                  data_type: string
+              - name: modifyDate
+                data_categories: [system.operations]
+                fidesops_meta:
+                  data_type: string
+              - name: marketplaceId
+                data_categories: [system.operations]
+                fidesops_meta:
+                  data_type: integer
+              - name: marketplace
+                data_categories: [system.operations]
+                fidesops_meta:
+                  data_type: string
+              - name: username
+                data_categories: [system.operations]
+                fidesops_meta:
+                  data_type: string
+          - name: tags
+      - name: list_orders
+        fields:
+          - name: orders
+            fidesops_meta:
+              data_type: "object[]"
+            fields:
+              - name: orderId
+                data_categories: [system.operations]
+                fidesops_meta:
+                  data_type: integer
+              - name: orderNumber
+                data_categories: [system.operations]
+                fidesops_meta:
+                  data_type: string
+              - name: orderKey
+                data_categories: [system.operations]
+                fidesops_meta:
+                  data_type: string
+              - name: orderDate
+                data_categories: [system.operations]
+                fidesops_meta:
+                  data_type: string
+              - name: createDate
+                data_categories: [system.operations]
+                fidesops_meta:
+                  data_type: string
+              - name: modifyDate
+                data_categories: [system.operations]
+                fidesops_meta:
+                  data_type: string
+              - name: paymentDate
+              - name: shipByDate
+              - name: orderStatus
+                data_categories: [system.operations]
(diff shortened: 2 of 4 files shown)


data/saas/config/shipstation_config.yml
```diff
@@ -0,0 +1,74 @@
...
+  version: 0.1.1
+
+  connector_params:
+    - name: api_key
+    - name: api_secret
+
+  external_references:
+    - name: customer_id
+      label: customerId Field
+      description: The dataset reference to the Shipstation customerId
```
galvana, line 15: Any UI-facing value should be in "sentence case", where we only capitalize the first letter, acronyms, or proper nouns

```suggestion
      label: Customer ID field
      description: The dataset reference to the Shipstation customer ID
```

data/saas/config/shipstation_config.yml
```diff
@@ -0,0 +1,75 @@
+saas_config:
+  fides_key: <instance_fides_key>
+  name: Shipstation
+  type: shipstation
+  description: A sample schema representing the Shipstation connector for Fides
+  version: 0.1.1
+
+  connector_params:
+    - name: api_key
+    - name: api_secret
```
galvana, line 10: A value like `api_key` will be converted to `Api key` in the UI, but it should be `API key`. To do this, we usually provide the `label` value. Also, any secret or sensitive value needs to be flagged so it's not displayed in the UI or exposed via the API.

```suggestion
  connector_params:
    - name: api_key
       label: API key
    - name: api_secret
       label: API secret
       sensitive: True
```

galvana: We also like to include `domain` and provide a default value. This is so the user can point to a different URL if needed.
```
  - name: domain
    default_value: ssapi.shipstation.com
```

data/saas/config/shipstation_config.yml
```diff
@@ -0,0 +1,75 @@
...
+              references:
+                - dataset: <instance_fides_key>
+                  field: customer.name
+                  direction: from
+            - name: customer_id_value
+              references:
+                - dataset: <instance_fides_key>
+                  field: customer.customerId
+                  direction: from
+        after: ["<instance_fides_key>.customers"]
```
galvana, line 66: We don't need to specify and `after` for access requests. Fides will determine the correct order based on the `param_values` and relationships between the collections. The order of the endpoints also doesn't affect execution, but I like to put the endpoints in the order that they'll be visited. So in this case, define `customer` first, followed by `orders`

src/fides/api/schemas/saas/shared_schemas.py
```diff
@@ -45,6 +45,12 @@ class IdentityParamRef(BaseModel):
     identity: str
 
 
+class DataSetRef(BaseModel):
```
galvana, line 48: Dataset is one word in Fides
```suggestion
class DatasetRef(BaseModel):
```

data/saas/dataset/shipstation_dataset.yml
galvana, line 1: Same as the previous comment, make sure the collections are in the same order as the endpoints in the SaaS config

src/fides/api/schemas/saas/shared_schemas.py
Linker44, line 48: i dont think this is a good name. Technically a reference to any param value will work, not just dataset reference values.

src/fides/api/service/connectors/saas_connector.py
```diff
@@ -310,6 +310,7 @@ def retrieve_data(
                             next_request,
                             privacy_request.get_cached_identity_data(),
                             read_request,
+                            input_data,
```
Linker44, line 313: for now i need to pass down input data to have the populated param values.

src/fides/api/schemas/saas/shared_schemas.py
Linker44: same thing for the field dataset_reference.

src/fides/api/service/connectors/saas_connector.py
```diff
@@ -476,7 +490,9 @@ def process_response_data(
                 postprocessor.strategy,  # type: ignore
             )
             try:
-                processed_data = strategy.process(processed_data, identity_data)
+                processed_data = strategy.process(
+                    processed_data, identity_data, param_values
+                )
```
galvana, line 495: Let's pass in the `privacy_request` in here instead. Privacy request has a `get_raw_access_results()` function that we can use to get a map of all the results we have up until now.

```suggestion
                processed_data = strategy.process(
                    processed_data, identity_data, privacy_request
                )
```

src/fides/api/service/connectors/saas_connector.py
```diff
@@ -456,16 +459,27 @@ def process_response_data(
...
+        if (
+            self.current_saas_request.param_values is not None
+            and input_data is not None
+        ):
+            for param_value in self.current_saas_request.param_values or []:
+                if param_value.references:
+                    input_list = input_data.get(param_value.name)
+                    if input_list:
+                        param_values[param_value.name] = input_list[0]
+
```
galvana, line 482: We shouldn't need to do anything with `param_values`. The `configuration` for the postprocessor will have the `dataset_reference` value we care about.
```
postprocessors:
            - strategy: filter
              configuration:
                field: customerId
                value:
                  dataset_reference: <instance_fides_key>.customer.id
```

Then somewhere in the postprocessor filter, we'd do something like this
```
access_data.get(self.value.dataset_reference)
```
See the next comment for where `access_data` should come from

galvana (changes requested):
@Linker44, I left some suggestions, some easier than others. Try to get as much as you can and we can chat tomorrow on the refactor for the filter postprocessor

data/saas/config/shipstation_config.yml
```diff
@@ -0,0 +1,75 @@
...
+    - name: api_secret
+
+  external_references:
+    - name: customer_id
+      label: customerId Field
+      description: The dataset reference to the Shipstation customerId
+
+  client_config:
+    protocol: https
+    host: ssapi.shipstation.com
```
galvana, line 19:
```suggestion
    host: <domain>
```

src/fides/api/service/connectors/saas_connector.py
Linker44: for future reference. access_data.get(self.value.dataset_reference) did not work per se since the access_data has a special format, that being "<datatset:collection>.field" so i had to accomodate to that.

data/saas/dataset/shipstation_dataset.yml
```diff
@@ -0,0 +1,359 @@
...
+          - name: createDate
+            data_categories: [system.operations]
+            fidesops_meta:
+              data_type: string
+          - name: modifyDate
+            data_categories: [system.operations]
+            fidesops_meta:
+              data_type: string
+          - name: name
+            data_categories: [system.operations]
```
galvana, line 21: You'll need to manually categorize these fields, here's an example for this `name` field
```suggestion
            data_categories: [user.name]
```
Check out this page for the possible data categories https://ethyca.github.io/fideslang/taxonomy/data_categories/
Also take a look at existing SaaS datasets if you want more examples

src/fides/api/service/processors/post_processor_strategy/post_processor_strategy_filter.py
```diff
@@ -52,10 +53,12 @@ def process(
         self,
         data: Union[List[Dict[str, Any]], Dict[str, Any]],
         identity_data: Optional[Dict[str, Any]] = None,
+        privacy_request: Optional[PrivacyRequest] = None,
     ) -> Union[List[Dict[str, Any]], Dict[str, Any]]:
         """
         - data: A list or a dict
         - identity_data: A dict of cached identity data
+        - param_values: A dict of cached param values
```
galvana, line 61: Update this to match `privacy_request`

src/fides/api/service/processors/post_processor_strategy/post_processor_strategy_filter.py
```diff
@@ -75,6 +78,24 @@ def process(
...
+                    self.value.dataset_reference,
+                    self.name,
+                )
+                return []
+            access_data = privacy_request.get_raw_access_results()
+            dataset_reference = self.value.dataset_reference.split(".")
+            dataset, collection, field = dataset_reference
+            filter_value = self._get_nested_values(
+                access_data, f"{dataset}:{collection}.{field}"
+            )
```
galvana, line 97: Very nice 👍

tests/ops/integration_tests/saas/test_shipstation_task.py
```diff
@@ -0,0 +1,21 @@
...
+    async def test_access_request(
+        self,
+        shipstation_runner: ConnectorRunner,
+        policy: Policy,
+        shipstation_identity_email: str,
+    ):
+        await shipstation_runner.access_request(
+            access_policy=policy,
+            identities={"email": shipstation_identity_email},
+        )
```
galvana, line 21: Let's add an assertion here to make sure we got the expected data. Check out the other tests for an example
```suggestion
        access_results = await shipstation_runner.access_request(
            access_policy=policy,
            identities={"email": shipstation_identity_email},
        )
        assert access_results["shipstation_instance:customer"]...
        assert access_results["shipstation_instance:orders"]...
```

tests/ops/service/processors/post_processor_strategy/test_post_processor_strategy_filter.py
```diff
@@ -366,3 +368,534 @@ def test_nested_array_path():
...
         }
     ]
+
+
+@mock.patch("fides.api.models.privacy_request.PrivacyRequest.get_raw_access_results")
+def test_filter_by_invalid_dataset_reference(mock_method):
+
+    config = FilterPostProcessorConfiguration(
+        field="customerId",
+        value={"dataset_reference": "<instance_fides_key>.customer"},
```
galvana, line 378: Although this string works for our tests, we should replace `<instance_fides_key>` with an actual value. I don't want to give the impression that the placeholder will be replaced inside the `process` function. The `<instance_fides_key>` placeholder will be replaced by an actual value when the connector is instantiated.
```suggestion
        value={"dataset_reference": "postgres.customer"},
```

galvana (commented):
Solid implementation! I just left some comments. Make sure your run `nox -s static_checks` locally to run the auto-formatter and to detect any mypy or pylint issues.

src/fides/api/service/processors/post_processor_strategy/post_processor_strategy_filter.py
Linker44: thanks!

tests/ops/integration_tests/saas/test_shipstation_task.py
```diff
@@ -15,7 +16,87 @@ async def test_access_request(
...
+                "advancedOptions",
+                "tagIds",
+                "userId",
+                "externallyFulfilled",
+                "externallyFulfilledBy",
+                "externallyFulfilledById",
+                "externallyFulfilledByName",
+                "labelMessages",
+            ],
+        )
```
galvana: These checks are done automatically as part of the `ConnectorRunner`'s `access_request` so we can remove them

data/saas/dataset/shipstation_dataset.yml
galvana, line 1: Good data categorization 💯 !

galvana (approved):
Congrats on unblocking yourself 😆 You identified functionality that was missing for your integration and you knocked it out! Good job 🙌

PR #5301, opened 2024-09-19
Fix broken fuzzy search tests

eastandwestwind:
Closes Unticketed

### Description Of Changes

Address pre-existing test failures with fuzzy search

### Code Changes

* [ ] Fix tests

### Steps to Confirm

* [ ] Run `nox -s dev -- shell` then `pytest tests/ops/api/v1/endpoints/test_privacy_request_endpoints.py::TestGetPrivacyRequests`
* [ ] Confirm tests pass

### Pre-Merge Checklist

* [ ] All CI Pipelines Succeeded
* Documentation:
  * [ ] documentation complete, [PR opened in fidesdocs](https://github.com/ethyca/fidesdocs/pulls)
  * [ ] documentation [issue created in fidesdocs](https://github.com/ethyca/fidesdocs/issues/new/choose)
  * [ ] if there are any new client scopes created as part of the pull request, remember to update public-facing documentation that references our scope registry
* [ ] Issue Requirements are Met
* [ ] Relevant Follow-Up Issues Created
* [ ] Update `CHANGELOG.md`
* [ ] For API changes, the [Postman collection](https://github.com/ethyca/fides/blob/main/docs/fides/docs/development/postman/Fides.postman_collection.json) has been updated
* If there are any database migrations:
  * [ ] Ensure that your downrev is up to date with the latest revision on `main`
  * [ ] Ensure that your `downgrade()` migration is correct and works
    * [ ] If a downgrade migration is not possible for this change, please call this out in the PR description!

PR #5321, opened 2024-09-23
PROD-2700 Implement Soft Delete for PrivacyRequests

erosselli:
Closes [PROD-2700](https://ethyca.atlassian.net/browse/PROD-2700)

### Description Of Changes

Adds support for soft deleting `PrivacyRequests`. The deletion is defined by the new column `deleted_at`, and we also stored who deleted it in the `deleted_by` column. A new `POST /privacy-request/id/soft-delete` endpoint is added in order to allow soft deleting requests, as well as an endpoint to bulk delete multiple requests at the same time `POST /privacy-request/bulk/soft-delete`. 

### Code Changes

* Added columns `deleted_at` and `deleted_by` to `PrivacyRequest` , both nullable 
* Added a new endpoint to allow soft deleting privacy requests, `POST /privacy-request/id/soft-delete`  and for bulk deleting them `POST /privacy-request/bulk/soft-delete`
* Added a new optional param to the existing search / list privacy requests endpoints, `include_deleted_requests`, that is `False` by default
* Added validation to the existing privacy request endpoints that modify privacy request data, to throw a 422 error when attempting to update any data related to a deleted privacy request

### Steps to Confirm

* Run the admin UI and the privacy center (or alternatively, just run the entire test project) 
* Create four different privacy requests, each with a different email, by clicking the "Access my data" blob of the privacy center
* If you go to the Admin UI, you should see all privacy requests 
* Open SwaggerUI (or your Postman collection if you prefer) and call the soft delete endpoint with the id of one of the privacy requests 
* Next, use the bulk soft delete endpoint to delete two other requests
* Go back to Admin UI -- you should now see only one privacy request, the one that hasn't been deleted 

#### Error cases

Trying to do anything that modifies the deleted privacy request should cause a 422 error. Some examples: 
- approving the request
- deleting the request 
- restarting the request

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

[PROD-2700]: https://ethyca.atlassian.net/browse/PROD-2700?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/fides/api/alembic/migrations/versions/75bb9ee843f5_add_soft_delete_for_privacy_request.py
```diff
@@ -0,0 +1,38 @@
...
+
+"""
+
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = "75bb9ee843f5"
+down_revision = "9de4bb76307a"
```
erosselli, line 15: note to self: remember to update before merging

src/fides/api/api/v1/endpoints/privacy_request_endpoints.py
```diff
@@ -188,7 +191,7 @@
 
 
 def get_privacy_request_or_error(
-    db: Session, privacy_request_id: str
+    db: Session, privacy_request_id: str, error_if_deleted: Optional[bool] = True
```
erosselli, line 194: If the privacy request is deleted, we error by default, unless the `error_if_deleted` argument is set to False

src/fides/api/api/v1/endpoints/privacy_request_endpoints.py
```diff
@@ -434,6 +443,7 @@ def _filter_privacy_request_queryset(
     external_id: Optional[str] = None,
     action_type: Optional[ActionType] = None,
     include_consent_webhook_requests: Optional[bool] = False,
+    include_deleted_requests: Optional[bool] = False,
```
erosselli, line 446: same here, by default we will filter out deleted requests, unless explicitly included

src/fides/api/api/v1/endpoints/privacy_request_endpoints.py
```diff
@@ -882,7 +902,7 @@ def get_request_status_logs(
 ) -> AbstractPage[ExecutionLog]:
     """Returns all the execution logs associated with a given privacy request ordered by updated asc."""
 
-    get_privacy_request_or_error(db, privacy_request_id)
+    get_privacy_request_or_error(db, privacy_request_id, error_if_deleted=False)
```
erosselli, line 905: I allow `get` operations on the deleted privacy request , but not update ones. TBH there weren't any specific requirements regarding this on the ticket, so I used my best judgement. Any feedback / opinions are welcome

src/fides/api/models/privacy_request.py
```diff
@@ -451,6 +460,15 @@ def delete(self, db: Session) -> None:
...
 
+    def soft_delete(self, db: Session, user_id: str) -> None:
+        """
+        Soft delete the privacy request, marking it as deleted and setting the user who deleted it.
+        """
+        self.deleted_at = datetime.utcnow()
+        self.deleted_by = user_id
+        self.save(db)
+        # TODO: do we need to remove cache keys?
+
```
erosselli, line 471: the question is because the "real" `delete` method (defined right above) does the following: 
```py
        cache: FidesopsRedis = get_cache()
        all_keys = get_all_cache_keys_for_privacy_request(privacy_request_id=self.id)
        for key in all_keys:
            cache.delete(key)
        for provided_identity in self.provided_identities:  # type: ignore[attr-defined]
            provided_identity.delete(db=db)
```

src/fides/api/service/privacy_request/email_batch_service.py
```diff
@@ -41,6 +41,7 @@ def send_email_batch(self: DatabaseTask) -> EmailExitState:
         privacy_requests: Query = (
             session.query(PrivacyRequest)
             .filter(PrivacyRequest.status == PrivacyRequestStatus.awaiting_email_send)
+            .filter(PrivacyRequest.deleted_at.is_(None))
             .order_by(PrivacyRequest.created_at.asc())  # oldest first
```
erosselli, line 45: again, there were no specfic requirements here, but I think it makes sense not to send emails for deleted requests

src/fides/api/service/privacy_request/request_runner_service.py
```diff
@@ -324,6 +324,13 @@ def run_privacy_request(
                 "Terminating privacy request {}: request canceled.", privacy_request.id
             )
             return
+
+        if privacy_request.deleted_at is not None:
+            logger.info(
+                "Terminating privacy request {}: request deleted.", privacy_request.id
+            )
+            return
```
erosselli, line 332: same here, I didn't think it made sense to run an access request for a deleted privacy reqyest

tests/ops/api/v1/endpoints/test_privacy_request_endpoints.py
```diff
@@ -1237,6 +1312,8 @@ def test_filter_privacy_requests_by_identity_no_request_id(
         assert len(resp["items"]) == 1
         assert resp["items"][0]["id"] == privacy_request.id
 
+    # FIXME: don't skip this
+    @pytest.mark.skip("temporary cause it hangs?")
```
erosselli, line 1316: Un related to this PR, see https://github.com/ethyca/fides/pull/5301 . I've skipped it just so that this file would run on the CI. I can remove the skip before merging

erosselli (commented):
Just wanted to call out that I tried to filter out the deleted requests from the places where it made sense, but also I don't know if I found all the places where we get a list of privacy requests. Wanted to point it out to get some extra eyes around that if possible

pattisdr:
starting review -

src/fides/api/api/v1/endpoints/privacy_request_endpoints.py
pattisdr: This is a great thing to run by Michael early on!   There's a case to be made either way.

Maybe I would strive for consistency here:`get_request_status_logs`, `view_uploaded_manual_webhook_data`, `view_uploaded_erasure_manual_webhook_data`, and `get_individual_privacy_request_tasks` return very specific details about potentially deleted privacy requests by default. 

But the search endpoint which is the place to get primary privacy request details does _not_ return deleted privacy requests by default.  This latter detail tips me towards wanting to hide everything by default or add query params to the GET ones so the default is "don't show" instead of "show"


Screenshot 2024-09-24 at 3 05 00 PM

src/fides/api/service/privacy_request/email_batch_service.py
pattisdr: smart

pattisdr (commented):
This looks pretty good but let's step back for a second and clarify with product about expected behavior around "soft deletes".  What should be able to be retrieved about a privacy request once it has been soft deleted?

src/fides/api/alembic/migrations/versions/75bb9ee843f5_add_soft_delete_for_privacy_request.py
```diff
@@ -0,0 +1,44 @@
...
+    )
+    op.add_column("privacyrequest", sa.Column("deleted_by", sa.String(), nullable=True))
+    op.create_foreign_key(
+        "privacyrequest_deleted_at_key",
+        "privacyrequest",
+        "fidesuser",
+        ["deleted_by"],
+        ["id"],
+        ondelete="SET NULL",
+    )
```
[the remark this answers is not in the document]
erosselli: ah that's a good idea I think, it may be overkill to have the FK constraint anyways -- I was following what was done for the approved_by column (I think), but it could be argued that this is a different use case anyways

src/fides/api/api/v1/endpoints/privacy_request_endpoints.py
erosselli: Confirmed the requirements here https://ethyca.atlassian.net/browse/PROD-2700?focusedCommentId=44123

pattisdr:
reviewing!

pattisdr (approved):
Changes look good to me but I didn't do an additional round of testing - leaving that to you @erosselli !

src/fides/api/api/v1/endpoints/privacy_request_endpoints.py
```diff
@@ -2441,3 +2488,86 @@ def request_task_async_callback(
...
+    """
+    Endpoint for soft deleting a privacy request. The request's deleted_at field will be populated with the current datetime
+    and its deleted_by field will be populated with the user_id of the user who initiated the deletion.
+    """
+    privacy_request: PrivacyRequest = get_privacy_request_or_error(
+        db, privacy_request_id
+    )
+    user_id = client.user_id
+    if client.id == CONFIG.security.oauth_root_client_id:
+        user_id = "root"
```
pattisdr, line 2571: I just saw there is a `root_username` field as well if that is useful

erosselli: hmm yeah I wonder which one is more clear: to store user_id for all users but username for the root user, or just this sort of "fake root user id"

pattisdr: I think "root" is clear too, up to you

PR #5325, opened 2024-09-24
PROD-2734 - initial partitioning support (user-defined windows)

adamsachs:
Closes PROD-2734

### Description Of Changes

- Depends on https://github.com/ethyca/fideslang/pull/21
- Accepts array of strings in a `fides_meta.partitioning.where_clauses` (maybe there's a better name?) field on the `DatasetCollection` config, which are (for BQ DSR execution) injected into access and erasure queries
- if configured on a BQ dataset collection, DSR execution now executes one query per specified `where_clauses` string per node per DSR request, rather that just one query per node per DSR request.
- validation of `where_clauses` input to prevent SQL injection

Configuration is within `fides_meta` of a `DatasetCollection`, e.g.: 

```yaml
dataset:
  - fides_key: bigquery_example_test_dataset_with_partitioning
    name: BigQuery Example Test Dataset Demonstrating Partitioning
    description: Example of a BigQuery dataset containing a variety of related tables like customers, products, addresses, etc.
    collections:
      - name: customer_partitioned
        fides_meta:
          partitioning:
            where_clauses: [
                "`created` > DATETIME_SUB(CURRENT_DATETIME(), INTERVAL 1000 DAY) AND `created` <= CURRENT_DATETIME()",
                "`created` > TIMESTAMP_SUB(CURRENT_DATETIME(), INTERVAL 2000 DAY) AND `created` <= TIMESTAMP_SUB(CURRENT_DATETIME(), INTERVAL 1000 DAY)"
            ]
        fields:
       ...
```

- note that `where_clauses` values _must_ strictly match an expected format/structure:
```
"`column_name` >(=) [value expression] (AND `column_1` <(=) [value expression])"
```
or
```
"`column_name` <(=) [value expression] (AND `column_1` >(=) [value expression])"
```

### Code Changes

- [x] update dataset <--> request graph (de)serialization to correctly extract `partitioning` metadata from dataset config
- [x] perform rigid validation of `where_clauses` values in the `partitioning` metadata to prevent malicious string injection
- [x] ensure partitioning can only be used with BigQuery DSR execution for now, as this is still an experimental feature
- [x] provide a hook in generic `retrieve_data` to call a `partitioned_retrieval` method if the collection has `partitioning` defined (and the connector supports partitioning, i.e. only BigQuery for now)
- [x] implement the `partitioned_retrieval` hook in the `BigQueryConnector` to support executing multiple partitioned queries for DSR access requests against BQ
- [x] also update existing BQ-specific methods to support executing multiple partitioned queries for DSR erasure (update/delete) requests against BQ

### Steps to Confirm

* [ ] testing steps here: https://ethyca.atlassian.net/wiki/spaces/EN/pages/3316875290/Testing+BQ+DSR+Partitioning

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`


data/dataset/postgres_example_test_dataset.yml
```diff
@@ -127,6 +127,19 @@ dataset:
             data_categories: [system.operations]
 
       - name: payment_card
+        # a sample partitioning for functional testing - not realistic!
+        fides_meta:
+          partitioning:
+            field: billing_address_id
+            windows:
+              - start: "0"
```
adamsachs, line 135: might be nice to allow proper ints here, but we do also need to support timestamps, and easiest for now is just to accept generic strings...

requirements.txt
```diff
@@ -16,7 +16,7 @@ types-defusedxml==0.7.0.20240218
 expandvars==0.9.0
 fastapi[all]==0.111.0
 fastapi-pagination[sqlalchemy]==0.12.25
-fideslang==3.0.4
+fideslang[all] @ git+https://github.com/ethyca/fideslang@64c15da6e742e5c7cecd3b2484d2b66dbb94fce8
```
adamsachs, line 19: TODO: update once https://github.com/ethyca/fideslang/pull/21 is finalized and merged

src/fides/api/service/connectors/query_config.py
```diff
@@ -458,7 +466,7 @@ def get_update_stmt(
         pk_clauses: List[str],
     ) -> str:
         """Returns a SQL UPDATE statement to fit SQL syntax."""
-        return f"UPDATE {self.node.address.collection} SET {', '.join(update_clauses)} WHERE {' AND '.join(pk_clauses)}"
+        return f"UPDATE {self.node.address.collection} SET {', '.join(update_clauses)} WHERE {' AND '.join(pk_clauses + ([PARTITION_CLAUSE_TEMPLATE] if self.partitioning else []))}"
```
adamsachs, line 469: TODO: could be cleaner

src/fides/api/service/connectors/query_config.py
```diff
@@ -476,7 +484,7 @@ def format_key_map_for_update_stmt(self, fields: List[str]) -> List[str]:
 
     def generate_update_stmt(
         self, row: Row, policy: Policy, request: PrivacyRequest
-    ) -> Optional[T]:
+    ) -> Optional[T] | List[T]:
```
adamsachs, line 487: TODO: probably could be cleaner

src/fides/api/service/connectors/query_config.py
```diff
@@ -509,8 +523,62 @@ def generate_update_stmt(
...
+                )
+                partitioned_queries.append(
+                    self.format_query_stmt(query_str, partitioned_update_value_map)
+                )
+            return partitioned_queries
+        else:
+            logger.info(
+                "query = {}, params = {}", Pii(query_str), Pii(update_value_map)
+            )
+            return self.format_query_stmt(query_str, update_value_map)
```
adamsachs, line 549: same as before, just moved down here into the `else`.

src/fides/api/service/connectors/query_config.py
```diff
@@ -509,8 +523,62 @@ def generate_update_stmt(
...
+                partitioned_update_value_map.update(partition_var_set)
+                logger.info(
+                    "query = {}, params = {}",
+                    Pii(query_str),
+                    Pii(partitioned_update_value_map),
+                )
+                partitioned_queries.append(
+                    self.format_query_stmt(query_str, partitioned_update_value_map)
+                )
+            return partitioned_queries
```
adamsachs, line 544: injection of the partitioning vars, a new piece

src/fides/api/service/connectors/sql_connector.py
```diff
@@ -178,14 +180,30 @@ def retrieve_data(
...
+                )
+                for partition_var_set in partition_var_sets:
+                    logger.debug(
+                        f"Executing partition query with start '{partition_var_set['partition_start']}' and end ''{partition_var_set['partition_end']}'"
+                    )
+                    results = connection.execute(stmt, partition_var_set)
+                    rows.extend(self.cursor_result_to_rows(results))
+            else:
+                results = connection.execute(stmt)
+                rows = self.cursor_result_to_rows(results)
```
adamsachs, line 205: same as before

src/fides/api/service/connectors/sql_connector.py
```diff
@@ -178,14 +180,30 @@ def retrieve_data(
...
+            if partition_var_sets:
+                logger.info(
+                    f"Executing {len(partition_var_sets)} partition queries for node '{node.address}' in DSR execution"
+                )
+                for partition_var_set in partition_var_sets:
+                    logger.debug(
+                        f"Executing partition query with start '{partition_var_set['partition_start']}' and end ''{partition_var_set['partition_end']}'"
+                    )
+                    results = connection.execute(stmt, partition_var_set)
+                    rows.extend(self.cursor_result_to_rows(results))
```
adamsachs, line 202: injection of partitioning vars, the new piece

src/fides/api/service/connectors/sql_connector.py
```diff
@@ -507,23 +525,21 @@ def build_uri(self) -> str:
...
 
-        return create_engine(
-            uri,
-            credentials_info=credentials_info,
-            hide_parameters=self.hide_parameters,
-            echo=not self.hide_parameters,
+        return BigQueryClient(
+            credentials=service_account.Credentials.from_service_account_info(
+                credentials_info  # pylint: disable=no-member
+            )
```
adamsachs, line 542: whoops, need to clean up. mistake.

adamsachs (commented):
some self review -

src/fides/api/service/connectors/query_config.py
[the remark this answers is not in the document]
adamsachs: ah ok, well i've done this on the abstract class, but looks like i missed updating the subclasses that override this method - most importantly, bigquery.

thanks for calling that out!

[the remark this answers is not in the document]
adamsachs: oh i see, i think this also changed on the delete override PR 👍 

probably was a bit more overlap here than anticipated...

src/fides/api/service/connectors/query_config.py
```diff
@@ -940,7 +1000,22 @@ def generate_delete(self, row: Row, client: Engine) -> Optional[Delete]:
...
+            )
+
+            for partition_clause in partition_clauses:
+                partitioned_queries.append(
+                    table.delete().where(*(pk_clauses + [text(partition_clause)]))
+                )
+
+            return partitioned_queries
+
+        return [table.delete().where(*pk_clauses)]
```
adamsachs, line 1018: fall back to existing behavior if no partitioning

src/fides/api/service/connectors/query_config.py
```diff
@@ -913,13 +953,33 @@ def generate_update(
...
+            for partition_clause in partition_clauses:
+                partitioned_queries.append(
+                    table.update()
+                    .where(*(pk_clauses + [text(partition_clause)]))
+                    .values(**update_value_map)
+                )
+
+            return partitioned_queries
+
+        return [table.update().where(*pk_clauses).values(**update_value_map)]
```
adamsachs, line 972: fall back to existing behavior if no partitioning

src/fides/api/models/datasetconfig.py
```diff
@@ -339,6 +343,7 @@ def convert_dataset_to_graph(
             data_categories=(
                 set(collection.data_categories) if collection.data_categories else set()
             ),
+            partitioning=collection_partitioning,
```
pattisdr, line 346: This reminds me, if we end up _removing_ support for the where clauses since it's experimental, we'll have to do it in a way that still allows DSR's in process to run:

tDSR 3.0 (which are customers should all be running by now) stores a representation of this collection on the `requesttask` backing the node in the database, so we don't have to rebuild the graph.  If details change about the partitioning, there could still be `requesttasks` that haven't run in the database with the old partitioning configuration which would fail when they are queued by the worker and the worker goes to rehydrate the collection off of the request task.

src/fides/api/service/connectors/query_config.py
```diff
@@ -51,6 +51,16 @@ class QueryConfig(Generic[T], ABC):
...
         self.node = node
 
+    @property
+    def partitioning(self) -> Optional[Dict]:  # pylint: disable=R1711
+        # decided to de-scope partitioning support to only bigquery as this grew more complex,
+        # but keeping more generic support stubbed out feels like a reasonable step.
+        if self.node.collection.partitioning:
+            logger.warning(
+                "Partitioning is only supported on BigQuery connectors at this time!"
+            )
```
pattisdr, line 61: this seems useful, I could have added that for bigquery row-level deletes as well

src/fides/api/service/connectors/query_config.py
```diff
@@ -537,6 +547,33 @@ def generate_update_stmt(
         logger.info("query = {}, params = {}", Pii(query_str), Pii(update_value_map))
         return self.format_query_stmt(query_str, update_value_map)
 
+    def get_partition_clauses(
+        self,
+    ) -> List[str]:
+        """
```
pattisdr, line 553: Since we for sure only support bigquery at the moment, it felt too early to move this to the generic location, but that's just my opinion

src/fides/api/service/connectors/query_config.py
```diff
@@ -934,13 +998,28 @@ def generate_delete(self, row: Row, client: Engine) -> Optional[Delete]:
...
+            logger.info(
+                f"Generating {len(partition_clauses)} partition queries for node '{self.node.address}' in DSR execution"
+            )
+
+            for partition_clause in partition_clauses:
+                partitioned_queries.append(
+                    table.delete().where(*(pk_clauses + [text(partition_clause)]))
+                )
+
+            return partitioned_queries
```
pattisdr, line 1020: might be able to dry some of this up with the update/deletes, just a thought

src/fides/api/service/connectors/sql_connector.py
```diff
@@ -199,11 +199,30 @@ def retrieve_data(
...
+                for partition_clause in partition_clauses:
+                    logger.debug(
+                        f"Executing partition query with partition clause '{partition_clause}'"
+                    )
+                    existing_bind_params = stmt.compile().params
+                    partitioned_stmt = text(
+                        f"{stmt} AND ({text(partition_clause)})"
+                    ).params(existing_bind_params)
+                    results = connection.execute(partitioned_stmt)
+                    rows.extend(self.cursor_result_to_rows(results))
```
pattisdr, line 221: This feels like it shouldn't be here yet while it's only supported for bigquery -

pattisdr (commented):
generally this makes sense - we're trying to support querying on a partitioned database with a very narrow scope, bigquery only to start, and the where clause is defined with bigquery syntax in the yaml that is added directly -

Given all that, I'd lean to keeping everything defined on the bigquery connector/bigquery queryconfig for now, as this is a super-specific implementation

src/fides/api/service/connectors/sql_connector.py
adamsachs: that's fair, but i'd generally prefer not to create parallel implementations for `retrieve_data` (for maintenance purposes) and i'm not seeing a simple way to refactor this otherwise - but maybe i'm not being creative enough!

src/fides/api/models/datasetconfig.py
adamsachs: thanks for calling that out, great edge case to consider. i don't think there's a whole lot to do right now to anticipate that future state, beyond some code comments to warn the future dev that they'll need to handle that backward compatible case, right? if you have other ideas, would love to hear them to make this as seamless as possible a transition in the (hopefully near) future!

pattisdr: just a callout would be fine!

src/fides/api/graph/config.py
```diff
@@ -647,6 +653,31 @@ def serialize_erase_after(self, erase_after: Set[CollectionAddress]) -> Set[str]
...
+        if not partitioning:
+            return partitioning
+        if where_clauses := partitioning.get("where_clauses"):
+            if not isinstance(where_clauses, List) or not all(
+                isinstance(where_clause, str) for where_clause in where_clauses
+            ):
+                raise ValueError("`where_clauses` must be a list of strings!")
+            for partition_clause in where_clauses:
+                if matching := match(PARTITION_CLAUSE_PATTERN, partition_clause):
+                    if matching["field_1"] != matching["field_2"]:
```
pattisdr, line 670: what is field_1 and field_2? can you add an example?

pattisdr: what is there is no match then, something didn't fit the pattern and the match was None?

pattisdr: are all partitioning clauses expected to be on the same field as well, or do we just care about the pairs?

src/fides/api/graph/config.py
```diff
@@ -443,6 +444,9 @@ class MaskingOverride:
     length: Optional[int]
 
 
+PARTITION_CLAUSE_PATTERN = r"`(?P<field_1>[a-zA-Z0-9_]*)` ([<|>][=]?) ([a-zA-Z0-9_\s(),.]*) AND `(?P<field_2>[a-zA-Z0-9_]*)` ([<|>][=]?) ([a-zA-Z0-9_\s(),.]*)$"
```
pattisdr, line 447: is any of this bigquery specific? if so we might note that

src/fides/api/graph/config.py
```diff
@@ -647,6 +653,31 @@ def serialize_erase_after(self, erase_after: Set[CollectionAddress]) -> Set[str]
         },
     )
 
+    @field_validator("partitioning")
+    @classmethod
+    def validate_partitioning(
```
pattisdr, line 658: I think this location makes sense given how we're leaving it flexible and open-ended right now -

src/fides/api/graph/config.py
adamsachs: > ~what is field_1 and field_2? can you add an example?~
> 
> EDIT okay got it, parsing your PARTITION_CLAUSE_PATTERN now

let me still clarify things with some comments so this is easier to understand from a quick glance 👍 


> are all partitioning clauses expected to be on the same field as well, or do we just care about the pairs?

yeah, i do think we'd expect that, but i'm not sure that checking for it really helps in preventing malicious input. granted, i don't have a super concrete example of the type of malicious input that the current check prevents, but it does at least put some more bounds on each of the clause strings. i wasn't think that checking across strings provides a whole lot of value - but i guess it could help prevent/clarify user error?

just typing out my thoughts...

src/fides/api/graph/config.py
adamsachs: yeah good call, i think the back ticks may be BQ-specific, and we're already scoped down to BQ in other respects so let me just make that more explicit 👍

src/fides/api/service/connectors/query_config.py
adamsachs: i hear you and agree, but the codepaths were already a bit duplicated to begin with, and it'll take a bit more effort to pull apart at the moment. i'll mark as a TODO for us to come back to generally DRYing up the two methods in a future iteration 👍

requirements.txt
```diff
@@ -16,7 +16,7 @@ types-defusedxml==0.7.0.20240218
 expandvars==0.9.0
 fastapi[all]==0.111.0
 fastapi-pagination[sqlalchemy]==0.12.25
-fideslang==3.0.6
+fideslang @ git+https://github.com/ethyca/fideslang@3d919f613072947f6fd6a20fcfe055ac16bd48fb
```
adamsachs, line 19: TODO: update to release

adamsachs: updated

pattisdr:
Reviewing/testing -

src/fides/api/graph/config.py
```diff
@@ -443,6 +444,22 @@ class MaskingOverride:
...
+    "UNION",
+    "INSERT",
+    "UPDATE",
+    "CREATE",
+    "DROP",
+    "SELECT",
+    "CHAR",
+    "HAVING",
+    "EXEC",
+]
```
pattisdr, line 460: great this helps add confidence

src/fides/api/graph/config.py
```diff
@@ -647,6 +666,74 @@ def serialize_erase_after(self, erase_after: Set[CollectionAddress]) -> Set[str]
...
+
+        To be clear, some notable constraints on the input:
+        - the clause string must begin by referencing a column name wrapped by backticks (`)
+        - the clause string must compare that first column with a `<>(=)` operator, and may
+        include at most one other conditional with a `<>(=)` operator that's joined to the first
+        conditional via an AND operator
+        - if the clause string contains a second conditional, it must reference the same column name
+        as the first conditional, also wrapped by backticks
+        - column names (wrapped by backticks) must always be on the _left_ side of the `<>(=)`operator
+        in its conditional
```
pattisdr, line 697: V. helpful

src/fides/api/graph/config.py
```diff
@@ -647,6 +666,74 @@ def serialize_erase_after(self, erase_after: Set[CollectionAddress]) -> Set[str]
...
+                        raise ValueError(
+                            f"Partition clause must have matching fields. Identified non-matching field references '{matching['field_1']}' and '{matching['field_2']}"
+                        )
+
+                    for prohibited_keyword in PROHIBITED_KEYWORDS:
+                        search_str = prohibited_keyword.lower() + r"\s"
+                        if search(search_str, partition_clause.lower()):
+                            raise ValueError(
+                                "Prohibited keyword referenced in partition clause"
+                            )
```
pattisdr, line 729: I think you've done what you can given that we're allowing a where clause to be entered directly here! Lots of good checks.

src/fides/api/service/connectors/sql_connector.py
```diff
@@ -199,9 +199,15 @@ def retrieve_data(
...
             return []
+
         logger.info("Starting data retrieval for {}", node.address)
         with client.connect() as connection:
             self.set_schema(connection)
+            if (
+                query_config.partitioning
+            ):  # only BigQuery supports partitioning, for now
+                return self.partitioned_retrieval(query_config, connection, stmt)
+
```
pattisdr, line 210: I do like this, thanks

pattisdr (approved):
`test_bigquery_example_data` looks like a new test failure? that should be quick to resolve, I think it's just looking at table names

adamsachs:
merging as all CI failures already exist in `main` 👍

PR #5333, opened 2024-09-25
Include overrides when loading full experience

gilluminate:
Closes [PROD-2812](https://ethyca.atlassian.net/browse/PROD-2812)

### Description Of Changes

Custom overrides were being applied to the minimal TCF experience, but once the full experience was loaded custom overrides were being ignored. This fix passes the overrides along to i18n as part of the full experience's re-initialization, similar to what happens during first `init()` phase.

### Code Changes

* Pass along the `translationOverrides` into the TCFOverlay to be used while loading the messages from experience.

### Steps to Confirm

* edit `fides-js-demo.html`:

Replace
```
// window.fides_overrides = {
//   fides_embed: true,
// };
```
With
```
window.fides_overrides = {
  fides_title: "This is my new title",
  fides_description: "This is my new description",
  fides_privacy_policy_url: "https://example.com/privacy-policy",
  fides_override_language: "en",
};
```
* load the demo with a TCF enabled experience (eg. http://localhost:3001/fides-js-demo.html?geolocation=eea)
* The banner's title, description, and privacy policy link should all be updated with the overrides even after the full experience has loaded (once the `/privacy-experience` endpoint has resolved)
* Now change languages (eg. http://localhost:3001/fides-js-demo.html?geolocation=eea&fides_locale=fr) and note that the overrides only apply to the language specified in `fides_override_language`

### Pre-Merge Checklist

* [ ] Update `CHANGELOG.md`

[PROD-2812]: https://ethyca.atlassian.net/browse/PROD-2812?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/clients/fides-js/src/components/tcf/TcfOverlay.tsx
+++ b/clients/fides-js/src/components/tcf/TcfOverlay.tsx
@@ -56,6 +56,7 @@ export const TcfOverlay = ({
   cookie,
   savedConsent,
   propertyId,
+  translationOverrides,
 }: TcfOverlayProps) => {
   const {
     i18n,
@@ -144,7 +145,7 @@ export const TcfOverlay = ({
         const fullExperience: PrivacyExperience = { ...result, ...userPrefs };
 
         setExperience(fullExperience);
-        loadMessagesFromExperience(i18n, fullExperience);
+        loadMessagesFromExperience(i18n, fullExperience, translationOverrides);
         if (!userlocale || bestLocale === defaultLocale) {
           // English (default) GVL translations are part of the full experience, so we load them here.
           loadGVLMessagesFromExperience(i18n, fullExperience);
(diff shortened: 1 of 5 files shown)


clients/fides-js/src/components/tcf/TcfOverlay.tsx
```diff
@@ -144,7 +145,7 @@ export const TcfOverlay = ({
         const fullExperience: PrivacyExperience = { ...result, ...userPrefs };
 
         setExperience(fullExperience);
-        loadMessagesFromExperience(i18n, fullExperience);
+        loadMessagesFromExperience(i18n, fullExperience, translationOverrides);
```
eastandwestwind, line 148: Do we need an equivalent for non-TCF overlay too?

gilluminate: no, the minimal -> full experience is TCF only.

gilluminate: Or, a better answer would be, the non-TCF experience already does this on initialization (see https://github.com/ethyca/fides/blob/main/clients/fides-js/src/lib/i18n/i18n-utils.ts#L496)

eastandwestwind: ah yes, thank you!

eastandwestwind (approved):

PR #5336, opened 2024-09-26
Updating hash migration status check

galvana:
Closes [PROD-2828](https://ethyca.atlassian.net/browse/PROD-2828)

### Description Of Changes

Updates the hash migration status check query to use the `is_hash_migrated IS FALSE` clause instead of `is_hash_migrated = FALSE`. The indexes for this query are created with `IS FALSE` so the status check queries were running an inefficient sequential scan on the entire table instead of an optimized index scan.

### Steps to Confirm

* [ ] This is tricky to test on a dev environment but as long as the server starts up it means the query still ran and the intended behavior is still preserved.

### Pre-Merge Checklist

* [ ] Update `CHANGELOG.md`

[PROD-2828]: https://ethyca.atlassian.net/browse/PROD-2828?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,9 @@ The types of changes are:
 ### Developer Experience
 - Migrate toggle switches from Chakra to Ant Design [#5323](https://github.com/ethyca/fides/pull/5323)
 
+### Fixed
+- Updating the hash migration status check query to use the available indexes [#5336](https://github.com/ethyca/fides/pull/5336)
+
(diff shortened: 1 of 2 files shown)


NevilleS (approved):
👍

PR #5338, opened 2024-09-27
Fix cookies not deleting on opt-out

jpople:
Closes PROD-2822

Associated fidesplus PR- https://github.com/ethyca/fidesplus/pull/1653

### Description Of Changes

Fixes a bug where cookies were not being deleted when a user opted out due to js-cookie attempting to exactly match a provided domain even when it was `null`.

Adds a new field on privacyexperienceconfig table, `auto_subdomain_cookie_deletion`; when `true`, opting out will cause cookies from subdomains (prefixed with `.`s) to be deleted as well as cookies from the top-level domain.

Note that I've manually tested the migration using the following steps:
1. Added some privacy experience configs to mock pre-existing customers' configs
2. Ran migration
3. Confirmed the `auto_subdomain_cookie_deletion` was `false`
4. Added net new privacy experience config
5. Confirmed the auto_subdomain_cookie_deletion` was `true` ✅ 

### Code Changes

* [ ] Add toggle to experience config in admin-ui that controls whether we automatically delete subdomain cookies
* [ ] Updates appropriate models across the fides ecosystem to account for this new field in the BE
* [ ] Adds migration to write the `auto_subdomain_cookie_deletion` to "false" for existing privacy experience configs

### Steps to Confirm

Test using local fides-js against a [staging Ethyca site](https://ethyca-site-2ccue3cye-ethyca.vercel.app/) which GETs localhost:3001 for fides-js.

Run the [associated Fidesplus branch](https://github.com/ethyca/fidesplus/pull/1653) (this is pointed to the `aa28ea20aeee81d4f73db25538af5413718b9ef1` fides commit since there are BE changes here)

Toggle the `auto_subdomain_cookie_deletion` field on your privacy experience to `false`:
* [ ] Opt in to a privacy notice
* [ ] Should save cookies
* [ ] Opt out
* [ ] Should delete only top-level cookies with domain of `ethyca.com`, _not_ subdomain cookies with domain of `.ethyca.com`

Toggle `auto_subdomain_cookie_deletion` on your privacy experience to `true` and re-run, then:
* [ ] Opt in, should save as before
* [ ] On opt out, should delete cookies both with top-level and subdomains

### Pre-Merge Checklist

* [ ] Update `CHANGELOG.md`


tvandort (commented):
Neat! Thanks! 

I'm curious, was the `js-cookie` library not working for removing these if we passed the same domains?

clients/fides-js/src/lib/cookie.ts
```diff
@@ -331,13 +331,18 @@ export const makeConsentDefaultsLegacy = (
...
+  cookiesToRemove: CookiesType[],
+  removeSubdomainCookies: boolean = false,
+) => {
   cookiesToRemove.forEach((cookie) => {
-    cookies.remove(cookie.name, {
-      path: cookie.path ?? "/",
-      domain: cookie.domain,
-    });
+    const { hostname } = window.location;
+    cookies.remove(cookie.name);
```
eastandwestwind, line 342: I'm thinking for this line we could keep the previous logic of: 
```
cookies.remove(cookie.name, {
      path: cookie.path ?? "/",
      domain: cookie.domain,
    });
```

This way, if we ever were somehow deleting compass cookies before (unlikely but still), we won't change functionality unintentionally. what do you think? Then when we follow-up with handling compass cookies, too, we can update this bit.

clients/privacy-center/app/server-utils/loadEnvironmentVariables.ts
```diff
@@ -81,6 +81,11 @@ const loadEnvironmentVariables = () => {
     FIDES_CLEAR_COOKIE: process.env.FIDES_PRIVACY_CENTER__FIDES_CLEAR_COOKIE
       ? process.env.FIDES_PRIVACY_CENTER__FIDES_CLEAR_COOKIE === "true"
       : false,
+    AUTOMATIC_SUBDOMAIN_COOKIE_DELETION: process.env
+      .FIDES_PRIVACY_CENTER__AUTOMATIC_SUBDOMAIN_COOKIE_DELETION
+      ? process.env
+          .FIDES_PRIVACY_CENTER__AUTOMATIC_SUBDOMAIN_COOKIE_DELETION === "true"
+      : false,
```
eastandwestwind, line 88: 👍

eastandwestwind (changes requested):
Looking good @jpople ! In addition to my comment below:

Could we add a test that basically asserts we do not support cookie deletion for compass cookies yet, and link to https://ethyca.atlassian.net/browse/PROD-2830 as the follow-up? E.g. when cookie domain is `["google.com", "googleadservices.com"]'` cookie removal won't work. 

If that's gonna take too much time, I'd say just leave a link to the PROD ticket in the cookie removal logic 👍

jpople:
@eastandwestwind I added a test case to cover a cookie with a passed-in domain, but the tests on `removeCookieFromBrowser` are just detecting whether `cookies.remove()` is called with the right arguments as they were before.  I'm not sure if it's possible to test whether that actually results in a cookie being deleted or not.

clients/admin-ui/src/features/privacy-experience/PrivacyExperienceForm.tsx
```diff
@@ -315,6 +315,12 @@ export const PrivacyExperienceForm = ({
           Edit experience text
         </Button>
       )}
+      <CustomSwitch
```
jpople, line 318: Can we add a tooltip that explains this in a bit more depth?

clients/fides-js/src/lib/cookie.ts
```diff
@@ -331,13 +331,21 @@ export const makeConsentDefaultsLegacy = (
 
 /**
  * Given a list of cookies, deletes them from the browser
+ * Optionally removes subdomain cookies as well
  */
-export const removeCookiesFromBrowser = (cookiesToRemove: CookiesType[]) => {
+export const removeCookiesFromBrowser = (
+  cookiesToRemove: CookiesType[],
+  removeSubdomainCookies: boolean = true,
```
jpople, line 338: Do we want this to default to `true` here?

eastandwestwind: yes, this is stored in the DB-level and is unlikely to ever be undefined here. When a user configs their experience in Admin-UI, then that removeSubdomainCookies toggle will be either true or false depending on their choice. 

Additionally, our current customers will get the `false` val written to the DB, so we're safe.

eastandwestwind (approved):

PR #5344, opened 2024-10-01
Custom report models

galvana:
Closes [PROD-2673](https://ethyca.atlassian.net/browse/PROD-2673)

### Description Of Changes

Introduces the `CustomReport` model with the following properties:

- id
- name
- type
- created_by
- config

A custom report, for now, is defined as an object containing the table state (to be determined by the front-end) and a mapping of column keys to custom labels.

### Code Changes

* [ ] New `CustomReport` model

### Steps to Confirm

* [ ] This will be used by the front-end but steps to verify back-end only functionality can be found at https://github.com/ethyca/fidesplus/pull/1641

### Pre-Merge Checklist

* [ ] Update `CHANGELOG.md`

[PROD-2673]: https://ethyca.atlassian.net/browse/PROD-2673?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/ops/models/test_custom_report.py
```diff
@@ -0,0 +1,99 @@
...
+                "created_by": user.id,
+                "config": CustomReportConfig(
+                    column_map={"system": "Vendor"}
+                ).model_dump(mode="json"),
+            },
+        )
+        yield custom_report
+        custom_report.delete(db)
+
+    def test_create_custom_report(self, db: Session, user: FidesUser):
```
erosselli, line 30: this test doesn't have any assertions -- is that expected?

src/fides/api/alembic/migrations/versions/c9abc9bbaa49_add_custom_report_table.py
```diff
@@ -0,0 +1,51 @@
...
+        sa.Column(
+            "updated_at",
+            sa.DateTime(timezone=True),
+            server_default=sa.text("now()"),
+            nullable=True,
+        ),
+        sa.Column("name", sa.String(), nullable=True),
+        sa.Column("type", sa.String(), nullable=False),
+        sa.Column("created_by", sa.String(), nullable=True),
+        sa.Column("config", postgresql.JSONB(astext_type=sa.Text()), nullable=True),
```
erosselli, line 39: Are we storing the `table_state` and `column_map` within the `config` (as opposed to using two separate columns) because we expect that we'll need more things in the config if/when we implement custom reports in other parts of the UI?

galvana: That's right, especially for `table_state`, Jason is still trying to figure out the schema for the front-end. It's tricky because the metadata to capture a Tanstack table is different from an ANT table so I just provided a flexible solution for now.

tests/ops/models/test_custom_report.py
galvana: I was just asserting that we could create it with no errors but I added field assertions now

erosselli (approved):

PR #5352, opened 2024-10-04
Introduce custom reports to Data map

gilluminate:
Closes PROD-2680

### Description Of Changes

Allow customers to save custom reporting templates. Saved reports are accessible in reporting views with specific filter/grouping/column setting applied. Reports are accessible to everyone in an org/workspace.

The Data map report can be broken down to these various states:

* Group By
* Filters
* Column Ordering
* Column Visibility
* Column Sizing
* Column Expand/Collapse (view more applied)
* Column Wrap Results

All 7 should persist across sessions (local storage) for when the user reloads their page or returns to the app later.

The saved report should **exclude** _Column Sizing_, _Column Expand/Collapse_, and _Column Wrap Results_.

### Code Changes

* Refactor `DatamapReportTable` component, breaking it up into multiple files for better maintainability
* New `CustomReportsTemplate` custom popover component with ability to create, load, apply, and delete custom reports
* New datamap react context/provider to more easily maintain and access data stored in local storage
* Pass `report_id` to endpoint for downloading reports for backend support
* Cypress tests

### Steps to Confirm

* Log in to Admin UI > Data map report (eg. http://localhost:3000/reporting/datamap)
* Adjust some of the options:
  * Group by Data Use
  * Reorder columns
  * Hide a column or two
  * Apply a filter
* Reload the page (all settings should remain applied)
* Open the Reports popover
* Create a new report by giving it a name
* Log in to Admin UI > Data map report in an Incognito window where none of the customization has been applied.
* Open the Reports popover
* Select the newly created report and click Apply
* All of the customization should now be applied to the table
* The popover trigger button should be renamed to the selected report (with ellipsis if it's long)
* Verify that each selection is reflected in the various menus
* Make a different change (eg. hide another column)
* The applied report is no longer active, the popover trigger button should be named Reports again, when opened there should be no filter applied
* Select the report again and click reset (selection should be cleared)
* Select the report again and click outside of the popover
* Report should be applied even without clicking the Apply button
* Open the popover again and attempt to create a new report
* Give the report the "exact" same name as before
* You should see an error disallowing duplicate names
* Change the name and save a new report
* Now click the trash icon next to one of the saved reports
* You should see a dialog confirming
* Canceling the dialog should close the modal without deleting
* Continue button should close the modal and delete the report from the list
* Apply a report again
* Resizing columns, expanding/collapsing columns, and wrapping columns should **not** reset the applied report and should not have any bearing on the report itself (doesn't get saved with the report) but reloading the page should retain these settings for the current user only.

### Pre-Merge Checklist

* [x] Update `CHANGELOG.md`


clients/admin-ui/src/features/common/custom-reports/CustomReportCreationModal.tsx
```diff
@@ -0,0 +1,142 @@
+import {
+  Button,
```
jpople, line 2: Can we use AntButton here?

clients/admin-ui/src/features/common/custom-reports/CustomReportTemplates.tsx
```diff
@@ -0,0 +1,426 @@
+import {
+  Button,
```
jpople, line 2: Also update to AntButton here.

jpople (approved):
:100:

PR #5383, opened 2024-10-15
FIX: separate urn with non-nested-pattern for now

gilluminate:
Closes HJ-36

### Description Of Changes

This is a bit of a bandaid fix to handle nested fields on the frontend. We are specifying something other than `.` used in nested field names for the URN so that we can distinguish those. More robust fixes are happening on both the BE and FE in separate tickets.

### Code Changes

* Specify a non-period separator for FE logic
* Apply some best practices to the Edit drawer and Breadcrumbs to avoid race conditions during E2E testing and for optimization in general.

### Steps to Confirm

* In Fides Plus Admin UI, visit the Manage datasets page (eg. http://localhost:3000/dataset)
* **If the dataset `example_dataset_issue_hj36` exists**
1. Click it
2. Click `example_table`
3. Ensure that you can still add/remove Categories from the "Can save data category changes" fields.
4. Click through to the nested field `example_nested_field`
5. The field `example_failure_nested_field.1` with description "Fail to save data category changes on nested field with '.' in name" should now be able to add/remove categories.
6. Click "edit" on the same `example_failure_nested_field.1` row and make sure you can still "Save"
* **If the dataset `example_dataset_issue_hj36` does not exist**
1. Click "+ Add Dataset" button
2. Click "Upload a Dataset YML"
3. Paste the following into the editor:
```
- fides_key: example_dataset_issue_hj36
  organization_fides_key: default_organization
  tags: null
  name: example_dataset_issue_hj36
  description: Minimal dataset reproducing issue (HJ-36)
  meta: null
  data_categories: null
  fides_meta: null
  collections:
    - name: example_table
      description: null
      data_categories: null
      fields:
        - name: example_nested_field
          description: null
          data_categories: null
          fides_meta: null
          fields:
            - name: example_success_nested_field
              description: Can save data category changes on nested field
              data_categories:
                - user.device
              fides_meta: null
              fields: []
            - name: example_failure_nested_field.1
              description: >-
                Fail to save data category changes on nested field with '.' in
                name
              data_categories:
                - user.location
                - system
              fides_meta: null
              fields: []
        - name: example_success_field
          description: Can save data category changes
          data_categories:
            - user.device
          fides_meta: null
          fields: null
        - name: example_success_field.1
          description: Can save data category changes
          data_categories:
            - user.device
          fides_meta: null
          fields: null
      fides_meta: null
```
4. Click "Create dataset" button
5. Proceed to follow steps 1-6 above for the `example_dataset_issue_hj36` dataset

### Pre-Merge Checklist

* [x] Update `CHANGELOG.md`

### DIFF
--- a/clients/admin-ui/src/features/dataset/EditCollectionDrawer.tsx
+++ b/clients/admin-ui/src/features/dataset/EditCollectionDrawer.tsx
@@ -1,5 +1,6 @@
 import { errorToastParams, successToastParams } from "common/toast";
 import { ConfirmationModal, Text, useDisclosure, useToast } from "fidesui";
+import { useMemo } from "react";
 
 import EditDrawer, {
   EditDrawerFooter,
@@ -19,8 +20,8 @@ import {
 const DESCRIPTION =
   "Collections are an array of objects that describe the Dataset's collections. Provide additional context to this collection by filling out the fields below.";
 interface Props {
-  dataset?: Dataset;
-  collection?: DatasetCollection;
+  dataset: Dataset;
+  collection: DatasetCollection;
   isOpen: boolean;
   onClose: () => void;
 }
@@ -30,7 +31,11 @@ const EditCollectionDrawer = ({
   isOpen,
   onClose,
 }: Props) => {
-  const collectionIndex = dataset?.collections.indexOf(collection!);
+  const collectionIndex = useMemo(
+    () => dataset?.collections.indexOf(collection),
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [],
+  );
   const [updateDataset] = useUpdateDatasetMutation();
   const toast = useToast();
   const {
@@ -42,7 +47,7 @@ const EditCollectionDrawer = ({
   const handleSubmit = async (
     values: Pick<DatasetCollection, "description" | "data_categories">,
   ) => {
-    const updatedCollection = { ...collection!, ...values };
+    const updatedCollection = { ...collection, ...values };
     const updatedDataset = getUpdatedDatasetFromCollection(
       dataset!,
       updatedCollection,
@@ -92,7 +97,7 @@ const EditCollectionDrawer = ({
         }
       >
         <EditCollectionOrFieldForm
-          values={collection!}
+          values={collection}
           onSubmit={handleSubmit}
           dataType="collection"
           showDataCategories={false}
--- a/clients/admin-ui/src/features/dataset/constants.ts
+++ b/clients/admin-ui/src/features/dataset/constants.ts
@@ -29,3 +29,5 @@ export const FIELD = {
       "Arrays of Data Category resources, identified by fides_key, that apply to this field.",
   },
 };
+
+export const URN_SEPARATOR = "/";
--- a/clients/admin-ui/src/pages/dataset/[datasetId]/[collectionName]/[subfieldUrn].tsx
+++ b/clients/admin-ui/src/pages/dataset/[datasetId]/[collectionName]/[subfieldUrn].tsx
@@ -46,6 +46,7 @@ import {
   useGetDatasetByKeyQuery,
   useUpdateDatasetMutation,
 } from "~/features/dataset";
+import { URN_SEPARATOR } from "~/features/dataset/constants";
 import DatasetBreadcrumbs from "~/features/dataset/DatasetBreadcrumbs";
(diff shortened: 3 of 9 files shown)


clients/admin-ui/src/features/dataset/EditCollectionDrawer.tsx
```diff
@@ -30,7 +31,11 @@ const EditCollectionDrawer = ({
   isOpen,
   onClose,
 }: Props) => {
-  const collectionIndex = dataset?.collections.indexOf(collection!);
+  const collectionIndex = useMemo(
+    () => dataset?.collections.indexOf(collection),
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
lucanovera, line 36: i'm curious, did it cause an issue trying to add the deps?

clients/admin-ui/src/features/dataset/constants.ts
```diff
@@ -29,3 +29,5 @@ export const FIELD = {
       "Arrays of Data Category resources, identified by fides_key, that apply to this field.",
   },
 };
+
+export const URN_SEPARATOR = "/";
```
lucanovera, line 33: Nice quick solution changing the separator to something different. 
Although this makes me wonder, could we expect to have "/" as a part of the name too? 😂 

Anyways, we shouldn't have accepted dots as part of the name either, and kirk is working on a fix for that.

clients/admin-ui/src/pages/dataset/[datasetId]/[collectionName]/[subfieldUrn].tsx
```diff
@@ -260,46 +261,48 @@ const FieldsDetailPage: NextPage = () => {
     DatasetField | undefined
   >();
 
+  const breadcrumbs = useMemo(() => {
```
Comment (lucanovera) [inline clients/admin-ui/src/pages/dataset/[datasetId]/[collectionName]/[subfieldUrn].tsx:264]: Good optimization!

clients/admin-ui/src/pages/dataset/[datasetId]/index.tsx
```diff
@@ -174,12 +176,14 @@ const DatasetDetailPage: NextPage = () => {
         </Box>
       )}
 
-      <EditCollectionDrawer
-        dataset={dataset!}
-        collection={selectedCollectionForEditing}
-        isOpen={isEditingCollection}
-        onClose={() => setIsEditingCollection(false)}
-      />
+      {dataset && selectedCollectionForEditing && (
```
Comment (lucanovera) [inline clients/admin-ui/src/pages/dataset/[datasetId]/index.tsx:179]: Nice, avoiding needing to use the ! or handling undefined values further down.

lucanovera (approved):
I reproduced the issue and confirmed that it's fix in this branch. The code looks good. left a minor question below. approved!

lucanovera:
@gilluminate While testing I found another bug that would be very easy to fix, can we include it here?

The bug is that in the table view if you try to add a category by adding the plus icon, it will navigate down to the nested field level.

In src/pages/dataset/[datasetId]/[collectionName]/index.tsx the data categories column is missing this option:
`meta: {
          disableRowClick: true,
        },`
        
It won't cause an issue here src/pages/dataset/[datasetId]/[collectionName]/[subfieldUrn].tsx but we could add it for consistency too.

clients/admin-ui/src/features/dataset/EditCollectionDrawer.tsx
gilluminate: This was causing a test to fail and I believe it's because the supplied collection was no longer matching anything in the dataset. Since we're now requiring both dataset and collection to be available from the start, we want to snag this index and then stop worrying about it once either gets updated. Including the dependencies recreates the problem.

clients/admin-ui/src/features/dataset/constants.ts
gilluminate: Yes, it's party why I made this a constant we can update. We can come back and tweak if it's still a problem for other data.

clients/admin-ui/src/pages/dataset/[datasetId]/index.tsx
Comment (gilluminate) [reply clients/admin-ui/src/pages/dataset/[datasetId]/index.tsx:179]: IMO we should avoid `!` always. It's usually an indication for a potential bug, as was the case here.

clients/admin-ui/src/features/dataset/EditCollectionDrawer.tsx
gilluminate: I think if I were to rewrite this, I'd pass in the `index` rather than the `collection` but this seemed like the less risky fix.

PR #5387, opened 2024-10-16
HJ-23 Add connection_type to NamespaceMeta

erosselli:
Part of [HJ-23](https://ethyca.atlassian.net/browse/HJ-23)

### Description Of Changes

Migration to add the `connection_type` field to all existing `namespace` dicts in the `fides_meta` field of the `Dataset` model.

### Code Changes

* Migration to populate `connection_type` for existing datasets with a namespace 

### Steps to Confirm

* Make sure you have at least 4 datasets created (but the more the merrier!): 
    *  A: One without a `namespace` key in its `fides_meta` field 
    *  B: One with a `namespace` key and value `{ "project_id": "some-id" }` (simulates a BigQuery dataset created through D&D)
    *  C: One  with a `namespace` key and value `{ "database_instance_id": "some-id" }` (simulates an RDS MySQL dataset created through D&D)
    * D: One with a `namespace` key and value `{ "other": "test" }` 
* Run the migration (if it was already run when you started the server, you can just downgrade and then upgrade again using the alembic CLI)
* Check that the dataset in created in (B) now has a `connection_type` field with value `bigquery` as part of its `namespace`
* Check that the dataset in created in (C) now has a `connection_type` field with value `rds_mysql` as part of its `namespace`
* The other two datasets should remain unchanged 

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

[HJ-23]: https://ethyca.atlassian.net/browse/HJ-23?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/fides/api/alembic/migrations/versions/4ebe0766021b_add_connection_type_to_dataset_namespace.py
+++ b/src/fides/api/alembic/migrations/versions/4ebe0766021b_add_connection_type_to_dataset_namespace.py
@@ -1,7 +1,7 @@
 """Add connection_type to dataset namespace meta
 
 Revision ID: 4ebe0766021b
-Revises: 5dafbc1818ae
+Revises: 36ad82edb38e
 Create Date: 2024-10-16 15:25:08.544946
 
 """
@@ -11,7 +11,7 @@ from alembic import op
 
 # revision identifiers, used by Alembic.
 revision = "4ebe0766021b"
-down_revision = "5dafbc1818ae"
+down_revision = "36ad82edb38e"
 branch_labels = None
 depends_on = None
 
--- a/src/fides/api/alembic/migrations/versions/4ebe0766021b_add_connection_type_to_dataset_namespace.py
+++ b/src/fides/api/alembic/migrations/versions/4ebe0766021b_add_connection_type_to_dataset_namespace.py
@@ -19,13 +19,22 @@ depends_on = None
 def upgrade():
     # Populate the connection_type field in the namespace meta for all datasets
     # that have a namespace key in their fides_meta json field.
-    # We use bigquery as the initial value since only datasets generated from
-    # BigQuery monitors will have a namespace key in their fides_meta json field.
+
+    # If the namespace dict has a project_id key, we know it's a BigQuery dataset.
     op.execute(
         """
         update ctl_datasets
         set fides_meta = jsonb_set(fides_meta::jsonb, '{namespace, connection_type}', '"bigquery"', true)
-        where fides_meta::jsonb ? 'namespace';
+        where (fides_meta::jsonb ? 'namespace') and ((fides_meta::jsonb ->> 'namespace')::jsonb ? 'project_id');
+        """
+    )
+
+    # If the namespace dict has a database_instance_id, we know it's an RDS MySQL dataset.
+    op.execute(
+        """
+        update ctl_datasets
+        set fides_meta = jsonb_set(fides_meta::jsonb, '{namespace, connection_type}', '"rds_mysql"', true)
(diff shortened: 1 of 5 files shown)


src/fides/api/alembic/migrations/versions/4ebe0766021b_add_connection_type_to_dataset_namespace.py
```diff
@@ -0,0 +1,42 @@
...
+def upgrade():
+    # Populate the connection_type field in the namespace meta for all datasets
+    # that have a namespace key in their fides_meta json field.
+    # We use bigquery as the initial value since only datasets generated from
+    # BigQuery monitors will have a namespace key in their fides_meta json field.
+    op.execute(
+        """
+        update ctl_datasets
+        set fides_meta = jsonb_set(fides_meta::jsonb, '{namespace, connection_type}', '"bigquery"', true)
+        where fides_meta::jsonb ? 'namespace';
```
erosselli, line 28: I just want to call out that I am a bit concerned about the performance of this migration -- opening the json to check whether it has a `namespace` key seems like it might not scale well.

erosselli: okay tested on 500,000 rows and it took like 7s and on 1,120,000 rows it took 20s. we should be okay?

thingscouldbeworse: Thanks for doing that, makes for much higher confidence

src/fides/api/alembic/migrations/versions/4ebe0766021b_add_connection_type_to_dataset_namespace.py
```diff
@@ -0,0 +1,51 @@
...
+    op.execute(
+        """
+        update ctl_datasets
+        set fides_meta = jsonb_set(fides_meta::jsonb, '{namespace, connection_type}', '"bigquery"', true)
+        where (fides_meta::jsonb ? 'namespace') and ((fides_meta::jsonb ->> 'namespace')::jsonb ? 'project_id');
+        """
+    )
+
+    # If the namespace dict has a database_instance_id, we know it's an RDS MySQL dataset.
+    op.execute(
```
thingscouldbeworse, line 33: I'm missing context from the ticket, sorry, but do we care about any other dialects? We'll fill those in later?

thingscouldbeworse (commented):
How do we get these back out of the database? Does the app just parse everything in `fides_meta` already?

src/fides/api/alembic/migrations/versions/4ebe0766021b_add_connection_type_to_dataset_namespace.py
erosselli: My fidesplus PR https://github.com/ethyca/fidesplus/pull/1673 (not ready for review yet) makes the necessary changes to start storing it for all datasets created through D&D . This migration just takes advantage of the fact that BQ and RDS MySQL were both already using the `namespace` key so we have a reliable way to identify existing datasets from these integrations. for other integrations we'll start storing the connection_type from now on but we don't have a way to populate their values for existing datasets

erosselli:
> How do we get these back out of the database? Does the app just parse everything in `fides_meta` already?

@thingscouldbeworse yes, fides_meta is a JSON column so it gets parsed as a dictionary.

thingscouldbeworse (approved):
nice 👍

PR #5388, opened 2024-10-16
Allow Responsys and Firebase connectors to ignore extra identities

eastandwestwind:
Closes https://ethyca.atlassian.net/browse/LA-34

### Description Of Changes

Allow Responsys and Firebase connectors to ignore extra identities

### Code Changes

* [ ] Remove code that throws err if extra identities are found
* [ ] Update tests accordingly

### Steps to Confirm

We ran the following steps manually with @RobertKeyser and confirmed:

* [ ] Remove `@pytest.mark.skip(reason="Enterprise account only")` from oracle_responsys_fixtures.py
* [ ] Run `nox -s dev -- shell`
* [ ] Run `pytest -k TestOracleResponsysConnector -s`
* [ ] Confirm the `test_access_request_by_email_and_extra_identities` test no longer has the invalid identity err, and instead gets as far as the other tests
* [ ] Separately, run `pytest tests/ops/integration_tests/saas/request_override/test_firebase_auth_task.py` and see if we have failing tests

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`


src/fides/api/service/saas_request/override_implementations/oracle_responsys_request_overrides.py
```diff
@@ -147,9 +147,7 @@ def oracle_responsys_profile_list_recipients_read(
                 query_ids[idx] = query_id[1:] if query_id.startswith("+") else query_id
             query_attribute = "m"
         else:
-            raise FidesopsException(
-                "Unsupported identity type for Oracle Responsys connector. Currently only `email` and `phone_number` are supported"
-            )
+            continue
```
eastandwestwind, line 150: skip this for loop entirely if identity is not "email" or "phone_number"

Vagoasdf (approved):
Looks good! 👍🏽 

just out a few comments out of curiosity, like why are we replacing the value `extra_identities`  for `ga_client_id`, or that key usage

tests/ops/integration_tests/saas/test_oracle_responsys_task.py
```diff
@@ -59,6 +59,34 @@ async def test_access_request_by_phone_number(
...
+        request.getfixturevalue(dsr_version)  # REQUIRED to test both DSR 3.0 and 2.0
+
+        access_results = await oracle_responsys_runner.access_request(
+            access_policy=policy,
+            identities={
+                "email": oracle_responsys_identity_email,
+                "ga_client_id": "extra_value",
+            },
+        )
+        assert (
```
Vagoasdf, line 83: Should we also assert the `ga_client_id` value?

Vagoasdf (approved):

tests/ops/integration_tests/saas/test_oracle_responsys_task.py
eastandwestwind: For now I think its ok to just do positive assertions (as opposed to which identities are _not_ returned in the access_results). The main thing we want here is to ensure we don't throw an identity err before accessing data when extra identities are provided.

eastandwestwind:
@Vagoasdf to answer your Q:

If we used the literal key `extra_identities` as an identity key, this would hit our validator as an unsupported identity type, and require additional params to be passed in like `label` in addition to `value`. For simplicity of testing, I chose to use an existing identity field, `ga_client_id`.

PR #5429, opened 2024-10-29
Zendesk request override get tasks

Vagoasdf:
Closes [LA#32](https://ethyca.atlassian.net/browse/LA-32)

Pairs With [FidesPlus#1685 ](https://github.com/ethyca/fidesplus/pull/1685)

### Description Of Changes

Creating a new Request Override that would check for the  status of the ticket on a delete request and halt it if there is an open ticket

Currently we are using a default list to determine if a ticket is open. Basically, anything that its not "closed" its open. The list is ["new", "open", "pending", "hold", "solved"]

It could be a parametrized list on the future, but the current fides iteration does not support indicating a multi-select field from the config.yml file

### Code Changes

* [ ] Creating new Request Override
* [ ] Creating tests for the new functionality
* [ ] Adding fixtures that manage the new state

### Steps to Confirm

* [ ] Create a new user with a known email on Zendesk via Postman
* [ ] Create a few new tickets for that username via Postman. 
* [ ] Set up the Zendesk Integration (This can be done copying the relevant config.yml on the paired PR)
* [ ] Try to make a Delete request through the test site with the known email
* [ ] Check the Status of the request. It should fail with the message ""User still has open tickets, halting request"
* [ ] Update the tickets  status to closed via Postman
* [ ] Try to make a Delete Request through the test site with the known email
* [ ] Check the status of the request. It should've succesed
* [ ] Check on Postman and/or Zendesk if the user still exists. It should not

### Pre-Merge Checklist

* [ ] Update `CHANGELOG.md`


Vagoasdf:
I'm not sure on which of the Two PR's the Changelog should be updated

Vagoasdf:
Is it possible that Codecov is missing the test coverage because we are missing the Zendesk_config.yml file?

src/fides/api/service/saas_request/override_implementations/zendesk_request_overrides.py
```diff
@@ -0,0 +1,76 @@
...
+    """
+    if policy.get_rules_for_action(action_type=ActionType.erasure):
+        for ticket in tickets:
+            ## TODO: Check with Ramp that the Statuses list is the expected one
+            if ticket["status"] in ["new", "open", "pending", "hold", "solved"]:
+                raise FidesopsException("User still has open tickets, halting request")
+
+
+@register("zendesk_logitech_tickets_read", [SaaSRequestType.READ])
+def zendesk_logitech_tickets_read(
```
galvana, line 37: Same here, let's just call it `zendesk_tickets_read`
```suggestion
def zendesk_tickets_read(
```

tests/fixtures/saas/zendesk_fixtures.py
```diff
@@ -40,7 +40,7 @@ def zendesk_erasure_identity_email() -> str:
 class ZendeskClient:
     def __init__(self, secrets: Dict[str, Any]):
         self.base_url = f"https://{secrets['domain']}"
-        self.auth = secrets["username"], secrets["api_key"]
+        self.auth = secrets["username"] + "/token", secrets["api_key"]
```
galvana, line 43: I commented on this somewhere else too, but if we don't have a migration in place to migrate the username, we should just leave things as they are. I agree this is cleaner, but we would break existing integrations

CHANGELOG.md
```diff
@@ -21,6 +21,9 @@ The types of changes are:
 - Added DataHub integration config [#5401](https://github.com/ethyca/fides/pull/5401)
 - Added keepalive settings to the Redshift integration [#5433](https://github.com/ethyca/fides/pull/5433)
 
+### Changed
+- Adding a Request Override on Zendesk Ticket's that checks for the tickets statuses on a Delete Request
```
galvana, line 25:
```suggestion
- Preventing erasures for the Zendesk integration if there are any open tickets [#5429](https://github.com/ethyca/fides/pull/5429)
```

src/fides/api/service/saas_request/override_implementations/zendesk_request_overrides.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+def _check_tickets(tickets: List[Row], policy: Policy) -> None:
+    """
+    Raises an exception if there are any "open" tickets to halt the privacy request.
+    Will only halt if the policy contains an erasure action.
+    """
+    if policy.get_rules_for_action(action_type=ActionType.erasure):
+        for ticket in tickets:
+            ## TODO: Check with Ramp that the Statuses list is the expected one
```
galvana, line 31: Same here, let's avoid putting customer names in the comments

galvana: Once we move to fidesplus, we're ok 😄

tests/ops/integration_tests/saas/test_zendesk_task.py
```diff
@@ -87,3 +88,26 @@ async def test_non_strict_erasure_request(
...
+        policy: Policy,
+        erasure_policy_string_rewrite: Policy,
+        zendesk_erasure_identity_email: str,
+        zendesk_erasure_data_with_open_comments,
+        zendesk_client,
+    ):
+
+        with pytest.raises(
+            FidesopsException, match="User still has open tickets, halting request"
+        ):
```
galvana, line 105: Nice assertion!

tests/ops/integration_tests/saas/test_zendesk_task.py
```diff
@@ -87,3 +88,26 @@ async def test_non_strict_erasure_request(
...
+            (
+                access_results,
+                erasure_results,
+            ) = await zendesk_runner.non_strict_erasure_request(
+                access_policy=policy,
+                erasure_policy=erasure_policy_string_rewrite,
+                identities={"email": zendesk_erasure_identity_email},
+            )
```
galvana, line 113: We can simplify this since we're not using the results
```suggestion
           await zendesk_runner.non_strict_erasure_request(
                access_policy=policy,
                erasure_policy=erasure_policy_string_rewrite,
                identities={"email": zendesk_erasure_identity_email},
            )
```

galvana (changes requested):

tests/fixtures/saas/zendesk_fixtures.py
Vagoasdf: In that case, i think we should add it to the hoverable description on the config.  Something that points to the user that they hadd to add the "/token" at the end without lookinv over the Zendesk Docs

galvana (changes requested):
Just one small change and we're good to go!

galvana (approved):

PR #5438, opened 2024-10-30
HJ-127 - Fix API router sanitizer being too aggressive with NextJS Catch-all Segments

andres-torres-marroquin:
Closes #HJ-127

### Description Of Changes

Fix API router sanitizer being too aggressive with NextJS Catch-all Segments

### Code Changes

* [ ] Updated `sanitise_url_path` on `src/fides/api/main.py` so it is more permissive with NextJS urls.

### Steps to Confirm

* [ ] Run the test suite

### Pre-Merge Checklist

* Documentation:
* [ ] Update `CHANGELOG.md`

### DIFF
--- a/src/fides/api/main.py
+++ b/src/fides/api/main.py
@@ -57,7 +57,7 @@ from fides.cli.utils import FIDES_ASCII_ART
 from fides.config import CONFIG, check_required_webserver_config_values
 
 IGNORED_AUDIT_LOG_RESOURCE_PATHS = {"/api/v1/login"}
-NEXT_JS_CATCH_ALL_SEGMENTS_RE = r"^\[\.\.\.\w+\]"  # https://nextjs.org/docs/pages/building-your-application/routing/dynamic-routes#catch-all-segments
+NEXT_JS_CATCH_ALL_SEGMENTS_RE = r"^\[{1,2}\.\.\.\w+\]{1,2}"  # https://nextjs.org/docs/pages/building-your-application/routing/dynamic-routes#catch-all-segments
 
 VERSION = fides.__version__
 
--- a/tests/ops/util/test_api_router.py
+++ b/tests/ops/util/test_api_router.py
@@ -105,6 +105,22 @@ class TestApiRouter:
                 "[datasetName]/[collectionName]/[...subFields]-js/../../../../etc/passwd",
                 True,
             ),
+            (
+                "/_next/static/chunks/pages/dataset/[datasetId]/[collectionName]/[[...subfieldNames]]-6596c3d4607847d0.js",
+                False,
+            ),
+            (
+                "/_next/static/chunks/pages/dataset/[datasetId]/[[...collectionName]]/6596c3d4607847d0.js",
+                False,
+            ),
+            (
+                "/_next/static/chunks/pages/dataset/[[...datasetId]]/[collectionName]/6596c3d4607847d0.js",
+                False,
+            ),
+            (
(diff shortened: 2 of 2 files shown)


src/fides/api/main.py
```diff
@@ -248,8 +250,9 @@ def sanitise_url_path(path: str) -> str:
     """Returns a URL path that does not contain any ../ or //"""
     path = unquote(path)
     path = os.path.normpath(path)
+
     for token in path.split("/"):
-        if ".." in token:
+        if ".." in token and not re.search(NEXT_JS_CATCH_ALL_SEGMENTS_RE, token):
```
gilluminate, line 255: won't this still allow the `example.com/[datasetName]/[collectionName]/[...subFields]-js/../../../../etc.` scenario?

tests/ops/util/test_api_router.py
```diff
@@ -83,3 +83,22 @@ def test_malicious_url(
...
+    ) -> None:
+        """
+        Assert that NextJS Catch-all Segments aren't blocked by our sanitise_url_path.
+        """
+
+        non_malicious_paths = [
+            "/_next/static/chunks/pages/dataset/[datasetId]/[collectionName]/[...subfieldNames]-6596c3d4607847d0.js",
+            "/_next/static/chunks/pages/dataset/[datasetId]/[...collectionName]/6596c3d4607847d0.js",
+            "/_next/static/chunks/pages/dataset/[...datasetId]/[collectionName]/6596c3d4607847d0.js",
+        ]
```
NevilleS, line 100: Let's add a few tests for the positive case (blocking malicious paths)

tests/ops/util/test_api_router.py
```diff
@@ -83,3 +83,22 @@ def test_malicious_url(
...
+        for path in non_malicious_paths:
+            resp = api_client.get(f"{url}/{path}")
+            assert resp.status_code == 404
+            assert resp.text == '{"detail":"Item not found"}'
```
NevilleS, line 104: Is this really a good test? Doesn't seem like this assertion is very helpful, what is it testing?

I'd recommend dropping down to the unit-level and calling `sanitise_url_path` directly instead. That way you can just craft ~10 different `path` arguments to test a bunch of different cases.

src/fides/api/main.py
NevilleS: I don't think so, because of the `.split("/")` above. So it's testing each path segment like:
* token = `example.com`
* token = `[datasetName]`
* token = `[collectionName]`
* token = `[...subFields]-js`
* token = `..`
* token = `..`
* token = `..`
* token = `..`

NevilleS (changes requested):
I think the code _might_ be alright here, but I'd like the test coverage to be better. See comments

src/fides/api/main.py
gilluminate: Yeah, makes sense! Thanks for the illustration :)

daveqnet:
If this is not helpful, tell me, but an alternative approach we could take here would be to side-step all the checking for `..` strings and fully resolve the path first using `os.path.abspath` or possibly `os.path.realpath` (instead of using `os.path.normpath` to get a relative path), and only proceed if the resolved (canonical) path matches an allow-list.

This at least applies to files that can be served, not API or UI routes, as we only need path traversal protection when serving real files.

NevilleS (approved):
Thanks. I like the tests 👍

PR #5439, opened 2024-10-31
Fix nested field name displays in D&D tables

jpople:
Closes HJ-59

### Description Of Changes

Changes D&D's default result cell to display "un-nested" subfield names with multiple parts derived from the URN instead of just the "terminal" field's name.

Before:
Screenshot 2024-10-31 at 02 37 57

After:
Screenshot 2024-10-31 at 02 38 19

Non-subfield resources' naming is unchanged.

### Code Changes

* [ ] Adds a helper method to derive a resource's name

### Steps to Confirm

* [ ] Populate D&D tables with data in nested fields
* [ ] Check cell names beneath nested fields
* [ ] Fields that are nested _more than one level deep_ should show a multiple-part name separated by a `.`

### Pre-Merge Checklist

* [ ] Update `CHANGELOG.md`


clients/admin-ui/src/features/data-discovery-and-detection/utils/getResourceName.ts
```diff
@@ -0,0 +1,19 @@
+import { DiscoveryMonitorItem } from "~/features/data-discovery-and-detection/types/DiscoveryMonitorItem";
+
+const TOP_LEVEL_FIELD_URN_PARTS = 5;
+
+const getResourceName = (resource: DiscoveryMonitorItem) => {
+  const splitUrn = resource.urn.split(".");
```
gilluminate, line 6: we had an issue we fixed in datasets where Mozilla had a "." in one of their field names which broke the URN. Is that not a concern with the D&D stuff? Like can we use a hex code that looks like a "." instead or something?

gilluminate (approved):

PR #5446, opened 2024-11-01
Added masking_strategy_override at field level

Linker44:
Closes #[LA-60](https://ethyca.atlassian.net/browse/LA-60)

### Description Of Changes

This pr adds masking_strategy_overrides at the dataset field-level.

Many masking strict erasure requests fail because some fields dont have the correct format. This is caused by applying the same masking to each field when some of them should remain with a certain format. 

ex: an endpoint will only succeed at updating an email field if it has an email format.

```
          - name: email_address
            fidesops_meta:
              data_type: string
              masking_strategy_override:
                strategy: random_string_rewrite
                configuration:
                  length: 10
                  format_preservation:
                    suffix: "@test.com"
```

### Code Changes

* [x]  Added field-level masking strategy override.
* [x] Added Testing for update_value_map method in QueryConfig.
* [x] Added Testing for dsr database erasure request.
* [x] Added Validation for masking strategies that require secrets.
* [x] Added Testing for masking strategies that require secrets validation.
* [x] Renamed class MaskingOverride to something more appropiate (MaskingTruncation).

### Steps to Confirm

Confirm masking override validation on fides admin:
* [ ] Try to create a dataset that contains a field-level masking_strategy_override that requires secrets. It should fail to create it
* [ ] Try to update a dataset so that it contains a field-level masking_strategy_override that requires secrets. It should fail to create it
* [ ] Try to create a dataset that contains a field-level masking_strategy_override that doesn’t require secrets. It should succeed 

Confirm masking override validation on startup (nox -s dev):

* [ ]  Update one of the existing datasets so that it contains a field-level masking_strategy_override that requires secrets. It should raise a validation error on the terminal and fail to create it.
* [ ] Update one of the existing datasets so that it contains a field-level masking_strategy_override that  doesn't requires secrets. It should succeed.

Confirm masking override is working:

* [ ]  Alter a dsr saas dataset in a user collection field so that it contains a field_level_masking override that doesnt require secrets -> create a system with the integration. (make sure to have the enviroment variable FIDES__DEV_MODE set to true to see the request body in the terminal) -> make an erasure request -> Confirm that the request has the correct masking on the specified field.

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

[LA-60]: https://ethyca.atlassian.net/browse/LA-60?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/fides/api/models/datasetconfig.py
```diff
@@ -403,3 +410,28 @@ def validate_dataset_reference(
         raise ValidationError(
             f"Unknown field '{dataset_reference.field}' in dataset '{dataset_config.fides_key}' referenced by external reference"
         )
+
+
+def validate_masking_strategy_override(dataset: Dataset) -> None:
```
galvana, line 415: We should add a docstring explaining the need for this validation. When we receive a privacy request, we look at the policies to determine which masking strategies we need to generate and cache secret keys. For now, we don't want to add the extra scope of also having to scan the datasets for masking overrides. You can site the `cache_data` function in https://github.com/ethyca/fides/blob/field_level_masking/src/fides/api/service/privacy_request/request_service.py#L59

tests/ops/api/v1/endpoints/test_dataset_endpoints.py
```diff
@@ -1232,6 +1233,152 @@ def test_patch_datasets_bulk_create(
...
+            == "Masking strategy 'hash' with required secrets not allowed as an override."
+        )
+        field_masking_override_config = DatasetConfig.get_by(
+            db=db,
+            field="fides_key",
+            value="postgres_example_invalid_masking_strategy_override",
+        )
+        assert field_masking_override_config is None
+
+    def test_patch_datasets_invalid_field_masking_strategy_override_update(
```
galvana, line 1318: I see tests for create and update, do we need one for upsert?

tests/ops/service/connectors/test_queryconfig.py
```diff
@@ -57,6 +60,54 @@
 privacy_request = PrivacyRequest(id="234544")
 
 
+@mock.patch.multiple(QueryConfig, __abstractmethods__=set())
```
galvana, line 63: Fancy! ✨

galvana (dismissed):
Good work finding all the spots that needed validation! I just left some minor suggestions but the core of this change is solid 👍

tests/ops/api/v1/endpoints/test_dataset_endpoints.py
Linker44: Not really. The endpoint we are testing does uses an upsert function. So by testing creation and update we are esentially testing upsert. Tho i did create two more tests thanks to this comment. One for a succeded  patch update and one for a failed put.

noxfiles/setup_tests_nox.py
```diff
@@ -79,12 +79,15 @@ def pytest_ctl(session: Session, mark: str, coverage_arg: str) -> None:
...
+            "timeout",
+            "--signal=INT",
+            "300",
             "pytest",
+            "--full-trace",
             coverage_arg,
             "-m",
             "external",
             "tests/ctl",
-            "--tb=no",
```
galvana, line 87: Can you revert these debugging changes? Then we're good to go!

PR #5447, opened 2024-11-01
HJ-143: Fix ValidationError for datasets with a connection_type

erosselli:
Closes #<issue>

### Description Of Changes

Trying to parse a correct `fides_meta.namespace` field from a BigQuery dataset (i.e with a `connection_type: 'bigquery'`) was raising a Pydantic validation error, because Pydantic was expecting an Enum instance for `connection_type` rather than a string.  We now use the string literal for the type rather than the enum value. 

I've also added a test for this case (this was an oversight on my part when I implemented the original PR) and updated two fixtures to include the `connection_type` as well.  

### Steps to Confirm

If you run the test added in `tests/ops/service/connectors/test_bigquery_queryconfig.py` without the other changes, it should fail with the validation error. Adding the changes to the `BigQueryNameSpaceMeta` should have the test passing. 

**Manual check: ** 
For easier testing with D & D features , pull down this fidesplus branch https://github.com/ethyca/fidesplus/pull/1705

1. Create a bigquery dataset through D & D. Make sure the dataset is reachable, e.g it has an email identity declared on it
2. Link the dataset to a bigquery integration. 
3. Create a privacy request
4. Request should not fail because of a validation error

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`

### DIFF
--- a/tests/fixtures/bigquery_fixtures.py
+++ b/tests/fixtures/bigquery_fixtures.py
@@ -137,6 +137,7 @@ def bigquery_example_test_dataset_config_with_namespace_meta(
         "namespace": {
             "project_id": "silken-precinct-284918",
             "dataset_id": "fidesopstest",
+            "connection_type": "bigquery",
         }
     }
     fides_key = bigquery_dataset["fides_key"]
@@ -170,6 +171,7 @@ def bigquery_example_test_dataset_config_with_namespace_and_partitioning_meta(
         "namespace": {
             "project_id": "silken-precinct-284918",
             "dataset_id": "fidesopstest",
+            "connection_type": "bigquery",
         },
     }
     # update customer collection to have a partition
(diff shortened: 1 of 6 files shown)


tests/fixtures/bigquery_fixtures.py
```diff
@@ -137,6 +137,7 @@ def bigquery_example_test_dataset_config_with_namespace_meta(
         "namespace": {
             "project_id": "silken-precinct-284918",
             "dataset_id": "fidesopstest",
+            "connection_type": "bigquery",
```
thingscouldbeworse, line 140: This looks like it works, I'm surprised that we didn't break any tests before? Or at least the `pylint` run given it was a pydantic validation error? Either way, if this test reproduced and now checks that's all good. Maybe add the `connection_type` anywhere else we don't have it already to be sure.

thingscouldbeworse (approved):

tests/fixtures/bigquery_fixtures.py
erosselli: ticketed in https://ethyca.atlassian.net/browse/HJ-144

PR #5452, opened 2024-11-04
LA 83 Fix Salesforce Erasure Data Flow

Vagoasdf:
[Closes LA#83](https://ethyca.atlassian.net/browse/LA-83)

Pairs With [fidesplus#1708](https://github.com/ethyca/fidesplus/pull/1708)

### Description Of Changes

Started as a removal of the Account endpoint since that was data from the Business, not a person. Ended up creating custom overrides requests so the validation rules for the other endpoints were passing. 

currently the Requests Overrides do two things 
* Checks if the field is an email. If it is, changes it to `masked@company.com`
* Checks if the field is over 40 characters after masking. If it is, truncating it

### Code Changes

* [ ]  Request overrides for the Salesforce endpoints created

### Steps to Confirm

* [ ] Create a new Business Account on Salesforces from the Admin Template
* [ ] Create two contacts for that Business Account
* [ ] Set up the Salesforces Integration 
* [ ] Mask one of the Contacts by Email
* [ ] The Other Contact should not be masked
* [ ] The Account should not be masked

imagen

### Pre-Merge Checklist

* [ ] Update `CHANGELOG.md`


src/fides/api/service/saas_request/override_implementations/salesforce_request_overrides.py
```diff
@@ -0,0 +1,159 @@
...
+# if so, truncate it to 40 characters
+def truncateFieldsTo40Characters(masked_object_fields: Dict) -> Dict:
+    for key in masked_object_fields:
+        logger.info(key)
+        logger.info((masked_object_fields[key]))
+        if not isinstance(masked_object_fields[key], str):
+            continue
+        if len(masked_object_fields[key]) > 40:
+            masked_object_fields[key] = masked_object_fields[key][:40]
+    return masked_object_fields
```
galvana, line 26: We can simplify this a bit and only log a message when we need to truncate. The comment you added is helpful but it would make sense to add it as a doc string. Finally, make sure to use snake case since that's the convention for Python.
```suggestion
def truncate_fields(masked_object_fields: Dict) -> Dict:
    """
    Check if the masked field is over 40 characters long, if so truncate it to 40 characters.
    """
    for key in masked_object_fields:
        value = masked_object_fields[key]
        if isinstance(value, str) and len(value) > 40:
            logger.info("Truncating {key} field to 40 characters")
            masked_object_fields[key] = value[:40]

    return masked_object_fields
```

src/fides/api/service/saas_request/override_implementations/salesforce_request_overrides.py
```diff
@@ -0,0 +1,159 @@
...
+
+
+# Masking Email properly so it does not breaks validation rules
+# using the privacy request id to have an unique id
+def maskEmail(masked_object_fields: Dict, email_field: str) -> Dict:
```
galvana, line 31: Same as above
```suggestion
def mask_email(masked_object_fields: Dict, email_field: str) -> Dict:
    """Masking email fields properly so they don't break validation rules."""
```

tests/ops/integration_tests/saas/test_salesforce_task.py
galvana, line 1: I know we have these tests flagged as `@pytest.mark.skip(reason="Currently unable to test OAuth2 connectors")` but can you try the following:
1. Run the tests by hard-coding the `access_token` in your local `saas_config.toml`?
2. Use the `erasure_policy_string_rewrite_name_and_email` policy fixture so we can see the email being masked.
3. Use the new `ConnectorRunner` pattern for the task tests
4. Verify we have multiple users in Salesforce to verify we only return results for one user
5. Verify that multiple entities can have the same email (the masked@company email we're using)

galvana (changes requested):

src/fides/api/service/saas_request/override_implementations/salesforce_request_overrides.py
```diff
@@ -0,0 +1,159 @@
...
+        masked_object_fields = truncateFieldsTo40Characters(masked_object_fields)
+
+        update_body = dumps(masked_object_fields)
+        contact_id = row_param_values["contact_id"]
+        client.send(
+            SaaSRequestParams(
+                method=HTTPMethod.PATCH,
+                headers={"Content-Type": "application/json"},
+                path=f"/services/data/v54.0/sobjects/Contact/{contact_id}",
+                body=update_body,
```
galvana, line 62: We should be able to just pass the JSON directly, same for the other request overrides
```suggestion
        contact_id = row_param_values["contact_id"]
        client.send(
            SaaSRequestParams(
                method=HTTPMethod.PATCH,
                headers={"Content-Type": "application/json"},
                path=f"/services/data/v54.0/sobjects/Contact/{contact_id}",
                json=masked_object_fields,
```

Vagoasdf: Actually we can't. If we pass the JSON value directly we get the following 400 error 
`'[{"message":"The HTTP entity body is required, but this request has no entity body.","errorCode":"JSON_PARSER_ERROR"}]'`

So we have to pass the data in the body.

Vagoasdf:
By the looks of the [Campaings Member API ](https://developer.salesforce.com/docs/atlas.en-us.object_reference.meta/object_reference/sforce_api_objects_campaignmember.htm) docs, it looks like the Campaing Members represents a relationship between Leads and Contacts, and as such, their fields should be updated on their respective endpoints. We already have some fields on the campaign_members dataset collection as read_only, and those that are not, are throwing an 400 error as we are not able to update them. 

removing the Update endpoint on campaign member. We still want to read it, even if it's a bit of duplicate data, because its relevant to see if its on a campaign, but it should not be updated at all

tests/ops/integration_tests/saas/test_salesforce_task.py
```diff
@@ -1,1142 +1,186 @@
...
 from fides.api.service.connectors import get_connector
 from fides.api.task.graph_task import get_cached_data_for_erasures
 from fides.config import CONFIG
 from tests.conftest import access_runner_tester, erasure_runner_tester
 from tests.ops.graph.graph_test_util import assert_rows_match
+from tests.ops.integration_tests.saas.connector_runner import ConnectorRunner
 
 
-@pytest.mark.skip(reason="Currently unable to test OAuth2 connectors")
+#@pytest.mark.skip(reason="Currently unable to test OAuth2 connectors")
```
galvana, line 15:
```suggestion
@pytest.mark.skip(reason="Currently unable to test OAuth2 connectors")
```

galvana (approved):

PR #5465, opened 2024-11-05
Address GitHub CI issues

gilluminate:
Cypress tests are inconsistent, randomly fail, are quite slow. To help address this, lets:

* Upgrade older versions of Cypress and Turbo
* Upgrade `cypress-io/github-action` to v6
* Update lots of Cypress stubs to avoid `ECONNREFUSED` errors with no BE running (`/api/v1/system/`, `/health`, etc)
* Fix error handling on Privacy Center to avoid crashes during fides-js tests with no BE running
* Minimize run time by avoiding redundant builds


clients/admin-ui/src/features/plus/plus.slice.ts
```diff
@@ -262,8 +262,16 @@ const plusApi = baseApi.injectEndpoints({
         url: `plus/custom-metadata/custom-field-definition/resource-type/${resource_type}`,
       }),
       providesTags: ["Custom Field Definition"],
-      transformResponse: (list: CustomFieldDefinitionWithId[]) =>
-        list.sort((a, b) => (a.name ?? "").localeCompare(b.name ?? "")),
+      transformResponse: (
+        response: CustomFieldDefinitionWithId[] | { detail: string },
+      ) => {
+        if ("detail" in response) {
```
lucanovera, line 268: I don't get the context for this change. Could you explain?

gilluminate: That's a great question that I'm trying to investigate and remember...

gilluminate: @lucanovera I updated this code with a comment to explain why. For some reason, instead of an error code or an empty list, the BE is returning { detail: "No custom metadata fields found with resource type system" } which was causing a console error in this original piece of code.

gilluminate: Good idea! Thanks for doing that.

jpople (approved):
Everything is passing for me-- great job cleaning all this up.

jpople (approved):
Looks good and passing all tests for me.

PR #5468, opened 2024-11-06
LA-41  Update POST endpoints for data categories, subjects and uses for new taxonomy functionality

eastandwestwind:
Closes https://ethyca.atlassian.net/browse/LA-41

### Description Of Changes

 Update POST endpoints for data categories, subjects and uses for new taxonomy functionality. Specifically enables us to do 2 things:
1. Create a new resource without having to specify a fides_key. We generate fides_key in the BE in this case
2. When we attempt to create a new resource that has the same name as a disabled resource, we automatically re-enable that existing resource

### Code Changes

* [ ] Adds new overrides to our generic endpoints for taxonomy
* [ ] Adds test assertions for all new use cases

### Steps to Confirm

* [ ] Confirm that all unit tests pass
* [ ] To manually test, you'll need to manually ping the API for POST `/data_use`, POST `/data_category`, and POST `/data_subejct`. Confirm that you can create a resource without having to specify a fides key

### Pre-Merge Checklist

* Documentation:
* [x] Update `CHANGELOG.md`


src/fides/api/api/v1/endpoints/generic_overrides.py
```diff
@@ -85,5 +96,92 @@ async def list_dataset_paginated(
...
+    Helper to create taxonomy resource when not given a fides_key.
+    Automatically re-enables disabled resources with the same name.
+    """
+    # If data with same name exists but is disabled, re-enable it
+    disabled_resource_with_name = db.query(model).filter(
+        model.key == data.name,  # type: ignore[union-attr]
+        model.active is False,
+    )
+    if disabled_resource_with_name:
+        return model.update(db=db, data=disabled_resource_with_name, active=True)  # type: ignore[union-attr]
```
galvana, line 114: What happens if the incoming `data` doesn't match the DB model we're reactivating? Should we pass in the incoming `data` instead of the `disabled_resource_with_name`?
```suggestion
        return model.update(db=db, data=data, active=True)  # type: ignore[union-attr]
```

eastandwestwind: yes, good catch!

galvana (approved):
Thanks for making the recommended changes, approved!

PR #5478, opened 2024-11-07
HJ-131 Delete monitors before deleting integration

erosselli:
Closes  [HJ-131](https://ethyca.atlassian.net/browse/HJ-131)

### Description Of Changes

Before, trying to delete an integration (i.e `ConnectionConfig`) that had a related monitor (i.e `MonitorConfig`) would throw a 500 error. Intended behavior is to cascade the deletion so that all monitors associated to the given `ConnectionConfig` are deleted. 

### Code Changes

- Updated `ConnectionConfig` model to explicitly define its `monitors` reference, and updated the class' `delete` method to iterate over its related `MonitorConfig`s and delete them as well
- Added a test for this case
- Created a `detection_discovery_fixtures.py` file to organize D&D fixtures in one single place

### Steps to Confirm

1. Pull down [this fidesplus branch](https://github.com/ethyca/fidesplus/pull/1717) so that you have all D&D features. 
2. Create an integration, or click on an existing one, e.g a BigQuery integration. 
3. Add a couple (i.e more than 1) monitors to the integration
4. Click `Delete integration`
5. Integration and monitors should both be deleted 

### Pre-Merge Checklist

* [ ] Issue requirements met
* [x] `CHANGELOG.md` updated
* Followup issues:
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:

[HJ-131]: https://ethyca.atlassian.net/browse/HJ-131?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/fides/api/models/connectionconfig.py
+++ b/src/fides/api/models/connectionconfig.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import enum
 from datetime import datetime
+from loguru import logger
 from typing import Any, Dict, Optional, Type
 
 from sqlalchemy import Boolean, Column, DateTime, Enum, ForeignKey, String, event
@@ -162,6 +163,12 @@ class ConnectionConfig(Base):
         cascade="all, delete",
     )
 
+    monitors = relationship(
+        "MonitorConfig",
+        back_populates="connection_config",
+        cascade="all, delete",
+    )
+
     access_manual_webhook = relationship(  # type: ignore[misc]
         "AccessManualWebhook",
         back_populates="connection_config",
@@ -276,8 +283,15 @@ class ConnectionConfig(Base):
 
     def delete(self, db: Session) -> Optional[FidesBase]:
         """Hard deletes datastores that map this ConnectionConfig."""
+        logger.info(
+            "Deleting connection config {} and its associated datasets and monitors",
+            self.key,
+        )
         for dataset in self.datasets:
             dataset.delete(db=db)
+
+        for monitor in self.monitors:
+            monitor.delete(db=db)
 
         return super().delete(db=db)
 
--- a/src/fides/api/models/connectionconfig.py
+++ b/src/fides/api/models/connectionconfig.py
@@ -2,9 +2,9 @@ from __future__ import annotations
 
 import enum
 from datetime import datetime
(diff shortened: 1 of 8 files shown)


tests/ops/models/test_detection_discovery.py
```diff
@@ -190,34 +190,14 @@ def test_staged_resource_helpers(self, db: Session, create_staged_resource):
 
 
 class TestMonitorConfigModel:
-    @pytest.fixture
-    def create_monitor_config(self, db: Session, connection_config: ConnectionConfig):
```
erosselli, line 194: moved this fixture to the new `detection_discovery_fixtures.py` file I created and just renamed it to `monitor_config`, to be more consistent with our existing fixture names

src/fides/api/models/connectionconfig.py
```diff
@@ -21,6 +22,9 @@
 from fides.api.schemas.saas.saas_config import SaaSConfig
 from fides.config import CONFIG
 
+if TYPE_CHECKING:
+    from fides.api.models.detection_discovery import MonitorConfig
+
```
erosselli, line 27: this is so mypy doesn't complain about not recognizing the `MonitorConfig` model. The `if TYPE_CHECKING` is needed to avoid a circular import

tests/fixtures/detection_discovery_fixtures.py
```diff
@@ -0,0 +1,84 @@
...
+            "databases": ["db1", "db2"],
+            "execution_frequency": None,
+            "execution_start_date": None,
+        },
+    )
+    yield mc
+
+
+@pytest.fixture(scope="function")
+def monitor_config_2_no_deletion(
```
thingscouldbeworse, line 61: Why do we need these to not delete themselves? Just to avoid an error when it tries to delete itself after the test already got it?

erosselli: yes -- if you think it's neater, I can add the delete back to the function and just catch the `ObjectDeleted` error so the fixture doesn't break

PR #5479, opened 2024-11-07
HJ-166: Support BlueConic objectives

tvandort:
Closes [HJ-166]

### Description Of Changes

Support BlueConic Objectives

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:

[HJ-166]: https://ethyca.atlassian.net/browse/HJ-166?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/clients/fides-js/src/integrations/blueconic.ts
+++ b/clients/fides-js/src/integrations/blueconic.ts
@@ -50,7 +50,7 @@ const configureObjectives = () => {
 };
 
 export const blueconic = (
-  { approach }: { approach: string } = { approach: "onetrust" },
+  { approach }: { approach: "onetrust" } = { approach: "onetrust" },
 ) => {
   if (approach !== "onetrust") {
     throw new Error("Unsupported approach");
--- a/clients/fides-js/src/integrations/blueconic.ts
+++ b/clients/fides-js/src/integrations/blueconic.ts
@@ -1,0 +1,55 @@
+declare global {
+  interface Window {
+    blueConicClient?: {
+      profile: {
+        setConsentedObjectives: (objectives: string[]) => void;
+        setRefusedObjectives: (objectives: string[]) => void;
+        updateProfile: () => void;
+      };
+      event: {
+        subscribe: any;
+      };
+    };
+  }
+}
+
+// https://support.blueconic.com/hc/en-us/articles/202605221-JavaScript-front-end-API
+const blueConicLoaded = () =>
+  typeof window.blueConicClient !== "undefined" &&
+  typeof window.blueConicClient.event !== "undefined" &&
+  typeof window.blueConicClient.event.subscribe !== "undefined";
+
+const configureObjectives = () => {
+  if (!blueConicLoaded() || !window.blueConicClient?.profile) {
+    return;
+  }
+
+  const profile = window.blueConicClient?.profile;
+  const { consent } = window.Fides;
+  if (consent.targeted_advertising_gpp_us_state) {
+    profile.setConsentedObjectives([
+      "iab_purpose_1",
+      "iab_purpose_2",
+      "iab_purpose_3",
+      "iab_purpose_4",
+    ]);
+    profile.setRefusedObjectives([]);
+  } else {
+    profile.setConsentedObjectives(["iab_purpose_1"]);
+    profile.setRefusedObjectives([
+      "iab_purpose_2",
+      "iab_purpose_3",
+      "iab_purpose_4",
+    ]);
+  }
+
+  profile.updateProfile();
+};
+
+export const blueconic = () => {
+  window.addEventListener("FidesInitialized", configureObjectives);
+  window.addEventListener("FidesUpdated", configureObjectives);
+  window.addEventListener("onBlueConicLoaded", configureObjectives);
+
+  configureObjectives();
+};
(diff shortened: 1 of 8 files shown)


clients/fides-js/src/integrations/blueconic.ts
```diff
@@ -0,0 +1,55 @@
...
+  typeof window.blueConicClient.event.subscribe !== "undefined";
+
+const configureObjectives = () => {
+  if (!blueConicLoaded() || !window.blueConicClient?.profile) {
+    return;
+  }
+
+  const profile = window.blueConicClient?.profile;
+  const { consent } = window.Fides;
+  if (consent.targeted_advertising_gpp_us_state) {
```
tvandort, line 29: I feel like I need to do something more expansive here. Should I be looking to do something like [this.](https://github.com/ethyca/fides/blob/42be590e3b64e566642d22b1c831bffa8cac4609/clients/fides-js/src/integrations/shopify.ts#L20)

tvandort:
Are there any tests I should add this to?

clients/fides-js/src/integrations/blueconic.ts
```diff
@@ -0,0 +1,55 @@
...
+  typeof window.blueConicClient.event !== "undefined" &&
+  typeof window.blueConicClient.event.subscribe !== "undefined";
+
+const configureObjectives = () => {
+  if (!blueConicLoaded() || !window.blueConicClient?.profile) {
+    return;
+  }
+
+  const profile = window.blueConicClient?.profile;
+  const { consent } = window.Fides;
```
gilluminate, line 28: This feels like a potential race condition, assuming this `window.Fides` will already be updated after the event. Feels safer to get the consent object from the event itself if possible.

```
  configureObjectives = (event?: FidesEvent) => {
```

and then

```suggestion
  const { consent } = event?.details || window.Fides;
```

tvandort: I'm down to change it that way but... our docs are _very_ explicit that this would not be a race condition:

> **FidesUpdated:** Dispatched when a user action (e.g. accepting all, saving changes, applying GPC) has finished updating the user's consent preferences. **This event is dispatched once the changes are fully saved to the Fides object, fides_consent cookie on the user's device, and the Fides API.** To receive an event that fires before these changes are saved, use the FidesUpdating event instead.

> **FidesInitialized**: Dispatched when initialization is complete and the current user's consent preferences - either previously saved or applicable defaults - **have been set on the Fides global object.**

Emphasis mine.

gilluminate: Ah perfect. Thanks for the sanity check. I can't think of any reason not to do it this way in that case. It's definitely cleaner code.

clients/fides-js/src/integrations/shopify.ts
```diff
@@ -18,14 +19,7 @@ declare global {
...
 const CONSENT_MAP = {
-  marketing: [
-    "marketing",
-    "data_sales_and_sharing",
-    "data_sales_sharing_gpp_us_state",
-    "data_sharing_gpp_us_state",
-    "data_sales_gpp_us_state",
-    "targeted_advertising_gpp_us_state",
-  ],
+  marketing: MARKETING_CONSENT_KEYS,
```
gilluminate, line 22: 👍

gilluminate (approved):
This looks great. Love the new precedent set by having an integrations unit test.

clients/fides-js/src/lib/consent-types.ts
```diff
@@ -161,6 +162,7 @@ export interface FidesGlobal extends Fides {
   options: FidesInitOptions;
   saved_consent: NoticeConsent;
   tcf_consent: TcfOtherConsent;
+  blueconic: typeof blueconic;
```
NevilleS, line 165: nit: I'd prefer you add this to the src/docs/fides.ts `Fides` interface as well, even if you mark it `@internal` there. That interface is what generates the developer docs here: https://ethyca.com/docs/dev-docs/js/reference/interfaces/Fides

This isn't a must-have here, esp. since we're POC'ing this integration first, but I'd like us to try to keep those docs updated as we build

clients/fides-js/src/lib/consent-constants.ts
```diff
@@ -125,3 +125,12 @@ export const FIDES_OVERRIDE_EXPERIENCE_LANGUAGE_VALIDATOR_MAP: {
...
 export const FIDES_OVERLAY_WRAPPER = "fides-overlay-wrapper";
 export const FIDES_I18N_ICON = "fides-i18n-icon";
+
+export const MARKETING_CONSENT_KEYS = [
+  "marketing",
+  "data_sales_and_sharing",
+  "data_sales_sharing_gpp_us_state",
+  "data_sharing_gpp_us_state",
+  "data_sales_gpp_us_state",
+  "targeted_advertising_gpp_us_state",
```
NevilleS, line 135:
```suggestion
  "targeted_advertising_gpp_us_state",
  "sales_sharing_targeted_advertising_gpp_us_national",
```

NevilleS (approved):
One extra notice to add here and a nitpick about docs, but otherwise this is solid. Nice!

clients/fides-js/src/lib/consent-types.ts
tvandort: I'll move it.

tvandort: function call is now included in the docs interface. I did mark it as `@internal` because I'm not sure what I want to write for it yet docs wise. 

Going to wait for tests to pass after the most recent change, then will merge.

PR #5483, opened 2024-11-11
Add branding link to privacy center & CMP modal

jpople:
Closes HJ-157, HJ-165

### Description Of Changes

Adds "powered by Ethyca" branding (conditional on environment variables) to the consent modal, both TCF and non-TCF:

Screenshot 2024-11-11 at 11 44 58

Screenshot 2024-11-11 at 11 45 13

And to the privacy center:

Screenshot 2024-11-11 at 11 45 35

### Code Changes

* Adds two new environment variables to the privacy center, `FIDES_PRIVACY_CENTER__SHOW_BRAND_LINK` and `FIDES_PRIVACY_CENTER__BRAND_LINK_URL`

### Steps to Confirm

1. Add the following to your privacy center `.env`: 
```
FIDES_PRIVACY_CENTER__SHOW_BRAND_LINK=false
FIDES_PRIVACY_CENTER__BRAND_LINK_URL=https://fid.es #(or a URL of your choice)
```
2. View the privacy center and a consent modal
3. Branding link should not render in either
4. In privacy center `.env`, change `FIDES_PRIVACY_CENTER__SHOW_BRAND_LINK` to `true`
5. View the privacy center
6. Branding link should render in lower right corner at all screen sizes
7. Clicking branding link should lead to configured URL
8. View a consent modal
9. Branding link should render in lower right corner of modal at all screen sizes
10. Clicking branding link should lead to configured URL

### Pre-Merge Checklist

* [x] Issue requirements met
* [x] `CHANGELOG.md` updated
* Followup issues:
  * [x] No followup issues
* Database migrations:
  * [x] No migrations
* Documentation:


clients/fides-js/src/components/ConsentButtons.tsx
```diff
@@ -56,6 +57,8 @@ export const ConsentButtons = ({
       onManagePreferencesClick();
     }
   };
+  const includeBrandLink =
+    isInModal && options.showFidesBrandLink && options.fidesBrandUrl;
```
NevilleS, line 61: nit: Can this just be a single config `fidesShowBrandingLink`, and just set the brand URL to ttps://fid.es/powered instead of making that a config option? (I just created that shortlink now)

Having two config options just makes this more complicated to document & support, and I'd rather have one clear boolean `FIDES_PRIVACY_CENTER__SHOW_BRANDING_LINK` that is an on/off switch.

clients/fides-js/src/components/EthycaLogo.tsx
```diff
@@ -0,0 +1,9 @@
+import { h } from "preact";
+
+const EthycaLogo = () => (
+  <svg xmlns="http://www.w3.org/2000/svg" width="31" height="20" fill="#2b2e35">
```
NevilleS, line 4: let's set this fill color with a CSS class instead, so you can just change a var like `--fides-overlay-brand-link-font-color` to style this

an inline style fill color like this is hard to change

clients/fides-js/src/components/fides.css
```diff
@@ -13,6 +13,7 @@
   --fides-overlay-gpc-applied-text-color: white;
   --fides-overlay-gpc-overridden-background-color: #e53e3e;
   --fides-overlay-gpc-overridden-text-color: white;
+  --fides-overlay-brand-link-font-color: #a0aec0;
```
NevilleS, line 16: Do we need the font & logo to be different colors? They are very similar here, and it'll be less error-prone to have them match.

If we want the logo color to be specified, let's have the font color reuse an existing text color. That'll make it more likely to work in existing customizations

clients/fides-js/src/components/fides.css
```diff
@@ -477,6 +480,13 @@ div#fides-consent-content .fides-modal-description {
   justify-content: center;
 }
 
+.fides-modal-secondary-actions #fides-brand-link {
+  position: absolute;
+  right: var(--fides-overlay-padding);
+  color: var(--fides-overlay-brand-link-font-color) !important;
```
NevilleS, line 486: nit: `!important` suggests you're fighting against something... what is it?

clients/fides-js/src/lib/consent-types.ts
```diff
@@ -136,6 +136,12 @@ export interface FidesInitOptions {
 
   // Shows fides.js overlay UI on load deleting the fides_consent cookie as if no preferences have been saved
   fidesClearCookie: boolean;
+
+  // Whether to render the brand link in the footer of the modal
+  showFidesBrandLink: boolean;
+
+  // URL for the brand link
+  fidesBrandUrl: string | null;
```
NevilleS, line 144: As suggested above- I think we should remove this config option. Less is more

clients/privacy-center/app/server-utils/loadEnvironmentVariables.ts
```diff
@@ -14,6 +14,10 @@ const loadEnvironmentVariables = () => {
     CONFIG_CSS_URL:
       process.env.FIDES_PRIVACY_CENTER__CONFIG_CSS_URL ||
       "file:///app/config/config.css",
+    SHOW_BRAND_LINK:
+      process.env.FIDES_PRIVACY_CENTER__SHOW_BRAND_LINK === "true" || false,
```
NevilleS, line 18: Can we complicate this rollout slightly?

I'd like this to _default_ to `true`, not `false`. But I'd like to have one release to say that we're planning to do that, so we can ensure that a couple heavily-styled existing clients set this explicitly to `false` in their environment. Can you do that? Just add a tiny PR to the next release to flip this default.

NevilleS (changes requested):
Overall this is solid- but I'd like to simplify the config a bit. We already have a lot of undocumented options so I'd like to keep it as simple as possible with a single boolean instead.

Couple other nits inline

clients/fidesui/src/icons/EthycaLogo.tsx
```diff
@@ -0,0 +1,11 @@
+import { createIcon } from "@chakra-ui/icons";
+
+export default createIcon({
+  displayName: "EthycaLogo",
+  viewBox: "0 0 31 20",
+  defaultProps: {
+    width: "20px",
```
gilluminate, line 7: shouldn't this match the view box dimensions?
```suggestion
    width: "31px",
```

gilluminate (changes requested):

clients/fides-js/src/components/ConsentButtons.tsx
gilluminate: @NevilleS What about a default URL and an optional 2nd config? Or nah, just axe the 2nd config?

clients/privacy-center/app/server-utils/loadEnvironmentVariables.ts
jpople: Sure-- ticketed flipping as [HJ-184](https://ethyca.atlassian.net/browse/HJ-184).

clients/fides-js/src/components/fides.css
jpople: The colors came from the [design](https://www.figma.com/design/MFzrOaPcszfJTcx4djQqMI/HJ-157%3A-Add-Ethyca-Logo-to-CMP-%2F-Privacy-Center?node-id=0-1&node-type=canvas&t=sRzlq5SAVaCIA76p-0); the logo is in Minos which I think we want to maintain because it's a brand color.  I can change the text to match that or be in the default text color instead if we prefer, though.

clients/fides-js/src/components/ConsentButtons.tsx
NevilleS: It's my **Strong Opinion**™️ (but weakly held) that we shouldn't add two config options for this feature unless we have a really compelling reason. I'd question the utility of the URL config- since the branding logo itself isn't configurable in the same way, why would I change it to anything else? Instead it just creates possibilities for invalid configurations where a client doesn't realize the two options are dependent on each other.

clients/fides-js/src/components/fides.css
NevilleS: There'll be cases where the background color will be too dark for the logo to be visible, so having an option to switch it from minos -> white will need to be available is all

jpople: Gotcha-- I've updated Minos to be a variable in `fides.css` over using an inline style so it should be easier to change and switched the text over to use the default font color instead of defining a new one:

Screenshot 2024-11-11 at 14 35 02

clients/fides-js/src/components/fides.css
```diff
@@ -477,6 +481,28 @@ div#fides-consent-content .fides-modal-description {
...
+
+.fides-modal-secondary-actions .fides-brand-link {
+  display: flex;
+  align-items: center;
+  gap: 4px;
+  text-decoration: none;
+  color: var(--fides-overlay-font-color);
+}
+
+#ethyca-logo {
```
NevilleS, line 502: super-nit: this CSS rule is global. `fides.css` is installed on lots of customer websites, some of which might have their own reasons to include a div with an id like `#ethyca-logo`... can we adjust this to be less likely to conflict? Let's use a class name and a selector that scopes to our own div.

NevilleS (approved):
I have one nit remaining here- but rather than block, can you address that and then merge? Thanks!

PR #5484, opened 2024-11-11
HJ-86 system endpoint new lifecycle table

thingscouldbeworse:
Closes #HJ-86

### Description Of Changes

Necessary changes for lifecycle tables. Option to retrieve only DnD relevant systems when paginating/filtering, logic to filter out "hidden" systems, and corresponding migration to add `hidden` bool column

### Code Changes

* inside the conditional block for filtering/paginating systems in the `GET` endpoint also allow for `dnd_relevant` systems, decided as 
    - systems with any connection config attached
    - systems with any dataset references
* column on `ctl_systems` `hidden`, defaults to `False`
* by default, filter out these hidden systems in the `GET` endpoint filtering, but provide a param to optionally include them

### Steps to Confirm

* [ ] using either the swagger UI and `/system/ update/create/upsert endpoints (or DB statements) create (or verify that you already have) 
    - [ ] a system connected to a `ConnectionConfig` object (via `connectionconfig.system_id`)
    - [ ] a system connected to a dataset reference (via `ctl_systems.dataset_references`)
    - [ ] a system without either of these connections
    - [ ] any system with the `hidden` column set to `True`
* [ ] use the `/system` endpoint with `GET` to retrieve systems with pagination parameters and with `dnd_relevant` set to `true`, and also `false`. `true` should exclude the system without a connection config or dataset reference
* [ ] repeat, but flipping the `show_hidden` property on/off. When on, the hidden system should be returned, otherwise not

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:


src/fides/api/api/v1/endpoints/system.py
```diff
@@ -389,37 +389,62 @@ async def ls(  # pylint: disable=invalid-name
     data_uses: Optional[List[FidesKey]] = Query(None),
     data_categories: Optional[List[FidesKey]] = Query(None),
     data_subjects: Optional[List[FidesKey]] = Query(None),
+    dnd_relevant: Optional[bool] = Query(None),
+    show_hidden: Optional[bool] = Query(False),
 ) -> List:
     """Get a list of all of the Systems.
-    If any pagination parameters (size or page) are provided, then the response will be paginated
-    & provided filters (search, taxonomy fields) will be applied.
+    If any parameters or filters are provided the response will be paginated and/or filtered.
```
erosselli, line 396: Can we check with someone in the frontend team that this change in behavior won't affect any existing functionality? e.g if we're using this endpoint anywhere in the frontend without pagination params but some filters like `data_categories`, then this will break cause that response will now be paginated but the frontend won't expect that

src/fides/api/api/v1/endpoints/system.py
```diff
@@ -389,37 +389,62 @@ async def ls(  # pylint: disable=invalid-name
...
+        PrivacyDeclaration, System.id == PrivacyDeclaration.system_id
+    )
+
+    # Fetch any system that is relevant for Detection and Discovery, ie any of the following:
+    # - has connection configurations (has some integration for DnD or SaaS)
+    # - has dataset references
+    if dnd_relevant:
+        query = query.filter(
+            (System.connection_configs != None)
+            | (System.dataset_references.any())  # noqa: E712
```
erosselli: why do we use `!= None` for one but `.any()` for the other condition?

tests/ctl/core/test_api.py
```diff
@@ -1536,6 +1536,76 @@ def test_list_with_pagination_and_multiple_filters(
...
+
+        assert result.status_code == 200
+        result_json = result.json()
+        assert result_json["total"] == 1
+        assert len(result_json["items"]) == 1
+
+        # only "system_with_cleanup" has a connection config attached to it in fixtures
+        assert result_json["items"][0]["fides_key"] == system_with_cleanup.fides_key
+
+    def test_list_with_show_hidden(
```
erosselli, line 1564: can we add a test that has BOTH show_hidden true and the dnd_filter so we test that case?

src/fides/api/api/v1/endpoints/system.py
erosselli: well yes but we can't release something if it's broken 😅 I agree that it _shouldn't_ be the case that the frontend was sending the filter params but not the pagination ones, but we should check with the frontend team just to make sure , cause returning a paginated response when the frontend isn't expecting would break it

thingscouldbeworse: we only use the systems api in one place as far as I can tell https://github.com/ethyca/fides/blob/ea84dc29bfd1b5e51b41995a9ec4e52cc311aeab/clients/admin-ui/src/pages/systems/index.tsx#L105

https://github.com/search?q=repo%3Aethyca%2Ffides+%22getSystems%22&type=code

I'll ask FE if anyone knows of more places but I don't think we have a place that's expecting an un-paginated response with filters that it also doesn't actually want

erosselli (approved):

PR #5499, opened 2024-11-15
LA-33 use shopify graphql api

Vagoasdf:
Closes [LA#33](https://ethyca.atlassian.net/browse/LA-33)

Pairs With  https://github.com/ethyca/fidesplus/pull/1728

### Description Of Changes

We are updating the Shopify Integration to use GraphQL. This allows us to save processing power by running specific requests with the data we want to use, and we also keep up to date with the current versions of the API, upgrading from one dating to 2022 to one recent from november 2024.  This also allow us to leverage the  dedicated customer data removal endpoint from shopify, instead of relying on masking each on of the customer orders and addressed directly

### Code Changes

* Introducing request overrides for Shopify Integration in order to properly use GraphQL
* Implementing pagination for said  overrides
* Removing unused endpoints: Customer_order_Transaccions, blog, and blog_articles 
* Updating tests to use the newest version
* Updating test fixtures to generate data during testing

### Steps to Confirm

1.  Pair with the new Config and Dataset from fidesplus PR  https://github.com/ethyca/fidesplus/pull/1728
2. Create a new Shopify integration
3. Make an access request with the identity connectors@ethyca.com
4. Check the data is properly accessed
5. make a delete request with the identity connectors@ethyca.com
6. On Postman, cancel the erasure requests corresponding to the customers recognized by connectors@ethyca.com:
     * gid://shopify/Customer/5692184199261 
     * gid://shopify/Customer/5753359859805
     * gid://shopify/Customer/5765598150749
     * gid://shopify/Customer/5829835063389
 
(Note: If we attempt to create another delete request without canceling the previous ones, the request will fail with the error _Failed to request customer data erasure_  )

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] No followup issues
* Documentation:
  * [ ] If there are any new client scopes created as part of the pull request, remember to update public-facing


tests/ops/integration_tests/saas/test_shopify_task.py
```diff
@@ -1,828 +1,68 @@
...
+        for customer in access_results[f"{key}:customers"]:
+            assert customer["email"] == shopify_identity_email
+
+        ## Assert for Orders by Customer
+        orders = access_results[f"{key}:customer_orders"]
+        assert len(orders) == len(shopify_access_data["orders"])
+        for order in orders:
+            assert order["email"] == shopify_identity_email
+
+        ## Assert for blog article comments
```
galvana, line 36: Comments should only start with a single `#`
```suggestion
        # Assert for blog article comments
```

tests/fixtures/saas/shopify_fixtures.py
```diff
@@ -1,30 +1,20 @@
...
 
 from fides.api.cryptography import cryptographic_util
-from fides.api.db import session
-from fides.api.models.connectionconfig import (
-    AccessLevel,
-    ConnectionConfig,
-    ConnectionType,
-)
 from fides.api.models.datasetconfig import DatasetConfig
 from fides.api.models.sql_models import Dataset as CtlDataset
```
galvana, line 11: It looks like these imports are no longer used, we can get rid of them
```suggestion
```

tests/fixtures/saas/shopify_fixtures.py
```diff
@@ -33,102 +23,163 @@ def shopify_secrets(saas_config):
         "domain": pydash.get(saas_config, "shopify.domain") or secrets["domain"],
         "access_token": pydash.get(saas_config, "shopify.access_token")
         or secrets["access_token"],
+        "fixtures_access_token": pydash.get(
```
galvana, line 26: How is this `fixtures_acess_token` different from the other `access_token`? If it has write access for example, we should rename it to be able to tell how it's different

galvana (commented):
Nice work figuring out GraphQL, pagination, and the test data seeding! I made some minor recommendations but I'd like to meet in person to discuss some of the GraphQL queries

tests/fixtures/saas/shopify_fixtures.py
Vagoasdf: Yes, it has write access for the fixture data operations: creating customers, orders, blogs, articles and comments. 

What would be a good name? I was using `fixtures_access_token` to signal that it was used on the fixtures only.  Other prefixes that i can think is 
*  `write_`
*  `data_creation_`
* `seeding_`

I still think that `fixtures_` is better, since it signals _where_ it's used

galvana (approved):

PR #5524, opened 2024-11-20
Updating dataset PUT to allow deleting all datasets

galvana:
Closes [LA-84](https://ethyca.atlassian.net/browse/LA-84)

### Description Of Changes

Adding manual commit to the `PUT /dataset` endpoint to be able to persist the deletes before updating or creating datasets.

### Code Changes

* Updated the affected endpoint and added a test

### Steps to Confirm

1.  Create a system with a DB integration
2. Add a dataset and save
3. Remove all datasets and save, the datasets should still be removed after a page refresh

### Pre-Merge Checklist

* [x] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [x] No followup issues
* Database migrations:
  * [x] No migrations
* Documentation:

[LA-84]: https://ethyca.atlassian.net/browse/LA-84?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/fides/api/api/v1/endpoints/dataset_endpoints.py
```diff
@@ -214,7 +214,8 @@ def put_dataset_configs(
         db.query(DatasetConfig).filter(
             DatasetConfig.connection_config_id == connection_config.id,
             DatasetConfig.fides_key.in_(config_keys_to_remove),
-        ).delete()
+        ).delete(synchronize_session=False)
```
eastandwestwind, line 217: what's the default `synchronize_session` if not provided?

eastandwestwind (approved):
Looks good, just 1 question for my understanding.

src/fides/api/api/v1/endpoints/dataset_endpoints.py
galvana: The default is `evaluate`. We're using `False` here since we don't need to maintain the deleted objects in the session state, we're just removing records that match our filter criteria.

PR #5532, opened 2024-11-22
Send property id and also override path.

tvandort:
### Description Of Changes

Fetch correct experience when a privacy center property is being use. Add env variable to override use a specific property for the root domain (when no property paths are used).

### Code Changes

* If property_id is available, send it as a paramater to the fetch experience function
* Add env variable to override use a specific property 

### Steps to Confirm

### Pre-Merge Checklist

* [x] Issue requirements met
* [x] `CHANGELOG.md` updated
* Followup issues:
  * [ ] No followup issues

### DIFF
--- a/clients/privacy-center/app/server-utils/loadEnvironmentVariables.ts
+++ b/clients/privacy-center/app/server-utils/loadEnvironmentVariables.ts
@@ -16,7 +16,9 @@ const loadEnvironmentVariables = () => {
       "file:///app/config/config.css",
     SHOW_BRAND_LINK:
       process.env.FIDES_PRIVACY_CENTER__SHOW_BRAND_LINK === "true" || false,
-    CUSTOM_PROPERTIES: process.env.CUSTOM_PROPERTIES === "true" || true,
+    CUSTOM_PROPERTIES: process.env.CUSTOM_PROPERTIES !== "false", // default: true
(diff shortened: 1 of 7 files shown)


clients/privacy-center/app/server-utils/loadEnvironmentVariables.ts
```diff
@@ -16,7 +16,9 @@ const loadEnvironmentVariables = () => {
       "file:///app/config/config.css",
     SHOW_BRAND_LINK:
       process.env.FIDES_PRIVACY_CENTER__SHOW_BRAND_LINK === "true" || false,
-    CUSTOM_PROPERTIES: process.env.CUSTOM_PROPERTIES === "true" || true,
+    CUSTOM_PROPERTIES: process.env.CUSTOM_PROPERTIES !== "false", // default: true
```
tvandort, line 19: I'm not sure I understand the "why" behind this change, can you elaborate?

lucanovera: Fixing a bug I noticed. || true will always be true even if you set it to "false". So I fix that.

It shouldn't matter too much, the default is true and I doubt anyone will set it to false. But just in case I wanted to fix it.

lucanovera (approved):
Tested the fix for fetching the correct experience when using properties and it works correctly. Tested the override env variable and works correctly too. Approved.

PR #5545, opened 2024-11-26
FIDES-1606: Change BlueConic Consent Behavior

tvandort:
Set Consented Objects to all if Fides detects that a user is in a region with no consent preferences.

Closes [FIDES-1606]

### Description Of Changes

Update BlueConic integration to communicate that objectives are consented to when Fides detects that the user is in a state with no consent options.

### Code Changes

### Steps to Confirm

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:

[FIDES-1606]: https://ethyca.atlassian.net/browse/FIDES-1606?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/clients/fides-js/__tests__/integrations/blueconic.test.ts
+++ b/clients/fides-js/__tests__/integrations/blueconic.test.ts
@@ -1,5 +1,3 @@
-import { addSyntheticLeadingComment } from "typescript";
-
 import { FidesGlobal } from "../../src/fides";
 import { blueconic } from "../../src/integrations/blueconic";
 import { MARKETING_CONSENT_KEYS } from "../../src/lib/consent-constants";
--- a/clients/fides-js/__tests__/integrations/blueconic.test.ts
+++ b/clients/fides-js/__tests__/integrations/blueconic.test.ts
@@ -35,6 +35,12 @@ const setupFidesWithConsent = (key: string, optInStatus: boolean) => {
   } as any as FidesGlobal;
 };
 
+const setupFidesWithoutConsent = () => {
+  window.Fides = {
+    consent: {},
+  } as any as FidesGlobal;
+};
+
 describe("blueconic", () => {
   afterEach(() => {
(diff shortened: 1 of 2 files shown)


clients/fides-js/__tests__/integrations/blueconic.test.ts
```diff
@@ -58,6 +64,22 @@ describe("blueconic", () => {
     ).not.toHaveBeenCalled();
   });
 
+  test("when fides configures no consent, blueconic sets consent for all purposes", () => {
```
NevilleS, line 67: nit: I'd add an explicit test outside of this case for something like:
```
test("when no approach is specified, 'onetrust' is used by default", ...)
```

...and then I'd wrap these other tests with a describe like:
```
describe("when using the default approach", () => { ... }
```

Right now it's subtle that there are some existing paved roads here to allow creating different approaches in the future.

NevilleS (approved):
Couple nits, but this makes sense.

clients/fides-js/__tests__/integrations/blueconic.test.ts
tvandort: Good call, will do that.

tvandort: I couldn't _quite_ do the first suggestion because there was no direct observable way with this code to check that `'onetrust'` is a "default". So instead I've rewritten the tests to assert that `undefined` and `"onetrust"` approaches being passed in have identical behavior.

gilluminate (approved):
Looks good to me. Tests are passing locally for me.

PR #5553, opened 2024-12-02
HJ-97 - Update systems endpoint to filter vendor deleted systems

andres-torres-marroquin:
Closes #HJ-97

### Description Of Changes

Update systems endpoint to filter vendor deleted systems

### Code Changes

* Updates `/system` list endpoint to filter out vendor deleted systems.

### Steps to Confirm

For System resource:
1.  Create a system
2. Update the `vendor_deleted_at` field for the newly created system to something in the past (if it is null or in the future it won't be filtered out)
3. List `/system` the created system should be filtered out.

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,9 +19,12 @@ The types of changes are:
 
 ### Added
 - Added new column for Action Type in privacy request event logs [#5546](https://github.com/ethyca/fides/pull/5546)
+- Added `fides_consent_override` option in FidesJS SDK [#5541](https://github.com/ethyca/fides/pull/5541)
+- Added new `script` ConsentMethod in FidesJS SDK for tracking automated consent [#5541](https://github.com/ethyca/fides/pull/5541)
 
 ### Changed
 - Adding hashes to system tab URLs [#5535](https://github.com/ethyca/fides/pull/5535)
+- Updated Cookie House to be responsive [#5541](https://github.com/ethyca/fides/pull/5541)
 - Updated `/system` endpoint to filter vendor deleted systems [#5553](https://github.com/ethyca/fides/pull/5553)
 
 ### Developer Experience
--- a/clients/README.md
+++ b/clients/README.md
@@ -1,5 +1,3 @@
 # Clients
 
 The clients directory houses all front-end packages and shared code amongst clients, and also includes e2e tests.
-
-See the [UI Contribution Guide](http://localhost:8000/fides/development/ui/overview/) for more information
--- a/clients/admin-ui/src/types/api/models/ConsentMethod.ts
+++ b/clients/admin-ui/src/types/api/models/ConsentMethod.ts
@@ -6,6 +6,7 @@ export enum ConsentMethod {
   BUTTON = "button",
   REJECT = "reject",
   ACCEPT = "accept",
+  SCRIPT = "script",
   SAVE = "save",
   DISMISS = "dismiss",
   GPC = "gpc",
--- a/clients/fides-js/docs/interfaces/FidesOptions.md
+++ b/clients/fides-js/docs/interfaces/FidesOptions.md
@@ -1,7 +1,7 @@
 # Interface: FidesOptions
 
 FidesJS supports a variety of custom options to modify it's behavior or
-enabled more advanced usage. For example, the `fides_locale` option can be
+enable more advanced usage. For example, the `fides_locale` option can be
 provided to override the browser locale. See the properties list below for
 the supported options and example usage for each.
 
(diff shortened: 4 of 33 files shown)


src/fides/api/api/v1/endpoints/system.py
```diff
@@ -446,6 +446,20 @@ async def ls(  # pylint: disable=invalid-name
...
+        or page
+        or search
+        or data_uses
+        or data_categories
+        or data_subjects
+        or dnd_relevant
+        or show_hidden
+    ):
+        result = await db.execute(duplicates_removed)
+        return result.scalars().all()
```
adamsachs, line 461: any particular reason that we're changing the functionality here? if it's just a general optimization, i understand that motivation - but i think i'd lean toward keeping this using the simple `return await list_resource(System, db)`, because i _believe_ this was in place for backward compatibility of this endpoint with uses in the fides CLI, etc - when it specifically did not paginate, filter, or deduplicate (or do anything besides list all the systems, raw).

granted, that backward compatibility is not well commented in the code, so if we do keep this as it was, then a nice addition would be a code comment to clarify why it's kept as `list_resource`, i.e. for backward compatibility reasons.

i see/know that @erosselli and @galvana worked on refactoring this endpoint a little while back, may be worth getting them to chime in quickly.

also, as more of a stylistic comment - i preferred this conditional coming at the _beginning_ of the endpoint function, as it had been previously. if there's a branch of the function that's a "special case" and returns immediately, i'd prefer to see that at the beginning - i just find that easier to make sense of when reading the function...

tests/ctl/core/test_api.py
```diff
@@ -1665,6 +1665,40 @@ def test_list_with_pagination_and_multiple_filters_2(
...
 
+    @pytest.mark.parametrize(
+        "vendor_deleted_date, expected_systems_count, show_deleted",
+        [
+            (datetime.now() - timedelta(days=1), 1, True),
+            (datetime.now() - timedelta(days=1), 0, None),
+            (datetime.now() + timedelta(days=1), 1, None),
+            (None, 1, None),
+        ],
+    )
```
adamsachs, line 1676: nice test parameterization 👍

tests/ctl/core/test_api.py
```diff
@@ -1665,6 +1665,40 @@ def test_list_with_pagination_and_multiple_filters_2(
 
         assert result_json["items"][0]["fides_key"] == tcf_system.fides_key
 
+    @pytest.mark.parametrize(
+        "vendor_deleted_date, expected_systems_count, show_deleted",
+        [
+            (datetime.now() - timedelta(days=1), 1, True),
+            (datetime.now() - timedelta(days=1), 0, None),
```
adamsachs, line 1672: nit - reads a bit weird to me that the value here is `True` or `None`, although i know that `None` is falsey 😛 
```suggestion
            (datetime.now() - timedelta(days=1), 0, False),
```

adamsachs (approved):
OK, this is looking good in some brief manual testing - I'll tentatively approve, but a few notes:

- i left a [comment](https://ethyca.atlassian.net/browse/HJ-97?focusedCommentId=45290) on the ticket to confirm with product that we're OK filtering these out of the system list endpoint (and system inventory page) while _not_ filtering out of the reporting pages
- there are some minor points in the code about backward compatibility/stylistic points that'd be nice to address
- it's probably worth improving the "steps to confirm" in the PR description, since adding a "vendor deleted date" is not a self-explanatory step, since the UI doesn't support manually editing this field (i don't think?). i'd either:
    - provide steps on how to test this integrated with fidesplus - which would consist of creating a system based on a compass vendor who actually has a`vendor_deleted_date` populated. this is the _best_ test, since it more closely mimics the actual use case in the ticket. note that i didn't complete this testing myself.
    - provide a sample API request that you used to set a `vendor_deleted_date` "manually" on a system. this is probably good enough to get the PR merged, but i do think that generally we should make sure there are some testing steps listed out here for the "end to end" functionality, which includes vendors from compass with `vendor_deleted_date`s.

src/fides/api/api/v1/endpoints/system.py
galvana: Hey guys, yes, that's the reason we kept the non-paginated version of the system retrieval, for backwards compatibility. I also agree with @adamsachs's comment about having this higher up so it's a little more obvious that there's a possible early return.

PR #5555, opened 2024-12-03
LA-166 ux issues when saving an integration

lucanovera:
### Description Of Changes

Fix unexpected re-enabling of an integration after saving. Fix validation issues with boolean field.

### Code Changes

* Remove hardcoded "enabled" value during connection patch
* Replace text input with a select for boolean values in connection form

### Steps to Confirm

1.  Go to admin, add system
2. Add an integration of postgres type
3. Fill required fields and save
4. Check that the SSH required field is now a select with true/false values instead of a text input
5. Toggle Disable Integration
6. Now, save the form again
7. Check that the integration is still disabled after saving

### Pre-Merge Checklist

* [x] Issue requirements met
* [x] `CHANGELOG.md` updated


clients/admin-ui/cypress/support/stubs.ts
```diff
@@ -240,6 +240,10 @@ export const stubPlus = (available: boolean, options?: HealthCheck) => {
       statusCode: 400,
       body: {},
     }).as("getPlusHealth");
+    cy.intercept("GET", "/api/v1/plus/*", {
```
lucanovera, line 243: If we're not in plus, all plus endpoints should return 404.

clients/admin-ui/src/features/datastore-connections/system_portal_config/forms/ConnectorParametersForm.tsx
```diff
@@ -7,6 +7,7 @@ import { useLazyGetDatastoreConnectionStatusQuery } from "datastore-connections/
 import DSRCustomizationModal from "datastore-connections/system_portal_config/forms/DSRCustomizationForm/DSRCustomizationModal";
 import {
   AntButton as Button,
+  AntSelect,
```
gilluminate, line 10: looks like this is now being imported twice. Can we just use the `Select` version?

clients/admin-ui/src/features/datastore-connections/system_portal_config/forms/ConnectorParametersForm.tsx
```diff
@@ -179,7 +183,17 @@ export const ConnectorParametersForm = ({
                   size="sm"
                 />
               )}
-              {item.type === "integer" && (
+              {isBoolean && (
+                <AntSelect
```
gilluminate, line 187: is there a reason we chose to use a select rather than a checkbox which is typically used for booleans on a form?

clients/admin-ui/src/features/datastore-connections/system_portal_config/forms/ConnectorParametersForm.tsx
lucanovera: Yess! I left just the Select one now.

clients/admin-ui/src/features/datastore-connections/system_portal_config/forms/ConnectorParametersForm.tsx
lucanovera: I used a checkbox initially, which I think it's better UX. 
But with the current layout for the form, it looks weird because of the empty space it creates and having the ? icon all the way to the right. Jack asked me to change that to a Select instead.
Captura de pantalla 2024-12-03 a la(s) 1 09 50 p  m

I think this is still an improvement over the input text, but in the future I would like to see a layout that supported the checkbox better.

gilluminate (approved):
Tested locally and it is working well!

PR #5566, opened 2024-12-06
Minor Demo fixes

gilluminate:
Some minor fixes for the Embedded Demo page.

### DIFF
--- a/clients/fides-js/src/components/Overlay.tsx
+++ b/clients/fides-js/src/components/Overlay.tsx
@@ -153,7 +153,7 @@ const Overlay: FunctionComponent<Props> = ({
   }, [showBanner, setBannerIsOpen]);
 
   useEffect(() => {
-    if (!experience || options.modalLinkId === "") {
+    if (options.fidesEmbed || !experience || options.modalLinkId === "") {
       // If empty string is explicitly set, do not attempt to bind the modal link to the click handler.
       // developers using `Fides.showModal();` can use this to prevent polling for the modal link.
       return () => {};
@@ -211,7 +211,13 @@ const Overlay: FunctionComponent<Props> = ({
       }
       window.Fides.showModal = defaultShowModal;
     };
-  }, [options.modalLinkId, options.debug, handleOpenModal, experience]);
+  }, [
+    options.fidesEmbed,
+    options.modalLinkId,
+    options.debug,
+    handleOpenModal,
+    experience,
+  ]);
 
   const handleManagePreferencesClick = (): void => {
     handleOpenModal();
--- a/clients/fides-js/src/components/fides.css
+++ b/clients/fides-js/src/components/fides.css
@@ -485,7 +485,7 @@ div#fides-consent-content .fides-modal-description {
   justify-content: center;
 }
 
-.fides-modal-container .fides-button-group-brand {
+.fides-modal-footer .fides-button-group-brand {
   min-height: var(--fides-overlay-modal-secondary-button-group-height);
 }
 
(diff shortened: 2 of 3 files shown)


clients/fides-js/src/components/Overlay.tsx
```diff
@@ -153,7 +153,7 @@ const Overlay: FunctionComponent<Props> = ({
   }, [showBanner, setBannerIsOpen]);
 
   useEffect(() => {
-    if (!experience || options.modalLinkId === "") {
+    if (options.fidesEmbed || !experience || options.modalLinkId === "") {
```
gilluminate, line 156: we don't want to be constantly polling for a modal link when the CMP is embedded in the page!

clients/fides-js/src/components/fides.css
```diff
@@ -485,7 +485,7 @@ div#fides-consent-content .fides-modal-description {
   justify-content: center;
 }
 
-.fides-modal-container .fides-button-group-brand {
+.fides-modal-footer .fides-button-group-brand {
```
gilluminate, line 488: `.fides-modal-container` doesn't exist when things are embedded, so let's target the modal footer class instead.

clients/sample-app/src/pages/embedded-consent.tsx
```diff
@@ -60,7 +60,17 @@ const IndexPage = ({ gtmContainerId, privacyCenterUrl }: Props) => {
...
+        <style>{`
+          #fides-embed-container {
+            --fides-overlay-width: 'auto'
+          }
+
+          body {
+            font-family: "Inter", sans-serif;
+            color: #171923;
+            font-size: 14px;
+          }
```
gilluminate, line 72: fonts were all wonky on this page otherwise. This mimics the font choices of the Cookie House homepage.

clients/fides-js/src/components/fides.css
jpople: Since this component exists in both the modal and the embed, is it worth trying to rename it to something more generic or would that break too much stuff?

clients/fides-js/src/components/fides.css
```diff
@@ -1023,13 +1023,13 @@ div#fides-overlay-wrapper .fides-toggle .fides-toggle-display {
   position: relative;
 }
 
-.fides-modal-container .fides-i18n-menu {
+.fides-modal-footer .fides-i18n-menu {
```
jpople, line 1026: Screenshot 2024-12-06 at 13 52 34

These styles still don't seem to be applying for me and the i18n menu is still centered, is that intentional?

clients/fides-js/src/components/fides.css
gilluminate: which component are you referring to, the footer or the brand?

clients/fides-js/src/components/fides.css
gilluminate: weird, that was working for me. Try rebuilding/clearing your cache?

clients/fides-js/src/components/fides.css
jpople: Sorry, the modal footer, that is.  Or anything else that's getting used here that has `modal` in the name, I suppose.

gilluminate: Yeah, I think eventually we'll want to rename everything as "layer1" and "layer2" instead for that reason. I don't think this now is the time to do that. 👍

jpople (approved):

PR #5569, opened 2024-12-06
LA 194 refreshing oauth credentials

Vagoasdf:
Closes [LA#194](https://ethyca.atlassian.net/browse/LA-194)

### Description Of Changes

We were seeing trouble with OAuth Client Credentials strategies, as the token is not being refreshed properly, as the base class was expecting a refresh request to exist. For that, we are adding an override function on the Client Credentials strategy that can properly refresh the token

### Code Changes

Adding a override function on OAuth2ClientCredentialsAuthenticationStrategy for refreshing the access token, using the base token request instead of the refresh request, since the Client Credentials OAuth connection does not uses a separate refresh request call.

### Steps to Confirm

1.  Set up a local OAuth2 Credentials integration. Currently PowerReviews is our only working Oauth2 Integration 
2. Execute an erasure request to confirm a working integration
3. Wait for the Expiration of the access_token request. PowerReviews by defaults sets their expires_in at 3600 seconds
4. After the time has passed, execute another erasure request. Previously, this would've return a 401 expired token request. Now it should refresh the token properly

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:

### DIFF
--- a/src/fides/api/service/authentication/authentication_strategy_oauth2_client_credentials.py
+++ b/src/fides/api/service/authentication/authentication_strategy_oauth2_client_credentials.py
@@ -42,8 +42,6 @@ class OAuth2ClientCredentialsAuthenticationStrategy(OAuth2AuthenticationStrategy
         For Client Credentials OAuth we are not using a refresh request
         """
 
-        logger.info("Checking if the access token for {} needs to be refreshed for a Client Credentials", connection_config.key)
-        logger.info('Value of Expires at is {}', connection_config.secrets.get("expires_at"))
 
         expires_at = connection_config.secrets.get("expires_at")  # type: ignore
         if self._close_to_expiration(expires_at, connection_config):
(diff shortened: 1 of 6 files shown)


src/fides/api/service/authentication/authentication_strategy_oauth2_client_credentials.py
```diff
@@ -33,3 +33,21 @@ def add_authentication(
         # add access_token to request
         request.headers["Authorization"] = "Bearer " + access_token
         return request
+
+    def _refresh_token(self, connection_config: ConnectionConfig) -> str:
+        """
+        Persists and returns a refreshed access_token if the token is close to expiring.
+        Otherwise just returns the existing access_token.
+        For Client Credentials OAuth we are not using a refresh request
```
galvana, line 41:
```suggestion

        For the Client Credentials OAuth flow we reuse the access request to get a new token.
```

PR #5573, opened 2024-12-06
Putting dataset test results behind security flag

galvana:
Closes [LA-198](https://ethyca.atlassian.net/browse/LA-198)

### Description Of Changes

Adding the `dsr_testing_tools_enabled` security setting and using it to guard two endpoints:
- `POST /api/v1/connection/{connection_key}/dataset/{fides_key}/test`
- `POST /api/v1/privacy-request/{privacy_request_id}/filtered-results`

### Steps to Confirm

1.  Set `FIDES__SECURITY__DSR_TESTING_TOOLS_ENABLED` to `False`
2. Start Fides
3. Attempt to run a test, you should see an error toast with the message "DSR testing tools are not enabled."
image
The user will still be able to use the dataset editor and reachability status checker since they don't involve any user data.

### Pre-Merge Checklist

* [x] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [x] No followup issues
* Database migrations:
  * [x] No migrations
* Documentation:

[LA-198]: https://ethyca.atlassian.net/browse/LA-198?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


eastandwestwind (approved):
Looks good, thanks for adding tests too!

src/fides/api/api/v1/endpoints/privacy_request_endpoints.py
```diff
@@ -2662,10 +2662,14 @@ def get_test_privacy_request_results(
...
+    results = json.loads(escaped_json)
 
     return {
         "privacy_request_id": privacy_request.id,
         "status": privacy_request.status,
-        "results": escaped_data,
+        "results": (
+            results
+            if CONFIG.security.dsr_testing_tools_enabled
+            else "DSR testing tools are not enabled, results will not be shown."
```
eastandwestwind, line 2673: Hm, I feel like since this whole endpoint deals with the privacy request results, we should just return 403 instead of returning just the id and status of the request. Feels weird to create the results, too, even if the flag is disabled.

Thoughts?

eastandwestwind (changes requested):
Just 1 nit, but otherwise looks good!

src/fides/api/api/v1/endpoints/privacy_request_endpoints.py
galvana: I did this as a quick way to hide the results, but not change the way the front-end checks for a completed test run. The front-end checks for the status so if it can't find it, it doesn't know to stop polling. I think a better approach would be for the 403 to stop the polling but I wanted to minimize the changes so close to release. I can do this in a follow up ticket if that's ok?

eastandwestwind (approved):

PR #5575, opened 2024-12-09
Allow masking without primary keys

galvana:
Closes [LA-95](https://ethyca.atlassian.net/browse/LA-95)

### Description Of Changes

Removed the need to include primary keys for erasure requests for BigQuery and Postgres. Before this change, all connectors required collections to have one field that contained `primary_key: True` for that collection to be masked. In order to make this an incremental change, I added a `requires_primary_keys` property to `BaseConnector` which I defaulted to `True`. Every connector can either override this, or keep the default behavior.  

### Code Changes

* BigQuery - Set `requires_primary_keys` to False
* Postgres - Set `requires_primary_keys` to False
* ScyllaDB - Set `requires_primary_keys` to True and included a comment about ScyllaDB requiring the where clause for an update to be a primary key

### Steps to Confirm

1. Remove all the `primary_key` annotations from `src/fides/data/sample_project/sample_resources/postgres_example_test_dataset.yml`
2. Run `nox -s "fides_env(test)"
3. Run an access and erasure request for `jane@example.com`. Verify both are still possible without primary keys.

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:

[LA-95]: https://ethyca.atlassian.net/browse/LA-95?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/fides/api/task/graph_task.py
```diff
@@ -603,12 +603,19 @@ def erasure_request(
...
+
         # if there is no primary key specified in the graph node configuration
         # note this in the execution log and perform no erasures on this node
-        if not self.execution_node.collection.contains_field(lambda f: f.primary_key):
+        if (
+            self.connector.requires_primary_keys
+            and not self.execution_node.collection.contains_field(
+                lambda f: f.primary_key
+            )
+        ):
```
galvana, line 614: Key change - This is where I added the additional check for `requires_primary_keys`

src/fides/api/service/connectors/query_configs/query_config.py
```diff
@@ -100,6 +100,15 @@ def primary_key_field_paths(self) -> Dict[FieldPath, Field]:
...
         }
 
+    @property
+    def reference_field_paths(self) -> Dict[FieldPath, Field]:
+        """Mapping of FieldPaths to Fields that have incoming identity or dataset references"""
+        return {
+            field_path: field
+            for field_path, field in self.field_map().items()
+            if field_path in {edge.f2.field_path for edge in self.node.incoming_edges}
+        }
```
galvana, line 110: Very important change - this is what allows us to use the dataset references instead of the primary key fields for the where clauses

.github/workflows/backend_checks.yml
```diff
@@ -256,6 +256,7 @@ jobs:
 
     runs-on: ubuntu-latest
     timeout-minutes: 45
+    continue-on-error: true
```
eastandwestwind, line 259: this was previously removed because it caused tests that timed out to pass. Instead they should fail on timeout. More context here- https://ethyca.slack.com/archives/CBKK8TS74/p1733926881926369. Why did we need to add this back?

eastandwestwind (commented):
Manually tested and this is working great!

Just 1 q to address, below

.github/workflows/backend_checks.yml
galvana: I removed it! I was seeing some weird behavior earlier where a test failures caused the rest of the test suite to be skipped

eastandwestwind (approved):

PR #5596, opened 2024-12-11
La 175 segment unify pagination loop fix

Vagoasdf:
Closes LA#175

### Description Of Changes

Sets up a new parameter on link pagination, to check for a "has next propiety", optionally included on the response parameters, which determines if there is a next link to be searched for

Some endpoints (see [Segment Unify](https://segment.com/docs/unify/profile-api/#api-reference) ) rely on this response parameter, and they return the current page on the has next value

### Code Changes

* updated pagination strategy

### Steps to Confirm

 Ideally we would check with Segment Unify. Currently being tested with Unit Testing

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:


tests/ops/service/pagination/test_pagination_strategy_link.py
```diff
@@ -49,6 +49,42 @@ def response_with_empty_string_link():
...
+@pytest.fixture(scope="function")
+def response_with_has_next_conditional_true():
+    response = Response()
+    response._content = bytes(
+        json.dumps(
+            {
+                "customers": [{"id": 1}, {"id": 2}, {"id": 3}],
+                "links": {
+                    "next": "https://domain.com/customers?page=def",
+                    "hasNext": "true",
```
galvana, line 61: Now that we know what the actual payload looks like, we should replace this with a boolean
```suggestion
                    "hasNext": True,
```
Same thing for the other fixture

src/fides/api/service/pagination/pagination_strategy_link.py
```diff
@@ -40,6 +41,12 @@ def get_next_request(
         if not response_data:
             return None
 
+        if self.has_next:
+            has_next = pydash.get(response.json(), self.has_next)
+            logger.info("The hasNext value is {}", has_next)
```
galvana, line 46:
```suggestion
            logger.info(f"The {self.has_next} field has a value of {has_next}")
```

tests/ops/service/pagination/test_pagination_strategy_link.py
Vagoasdf: Replacing them with base boolean. 

I was thinking, would it be redundant to also test that it keeps catching these string values correctly? I.E if a API returns TRUE/FALSE in all caps, or something similar

galvana (approved):

tests/ops/service/pagination/test_pagination_strategy_link.py
galvana: I think this is ok for now, we can handle those use cases if we run into them

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ The types of changes are:
 ### Added
 - New page in the Cookie House sample app to demonstrate the use of embedding the FidesJS SDK on the page [#5564](https://github.com/ethyca/fides/pull/5564)
 - Added new erasure tests for BigQuery Enterprise [#5554](https://github.com/ethyca/fides/pull/5554)
+- Added support for boolean parameter that checks next page on linked pagination strategy [#5596](https://github.com/ethyca/fides/pull/5596)
```
galvana, line 23:
```suggestion
- Added new has_next parameter for the "link" pagination strategy [#5596](https://github.com/ethyca/fides/pull/5596)
```

PR #5613, opened 2024-12-17
HJ-319 Add DBCache model and table

erosselli:
Implements part of [HJ-319](https://ethyca.atlassian.net/browse/HJ-319)

### Description Of Changes

Adds a `DBCache` model and table to be used for persistent caching. The table has the `namespace` and `cache_key` columns that uniquely determine a row, and the column `cache_value` is used for storing an arbitrary value. For now we'll only use this for the list privacy experiences endpoint, but in the future it should be easy to add other things as well. 

### Code Changes

- Create `DBCache` model and its corresponding migration 
  -  We index both over namespace (since we may want to implement a "clear all entries for a namespace") and over the (namespace, cache_key) pair (for lookups) 

### Steps to Confirm

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:

[HJ-319]: https://ethyca.atlassian.net/browse/HJ-319?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/fides/api/models/db_cache.py
+++ b/src/fides/api/models/db_cache.py
@@ -1,7 +1,7 @@
 from enum import Enum
 from typing import ByteString, Optional
 
-from sqlalchemy import Column, String, Index
+from sqlalchemy import Column, Index, String
 from sqlalchemy.dialects.postgresql import BYTEA
 from sqlalchemy.orm import Session
 


src/fides/api/models/db_cache.py
```diff
@@ -0,0 +1,82 @@
...
+    """
+    Cache table for storing arbitrary data, useful when in-memory caches aren't enough.
+    For example if cache contents need to be persisted across server restarts,
+    or if the cache needs to be shared across different Fides instances.
+    """
+
+    namespace = Column(
+        String, nullable=False, index=True
+    )  # Add a namespace since the same cache key could technically be used for different contexts
+    cache_key = Column(String, nullable=False, index=True)
```
adamsachs, line 27: hmm - i'm not sure how much it matters, but i _think_ for a composite index (which this is, right?) it's generally [specified a bit differently](https://docs.sqlalchemy.org/en/20/core/constraints.html#index-api)?

and we have some examples in our code base, e.g. [here](https://github.com/ethyca/fides/blob/main/src/fides/api/models/system_history.py#L30-L34)

src/fides/api/models/db_cache.py
```diff
@@ -0,0 +1,82 @@
...
+    cache_value = Column(BYTEA, nullable=False)
```
adamsachs, line 28: ok, i hope this works well and i wasn't silly to suggest a byte column :)

adamsachs (approved):
this looks solid to me! just a nit on the sqlalchemy representation of the composite index 👍

src/fides/api/models/db_cache.py
erosselli: ah I forgot to specify the composite index on the model definition, I just added it in the migration 🤦‍♀️  I did add index=True just to namespace thinking of queries like "Let's remove all entries for a namespace" , that don't have a specific cache_key but maybe the composite index still works for those kinds of queries? I'm a bit rusty on the specifics of composite indexes 😅

adamsachs: heh, i'm not really an expert either. from my general understanding, i don't _think_ the composite index would work well specifically for that use case - but i haven't validated that assumption at all.

i was aware of your intention with the two indexes there, and i didn't really question it - so i'd say it's fine to keep that in place! 👍

PR #5616, opened 2024-12-17
initial web monitor support (fides)

adamsachs:
Closes #<issue>

### Description Of Changes

### Code Changes

### Steps to Confirm

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:


adamsachs:
@thingscouldbeworse had you taken a look at this fides-side PR? if so, and it looks good, would you mind giving an approval? thanks!

src/fides/api/alembic/migrations/versions/021166731846_add_asset_table.py
```diff
@@ -0,0 +1,105 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "021166731846"
+down_revision = "58f8edd66b69"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # stored separately to be referenced in md5 expression on index
```
thingscouldbeworse, line 21: What is stored separately?

src/fides/api/alembic/migrations/versions/c961528edfc6_add_website_connection_type.py
```diff
@@ -0,0 +1,108 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "c961528edfc6"
+down_revision = "021166731846"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # Add 'datahub' to ConnectionType enum
```
thingscouldbeworse, line 20: copypasta comment from the datahub migration

src/fides/api/alembic/migrations/versions/c961528edfc6_add_website_connection_type.py
```diff
@@ -0,0 +1,108 @@
...
+        """
+        ALTER TABLE connectionconfig ALTER COLUMN connection_type TYPE connectiontype USING
+        connection_type::text::connectiontype
+    """
+    )
+    op.execute("DROP TYPE connectiontype_old")
+
+
+def downgrade():
+    # Remove 'datahub' from ConnectionType enum
```
thingscouldbeworse, line 66: copypasta comment

src/fides/api/models/asset.py
```diff
@@ -0,0 +1,156 @@
...
+                cls.domain == data["domain"],
+                cls.system_id == data["system_id"],
+                cls.base_url == data.get("base_url"),
+            )
+        )
+        existing_record = result.scalars().first()
+        record_id: str
+        if existing_record:
+            await async_session.execute(
+                update(cls).where(cls.id == existing_record.id).values(data)  # type: ignore[arg-type]
```
thingscouldbeworse, line 121: I haven't tried this myself just yet but do we know that this `update(...).where(...)` will properly update any field, even `meta`? Or does it have the weird behavior with sqlalchemy where changing a json property of an object and actually getting it saved requires the `flag_modified` call?

src/fides/api/models/detection_discovery.py
```diff
@@ -232,11 +233,38 @@ class StagedResource(Base):
...
+    system_id = Column(
+        String,
+        ForeignKey(System.id_field_path),
+        nullable=True,
+        index=True,
+    )
+
+    # TODO: we should be able to enable the below relationship, but it
+    # but it confuses different functionality since 'system' means different
+    # things depending on whether the resource is a database or web monitor
```
thingscouldbeworse, line 252:
```suggestion
    # things depending on whether the resource is a datastore or web monitor
```

src/fides/api/models/detection_discovery.py
```diff
@@ -352,7 +380,6 @@ class MonitorExecution(Base):
     configuration details used in connecting to the external data store.
     """
 
-    id = Column(String, primary_key=True)
```
thingscouldbeworse, line 355: What's happening here? We're removing the `id` property of `MonitorExecution`?

src/fides/api/service/connectors/website_connector.py
```diff
@@ -0,0 +1,81 @@
...
+            website_url,
+        )
+        try:
+            response = head(self.configuration.secrets["url"])
+        except Exception as error:
+            raise ConnectionException(str(error))
+
+        if response.status_code >= 400:
+            raise ConnectionException(
+                f"HEAD request to '{website_url}' resulted in error status code: '{response.status_code}'"
```
thingscouldbeworse, line 50: Do we expect that this will often fail if you're configuring the monitor from outside the proxy the web monitor service is running in? Or more the reverse, that this will succeed but the WM service will have trouble reaching the site?

src/fides/api/alembic/migrations/versions/c961528edfc6_add_website_connection_type.py
adamsachs: you caught me! 😄

src/fides/api/service/connectors/website_connector.py
adamsachs: yeah it's a totally valid point, we don't _expect_ major differences afaik but it's certainly possible. that being said, i think _some_  validation that the website (URL) is valid is better than none, and this is the best thing we can do at the moment. any suggestions on how to clarify/improve that?

src/fides/api/models/detection_discovery.py
adamsachs: so my intention here was to clean up from something we missed when creating the `MonitorExecution` model class. instead of defining our own `id` field/column in the `MonitorExecution` class, we should have relied on the `id` field/column that's defined in the `FidesBase` ORM class that we inherit from. this provides some 'builtin' support for randomized ID generation, and promotes consistency.

so i'm trying to get us back to that better standard here. that being said, i need to be a bit careful i don't break anything on deployments that have already upgraded and have some `MonitorExecution` records with the 'old' `id` definition. i'll do some testing on that before we push this through...

but does that make sense?

tests/ctl/models/test_asset.py
```diff
@@ -0,0 +1,346 @@
...
+@pytest.fixture()
+async def javascript_asset_data(system_async) -> Dict[str, Any]:
+    return {
+        "name": "gtm.js",
+        "asset_type": "javascript",
+        "base_url": "https://www.googletagmanager.com/gtm.js",
+        "domain": "www.googletagmanager.com",
+        "system_id": system_async.id,
+        "locations": ["US", "Canada"],
+        "data_uses": ["analytics"],
```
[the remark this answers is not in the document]
adamsachs: i think we're good here, check out https://github.com/ethyca/fides/pull/5616/commits/9593bdd54de019275c3d02e5fed83a2b941da089

src/fides/api/models/asset.py
adamsachs: great to call this out, i think we're OK here based on https://github.com/ethyca/fides/pull/5616/commits/9593bdd54de019275c3d02e5fed83a2b941da089

src/fides/api/models/detection_discovery.py
thingscouldbeworse: Makes sense, I'll definitely be on the lookout for that in the future

src/fides/api/service/connectors/website_connector.py
thingscouldbeworse: I think this is fine as is, I take it the front end will call this and raising the error is exactly what we're expecting from there? I was just kind of spitballing

thingscouldbeworse (approved):

src/fides/api/service/connectors/website_connector.py
adamsachs: yup. it's a good thing to callout, we may need to iterate if/when this discrepancy comes up in the wild but i'm good with putting this in and letting that occur 👍

PR #5637, opened 2024-12-28
Updating static checks

galvana:
### Code Changes

* Breaks out the static checks into their own step so we can run always run them, even if there aren't any backend changes. This is done so we can prevent a branch from being merged if the static checks don't pass.
* Fixing silent error with Xenon

### Steps to Confirm

1.  CI tests should pass

### Pre-Merge Checklist

* [x] Issue requirements met
* Followup issues:
  * [x] No followup issues
* Database migrations:
  * [x] No migrations
* Documentation:

### DIFF
--- a/noxfiles/ci_nox.py
+++ b/noxfiles/ci_nox.py
@@ -102,10 +102,10 @@ def xenon(session: nox.Session) -> None:
         "--exclude=src/fides/_version.py",
     )
     session.run(*command, success_codes=[0, 1])
-    session.log(
+    session.warn(
         "Note: This command was malformed so it's been failing to report complexity issues."
     )
-    session.log(
+    session.warn(
         "Intentionally suppressing the error status code for now to slowly work through the issues."
     )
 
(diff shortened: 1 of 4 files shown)


src/fides/api/main.py
```diff
@@ -99,9 +99,7 @@ async def lifespan(wrapped_app: FastAPI) -> AsyncGenerator[None, None]:
 
     await send_analytics_event(
         AnalyticsEvent(
-            docker=in_docker_container(),
-            event=Event.server_start.value,
-            event_created_at=datetime.now(tz=timezone.utc),
+            docker=in_docker_container(), event=Event.server_start.value, event_created_at=datetime.now(tz=timezone.utc),
```
galvana, line 102: Intentionally mis-formatting this to trigger a CI failure

noxfiles/ci_nox.py
```diff
@@ -95,13 +95,19 @@ def xenon(session: nox.Session) -> None:
...
+        "--ignore=data,docs",
+        "--exclude=src/fides/_version.py",
+    )
+    session.run(*command, success_codes=[0, 1])
+    session.log(
+        "Note: This command was malformed so it's been failing to report complexity issues."
+    )
+    session.log(
+        "Intentionally suppressing the error status code for now to slowly work through the issues."
     )
```
galvana, line 110: This is the only thing we didn't discuss on Slack, let me know your thoughts on this one

PR #5647, opened 2025-01-06
Modify Logging Levels for Read-only actions

RobertKeyser:
### Description Of Changes

I don't find much value in logging messages like `Finding all datasets for connection '{}' with pagination params {}` in a production setting, when the HTTP logs cover almost the same information.

This PR converts all of those messages that are spawned from a read-type action (so not creates/updates/deletes) to a `debug` level. While they aren't hugely problematic, it will help our logs stay a bit cleaner.

IMO we're not really losing too much here, because when HTTP logs are logged with context, I don't need a second message telling me that it's doing what it should be doing. I did, however, keep those `finding xyz` logs on PUTs and DELETEs, etc to help show progress for those actions.

### Code Changes

* info -> debug for read actions

### Steps to Confirm

1.  launch Fides w info logging enabled
2. go to the request manager tab
3. check logs

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:

### DIFF
--- a/src/fides/api/api/v1/endpoints/dataset_endpoints.py
+++ b/src/fides/api/api/v1/endpoints/dataset_endpoints.py
@@ -98,7 +98,7 @@ router = APIRouter(tags=["Dataset Configs"], prefix=V1_URL_PREFIX)
 def _get_connection_config(
     connection_key: FidesKey, db: Session = Depends(deps.get_db)
 ) -> ConnectionConfig:
-    logger.info("Finding connection config with key '{}'", connection_key)
+    logger.debug("Finding connection config with key '{}'", connection_key)
     connection_config = ConnectionConfig.get_by(db, field="key", value=connection_key)
     if not connection_config:
         raise HTTPException(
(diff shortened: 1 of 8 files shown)


src/fides/api/api/v1/endpoints/messaging_endpoints.py
```diff
@@ -358,7 +358,7 @@ def put_config_secrets(
     """
     Add or update secrets for messaging config.
     """
-    logger.info("Finding messaging config with key '{}'", config_key)
+    logger.debug("Finding messaging config with key '{}'", config_key)
```
RobertKeyser, line 361:
```suggestion
    logger.info("Finding messaging config with key '{}'", config_key)
```

adamsachs (approved):
this looks good, thanks @RobertKeyser !

src/fides/api/api/v1/endpoints/dataset_endpoints.py
```diff
@@ -631,7 +631,7 @@ def delete_dataset(
 ) -> None:
     """Removes the DatasetConfig based on the given key."""
 
-    logger.info(
+    logger.debug(
```
RobertKeyser, line 634:
```suggestion
    logger.info(
```

PR #5673, opened 2025-01-15
Convert `rem` to base font size

gilluminate:
Closes [HJ-382]

### Description Of Changes

Provides a single css var for customers to use to override the font sizes.

This allows for customers to "opt in" to accessibility compliance by setting this as "1rem" while avoiding the footgun probability that most are hard coding an HTML font size anyway which has the potential to cause sizing issues with fides.js out of the box.

### Code Changes

* Adds a new `--fides-base-font-size`
* Converts all instances of the use of `rem` to calculate from the base size instead.

### Steps to Confirm

1. Visit the fidej.js component demo page (http://localhost:3001/fides-js-components-demo.html)
2. Font sizes should remain exactly how they have always been.

### Pre-Merge Checklist

* [x] Issue requirements met
* [x] `CHANGELOG.md` updated
  * [x] Add a https://github.com/ethyca/fides/labels/high-risk label to the entry if your change includes a high-risk change (i.e. potential for performance impact or unexpected regression) that should be flagged
* Followup issues:
  * [x] No followup issues
* Database migrations:
  * [x] No migrations
* Documentation:

[HJ-382]: https://ethyca.atlassian.net/browse/HJ-382?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/clients/fides-js/src/components/fides.css
+++ b/clients/fides-js/src/components/fides.css
@@ -56,11 +56,11 @@
 
   /* Everything else */
   --fides-overlay-font-family: Inter, sans-serif;
-  --8px: 0.5rem;
-  --12px: 0.75rem;
-  --14px: 0.875rem;
-  --15px: 0.9375rem;
-  --16px: 1rem;
+  --fides-base-font-size: 1rem;
+  --8px: calc(var(--fides-base-font-size) * 0.5);
+  --12px: calc(var(--fides-base-font-size) * 0.75);
+  --14px: calc(var(--fides-base-font-size) * 0.875);
+  --16px: calc(var(--fides-base-font-size) * 1);
   --fides-overlay-font-size-body-xs: var(--8px);
   --fides-overlay-font-size-body-small: var(--12px);
   --fides-overlay-font-size-body: var(--14px);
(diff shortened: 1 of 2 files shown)


clients/fides-js/src/components/fides.css
```diff
@@ -737,7 +735,7 @@ div#fides-overlay-wrapper .fides-toggle .fides-toggle-display {
   display: inline-block;
   min-width: 8px;
   height: 8px;
-  margin-right: calc(0.5rem + 2px);
+  margin-right: calc(calc(var(--fides-base-font-size) * 0.5) + 2px);
```
tvandort, line 738: This originally was not tied to the variable, do we want it to be?

tvandort (commented):
Good to approve but a quesiton first, I see there are a number of margins (etc) that seem to now be tied to variables that previously were not. 

Are we sure we want to tie those to these variables now? Especially if a client did decide to override the `--fides-base-font-size` with a rem unit?

clients/fides-js/src/components/fides.css
gilluminate: Good callout. Yeah, that makes sense to me. This just controls how much space is between the down arrow icon and the text so the space there is uniform with the text spacing.

tvandort (approved):

PR #5698, opened 2025-01-21
ensure `decode_password` function properly handles plaintext but valid base64 passwords 

adamsachs:
Closes [HJ-397]

### Description Of Changes

Handles the `UnicodeDecodeError` that is raised by `decode_password` if it is given a plaintext password that's also valid base64. The exception is caught, and the input string is returned as it was provided, under the assumption that it is a plaintext (and _not_ base64-encoded) password.

See issue description for more details. This impacts some user-related endpoints (e.g. reset password, accept invite) if the user provides a password that also happens to be valid base64. Notably, the utilities also _ignore_ invalid base64 characters like `_`, so a value like  e.g. `Test_1234` is treated as valid base64, since the `_` is ignored and its length of `8` is a multiple of 4.

### Code Changes

* catch and ignore `UnicodeDecodeError` in `decode_password` utility function, assuming input string is a plaintext password

### Steps to Confirm

- [x] confirmed locally that I can (still) create a user and manually set their password to `Test_1234` (i think this was working before, because we send a base64 encoded password in the request from the UI in this workflow)
- [x] confirmed i can now accept a user invite and set my password to `Test_1234`, which i also confirmed was broken before: 
image
 

### Pre-Merge Checklist

* [x] Issue requirements met
* [x] `CHANGELOG.md` updated
* Followup issues:
  * [x] No followup issues
* Database migrations:
  * [x] No migrations
* Documentation:

[HJ-397]: https://ethyca.atlassian.net/browse/HJ-397?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/fides/api/cryptography/cryptographic_util.py
+++ b/src/fides/api/cryptography/cryptographic_util.py
@@ -13,7 +13,7 @@ def decode_password(password: str) -> str:
     """
     try:
         return b64_str_to_str(password)
-    except Error:
+    except (Error, UnicodeDecodeError):
         return password
 
 
--- a/tests/lib/test_cryptography_util.py
+++ b/tests/lib/test_cryptography_util.py
@@ -104,7 +104,12 @@ def test_str_to_b64_str() -> None:
     "password, expected",
     [
         ("Testpassword1!", "Testpassword1!"),
+        (
+            "Test_1234",
+            "Test_1234",
+        ),  # this is actually valid base64 (but should be treated as plaintext), so this represents an edge case
         (str_to_b64_str("Testpassword1!"), "Testpassword1!"),
+        (str_to_b64_str("Test_1234"), "Test_1234"),
     ],
 )
 def test_decode_password(password, expected):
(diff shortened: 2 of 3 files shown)


tests/lib/test_cryptography_util.py
```diff
@@ -104,7 +104,12 @@ def test_str_to_b64_str() -> None:
     "password, expected",
     [
         ("Testpassword1!", "Testpassword1!"),
+        (
```
adamsachs, line 107: this new test was failing before the fix

src/fides/api/cryptography/cryptographic_util.py
```diff
@@ -13,7 +13,7 @@ def decode_password(password: str) -> str:
     """
     try:
         return b64_str_to_str(password)
-    except Error:
+    except (Error, UnicodeDecodeError):
```
adamsachs, line 16: decided to keep this minimal in scope - only catching the error i know can be thrown in the error case. handling more generic errors/exceptions would be more likely to lead to unintended side effects, IMO.

also didn't add any sort of logging here as it could get very noisy, but i'd be open to an argument for that

adamsachs (commented):
some quick notes -

Kelsey-Ethyca (approved):
LGTM

adamsachs:
failing tests look flaky to me, merging 👍

PR #5668, opened 2025-01-14
LA-226 Updates UI to use the new Ethyca brand colors

lucanovera:
### Description Of Changes

Update Chakra theme and components to use new Ethyca brand colors and standardize the looks of some components like Badges.

### Code Changes

* Update Chakra theme primary, secondary, complementary palettes
* Added new colors to Chakra theme
* Standardized the look of badges
* Updated admin homepage to use Cards

### Steps to Confirm

Admin-UI
1.  Browse around admin-ui pages
2. Check no purple or deep blue colors are present, they should be replaced with minos black
3. Check that the Fides logo is updated to it's uppercase version
4. Check that the homepage doesn't show an image, the links should now be gray cards with a left border of different colors

Privacy Center
1. Open privacy center homepage
2. Check icons are now Minos black instead of purple
3. Go to manage consent
4. Check the toggles are Minos black instead of purple

Fides-JS
1. Make sure you have a valid US Experience to test the fides-js modal:
2. Run the fides-js build, go to `clients/fides-js` and run `npm run build`
5. Go to http://localhost:3001/fides-js-demo.html?geolocation=us-ca&fides_locale=us
6. Click on Manage preferences
7. Check that the the modal now displays the button and the toggles in Minos black color instead of purple

### Screenshots
Admin-UI
Screen Shot 2025-01-14 at 12 04 12

Privacy Center
Screen Shot 2025-01-14 at 12 05 17
Screen Shot 2025-01-14 at 12 44 00

Fides JS Banner, Modal, TCF
Screen Shot 2025-01-14 at 12 19 01
Screen Shot 2025-01-14 at 12 23 57
Screen Shot 2025-01-14 at 12 24 02
Screen Shot 2025-01-14 at 12 25 08

### Pre-Merge Checklist

* [x] Issue requirements met
* [ ] All CI pipelines succeeded
* [x] `CHANGELOG.md` updated
  * [ ] Add a https://github.com/ethyca/fides/labels/db-migration label to the entry if your change includes a DB migration
  * [ ] Add a https://github.com/ethyca/fides/labels/high-risk label to the entry if your change includes a high-risk change (i.e. potential for performance impact or unexpected regression) that should be flagged
* Followup issues:
  * [ ] Followup issues created (include link)
  * [x] No followup issues
* Database migrations:
  * [ ] Ensure that your downrev is up to date with the latest revision on `main`
  * [ ] Ensure that your `downgrade()` migration is correct and works
[...]

PR #5702, opened 2025-01-27
Fix ClassificationCategoryBadge icon spacing

gilluminate:
### Description Of Changes

Changes made along with the color updates (https://github.com/ethyca/fides/pull/5668) caused an issue with icon spacing in these badges. This fix addresses that and restores the badge layout.

### Code Changes

* Restores flex properties on the badge

Before:
image

After:
image


lucanovera (approved):
Good catch! Thanks for the quick fix. Approved!

PR #5704, opened 2025-01-28
HJ-92: add table for storing monitor execution records

thingscouldbeworse:
Prereq for HJ-92

### Code Changes

* Migration to add table that can store `MonitorExecution` objects during Detection

### Steps to Confirm

1.  run migrations, check if table exists

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:


src/fides/api/models/detection_discovery.py
```diff
@@ -348,6 +338,41 @@ def mark_as_addition(
...
+    monitor_config_key = Column(
+        String,
+        ForeignKey(MonitorConfig.key),
+        nullable=False,
+        index=True,
+    )
+    status = Column(String, nullable=True)
+    started = Column(
+        DateTime(timezone=True), nullable=True, default=datetime.now(timezone.utc)
+    )
```
adamsachs, line 359: nit: we _could_ just use `created_at` as our de-facto "started" timestamp i think? we do that on some other models, and reduces some redundancy. but it's OK if we wanna keep it as a separate property/attribute

src/fides/api/models/detection_discovery.py
```diff
@@ -348,6 +338,41 @@ def mark_as_addition(
...
+        unique=False,
+        nullable=False,
+        default=list,
+    )
+    created_at = Column(
+        DateTime(timezone=True), nullable=False, default=datetime.now(timezone.utc)
+    )
+    updated_at = Column(
+        DateTime(timezone=True), nullable=False, default=datetime.now(timezone.utc)
+    )
```
adamsachs, line 373: pretty sure these are inferred from the `Base` model class, a bit cleaner to not re-define here unless it's needed for some reason?

src/fides/api/alembic/migrations/versions/ed96417b07d8_create_table_for_persisting_.py
```diff
@@ -0,0 +1,48 @@
...
+down_revision = "d9237a0c0d5a"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.create_table(
+        "monitorexecution",
+        sa.Column("id", sa.String(), primary_key=True),
+        sa.Column("monitor_config_key", sa.String(), nullable=False),
```
adamsachs, line 25: i think it'd be worth indexing on this column, since we're going to be looking up against it quite a bit

CHANGELOG.md
```diff
@@ -33,6 +33,7 @@ Changes can also be flagged with a GitHub label for tracking purposes. The URL o
 
 ### Added
 - Migration to add the `data_uses` column to `stagedresource` table, prereqs for Data Catalog work in Fidesplus [#5600](https://github.com/ethyca/fides/pull/5600/) https://github.com/ethyca/fides/labels/db-migration
+- Migration to add the `monitorexecution` table used by fidesplus to persist `MonitorExecution` records to DB [#5704](https://github.com/ethyca/fides/pull/5704) https://github.com/ethyca/fides/labels/db-
```
adamsachs, line 36: looks like the label URL got clipped somehow 
```suggestion
- Migration to add the `monitorexecution` table used by fidesplus to persist `MonitorExecution` records to DB [#5704](https://github.com/ethyca/fides/pull/5704) https://github.com/ethyca/fides/labels/db-migration
```

adamsachs (approved):
generally looks good, but a few minor comments that are worth addressing i think!

CHANGELOG.md
thingscouldbeworse: oop, fixed here

src/fides/api/models/detection_discovery.py
thingscouldbeworse: Swapped to how we did it in the most recent migration here: https://github.com/ethyca/fides/pull/5704/commits/af165b5cd6cdc558506a12a03d4081b274980614

src/fides/api/models/detection_discovery.py
thingscouldbeworse: We could, I left it this way just to make the serialization to the pydantic model easier but I can make it mapped manually if we prefer to just rely on one column. Or I could remove `created_at`?

adamsachs: don't think we should remove `created_at` because it's basically on all of our tables. the pydantic model serialization makes sense, though another option for that is to provide an `alias` on the pydantic field so that e.g. in the API response it come through as `started` even though underneath it's being referred to as `created_at`.

that being said, i think it's fine to leave as-is 👍

adamsachs (approved):
just re-approving after last round of updates, things look good here. i'd go ahead and merge this into `main` and push a beta tag so the f+ PR can reference a beta tag and not a commit hash, to get that one ready for merge.

PR #5752, opened 2025-02-11
Update consent-banner-tcf-custom-notices.cy.ts

gilluminate:
### Description Of Changes

Recently fixed this flaky test in the `consent-banner-tcf.cy.ts` tests but it made its way to this custom notices test file as well. It really isn't relevant at all to the custom notices to I'm just removing it to fix the reintroduction of flakiness.


thingscouldbeworse (approved):
🩸

PR #5753, opened 2025-02-12
Update incorrect APIRouter import

galvana:
Closes [LJ-373](https://ethyca.atlassian.net/browse/LJ-373)

### Description Of Changes

Fides uses a custom `APIRouter` to create two versions of each endpoint, a version with and without slashes (`/dataset` and `/dataset/`). This allows either version of the URL to be routed to the same endpoint logic. However, the routes in `generic_overrides.py` use the default `APIRouter` from FastAPI. This means that we only create the non-slash versions of the endpoints in `generic_overrides.py` and when we override the base routes from `router_factory.py`, we only replace the non-slash versions, not the slash versions.
```
`/dataset` -> uses the new implementation from `generic_overrides.py`
`/dataset/` -> uses the default implementation from `router_factory.py`
```
The result was that the Fides tests were still hitting the endpoints in `router_factory.py` but not the endpoints with extra validation in `generic_overrides.py`. Now that we fixed the `APIRouter` we need to update the logic and tests to work with the updated logic from the `generic_overrides.py`

**TLDR: It's possible to invoke taxonomy endpoints that don't use extra validations we've added in the last few months, leading to the creation of invalid data.**

### Code Changes

* Misc fixes to get the existing tests to pass now that we're pointing to the correct, validated endpoints.

### Steps to Confirm

1.  Tests should pass

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:

[LJ-373]: https://ethyca.atlassian.net/browse/LJ-373?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/fides/api/schemas/taxonomy_extensions.py
```diff
@@ -31,7 +31,7 @@ class DataSubject(BaseDataSubject):
 
 class TaxonomyCreateOrUpdateBase(DefaultModel, BaseModel):
     name: Optional[str] = None
-    description: Optional[str]
+    description: Optional[str] = None
```
galvana, line 34: We were missing the `None` so `description` was required and was causing some fixtures to fail

tests/ctl/core/test_api_helpers.py
```diff
@@ -32,9 +32,9 @@ def created_resources(
     resource_type = request.param
     for _ in range(RESOURCE_CREATION_COUNT):
         base_resource = resources_dict[resource_type].model_copy()
-        base_resource.fides_key = "{}_{}".format(
-            base_resource.fides_key, str(uuid.uuid4())[:6]
-        )
+        uuid_suffix = str(uuid.uuid4())[:6]
+        base_resource.fides_key = "{}_{}".format(base_resource.fides_key, uuid_suffix)
+        base_resource.name = "{} {}".format(base_resource.name, uuid_suffix)
```
galvana, line 37: We check name now for uniqueness, so I just added the UUID to the end of the name

src/fides/api/api/v1/endpoints/generic_overrides.py
```diff
@@ -28,15 +29,16 @@
     DataUse,
     DataUseCreateOrUpdate,
 )
+from fides.api.util.api_router import APIRouter
```
galvana, line 32: Key change - this is the fix

adamsachs (approved):
looks good on a quick glance, thanks for the catch!

PR #5766, opened 2025-02-15
Dataset test UI improvements

galvana:
Closes [LJ-348](https://ethyca.atlassian.net/browse/LJ-348)

### Description Of Changes

This PR adds a `Test logs` section to the dataset test UI
test-logs

### Code Changes

* Adds a new log destination (a "sink") that filters for logs that belong to test privacy requests. Each log entry for a given privacy request is stored in a list under the Redis key `log_{privacy_request_id}`
* Adds a new `GET /privacy_request/{id}/logs` endpoint to retrieve the logs
* New test logs component for the front-end

### Steps to Confirm

1. Set `FIDES__SECURITY__DSR_TESTING_TOOLS_ENABLED=true` in your `.env` file
2. Start Fides by running `nox -s dev -- postgres` (you might need to run `nox -s build` if you get an error)
3. Create a new system called `Postgres` and add a Postgres integration to the `postgres_example` database
```
Host: host.docker.internal
Port: 6432
Username: postgres
Password: postgres
Database: postgres_example
```
4. Create and attach the [postgres_example_test_dataset.txt](https://github.com/user-attachments/files/18895454/postgres_example_test_dataset.txt) dataset
5. Navigate to http://localhost:3000/systems/configure/postgres/test-datasets. There is a button from the integration config if you're running plus, but this is so we can test in fides
6. Select the default policy, and set the email to `jane@example.com`
7. Click on run, the request should succeed and the logs should be shown at the bottom

### Pre-Merge Checklist
* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:

[LJ-348]: https://ethyca.atlassian.net/browse/LJ-348?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


clients/admin-ui/src/features/test-datasets/TestLogsSection.tsx
galvana, line 1: New component with color-coded logs and auto-scrolling as new logs are available

src/fides/api/api/v1/endpoints/drp_endpoints.py
```diff
@@ -19,7 +19,7 @@
     get_privacy_request_or_error,
 )
 from fides.api.models.policy import Policy
-from fides.api.models.privacy_request import PrivacyRequest, PrivacyRequestStatus
+from fides.api.models.privacy_request import PrivacyRequest
```
galvana, line 22: You'll notice a lot of these files just update the imports. I was running into cyclic dependency issues with some Pydantic schemas that were being defined under the `models` directory. This is kind of a noisy change but I figured it was the right thing to do in the long term, to move these schemas under the `schema` directory instead

src/fides/api/api/v1/endpoints/privacy_request_endpoints.py
```diff
@@ -2179,15 +2180,21 @@ def get_test_privacy_request_results(
...
-        "results": (
-            filtered_results
-            if CONFIG.security.dsr_testing_tools_enabled
-            else "DSR testing tools are not enabled, results will not be shown."
-        ),
-    }
+    with logger.contextualize(
+        privacy_request_id=privacy_request.id,
+        privacy_request_source=privacy_request.source.value,
+    ):
```
galvana, line 2186: This is another key change. There are instances where we can't use the `@log_context` decorator because we're not passing in any params of type `PrivacyRequest` that we can automatically extract the ID and source from. In those cases we use a manual `with logger.contextualize` context

galvana (commented):
Here are some self-review comments, let me know if you have any questions!

JadeCara (approved):
Legit!

PR #5776, opened 2025-02-18
HA-399: migrate existing `Cookies` resources to `Asset` resources of type `Cookie`

thingscouldbeworse:
Closes #HA-399

### Description Of Changes

migrate cookies to asset

### Code Changes

* data migration to turn every entry in the `cookies` table into an entry in `asset` with `asset_type == 'Cookie'`
    - pull the privacy declaration attached to each cookie if it exists and combine the `data_use` value there into the array `data_uses` on `asset`
 * rip out where we previously referenced a `cookies` property on `System` and `PrivacyDeclaration` and prevent creating cookies when upserting these resources
 * retain a privacy notice's ability to have a `cookies` property which collects assets of type `Cookie` with a permissive matching data use

### Steps to Confirm

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:


src/fides/api/alembic/migrations/versions/c9c72b3d550b_data_migration_converting_existing_.py
```diff
@@ -0,0 +1,78 @@
...
+                "INSERT INTO asset (id, created_at, updated_at, name, domain, system_id, locations, data_uses, asset_type, with_consent) "
+                "VALUES (:id, :created_at, :updated_at, :name, :domain, :system_id, :locations, :data_uses, :asset_type, false)"
+            ),
+            id=str(uuid.uuid4()),
+            created_at=row.created_at,
+            updated_at=row.updated_at,
+            name=row.name,
+            domain=row.domain,
+            system_id=row.system_id,
+            locations=[row.path],
```
adamsachs, line 50: mm i think this is incorrect, `locations` are going to be geographic locations from which the asset was detected. `path` _may_ refer to `parent` on the new `Asset` model, but we should double check that.

i think we'll just need to leave `locations` as an empty list for the migrated cookies 👍

src/fides/api/alembic/migrations/versions/c9c72b3d550b_data_migration_converting_existing_.py
```diff
@@ -0,0 +1,78 @@
...
+                "INSERT INTO cookies (id, created_at, updated_at, name, domain, system_id, path, privacy_declaration_id) "
+                "VALUES (:id, :created_at, :updated_at, :name, :domain, :system_id, :path, :privacy_declaration_id)"
+            ),
+            id=str(uuid.uuid4()),
+            created_at=row.created_at,
+            updated_at=row.updated_at,
+            name=row.name,
+            domain=row.domain,
+            system_id=row.system_id,
+            path=row.locations[0],
```
adamsachs, line 76: same here but on the downgrade, just marking so that we don't accidentally forget to update the downgrade

src/fides/api/models/privacy_notice.py
```diff
@@ -191,19 +192,16 @@ def default_preference(self) -> UserConsentPreference:
...
-                PrivacyDeclaration,
-                PrivacyDeclaration.id == Cookies.privacy_declaration_id,
-            )
+            db.query(Asset)
             .filter(
+                Asset.type == "Cookie",
                 or_(
                     *[
-                        PrivacyDeclaration.data_use.like(f"{notice_use}%")
+                        Asset.data_use.like(f"{notice_use}%")
```
adamsachs, line 204: pretty sure the expression here will need to at least be tweaked, right? for one, the column is now `Asset.data_uses`, and beyond that, it's now an array of strings, so i think the syntax/comparison will need to adjust accordingly?

a bit more substantively, here's where we need to be careful about performance impact. previously, the `data_use` column of `PrivacyDeclaration` was indexed, so i _think_ this lookup avoided a full table scan? (although i'm questioning that now since it's not an exact string match....) if we need to get an indexed lookup for our `Asset`s, then we may need to look into creating an auxiliary table that stores the `data_use` strings associated with an `Asset`, so that it can be indexed and looked up more efficiently. unless there's some way to index the elements of arrays in an array column? this is a postgres optimization question we may want to dig into!

adamsachs (commented):
@thingscouldbeworse looks like you're on the right track here, but there's some nuance/detail that needs to be worked through on the `cookies` property method 👍

src/fides/api/alembic/migrations/versions/c9c72b3d550b_data_migration_converting_existing_.py
thingscouldbeworse: changed here https://github.com/ethyca/fides/pull/5776/commits/157b5355ae32cdbf6bc3ff0780036bbc779da63f#diff-9a71bb9fcb8781cd9b038e712d899749d3e99ec8344ce94cfa617e6dbb0785a6L41

src/fides/api/alembic/migrations/versions/c9c72b3d550b_data_migration_converting_existing_.py
thingscouldbeworse: 👍 https://github.com/ethyca/fides/pull/5776/commits/157b5355ae32cdbf6bc3ff0780036bbc779da63f#diff-9a71bb9fcb8781cd9b038e712d899749d3e99ec8344ce94cfa617e6dbb0785a6L76

src/fides/api/models/privacy_notice.py
thingscouldbeworse: Yeah that was almost kind of pseudocode on my part, cleaned up here https://github.com/ethyca/fides/pull/5776/commits/157b5355ae32cdbf6bc3ff0780036bbc779da63f#diff-163f9e368e31215ee24ddaed03376f07313bea4488f74387edd0393bd958c334R204

src/fides/api/alembic/migrations/versions/c9c72b3d550b_data_migration_converting_existing_.py
```diff
@@ -0,0 +1,214 @@
...
+depends_on = None
+
+
+def upgrade():
+    try:
+        op.add_column("asset", sa.Column("description", sa.String(), nullable=True))
+    except psycopg2.errors.DuplicateColumn:
+        logger.warning(
+            "Column 'description' already exists in 'asset' table. Skipping column addition."
+        )
```
adamsachs, line 32: hm, seems weird we'd need to handle this? i suppose there's no harm in being extra safe

src/fides/api/alembic/migrations/versions/c9c72b3d550b_data_migration_converting_existing_.py
```diff
@@ -0,0 +1,214 @@
...
+                data_uses=[],
+                asset_type="Cookie",
+            )
+            continue
+        pud_system_id_data_use = connection.execute(
+            sa.text(
+                "SELECT system_id, data_use FROM privacydeclaration WHERE id = :privacy_declaration_id"
+            ),
+            privacy_declaration_id=row.privacy_declaration_id,
+        ).fetchone()
```
adamsachs, line 62: ideally we'd pull this data in with our original cookie query in an outer join on the `privacy_declaration_id`? as it stands now, i think this presents an an n+1 query pattern. the performance implications for a migration are obviously not as important as for normal app runtime, but in the past we've had migrations time out, and it's not inconceivable that someone could have tens of thousands of cookies on their server (which would result in tens of thousands of queries being executed here). so i do think we should try to optimize this

src/fides/api/alembic/migrations/versions/c9c72b3d550b_data_migration_converting_existing_.py
```diff
@@ -0,0 +1,214 @@
...
+            ),
+            id=str(uuid.uuid4()),
+            created_at=asset.created_at,
+            updated_at=asset.updated_at,
+            name=asset.name,
+            domain=asset.domain,
+            system_id=asset.system_id,
+            data_uses=asset.data_uses,
+            asset_type="Cookie",
+        )
```
adamsachs, line 111: again, this'll issue `n` queries for `n` assets, right? the number of assets we're creating here could be quite large, i think we should really try to get this done in a single bulk query 👍

src/fides/api/alembic/migrations/versions/c9c72b3d550b_data_migration_converting_existing_.py
```diff
@@ -0,0 +1,214 @@
...
+        # Find the privacy declaration ID matching the system ID and data use
+        privacy_declaration_ids = []
+        for data_use in row.data_uses:
+            pud = connection.execute(
+                sa.text(
+                    "SELECT id FROM privacydeclaration WHERE system_id = :system_id AND data_use = :data_use"
+                ),
+                system_id=row.system_id,
+                data_use=data_use,
+            ).fetchone()
```
adamsachs, line 183: ok so here it may not be quite as simple to 'join' with the privacy declaration table. but that doesn't mean we should issue a query for every asset! that can be problematic. 

we can just query the full privacy declaration table up front, store that in memory, and do the "joining" logic (which is admittedly more complex here, since we don't have a direct FK link) using python. the key thing i'd want to avoid though is issuing a query for every asset that we're migrating - there could be thousands.

src/fides/api/alembic/migrations/versions/c9c72b3d550b_data_migration_converting_existing_.py
```diff
@@ -0,0 +1,214 @@
...
+            connection.execute(
+                sa.text(
+                    "INSERT INTO cookies (id, created_at, updated_at, name, domain, privacy_declaration_id) "
+                    "VALUES (:id, NOW(), NOW(), :name, :domain, :privacy_declaration_id)"
+                    "ON CONFLICT DO NOTHING"
+                ),
+                id=cookie_id,
+                name=row.name,
+                domain=row.domain,
+                privacy_declaration_id=privacy_declaration_id,
```
adamsachs, line 213: again, just want to make sure we do a bulk add here rather than one query per asset. (i know i'm a broken record, but erring on the side of thoroughness even at the expense of being repetitive 😅 )

adamsachs (commented):
this is looking solid @thingscouldbeworse, but two things i think we'll need to address further:
1. we need to be careful about our query patterns in the data migration. as it stands now, i'm pretty sure on both upgrade and downgrade we're going to be executing `n` queries for `n` cookie records on Fides. in a worst case scenario, there could be thousands of cookie records on Fides - it's not likely, but there's at least one GVL vendor who declares thousands of cookies (yuck). in any case, it shouldn't be hard for us to gather the data we need "up front" (either via a JOIN in the query, or via a query up front to pull all the privacy declaration data into memory) rather than execute `n` queries, so i'd encourage us to do that
2. it'd be good to verify how this interacts with the `fideslang` changes you've got queued up, as mentioned on that PR 👍

src/fides/api/alembic/migrations/versions/c9c72b3d550b_data_migration_converting_existing_.py
thingscouldbeworse: Addressed here: https://github.com/ethyca/fides/pull/5776/commits/54e4eac935d25653649ec9f2424b76047206c7bb

src/fides/api/alembic/migrations/versions/c9c72b3d550b_data_migration_converting_existing_.py
thingscouldbeworse: Good point, done here: https://github.com/ethyca/fides/pull/5776/commits/54e4eac935d25653649ec9f2424b76047206c7bb#diff-9a71bb9fcb8781cd9b038e712d899749d3e99ec8344ce94cfa617e6dbb0785a6R165

src/fides/api/alembic/migrations/versions/c9c72b3d550b_data_migration_converting_existing_.py
thingscouldbeworse: Mostly a "just in case". I ran into it once or twice migrating down/up/down/up and I figured it wouldn't hurt

src/fides/api/alembic/migrations/versions/c9c72b3d550b_data_migration_converting_existing_.py
thingscouldbeworse: bulk insert here: https://github.com/ethyca/fides/pull/5776/files#diff-9a71bb9fcb8781cd9b038e712d899749d3e99ec8344ce94cfa617e6dbb0785a6R101

src/fides/api/alembic/migrations/versions/c9c72b3d550b_data_migration_converting_existing_.py
thingscouldbeworse: removed the n+1 queries, bulk fetch, bulk insert for upgrade/downgrade

adamsachs (commented):
@thingscouldbeworse nice work on the migration updates, that's looking solid now (just wanna confirm you've been able to manually test that since we don't have any automated test coverage [not that we need it] 👍 )

then just a couple of other small cleanup items to work through that i've commented on. i'd like to find some time tomorrow to play around a bit with the API behavior myself if an API client specifies `cookies`, just to see what happens - but that shouldn't be a blocker.

oh, and looks like you need to bump the migration revision head due to other PRs with migrations!

src/fides/api/models/privacy_notice.py
```diff
@@ -191,24 +190,21 @@ def default_preference(self) -> UserConsentPreference:
...
+        or_queries = [
+            f"array_to_string(data_uses, ',') ILIKE '%{data_use}%'"
+            for data_use in self.data_uses
+        ]
+
+        query = db.query(Asset).filter(
+            Asset.asset_type == "Cookie",
+            or_(*[text(query) for query in or_queries]),
+        )
+        print(f"Query: {query}")
```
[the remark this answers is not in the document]
thingscouldbeworse: removed

adamsachs (commented):
@thingscouldbeworse some minor comments on a fresh pass through -

CHANGELOG.md
```diff
@@ -208,6 +208,12 @@ Changes can also be flagged with a GitHub label for tracking purposes. The URL o
 - Modified Dependabot configuration to support monorepo security updates [#5810](https://github.com/ethyca/fides/pull/5810)
 - Fix load_samples to correctly collect & load sample connections with "False" secret values [#5828](https://github.com/ethyca/fides/pull/5828)
 
+### Added
+- Data migration to convert existing `Cookies` resources to `Asset` resources of type `Cookie` [#5776](https://github.com/ethyca/fides/pull/5776) https://github.com/ethyca/fides/labels/db-migration https://github.com/ethyca/fides/labels/high-risk
+
+### Added
+- Data migration to convert existing `Cookies` resources to `Asset` resources of type `Cookie` [#5776](https://github.com/ethyca/fides/pull/5776) https://github.com/ethyca/fides/labels/db-migration https://github.com/ethyca/fides/labels/high-risk
```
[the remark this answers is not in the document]
thingscouldbeworse: Fixed here: https://github.com/ethyca/fides/pull/5776/commits/ec3ff019413fe1984d7aabe35e01a37abd1e750f#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR27

tests/ctl/core/test_system.py
```diff
@@ -20,6 +18,70 @@
...
+
+    # Verify the error message contains the expected text
+    assert "Error adding privacy declarations" in str(exc_info.value)
+    assert "reverting system creation" in str(exc_info.value)
+
+
+@pytest.mark.skip(
+    reason="This test blows up CI for some reason and is only needed for codecov on exception catching"
+)
+async def test_upsert_system_reverts_creation_on_privacy_declaration_error(
```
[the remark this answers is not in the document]
thingscouldbeworse: removed here: https://github.com/ethyca/fides/pull/5776/commits/ec3ff019413fe1984d7aabe35e01a37abd1e750f#diff-93b4b681c875473b0b3b58dc05b9391f71f4067f715240ba830c218e9080510cL28

tests/ops/models/test_privacy_notice.py
```diff
@@ -582,69 +582,6 @@ def test_dry_update_copies_cleanly(
...
-            (
-                ["marketing.advertising", "third_party_sharing"],
-                [{"name": "test_cookie"}, {"name": "another_cookie"}],
-                [CookieSchema(name="test_cookie"), CookieSchema(name="another_cookie")],
-                "Test multiple cookies",
-            ),
-            (["marketing.advertising"], [], [], "No cookies returns an empty set"),
-        ],
-    )
-    def test_relevant_cookies(
```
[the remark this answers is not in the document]
thingscouldbeworse: Revived here: https://github.com/ethyca/fides/pull/5776/commits/ec3ff019413fe1984d7aabe35e01a37abd1e750f#diff-5b0b0cbfdfd131da4ddff237a67554dceac652be2dcdc4c3c50c93095f68bc9fR585

clients/admin-ui/src/features/datamap/reporting/DatamapReportTableColumns.tsx
```diff
@@ -451,21 +451,6 @@ export const getDatamapReportColumns = ({
...
-          suffix="cookies"
-          value={props.getValue()}
-          {...props}
-        />
-      ),
-      meta: {
-        showHeaderMenu: !isRenaming,
-        width: "auto",
-      },
-    }),
```
adamsachs, line 468: we shouldn't be removing this @jpople !

adamsachs (approved):
OK, i think this is generally looking good! will give a tentative approval since i'm going OOO and don't want to hold this up, though there's at least a couple of things we need to adjust:
- [ ] @jpople  i think we accidentally removed the `cookies` column from the datamap report view, we want to keep that there! can you ensure we add that back in sensibly? probably a simple update but i want to let you do that
- [ ] updating the fideslang dep 👍 


more substantively: i ran through a migration test of existing cookies just now and the resulting asset table looked good from a spot check, but i would also encourage you @thingscouldbeworse and any ohters to run through a few manual tests with migrating locally - it's the aspect of this body of work that is least reversible and hardest to test (repeatedly) on deployed environment, so it carries the most inherent risk.

the other area is the still-unexplained CI adjustments that were needed. i don't think that's a blocker, but if you've got time to dig on it more, i'd do so, and i'd also just keep our eye out for any otherwise unexplained differences in behavior in `main` that may arise after we do merge this in...

tests/ctl/api/test_seed.py
```diff
@@ -1,7 +1,7 @@
 import io
 import os
 from textwrap import dedent
-from typing import Generator
+from typing import Generator, List
```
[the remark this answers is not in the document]
thingscouldbeworse: removed in https://github.com/ethyca/fides/pull/5776/commits/f66d5e8f78c20ddceb637ab4dc0605b5eb8eedf2

PR #5786, opened 2025-02-19
HA-423 - Fix Compass returning error for deleted GVL systems

andres-torres-marroquin:
Closes [HA-423]

### Description Of Changes

Fix Compass returning error for deleted GVL systems

### Code Changes

* modified `create_system` in `src/fides/api/db/system.py` so `model_dump` doesn't use `mode=json`
* modified `create_resource` in `src/fides/api/db/crud.py` so a `SQLAlchemyError` exception logs.

### Steps to Confirm

1.  Try adding a compass system that has a `vendor_deleted_date` from the "add a system" form, eg AB InBev, more examples are on the jira ticket.

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:

[HA-423]: https://ethyca.atlassian.net/browse/HA-423?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ Changes can also be flagged with a GitHub label for tracking purposes. The URL o
 ### Added
 - Added `reject_all_mechanism` to `PrivacyExperienceConfig` [#5952](https://github.com/ethyca/fides/pull/5952) https://github.com/ethyca/fides/labels/db-migration
 - Added DataHub dataset sync functionality UI with feedback and error handling [#5949](https://github.com/ethyca/fides/pull/5949)
+- Added support for TCF preview in Admin UI experience form [#5962](https://github.com/ethyca/fides/pull/5962)
 - Added `opt_in_only` to `Layer1ButtonOption` [#5958](https://github.com/ethyca/fides/pull/5958)
 - Added support for links in `<a>` tags on the custom HTML description [#5960](https://github.com/ethyca/fides/pull/5960)
 - Added "Reject all" behavior and visibility options to TCF Experience config form [#5964](https://github.com/ethyca/fides/pull/5964)
--- a/clients/admin-ui/cypress/e2e/privacy-experiences.cy.ts
+++ b/clients/admin-ui/cypress/e2e/privacy-experiences.cy.ts
@@ -266,7 +266,7 @@ describe("Privacy experiences", () => {
         cy.get("#preview-container")
           .find("#fides-banner")
           .find("#fides-banner-notices")
-          .contains("Essential");
+          .contains("Data Sales and Sharing");
       });
 
       it("does not show option to display privacy notices in modal preview when clicked", () => {
@@ -322,7 +322,7 @@ describe("Privacy experiences", () => {
         );
       });
 
-      it("doesn't show a preview while editing TCF experience", () => {
+      it("shows a preview while editing TCF experience", () => {
         cy.fixture("privacy-experiences/experienceConfig.json").then((data) => {
           cy.intercept("GET", "/api/v1/experience-config/pri*", {
             ...data,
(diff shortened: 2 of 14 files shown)


src/fides/api/db/system.py
```diff
@@ -406,9 +406,7 @@ async def create_system(
 
     # create the system resource using generic creation
     # the system must be created before the privacy declarations so that it can be referenced
-    resource_dict = resource.model_dump(
-        mode="json"
-    )  # mode=json helps Url fields be converted to strings before saving to db
+    resource_dict = resource.model_dump()
```
adamsachs, line 409: so did we validate that a `resource` here (i.e., a `System`) with an `AnyUrl` field that's populated actually gets created successfully? just trying to make sense of the comment that was clearly left here.

i'm also seeing that `update_system` in this module does a similar operation - 

```
    existing_system_dict = copy.deepcopy(
        SystemSchema.model_validate(system)
    ).model_dump(mode="json")
```
is that not going to cause a similar problem? do we need to adjust that?

adamsachs (changes requested):
thanks for finding the issue @andres-torres-marroquin! 

this may be an OK path to go down, but i've got some questions that we should clarify. generally, we need to add test coverage here - tests that surface the bug we're fixing, and i think also tests that prove we haven't caused a new regression with `Url` fields, as the comment suggests we may be.

relatedly, we need to make sure (ideally with automated tests) that `update_system` isn't being left with a similar bug to what you're fixing here...

src/fides/api/db/system.py
adamsachs: @andres-torres-marroquin two things:
1. what i was saying is that we should validate (ideally with automated test coverage) that with this change we can create a system with an `AnyUrl` field. i know that your change fixes the bug with the `vendor_deleted_at` field, but the comment around why we use `mode="json"` in the first place indicates that it was there for a reason, i.e. to prevent issues around `AnyUrl` fields - so what we should do here is make sure that by fixing the bug with the `vendor_deleted_at` (datetime) field, we haven't introduced a bug related to an `AnyUrl` field on `System
2. OK, i see that the output from https://github.com/ethyca/fides/blob/main/src/fides/api/db/system.py#L274C5-L274C25 isn't actually used to update the resource - but it _is_ used for updating the audit log. so it's fine if we want to change to just not use `mode=json` in any of these places, IF we confirm the point above - both in relation to creating the `System` with an `AnyUrl` field, and in relation to the audit log entries that are created by updating a `System` with an `AnyUrl` field.

tests/ctl/core/test_api.py
```diff
@@ -965,6 +965,21 @@ async def test_system_create(
         assert privacy_decl.shared_categories == ["user"]
         assert privacy_decl.flexible_legal_basis_for_processing is True
 
+    def test_system_create_with_privacy_policy(
```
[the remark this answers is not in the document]
adamsachs: yeah @andres-torres-marroquin it does, actually looking more closely i'm not even sure we need the additional - it looks like we already had test coverage in place since this fixture always had URL fields being populated...apologies that i hadn't realized that initially. so i think you can probably go ahead and remove this additional test...

i _would_ however ensure you add in test coverage for the errored case your fixing - namely, trying to create a system with a `vendor_deleted_date` populated. i assume we don't have test coverage on that already, because if we did, i'd expect it to have been failing previously?

adamsachs (commented):
OK @andres-torres-marroquin it looks like we had more test coverage initially on the URL values than i realized. but i still think we should get test coverage in place for the problem you're actually solving.

and we should also make the corresponding changes to the `update_system` function...

src/fides/api/db/system.py
adamsachs: sure, but the `vendor_deleted_date` or any other date field could still be referenced here in the `system` payload when updating a system, couldn't it? so then `_audit_system_changes` would get some dates that are not serialized well?

if you're 100% sure this isn't something we need to worry about at all, i'll let it go :)

adamsachs (approved):
👍

PR #5790, opened 2025-02-20
Read only fields correctly removed from row

Linker44:
### Description Of Changes

When evaluating read only fields to remove them from the row before masking we used their string_path. This meant no problem for fields that were not inside of lists so for example if i had a nested object
```
{
  "name": "Jhon"
  "address": {
         "street": "bleaker 123"
  }
}
```
the string path of the field street would resolve to "address.street"

But if i had a list of objects
```
{
  "name": "Jhon"
  "addresses": [{
         "street": "bleaker 123"
  }]
}
```
the string path would still resolve to "addresses.street"

Which to pydash is path that leads to nothing.

the correct string path would be "addresses.0.street"

### Code Changes

* [ ] correct functionality of generate_update_param_values

### Steps to Confirm

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:


galvana (approved):
Nice 👍

PR #5810, opened 2025-02-25
Modify dependabot.yaml to support frontend monorepo security updates

daveqnet:
Closes OPS-751

### Description Of Changes

- Dependabot security update PRs initiated manually from security alerts consistently fail with the error `The lockfile might be out of sync?` for dependencies in the frontend turbo monorepo
- This happens even though running `npm install --package-lock-only` in the `clients` folder results in `up to date, audited 2224 packages in 4s` (no sync issues)
- Security update PRs work for npm projects that are not in the monorepo e.g. `clients/sample-app`
- This may be the same issue described at https://github.com/dependabot/dependabot-core/issues/4993 and https://github.com/dependabot/dependabot-core/issues/5226

### Code Changes

* Removes monorepo workspaces from dependabot config file
* Adds monorepo root
* Removes IAB dependency version update config (Dependabot does not support two rules for the same directory)
* Adds cypress-e2e non-monorepo project (previously missing)

### Steps to Confirm

* Once merged to main, attempt to initiate a Dependabot security update PR from a security alert

### Pre-Merge Checklist

* [x] Issue requirements met
* [x] `CHANGELOG.md` updated
* Followup issues:
  * [x] No followup issues
* Database migrations:
  * [x] No migrations
* Documentation:


.github/dependabot.yaml
```diff
@@ -3,45 +3,49 @@ updates:
...
     schedule:
-      interval: "weekly"
+      interval: "daily"
     labels:
       - dependencies
       - python
       - dependabot
-  - package-ecosystem: npm
-    directory: "/clients/admin-ui"
+    open-pull-requests-limit: 0  # Security updates only
```
gilluminate, line 11: does this mean unlimited, or none?

daveqnet: Hi @gilluminate, it means "none", but it's a little more complicated than this even. Let me try to explain. 🙂 

Dependabot differentiates between ["security updates"](https://docs.github.com/en/code-security/dependabot/dependabot-security-updates/about-dependabot-security-updates) and ["version updates"](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/about-dependabot-version-updates).

A "security update" is a PR raised to bump a vulnerable dependency up to a version where the vulnerability is patched.

A "version update" is a PR raised to bump a dependency up to the latest version whenever a new version is released, regardless of vulnerabilities.

`dependabot.yaml` is only really supposed to be used to manage version updates, not security updates. You can see that I've actually enabled security updates as a repo setting, not via this file.

> Screenshot 2025-02-26 at 16 13 10

That would normally be fine, and we wouldn't have to even keep `dependabot.yaml` in the repo if we just wanted security updates and no version updates.

However, the problem I'm having is that Dependabot can't seem to create security updates when I manually ask it to in a security alert.

> Screenshot 2025-02-26 at 16 21 25


To try to work around this, I'm configuring `dependabot.yaml`to automatically create PRs for the deps in the `/clients` directory, but setting `open-pull-requests-limit: 0` is a kind of hack that will only allow security update PRs to be created. No version update PRs.

The reason this works is hinted at here in the docs: https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#open-pull-requests-limit-

> Security updates have a separate, internal limit of ten open pull requests which cannot be changed.

Confusing, I know!! 😅

gilluminate: Sounds good. I wish you all the luck in the world :)

PR #5812, opened 2025-02-25
LJ-367 S3 and local uploads

JadeCara:
Closes [LJ-367](https://ethyca.atlassian.net/browse/LJ-367?atlOrigin=eyJpIjoiYjNiYmQ4NzRlMjgxNDNlY2FjY2RkMzFiYWI1MzA0ZTgiLCJwIjoiaiJ9)

### Description Of Changes

These changes add the upload/download/retrieval functionality to the attachments class.

### Code Changes

* Introduced `@withconfig` StorageConfig retrieval wrapper/error checker
* Introduced S3 client functionality
* Introduced local storage functionality
* Updated Attachments to require use of `create_with_upload` function and added error handling for attempting attachment creation without attached file. 
* Added tests for all

### Steps to Confirm

1.  All changes are covered by tests
2. No functionality changes expected.

### Pre-Merge Checklist

* [x] Issue requirements met
* [x] `CHANGELOG.md` updated
* Followup issues:
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:

[LJ-367]: https://ethyca.atlassian.net/browse/LJ-367?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/fides/api/tasks/storage.py
JadeCara, line 1: Updating the type to `AnyHttpUrlString` to appease mypy.

dev-requirements.txt
```diff
@@ -3,6 +3,7 @@ debugpy==1.6.3
 Faker==14.1.0
 GitPython==3.1.41
 isort==5.12.0
+moto[s3]==5.1.0
```
JadeCara, line 6: Allows testing of S3 with accurate response types.

tests/ctl/models/test_attachment.py
```diff
@@ -9,6 +13,32 @@
...
+def s3_client(storage_config):
+    with mock_aws():
+        session = boto3.Session(
+            aws_access_key_id="fake_access_key",
+            aws_secret_access_key="fake_secret_key",
+            region_name="us-east-1",
+        )
+        s3 = session.client("s3")
+        s3.create_bucket(Bucket=storage_config.details[StorageDetails.BUCKET.value])
+        yield s3
```
JadeCara, line 29: mock_S3 fixture allows us to test with s3 calls and expected bucket names.

erosselli: this is very cool! 😎

erosselli (commented):
This is looking good :) I did leave some comments and suggestions. Thanks for adding so many tests 🤩

src/fides/api/models/attachment.py
```diff
@@ -90,64 +125,114 @@ class Attachment(Base):
...
-        #         "message": "",
-        #     },
-        # )
+    def config(self, db: Session) -> Optional[StorageConfig]:
+        """Returns the storage config for the attachment."""
+        return StorageConfig.get_by(db=db, field="key", value=self.storage_key)
+
+    def config_error_handler(self) -> None:
+        log.error(f"Storage config not found for key: {self.storage_key}")
+        raise ValueError(f"Storage config not found for key: {self.storage_key}")
```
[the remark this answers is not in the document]
JadeCara: Its a good point. I will make this update :)

src/fides/api/models/attachment.py
```diff
@@ -9,6 +14,36 @@
...
+    def wrapper(
+        self: "Attachment", db: Session, *args: Any, **kwargs: Any
+    ) -> Optional[StorageConfig]:
+        config = self.config(db)
+        if config is None:
+            self.config_error_handler()
+            return None
+        return method(self, db, config, *args, **kwargs)
+
+    return wrapper  # type: ignore
```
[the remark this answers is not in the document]
JadeCara: Ended up removing this. Adding the Foreign Key made this unnecessary.

src/fides/api/models/attachment.py
```diff
@@ -90,64 +125,114 @@ class Attachment(Base):
...
+    def config(self, db: Session) -> Optional[StorageConfig]:
+        """Returns the storage config for the attachment."""
+        return StorageConfig.get_by(db=db, field="key", value=self.storage_key)
+
+    def config_error_handler(self) -> None:
+        log.error(f"Storage config not found for key: {self.storage_key}")
+        raise ValueError(f"Storage config not found for key: {self.storage_key}")
+
+    @with_config
+    def upload(self, db: Session, config: Any, attachment: bytes) -> None:
```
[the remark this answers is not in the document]
JadeCara: Again, the FK made this all so much neater. :)

src/fides/api/models/attachment.py
```diff
@@ -90,64 +125,114 @@ class Attachment(Base):
...
+
+            filename = f"{LOCAL_FIDES_UPLOAD_DIRECTORY}/{self.id}"
+            with open(filename, "wb") as file:
+                file.write(attachment)
+            return
+
+        raise ValueError(f"Unsupported storage type: {config.type}")
+
+    @with_config
+    def download_attachment_from_s3(
```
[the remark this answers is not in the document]
JadeCara: There is a [storage_uploader_service](https://github.com/ethyca/fides/blob/c76f461d33bc70a1f0c8c7bd346071c3cc84d11c/src/fides/api/service/storage/storage_uploader_service.py) - it is pretty specific to `PrivacyRequests.`
I will see if I can pull out some functionality to make it more abstract and house it here.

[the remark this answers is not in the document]
JadeCara: I think I might remove this specific function though and handle it at an endpoint. It is S3 specific and returns a presigned url to download the original attachment using an existing function. I don't think it needs to live here and could be handled by an endpoint.

src/fides/api/models/attachment.py
```diff
@@ -90,64 +125,114 @@ class Attachment(Base):
...
+
+        if config.type == StorageType.local:
+            if not os.path.exists(LOCAL_FIDES_UPLOAD_DIRECTORY):
+                os.makedirs(LOCAL_FIDES_UPLOAD_DIRECTORY)
+
+            filename = f"{LOCAL_FIDES_UPLOAD_DIRECTORY}/{self.id}"
+            with open(filename, "wb") as file:
+                file.write(attachment)
+            return
+
```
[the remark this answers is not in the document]
JadeCara: There is a lot of coupling there which makes me reticent to go in and update it now. I might be able to sneak some of that work in when I update the privacy requests to expect comments and attachments. I have included some generic S3 operations that are less restrictive and tests for them as well in this PR.

src/fides/api/alembic/migrations/versions/6ea2171c544f_change_attachment_storage_key_to_.py
```diff
@@ -0,0 +1,53 @@
...
+        nullable=False,
+    )
+    op.create_foreign_key(
+        "fk_attachment_storage_key",
+        "attachment",
+        "storageconfig",
+        ["storage_key"],
+        ["key"],
+        ondelete="CASCADE",
+    )
```
erosselli, line 37: just double-checking here: the Attachment model was created this sprint right? asking cause the FK constraint would fail if the DB has any Attachments created that have a `storage_key` that is not a "real" `StorageConfig` key.  Not concerned about local envs but did want to flag it so we can just make sure this hasn't been done in our staging envs (will depend on whether the Attachment model was in a previous release or not)

src/fides/api/tasks/storage.py
```diff
@@ -119,23 +122,112 @@ def create_presigned_url_for_s3(s3_client: Any, bucket_name: str, file_key: str)
...
     """Uploads arbitrary data to s3 returned from an access request"""
     logger.info("Starting S3 Upload of {}", file_key)
 
+    if privacy_request is None and document is not None:
+        return generic_upload_to_s3(
+            storage_secrets, bucket_name, file_key, auth_method, document
+        )
+
+    if privacy_request is None:
+        raise ValueError("Privacy request must be provided")
```
erosselli, line 227: I see we're changing `upload_to_s3` to include this , but in the Attachment model we call `generic_upload_to_s3` so I don't know if I understand why this change was needed

src/fides/api/util/aws_util.py
```diff
@@ -70,3 +70,22 @@ def get_aws_session(
             raise
     else:
         return session
+
+
+def get_s3_client(
```
erosselli, line 75: 🤩

erosselli (approved):
Thanks for making the changes! I'm approving, left a question and then just a suggestion for future work

src/fides/api/models/attachment.py
```diff
@@ -90,64 +103,108 @@ class Attachment(Base):
...
+                auth_method=auth_method,
+                document=attachment,
+            )
+            log.info(f"Uploaded {self.file_name} to S3 bucket {bucket_name}/{self.id}")
+            return
+
+        if self.config.type == StorageType.local:
+            filename = get_local_filename(self.id)
+            with open(filename, "wb") as file:
+                file.write(attachment)
```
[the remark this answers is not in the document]
JadeCara: I think this would be a good thing to work on in my next [Comments and Attachments ticket ](https://ethyca.atlassian.net/browse/LJ-369?atlOrigin=eyJpIjoiOGNiNjdiZTRkNDNjNDBlZWEwZjhjZGY1NDkxOWExMTgiLCJwIjoiaiJ9) since I could do the same for privacy requests and clean it all up at the same time. I will add the details to that ticket.

src/fides/api/tasks/storage.py
JadeCara: Good call 😆

src/fides/api/alembic/migrations/versions/6ea2171c544f_change_attachment_storage_key_to_.py
JadeCara: Yeah, the attachment models got merged after the release. There is also no way to actually make an attachment yet, no active endpoints etc. So we are probably safe.

src/fides/api/service/storage/storage_uploader_service.py
```diff
@@ -88,14 +86,13 @@ def _s3_uploader(
     config: StorageConfig,
     data: Dict,
     privacy_request: PrivacyRequest,
-    data_category_field_mapping: Optional[DataCategoryFieldMapping] = None,
-    data_use_map: Optional[Dict[str, Set[str]]] = None,
```
JadeCara, line 92: These are passed around but not actually used in these uploads. Removing for cleanliness.

PR #5813, opened 2025-02-25
Fix action column width on D&D tables

jpople:
Closes HA-179

### Description Of Changes

Fixes a bug where D&D action buttons were sometimes cut off at certain screen sizes and after resizing other columns.

https://github.com/user-attachments/assets/1206a8f2-170f-4279-bb03-7a1302044c40

### Code Changes

* Changes all "Actions" columns in D&D tables to have width `auto`
* Also removes stopPropagation calls to move to updated `disableRowClick`pattern

### Steps to Confirm

1. Populate the D&D tables with some items
2. On both detection tables and discovery tables, make sure action buttons are visible and don't get cut off when resizing

### Pre-Merge Checklist

* [x] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [x] No followup issues
* Database migrations:
  * [x] No migrations
* Documentation:

### DIFF
--- a/clients/admin-ui/src/features/data-discovery-and-detection/hooks/useDiscoveryResultColumns.tsx
+++ b/clients/admin-ui/src/features/data-discovery-and-detection/hooks/useDiscoveryResultColumns.tsx
@@ -76,6 +76,7 @@ const useDiscoveryResultColumns = ({
         header: "Actions",
         meta: {
           width: "auto",
+          disableRowClick: true,
         },
       }),
     ];
@@ -141,6 +142,7 @@ const useDiscoveryResultColumns = ({
         header: "Actions",
         meta: {
           width: "auto",
+          disableRowClick: true,
         },
       }),
     ];
@@ -201,6 +203,7 @@ const useDiscoveryResultColumns = ({
         header: "Actions",
         meta: {
           width: "auto",
+          disableRowClick: true,
         },
       }),
     ];
--- a/clients/admin-ui/src/features/data-discovery-and-detection/hooks/useDiscoveryResultColumns.tsx
+++ b/clients/admin-ui/src/features/data-discovery-and-detection/hooks/useDiscoveryResultColumns.tsx
@@ -17,6 +17,9 @@ import DiscoveryItemActionsCell from "../tables/cells/DiscoveryItemActionsCell";
 import EditCategoryCell from "../tables/cells/EditCategoryCell";
 import ResultStatusCell from "../tables/cells/ResultStatusCell";
 
+const NAME_COLUMN_SIZE = 300;
+const ACTION_COLUMN_SIZE = 220;
+
 const useDiscoveryResultColumns = ({
   resourceType,
 }: {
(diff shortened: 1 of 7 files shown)


lucanovera (commented):
Thanks for fixing the stop propagation too, disableRowClick is the way to go. I've tested in Chrome and Firefox and it's not cutting off anymore. 

The variable auto width should be the more correct option, but for some reason it's making the column fixed and too wide for just the two buttons. Can we make a last effort to try to reduce that space before merging the fix? I've played around width using size: 250 instead of 180.

clients/admin-ui/src/features/data-discovery-and-detection/hooks/useDiscoveryResultColumns.tsx
```diff
@@ -17,6 +17,9 @@ import DiscoveryItemActionsCell from "../tables/cells/DiscoveryItemActionsCell";
 import EditCategoryCell from "../tables/cells/EditCategoryCell";
 import ResultStatusCell from "../tables/cells/ResultStatusCell";
 
+const NAME_COLUMN_SIZE = 300;
+const ACTION_COLUMN_SIZE = 220;
```
jpople, line 21: This one is slightly wider since it potentially needs to accommodate a kebab menu also.

jpople:
@lucanovera The column width behavior is pretty confusing and I think we've been using `width: auto` wrong in a few places-- putting `auto` on any column essentially forces all the non-auto columns to be their minimum sizes and allocates the rest of the space to the `auto` column(s).  

I've removed `auto` from the actions column in the discovery tables and added static widths, which has largely fixed the excess space issue; it doesn't seem feasible to guarantee zero excess space, but at least it won't be taking up the entire table now.

lucanovera:
> @lucanovera The column width behavior is pretty confusing and I think we've been using `width: auto` wrong in a few places-- putting `auto` on any column essentially forces all the non-auto columns to be their minimum sizes and allocates the rest of the space to the `auto` column(s).
> 
> I've removed `auto` from the actions column in the discovery tables and added static widths, which has largely fixed the excess space issue; it doesn't seem feasible to guarantee zero excess space, but at least it won't be taking up the entire table now.

Thanks @jpople . I'm testing and it's looking great. No cutoffs or excess spaces.

clients/admin-ui/src/features/data-discovery-and-detection/hooks/useDiscoveryResultColumns.tsx
lucanovera:
```suggestion
const ACTION_COLUMN_SIZE = 235;
```
Good catch! I did found that if I intentional shink the column down to it's minimum width it will still generate a bit of scrolling. Bringing it up a to 235 removes that scroll.

Captura de pantalla 2025-02-26 a la(s) 2 57 04 p  m

lucanovera (approved):
Looking great! Tested in Chrome and Firefox, switching resolutions and column widths. Code looks good. Approved!

PR #5819, opened 2025-02-27
Fix pagination issues on some tables

jpople:
Closes HA-421

### Description Of Changes

Fixes a bug on some tables where the displayed "total" value in the pagination bar was incorrect.

Before:

Screenshot 2025-02-27 at 12 32 15

After:

Screenshot 2025-02-27 at 12 40 33

Also fixes a bug where users could page forward on empty tables.

Before: 

Screenshot 2025-02-27 at 12 31 49

After: 

Screenshot 2025-02-27 at 15 07 34

### Code Changes

* Fix web monitor tables to use `data.total` instead of `data.items.length` to display end range of current page
* Fix some values and logic in `useServerSidePagination`

### Steps to Confirm

1.  Seed the web monitor tables with sample data using `nox -s demo`
2. View assets table; should be able to page through table and range (e.g. "1-25 of 316") should display correct values
3. View an empty table (e.g. detection or discovery results)
4. Range should display "0-0 of 0" and page buttons should be disabled

### Pre-Merge Checklist

* [x] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [x] No followup issues
* Database migrations:
  * [x] No migrations
* Documentation:


clients/admin-ui/src/features/common/table/v2/PaginationBar.tsx
```diff
@@ -53,13 +53,13 @@ export const useServerSidePagination = () => {
     setPageIndex((prev) => prev + 1);
   }, [setPageIndex]);
   const isNextPageDisabled = useMemo(
-    () => !!totalPages && (pageIndex === totalPages || totalPages < 2),
+    () => totalPages === 0 || pageIndex === totalPages,
```
jpople, line 56: Drive-by fix; this wasn't handling the case with 0 pages correctly.  When there's 1 page `pageIndex` should always be 1 so the other check was redundant.

clients/admin-ui/src/features/common/table/v2/PaginationBar.tsx
```diff
@@ -53,13 +53,13 @@ export const useServerSidePagination = () => {
...
     [pageIndex, totalPages],
   );
 
-  const startRange =
-    (pageIndex - 1) * pageSize === 0 ? 1 : (pageIndex - 1) * pageSize;
-  const endRange = (pageIndex - 1) * pageSize + pageSize;
+  const pageStartIndex = (pageIndex - 1) * pageSize;
+  const startRange = totalPages !== 0 ? pageStartIndex + 1 : 0;
+  const endRange = pageStartIndex + pageSize;
```
jpople, line 62: Only functional changes here are:

* when total is 0, should show "0-0 of 0" instead of "1-0 of 0"
* should show, e.g., "26-50" instead of "25-50", since the 25th item is on the first page

clients/admin-ui/src/features/common/table/v2/PaginationBar.tsx
gilluminate: Having `!!totalPages` instead of `totalPages === 0` was meant to handle cases where totalPages is `null` or `undefined` in cases where there's an error in the response or it returns a `null` value. This was happening to me when I made that update.

clients/admin-ui/src/features/common/table/v2/PaginationBar.tsx
gilluminate: how does this handle the `null` or `undefined` cases? Should total be 0 by default instead of 1, where the useState is defined?

jpople:
@gilluminate Thanks for the clarification-- I've updated the logic a bit to handle `null` and `undefined` better.

clients/admin-ui/src/features/common/table/v2/PaginationBar.tsx
```diff
@@ -52,14 +52,23 @@ export const useServerSidePagination = () => {
...
+  let startRange;
+  let endRange;
+  if (totalPages) {
+    const pageStartIndex = (pageIndex - 1) * pageSize;
+    startRange = pageStartIndex + 1;
+    endRange = pageStartIndex + pageSize;
+  } else {
+    startRange = 0;
+    endRange = 0;
+  }
```
gilluminate, line 71: Nice and clean. Love it.

gilluminate (approved):

PR #5820, opened 2025-02-27
LJ-477 Improvements to GraphTask and poll_for_exited_privacy_request_tasks

erosselli:
Closes [LJ-447](https://ethyca.atlassian.net/browse/LJ-447)

### Description Of Changes

**Issue**: We were seeing a lot of data being transferred from the DB to the fides webserver 

**Explanation**:
We have a recurring task `poll_for_exited_privacy_request_tasks` that checks all in-progress privacy requests to see if they have any errored tasks. This task was inadvertently querying for all fields in the request tasks table (including task results), instead of for just task status. 

At the same time, we have long-running privacy request that has a LOT of request tasks, and since it's stuck in "in progress", all its tasks keep getting polled by the recurring check . This long-running request fails to move to a completed status because of access tasks that take a long time to process: the task processes fine, but when it goes to update its own status in the DB, the DB session it was using has already been closed .

### Code Changes

* Update code in `src/fides/api/task/graph_task.py` to use a new DB session to log updates to the request task status
* Update `poll_for_exited_privacy_request_tasks` to only query for `RequestTask.status` rather than the entire row
* I had to move some service functions from `api.api.deps` to a new file `api.service.deps` so that I could avoid circular imports

### Steps to Confirm

1.  Update your `fides.toml` to run DSR 3.0 by setting `use_dsr_3_0 = true` under `[execution]` 
2. Create a System (or choose an existing one) and link an integration of your choosing to it (e.g Postgres ) 
3. Create a Dataset for that integration and link it to the integration 
4. Run a DSR
5. Privacy Request should run OK 
6. Change the credentials for the integration so that they are wrong (i.e integration will fail)
7. Run another DSR 
8. Privacy Request should show in errored status as expected, with the Activity Timeline showing correctly 
9. Update your `fides.toml` back to DSR 2.0 , and re-run steps 2-8

### Pre-Merge Checklist

* [ ] Issue requirements met
* [x] `CHANGELOG.md` updated
* Followup issues:
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:

[LJ-447]: https://ethyca.atlassian.net/browse/LJ-447?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/fides/api/api/deps.py
+++ b/src/fides/api/api/deps.py
@@ -10,10 +10,6 @@ from fides.api.util.cache import get_cache as get_redis_connection
 from fides.config import CONFIG, FidesConfig
 from fides.config import get_config as get_app_config
 from fides.config.config_proxy import ConfigProxy
-from fides.service.dataset.dataset_config_service import DatasetConfigService
-from fides.service.dataset.dataset_service import DatasetService
-from fides.service.messaging.messaging_service import MessagingService
-from fides.service.privacy_request.privacy_request_service import PrivacyRequestService
 
 _engine = None
 
@@ -70,27 +66,3 @@ def get_cache() -> Generator:
             "Application redis cache required, but it is currently disabled! Please update your application configuration to enable integration with a Redis cache."
         )
     yield get_redis_connection()
-
-
-def get_messaging_service(
-    db: Session = Depends(get_db),
-    config: FidesConfig = Depends(get_config),
-    config_proxy: ConfigProxy = Depends(get_config_proxy),
-) -> MessagingService:
-    return MessagingService(db, config, config_proxy)
-
-
-def get_privacy_request_service(
-    db: Session = Depends(get_db),
-    config_proxy: ConfigProxy = Depends(get_config_proxy),
-    messaging_service: MessagingService = Depends(get_messaging_service),
-) -> PrivacyRequestService:
-    return PrivacyRequestService(db, config_proxy, messaging_service)
-
-
-def get_dataset_service(db: Session = Depends(get_db)) -> DatasetService:
-    return DatasetService(db)
-
-
-def get_dataset_config_service(db: Session = Depends(get_db)) -> DatasetConfigService:
-    return DatasetConfigService(db)
--- a/src/fides/api/service/privacy_request/request_service.py
+++ b/src/fides/api/service/privacy_request/request_service.py
@@ -8,7 +8,6 @@ from typing import Any, Dict, List, Optional, Set
 from httpx import AsyncClient
 from loguru import logger
 from sqlalchemy import text
-from sqlalchemy.orm import Query
 from sqlalchemy.sql.elements import TextClause
 
 from fides.api.common_exceptions import PrivacyRequestNotFound
@@ -224,7 +223,7 @@ def poll_for_exited_privacy_request_tasks(self: DatabaseTask) -> Set[str]:
                 RequestTask.action_type == task_type,
             )
 
-            statuses = set([status for status, in tasks_statuses_query.all()])
+            statuses = set(status for status, in tasks_statuses_query.all())
             all_exited = all(
                 status in EXITED_EXECUTION_LOG_STATUSES for status in statuses
             )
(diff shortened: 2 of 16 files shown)


src/fides/api/task/graph_task.py
```diff
@@ -217,6 +218,7 @@ def __init__(
     ):  # cache config, log config, db store config
         super().__init__()
         self.request_task = resources.privacy_request_task
+        self.request_task_id = resources.privacy_request_task.id
```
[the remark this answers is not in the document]
erosselli: I had to change this up a bit to use `db.merge` because there were some changes to the `self.request_task` object that hadn't been committed yet, it seems like they're saved in the `update_status` call so tests were breaking because those changes were "lost"

src/fides/api/api/deps.py
```diff
@@ -70,27 +66,3 @@ def get_cache() -> Generator:
             "Application redis cache required, but it is currently disabled! Please update your application configuration to enable integration with a Redis cache."
         )
     yield get_redis_connection()
-
-
-def get_messaging_service(
```
adamsachs, line 75: just noting for self/posterity that these were not removed, but rather moved over to `api/service/deps.py` 👍

src/fides/api/service/privacy_request/request_service.py
```diff
@@ -210,34 +210,48 @@ def poll_for_exited_privacy_request_tasks(self: DatabaseTask) -> Set[str]:
...
         )
 
-        def some_errored(tasks: Query) -> bool:
-            """All statuses have exited and at least one is errored"""
-            statuses: List[ExecutionLogStatus] = [tsk.status for tsk in tasks]
+        def privacy_request_has_errored_tasks(
+            privacy_request: PrivacyRequest, task_type: ActionType
+        ) -> bool:
+            """Check if a privacy request has all exited all its tasks of the given type,
+            and at least on the tasks has errored.
```
adamsachs, line 217: nit: typo
```suggestion
            and at least one of  the tasks has errored.
```

adamsachs: but nice docstring here to explain the context + optimization :)

src/fides/api/service/privacy_request/request_service.py
```diff
@@ -210,34 +210,48 @@ def poll_for_exited_privacy_request_tasks(self: DatabaseTask) -> Set[str]:
...
+            We specifically only query for the RequestTask.status column to avoid
+            querying for the entire RequestTask row.
+            """
+            tasks_statuses_query = db.query(RequestTask.status).filter(
+                RequestTask.privacy_request_id == privacy_request.id,
+                RequestTask.action_type == task_type,
+            )
```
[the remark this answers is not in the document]
adamsachs: yup, this is nice!

somewhat relatedly: this is likely out of scope for this patch PR, and likely not going to make as much of an impact (at least for our current problem) - but i do wonder if we should also be trying to run the `RequestTask` queries up front as a 'bulk' query. it seems like in our current situation, the # of `RequestTask`s dwarf the # of in progress `PrivacyRequest`s, and that may generally be the case. but i could imagine a scenario where we have many in progress `PrivacyRequest`s, and we're going to basically be issuing `3*n` queries where `n` is the # of in progress `PrivcayRequest`s, right? since we do know all the in progress `PrivacyRequest`s up front, an optimization here would be to just query for _all_ of their `RequestTasks`s up front in a bulk query, and then do the filtering/processing in-memory.

i know this isn't the source of the problem we're solving today, but trying to think one step ahead and here and look for optimizations while i've got close 👀 on this code - so maybe we leave a TODO comment here, if you all agree that it'd be a worthwhile optimization at some point?

adamsachs (approved):
this looks good to me! one nit on a comment and something to think about moving forward, but should be good to go once CI is passing 👍

PR #5821, opened 2025-02-28
Fides GTM & Event Origination.

tvandort:
Closes [<issue>]

### Description Of Changes

### Code Changes

### Steps to Confirm

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:


clients/fides-js/src/integrations/gtm.ts
```diff
@@ -38,16 +38,26 @@ const pushFidesVariableToGTM = (fidesEvent: {
...
+    FidesInitializing: true,
+    FidesInitialized: true,
+    FidesUpdating: true,
+    FidesUpdated: true,
+    FidesUIChanged: true,
+    FidesUIShown: true,
+    FidesModalClosed: true,
+  };
+
+  const events = Object.keys(fidesEvents) as FidesEventType[];
```
gilluminate, line 53: I would consider bumping this to a constants file that can be imported here and in the test (and anywhere else we need it in the future). Would it make more sense as an array instead of an object? Removes the need to coerce it to an array on the next line.
```suggestion
  const fidesEvents: FidesEventType[] = [
    FidesInitializing,
    FidesInitialized,
    FidesUpdating,
    FidesUpdated,
    FidesUIChanged,
    FidesUIShown,
    FidesModalClosed,
  };
```

tvandort: @gilluminate 

I don't think that an `Array` would do what I'm aiming to do here but I'm not a type system expert. The point of this change wasn't to reduce typing. The point of this change is to make sure that the event list was exhaustive. 

With the way this is written if anyone ever adds a new type to `FidesEventType` what ends up happening is Typescript realizes that this object is missing a key since it MUST be implemented. AFAIK there's no way to write "this array is an exhaustive list of keys for `FidesEventType`".

gilluminate: I see what you are saying. I think the more correct thing might be to use an `enum` to define the event types instead of a list of strings. That way you can iterate over the enum here, and any additions/changes to the event list is only happening in 1 place. That will require a bit of refactor though, to use the enum in all of those places instead of a string.

tvandort: I would still have to cast in that case, no? 

image

Also the downside of that approach (as far as I can project) would be that it would make ALL events push to GTM. Maybe there's a reason we would want _some_ events not to go to GTM? I'm not sure. I also didn't want to introduce behavior that would have an author not be aware of the fact that they're changing the GTM integration when adding events. The current structure alerts the author to the existence of the connection to GTM without forcing GTM adoption on them (they could use `Pick<>` etc.)

I can see an argument for "yes we should ALWAYS push every Fides event to GTM" but I was trying to be more circumspect than that.

gilluminate: Ok I'm getting it now. I'm on board with the original code now, thanks for walking me through it.

clients/fides-js/__tests__/integrations/gtm.test.ts
```diff
@@ -0,0 +1,53 @@
...
+      );
+
+      expect(
+        window.dataLayer?.filter((event) => event.event === eventName).length,
+      ).toBeGreaterThanOrEqual(1);
+    },
+  );
+
+  test.each(eventsThatDoNotFire)(
+    "that fides forwards %s event to gtm if appropriate",
```
gilluminate, line 41: Should the name of this be updated to reflect the split?
```suggestion
    "that fides doesn't forward %s event to gtm",
```

clients/fides-js/docs/interfaces/FidesEvent.md
gilluminate, line 1: My local build has more changes to this file. Be sure the build fides.js locally to get an updated version to push.

clients/fides-js/src/components/tcf/TcfPurposes.tsx
```diff
@@ -174,14 +178,17 @@ const TcfPurposes = ({
...
-            modelType: item.bestTranslation
-              ? "customPurposesConsent"
-              : activeData.purposeModelType,
-          })
+        onToggle={(newEnabledIds, item, toggleDetails) =>
+          onChange(
+            {
+              newEnabledIds,
+              // @ts-ignore
+              modelType: item.bestTranslation
```
gilluminate, line 186: I know this isn't your doing, but it would be nice to make this a bit safer while we're here
```suggestion
              modelType: (item as PrivacyNoticeWithBestTranslation).bestTranslation
```

clients/fides-js/src/lib/events.ts
```diff
@@ -8,12 +8,11 @@ declare global {
...
 /**
  * Defines the type of "extra" details that can be optionally added to certain
- * events. This is intentionally vague, but constrained to be basic (primitive)
- * values for simplicity.
+ * events. This is intentionally vague. See the /docs/fides-event.ts
  */
 export type FidesEventExtraDetails = Record<
   string,
-  string | number | boolean | undefined
+  string | number | boolean | object | undefined
```
gilluminate, line 15: In TypeScript, using object as a type is generally discouraged because it's too broad and can lead to type safety issues. The object type includes everything except primitives (string, number, boolean, symbol, null, undefined), which means it could accept arrays, functions, and other complex types that might not be what you expect. Unless that _is_ what we're expecting and that's part of the "vague for simplicity" we're talking about?

```suggestion
  string | number | boolean | Record<string, unknown> | undefined
```

clients/fides-js/src/components/tcf/TcfPurposes.tsx
tvandort: Isn't that a 6 of one, half a dozen of another sort of change?

clients/fides-js/src/lib/events.ts
tvandort: Hah, I swear I looked that up. Thank you!

clients/fides-js/src/components/tcf/TcfPurposes.tsx
gilluminate: Sort of. If there were other TS issues besides this one, they would not get seen. This is easier for future devs, basically.

gilluminate: An example of where this could be problematic is if `bestTranslation` were also removed from `PrivacyNoticeWithBestTranslation` and that wouldn't get picked up here because of the `//@ts-ignore`

tvandort: Ah ok.

gilluminate (approved):

clients/fides-js/docs/interfaces/FidesEvent.md
NevilleS: Updated!

NevilleS:
OK @gilluminate I pushed some updates to the Cypress tests to update those to include the new GTM events and do some (light) cleanup.

NevilleS (approved):
Overall, this is looking great. Thanks gang

clients/privacy-center/cypress/e2e/consent-banner.cy.ts
```diff
@@ -2122,7 +2122,19 @@ describe("Consent overlay", () => {
 
         // Toggle the notice, but don't save yet
         cy.getByTestId("toggle-Advertising").click();
-        cy.get("@FidesUIChanged").should("have.been.calledOnce");
+        cy.get("@FidesUIChanged").its("callCount").should("equal", 1);
```
gilluminate, line 2125: @NevilleS Just curious what the difference is here? Just for consistency?

gilluminate (approved):
Great detail added in the tests @NevilleS!

clients/privacy-center/cypress/e2e/consent-banner.cy.ts
NevilleS: Yeah, that's all

PR #5828, opened 2025-02-28
Fix load_samples to correctly collect & load sample connections with "False" secret values

NevilleS:
### Description Of Changes

When `FIDES__DATABASE__LOAD_SAMPLES` is set, the `load_samples` function will run at startup to load sample resources: systems, datasets, etc. It also loads various sample connections (e.g. Postgres, Mongo, Stripe), but it does this _conditionally_ only if the environment is configured with all the expected secrets for each connection.

This logic had a subtle footgun; if a secret value was a boolean like `FIDES_DEPLOY__CONNECTORS__POSTGRES__SSH_REQUIRED=False`, the code would interpret that as a missing secret and skip the sample connection! Whoops.

### Code Changes

* Fix `load_sample_connections_from_project()` to only ignore `None` or `""` as missing secrets
* Improve DEBUG logs to indicate exactly what sample connections are found, skipped, and collected 

### Steps to Confirm

1. Set `FIDES_DEPLOY__CONNECTORS__POSTGRES__SSH_REQUIRED=False` and confirm that the Postgres sample connection is configured
2. Remove `FIDES_DEPLOY__CONNECTORS__POSTGRES__SSH_REQUIRED` and confirm that the Postgres sample connection is skipped

### Pre-Merge Checklist

* [x] Issue requirements met
* [x] `CHANGELOG.md` updated
* Followup issues:
  * [x] No followup issues
* Database migrations:
  * [x] No migrations
* Documentation:


src/fides/api/db/samples.py
```diff
@@ -106,17 +108,41 @@ def load_sample_connections_from_project() -> List[SampleConnection]:
...
+        ]
+        if len(missing_keys) > 0:
+            log.debug(
+                f"Skipping sample connection {connection.key}. To include this sample, configure ENV secrets for these missing keys: {missing_keys} (see {sample_connections_path}/* for expected ENV vars)"
+            )
             continue
 
-        # Check if all secret values are present and non-empty
-        if all(value and value != "" for value in connection.secrets.values()):  # type: ignore
-            valid_connections.append(connection)
```
NevilleS, line 117: This was the bug: when `value == False` intentionally, this doesn't work, because `all([False, "localhost", "fides", ...])` will be false.

When a secret value isn't set in the ENV, it'll be `None` (or maybe `""`), so this check needed to be fixed. Whoops

src/fides/api/db/samples.py
```diff
@@ -106,17 +108,41 @@ def load_sample_connections_from_project() -> List[SampleConnection]:
...
+        # If all secrets are configured, include this connection
+        log.debug(
+            f"Including sample connection {connection.key} because all {len(expected_keys)} required secrets are configured"
+        )
+        valid_connections.append(connection)
 
     # Exclude any invalid connections from the final results
+    log.info(
+        f"Collected {len(valid_connections)} sample connections with configured ENV secrets: {[connection.key for connection in valid_connections]}"
+    )
```
NevilleS, line 145: nb. this is the only `INFO` log in this function, so for regular usage you should just see a log that tells you $N connections were collected and the list of expected connection keys.

DEBUG logs explain why each connection was either included or skipped

src/fides/api/db/samples.py
```diff
@@ -106,17 +108,41 @@ def load_sample_connections_from_project() -> List[SampleConnection]:
...
+                f"Skipping sample connection {connection.key}! No secrets exist, cannot create"
+            )
+            continue
+
+        # Check if all secret values exist and are non-empty
+        expected_keys = connection.secrets.keys()  # type: ignore
+        missing_keys = [
+            key
+            for key, value in connection.secrets.items()  # type: ignore
+            if value is None or value == ""
```
NevilleS, line 128: This is the right way to check if the secret value was found 👍

tests/ctl/api/test_seed.py
```diff
@@ -460,7 +460,7 @@ class TestLoadSamples:
         "FIDES_DEPLOY__CONNECTORS__POSTGRES__DBNAME": "test-var-db",
         "FIDES_DEPLOY__CONNECTORS__POSTGRES__USERNAME": "test-var-user",
         "FIDES_DEPLOY__CONNECTORS__POSTGRES__PASSWORD": "test-var-password",
-        "FIDES_DEPLOY__CONNECTORS__POSTGRES__SSH_REQUIRED": "True",
+        "FIDES_DEPLOY__CONNECTORS__POSTGRES__SSH_REQUIRED": "false",
```
NevilleS, line 463: Going forward, this test now reproduces the error we saw when deploying this too

NevilleS (commented):
Few comments!

galvana (approved):

PR #5833, opened 2025-03-03
LJ-365 New Comments models

JadeCara:
Closes [LJ-365](https://ethyca.atlassian.net/browse/LJ-365?atlOrigin=eyJpIjoiYTY0N2MwMjk0YTIzNDdlZWFkN2JmYWViYWY4ZjYzMGYiLCJwIjoiaiJ9)

### Description Of Changes

Added Comments and Comments reference models with tests. 

### Code Changes

* Added Comment model
* Added CommentReference model
* Added Tests for all included functionality. 

### Steps to Confirm

1.  No expected functionality changes. 
2. Introduced changes should be covered by tests. 

### Pre-Merge Checklist

* [x] Issue requirements met
* [x] `CHANGELOG.md` updated
* Followup issues:
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:

[LJ-365]: https://ethyca.atlassian.net/browse/LJ-365?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/ctl/models/test_attachment.py
JadeCara, line 1: Moving these fixtures up to `test/ctl/models/conftest.py` so that they are available for the comments tests as well.

src/fides/api/models/comment.py
```diff
@@ -0,0 +1,110 @@
...
+        """Overriding base class method to set the table name."""
+        return "comment_reference"
+
+    comment_id = Column(String, ForeignKey("comment.id"), nullable=False)
+    reference_id = Column(String, nullable=False)
+    reference_type = Column(EnumColumn(CommentReferenceType), nullable=False)
+
+    __table_args__ = (
+        UniqueConstraint("comment_id", "reference_id", name="_attachment_reference_uc"),
+    )
```
galvana, line 52: Possible copy/paste error?
```suggestion
    __table_args__ = (
        UniqueConstraint("comment_id", "reference_id", name="comment_reference_uc"),
    )
```

src/fides/api/models/comment.py
```diff
@@ -0,0 +1,110 @@
...
+    - reply comments are public and may cause an email or other communciation to be sent
+    """
+
+    note = "note"
+    reply = "reply"
+
+
+class CommentReferenceType(str, EnumType):
+    """
+    Enum for attachment reference types. Indicates where attachment is referenced.
```
galvana, line 29:
```suggestion
    Enum for comment reference types. Indicates where the comment is referenced.
```

src/fides/api/models/comment.py
```diff
@@ -0,0 +1,110 @@
...
+        )
+        return db.execute(stmt).scalars().all()
+
+    def delete(self, db: Session) -> None:
+        """Delete the comment and all associated references."""
+        attachments = self.get_attachments(db)
+        for attachment in attachments:
+            attachment.delete(db)
+        db.delete(self)
+        db.commit()
```
galvana, line 110: We should remove this manual `db.commit()` so we can roll-back any transaction this may be a part of. This means you might need to call `db.commit()` in your tests but that's preferred over committing this to the DB mid-transaction.
```suggestion
```

galvana (changes requested):
Nice work! I caught some typos and a minor DB issue but other than that pretty good 👍 The one thing I couldn't add inline was index considerations for the new `Comment` and `Attachment` models. What do you think the most common access pattern will be? I think most of the time we're going to be looking up these records by `reference_id` and `reference_type`. If that's the case, we should look into adding an index to make these lookups faster as these tables start getting some real use. I would usually lean away from premature optimization but it's better to add the indexes now before the tables start growing.

JadeCara:
> Nice work! I caught some typos and a minor DB issue but other than that pretty good 👍 The one thing I couldn't add inline was index considerations for the new `Comment` and `Attachment` models. What do you think the most common access pattern will be? I think most of the time we're going to be looking up these records by `reference_id` and `reference_type`. If that's the case, we should look into adding an index to make these lookups faster as these tables start getting some real use. I would usually lean away from premature optimization but it's better to add the indexes now before the tables start growing.

Such a good point. I will update these and I also added the same indexes on the attachments models in #5812

tests/ctl/models/test_attachment.py
JadeCara, line 1: Moved fixtures to conftest.py so they can be used by the comments tests as well.

tests/ops/util/test_consent_util.py
```diff
@@ -105,6 +105,9 @@ def test_notice_based_consent_multiple_preferences(
         )
         pref_1.privacy_request_id = privacy_request_with_consent_policy.id
         pref_1.save(db)
+        print(pref_1.created_at)
+        print(pref_1.updated_at)
+        print(pref_1.affected_system_status)
```
JadeCara, line 110:
```suggestion
```

galvana (approved):
Thanks for making the changes! I don't want to hold this one up any further, so let's make sure we ticket adding an index to the attachment table as well 👍

JadeCara:
> Thanks for making the changes! I don't want to hold this one up any further, so let's make sure we ticket adding an index to the attachment table as well 👍 

I actually added it [here](https://github.com/ethyca/fides/blob/728b4c97c380b9aeed0b8e6d0d8bd71adf4d456c/src/fides/api/alembic/migrations/versions/6ea2171c544f_change_attachment_storage_key_to_.py#L39-L51) 👍

PR #5845, opened 2025-03-05
Fides String Support for GPP

gilluminate:
Closes [LJ-296]

### Description Of Changes

Fides.js must support a GPP string as a new section in the existing fides_string, so that GPP can be passed back and forth between the server and Fides.js and used to initialize the CMP with existing preferences.

### Code Changes

* Added support for GPP string handling in fides_string (now contains TC_STRING, AC_STRING, GPP_STRING)
* Added functions for decoding and working with GPP strings
* fides_string to consent Supports US National, US state-specific, and TCF GPP sections
* Improved error handling and debugging
* Added comprehensive both unit and Cypress test cases for fides_string to consent functionality
* Updated developer docs with new fides_string format documentation including examples for different string formats.

### Steps to Confirm

1. In Admin UI set up US Modal and TCF experiences
2. In the demo page, manually save various opt in and opt out, noting the Fides String for each by using the `window.Fides.fides_string` value and/or cookie. Fides.js should store GPP as part of the Fides string in a new section after the TCF and AC strings (Example: fides_string: ,,DBABLA~BVAUAAAAAWA.QA) The GPP string should be placed at a third index after TCF and AC, even if there are no values for those strings.
3. Revisit the demo page with the `fides_string=...` query string, and use each GPP string noted above as the value
4. Ensure that
  * ...the banner/modal never shows when the fides_string is passed to the _correct_ experience (including for TCF)
  * ...the preferences are identical between the manual and automatic. Open the modal using the modal link after automatic has run to ensure the same preferences were saved (toggle switches match)
  * ...the cookies are the same
  * ...the payload of the `api/v1/privacy-preferences` endpoint is the same.

### Pre-Merge Checklist

* [x] Issue requirements met
* [x] `CHANGELOG.md` updated
* Followup issues:
  * [x] No followup issues
* Database migrations:
  * [x] No migrations
* Documentation:
  * [x] If there are any new client scopes created as part of the pull request, remember to update public-facing documentation that references our scope registry

[LJ-296]: https://ethyca.atlassian.net/browse/LJ-296?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


eastandwestwind:
Are there corresponding docs changes or a ticket to capture this?

clients/fides-js/rollup.config.mjs
```diff
@@ -24,8 +24,8 @@ const GZIP_SIZE_HEADLESS_ERROR_KB = 25;
 const GZIP_SIZE_HEADLESS_WARN_KB = 20;
 
 // GPP
-const GZIP_SIZE_GPP_ERROR_KB = 25;
-const GZIP_SIZE_GPP_WARN_KB = 15;
+const GZIP_SIZE_GPP_ERROR_KB = 40;
```
eastandwestwind, line 27: Confirming this size bump via client-services... stay tuned

gilluminate:
> Are there corresponding docs changes or a ticket to capture this?

Doc updates are in this same PR

gilluminate:
As per my latest commit, I've been thinking about it and I have to agree with you on the region -> API mapping problem. I'm reverting that portion of the code but keeping the other part that relies on it so that it's still the source of truth for all Fides supported regions (only need to update this one place).

clients/fides-js/src/lib/initialize.ts
```diff
@@ -394,17 +393,20 @@ export const initialize = async ({
...
-        experience: fides.experience,
-        isExperienceClientSideFetched: fetchedClientSideExperience,
-      });
-      fidesDebugger(
-        "Updated experience from saved preferences",
-        updatedExperience,
-      );
-      // eslint-disable-next-line no-param-reassign
-      fides.experience = { ...fides.experience, ...updatedExperience };
+      if (fetchedClientSideExperience) {
```
gilluminate, line 396: This check happens here now, that's why I removed passing `isExperienceClientSideFetched` everywhere. It's essentially the same outcome but cleaner.

clients/fides-js/src/lib/initialize.ts
```diff
@@ -442,6 +444,8 @@ export const initialize = async ({
         options,
         overrides?.experienceTranslationOverrides,
       );
+      // eslint-disable-next-line no-param-reassign
+      fides.locale = i18n.locale || DEFAULT_LOCALE;
```
gilluminate, line 448: had to add this to the window.Fides object so that the GPP extension could have access to it, since it has no i18n capabilities on its own (just event handling). It's needed in the string -> consent process to get the correct history ID for the privacy-consent payload.

clients/fides-js/src/lib/tcf.ts
```diff
@@ -28,26 +29,35 @@ const FORBIDDEN_LEGITIMATE_INTEREST_PURPOSE_IDS = [1, 3, 4, 5, 6];
...
-  const vendorIds = uniqueIds.sort((a, b) => Number(a) - Number(b)).join(".");
-
+  const vendorIds = [
+    ...tcStringPreferences.vendorsConsent,
+    ...tcStringPreferences.vendorsLegint,
+  ]
+    .filter(vendorIsAc)
+    .map((id) => decodeVendorId(id).id)
+    .sort((a, b) => Number(a) - Number(b))
+    .join(".");
```
gilluminate, line 60: a little AI assisted clean up here. No change in functionality. :)

clients/fides-js/src/lib/fidesString.ts
gilluminate, line 1: This file was actually moved from `clients/fides-js/src/lib/tcf/fidesString.ts` and updated to include GPP. GitHub is making it look like this is all brand new.

clients/privacy-center/cypress/e2e/consent-banner-gpp.cy.ts
```diff
@@ -57,6 +58,24 @@ describe("Fides-js GPP extension", () => {
     }).as("patchNoticesServed");
   });
 
+  describe("with GPP forced", () => {
```
gilluminate, line 61: moved up from below just because it's more "global" and felt like it belonged here with the other "global" tests.

clients/privacy-center/public/fides-js-components-demo.html
```diff
@@ -312,7 +312,11 @@ <h2>Property id</h2>
         console.log("Fides has been initialized!");
         // Pretty-print the fides consent object and add it to the page.
         document.getElementById("consent-json").textContent = JSON.stringify(
-          Fides.consent,
+          Fides.experience &&
+            Fides.experience.experience_config &&
+            Fides.experience.experience_config.component === "tcf_overlay"
+            ? Fides.tcf_consent
+            : Fides.consent,
```
gilluminate, line 319: This stuff might not be needed. It relates to my question on our call about what the `tcf_consent` string is. Doesn't hurt. But probably doesn't help either.

clients/privacy-center/public/fides-js-demo.html
```diff
@@ -17,8 +17,11 @@
         window.render_delay = 0; // increment the delay to test/mimic a slow loading or dynamic page
         const params = new URLSearchParams(window.location.search);
         const geolocation = params.get("geolocation");
+        const gpp = params.get("gpp");
         const src = `/fides.js${!!params.size ? `?${params.toString()}` : ""}`;
-
+        if (gpp === "debug") {
+          document.write('<script src="/fides-ext-gpp.js"><\/script>');
+        }
```
gilluminate, line 24: when `gpp=debug` we're loading the js file instead of including it as part of the fides.js payload. That way we can benefit from the SourceMaps :)

clients/fides-js/src/lib/gpp/string-to-consent.ts
```diff
@@ -0,0 +1,308 @@
...
+  } = experience;
+  const usApproach = gppSettings?.us_approach;
+  let gppRegion = deriveGppFieldRegion({
+    experienceRegion,
+    usApproach,
+  });
+  let gppSection = FIDES_US_REGION_TO_GPP_SECTION[gppRegion];
+
+  if (!gppSection && usApproach === GPPUSApproach.ALL) {
+    // if we're using the all approach, and the current state isn't supported, we should default to national
```
eastandwestwind, line 59: could be helpful to have a debug line here

clients/fides-js/src/lib/gpp/string-to-consent.ts
```diff
@@ -0,0 +1,308 @@
...
+    gppRegion = US_NATIONAL_REGION;
+    gppSection = FIDES_US_REGION_TO_GPP_SECTION[gppRegion];
+  }
+
+  if (
+    !gppSection ||
+    (gppRegion === US_NATIONAL_REGION && usApproach === GPPUSApproach.STATE)
+  ) {
+    // If we don't have a section, we can't get anything.
+    // If we're using the state approach, we shouldn't use the national section, even if region is set to national.
```
eastandwestwind, line 69: same as above with debug line- in addition I think I'm confused by the comment and overall logic for this particular statement. Could you help me understand?

clients/fides-js/src/lib/gpp/string-to-consent.ts
```diff
@@ -0,0 +1,308 @@
...
+    return consent;
+  }
+
+  notices.forEach((notice) => {
```
eastandwestwind, line 73: At this point, `gppSection` could still be undefined right? What do we want to do in this case?

clients/fides-js/src/lib/fidesString.ts
```diff
@@ -0,0 +1,82 @@
...
+ * // Returns { tc: "CPzvOIA.IAAA", ac: "", gpp: "" }
+ *
+ * // GPP string only (with empty TC and AC)
+ * decodeFidesString(",,DBABLA~BVAUAAAAAWA.QA")
+ * // Returns { tc: "", ac: "", gpp: "DBABLA~BVAUAAAAAWA.QA" }
+ *
+ * @param fidesString - The combined Fides string to decode
+ * @returns An object containing the decoded TC, AC, and GPP strings
+ */
+export const decodeFidesString = (fidesString: string) => {
```
eastandwestwind, line 39: Could we add return typing and formalize this? Then we can enforce on each usage of `decodeFidesString`

clients/fides-js/src/lib/gpp/string-to-consent.ts
```diff
@@ -0,0 +1,308 @@
...
+
+        // Check if the value matches any of the defined states
+        if (valueStr === gppMechanism.opt_out) {
+          return true; // User has opted out
+        }
+        if (valueStr === gppMechanism.not_opt_out) {
+          return false; // User has not opted out
+        }
+        // If value matches not_available or is undefined, we'll skip this mechanism
+        return undefined;
```
eastandwestwind, line 100: We're in a `some` fn which only looks at true / false, otherwise converts `undefined` to `false`- Just making sure we want to use the implicit `undefined` here?

clients/fides-js/src/lib/gpp/string-to-consent.ts
```diff
@@ -0,0 +1,308 @@
...
+      null,
+  }));
+
+  let updateCookie: (oldCookie: FidesCookie) => Promise<FidesCookie>;
+  let tcf: TcfSavePreferences | undefined;
+  let consentPreferencesToSave: SaveConsentPreference[];
+
+  if (!isTCF) {
+    // Handle non-TCF case
+    const consent = getConsentFromGppCmpApi({
```
eastandwestwind, line 240: can we add a type annotation here?

clients/fides-js/src/lib/gpp/string-to-consent.ts
```diff
@@ -0,0 +1,308 @@
...
+      cmpApi,
+      experience: experience as PrivacyExperience,
+    });
+
+    const enabledPrivacyNoticeKeys = Object.entries(consent)
```
eastandwestwind, line 245: same as above- add type annotation

clients/fides-js/src/lib/gpp/string-to-consent.ts
```diff
@@ -0,0 +1,308 @@
...
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      .filter(([_, value]) => value)
+      .map(([key]) => key);
+
+    consentPreferencesToSave = createConsentPreferencesToSave(
```
eastandwestwind, line 250: type annotation

clients/fides-js/src/lib/preferences.ts
```diff
@@ -67,6 +69,19 @@ async function savePreferencesApi(
  * 6. Remove any cookies from notices that were opted-out from the browser
  * 7. Dispatch a "FidesUpdated" event
  */
+export interface UpdateConsentPreferences {
```
eastandwestwind, line 72: 💯

clients/fides-js/src/lib/tcf.ts
```diff
@@ -28,26 +29,35 @@ const FORBIDDEN_LEGITIMATE_INTEREST_PURPOSE_IDS = [1, 3, 4, 5, 6];
...
+ * filters for AC vendors only (those with 'gacp' prefix), extracts their numeric IDs,
+ * and joins them in ascending order.
+ *
+ * @example
+ * // Given vendors ["gacp.42", "gacp.33", "gvl.12", "gacp.49"]
+ * generateAcString({tcStringPreferences: {
+ *   vendorsConsent: ["gacp.42"],
+ *   vendorsLegint: ["gacp.33", "gvl.12", "gacp.49"]
+ * }})
+ * // Returns "1~33.42.49"
```
eastandwestwind, line 46: really nice docstring!

clients/fides-js/src/lib/tcf.ts
```diff
@@ -187,6 +199,14 @@ export const initializeTcfCmpApi = () => {
...
   // see https://github.com/InteractiveAdvertisingBureau/iabtcf-es/tree/master/modules/cmpapi#dont-show-ui--tc-string-does-not-need-an-update
   window.addEventListener("FidesInitialized", (event) => {
+    // Set default GPP string if GPP is enabled. This will get updated by the GPP extension
+    if (window.Fides.experience?.gpp_settings?.enabled) {
+      window.Fides.fides_string =
+        window.Fides.fides_string ||
+        window.Fides.cookie?.fides_string ||
+        formatFidesStringWithGpp();
+    }
+
```
eastandwestwind, line 209: it's assumed that the fides_string will already have the `DEFAULT_GPP_STRING`? 

How about for users that have already visited a page with fides and have an existing fides_string where we haven't yet added the default GPP str from this PR?

eastandwestwind (changes requested):
Wow, nice work @gilluminate ! Lots of moving parts here, but the key changes look solid.

My comments / Qs above mostly deal with code readability / possible err cases

clients/fides-js/src/lib/gpp/decoder.ts
```diff
@@ -0,0 +1,103 @@
+import { CmpApi, GppModel } from "@iabgpp/cmpapi";
+
+import { ETHYCA_CMP_ID } from "../tcf/constants";
+import { CMP_VERSION } from "./constants";
+
+/**
```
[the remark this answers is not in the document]
gilluminate: new ticket here: https://ethyca.atlassian.net/browse/LJ-476

clients/fides-js/src/lib/gpp/string-to-consent.ts
gilluminate: We `return consent;` in the line above so we don't reach this spot in that case.

clients/fides-js/src/lib/gpp/string-to-consent.ts
gilluminate: Good catch!

clients/fides-js/src/lib/gpp/string-to-consent.ts
gilluminate: it gets inferred well here but I'll add an explicit if that's helpful

clients/fides-js/src/lib/gpp/string-to-consent.ts
gilluminate: I generally prefer inferred typing as appropriate. If the TS knows what type it is I find it redundant to tell it again.

clients/fides-js/src/lib/gpp/string-to-consent.ts
gilluminate: I generally prefer inferred typing as appropriate. If the TS knows what type it is I find it redundant to tell it again.

clients/fides-js/src/lib/gpp/decoder.ts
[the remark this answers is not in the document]
gilluminate: Note added to the JSDoc

clients/fides-js/src/lib/tcf.ts
gilluminate: Great call out! I always felt a little strange to do this and I think you've nailed down why. Removed this and updated the `formatFidesStringWithGpp` method to require `cmpapi` and no longer sets a default.

clients/fides-js/rollup.config.mjs
gilluminate: any lingering concern here, or are we ok with the increase?

clients/fides-js/src/lib/gpp/string-to-consent.ts
```diff
@@ -0,0 +1,308 @@
...
+  preferences.customPurposesConsent = [];
+
+  return preferences;
+};
+
+interface FidesStringToConsentArgs {
+  fidesString: string;
+  cmpApi: CmpApi;
+}
+export const fidesStringToConsent = ({
```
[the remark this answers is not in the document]
gilluminate: added detailed JSDocs

PR #5853, opened 2025-03-07
HA-481: improve CORS origin handling

adamsachs:
Closes HA-481

### Description Of Changes

Improvements to CORS origin setting handling:
- only use a single validation code path for validating values set via 'traditional' config and set via API config (config proxy)
- allow `*` as a special-case origin URL
- do not rely on `validators` package for any URL validation
    - this has the side effect of allowing the `0.0.0.0` IP address as a valid origin, e.g. `http://0.0.0.0:8000`. this is a nonstandard CORS origin, but it's not something we think should be _disallowed_ for any particular reason

### Code Changes

### Steps to Confirm

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:

### DIFF
--- a/src/fides/api/custom_types.py
+++ b/src/fides/api/custom_types.py
@@ -2,10 +2,18 @@
 """Logic related to sanitizing and validating user application input."""
 from html import escape
 from re import compile as regex
-from typing import Annotated
+from typing import Annotated, Any
 
 from nh3 import clean
-from pydantic import AfterValidator, AnyHttpUrl, AnyUrl, BeforeValidator
+from pydantic import (
+    AfterValidator,
+    AnyHttpUrl,
+    AnyUrl,
+    BeforeValidator,
+    ValidationInfo,
+    ValidatorFunctionWrapHandler,
+    WrapValidator,
+)
 
 from fides.api.util.unsafe_file_util import verify_css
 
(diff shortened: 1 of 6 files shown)


src/fides/api/custom_types.py
```diff
@@ -116,6 +116,20 @@ def validate_path_of_url(value: AnyUrl) -> str:
...
+    rejected by the AnyUrl validator.
+
+    Note that `*` _is_ considered a valid origin value
+    (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin),
+    but we specifically disallow it to be set on the `cors_origins` security setting
+    in our application. This is because non-owner users (contributors)
+    are able to edit the `cors_origins` security setting (via API), and we do not
+    want to allow them to set a wildcard origin.
+
+    Instead, `*` can be set via the `cors_origin_regex` security setting.
```
tvandort, line 131:
```suggestion
    Instead, `.*` can be set via the `cors_origin_regex` security setting as an equivalent to `"*"`. There is an behavioral difference in that the `Access-Control-Allow-Origin` in the response will be set to the `Access-Control-Allow-Origin` in the request rather than `"*"`.
```

tvandort (changes requested):

src/fides/api/custom_types.py
adamsachs: thanks for the catch! put in something inspired by your suggestion in https://github.com/ethyca/fides/pull/5853/commits/e92d25369defb8b26537e2ca47b92a3c038107a8

tvandort (approved):

PR #5858, opened 2025-03-08
Fixed load_samples to wrap variables in quotes to prevent YAML parsing errors

NevilleS:
### Description Of Changes

Unfortunately the variable expansion logic in the `FIDES__DATABASE__LOAD_SAMPLES` handling can result in loading a character like `&`, which YAML parsing then tries to interpret as an anchor, and then it fails to load the sample data.

Fides' startup routine is resilient to that failure, it just continues starting up but aborts loading the rest of the sample data, which is a subtle footgun and caught us in a test environment recently!

This addresses this by wrapping all the variables we load from the template connections in quotes, and adds some tests to exercise that.

### Code Changes

* Update sample connections YML to wrap variables in quotes
* Add test to ensure that sample connections YML loads safely with special characters in variables
* Add test to demonstrate what will/won't fail to load, for future engineers

### Steps to Confirm

1.  Run unit tests and confirm
2. Use `FIDES__DATABASE__LOAD_SAMPLES=1` and insert ENV variables like `FIDES_DEPLOY__CONNECTORS__POSTGRES__HOST=&123` and confirm that YAML errors do not occur

### Pre-Merge Checklist

* [X] Issue requirements met
* [X] All CI pipelines succeeded
* [X] `CHANGELOG.md` updated
* Followup issues:
  * [x] No followup issues
* Database migrations:
  * [x] No migrations
* Documentation:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@ Changes can also be flagged with a GitHub label for tracking purposes. The URL o
 
 ### Fixed
 - Fixed pagination bugs on some tables [#5819](https://github.com/ethyca/fides/pull/5819)
+- Fixed load_samples to wrap variables in quotes to prevent YAML parsing errors [#5857](https://github.com/ethyca/fides/pull/5857)
 
 ## [2.56.2](https://github.com/ethyca/fides/compare/2.56.1...2.56.2)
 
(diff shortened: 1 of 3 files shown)


adamsachs (approved):
👍 thanks for the good test coverage!

PR #5859, opened 2025-03-08
LJ 464: Update all FidesJS events with `timestamp`, and all `FidesUIChanged` events with `extraDetails.trigger` and `extraDetails.preference`

NevilleS:
Closes [LJ-464]

### Description Of Changes

This makes a few changes to our FidesJS SDK in an attempt to make the `FidesEvents` we trigger more rich and - where possible - more consistent in the properties they include. Still lots of gaps though.

The main changes we've done here are:
* Add a new, dedicated spec that exercises most of the FidesEvent types and allows us to clearly see their format and write tests against them
* Add a `timestamp` property to all events, using `performance.mark()` as the clock, so we can see timing data on the events for analytics
* Add richer metadata to the `FidesUIChanged` event in two ways:
  * `extraDetails.trigger`: this includes the type & label of the element that triggered the UI change. In our UI today this will always be a toggle, so this property looks something like  `{ type: "toggle", label: "Analytics" }`
  * `extraDetails.preference` this includes the preference being changed, which can be many types: notices, TCF purposes, TCF vendors, etc. We also include some detail about the preference for reporting, which looks like this: `{ type: "notice", key: "analytics" }` 

### Code Changes

* Added `consent-events.cy.ts` with tests for both banner & modal and TCF component types
* Rewired the TcfOverlay and NoticeOverlay components to collect trigger & preference details on changes and fire those along with the `FidesUIChanged` event
* Rewired all the child components to pass around the context about toggle changes on the various `onChange`, `onToggle`, etc. callbacks
* Regenerated the FidesJS SDK docs with the new event properties
* Some minor quality-of-life refactoring along the way...

### Steps to Confirm

1. Launch the Cookie House demo app and subscribe to all the `FidesEvents` on the window
2. Interact with the banner, modal, and TCF components to trigger events
3. Inspect the event details to see the `timestamp` property
4. Inspect the event details on `FidesUIChanged` events to see the `trigger` and `preference` properties
5. Ensure that the `preference` type, in particular, matches the type and metadata of the toggle you trigger:
  a. Notices
  b. TCF Purposes (Consent)
  c. TCF Purposes (Legitimate Interest)
  d. TCF Special Purposes
  e. TCF Vendors (Consent)
  f. TCF Vendors (Legitimate Interest)

### Pre-Merge Checklist

* [x] Issue requirements met
* [x] `CHANGELOG.md` updated
* Followup issues:
  * [x] Followup issues created: [LJ-478] to further clean-up the events to make them more consistent
  * [ ] No followup issues
* Database migrations:
  * [x] No migrations
* Documentation:
  * [x] Documentation complete, [PR opened in fidesdocs](https://github.com/ethyca/fidesdocs/pulls)

[LJ-464]: https://ethyca.atlassian.net/browse/LJ-464?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[LJ-478]: https://ethyca.atlassian.net/browse/LJ-478?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


clients/fides-js/src/docs/fides-event.ts
```diff
@@ -122,12 +130,30 @@ export interface FidesEvent extends CustomEvent {
       consentMethod?: "accept" | "reject" | "save" | "dismiss" | "gpc";
 
       /**
-       * What toggle (if any) caused this event.
+       * What UI element (if any) triggered this event.
```
tvandort, line 133:
```typescript
/**
       * What UI element (if any) triggered this event.
       */
      trigger?:
        | {
            type: "toggle" /**
             * The checked state of the element that triggered the event, when type
             * is `toggle`.
             */;
            checked: boolean;
          }
        | {
            /**
             * The type of element that triggered the event. Additional types may be
             * added over time, so expect this type to grow.
             */
            type: "button" | string;

            /**
             * The UI label of the element that triggered the event.
             */
            label?: string;

            /**
             * The id of the element that triggered the event.
             */
            id: string;
          };
```

Out of curiosity what does the docs generation think about types defined like this?

NevilleS: Yeah, I like that union type, but the generated docs aren't great, it just shows `object | object`:
```
#### extraDetails.trigger?

> `optional` **trigger**: `object` \| `object`

What UI element (if any) triggered this event.
```

After a small Google, looks like better union type inlining is added in newer version of `typedoc` via an `@expand` annotation that'll inline those union type definitions ([link](https://github.com/TypeStrong/typedoc/issues/2303)), but I'm reticent to upgrade as there are breaking changes in 0.27.0+ and it would require bumping our minimum TS version.

clients/fides-js/src/docs/fides-event.ts
```diff
@@ -122,12 +130,30 @@ export interface FidesEvent extends CustomEvent {
...
        */
-      servingToggle?: {
-        label: string;
+      trigger?: {
+        /**
+         * The type of element that triggered the event. Additional types may be
+         * added over time, so expect this type to grow.
+         */
+        type: "toggle" | "button" | string;
```
gilluminate, line 140: I would suggest keeping this strict until additional types are needed and are explicitly added to the type.
```suggestion
        type: "toggle" | "button";
```

clients/privacy-center/cypress/e2e/consent-events.cy.ts
```diff
@@ -0,0 +1,571 @@
...
+          type: "FidesUpdating",
+          detail: {
+            extraDetails: {
+              // servingComponent: "modal",
+              consentMethod: "save",
+              // trigger: {
+              //   type: "button",
+              //   label: "Save",
+              //   id: "save-preferences",
+              // },
```
gilluminate, line 162: can we clean up these commented lines? seeing this in several places in this test. Are we planning to bring these back in at some point?

clients/fides-js/src/docs/fides-event.ts
NevilleS: The reason I did this is because that might be considered a breaking change to code that uses this today. Does that make sense?

clients/privacy-center/cypress/e2e/consent-events.cy.ts
NevilleS: So, I'd like to have the events include these properties. However, today our Updated/Updating events are completely disjointed from the buttons that trigger them, so it was really gnarly to try to include these.

In other words the tests only pass if these are commented out, but that's because our events are currently pretty inconsistent.

I don't think I'll have time to fix this tech debt quite yet, I tried for a bit and got nervous about the changes!

clients/fides-js/src/docs/fides-event.ts
gilluminate: what code uses it today? Isn't this net new to our system? Regardless it's ok I guess for documentation, but not for the benefit of TS. TS will just allow string now and there's no real reason to include the first 2 stricter checks.

clients/privacy-center/cypress/e2e/consent-events.cy.ts
gilluminate: Maybe just a follow up ticket with a comment here then?

clients/fides-js/src/docs/fides-event.ts
NevilleS: OK, agreed! I narrowed this down to the only trigger type currently defined: `"toggle"`.

clients/privacy-center/cypress/e2e/consent-events.cy.ts
NevilleS: Done!

clients/fides-js/src/components/tcf/TcfPurposes.tsx
```diff
@@ -155,6 +160,57 @@ const TcfPurposes = ({
...
+    triggerDetails: FidesEventDetailsTrigger,
+  ) => {
+    // Determine the preference being changed based on the model type:
+    // - customPurposesConsent -> notice
+    // - purposesConsent -> tcf_purpose_consent
+    // - purposesLegint/specialPurposes -> tcf_purpose_legitimate_interest
+    //
+    // NOTE: TCF purposes don't have equivalent notice "keys" today, but we
+    // expect to add those in the future. When we do, we'll prefix like this:
+    // "tcf_purpose_consent_1", "tcf_purpose_legitimate_interest_2", etc.
```
NevilleS, line 187: Note this! This is the first place we've exposed a TCF purpose id in our FidesJS SDK and I didn't want to ship vanilla numbers like `1` or `2` because they lack context. Therefore I've prefixed them here with the preference type so you can separate between purpose consents, special features, etc.

clients/fides-js/src/lib/events.ts
```diff
@@ -59,19 +74,26 @@ export const dispatchFidesEvent = (
...
   if (typeof window !== "undefined" && typeof CustomEvent !== "undefined") {
     // Extracts consentMethod directly from the cookie instead of having to pass in duplicate data to this method
     const constructedExtraDetails: FidesEventExtraDetails = {
-      consentMethod: cookie?.fides_meta.consentMethod,
+      consentMethod: cookie?.fides_meta
+        .consentMethod as FidesEventExtraDetails["consentMethod"],
       ...extraDetails,
     };
+    const perfMark = performance?.mark?.(type);
+    const timestamp = perfMark?.startTime;
```
NevilleS, line 87: This was easy to add 👍

clients/fides-js/docs/interfaces/FidesEvent.md
```diff
@@ -118,23 +127,100 @@ Whether the user should be shown the consent experience. Only available on Fides
 
 What consent method (if any) caused this event.
 
-#### extraDetails.servingToggle?
+#### extraDetails.trigger?
```
NevilleS, line 130: NOTE: I renamed this from servingToggle -> trigger. We haven't released servingToggle yet, so this doesn't affect any public consumers.

I did this to make the `trigger` property more future-proof: we definitely won't _only_ have toggles triggering events.

clients/privacy-center/cypress/e2e/consent-events.cy.ts
```diff
@@ -0,0 +1,640 @@
...
+                type: "toggle",
+                label: "Advertising",
+                checked: true,
+              },
+              preference: {
+                key: "advertising",
+                type: "notice",
+              },
+            },
+          },
```
NevilleS, line 116: See? This event is rich and useful, it has both the UI context from the `trigger` and the business-logic context from the `preference`.

clients/privacy-center/cypress/e2e/consent-events.cy.ts
```diff
@@ -0,0 +1,640 @@
...
+          type: "FidesUpdating",
+          detail: {
+            extraDetails: {
+              consentMethod: "save",
+              // DEFER(LJ-478): Restore these checks after updating events to be more consistent
+              // servingComponent: "modal",
+              // trigger: {
+              //   type: "button",
+              //   label: "Save",
+              // },
```
NevilleS, line 171: Note that our `FidesUpdating` and `FidesUpdated` events don't include `servingComponent` or any other metadata about where/how the update was triggered.

I looked into adding this context, but it was _very_ intimidating to do because of how separated the "preference update" code is from the UI hierarchy.

clients/privacy-center/cypress/e2e/consent-events.cy.ts
```diff
@@ -0,0 +1,640 @@
+import type { FidesEventDetail, FidesEventType } from "fides-js/src/lib/events";
+
+import { stubConfig, stubTCFExperience } from "../support/stubs";
+
+interface FidesEventExpectation {
+  type: FidesEventType;
+  detail: Partial<FidesEventDetail>;
+}
+
+describe("Consent FidesEvents", () => {
```
NevilleS, line 10: New spec file that exercises both banner+modal and TCF components, triggering all the FidesEvents along the way.

It's hard to reason about this behaviour from the other specs that are much more focused on the core logic of those components instead of the events they trigger. There's some duplicate tests now, but more coverage isn't necessarily bad.

NevilleS (commented):
Self-review!

clients/fides-js/src/components/DataUseToggle.tsx
```diff
@@ -21,10 +21,7 @@ const DataUseToggle = ({
   noticeKey: string;
   title: string;
   checked: boolean;
-  onToggle: (
-    noticeKey: string,
-    toggleDetails: FidesServingToggleDetails,
-  ) => void;
+  onToggle: (noticeKey: string, eventTrigger: FidesEventDetailsTrigger) => void;
```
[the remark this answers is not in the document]
gilluminate: makes sense.

clients/fides-js/src/docs/fides-event.ts
```diff
@@ -102,6 +102,14 @@ export interface FidesEvent extends CustomEvent {
      */
     fides_string?: string;
 
+    /**
+     * High-precision timestamp from {@link https://developer.mozilla.org/en-US/docs/Web/API/Performance/mark performance.mark()}
+     * representing when this event was created. The timestamp is measured in milliseconds since page load.
+     *
+     * May be undefined if the Performance API is not available.
+     */
+    timestamp?: number;
```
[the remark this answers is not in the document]
NevilleS: 👏 👏 👏 

This is why we code review.

gilluminate (approved):
Looks great

PR #5868, opened 2025-03-11
LJ-483 Update hover menu state

lucanovera:
### Description Of Changes
Update the hover state of the sub items in the menu to have a neutral-800 background and neutral-200 text color.

### Code Changes
* Added scss rule to use neutral 800 color upon hovering dark subitems
* Removed scss rule that changed the hover color to full-white (the result is that it keeps the neutral-200 color now)

### Steps to Confirm

1.  Login to admin ui
2. Hover over the menu items
3. Check the background is neutral-800 and the text color stays with the neutral-200 color
4. Check hover state doesn't apply to group level items, only to subitems in the menu

### Screenshots
Before: 
Captura de pantalla 2025-03-11 a la(s) 10 35 46 a  m

After:
Captura de pantalla 2025-03-11 a la(s) 10 43 18 a  m

### Pre-Merge Checklist

* [x] Issue requirements met
* [x] `CHANGELOG.md` updated
* Followup issues:
  * [x] No followup issues
* Database migrations:
  * [x] No migrations
* Documentation:

### DIFF
--- a/.fides/db_dataset.yml
+++ b/.fides/db_dataset.yml
@@ -143,6 +143,46 @@ dataset:
     - name: user_id
       data_categories:
       - user.unique_id
+  - name: comment
+    fields:
+    - name: comment_text
+      data_categories:
+      - system.operations
+    - name: comment_type
+      data_categories:
+      - system.operations
+    - name: created_at
+      data_categories:
+      - system.operations
+    - name: id
+      data_categories:
+      - system.operations
+    - name: updated_at
+      data_categories:
+      - system.operations
+    - name: user_id
+      data_categories:
+      - system.operations
+  - name: comment_reference
+    fields:
+    - name: comment_id
+      data_categories:
+      - system.operations
+    - name: created_at
+      data_categories:
+      - system.operations
+    - name: id
+      data_categories:
+      - system.operations
+    - name: reference_id
+      data_categories:
+      - system.operations
+    - name: reference_type
+      data_categories:
+      - system.operations
+    - name: updated_at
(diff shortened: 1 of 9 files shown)


clients/admin-ui/src/theme/global.scss
```diff
@@ -44,18 +44,21 @@ h6 {
...
+    > .ant-menu-title-content a {
+      color: var(--fidesui-neutral-200);
+
+      &:focus-visible {
+        text-decoration: underline;
+      }
+    }
 
     &:hover,
     &:focus {
```
gilluminate, line 57: This `&:focus` will never apply because the `a` element is focusable but the menu item is not.
```suggestion
    &:focus-within {
```

gilluminate (changes requested):

clients/admin-ui/src/theme/global.scss
gilluminate: Playing around with this, it still doesn't seem to work, probably because Ant's stuff is overriding. Might just want to remove this :focus altogether in favor of the underline

lucanovera: yup, I think it's because I'm overriding --ant-menu-dark-item-hover-bg which doesn't apply if it's not hovering. I will try setting the background color directly instead.

lucanovera: @gilluminate setting the background-color property worked for both hovering and focusing with keyboard.

gilluminate (approved):

PR #5871, opened 2025-03-11
Split Cypress workflows to relevant PR changes

gilluminate:
### Description Of Changes

* Pulls Cypress jobs from general frontend checks workflow
* Creates separate workflow for respective Cypress jobs so that they only run when PR contains changes in the appropriate directories (pull_request.paths)

The goal is to reduce the need to wait for all Cypress tests to complete on a PR where only one codebase or the other has been changed. (Note: All tests still run after merge to main/release branches.)


lucanovera (approved):
This is a great first step into speeding up our tests. Approved!

PR #5886, opened 2025-03-13
fix vendor override handling

gilluminate:
Closes [<issue>]

### Description Of Changes

Some GVL vendors cannot be consented to in the TCF CMP. When encoding a tc string for vendors that only declare LI as their purposes we also need to add a publisher restriction when that vendor is set with a "Consent Only" override.

### Code Changes

Make the same checks that the TCF framework does when looking at the overrides: 
  * vendor is for purpose
  * vendor's purpose list is empty []
  * flexible purposes are present
When all 3 of those are true, we just need to also add Publisher Restrictions for those flexible purposes to indicate that the publisher has chosen to override the LI as consent only.

### Steps to Confirm

1. Setup a TCF experience with at least DoubleVerify added to your vendors.
2. In the Consent settings page, enable "Override vendor purposes"
3. For each purpose with the option to use consent instead of Legitimate Interest, toggle the Consent "on" and click Save at the bottom.
4. You may need to refresh the experience cache at this point
5. Load the demo page with your TCF experience
6. Open the modal and look at the Vendors tab. DoubleVerify should be in the Purposes list and not in the Legitimate Interests list because of the consent override
7. Click Accept All
8. Reload the page without clearing cookies so that the consent gets loaded from the TCF string stored in the cookie
9. Open the modal and double verify that DoubleVerify (see what I did there?) is toggled "On" (before this fix it would be in an "off" state)

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:


eastandwestwind (changes requested):
Do we have a follow-up ticket for a test? Also could you add a Changelog entry?

eastandwestwind (approved):

PR #5888, opened 2025-03-13
Add has next to strategy cursor configuration

Linker44:
### Description Of Changes

For the aftership integration we face a scenario where the cursor will remain present and a has_next_page bool value has to be evaluated to verify if pagination should continue. Similar to what happend in [segment unify pagination](https://github.com/ethyca/fides/pull/5596)

### Code Changes

* edit cursor pagination strategy to support has_next value

### Steps to Confirm

1.  verify tests are running

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:


src/fides/api/service/pagination/pagination_strategy_cursor.py
```diff
@@ -27,11 +29,15 @@ def get_next_request(
...
         )
         if object_list and isinstance(object_list, list):
-            cursor = pydash.get(object_list.pop(), self.field)
+            last_object = object_list.pop()
+            cursor = pydash.get(last_object, self.field)
+            if self.has_next:
+                has_next = pydash.get(last_object, self.has_next)
 
         # If the cursor value isn't found, try to find the value in response.json() using the field value
         if cursor is None:
```
Vagoasdf, line 43: So here we have two possible situations: 
1. The  cursor is on the last object of the response, self contained inside its field, and we can grab it with the previous if/else 

2. The cursor is somewhere else inside the response JSON, and we have to get it directly. 

The part that checks for  the existance of has_next however is divorced and separate from this code fragment. 

I Believe it should be here for readability, before the return since it's related to this logic block

tests/ops/service/pagination/test_pagination_strategy_cursor.py
```diff
@@ -36,6 +36,36 @@ def response_with_empty_list():
...
 
 
+@pytest.fixture(scope="function")
+def response_with_has_next_true():
+    response = Response()
+    response._content = bytes(
+        json.dumps(
+            {
+                "conversations": [{"id": 1}, {"id": 2}, {"id": 3}],
+                "meta": {"next_cursor": 123, "has_next": "true"},
```
Vagoasdf, line 46: Nice tests! 🙌🏽 

We could add the specific case when the values are not in the last "meta" object (AKA the one in my previous comment) for completion

Vagoasdf (changes requested):
Looking good! 

There are just some minor improvements to suggest

Vagoasdf (approved):

src/fides/api/service/pagination/pagination_strategy_cursor.py
Linker44: The idea here is that since has next could not be present. The most important thing to check for is the cursor, if there is a cursor we can evaluate if has next exists, but if there is no cursor we just return early.

src/fides/api/service/pagination/pagination_strategy_cursor.py
```diff
@@ -41,6 +47,14 @@ def get_next_request(
...
             return None
 
+        if self.has_next and has_next is None:
+            has_next = pydash.get(response.json(), self.has_next)
+
+        if self.has_next:
+            logger.info(f"The {self.has_next} field has a value of {has_next}")
```
galvana, line 54: We should make this a debug statement
```suggestion
            logger.debug(f"The {self.has_next} field has a value of {has_next}")
```

PR #5915, opened 2025-03-20
HA-515 - API - Ability to pull a minimal version of a Dataset, and filter by fides_meta.namespace.connection_type

andres-torres-marroquin:
Closes HA-515

### Description Of Changes

Ability to pull a minimal version of a Dataset, and filter by `fides_meta.namespace.connection_type`.

### Code Changes

* Added a new DatasetResponse pydantic class that inherits from fideslang.Dataset, so it supports having no collections.
* Added a behavior where the Dataset API can return a minimal version of the resource, it shouldn't return `collections`, nor `tags`, nor `data_categories`.
* Added a behavior where the Dataset API objects can be filtered by `fides_meta.namespace.connection_type`.

### Steps to Confirm

1.  Interact with the API `/api/v1/dataset` with `minimal=True`, it shouldn't return `collections`, nor `tags`, nor `data_categories`.
2. Interact with the API `/api/v1/dataset` with `connection_type=<something>`, it should return only datasets with that `connection_type`.

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:


src/fides/api/api/v1/endpoints/generic_overrides.py
```diff
@@ -156,6 +159,21 @@ async def list_dataset_paginated(
 
     query = select(CtlDataset)
 
+    if minimal:
+        query = query.options(
+            load_only(
```
galvana, line 164: What does the combination of `options` and `load_only` do?

src/fides/api/api/v1/endpoints/generic_overrides.py
```diff
@@ -165,6 +183,13 @@ async def list_dataset_paginated(
         filter_params=filter_params,
     )
 
+    # If applicable, filter by connection type
+    if connection_type:
+        filtered_query = filtered_query.where(
+            CtlDataset.fides_meta["namespace"]["connection_type"].as_string()
+            == connection_type.value
+        )
```
galvana, line 191: I don't think this is a reliable query. Only some datasets have a namespace and even if they do, the connection_type might not be present. Can you remind me why we need to retrieve datasets by connection type?

src/fides/api/schemas/dataset.py
```diff
@@ -72,3 +72,14 @@ class DatasetReachability(FidesSchema):
 
     reachable: bool
     details: Optional[Union[str, List[Dict[str, Any]]]]
+
+
+class DatasetResponse(Dataset):
```
galvana, line 77: You should add a doc string indicating why we have to do this

galvana (changes requested):
Good first pass but you'll run into some issues with the query in its current state. You should write some tests for this so you can see how you need to handle the different scenarios:

- datasets without namespace
- datasets with namespace but without connection type
- datasets with namespace and connection type

src/fides/api/api/v1/endpoints/generic_overrides.py
```diff
@@ -156,6 +159,21 @@ async def list_dataset_paginated(
 
     query = select(CtlDataset)
 
+    if minimal:
+        query = query.options(  # type: ignore[attr-defined]
+            load_only(
```
galvana, line 164: You missed one @andres-torres-marroquin, what does `options`+`load_only` do?

PR #5919, opened 2025-03-20
LJ-369 be update PrivacyRequests model with comments and attachments

JadeCara:
### Closes
[LJ-369](https://ethyca.atlassian.net/browse/LJ-369?atlOrigin=eyJpIjoiY2VhODE0YjZkZWVlNGNlYTgwYzM1ZmNjZjVlZjAwYTciLCJwIjoiaiJ9)

### Description Of Changes

Added Comments and Attachments as well as some helper functions to the PrivacyRequest model.

### Code Changes

* Added comments relationship to PrivacyRequests
* Added attachments relationship to PrivacyRequests
* Added helper functions for comments and attachments to PrivacyRequests
* Added tests for those helper functions.
* Moved some fixtures from conftest.py to application fixtures since they are used in both ctl and ops tests. 
* Cleaned up and moved generic S3 storage accessors to `api/service/storage/s3` and updated imports (this was necessary because of circular imports in the original location with PrivacyRequest and Attachment)
* Moved  `LOCAL_FIDES_UPLOAD_DIRECTORY` and `get_local_filename` to fides.api.service.storage.util and updated imports (this was necessary because of circular imports in the original location with PrivacyRequest and Attachment)
* Moved the generic_s3 tests to the `tests/api/service/storage` dir
* cleaned up some elements in Attachments due to s3 movement.
* Updated the comment/attachments relationship

### Steps to Confirm

1. There should be no change to functionality. All tests should pass.

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:

[LJ-369]: https://ethyca.atlassian.net/browse/LJ-369?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/fides/api/models/attachment.py
```diff
@@ -114,13 +115,10 @@ def upload(self, attachment: bytes) -> None:
         if self.config.type == StorageType.s3:
             bucket_name = f"{self.config.details[StorageDetails.BUCKET.value]}"
             auth_method = self.config.details[StorageDetails.AUTH_METHOD.value]
-            upload_to_s3(
+            generic_upload_to_s3(
```
galvana, line 118: I think now would be a good time to make the abstraction that @erosselli  mentioned on your last PR https://github.com/ethyca/fides/pull/5812#discussion_r1989143561

I'd recommend something like this (not opinionated on the signature so I put `...`)
```
class StorageProvider(ABC):
    @abstractmethod
    def upload_file(...):

    @abstractmethod
    def retrieve_file(...):

    @abstractmethod
    def delete_file(...):
```

Then we could create an `LocalStorageProvider`, `S3StorageProvider` and others in the future. These providers would only be concerned with interacting with the storage location. You could put it all together in a new `StorageService`
```
class StorageService:
    def __init__(self, db: Session, storage_provider: Optional[StorageProvider] = None):
        """
        Orchestrates interactions with the storage providers, database persistence, and other business logic in the future
        """
        self.db = db
        self.storage_provider = storage_provider
        # we can set a default storage provider if that makes sense based on `get_active_default_storage_config`

    def upload_attachment(
        self, attachment_data: CreateAttachment, attachment: bytes
    ) -> str:
        # create attachment model
        db_attachment = Attachment.create(
            self.db, data=attachment_data.model_dump(mode="json")
        )
        self.storage_provider.upload_file(db_attachment.id, attachment)

    def retrieve_attachment(self, file_key: str) -> BytesIO:
        self.storage_provider.retrieve_file(file_key)

    def delete_attachment(self, file_key: str) -> str:
        # delete attachment model
        self.storage_provider.delete_file(file_key)
```
Then the `Attachment` model can just focus on persisting IDs and relationships to the database. So then we'd have a clean separation of concerns:

- StorageProviders - Just focused on writing files, don't need to to worry about any business logic
- StorageService - Allows us to inject the correct storage provider, or even a mock provider for testing
- AttachmentModel - Super bare-bones, just deals with relationships between other models

tests/ops/models/privacy_request/test_privacy_request.py
```diff
@@ -1132,3 +1144,139 @@ def test_persist_custom_identities(self, db, privacy_request):
...
+
+    monkeypatch.setattr(
+        "fides.api.service.storage.s3.get_s3_client", mock_get_s3_client
+    )
+
+    attachment1 = Attachment.create_and_upload(
+        db=db,
+        data=attachment_data,
+        attachment_file=b"contents of test file 1",
+    )
```
galvana, line 1165: If we move this away from the `Attachment` model and to the `StorageService`, we could test this with
```
mock_s3_storage_provider = S3StorageProvider(s3_client=mock_s3_client)
storage_service = StorageService(storage_provider=mock_s3_storage_provider)
storage_service.upload_attachment(attachment_data, attachment_file)
```

src/fides/api/models/attachment.py
galvana: We could put it here so we continue our pattern for new services
image

galvana (changes requested):
All the pieces are coming together! I made a non-trivial recommendation to refactor the upload logic out of the `Attachment` model and into a `StorageProviders` and a `StorageService` let me know what you think

src/fides/api/models/attachment.py
JadeCara: @erosselli **did** mention that! I think its a great idea. I have a ticket for it [here](https://ethyca.atlassian.net/browse/LJ-489). I wanted to get some of the endpoints up and running so Lucano will be unblocked but I intend to do it as part of this epic.

JadeCara: I updated the ticket with the above suggestions! Thank you!

galvana (approved):
Approved! Requested changes will be made in a follow-up ticket

PR #5922, opened 2025-03-21
LJ-372 Updated Attachments/uploads to handle large files.

JadeCara:
### Description Of Changes

Made some updates to Attachments, as well as the generic s3 storage functions that will allow them to handle larger files. This will allow us to upload files without having to read into memory. 

### Code Changes
 
* Updated the Attachment class to expect attachment_file as IO[bytes].
* Added chunking for uploads
* Added filesize checking for downloads. 
* Added some docs and additional error handling functionality to s3 upload functions. 

### Steps to Confirm

1.  Fidesplus should load without any overlap warnings. 
2. The rest of the functionality is still not linked to the front end and can be verified through test coverage/passing tests.

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:


src/fides/api/models/attachment.py
```diff
@@ -191,6 +199,7 @@ def create_and_upload(
             return attachment_model
         except Exception as e:
             log.error(f"Failed to upload attachment: {e}")
+            log.error(f"{e}")
```
JadeCara, line 202:
```suggestion
```

tests/ctl/models/test_attachment.py
```diff
@@ -11,18 +13,37 @@
...
+def verify_attachment_created_uploaded_s3(attachment, attachment_file_copy):
+    retrieved_file, download_url = attachment.retrieve_attachment()
+    assert retrieved_file == attachment_file_copy
+    assert attachment.config.details[StorageDetails.BUCKET.value] in download_url
+
+
+def verify_attachment_created_uploaded_local(attachment, attachment_file_copy):
+    retrieved_attachment, download_path = attachment.retrieve_attachment()
+    assert retrieved_attachment == attachment_file_copy
+    assert download_path == f"{LOCAL_FIDES_UPLOAD_DIRECTORY}/{attachment.id}"
```
JadeCara, line 46: Simplifies below tests which want to verify the attachments have uploaded.

tests/ops/service/storage/test_s3.py
```diff
@@ -1,40 +1,116 @@
...
-def s3_client(storage_config):
-    with mock_aws():
-        session = boto3.Session(
-            aws_access_key_id="fake_access_key",
-            aws_secret_access_key="fake_secret_key",
-            region_name="us-east-1",
-        )
-        s3 = session.client("s3")
-        s3.create_bucket(Bucket=storage_config.details[StorageDetails.BUCKET.value])
-        yield s3
```
JadeCara, line 23: This is available through conftest.

src/fides/api/models/comment.py
```diff
@@ -98,6 +98,7 @@ class Comment(Base):
         secondaryjoin="Attachment.id == AttachmentReference.attachment_id",
         order_by="Attachment.created_at",
         uselist=True,
+        viewonly=True,
     )
```
erosselli, line 103: I'm not too SQLAlchemy-savvy , for my own benefit can you explain why we have to add `viewonly=True` here? I understand it gets rid of the fidesplus warnings but not _why_ it does

erosselli: I also don't know if I follow all the models here, can you explain the `Comment` , `CommentReference` , `AttachmentReference` and `Attachment` models and how they relate to each other?

src/fides/api/models/privacy_request/privacy_request.py
```diff
@@ -174,13 +174,15 @@ class PrivacyRequest(
         primaryjoin="PrivacyRequest.id == AttachmentReference.reference_id",
         secondaryjoin="Attachment.id == AttachmentReference.attachment_id",
         order_by="Attachment.created_at",
+        viewonly=True,
```
erosselli, line 177: same question here

src/fides/api/service/storage/s3.py
```diff
@@ -38,76 +61,112 @@ def generic_upload_to_s3(  # pylint: disable=R0913
     bucket_name: str,
     file_key: str,
     auth_method: str,
-    document: BytesIO,
+    document: BytesIO | IO[bytes],
```
erosselli, line 64: what's the difference between `BytesIO` and `IO[Bytes]` ?

src/fides/api/models/comment.py
JadeCara: Both totally fair questions! 
I looked at all the models and decided to be much clearer on how things link for both sqlalchemy and for us going forward!
I have updated the models with very clear relationship and delete paths. 

Comments and Attachments have reference tables. This is designed so that in the future when we decide we want to add a comment or attachment option to any other object it will be much simpler/require no migrations etc to add a new reference type to either. 
These reference tables mean that we are using them as a secondary join to link up to Privacy request. It also means we have that extra added layer of abstraction when doing deletes. I updated all the backrefs and added some very clear `delete` functionality and tests for it! 

I am sorry this added extra lines but I think its a really good addition for clarity!

src/fides/api/service/storage/s3.py
JadeCara: Another really good question! (And honestly this is a bit redundant. I am going to adjust - read on for why!)
They are pretty similar. 

BytesIO:
-  is a concrete class from the io module in Python.
- It creates an in-memory binary stream that behaves like a file object.
- You can read from and write to it just like you would with a regular file.
- It is primarily used when you want to manipulate binary data in memory without creating actual files.
- BytesIO behaves like a binary file.
- You can perform file-like operations (read, write, seek, etc.).
- It stores the data in memory rather than on disk.

IO[Bytes]: 
- is a type hint used in Python type annotations.
- It indicates that a function or variable expects an I/O stream that handles Bytes.
- It is part of Python’s type hinting system and comes from the typing module.
- It is not an actual class or object but a type alias used for static typing.
- IO[bytes] is only used for type checking and documentation.
- It indicates that the function takes any binary stream (not just BytesIO).
- The stream could be a BytesIO, a socket, or any object that adheres to the binary I/O interface.

TLDR:
BytesIO → Concrete class: For working with in-memory binary streams.
IO[Bytes] → Type hint: For annotating functions that accept any bytes-based I/O stream.
Since BytesIO is already a valid IO[bytes] stream, explicitly including it in the union is unnecessary.

erosselli: thanks for the super detailed explanation!

src/fides/api/service/storage/s3.py
```diff
@@ -12,11 +11,34 @@
 from fides.api.util.aws_util import get_s3_client
 from fides.config import CONFIG
 
+LARGE_FILE_THRESHOLD = 5 * 1024 * 1024  # 5 MB threshold for large files
```
erosselli, line 14: where does this number come from? would we want to to be configurable, or is that overkill?

erosselli (approved):
Thanks for making the changes! I approved with one question (non-blocking), and I think you're missing a changelog entry

PR #5934, opened 2025-03-25
Fix changelog organization and tag Next/React upgrade as high risk

gilluminate:


lucanovera (approved):
Thanks for fixing this. Approved!

PR #5947, opened 2025-03-27
Address deprecation warnings from sass during builds

gilluminate:
### Description Of Changes

There were several warnings coming up during the Privacy Center builds after the latest round of npm package updates.

This change follows the modern Sass module system guidelines and removes the deprecation warning for the SCSS import. The functionality remains exactly the same, but it's now using the recommended `@use` syntax for SCSS files.

The `@use` rule is preferred because it:
- Only imports the module once, no matter how many times you use it
- Creates a namespace for the imported members
- Makes it clearer where each member comes from
- Has better performance characteristics

### Code Changes

* Using the modern `@use` rule instead of deprecated `@import` for Sass
* Using namespaced functions from Sass modules


lucanovera (approved):
Thanks for updating this to the new guidelines. Tested around in admin ui and privacy center and found no issues. Approved!

PR #5952, opened 2025-03-28
LJ-501 Add reject_all_mechanism field to PrivacyExperienceConfig

erosselli:
Implements part of  [LJ-501](https://ethyca.atlassian.net/browse/LJ-501)

### Description Of Changes

Adds a new field `reject_all_mechanism` to the `PrivacyExperienceConfig` (and to the `PrivacyExperienceConfigHistory`) model. This field will be used by FidesJS to determine the behavior of the "Reject all" button in TCF banners/modals. 

### Code Changes

* Add `reject_all_mechanism` field to models, as well as a new Enum 
* Migration to add `reject_all_mechanism` column to tables 
* Add tests

### Steps to Confirm

1.  Make sure you have a TCF experience created in your `privacy_experiences` table (the easiest way to do this is by running fidesplus and enabling TCF ; you can then pull down this fides branch and use it on your local fidesplus) 
2. Run migration, check that the column `reject_all_mechanism` is present in both the `privacyexperienceconfig` and `privacyexperienceconfighistory`.  The TCF experience should have "REJECT_ALL" automatically set 
4. Reverse the migration, column should be gone and DB should be in a good state 

### Pre-Merge Checklist

* [ ] Issue requirements met
* [x] `CHANGELOG.md` updated
* Followup issues:
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:

[LJ-501]: https://ethyca.atlassian.net/browse/LJ-501?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/clients/fidesui/src/hoc/CustomTag.module.scss
+++ b/clients/fidesui/src/hoc/CustomTag.module.scss
@@ -1,4 +1,5 @@
-@import "../palette/palette.module.scss";
+@use "sass:map";
+@use "../palette/palette.module.scss" as palette;
 
 .tag {
   display: inline-flex;
@@ -20,13 +21,13 @@
   &:focus {
     :global(.ant-tag) {
       // !important is necessary because our custom component adds the background style to the tag inline giving it a higher specificity than our scss style here
-      background: map-get($colors, "neutral-50") !important;
+      background: map.get(palette.$colors, "neutral-50") !important;
     }
   }
   &:focus-visible {
     outline: none;
     :global(.ant-tag) {
-      border: 1px solid map-get($colors, "minos");
+      border: 1px solid map.get(palette.$colors, "minos");
     }
   }
 }
@@ -34,7 +35,7 @@
 .closeButton {
   border-radius: var(--ant-border-radius-xs);
   &:focus-visible {
-    outline: 1px solid map-get($colors, "minos");
+    outline: 1px solid map.get(palette.$colors, "minos");
     outline-offset: 2px;
   }
 }
--- a/clients/fidesui/src/palette/palette.module.scss
+++ b/clients/fidesui/src/palette/palette.module.scss
@@ -1,3 +1,5 @@
+@use "sass:string";
+
 $colors: (
   "full-black": #000000,
   "full-white": #ffffff,
@@ -58,12 +60,12 @@ $colors: (
 }
 
 @function str-replace($string, $search, $replace: "") {
-  $index: str-index($string, $search);
+  $index: string.index($string, $search);
 
   @if $index {
-    @return str-slice($string, 1, $index - 1) + $replace +
+    @return string.slice($string, 1, $index - 1) + $replace +
       str-replace(
-        str-slice($string, $index + str-length($search)),
+        string.slice($string, $index + string.length($search)),
         $search,
         $replace
       );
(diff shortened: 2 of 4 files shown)


src/fides/api/alembic/migrations/versions/67d01c4e124e_add_reject_all_mechanism_to_privacy_.py
```diff
@@ -0,0 +1,56 @@
...
+            "reject_all_mechanism",
+            sa.Enum("REJECT_ALL", "REJECT_CONSENT_ONLY", name="rejectallmechanism"),
+            nullable=True,
+        ),
+    )
+
+    # Data migration - set the reject_all_mechanism for all TCF experience configs to REJECT_ALL
+    op.execute(
+        "UPDATE privacyexperienceconfig SET reject_all_mechanism = 'REJECT_ALL' WHERE component = 'tcf_overlay'"
+    )
```
erosselli, line 47: open question here: should we also create a new PrivacyExperienceConfigHistory instance, since we're technically changing the TCF experience config?

src/fides/api/db/base.py
```diff
@@ -8,6 +8,7 @@
 from fides.api.models.audit_log import AuditLog
 from fides.api.models.authentication_request import AuthenticationRequest
 from fides.api.models.client import ClientDetail
+from fides.api.models.comment import Comment, CommentReference
```
erosselli, line 11: not related but all models should be imported here. I was hoping this would fix the alembic autogenerate command, but it didn't

tests/api/models/test_privacy_experience.py
erosselli, line 1: I moved the test file so it's no longer under `ops` per what we discussed in backend retro

src/fides/api/models/privacy_experience.py
```diff
@@ -179,6 +193,10 @@ class PrivacyExperienceConfig(PrivacyExperienceConfigBase, Base):
     The Privacy Experience Configuration model that stores shared configuration for Privacy Experiences.
 
     - Translations, Notices, and Regions (via Privacy Experiences) are linked to this resource.
+
+    If you're adding a new PrivacyExperienceConfig, make sure to use the `create` method since it has
```
eastandwestwind, line 197: I know this expectation wasn't added by your PR but can we easily add some code-based validation to prevent direct instantiation of `PrivacyExperienceConfig`? 

We could override `__new__()` and throw error that says to prefer .create()

src/fides/api/alembic/migrations/versions/67d01c4e124e_add_reject_all_mechanism_to_privacy_.py
eastandwestwind: you mean migrate privacyexperienceconfighistory just like we did with privacyexperienceconfig? Or you mean something else?

src/fides/api/models/privacy_experience.py
erosselli: hmm I'd be hesitant to do that just in case SQL alchemy internally instantiates the model at some point ? I can try it out and see if any tests break but it feels like it might be a bit overkill for this, we rarely create instances without using a `create` method in our code base and we're not doing it for the experience config model in any place

src/fides/api/alembic/migrations/versions/67d01c4e124e_add_reject_all_mechanism_to_privacy_.py
eastandwestwind: ah I see what you mean... tough call. I don't think it's necessary for this migration though, as we're just populating a net new column. I don't view it as a change in history, but rather a new option becoming available for the user to then make future changes.

eastandwestwind (approved):

PR #5962, opened 2025-03-31
TCF Preview

gilluminate:
Closes [LJ-612]

### Description Of Changes

Adds TCF preview and cleans up some layout issues in the Preview, especially for the mobile version.

### Code Changes

* Replaced `AntButton` with `Radio.Group` for mobile/desktop preview toggle
* Updated preview container styling
* Added TCF overlay support in preview configuration
* Fixed router navigation to prevent caching issues when switching between TCF and non-TCF experiences
* Refactored `buildBaseConfig` to support TCF-specific configuration options
* Removed redundant privacy notice fetching logic in favor of using existing data

### Steps to Confirm

1. Test the privacy experience configuration with both TCF and non-TCF experiences
2. Verify the preview works correctly in both mobile and desktop views
3. Check that switching between different privacy experiences doesn't cause caching issues (TCF modal isn't showing in US experience, and vice-versa)
5. Validate that banner and modal positioning is correct in all scenarios (Including the new change for showing banner on bottom, modal on top)

### Pre-Merge Checklist

* [x] Issue requirements met
* [x] `CHANGELOG.md` updated
* Followup issues:
  * [x] Followup issues created (include link)
  * [ ] No followup issues
* Database migrations:
  * [x] No migrations
* Documentation:

[LJ-612]: https://ethyca.atlassian.net/browse/LJ-612?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lucanovera (commented):
Great work here, it's great to finally be adding support for more previews!
I've tested through the previews, found no caching or other type of issues. I have just two questions:
1. Previews are suppose to be non-interactive, this makes it a lot easier to implement and saves us from a lot of issues, but I kind of expected the TCF preview to still let me scroll. See, for a non-tcf the preview let's you see all the notices, but for TCF scrolling is necessary for me to be able to see that all the notices are shown. Would it be possible to allow scrolling? I think the preview would be more useful if it allows you to check what notices are being displayed.
2. Regarding the caching issue, is the problem that the browser caches the request or that the window/cookies is left in a different state? I'm wondering if there's a way to solve the caching issue without using changing the window.location since that causes a blank screen during navigation in admin-ui. If it's the browser cache, can we add a random queryParam to avoid the cache?

gilluminate:
@lucanovera 
1. I don't think we can allow scrolling with the method we're currently using to disable everything. The entire wrapper window is given the css style of `pointer-events: none`. It might be possible to really dig in and allow some interaction but not others, but that's too far outside the scope of this right now I think, 
2. It's not really a caching issue, it's the fact that it loads `fides.js` in a script tag on the page and for tcf that needs to be `fides-tcf.js`. It doesn't work to swap out `<script>` tags and doing that doesn't wipe and reload the `window.Fides` anyway. It really needs to be a full reload to work properly. The white screen between pages is definitely less of an issue once this gets built by NextJS for prod, so I think the tradeoff is worth it.

PR #5969, opened 2025-04-01
LJ-177: Adds tests for Enterprise Bigquery DSR nested fields

eastandwestwind:
Closes https://ethyca.atlassian.net/browse/LJ-177

### Description Of Changes

Adds tests for Enterprise Bigquery DSR nested fields

### Code Changes

* (not in PR) Already updated our `enterprise_dsr_testing.users` BQ table to include nested fields
* Adjusts BQ Enterprise dataset
* Updates tests to assert access/deletion is fully functional with STRUCT format for nested fields

### Steps to Confirm

1.  Add `run unsafe ci checks` label, confirm tests pass

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:


galvana (changes requested):
There was an issue with the tags check but the test didn't report it because this line doesn't return any results
```
stmt = f"select display_name, location from enterprise_dsr_testing.users where id = {user_id};"
res = connection.execute(stmt).all()
```
The data category on `users.id` is `user.contact` so that means the column will be nulled by our erasure policy. Then when we try to retrieve the row by `id` we won't get any results. Also, the data category of `user` on tags is not specific enough to get masked by our policy so the tags will remain.

Overall, you might need to add some asserts on the expected result count for each retrieval in `validate_erasure_privacy_request`, that was we don't skip over any assertions

galvana (approved):

PR #5976, opened 2025-04-01
remove transitions from banner on FidesJS tests

gilluminate:
### Description Of Changes

Each time the banner loads in Cypress tests for FidesJS, it takes a full second before the tests can properly run while it waits for the animation to run. Removing the CSS transition can shave a couple minutes off the total Cypress run time.

### Code Changes

* Removes transitions for banner during Cypress tests.


lucanovera (approved):
👏 👏 👏  great work in this pr! it improves the tests runtime, the reliability of the tests and you also cleaned up the code. Approved!

PR #5998, opened 2025-04-04
Read db.

tvandort:
Closes [LJ-643](https://ethyca.atlassian.net/browse/LJ-643)

### Description Of Changes

### Code Changes

### Steps to Confirm

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:

[LJ-643]: https://ethyca.atlassian.net/browse/LJ-643?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/fides/api/api/deps.py
+++ b/src/fides/api/api/deps.py
@@ -12,6 +12,7 @@ from fides.config import get_config as get_app_config
 from fides.config.config_proxy import ConfigProxy
 
 _engine = None
+_read_engine = None
 
 
 def get_config() -> FidesConfig:
@@ -23,6 +24,15 @@ def get_db() -> Generator:
     """Return our database session"""
     try:
         db = get_api_session()
+        yield db
+    finally:
+        db.close()
+
+
+def get_readonly_db() -> Generator:
+    """Return our readonly database session"""
+    try:
+        db = get_readonly_api_session()
         yield db
     finally:
         db.close()
@@ -51,6 +61,26 @@ def get_api_session() -> Session:
             keepalives_count=CONFIG.database.api_engine_keepalives_count,
         )
     SessionLocal = get_db_session(CONFIG, engine=_engine)
+    db = SessionLocal()
+    return db
+
+
+def get_readonly_api_session() -> Session:
+    """Gets the shared readonly database session to use for API functionality"""
+    if not CONFIG.database.sqlalchemy_readonly_database_uri:
+        return get_api_session()
+
+    global _read_engine  # pylint: disable=W0603
+    if not _read_engine:
+        _read_engine = get_db_engine(
+            database_uri=CONFIG.database.sqlalchemy_readonly_database_uri,
+            pool_size=CONFIG.database.api_engine_pool_size,
+            max_overflow=CONFIG.database.api_engine_max_overflow,
+            keepalives_idle=CONFIG.database.api_engine_keepalives_idle,
+            keepalives_interval=CONFIG.database.api_engine_keepalives_interval,
+            keepalives_count=CONFIG.database.api_engine_keepalives_count,
+        )
+    SessionLocal = get_db_session(CONFIG, engine=_read_engine)
     db = SessionLocal()
     return db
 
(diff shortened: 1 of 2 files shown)


src/fides/config/database_settings.py
```diff
@@ -121,6 +130,14 @@ class DatabaseSettings(FidesSettings):
...
     )
 
+    # @field_validator("read_sever", mode="before")
+    # def copy_server_if_no_read_server(
+    #     cls, value: Optional[str], info: ValidationInfo
+    # ) -> str:
+    #     if not value:
+    #         return info.data.get("server")
+    #     return value
+
```
adamsachs, line 140: drive by comment: do we not need some fallback logic like this? i may be missing it, but it does feel like we want to ensure that things still work just fine if you don't set the `read_server`, i.e. it should fallback to what the `server` value is

src/fides/config/database_settings.py
```diff
@@ -218,6 +235,32 @@ def assemble_db_connection(cls, v: Optional[str], info: ValidationInfo) -> str:
             )
         )
 
+    @field_validator("sqlalchemy_read_database_uri", mode="before")
+    @classmethod
+    def assemble_read_db_connection(cls, v: Optional[str], info: ValidationInfo) -> str:
```
NevilleS, line 240:
```suggestion
    @field_validator("sqlalchemy_readonly_database_uri", mode="before")
    @classmethod
    def assemble_readonly_db_connection(cls, v: Optional[str], info: ValidationInfo) -> Optional[str]:
```

src/fides/api/api/deps.py
```diff
@@ -61,6 +62,22 @@ def get_api_session() -> Session:
     return db
 
 
+def get_read_api_session() -> Session:
+    global _read_engine  # pylint: disable=W0603
```
NevilleS, line 66: nit: what I'd recommend we do here is a more explicit check in this module like:
```python
if not CONFIG.database.settings.sqlalchemy_readonly_database_uri:
  # return the regular engine

# else, continue...
```

NevilleS (approved):

src/fides/config/database_settings.py
```diff
@@ -65,6 +65,10 @@ class DatabaseSettings(FidesSettings):
         default="default-db",
         description="The hostname of the application database server.",
     )
+    read_server: str = Field(
+        default="default-db",
+        description="The hostname of the application read database server.",
+    )
```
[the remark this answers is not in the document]
tvandort: I considered doing this but... In any scenario that you're dealing with a replica when would the user, password, or db be different? Wouldn't this only increase the chance of a breakage and make configuration more complicated?

I can _maybe_ see params and port being things that vary between a primary and replica.

src/fides/api/api/deps.py
tvandort: Maybe return the result of calling `get_api_session` instead so that we fall back to single connection behavior?

src/fides/config/database_settings.py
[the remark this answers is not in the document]
NevilleS: (We discussed live)

I think it's _highly_ likely that there will be a customer who needs a completely different set of user/pwd/server/etc. for their R/O DB, so I'd ask us to factor that in before we consider this feature "done" 👍 

That said, nothing wrong with merging an initial beta version that doesn't support that, to get it up & running and tested in our own environments. I'd just want to nitpick the customer-facing config settings for this before it's released & documented

src/fides/config/database_settings.py
tvandort: Implemented this how Neville suggested instead.

src/fides/config/database_settings.py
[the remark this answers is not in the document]
tvandort: Added TODOs

PR #6000, opened 2025-04-04
LJ-622 Refactor TCFPublisherRestrictions to use async methods

erosselli:
### Description Of Changes

Adds the `update_async` method in the `TCFPublisherRestriction` model, and also adds validation. The validation is as follows: a single TCF configuration may have multiple restrictions for the same purpose, as long as there are no overlapping vendor ranges for those restrictions (for the same purpose) 

Additionally, I replaced the implementation of `create` so it throws a `NotImplemented` error -- I think it's confusing to have two ways to create the instances and we want to move towards async sessions in the future anyways. The fidesplus logic for this feature uses exclusively async sessions so it won't be a problem, and this avoids having to duplicate all the validation logic for both sync and async sessions. 

No changelog entry since this model is brand new and will be new in this release .

### Steps to Confirm

1.  Tests pass 

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:

### DIFF
--- a/src/fides/api/models/tcf_publisher_restrictions.py
+++ b/src/fides/api/models/tcf_publisher_restrictions.py
@@ -4,7 +4,7 @@ from typing import Any, Dict, List, Optional
 from pydantic import BaseModel, Field, ValidationError, model_validator
 from sqlalchemy import Column
 from sqlalchemy import Enum as EnumColumn
-from sqlalchemy import ForeignKey, Index, Integer, String, insert, select
+from sqlalchemy import ForeignKey, Index, Integer, String, insert, select, update
 from sqlalchemy.dialects.postgresql import ARRAY, JSONB
 from sqlalchemy.ext.asyncio import AsyncSession
 from sqlalchemy.ext.declarative import declared_attr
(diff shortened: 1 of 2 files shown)


src/fides/api/models/tcf_publisher_restrictions.py
```diff
@@ -185,21 +185,56 @@ def validate_publisher_restriction_data(
...
+        )
+
     @classmethod
     def create(
         cls,
         db: Session,
         *,
         data: Dict[str, Any],
-        check_name: bool = False,
+        check_name: bool = True,
```
erosselli, line 235: this is just the default value for the base class -- not used so it doesn't matter a whole lot, but I wanted it to be consistent with the base definition

src/fides/api/models/tcf_publisher_restrictions.py
```diff
@@ -185,21 +185,56 @@ def validate_publisher_restriction_data(
...
     ) -> "TCFPublisherRestriction":
-        """
-        Create a new TCFPublisherRestriction with validated range_entries.
-        Validates each range entry using the RangeEntry Pydantic model.
-        """
-        data = cls.validate_publisher_restriction_data(data)
-
-        return super().create(db=db, data=data, check_name=check_name)
+        raise NotImplementedError("Use create_async instead")
```
erosselli, line 237: see PR description

JadeCara: This seems like a good decision.

JadeCara (approved):
Thanks for the super clear comments. I think I can wrap my head around the changes. It all looks logical and well orded. 🥂

PR #6022, opened 2025-04-09
LJ-392 Add comments and attachments to manual access webhooks

JadeCara:
Closes [LJ-392](https://ethyca.atlassian.net/browse/LJ-392?atlOrigin=eyJpIjoiMzU4ZDVkMTU2M2VjNDIwZDg1ODhiOGFmYWQ5YWIxZjEiLCJwIjoiaiJ9)

### Description Of Changes

Added attachments and comments relationships to Manual Webhooks as well as adding a new "types" field to manual webhooks. This will allow the user to indicate they want to require some kind of attachment to the manual step. This will satisfy the following requirement:

> - When configuring DSRs, I can add one or more “Manual steps” to my DSR process that can be configured to support file uploads and comments instead of the current fields.

Additionally added att

### Code Changes

* Added comment and attachment relationships to manual webhooks with tests
* Added attachment relationship to comments with tests
* Added "types" to fields in manual webhooks to allow users to indicate there they expect an attachment rather than just a string with tests. 
* Added backwards compatibility tests.

### Steps to Confirm

1. This will be fidesplus functionality. There should be no changes to fides functionality. Tests should cover all changes. 

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:

[LJ-392]: https://ethyca.atlassian.net/browse/LJ-392?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/fides/api/schemas/manual_webhook_schemas.py
```diff
@@ -49,68 +48,57 @@ class AccessManualWebhooks(FidesSchema):
...
-            field.dsr_package_label for field in value
+        unique_dsr_package_labels: Set[str] = {
+            field.dsr_package_label
+            for field in self.fields
+            if field.dsr_package_label is not None
         }
-        if len(value) != len(unique_dsr_package_labels):
-            # Postponing dsr_package_label uniqueness check in case we get overlaps
-            # above when we fallback to converting pii_fields to dsr_package_labels
+        if len(self.fields) != len(unique_dsr_package_labels):
```
JadeCara, line 72:
```suggestion
        if len(value) != len(unique_dsr_package_labels):
            # Postponing dsr_package_label uniqueness check in case we get overlaps
            # above when we fallback to converting pii_fields to dsr_package_labels
```

src/fides/api/schemas/manual_webhook_schemas.py
```diff
@@ -49,68 +48,57 @@ class AccessManualWebhooks(FidesSchema):
...
                 field.dsr_package_label = DSRLabelFieldType(
                     to_snake_case(field.pii_field)
                 )
 
-        unique_dsr_package_labels: Set[Optional[str]] = {
-            field.dsr_package_label for field in value
+        unique_dsr_package_labels: Set[str] = {
+            field.dsr_package_label
+            for field in self.fields
+            if field.dsr_package_label is not None
```
JadeCara, line 70:
```suggestion
        unique_dsr_package_labels: Set[Optional[str]] = {
            field.dsr_package_label for field in value
```

src/fides/api/schemas/manual_webhook_schemas.py
```diff
@@ -49,68 +48,57 @@ class AccessManualWebhooks(FidesSchema):
...
         Set the dsr_package_label to a snake_cased lower case version of pii field if it doesn't exist.
         """
-        unique_pii_fields: Set[str] = {field.pii_field for field in value}
-        if len(value) != len(unique_pii_fields):
+        unique_pii_fields: Set[str] = {field.pii_field for field in self.fields}
+        if len(self.fields) != len(unique_pii_fields):
             raise ValueError("pii_fields must be unique")
 
-        for field in value:
+        for field in self.fields:
```
JadeCara, line 61:
```suggestion
        for field in value:
```

src/fides/api/schemas/manual_webhook_schemas.py
```diff
@@ -49,68 +48,57 @@ class AccessManualWebhooks(FidesSchema):
...
+    def check_for_duplicates(self) -> "AccessManualWebhooks":
         """
         Verify that pii_fields and dsr_package_labels are unique.
-
         Set the dsr_package_label to a snake_cased lower case version of pii field if it doesn't exist.
         """
-        unique_pii_fields: Set[str] = {field.pii_field for field in value}
-        if len(value) != len(unique_pii_fields):
+        unique_pii_fields: Set[str] = {field.pii_field for field in self.fields}
+        if len(self.fields) != len(unique_pii_fields):
```
JadeCara, line 58:
```suggestion
        unique_pii_fields: Set[str] = {field.pii_field for field in value}
        if len(value) != len(unique_pii_fields):
```

src/fides/api/schemas/manual_webhook_schemas.py
```diff
@@ -49,68 +48,57 @@ class AccessManualWebhooks(FidesSchema):
     fields: ManualWebhookFieldsList
     model_config = ConfigDict(from_attributes=True)
 
-    @field_validator("fields")
-    @classmethod
-    def check_for_duplicates(
-        cls, value: List[ManualWebhookField]
-    ) -> List[ManualWebhookField]:
+    @model_validator(mode="after")
+    def check_for_duplicates(self) -> "AccessManualWebhooks":
```
JadeCara, line 52:
```suggestion
    @field_validator("fields")
    @classmethod
    def check_for_duplicates(
        cls, value: List[ManualWebhookField]
    ) -> List[ManualWebhookField]:
```

src/fides/api/schemas/manual_webhook_schemas.py
```diff
@@ -49,68 +48,57 @@ class AccessManualWebhooks(FidesSchema):
...
+        return field_definitions
+
+    def erasure_field_definitions(self) -> Dict[str, Any]:
+        """Shared erasure field definitions for manual webhook schemas.
+        Only string fields can be used for erasure confirmation."""
+        return {
+            field.dsr_package_label: (Optional[bool], None)
+            for field in self.fields or []
+            if field.dsr_package_label
+            and "string" in field.types  # only include string fields
```
JadeCara, line 95:
```suggestion
        return value

    @field_validator("fields")
    @classmethod
    def fields_must_exist(
        cls, value: List[ManualWebhookField]
    ) -> List[ManualWebhookField]:
        """
        Verify that pii_fields and dsr_package_labels are unique.
        Set the dsr_package_label to a snake_cased lower case version of pii field if it doesn't exist.
        """
        unique_pii_fields: Set[str] = {field.pii_field for field in value}
        if len(value) != len(unique_pii_fields):
            raise ValueError("pii_fields must be unique")

        for field in value:
            if not field.dsr_package_label:
                field.dsr_package_label = DSRLabelFieldType(
                    to_snake_case(field.pii_field)
                )

        unique_dsr_package_labels: Set[Optional[str]] = {
            field.dsr_package_label for field in value
```

src/fides/api/models/manual_webhook.py
```diff
@@ -30,21 +33,49 @@ class AccessManualWebhook(Base):
...
-            for field in self.fields or []
-        }
+        field_definitions = {}
+        for field in self.fields or []:
+            # Use types if present, otherwise default to ["string"]
+            field_types = field.get("types", ["string"])
+            # Only include string fields for erasure
+            if "dsr_package_label" in field and "string" in field_types:
+                field_definitions[field["dsr_package_label"]] = (Optional[bool], None)
+        return field_definitions
```
JadeCara, line 78: Updated these to verify backwards compatibility.

src/fides/api/models/manual_webhook.py
```diff
@@ -90,7 +121,7 @@ def erasure_fields_non_strict_schema(self) -> FidesSchema:
         )
 
     @property
-    def empty_fields_dict(self) -> Dict[str, None]:
+    def empty_fields_dict(self) -> dict[str, None]:
```
JadeCara, line 124: Updated `List` and `Dict` to `list` and `dict` since they updated this in 3.9 and later versions [PEP-585](https://peps.python.org/pep-0585/) and it cleans up our imports a little.

src/fides/api/schemas/manual_webhook_schemas.py
```diff
@@ -20,8 +20,9 @@ class ManualWebhookField(FidesSchema):
     pii_field: Annotated[
         str, StringConstraints(min_length=1, max_length=200, strip_whitespace=True)
     ]
+    types: list[Literal["string", "file"]] = ["string"]
```
JadeCara, line 23: This is the real change here - the rest are mostly [PEP-585](https://peps.python.org/pep-0585/) updates

src/fides/api/models/manual_webhook.py
```diff
@@ -30,21 +33,49 @@ class AccessManualWebhook(Base):
...
-            field["dsr_package_label"]: (Optional[bool], None)
-            for field in self.fields or []
-        }
+        field_definitions = {}
+        for field in self.fields or []:
+            # Use types if present, otherwise default to ["string"]
+            field_types = field.get("types", ["string"])
+            # Only include string fields for erasure
+            if "dsr_package_label" in field and "string" in field_types:
+                field_definitions[field["dsr_package_label"]] = (Optional[bool], None)
```
erosselli, line 77: I don't think I follow here - what is the `field_types` used for? just to check if `"string"` is in it?

erosselli (approved):
LGTM! Left a non-blocking question

src/fides/api/models/manual_webhook.py
JadeCara: This is primarily an Erasure vs Access thing:
For access requests, all fields can be returned (strings, files, etc.)
For erasure requests, you can only "erase" string fields - you can't meaningfully "erase" a file (attachment) in the same way. We use the same base for both but only string fields get erased through this mechanism while attachments need to be deleted through their own mechanism.

They also have different kinds of responses. The erasure field type is Optional[bool] because the response indicates whether the field should be erased (`True`) or not (`False`). This is different from access requests where you return the actual data (`Optional[str]`)

This allows explicit backwards compatibility. The .get("types", ["string"]) default ensures old-style fields (without types) are treated as string fields. This maintains compatibility with existing webhooks in the database.

PR #6027, opened 2025-04-10
LJ-672

tvandort:
Closes [LJ-672](https://ethyca.atlassian.net/browse/LJ-672?atlOrigin=eyJpIjoiNzA2NzYyNjE1Y2Y5NGI4MGE1MTFlMTkwODRlOWViYWEiLCJwIjoiaiJ9)

### Description Of Changes

This PR adds the ability to have DSRs use an assumed role to communicate with AWS S3 storage. It keeps backwards compatibility with existing configurations while adding the ability to assume roles when:

- Configuring a storage secret
- Using environment variables

### Code Changes

- **Assume role arn and region config.**
- **Update required secrets?**
- **Optional fields.**
- **Don't import things we don't need.**
- **Use config.**

### Steps to Confirm

1.  Update the `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` values in your .env file to the ones for `sandboxProgrammaticUserS3Assumer` in 1password and start up fides!

2. Add an S3 storage config 
```json
{
  "type": "s3",
  "details": {
    "naming": "request_id",
    "auth_method": "automatic",
    "bucket": "test_bucket",
    "max_retries": 0
  },
  "format": "json"
}
```
Screenshot 2025-04-15 at 3 54 08 PM

2. Add default storage secrets
```json
{
  "aws_access_key_id": "<available in 1password>",
  "aws_secret_access_key": "<available in 1password>",
  "region_name": "us-east-1",
  "assume_role_arn": "arn:aws:iam::<available in 1password>:role/S3FullAccessRole"
}
```
3. Verify that the secrets authenticated in logs and result:

Logs:

This part especially!!
` Assumed role arn:aws:iam::<account number>:role/S3FullAccessRole and got temporary credentials.`

```bash
2025-04-15 16:47:34 2025-04-15 22:47:34.943 | INFO     | fides.api.api.v1.endpoints.storage_endpoints:put_default_config_secrets:499 - Finding default config of storage type 's3'
2025-04-15 16:47:34 2025-04-15 22:47:34.965 | INFO     | fides.api.api.v1.endpoints.storage_endpoints:put_default_config_secrets:522 - Updating storage config secrets for default config of storage type 's3'
2025-04-15 16:47:34 2025-04-15 22:47:34.992 | INFO     | fides.api.util.aws_util:get_aws_session:44 - Successfully created automatic session
2025-04-15 16:47:35 2025-04-15 22:47:35.671 | INFO     | fides.api.util.aws_util:get_assumed_role_session:77 - Assumed role arn:aws:iam::<account number>:role/S3FullAccessRole and got temporary credentials.
2025-04-15 16:47:35 2025-04-15 22:47:35.680 | INFO     | fides.api.api.v1.endpoints.storage_endpoints:put_default_config_secrets:538 - Storage secrets are valid for default config of storage type 's3'
```

Result:
```bash

{
   "msg": "Secrets updated for default config of storage type: s3.",   
   "test_status": "succeeded",   
   "failure_reason": null 
}

```

### Pre-Merge Checklist

* [x] Issue requirements met
* [x] `CHANGELOG.md` updated
* Followup issues:
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:

[LJ-672]: https://ethyca.atlassian.net/browse/LJ-672?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/fides/api/schemas/storage/storage.py
```diff
@@ -80,10 +82,10 @@ class StorageSecretsLocal(BaseModel):
...
 
 class StorageSecretsS3(BaseModel):
-    """The secrets required to connect to an S3 bucket."""
-
-    aws_access_key_id: str
-    aws_secret_access_key: str
+    aws_access_key_id: Optional[str] = None
+    aws_secret_access_key: Optional[str] = None
+    region_name: Optional[str] = None
+    assume_role_arn: Optional[str] = None
```
tvandort, line 88: Better way to union this type?

JadeCara: We might want to add some kind of validator here that says we need one or the other.
Something like this: 
```suggestion
    aws_access_key_id: Optional[str] = None
    aws_secret_access_key: Optional[str] = None
    region_name: Optional[str] = None
    assume_role_arn: Optional[str] = None
    
    @model_validator(mode='after')
    def validate_aws_credentials(self) -> 'StorageSecretsS3':
        """
        Validate that either:
        1. Both AWS access key and secret key are provided, or
        2. Both region name and assume role ARN are provided
        """
        has_aws_secrets = all(
            [
                self.aws_access_key_id,
                self.aws_secret_access_key
            ]
        )
        has_assume_role = all(
            [
                self.region_name,
                self.assume_role_arn
            ]
        )

        if not (has_aws_secrets or has_assume_role):
            raise ValueError(
                "Must provide either both AWS credentials "
                "(aws_access_key_id and aws_secret_access_key) "
                "or both role assumption details "
                "(region_name and assume_role_arn)"
            )

        # I think we would want totally different storage configs for these
        # Meaning If they have the option to use either it would be one that is role based
        # and one that is secrets based.
        if has_aws_secrets and has_assume_role:
            raise ValueError(
                "Cannot provide both AWS credentials and role assumption details"
            )

        return self
```

JadeCara: Another option might be to have a different class that is just the role_arn stuff and say the secrets would be either/or, but I think this works for now.

tvandort: @JadeCara I don't _think_ assume role arn and access key / secret key are mutually exclusive. 

I think we'd want: 
- Any time access key is defined, so is secret key (and vice versa)
- The rest of the props are optional

I think I can adapt the above to that format.

JadeCara: There one question I have is - based on the storage config, how would it know which auth option to use? Can we assume users would for sure want one auth option over the other?

tvandort: When you post to: 
`/api/v1/storage/default`

The body looks like:
```json
{
    "type": "s3",
    "details": {
        "bucket": "something",
        "max_retries": 0,
        "naming": "request_id",
        "auth_method": "automatic"
    },
    "format": "html"
}
```

So the user chooses at that point. They then get yelled at if they haven't provided the matching secrets when you try to upload using a mode that you do not have the right secrets for.

I actually this that there's a strong argument to say that `assume_role_arn` should be posted to the above endpoint instead of `/api/v1/storage/default/:storage_type/secret`. Same with `region` since neither are secrets. But that being said there seems to be history for reading these values from secrets so I kept it working that way. Plus `get_aws_session` accepts the secrets but not the config and I didn't want to have to refactor _everywhere_ `get_aws_session` was used because it would make this a larger / riskier change (I think.)

tvandort: Oh! User would also get barked at here: https://github.com/ethyca/fides/blob/f6134146a57fafcb9033528edf62c7d5786b9297/src/fides/api/schemas/connection_configuration/connection_secrets_base_aws.py#L39-L51

So I think I'm fine to proceed with the change in this file.

JadeCara: The good news is I have a ticket to totally update the storage stuff into service. So we need to make sure it works, but I can fine tune it when I get to that ticket :D

JadeCara (approved):
:ship:

PR #6046, opened 2025-04-14
Improve Formik performance on datamap report

gilluminate:
Closes: [LJ-687]

### Description Of Changes

Cypress tests were becoming flaky on the data map report due to the fact that Formik was re-rendering the entire table with every form change, causing a react performance issue. Tests were timing out and failing eventually because of it.

This performance issue was also causing some visible slowness in the UI for the end user.

To address both the Cypress and user issues, I've gone ahead and refactored the Formik implementation to rely more heavily on `useFormikContext` rather than wrapping the entire Formik child in a function that passes those same values in.

The result is much faster and the tests are no longer flaky
CleanShot 2025-04-14 at 16 53 00@2x

### Code Changes

* Updated form handling in `DatamapReportTable` to use Formik context more effectively by removing the child wrapper function
* Added `resetForm` parameter to `onCustomReportSaved` callback to allow using `useFormikContext` instead of passing it around from the child function
* Extracted column renaming buttons into a new `RenameColumnsButtons` component to allow using `useFormikContext`
* Replaced `useField` with `FastField` in `EditableHeaderCell` for better performance
* Enabled previously skipped column customization tests by removing `.skip` from the test description

### Steps to Confirm
1. Navigate to the Data Map Report table
3. Verify column renaming functionality works:
   - Reset all button clears customizations
   - Cancel button reverts changes
   - Apply button saves changes
4. Confirm the column customization tests are now passing

### Pre-Merge Checklist

* [x] Issue requirements met
* [x] `CHANGELOG.md` updated
* Followup issues:
  * [x] No followup issues
* Database migrations:
  * [x] No migrations
* Documentation:

[LJ-687]: https://ethyca.atlassian.net/browse/LJ-687?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@ Changes can also be flagged with a GitHub label for tracking purposes. The URL o
 - Fixed typo in Vermont US state name [#6029](https://github.com/ethyca/fides/pull/6029)
 - Fixed two Georgia's in regions list and incorrect name for the state SD [#6036](https://github.com/ethyca/fides/pull/6036)
 - Fixed Privacy Center issue where unconfigured fields (eg. phone) were being passed as null form values [#6045](https://github.com/ethyca/fides/pull/6045)
+- Fixed performance issues in Data map report in Admin UI [#6046](https://github.com/ethyca/fides/pull/6046)
 
 ## [2.59.0](https://github.com/ethyca/fides/compare/2.58.2...2.59.0)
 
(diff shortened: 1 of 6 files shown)


jpople (approved):
Changes look good, thanks for the fix!

PR #6057, opened 2025-04-17
Lock editing system assets on GVL systems

jpople:
Closes HA-620

### Description Of Changes

Disables editing of system assets when the system is locked due to being a GVL vendor.

### Code Changes

* Refactors system asset columns
* Adds read-only variants of previously editable cells

### Steps to Confirm

Run a Fidesplus backend with TCF enabled.

1. Create a system without a GVL vendor
2. In "Assets" tab, should be able to add, edit and delete assets as normal
3. Create a new system using a GVL vendor with assets (e.g. "AA Media")
4. Form should lock and show GVL notice as normal
5. Save, then navigate to "Assets" tab
6. Should not be able to add, delete, or make any changes to assets

### Pre-Merge Checklist

* [x] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:


thingscouldbeworse (approved):

jpople:
@andres-torres-marroquin We also don't enforce the lock on the backend for the other system fields-- you can still manually edit the name, description or other information via the API if you want, for example.  I don't remember if that's specifically intentional but it's not a change to existing behavior, I don't think it should cause problems.

PR #6066, opened 2025-04-18
Allow non-flexible Purposes to add Purpose Restriction

gilluminate:
Closes [LJ-694]

### Description Of Changes

Removes restrictions on editing purpose restrictions for non-flexible purposes in TCF consent settings. Instead of blocking editing entirely, the UI now allows users to create restrictions for these purposes but limits them to only use Purpose Restriction type.

### Code Changes

* Removed code that hid the "Edit" button for non-flexible purposes in `PublisherRestrictionsTable`
* Added state and logic in `PurposeRestrictionFormModal` to disable the Restriction Type dropdown for non-flexible purposes
* Added a tooltip explanation for why the Restriction Type is disabled
* Removed the error toast and redirect in the purpose config page (`[purpose_id].tsx`) that prevented accessing non-flexible purposes

### Steps to Confirm
1. Navigate to a TCF consent configuration (`/settings/consent`)
2. Attempt to edit restrictions for a non-flexible purpose (those with "No" in the Flexible column)
3. Verify that the Edit button is visible and the form opens
4. Confirm the Restriction type is set to Purpose restriction and cannot be changed
NOTE: Backend validation with throw an error for now, @JadeCara is working to address this

### Pre-Merge Checklist

* [x] Issue requirements met
* [ ] `CHANGELOG.md` updated
  * [ ] Add a https://github.com/ethyca/fides/labels/high-risk label to the entry if your 
change includes a high-risk change (i.e. potential for performance impact or unexpected regression) that should be flagged
  * [x] Bug fix for current release work, no entry required
* Followup issues:
  * [x] No followup issues
* Database migrations:
  * [x] No migrations
* Documentation:

[LJ-694]: https://ethyca.atlassian.net/browse/LJ-694?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/clients/admin-ui/src/features/consent-settings/tcf/PurposeRestrictionFormModal.tsx
+++ b/clients/admin-ui/src/features/consent-settings/tcf/PurposeRestrictionFormModal.tsx
@@ -1,11 +1,13 @@
 import {
   AntButton as Button,
   AntFlex as Flex,
+  AntTooltip as Tooltip,
   Collapse,
   Text,
   useToast,
 } from "fidesui";
 import { Form, Formik } from "formik";
+import { useEffect, useState } from "react";
 import * as Yup from "yup";
 
 import { ControlledSelect } from "~/features/common/form/ControlledSelect";
@@ -19,6 +21,7 @@ import {
 } from "~/types/api";
 
 import {
+  FORBIDDEN_LEGITIMATE_INTEREST_PURPOSE_IDS,
   RESTRICTION_TYPE_LABELS,
   VENDOR_RESTRICTION_LABELS,
 } from "./constants";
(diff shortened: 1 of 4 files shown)


clients/admin-ui/src/features/consent-settings/tcf/PurposeRestrictionFormModal.tsx
```diff
@@ -65,6 +68,17 @@ export const PurposeRestrictionFormModal = ({
   const toast = useToast();
   const [createRestriction] = useCreatePublisherRestrictionMutation();
   const [updateRestriction] = useUpdatePublisherRestrictionMutation();
+  const [isPurposeFlexible, setIsPurposeFlexible] = useState(true);
```
lucanovera, line 71: I've tested the feature and it works correctly. However, I found a bit odd that the during the first render the initialValues are set to the original values, then the useEffect runs, and then the initialValues get changed.
Captura de pantalla 2025-04-21 a la(s) 4 51 45 p  m

Apparently this works fine with Formik, but I think with a couple small improvement this can be cleaned and avoid the changing values during rendering (and also avoiding the eslint rule disable).

The purpose flexible can be derived from the prop without needing a state
`
const isPurposeFlexible = !(
    purposeId && FORBIDDEN_LEGITIMATE_INTEREST_PURPOSE_IDS.includes(+purposeId)
  );
  `
  so it's value is the correct one even during the first render.
  
  and the initial values is only used in the Formik component so it can be overriden there:
  `<Formik
        initialValues={{
          ...initialValues,
          restriction_type: isPurposeFlexible
            ? initialValues.restriction_type
            : TCFRestrictionType.PURPOSE_RESTRICTION,
        }}`

What do you think? Let me know if you see any issues with this approach.

lucanovera (commented):
@gilluminate I've tested and everything is working correctly 💯 
Captura de pantalla 2025-04-21 a la(s) 4 38 16 p  m

I left a suggestion for the code and don't forget to add the changelog entry.

lucanovera (approved):
Approved 💯

PR #6078, opened 2025-04-23
HA-639 - Added ConnectionType.okta, OktaSchema, OktaConnector as support for the Okta Monitor

andres-torres-marroquin:
Closes [HA-639]

### Description Of Changes

Added ConnectionType.okta, OktaSchema, OktaConnector as support for the Okta Monitor

### Code Changes

### Steps to Confirm
These are the same steps as #6100 

1.  Enable Okta beta flag on the About page
2. Go to integrations
3. Create an Okta Integration
4. Use the credentials on 1password found under `Okta Fidesctl Integration Test Token`
5. Click on test connection

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:

[HA-639]: https://ethyca.atlassian.net/browse/HA-639?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/fides/api/alembic/migrations/versions/d0cbfec0b2dd_add_okta_to_connection_types.py
+++ b/src/fides/api/alembic/migrations/versions/d0cbfec0b2dd_add_okta_to_connection_types.py
@@ -27,7 +27,7 @@ def get_enum_values():
 
 def upgrade():
     # add 'okta' to ConnectionType enum
-    enum_values = [f"'{v.strip()}'" for v in get_enum_values() + [type_to_handle]]
+    enum_values = [f"'{value}'" for value in get_enum_values() + [type_to_handle]]
     enum_values.sort()  # Just to keep it fantastic
 
     op.execute("ALTER TYPE connectiontype RENAME TO connectiontype_old")


src/fides/api/models/connectionconfig.py
```diff
@@ -135,6 +137,7 @@ def system_type(self) -> "SystemType":
             ConnectionType.mongodb.value: SystemType.database,
             ConnectionType.mssql.value: SystemType.database,
             ConnectionType.mysql.value: SystemType.database,
+            ConnectionType.okta.value: SystemType.manual,  # Maybe implement a new system type for Okta?
```
adamsachs, line 140: yeah, that seems appropriate for sure

src/fides/api/alembic/migrations/versions/d0cbfec0b2dd_add_okta_to_connection_types.py
adamsachs, line 1: hmm. this is clever, and it does help avoid the annoyance of having to go back and find the "previous" enum state, which is pretty error-prone. 

i'll note that there's _some_ benefit in the previous approach in that it self-documents very clearly what the state of the enum will be at any given point in time, but that benefit feels pretty marginal because we also hav the enum represented in code, it's not _just_ a DB enum (those two should always be aligned). so i'm good with this 👍

src/fides/api/alembic/migrations/versions/d0cbfec0b2dd_add_okta_to_connection_types.py
```diff
@@ -0,0 +1,62 @@
...
+depends_on = None
+
+type_to_handle = "okta"
+
+
+def upgrade():
+    # add 'okta' to ConnectionType enum
+    conn = op.get_bind()
+    result = conn.execute("SELECT enum_range(NULL::connectiontype)").scalar()
+    current_values = result[1:-1].split(",")  # Remove the {} and split by comma
```
adamsachs, line 26: potential edge case (maybe less for the current migration, but if we wanted to reuse this for future migrations, as it seems is your intention): can a particular enum value have a comma in it? if so, i think it'd break this, right?

adamsachs (commented):
the new approach for connection type enum migration looks good to me, but i think there may be an edge case to consider if we hope to apply this more broadly moving forward. i'm also assuming you've given it a good test :)

src/fides/api/schemas/connection_configuration/connection_secrets_okta.py
```diff
@@ -0,0 +1,28 @@
...
+        title="Organization URL",
+        description="The URL of your Okta organization (e.g. https://your-org.okta.com)",
+    )
+    api_token: str = Field(
+        title="API Token",
+        description="The API token used to authenticate with Okta",
+        json_schema_extra={"sensitive": True},
+    )
+
+    _required_components: ClassVar[List[str]] = ["org_url", "api_token"]
```
thingscouldbeworse, line 24: What's this used for? Doesn't look like it's accessed anywhere. By virtue of `org_url` and `api_token` not being `Optional` aren't they required?

src/fides/api/alembic/migrations/versions/d0cbfec0b2dd_add_okta_to_connection_types.py
```diff
@@ -0,0 +1,61 @@
...
+
+def get_enum_values():
+    conn = op.get_bind()
+    result = conn.execute("SELECT unnest(enum_range(NULL::connectiontype))").fetchall()
+    return [row[0] for row in result]
+
+
+def upgrade():
+    # add 'okta' to ConnectionType enum
+    enum_values = [f"'{v.strip()}'" for v in get_enum_values() + [type_to_handle]]
```
thingscouldbeworse, line 30: No reason to have this as an fstring right? We could just do `str(v.strip())` or `str(v)` given the suggestion in `get_enum_values`. Small beans but I think we should keep this line as clean as possible

src/fides/api/alembic/migrations/versions/d0cbfec0b2dd_add_okta_to_connection_types.py
```diff
@@ -0,0 +1,61 @@
...
+branch_labels = None
+depends_on = None
+
+type_to_handle = "okta"
+
+
+def get_enum_values():
+    conn = op.get_bind()
+    result = conn.execute("SELECT unnest(enum_range(NULL::connectiontype))").fetchall()
+    return [row[0] for row in result]
```
thingscouldbeworse, line 25: I think we should do the `row[0].strip()` here instead of in the big list comprehension

```suggestion
    return [row[0].strip() for row in result]
```

thingscouldbeworse (approved):
In general I'm not crazy about a lot of string operations like `strip` and `', '.join` etc etc but I understand that we have had problems in the past with copy-pasting previous enum migrations so this is a fine compromise.

PR #6089, opened 2025-04-24
Lj 462 Reduce Logging for Disabled Cache on SQLAlchemy

Vagoasdf:
Closes LJ#462

### Description Of Changes

We were having trouble with unwanted SQLAlchemy log message related to caching issues with third-party dialects like Redshift.

### Code Changes

* Implemented a SQLAlchemyGeneratedFilter in sqlalchemy_filter.py to filter out specific unwanted log messages from the SQLAlchemy engine. (Thanks @galvana )
* Added a new test suite TestSQLAlchemyLogger in test_logger.py to ensure that the SQLAlchemyGeneratedFilter correctly filters out unwanted log messages.

### Steps to Confirm

1. Run the test suite in test_logger.py to ensure that the SQLAlchemyGeneratedFilter is correctly filtering out the specified log messages.

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:

### DIFF
--- a/src/fides/api/util/sqlalchemy_filter.py
+++ b/src/fides/api/util/sqlalchemy_filter.py
@@ -9,7 +9,7 @@ class SQLAlchemyGeneratedFilter(logging.Filter):
     which typically come from sqlalchemy.engine.Engine when executing cached queries.
     """
 
-    def __init__(self):
+    def __init__(self) -> None:
         super().__init__()
         # List of substrings to filter out
         self.patterns = [
@@ -18,7 +18,7 @@ class SQLAlchemyGeneratedFilter(logging.Filter):
             "does not support caching",
         ]
 
-    def filter(self, record):
+    def filter(self, record: logging.LogRecord) -> bool:
         # Only filter records from sqlalchemy.engine.Engine
         if record.name == "sqlalchemy.engine.Engine":
             message = record.getMessage()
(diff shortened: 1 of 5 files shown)


src/fides/api/util/sqlalchemy_filter.py
```diff
@@ -0,0 +1,30 @@
...
+    which typically come from sqlalchemy.engine.Engine when executing cached queries.
+    """
+
+    def __init__(self) -> None:
+        super().__init__()
+        # List of substrings to filter out
+        self.patterns = [
+            "cached since",
+            "caching disabled",
+            "does not support caching",
```
JadeCara, line 18: How certain are we that these are the only caching related substrings we need to catch? Are there any other substrings related to PII we might want to guard against (that may or may not be happening now?)

JadeCara (approved):
The tests work for all test cases. I had one question about if we are filtering strongly enough, but this will definitely solve the immediate problem. Approving.

src/fides/api/util/sqlalchemy_filter.py
Vagoasdf: The one we detected during this Issue was related to caching strategies for third party dialects. Docs over here: https://docs.sqlalchemy.org/en/14/core/connections.html#engine-thirdparty-caching point to the following structure for this error

> For all third party dialects that don’t support this attribute, the logging for such a dialect will indicate `dialect does not support caching`.

cached since and caching disabled are failsafe nets for similar error that might contain PII, but haven't showed on the logs yet

PR #6090, opened 2025-04-24
Treat non-applicable keys as defaults which then get overridden

gilluminate:
Closes [LJ-720]

### Description Of Changes

Fixed a bug in the GTM integration where duplicate notice keys between `privacy_notices` and `non_applicable_privacy_notices` arrays caused incorrect consent values to be reported. The issue occurred when custom notices were created with the same `notice_key` as existing notices.

### Code Changes

**GTM Integration Fix**: Completely reworked how consent values are initialized and applied in the GTM integration. Now it:
  1. First initializes an empty consent values object
  2. Sets defaults for non-applicable notices when needed
  3. Then overrides with actual consent values from configured notices
  4. This ensures proper prioritization even when duplicate keys exist

**Admin UI Preview Improvement**: Replaced the use of `systemsCount` with a proper `vendorCount` that is fetched from the vendor report API, providing more accurate vendor count data for previews.

### Steps to Confirm
1. Create a custom notice with the same `notice_key` as an existing notice (e.g., "data_sales_and_sharing")
2. Add only the custom notice to an experience
3. Verify the GTM integration correctly reports the consent value from the custom notice, not the default from the non-applicable notice
4. Confirm no duplicates appear in the `dataLayer` object

### Pre-Merge Checklist

* [x] Issue requirements met
* [x] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [x] No followup issues
* Database migrations:
  * [x] No migrations
* Documentation:

[LJ-720]: https://ethyca.atlassian.net/browse/LJ-720?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/clients/fides-js/src/integrations/gtm.ts
+++ b/clients/fides-js/src/integrations/gtm.ts
@@ -51,23 +51,12 @@ const pushFidesVariableToGTM = (
       nonApplicableFlagMode = GtmNonApplicableFlagMode.OMIT,
     flag_type: flagType = GtmFlagType.BOOLEAN,
   } = options ?? {};
-  const consentValues: FidesVariable["consent"] = JSON.parse(
-    JSON.stringify(consent),
-  );
+  const consentValues: FidesVariable["consent"] = {};
   const privacyNotices = window.Fides?.experience?.privacy_notices;
   const nonApplicablePrivacyNotices =
     window.Fides?.experience?.non_applicable_privacy_notices;
 
-  if (privacyNotices && flagType === GtmFlagType.CONSENT_MECHANISM) {
-    Object.entries(consent).forEach(([key, value]) => {
-      consentValues[key] = transformConsentToFidesUserPreference(
-        value,
-        privacyNotices.find((notice) => notice.notice_key === key)
-          ?.consent_mechanism,
-      );
-    });
-  }
-
+  // First set defaults for non-applicable privacy notices if needed
   if (
     nonApplicableFlagMode === GtmNonApplicableFlagMode.INCLUDE &&
     nonApplicablePrivacyNotices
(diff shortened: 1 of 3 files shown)


clients/fides-js/src/integrations/gtm.ts
```diff
@@ -78,6 +67,23 @@ const pushFidesVariableToGTM = (
     });
   }
 
+  // Then override with actual consent values
+  if (consent) {
```
tvandort, line 71: Small nit but I feel like we'd be better served by reducing the overall cyclomatic complexity of this code.

Instead of wrapping this in an if statement we could:

```
const nextConsent = consent ?? {}
if (privacyNotices && flagType === GtmFlagType.CONSENT_MECHANISM) {
  Object.entries(consent).forEach(([key, value]) => {
    consentValues[key] = transformConsentToFidesUserPreference(
      value,
      privacyNotices.find((notice) => notice.notice_key === key)
        ?.consent_mechanism,
    );
  });
} else {
  Object.entries(consent).forEach(([key, value]) => {
    consentValues[key] = value;
  });
}
```

clients/fides-js/src/integrations/gtm.ts
```diff
@@ -78,6 +67,23 @@ const pushFidesVariableToGTM = (
...
+    if (privacyNotices && flagType === GtmFlagType.CONSENT_MECHANISM) {
+      Object.entries(consent).forEach(([key, value]) => {
+        consentValues[key] = transformConsentToFidesUserPreference(
+          value,
+          privacyNotices.find((notice) => notice.notice_key === key)
+            ?.consent_mechanism,
+        );
+      });
```
tvandort, line 79: I'm wondering if we should take the time to extract a function to build the consent object that we pass an array of settings to. That function could compose the consent object from those arguments.

clients/fides-js/src/integrations/gtm.ts
gilluminate: if there's no consent, it doesn't do anything else, which I think is less cyclomatic, no?

gilluminate: Additionally, I'm betting the rollup algorythim is going to make this nice and clean anyway, and I'd favor readability in the source code.

tvandort: I find less `if` statements more readable personally. :)

Something that always happens is easier to reason about than something that _sometimes_ happens.

> if there's no consent, it doesn't do anything else, which I think is less cyclomatic, no?

No I don't believe so. At least to my understand cyclomatic complexity goes up with mutually exclusive execution paths in the code. If a path always executes it's less complex than one that does.

Let's take a simpler example case:

```js
function trim(str) {
	if (str) {
		return str.trim();
	}
	
	return "";
}
```

This code has two paths, one where a another function, `.trim()` will be executed and one where it will not. CC of 2.

```js
function trim(str) {
	var trimmedStr = str ?? "";
	return trimmedStr.trim();
}
```

This code has one path, `.trim()` is always called. Therefore (assuming I understand correctly) it has a lower cyclomatic complexity. CC of 1.

tvandort (approved):

PR #6095, opened 2025-04-28
Fides 1926

wadesdev:
Closes FIDES-1926

### Description Of Changes
Allows MySQL integration to complete Helios D&D steps for a DB using keywords in columns.

### Code Changes

* MySQLConfig updates generate_raw_query to add backticks to fields

### Steps to Confirm

1.  A MySQL DB using reserved keywords as column names can successfully complete Helios D&D

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:

### DIFF
--- a/src/fides/api/service/connectors/query_configs/mysql_query_config.py
+++ b/src/fides/api/service/connectors/query_configs/mysql_query_config.py
@@ -1,4 +1,7 @@
-from typing import List
+from typing import Any, Dict, List, Optional, Tuple
+
+from sqlalchemy import text
+from sqlalchemy.sql.elements import TextClause
 
 from fides.api.service.connectors.query_configs.query_config import SQLQueryConfig
 
@@ -7,6 +10,13 @@ class MySQLQueryConfig(SQLQueryConfig):
     """
     Generates SQL valid for MySQL
     """
+
+    def generate_raw_query(
+        self, field_list: List[str], filters: Dict[str, List[Any]]
+    ) -> Optional[TextClause]:
+        formatted_field_list = [f"`{field}`" for field in field_list]
+        raw_query = super().generate_raw_query(formatted_field_list, filters)
+        return raw_query  # type: ignore
 
     def get_formatted_query_string(
         self,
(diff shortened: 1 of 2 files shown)


src/fides/api/service/connectors/query_configs/mysql_query_config.py
```diff
@@ -8,6 +10,13 @@ class MySQLQueryConfig(SQLQueryConfig):
     Generates SQL valid for MySQL
     """
 
+    def generate_raw_query(
```
thingscouldbeworse, line 13: Does this do anything that `get_formatted_query_string` doesn't? It seems to me like we could just use `get_formatted_query_string` with the filters as `WHERE` clauses

wadesdev: @thingscouldbeworse `get_formatted_query_string` has already processed the fields so they're a string of comma separated values. To fit it in `get_formatted_query_string`, we'd need to split the string back into a list and then add backticks to each string in the list.

I stole this pattern from `snowflake_query_config.py`

thingscouldbeworse: They *should* then also already be escaped then, right? Since `get_formatted_query_string` puts backticks around `self.node.collection.name`? Or is the problem that the contents of `field_list` aren't escaped? If that's the case we should just update `get_formatted_query_string` to also escape the elements of `field_list`.

.fides/db_dataset.yml
```diff
@@ -8,21 +8,26 @@ dataset:
   - system.operations
```
thingscouldbeworse, line 8: I think we want to revert the changes to this file and the `.fides/` directory in general

wadesdev: @thingscouldbeworse for sure, I'm just being a git moron

PR #6099, opened 2025-04-30
Allow OT migration to auto write to fides preferences api

eastandwestwind:
Closes https://ethyca.atlassian.net/browse/LJ-722

### Description Of Changes

Allow OT migration to auto write to fides preferences api

### Code Changes

* Adds code to automatically write to consent prefs api when we have OT consent
* Adds test assertions

### Steps to Confirm

1. In fides-js-demo, uncomment out the `window.fides_overrides` declaration. Add a new field, `ot_fides_mapping` that will hold your OT -> fides mappings. 
Example core data: 
```
{
    C0001: ["essential"],
    C0002: ["analytics"],
    C0004: ["advertising", "marketing"],
  }
```
To prep this, you'll need to JSON.stringify, then URL-encode it.
Here's my resulting fides_overrides obj:
```
window.fides_overrides = {
          fides_embed: true,
          ot_fides_mapping: "%7B%22C0001%22%3A%5B%22essential%22%5D%2C%22C0002%22%3A%5B%22analytics%22%5D%2C%22C0004%22%3A%5B%22advertising%22%2C%22marketing%22%5D%7D"
        };
```
3. Manually add the OT cookie to your browser. Below where you set the overrides, you'll need to set:
```
document.cookie =
             "OptanonConsent=isGpcEnabled=0&datestamp=Fri+Mar+07+2025+19%3A51%3A06+GMT%2B0100+(Central+European+Standard+Time)&version=202409.1.0&browserGpcFlag=0&isIABGlobal=false&hosts=&consentId=f9c105b8-48e3-401a-87e9-363ad1f77531&interactionCount=1&isAnonUser=1&landingPath=NotLandingPage&groups=C0003%3A1%2CC0002%3A1%2CC0004%3A1%2CC0005%3A0%2CC0001%3A1%2CC0009%3A0%2CV2STACK42%3A0&AwaitingReconsent=false";
```
Look at the `groups` param. Take that value and url-decode it. These groups map to either 0 (opted out) or 1 (opted in). 
4. Based on our onetrust consent cookie and our mappings, we should see the essential / analytics / advertising / marketing notices opted-in by default, if they exist
5. We should see the `fides_consent` cookie written to the browser and consent vals should reflect the `window.Fides.consent` values. Note that the consent method in the cookie is "ot_migration"
6. We should see new entries in Consent Reporting in Admin UI for the "ot_migration" consent

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:

### DIFF
--- a/clients/fides-js/src/fides.ts
+++ b/clients/fides-js/src/fides.ts
@@ -24,11 +24,13 @@ import {
   OtToFidesConsentMapping,
   OverrideType,
   PrivacyExperience,
+  SaveConsentPreference,
 } from "./lib/consent-types";
 import {
   decodeNoticeConsentString,
   defaultShowModal,
   encodeNoticeConsentString,
+  isPrivacyExperience,
   shouldResurfaceBanner,
 } from "./lib/consent-utils";
 import {
@@ -51,7 +53,9 @@ import {
   UpdateExperienceFn,
 } from "./lib/initialize";
 import { initOverlay } from "./lib/initOverlay";
+import { updateConsentPreferences } from "./lib/preferences";
 import { renderOverlay } from "./lib/renderOverlay";
+import { transformConsentToFidesUserPreference } from "./lib/shared-consent-utils";
 import { customGetConsentPreferences } from "./services/external/preferences";
 
 declare global {
@@ -178,6 +182,7 @@ async function init(this: FidesGlobal, providedConfig?: FidesConfig) {
 
   // Check for an existing cookie for this device
   let consentFromOneTrust: NoticeConsent | undefined;
+  let hasAppliedOTConsent = false;
   if (optionsOverrides.otFidesMapping && !getFidesConsentCookie()) {
     consentFromOneTrust = readConsentFromOneTrust(config, optionsOverrides);
   }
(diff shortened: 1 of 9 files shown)


clients/fides-js/src/lib/consent-types.ts
```diff
@@ -782,6 +782,7 @@ export enum ConsentMethod {
   GPC = "gpc",
   INDIVIDUAL_NOTICE = "individual_notice",
   ACKNOWLEDGE = "acknowledge",
+  OT_MIGRATION = "ot_migration",
```
gilluminate, line 785: This feels a little too specific. If we started migrating a different competitor tomorrow, would we want to add another new method here? That's perfectly valid if we do, but want to call it out to make sure it's intentional that way.

clients/fides-js/src/fides.ts
```diff
@@ -249,6 +255,49 @@ async function init(this: FidesGlobal, providedConfig?: FidesConfig) {
   });
   Object.assign(this, updatedFides);
   updateWindowFides(this);
+
+  // Now that we have the final experience, we can save OneTrust preferences to the API, if they exist
+  if (
```
gilluminate, line 260: I think it might be work breaking all of the OT migration logic out of this file and into a dedicated file of its own. Similar to my previous comment, how much of this is specific to OT and how much of it can we make generic for future competitor support?

clients/fides-js/src/lib/consent-types.ts
eastandwestwind: This is something I'd asked Michael about, and we agreed on this naming. Our consent reporting today is very basic and does not allow us to see much granularity with various facets like source / method / etc, so I'm OK being more specific rather than the other way around here.

clients/fides-js/src/fides.ts
eastandwestwind: I think that's a great suggestion! I'll at least break out into a separate file for now

gilluminate (approved):
Looks good and works in testing. Thanks for updating to make it more maintainable through better abstraction! reminder to add a CHANGELOG entry :)

PR #6105, opened 2025-04-30
Ignore unstubbed requests

gilluminate:
Default to ignoring any request that hasn't been stubbed. Subsequent stubs will override as needed. The request gets logged in the Cypress run as "unstubbedRequest" so it's apparent while we're developing and no request ever makes it to our local server so it's a much more fool proof mimic of the GitHub Actions environment where there is no server.

Note: this is set at the Cypress e2e.ts file level, so there's no need to add this anywhere else. It's the first `beforeEach` to run, always.


clients/admin-ui/cypress/e2e/privacy-experiences.cy.ts
```diff
@@ -556,6 +556,9 @@ describe("Privacy experiences", () => {
         cy.intercept("GET", "/api/v1/experience-config/*", {
           fixture: "privacy-experiences/tcf-experience.json",
         }).as("getTCFExperience");
+        cy.intercept("GET", "/api/v1/privacy-experience/gvl/translations*", {
+          body: {},
+        }).as("getLanguage");
```
gilluminate: This was basically the only one that was causing a test failure after the default stubs were added.

lucanovera (approved):
Changes looks good! I tested it and it intercept requests without stubs correctly. I added a line for a console warn. Great improvement, approved!

PR #6122, opened 2025-05-06
Add environment variable to configure pool_recycling

wadesdev:
Closes HA-667
### Description Of Changes

This change adds the `pool_recycling` configuration variable to further control settings when connecting Fides to its postgres DB. We believe this may improve the performance issues recently seen on Surveymonkey's environment.

### Code Changes

* Update database_settings.py file to include `pool_recycling`. This change is pushed through to where the db engine is created

### Steps to Confirm

1.  Configure Fides to start with FIDES__DATABASE__POOL_RECYCLING=3600. Connections should be recycled every hour.

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:


src/fides/config/database_settings.py
```diff
@@ -43,6 +43,10 @@ class DatabaseSettings(FidesSettings):
         default=5,
         description="Maximum number of TCP keepalive retries before the client considers the connection dead and closes it.",
     )
+    api_engine_pool_recycle: int = Field(
+        default=-1,
```
tvandort, line 47: Is there documentation that you found somewhere that says the default is `-1`? I was trying to track that down myself and only kinda sorta found docs.

wadesdev: So funny enough I've been banging my head on this because of the CI error -- turns out the test is failing because of the default value in `session.py` file is None.

Found the default value here: https://docs.sqlalchemy.org/en/14/core/pooling.html#sqlalchemy.pool.Pool

PR #6124, opened 2025-05-06
LJ-745 Adding extension to file uploads

JadeCara:
Issue [LJ-745](https://ethyca.atlassian.net/browse/LJ-745?atlOrigin=eyJpIjoiYTI4Y2EzZTc3NzA2NGY1NDllZmE4ZTJkMmJkNzJhYmYiLCJwIjoiaiJ9)

### Description Of Changes

This PR enables using file extensions while uploading and retrieving documents. There were also some small usability changes associated with this. 
We can use the pre-signed urls to include document previews or provide a download link, the important distinction will be the size of the attachment so these functions now return the size of the attached file as well as the pre-signed url. These changes mean we do not need to distinguish between the retrieve and download functions so the download function was removed. 
The other requirement for this ticket is using pre-signed URLs which was already implemented. 

### Code Changes

* Added file type checking to the storage service utils
* Added file extension to the uploads and retrieval functions.
* Added tests for file type checking and verifiying extensions. 
* Updated functions to return file size rather than content or message on size being too large. 
* Updated tests across the storage/attachment/comments models etc

### Steps to Confirm

1.  All tests pass
2. The endpoints for this work are all in fidesplus so until those are updated to reflect these changes (PR in progress [here](https://github.com/ethyca/fidesplus/pull/2126) )there is no additional endpoint testing. Once that  PR is merged the test steps listed there can be used for these tests as well. 

### Pre-Merge Checklist

* [x] Issue requirements met
* [x] `CHANGELOG.md` updated
* Followup issues:
  * [x] Followup issues created
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:

[LJ-745]: https://ethyca.atlassian.net/browse/LJ-745?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/fides/api/models/attachment.py
```diff
@@ -237,8 +243,11 @@ def create(
     def delete(self, db: Session) -> None:
         """Deletes an attachment record from the database and deletes the attachment from S3."""
         self.delete_attachment_from_storage()
-        for attachment_reference in self.references:
-            attachment_reference.delete(db)
+
+        # Delete all references to the attachment
+        while self.references:
+            reference = self.references.pop()
+            reference.delete(db)
```
JadeCara, line 250: Safer delete pattern.

src/fides/api/models/attachment.py
```diff
@@ -276,5 +285,6 @@ def delete_attachments_for_reference_and_type(
             .all()
         )
 
-        for attachment in attachments:
+        while attachments:
+            attachment = attachments.pop()
```
JadeCara, line 289: Safer delete pattern.

src/fides/api/service/storage/s3.py
```diff
@@ -8,14 +8,13 @@
...
+    get_allowed_file_type_or_raise,
+)
 from fides.api.util.aws_util import get_s3_client
 from fides.config import CONFIG
 
-LARGE_FILE_THRESHOLD = 5 * 1024 * 1024  # 5 MB threshold for large files
-LARGE_FILE_WARNING_TEXT = (
-    "File is too large to display directly. Please use the provided link to download."
-)
-
```
JadeCara, line 18: Moved to utils so these can be used with GCS when that gets implemented for attachments.

src/fides/api/models/attachment.py
erosselli: I'm curious, why is this safer?

src/fides/api/service/storage/s3.py
```diff
@@ -35,7 +34,7 @@ def maybe_get_s3_client(
 
 
 def create_presigned_url_for_s3(
-    s3_client: Any, bucket_name: str, file_key: str
+    s3_client: Any, bucket_name: str, file_key: str, file_type: Optional[str] = None
```
erosselli, line 37: would it be useful to type `file_type` more strictly, either with an enum or with a union of the allowed strings?

src/fides/api/service/storage/s3.py
```diff
@@ -98,85 +100,53 @@ def generic_upload_to_s3(  # pylint: disable=R0913
...
-            )
-
-        # File is large, return warning text and a presigned URL
-        return (
-            LARGE_FILE_WARNING_TEXT,
-            create_presigned_url_for_s3(s3_client, bucket_name, file_key),
+        file_type = get_allowed_file_type_or_raise(file_key)
+        file_size = s3_client.head_object(Bucket=bucket_name, Key=file_key)[
+            "ContentLength"
+        ]
```
erosselli, line 146: we have this line in a couple of different places, do you think it's worth it to abstract it into a `get_file_size(s3_client)` function?  it's a one-liner so I'm on the fence about whether it's overkill or not

src/fides/api/service/storage/util.py
```diff
@@ -1,10 +1,46 @@
...
+}
 
 LOCAL_FIDES_UPLOAD_DIRECTORY = "fides_uploads"
 
+# Default to 5MB if not specified in environment
+DEFAULT_LARGE_FILE_THRESHOLD = 5 * 1024 * 1024  # 5 MB threshold
+# This checks to see if the environment variable is set and if it is, it uses that value
+# Otherwise, it uses the default value
+LARGE_FILE_THRESHOLD = int(
+    os.getenv("FIDES__LARGE_FILE_THRESHOLD", str(DEFAULT_LARGE_FILE_THRESHOLD))
```
erosselli, line 27: I don't think I've seen this `os.getenv` pattern a lot in our codebase, we usually use the ConfigProxys for retrieving env variables I think? I'm not an expert on the ConfigProxy but we have StorageSettingsProxy which seems appropriate here? If we use that, then it'd be possible to define this variable as part of the `fides.toml` file I think

 Also I think the naming `FIDES__LARGE_FILE_THRESHOLD` is missing a prefix , I imagine something like `FIDES__STORAGE__LARGE_FILE_THRESHOLD`

tests/ctl/models/test_attachment.py
```diff
@@ -39,20 +41,23 @@ def attachment_file(request):
...
+    assert download_path == get_local_filename(
+        f"{attachment.id}/{attachment.file_name}"
+    )
 
 
 def test_create_attachment_without_attachement_file_raises_error(db, attachment_data):
     """Test creating an attachment without an attachment file raises an error."""
-    with pytest.raises(ValueError):
+    with pytest.raises(ClientError):
         Attachment.create_and_upload(db, data=attachment_data, attachment_file=None)
```
erosselli, line 62: should we also test something related to the exception message?

src/fides/api/service/storage/util.py
JadeCara: I found that a lot of upload related things exist under security so I have it looking there for the environment variable. It is checking to see if it exists but not expecting it so it wont break any active environments where it does not exist.

src/fides/config/security_settings.py
```diff
@@ -154,6 +154,10 @@ class SecuritySettings(FidesSettings):
         default=10,
         description="The timeout in seconds for tunnel connection (open_channel timeout)",
     )
+    large_file_threshold: Optional[int] = Field(
+        default=None,
```
erosselli, line 158: maybe we can add the default here! instead of having it in `src/fides/api/service/storage/util.py`

tests/ops/service/storage/test_s3.py
```diff
@@ -7,318 +7,425 @@
...
-        "region_name": "us-east-1",
-    }
-    file_key = "test-file"
-    auth_method = storage_config.details[StorageDetails.AUTH_METHOD.value]
-    document = b"This is a test document."
-
-    with pytest.raises(Exception) as e:
-        generic_upload_to_s3(
-            storage_secrets=storage_secrets,
+    def test_upload_with_different_file_types(
```
erosselli, line 146: I feel like this is a dumb question but where in this test do we use the different file types?

tests/ops/service/storage/test_s3.py
```diff
@@ -7,318 +7,425 @@
...
+                storage_secrets=storage_config.secrets,
+                bucket_name=bucket_name,
+                file_key=file_key,
+                auth_method=auth_method,
+                document=document,
+            )
+
+            assert file_size == len(copy_document)
+            assert bucket_name in presigned_url
+        else:
```
erosselli, line 249: nit: I'm not a huge fan of having the success and the error cases in the same test. I feel it's cleaner to just split this into two tests, `test_upload_with_not_allowed_file_types` and `test_upload_with_allowed_file_types`.

src/fides/api/service/storage/util.py
```diff
@@ -1,10 +1,54 @@
...
 LOCAL_FIDES_UPLOAD_DIRECTORY = "fides_uploads"
 
+# Default to 5MB if not specified in environment
+DEFAULT_LARGE_FILE_THRESHOLD = 5 * 1024 * 1024  # 5 MB threshold
+# This checks to see if the environment variable is set and if it is, it uses that value
+# Otherwise, it uses the default value - corresponds to FIDES__STORAGE__LARGE_FILE_THRESHOLD
+LARGE_FILE_THRESHOLD = (
+    CONFIG.security.large_file_threshold or DEFAULT_LARGE_FILE_THRESHOLD
+)
+
```
erosselli, line 36: I wonder if security is the best place for this... I imagine at some point we'll want to group all storage-related settings under the same place, but I'm ok with ticketing that for some future work at some point since creating `CONFIG.storage` might be too much scope for this PR :)

erosselli (approved):
Thanks for making the changes -- left some nits but overall looking good! Thanks for the very thorough unit testing, as usual 🙂

tests/ops/service/storage/test_s3.py
JadeCara: Oh - yeah this name is confusing. A FastApi upload file is of `type(attachment) = SpooledTemporaryFile` when we have a larger file we use chunking which changes it into `type(attachment) = BytesIO` for each chunk of the attachment. This test just makes sure that the upload works with both types interchangeably... so not extension types, data types. I will update for clarity!

src/fides/config/security_settings.py
JadeCara: I am going to wait having the environment variable override  for now. I will add that separately.  
I added it to this [ticket](https://ethyca.atlassian.net/browse/LJ-489?atlOrigin=eyJpIjoiNTc3ZTcxYTU3YjBiNGIzY2E4MDZiMDc0OGQ5MDQ0NjYiLCJwIjoiaiJ9) so it wont get lost.

src/fides/api/service/storage/util.py
JadeCara: I think we may want to make storage a thing if we have more than one env var. I am going to kick this can down the road a bit for now. - Below I added this to another ticket so the work is still tracked.

PR #6133, opened 2025-05-08
FIDES-1977: Overhaul fides-privacy-center logging (JSON logging, configurable log levels) and add option to throw error when server-side experience lookup fails

tvandort:
Closes [FIDES-1977]

### Description Of Changes

Allow different empty experience fallback mechanisms.
Improve logging in Privacy Center.

### Steps to Confirm

1. Use cookiehouse while the Fides API isn't running with the following set:
```
FIDES_PRIVACY_CENTER__MISSING_EXPERIENCE_BEHAVIOR=throw
FIDES_PRIVACY_CENTER__LOG_LEVEL=debug
```
Observe that the PC 500s and logs per retry are listed.

2. Use cookiehouse with the Fides API running with the following set:
```
FIDES_PRIVACY_CENTER__MISSING_EXPERIENCE_BEHAVIOR=throw
```
Observe that everything works as it used to but 

3. Use cookiehouse with Fides API running with the following set:
```
FIDES_PRIVACY_CENTER__MISSING_EXPERIENCE_BEHAVIOR=empty_experience
```
Observe that an empty experience is returned.

4. Use cookiehouse while Fides API is not running with the following set:
```
FIDES_PRIVACY_CENTER__MISSING_EXPERIENCE_BEHAVIOR=empty_experience
```
Observe that an empty experience is returned.

5. Adjust the log level to info and observe that there are no debug logs.

6. Observe that each log made during a HTTP request logs a request ID that is the same across messages.

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:

[FIDES-1977]: https://ethyca.atlassian.net/browse/FIDES-1977?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/clients/fides-js/src/services/api.ts
+++ b/clients/fides-js/src/services/api.ts
@@ -28,14 +28,14 @@ interface FetchExperienceOptions {
   apiOptions?: FidesApiOptions | null;
   propertyId: string | null | undefined;
   requestMinimalTCF?: boolean;
-  missingExperienceHandler?: () => Record<string, never>;
+  missingExperienceHandler?: (error: unknown) => Record<string, never>;
 }
 
 /**
  * We allow this to be overriden in cases where falling back to an
  * empty experience is undesirable.
  */
-function createEmptyExperience() {
+export function createEmptyExperience() {
   return {};
 }
 
@@ -63,7 +63,7 @@ export const fetchExperience = async <T = PrivacyExperience>({
       );
     } catch (e) {
       fidesDebugger("Error fetching experience from custom API. Error: ", e);
-      return missingExperienceHandler();
+      return missingExperienceHandler(e);
     }
   }
 
@@ -117,7 +117,7 @@ export const fetchExperience = async <T = PrivacyExperience>({
     }
   } catch (error) {
     fidesDebugger("Error getting experience from Fides API. Error:", error);
-    return missingExperienceHandler();
+    return missingExperienceHandler(error);
   }
 
   try {
@@ -145,7 +145,7 @@ export const fetchExperience = async <T = PrivacyExperience>({
       "Error parsing experience response body from Fides API. Response:",
       response,
     );
-    return missingExperienceHandler();
+    return missingExperienceHandler(e);
   }
 };
 
--- a/clients/fides-js/src/services/api.ts
+++ b/clients/fides-js/src/services/api.ts
@@ -44,13 +44,17 @@ export const fetchExperience = async <T = PrivacyExperience>({
 }: FetchExperienceOptions): Promise<T | EmptyExperience> => {
   if (apiOptions?.getPrivacyExperienceFn) {
     fidesDebugger("Calling custom fetch experience fn");
-
-    return apiOptions.getPrivacyExperienceFn<T>(
-      userLocationString,
-      // We no longer support handling user preferences on the experience using fidesUserDeviceId.
-      // For backwards compatibility, we keep fidesUserDeviceId in fn signature but pass in null here.
-      null,
-    );
+    try {
+      return await apiOptions.getPrivacyExperienceFn<T>(
+        userLocationString,
+        // We no longer support handling user preferences on the experience using fidesUserDeviceId.
+        // For backwards compatibility, we keep fidesUserDeviceId in fn signature but pass in null here.
+        null,
+      );
+    } catch (e) {
+      fidesDebugger("Error fetching experience from custom API. Error: ", e);
+      throw e;
+    }
   }
 
   const headers = [
@@ -102,34 +106,36 @@ export const fetchExperience = async <T = PrivacyExperience>({
       throw new Error("Error fetching experience from Fides API");
     }
   } catch (error) {
-    fidesDebugger(
-      "Error getting experience from Fides API, returning {}. Error:",
-      error,
-    );
-    return {};
+    fidesDebugger("Error getting experience from Fides API. Error:", error);
+    throw error;
   }
 
   try {
     const body = await response.json();
+
     // returning empty obj instead of undefined ensures we can properly cache on
     // server-side for locations that have no relevant experiences.
+    if (body.items?.length < 1) {
+      return {};
+    }
+
     const experience:
       | PrivacyExperience
       | PrivacyExperienceMinimal
-      | EmptyExperience = (body.items && body.items[0]) ?? {};
+      | EmptyExperience = body.items && body.items[0];
 
     const firstLanguage =
       experience.experience_config?.translations?.[0].language;
     fidesDebugger(
-      `Recieved ${experience.minimal_tcf ? "minimal TCF " : ""}experience response from Fides API${experience.minimal_tcf ? ` (${firstLanguage})` : ""}`,
+      `Received ${experience.minimal_tcf ? "minimal TCF " : ""}experience response from Fides API${experience.minimal_tcf ? ` (${firstLanguage})` : ""}`,
     );
     return experience as T;
   } catch (e) {
     fidesDebugger(
-      "Error parsing experience response body from Fides API, returning {}. Response:",
+      "Error parsing experience response body from Fides API. Response:",
       response,
     );
-    return {};
+    throw e;
   }
 };
 
@@ -163,7 +169,7 @@ export const fetchGvlTranslations = async (
   }
   const gvlTranslations: GVLTranslations = await response.json();
   fidesDebugger(
-    `Recieved GVL languages response from Fides API (${
+    `Received GVL languages response from Fides API (${
       Object.keys(gvlTranslations).length
     })`,
     gvlTranslations,
--- a/clients/fides-js/src/services/api.ts
+++ b/clients/fides-js/src/services/api.ts
@@ -44,20 +44,13 @@ export const fetchExperience = async <T = PrivacyExperience>({
 }: FetchExperienceOptions): Promise<T | EmptyExperience> => {
   if (apiOptions?.getPrivacyExperienceFn) {
     fidesDebugger("Calling custom fetch experience fn");
-    try {
-      return await apiOptions.getPrivacyExperienceFn<T>(
-        userLocationString,
-        // We no longer support handling user preferences on the experience using fidesUserDeviceId.
-        // For backwards compatibility, we keep fidesUserDeviceId in fn signature but pass in null here.
-        null,
-      );
-    } catch (e) {
-      fidesDebugger(
-        "Error fetching experience from custom API, returning {}. Error: ",
-        e,
-      );
-      return {};
-    }
+
+    return apiOptions.getPrivacyExperienceFn<T>(
+      userLocationString,
+      // We no longer support handling user preferences on the experience using fidesUserDeviceId.
+      // For backwards compatibility, we keep fidesUserDeviceId in fn signature but pass in null here.
+      null,
+    );
   }
 
   const headers = [
--- a/clients/fides-js/src/services/api.ts
+++ b/clients/fides-js/src/services/api.ts
@@ -28,6 +28,15 @@ interface FetchExperienceOptions {
   apiOptions?: FidesApiOptions | null;
   propertyId: string | null | undefined;
   requestMinimalTCF?: boolean;
+  missingExperienceHandler?: () => Record<string, never>;
+}
+
+/**
+ * We allow this to be overriden in cases where falling back to an
+ * empty experience is undesirable.
+ */
+function createEmptyExperience() {
+  return {};
 }
 
 /**
@@ -41,6 +50,7 @@ export const fetchExperience = async <T = PrivacyExperience>({
   apiOptions,
   propertyId,
   requestMinimalTCF,
+  missingExperienceHandler = createEmptyExperience,
 }: FetchExperienceOptions): Promise<T | EmptyExperience> => {
   if (apiOptions?.getPrivacyExperienceFn) {
     fidesDebugger("Calling custom fetch experience fn");
@@ -53,7 +63,7 @@ export const fetchExperience = async <T = PrivacyExperience>({
       );
     } catch (e) {
       fidesDebugger("Error fetching experience from custom API. Error: ", e);
-      throw e;
+      return missingExperienceHandler();
     }
   }
 
@@ -107,7 +117,7 @@ export const fetchExperience = async <T = PrivacyExperience>({
     }
   } catch (error) {
     fidesDebugger("Error getting experience from Fides API. Error:", error);
-    throw error;
+    return missingExperienceHandler();
   }
 
   try {
@@ -135,7 +145,7 @@ export const fetchExperience = async <T = PrivacyExperience>({
       "Error parsing experience response body from Fides API. Response:",
       response,
     );
-    throw e;
+    return missingExperienceHandler();
   }
 };
 
--- a/clients/privacy-center/app/server-environment.ts
+++ b/clients/privacy-center/app/server-environment.ts
@@ -31,7 +31,9 @@ import {
 
 export type PrivacyCenterServerSettings = Pick<
   PrivacyCenterSettings,
-  "SERVER_SIDE_FIDES_API_URL" | "FIDES_JS_MAX_AGE_SECONDS"
+  | "SERVER_SIDE_FIDES_API_URL"
+  | "FIDES_JS_MAX_AGE_SECONDS"
+  | "MISSING_EXPERIENCE_BEHAVIOR"
 >;
 
 /**
@@ -283,6 +285,7 @@ export const loadServerSettings = (): PrivacyCenterServerSettings => {
     SERVER_SIDE_FIDES_API_URL:
       settings.SERVER_SIDE_FIDES_API_URL || settings.FIDES_API_URL,
     FIDES_JS_MAX_AGE_SECONDS: settings.FIDES_JS_MAX_AGE_SECONDS,
+    MISSING_EXPERIENCE_BEHAVIOR: settings.MISSING_EXPERIENCE_BEHAVIOR,
   };
 
   return serverSideSettings;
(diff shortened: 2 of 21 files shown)


clients/privacy-center/pages/api/fides-js.ts
```diff
@@ -32,6 +35,48 @@ let lastFetched: number = 0;
...
+    return await retry(() => fetchExperience(...args), {
+      delay: PREFETCH_RETRY_DELAY,
+      retries: PREFETCH_MAX_RETRIES,
+    });
+  } catch (error) {
+    // eslint-disable-next-line no-console
+    console.error(
+      "Privacy Center was not able to prefetch an experience.",
+      error,
+    );
```
NevilleS, line 75: 👌 thanks for the error log, that's preferable.

Just to be clear though, this changes the behaviour to not return an empty experience.

NevilleS (changes requested):

clients/privacy-center/pages/api/fides-js.ts
tvandort: I really don't think we ever _want_ to return an empty experience. Originally I was going to return empty after N+1 tries but the more I thought about it the less comfortable I was with doing that.

clients/fides-js/src/services/api.ts
```diff
@@ -44,20 +44,13 @@ export const fetchExperience = async <T = PrivacyExperience>({
 }: FetchExperienceOptions): Promise<T | EmptyExperience> => {
   if (apiOptions?.getPrivacyExperienceFn) {
     fidesDebugger("Calling custom fetch experience fn");
-    try {
```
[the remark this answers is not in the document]
tvandort: Ah yeah, I missed that. 

I'll go remove the try-catch that I care more about but I really think neither of these should have this behavior.

clients/privacy-center/pages/api/fides-js.ts
tvandort: On further review I think we want to return an empty experience ONLY when the experiences endpoint returns an empty array.

clients/fides-js/src/services/api.ts
```diff
@@ -109,34 +116,36 @@ export const fetchExperience = async <T = PrivacyExperience>({
...
       | PrivacyExperience
       | PrivacyExperienceMinimal
-      | EmptyExperience = (body.items && body.items[0]) ?? {};
+      | EmptyExperience = body.items && body.items[0];
 
     const firstLanguage =
       experience.experience_config?.translations?.[0].language;
     fidesDebugger(
-      `Recieved ${experience.minimal_tcf ? "minimal TCF " : ""}experience response from Fides API${experience.minimal_tcf ? ` (${firstLanguage})` : ""}`,
+      `Received ${experience.minimal_tcf ? "minimal TCF " : ""}experience response from Fides API${experience.minimal_tcf ? ` (${firstLanguage})` : ""}`,
```
lucanovera, line 140: haha good catch

clients/privacy-center/app/server-utils/loadEnvironmentVariables.ts
```diff
@@ -34,6 +44,9 @@ const loadEnvironmentVariables = () => {
     FIDES_JS_MAX_AGE_SECONDS:
       Number(process.env.FIDES_PRIVACY_CENTER__FIDES_JS_MAX_AGE_SECONDS) ||
       DEFAULT_FIDES_JS_MAX_AGE_SECONDS,
+    MISSING_EXPERIENCE_BEHAVIOR: defaultMissingExperienceBehavior(
```
lucanovera, line 47: could we follow the same pattern for env defaults as the others? that would get rid of the need for the defaultMissingExperienceBehavior function.
MISSING_EXPERIENCE_BEHAVIOR: process.env.FIDES_PRIVACY_CENTER__MISSING_EXPERIENCE_BEHAVIOR || "empty_experience"

clients/privacy-center/pages/api/fides-js.ts
```diff
@@ -32,6 +37,54 @@ let lastFetched: number = 0;
...
+    throw error;
+  },
+  empty_experience: () => {
+    return {};
+  },
+};
+
+const PREFETCH_RETRY_DELAY = 100;
+const PREFETCH_MAX_RETRIES = 10;
+async function retry<T>(
```
lucanovera, line 54: Let's consider using a package for the retry behavior. I found this one: https://www.npmjs.com/package/fetch-retry with 3.7M weekly downloads.

I think using a well tested package would be safer and easier to maintain than using custom code.

lucanovera (changes requested):
Good work! I left 1 suggestions for simplifying the code and a recommendation for using a library to handle the retry functionality.

clients/privacy-center/app/server-utils/loadEnvironmentVariables.ts
tvandort: Sadly no, the type of `process.env.FIDES_PRIVACY_CENTER__MISSING_EXPERIENCE_BEHAVIOR` is `string | undefined` and the `||` trick makes the resulting type `string | "empty_experience"` instead which is similarly wrong.

tvandort: Somewhat related... is there a reason we don't use `??` as our common pattern here? Feels like `||` might catch someone off guard someday with slightly unexpected behavior.

lucanovera: Oh I see. In that case, I have a tip: If you're sure of the type like in this case, you can force typescript to cast to a different type by using **as**. In this case it would be: `(process.env
        .FIDES_PRIVACY_CENTER__MISSING_EXPERIENCE_BEHAVIOR as MissingExperienceBehavior)`

clients/privacy-center/pages/api/fides-js.ts
tvandort: I like the idea of using an existing package, however the one you linked operates directly on the fetch library and produces a fetch client that retries. 

The function that I added retries a function.

If I were to use `fetch-retry` I would need to modify the behavior (and size etc) of `fides-js` instead of just applying this new behavior to Privacy Center.

If we'd like to make calls in `fides-js` resilient then I think we should have another ticket for it.

Alternatively is there a _function_ retry library that you'd like to see used in the Privacy Center?

clients/privacy-center/app/server-utils/loadEnvironmentVariables.ts
lucanovera: @tvandort Yeah you're right, by using || an explicit false value would get overriden by the default. But in this case, since every end variable is a string, booleans are handled by checking whether something === "false" or === "true", so it doesn't have that issue. Still, I would like to come back to this file later and clean it up.

clients/privacy-center/pages/api/fides-js.ts
lucanovera: Oh I see, we're using the fetchExperience of the fides-js. In that case, maybe we can use this one: https://www.npmjs.com/package/p-retry It's for retrying functions that return Promises (such as fetchExperience.)

clients/privacy-center/app/server-utils/loadEnvironmentVariables.ts
tvandort: > Oh I see. In that case, I have a tip: If you're sure of the type like in this case, you can force typescript to assert to a different type by using **as**. In this case it would be: `(process.env .FIDES_PRIVACY_CENTER__MISSING_EXPERIENCE_BEHAVIOR as MissingExperienceBehavior)`

Wouldn't doing that allow other strings through as well?

[Example:](https://www.typescriptlang.org/play/?#code/MYewdgzgLgBAbgQwDYFcCmAuG0BOBLMAcxgB8YUwATNAMwLUpgF4YAiMEKAfVBxzWBRWAbgBQoqAE8ADmhgAVBDkJoo8mXJate-Qa1JtkEEDxB8BQsaEiwKEBDTSm0OYHnBZFy1etnN4yOgwCBAKSipqGmKi1sZIaAB0SCCEABR2Dk6gLm7gAJQwAPSFMAAK+GBQoQAG7Jym5nrVADQwAEYosFAAFnJSfnihXhG+iUA)

```
const value: string | undefined = "not_correct";

type TargetType = "correct" | "also_correct";
const unsafe_coercion: TargetType = value as TargetType;

console.log(unsafe_coercion) // Prints `"not_correct"`, but the type is TargetType.
```

clients/privacy-center/pages/api/fides-js.ts
tvandort: Aha, I'll take a look at this. Always down to use something by Sindre.

clients/privacy-center/app/server-utils/loadEnvironmentVariables.ts
tvandort: > @tvandort Yeah you're right, by using || an explicit false value would get overriden by the default. But in this case, since every env variable is a string, booleans are handled by checking whether something === "false" or === "true", so it doesn't have that issue. Still, I would like to come back to this file later and clean it up.

Thanks! Another example of a case where this could go wrong is with: 

```
FIDES_JS_MAX_AGE_SECONDS:
      Number(process.env.FIDES_PRIVACY_CENTER__FIDES_JS_MAX_AGE_SECONDS) ||
      DEFAULT_FIDES_JS_MAX_AGE_SECONDS,
```

You wouldn't be able to set this to `0` if you wanted to.

lucanovera: Yes, that's correct. Currently, passing in invalid values as env variables would result in unexpected behavior. Runtime checking and throwing an error would be safer. We need a follow up ticket for refactoring this file into a standard pattern for handling env variable loading and doing runtime checking.

clients/privacy-center/app/server-utils/loggerContext.ts
```diff
@@ -0,0 +1,23 @@
+import { NextApiRequest } from "next";
+
+export const createLoggingContext = (request: NextApiRequest) => {
+  const id = request.headers["x-request-id"] ?? "null";
+  const logSuffix = `| { "Request-Id": ${id} }`;
+  const debug = (...args: Parameters<typeof console.debug>) =>
+    // eslint-disable-next-line no-console
+    console.debug(...args, logSuffix);
```
NevilleS, line 8: So I just realized (much too late!) that this PR will *fix* debug logging for Privacy Center in `/fides.js` endpoint, which currently don't get enabled in production builds. That's good 👍

However, that'll also generate a lot of debug logs and customers might not want that level of verbosity 👎 

So unfortunately I think we need to continue doing the Right Thing here and introduce `FIDES_PRIVACY_CENTER__LOG_LEVEL` to control log levels. We don't need have a _ton_ of logging today, so I think it's a pretty manageable change, but it is kind of _piling on_ to this one, sorry!

NevilleS: So in other words, I think we roll a _slightly better_ logger implementation like this:
```typescript
// Designed to be forwards-compatible with Pino: https://github.com/pinojs/pino/blob/main/docs/api.md#loggerlevels-object
const LOG_LEVELS = { fatal: 60, error: 50, warn: 40, info: 30, debug: 20, trace: 10 };

export const createLoggingContext = (request: NextApiRequest) => {
  // Set default level
  const logLevel: number = getLogLevelFromLabel(loadServerSettings().LOG_LEVEL);

  // ...other stuff

  // define wrapped log functions
  const debug = (...args: Parameters<typeof console.debug>) =>
    if (logLevel >= LOG_LEVELS.debug) {
      console.debug("DEBUG | ", ...args, logSuffix);
    }
    // no-op, log disabled
  }
};
```

However, I really want us to assume we should plan to adopt Pino, so I'd ask we try to plan a bit of "forwards-compatibility" with that approach to ensure we can upgrade the PC logging overall without too much hassle

NevilleS (changes requested):
Ugh, I apologize for this "request changes", knowing that we're piling QoL asks onto a request that was originally fairly simple 👍

clients/privacy-center/pages/api/fides-js.ts
tvandort: Switched to p-retry. We may want to tweak some of the values I've set up but I think what I've configured works for now.

clients/privacy-center/app/server-utils/loggerContext.ts
tvandort: I'd prefer that if we were intending on using pino to just implement pino instead of trying to mock the API. So I implemented pino in this PR and replaced all the non-shared debug and error calls.

lucanovera (approved):
Thanks for re-implementing with the library. Just tested this and found the logging, retry, throw/empty_experience all to be working as expected. Approved.

clients/privacy-center/pages/api/fides-js.ts
```diff
@@ -178,31 +201,60 @@ export default async function handler(
...
+              );
+            },
+          },
+        );
+        log.debug(
+          `Fetched relevant experiences from server-side (${userLanguageString}).`,
+        );
+        experienceIsValid(experience);
+      } catch (error) {
+        log.error(error);
```
NevilleS, line 248: Will this be a good error message? I'm assuming probably not by default (something super specific) so I'd do this:

```suggestion
        log.error("Error fetching experience from server-side Fides API. Error: ", error);
```

NevilleS (changes requested):
Couple nits and a few extra log requests 👍 

Main blocker though is to not use pino-pretty I have a feeling that'll gum up server logs

clients/privacy-center/app/server-utils/loggerContext.ts
```diff
@@ -0,0 +1,22 @@
+import { NextApiRequest } from "next";
+import pino from "pino";
+
+import { loadServerSettings } from "../server-environment";
+
+export const createLogger = () => {
+  const settings = loadServerSettings();
+  const logger = pino({
+    level: settings.LOG_LEVEL,
+    transport: { target: "pino-pretty" },
```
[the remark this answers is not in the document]
tvandort: Missed that. Good call.

clients/privacy-center/middleware.ts
```diff
@@ -1,14 +1,28 @@
...
   const logDict = {
     method: request.method,
     path: request.nextUrl.pathname,
+    requestId,
   };
 
   const debug = process.env.FIDES_PRIVACY_CENTER__DEBUG === "true";
   if (debug) {
     /* eslint-disable no-console */
     console.info(JSON.stringify(logDict));
```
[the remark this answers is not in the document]
tvandort: Feels like it should be `log.debug` without `privacy_center__debug` needing to be true.

clients/privacy-center/app/server-environment.ts
```diff
@@ -31,7 +31,10 @@ import {
 
 export type PrivacyCenterServerSettings = Pick<
   PrivacyCenterSettings,
-  "SERVER_SIDE_FIDES_API_URL" | "FIDES_JS_MAX_AGE_SECONDS"
+  | "SERVER_SIDE_FIDES_API_URL"
+  | "FIDES_JS_MAX_AGE_SECONDS"
+  | "MISSING_EXPERIENCE_BEHAVIOR"
+  | "LOG_LEVEL"
```
NevilleS, line 37: nit: There's some `console.debug`, `console.warn`, etc. usage in here that should swap over to the real logger, for consistency 👍

NevilleS: Oof, realizing this ask would create a circular dependency between server environment and logger modules.

clients/privacy-center/pages/api/fides-js.ts
```diff
@@ -137,7 +160,7 @@ export default async function handler(
       fidesString,
     );
   } catch (error) {
-    fidesError(error);
+    log.error(error);
```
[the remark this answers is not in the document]
tvandort: Handled.

clients/privacy-center/pages/api/fides-js.ts
```diff
@@ -178,31 +201,60 @@ export default async function handler(
       const userLanguageString =
         fidesLocale || req.headers["accept-language"] || DEFAULT_LOCALE;
 
-      fidesDebugger(
+      log.debug(
         `Fetching relevant experiences from server-side (${userLanguageString})...`,
       );
```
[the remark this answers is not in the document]
tvandort: Handled & structured output.

clients/privacy-center/pages/api/fides-js.ts
```diff
@@ -178,31 +201,60 @@ export default async function handler(
...
+              log.debug(
+                `Attempt to get privacy experience failed, ${error.retriesLeft} remain. Error message was: `,
+                error.message,
+              );
+            },
+          },
+        );
+        log.debug(
+          `Fetched relevant experiences from server-side (${userLanguageString}).`,
+        );
```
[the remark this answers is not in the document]
tvandort: Done.

clients/privacy-center/pages/api/fides-js.ts
tvandort: Done.

clients/privacy-center/pages/api/fides-js.ts
```diff
@@ -343,7 +393,6 @@ export default async function handler(
   `;
 
   // Instruct any caches to store this response, since these bundles do not change often
-  const serverSettings = loadServerSettings();
   const cacheHeaders: CacheControl = {
```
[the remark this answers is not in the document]
tvandort: Flip the parameters but yes, this works.

NevilleS:
Awesome - taking one more look at this now.

NevilleS (approved):
This is great - we have some issues with the fides-privacy-center Vercel deployment, but that's almost certainly a red herring.

PR #6134, opened 2025-05-08
Update to iabtcf v1.5.15

gilluminate:
Closes [LJ-415]

### Description Of Changes

Our `package-lock.json` file was already updated to the `1.5.15` version, but this change updates the `package.json` to be more precise and specific about versions so we don't accidentally upgrade when we're not ready in the future.

Version `1.5.15` also includes a fix that we need to tweak our own CMP code to support. That fix is also included in this PR

### Code Changes

* Add static version to iabtechlabtcf packages
* Remove checks for vendors with purpose consent when considering adding Special Purpose vendors to Legitimate Interest list.

### Steps to Confirm

1. In Admin UI remove all vendors, and add in the following vendors:
   - "SeenThis" `gvl.415` (purpose consent: no, LI: no, special purposes: yes)
   - "Adhese" `gvl.553` (purpose consent: yes, LI: no, special purposes: yes)
   - "M6 Publicite" `gvl.1407` (purpose consent: yes, LI: no, special purposes: no)
1. Enable the TCF experience
1. In the PC demo page for TCF, Opt in to all
1. Grab the TCF String from the cookie and plug it in to https://iabgpp.com/
1. Verify that the Vendor Consents list includes both `553` and `1407`
1. Verify that the Vendor Legitimate Interests list includes both `415` and `553`
1. Note that because "Adhese" `553` has both purpose consent and special interests it correctly appears in both lists.
1. Delete the cookie and reload the demo page and Opt out of all
1. Grab the TCF String from the cookie and plug it in to https://iabgpp.com/
1. Verify that the Vendor Consents list is empty
1. Verify that the Vendor Legitimate Interests list still includes both `415` and `553`

### Pre-Merge Checklist

* [x] Issue requirements met
* [x] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [x] No followup issues
* Database migrations:
  * [x] No migrations
* Documentation:

[LJ-415]: https://ethyca.atlassian.net/browse/LJ-415?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/clients/fides-js/__tests__/lib/tcf/tcf.test.ts
+++ b/clients/fides-js/__tests__/lib/tcf/tcf.test.ts
@@ -477,20 +477,33 @@ describe("generateFidesString", () => {
     expect(decodedTCString.publisherCountryCode).toBe("US");
   });
 
-  it("saves special purpose only vendor to legitimate interest regardless of consent choice", async () => {
-    // Add a vendor that only has special purposes
-    const experienceWithSpecialPurposeVendor = {
+  it("saves special purpose vendor to legitimate interest appropriately", async () => {
+    // This test only check for opt out, but is applicable regardless of consent choice
+    const experienceWithSpecialPurposeVendors = {
       ...experience,
       tcf_vendor_relationships: [
         {
+          id: "gvl.777",
+          name: "Vendor with Purposes but no Special Purposes",
+          special_purposes: [],
+          features: [],
+          special_features: [],
+          url: "https://test.com/privacy",
+        },
+        {
+          id: "gvl.888",
+          name: "Special Purpose Vendor with Purposes",
+          special_purposes: [1, 2, 3],
+          features: [],
+          special_features: [],
+          url: "https://test.com/privacy",
+        },
+        {
           id: "gvl.999",
           name: "Special Purpose Only Vendor",
-          purposes: [],
-          legitimate_interest_purposes: [],
           special_purposes: [1, 2],
           features: [],
           special_features: [],
-          flexible_purposes: [],
           url: "https://test.com/privacy",
         },
       ],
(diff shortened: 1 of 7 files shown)


clients/fides-js/src/lib/tcf.ts
```diff
@@ -161,25 +161,22 @@ export const generateFidesString = async ({
...
               vendor &&
-              vendor.specialPurposes?.length &&
-              (!vendor.purposes || vendor.purposes.length === 0) &&
-              (!vendor.flexiblePurposes ||
-                vendor.flexiblePurposes.length === 0) &&
-              (!vendor.legIntPurposes || vendor.legIntPurposes.length === 0) &&
-              !isInVendorConsents
+              (vendor.purposes.length === 0 || vendor.purposes.length > 0) &&
+              vendor.legIntPurposes.length === 0 &&
+              nonForbiddenSpecialPurposes.length > 0
```
gilluminate, line 179: This logic aligns with the logic now (as of 1.5.15) used to validate in the api https://github.com/InteractiveAdvertisingBureau/iabtcf-es/pull/472/files#diff-a213d54b9e29c153dab07ac82bbce45e7b5a5177b7c1d38bdcc836fab9ed5dc8R57-R74

clients/fides-js/__tests__/lib/tcf/tcf.test.ts
```diff
@@ -477,34 +477,77 @@ describe("generateFidesString", () => {
...
-    // Add a vendor that only has special purposes
-    const experienceWithSpecialPurposeVendor = {
+  it("saves special purpose vendor to legitimate interest appropriately", async () => {
+    // This test only check for opt out, but is applicable regardless of consent choice
+    const experienceWithSpecialPurposeVendors = {
       ...experience,
       tcf_vendor_relationships: [
+        {
+          id: "gvl.777",
+          name: "Vendor with Purposes but no Special Purposes",
```
erosselli, line 487: can we also add a `"Speical Purpose Vendor with Leg Int Purposes"` vendor?  it would have  some purposes in the `legIntPurposes` array in the `gvl` object below

clients/fides-js/src/lib/tcf.ts
erosselli: for my own understanding: what would be a "forbidden special purpose" ?  also, why do we not need to check `isInVendorConsents` any more?

erosselli, line 179: I don't fully understand this change, why would ` (vendor.purposes.length === 0 || vendor.purposes.length > 0)` work here?

gilluminate: Much like the Legitimate Interest list, the special purpose list can only contain flexible purposes, so this just does a check to make sure it includes at least one flexible purpose (non-forbidden)

gilluminate: Previously, a vendor with Special Interests would only get added to the Vendor Legitimate Interst list if they didn't have any Purpose Consents. That turned out to be a bug and it now adds vendors with empty OR populated purposes as long as the special interests are flexible purposes. That's also why we no longer need to check `!isInVendorConsents` anymore.

gilluminate: See the description of their bug fix https://github.com/InteractiveAdvertisingBureau/iabtcf-es/pull/472 for more context.

gilluminate: @eastandwestwind I wasn't sure why that was added. I don't see any documentation about flexible purposes relating to special purposes -> LI. Can you point me to why that came in to play in the first place?

clients/fides-js/__tests__/lib/tcf/tcf.test.ts
gilluminate: done!

PR #6139, opened 2025-05-09
Adding certifi to fix SSL CA Error

Vagoasdf:
Closes[ LJ-728](https://ethyca.atlassian.net/browse/LJ-728?atlOrigin=eyJpIjoiMDkyMGFkOWVmMjEzNDcyOThkNzhkZjhkNTdkZjg2YTgiLCJwIjoiamlyYS1zbGFjay1pbnQifQ)

### Description Of Changes

We were having trouble hitting Shopify subdomains recently, with the integration returning an SSL Error of an invalid CA. 
We confirmed that from the server, we were able to access those endpoints, so it was a problem with Fides not getting the Certificates correctly. 

This should point which certificates to use directly. 

### Code Changes

* Using [Certifi](https://pypi.org/project/certifi/) to get the proper CA 

### Steps to Confirm

1.  DSRs where we had problems with SSL should pass 

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,6 +5,7 @@ anyio==3.7.1
 APScheduler==3.9.1.post1
 asyncpg==0.27.0
 boto3==1.26.1
+certifi==2024.8.30
 celery[pytest]==5.2.7
 click_default_group==1.2.2
 cloud-sql-python-connector==1.9.2
--- a/src/fides/api/service/connectors/saas/authenticated_client.py
+++ b/src/fides/api/service/connectors/saas/authenticated_client.py
@@ -50,6 +50,7 @@ class AuthenticatedClient:
         rate_limit_config: Optional[RateLimitConfig] = None,
     ):
         self.session = Session()
+        self.session.verify = certifi.where()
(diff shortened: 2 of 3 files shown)


requirements.txt
```diff
@@ -5,6 +5,7 @@ anyio==3.7.1
 APScheduler==3.9.1.post1
 asyncpg==0.27.0
 boto3==1.26.1
+certifi==2024.8.30
```
galvana, line 8: @Vagoasdf is there a reason we're not using the latest?

Vagoasdf: This one was compatible with Requests `2.32.3` which i believe is the one we are currently running (throught FastAPI)

galvana: Makes sense, thanks for the explanation!

galvana (approved):

PR #6149, opened 2025-05-15
Forms POC

gilluminate:
Experiments for Formik vs Ant Form components


lucanovera (approved):
Thanks for creating this PR! This will be a good reference for new forms and/or changes to existing Formik forms. The cypress tests had a weird failure, it should be re-running now.

PR #6152, opened 2025-05-15
Fail if docs are different.

tvandort:


clients/fides-js/docs/interfaces/Fides.md
```diff
@@ -9,6 +9,8 @@ FidesJS integrations (e.g. `Fides.gtm()`), programmaticaly show the FidesJS
 UI (e.g. `Fides.showModal()`) and more. See the full list of properties below
 for details.
 
+bla bla bla
+
```
RobertKeyser, line 13: is this supposed to be in there?

tvandort: Not long term! I added this to prove that it was working. Will remove before merge. But I need a ticket # anyway.

PR #6006, opened 2025-04-07
HA-524 - Fides should support Google Cloud Storage as a storage option

vcruces:
Closes [HA-524](https://ethyca.atlassian.net/browse/HA-524)

### Description Of Changes

Google Cloud Storage (GCS) can now be chosen as storage for privacy requests.

### Code Changes

- Added GCS storage configuration UI component and related type definitions
- Extended storage schemas and endpoints to support GCS integration
- Created a new method to handle interactions with GCS: uploading files and generating signed URLs for secure access.
- Added authentication service for GCS.
- Updated storage upload service to handle GCS uploads
- Added test coverage for GCS functionality

### Steps to Confirm

1.  Create a default storage for GCS. Make a PUT request to `/api/v1/storage/default` with the following body:
```
{
  "type": "gcs",
  "details": {
    "naming": "request_id",
    "auth_method": "service_account_keys",
    "bucket": "prj-sandbox-55855-test-bucket",
    "max_retries": 0
  },
  "format": "json"
}
```
2. Make a PATCH request to `/api/v1/config` with the following body:
```
{
  "storage": {
    "active_default_storage_type": "gcs"
}
```
3. Add secrets for the new storage. Make a PUT request to `/api/v1/storage/default/{storage_type}/secret` with storage_type "gcp" and the body of the request is in 1password under the name "GCS prj-sandbox-55855-test-bucket".

**To test that the storage is working:**
1. Add a system
2. Add an Integration
3. Add dataset
4. Make sure the new storage is set as the default storage.
5. The environment variable `FIDES__SECURITY__SUBJECT_REQUEST_DOWNLOAD_UI_ENABLED` must be true to be able to download the files.
6. Create a new privacy request ("access your data")
7. Access the bucket in [https://console.cloud.google.com/storage/browser/prj-sandbox-55855-test-bucket](https://console.cloud.google.com/storage/browser/prj-sandbox-55855-test-bucket) and verify that the file is there with the correct content type
8. From the admin UI, access the privacy request and download the request results. It should open a new tab in the browser with the json.

**Another way to configure storage and its secrets:**
1. Go to `/privacy-requests/configure/storage`, select GCS and fill out the form. The secret details are in 1password under the name "GCS prj-sandbox-55855-test-bucket".

### Pre-Merge Checklist

* [ ] Issue requirements met
* [x] All CI pipelines succeeded
* [x] `CHANGELOG.md` updated
[...]

PR #6161, opened 2025-05-21
LJ-530 Attachments have access to google cloud storage

JadeCara:
Closes [LJ-530](https://ethyca.atlassian.net/browse/ENG-530?atlOrigin=eyJpIjoiODQ1NGYxODY3ZDZhNDAzYWE2ZmVmYWE0YzQ4YzA5ZWUiLCJwIjoiaiJ9)

### Description Of Changes

Google Cloud storage was added as a storage option after attachments were introduced and because we are missing a single storage service at this time, this option needed to be added independently. The delete statements for all storage types were updated to be more thorough, but there should be no change to the functionality.

### Code Changes

* Updated attachments to work with google cloud storage.
* Updated tests
* (Unrelated clean-up) removed Error handler service because we are using the fastapi version and this is unused code 🧽 

### Steps to Confirm

- [ ] All tests should pass
- [ ] With fidesplus pointed at this branch. Set up default storage config pointed at GCS prj-sandbox-55855-test-bucket. Instructions to do that can be found [here](https://github.com/ethyca/fides/pull/6006)
- [ ]  Create a DSR
- [ ] Add an attachment to the DSR
Screenshot 2025-05-21 at 10 43 31 AM

It should show up in the test bucket
Screenshot 2025-05-21 at 10 44 46 AM

- [ ] Verify you can retrieve the attachment by id.
Screenshot 2025-05-21 at 10 50 36 AM

- [ ] Delete the attachment works as expected - Before running delete command if you have gone into the created folder to check the attachment make sure you are no longer in it, because it will not auto delete a folder that is being accessed. 

- [ ] Verify the attachment is no longer in gcs.
Screenshot 2025-05-21 at 10 54 39 AM

- Attachments should work the same way for manual webhook steps and there should be no change in functionality for s3 or local storage. 

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:

[LJ-530]: https://ethyca.atlassian.net/browse/LJ-530?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/fides/api/models/attachment.py
```diff
@@ -204,9 +244,27 @@ def delete_attachment_from_storage(self) -> None:
...
         if self.config.type == StorageType.local:
-            filename = get_local_filename(f"{self.id}/{self.file_name}")
-            os.remove(filename)
+            folder_path = os.path.dirname(
+                get_local_filename(f"{self.id}/{self.file_name}")
+            )
+            if os.path.exists(folder_path):
+                import shutil
+
+                shutil.rmtree(folder_path)
```
galvana, line 267: What does the `folder_path` evaluate to? I want to make sure we don't delete the entire contents of `fides_uploads`

src/fides/api/service/storage/s3.py
```diff
@@ -169,7 +170,24 @@ def generic_delete_from_s3(
 
     s3_client = maybe_get_s3_client(auth_method, storage_secrets)
     try:
-        s3_client.delete_object(Bucket=bucket_name, Key=file_key)
+        # If the file_key ends with a '/', it's a folder prefix
+        if file_key.endswith("/"):
+            # List all objects with the prefix
+            objects_to_delete = s3_client.list_objects_v2(
```
galvana, line 176: I don't know how likely this would be, but if there are more than 1000 files, then we would need to paginate to get the next set of files https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/client/list_objects_v2.html#S3.Client.list_objects_v2. We probably don't need to do this, but just thought I'd call it out

src/fides/api/models/attachment.py
```diff
@@ -183,16 +205,34 @@ def retrieve_attachment(
...
+            )
+            return blob.size, url
+
         if self.config.type == StorageType.local:
             filename = get_local_filename(f"{self.id}/{self.file_name}")
             with open(filename, "rb") as file:
-                size = len(file.read())
+                content = file.read()
+                size = len(content)
                 return size, filename
```
galvana, line 230: We could do something like this to avoid having to load the entire file into memory
```suggestion
                size = os.path.getsize(filename)
                return size, filename
```

src/fides/api/models/attachment.py
JadeCara: I just went and double checked because ... paranoia...

`get_local_filename()` is called with `f"{self.id}/{self.file_name}"`.

 In `src/fides/api/service/storage/util.py` -  `get_local_filename()` adds `LOCAL_FIDES_UPLOAD_DIRECTORY` to the start of the path, and `LOCAL_FIDES_UPLOAD_DIRECTORY == fides_uploads`.

`os.path.dirname()`  is called on this path, which returns the directory portion of the path.
So for an attachment with:
id = `att_123`
file_name = `example.pdf`
The folder_path would evaluate to: `fides_uploads/att_123`

So this should be safe.  It only deletes the specific folder for this attachment (named by its ID)
It's contained within the fides_uploads directory
Each attachment gets its own subdirectory named by its ID - this follows the pattern I am using for `gcs` and `s3`

src/fides/api/service/storage/s3.py
JadeCara: I updated to use the paginate option just in case 
If we want to optimize this further, we could batch the deletes using `delete_objects` with multiple keys per page, but that would make the code more complex. 

The current implementation prioritizes reliability and simplicity over performance, which is probably the right tradeoff given that:
This is a deletion operation that doesn't need to be super fast
The number of files per attachment is likely to be small
The simpler code is easier to maintain and less prone to bugs

src/fides/api/models/attachment.py
galvana: Perfect, thanks for checking!

src/fides/api/service/storage/s3.py
```diff
@@ -169,6 +170,17 @@ def generic_delete_from_s3(
 
     s3_client = maybe_get_s3_client(auth_method, storage_secrets)
     try:
+        # If the file_key ends with a '/', it's a folder prefix
+        if file_key.endswith("/"):
+            # List all objects with the prefix, handling pagination
+            paginator = s3_client.get_paginator("list_objects_v2")
```
galvana, line 176: Wow, so easy! 👍

galvana (approved):
Thanks for making the changes, looks good!

PR #6166, opened 2025-05-22
Table migration mini POC using Cursor

gilluminate:
Closes [ENG-437]

### Description Of Changes

Cursor prompts:

1. `@ConsentManagementTable.tsx` uses custom table components built with TanStack and Chakra. We want to recreate this exact table functionality, but ONLY using AntTable from Fides UI. You can ignore the filter for now. Add this new Ant Table version to the `@table-migration.tsx` page. Try to avoid using non-Ant components
2. Instead of a filter button that will open a custom modal, can we implement the same filter functionality using Ant Table's built in Column Filtering?
3. the filter looks good, but the api query doesn't seem to be configured right. The end result of selecting and applying the filters should look something like `...plus/system/consent-management/report?page=1&size=25&purposes=2&purposes=3` but it currently looks something like `...plus/system/consent-management/report?page=1&size=25&2,3,2,3`
4. let's debounce the search input at 300ms

### Code Changes

* 100% AI generated code

[ENG-437]: https://ethyca.atlassian.net/browse/ENG-437?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/clients/admin-ui/src/features/common/nav/nav-config.tsx
+++ b/clients/admin-ui/src/features/common/nav/nav-config.tsx
@@ -311,6 +311,11 @@ if (process.env.NEXT_PUBLIC_APP_ENV === "development") {
         path: routes.FORMS_POC_ROUTE,
         scopes: [],
       },
+      {
+        title: "Table Migration POC",
+        path: routes.TABLE_MIGRATION_POC_ROUTE,
+        scopes: [],
+      },
     ],
   });
 }
--- a/clients/admin-ui/src/features/common/nav/routes.ts
+++ b/clients/admin-ui/src/features/common/nav/routes.ts
@@ -73,4 +73,5 @@ export const OPENID_AUTHENTICATION_ROUTE = "/settings/openid-authentication";
 
 export const ANT_POC_ROUTE = "/poc/ant-components";
 export const FORMS_POC_ROUTE = "/poc/forms";
+export const TABLE_MIGRATION_POC_ROUTE = "/poc/table-migration";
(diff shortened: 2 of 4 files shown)


lucanovera (approved):
Thanks for running this experiment 💯 . I'm pleasantly surprised by the results. A few more improvements and I'd say it's ready to ship. 
Looking at the code, it's a very good first pass, there's some hardcoding and I'd replace the way urls params are formed to avoid string manipulation. 
I also noticed we're missing the button to expand the contracted values in the table and that the Legal basis filter shows limited options, but.... those are issues with the current table too.
Let's get this merged so we can have a working example of an ant table for reference. Maybe we can create a follow up ticket for adding the Collapse/Expand column functionality?

gilluminate:
> Thanks for running this experiment 💯 . I'm pleasantly surprised by the results. A few more improvements and I'd say it's ready to ship. Looking at the code, it's a very good first pass, there's some hardcoding and I'd replace the way urls params are formed to avoid string manipulation. I also noticed we're missing the button to expand the contracted values in the table and that the Legal basis filter shows limited options, but.... those are issues with the current table too. Let's get this merged so we can have a working example of an ant table for reference. Maybe we can create a follow up ticket for adding the Collapse/Expand column functionality?

Thanks! Yeah, I totally agree there's room for improvement but sufficient for the purpose of the POC. The original table is still in tact so we don't need to make improvements to this POC page per-se. Lot's to improve on during the real migration.

PR #6173, opened 2025-05-27
Refactor init for better clarity + TCF separation

gilluminate:
### Description Of Changes

There was some developer confusion around when and how existing consent gets set and how it fits in to the initialization process. This refactor is meant to clear up confusion by moving some of the logic to the main `init` thread. There's a different set of logic needed for regular fides vs. TCF vs. Headless, and this refactor helps make that clear without changing any of the existing functionality.

### Code Changes

* Rename `getInitialFides` to `getInitialFidesFromConsentCookie` for clarity
* Add lots of extra comments explaining why things are the way they are
* Move the checks for existing cookie, TCF string, and Migratable consent (3rd party cookies) up to the main init thread rather than burying it in `getInitialFidesFromConsent` where that context isn't obvious and can be confusing.

### Steps to Confirm

1. No change in functionality, so general smoke tests and passing CI tests are warranted.

### Pre-Merge Checklist

* [x] Issue requirements met
* [x] `CHANGELOG.md` updated
* Followup issues:
  * [x] Followup issues created
  * [ ] No followup issues
* Database migrations:
  * [x] No migrations
* Documentation:

### DIFF
--- a/clients/fides-js/src/fides.ts
+++ b/clients/fides-js/src/fides.ts
@@ -37,7 +37,7 @@ import {
 } from "./lib/init-utils";
 import {
   getInitialCookie,
-  getInitialFidesFromConsent,
+  getInitialFidesFromConsentCookie,
   getOverridesByType,
   initialize,
 } from "./lib/initialize";
@@ -157,7 +157,7 @@ async function init(this: FidesGlobal, providedConfig?: FidesConfig) {
      * In this case, the experience is less important because the user has already consented to something.
      * The earlier we can communicate consent to the vendor, the better.
      */
-    const initialFides = getInitialFidesFromConsent({
+    const initialFides = getInitialFidesFromConsentCookie({
       ...config,
       cookie: this.cookie,
       savedConsent: this.saved_consent,
--- a/clients/fides-js/src/lib/initialize.ts
+++ b/clients/fides-js/src/lib/initialize.ts
@@ -139,7 +139,7 @@ export const getInitialCookie = ({ consent, options }: FidesConfig) => {
 /**
  * If saved preferences are detected, immediately initialize from local cache
  */
-export const getInitialFidesFromConsent = ({
+export const getInitialFidesFromConsentCookie = ({
   cookie,
   savedConsent,
   experience,
--- a/clients/fides-js/src/fides-headless.ts
+++ b/clients/fides-js/src/fides-headless.ts
@@ -31,7 +31,7 @@ import {
 } from "./lib/init-utils";
 import {
   getInitialCookie,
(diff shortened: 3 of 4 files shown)


clients/fides-js/src/lib/initialize.ts
```diff
@@ -160,23 +155,10 @@ export const getInitialFides = ({
...
     }) => PrivacyExperience;
-  }): Partial<FidesGlobal> | null => {
-  const hasExistingCookie = !isNewFidesCookie(cookie);
-
-  // Register any configured consent migration providers
-  registerDefaultProviders(options);
-
-  // Check if there's consent available to migrate from any provider
-  const { consent: migratedConsent } = readConsentFromAnyProvider(options);
-  const hasMigratableConsent = !!migratedConsent;
```
NevilleS, line 171: Were we doing this part twice before (`registerDefaultProviders` and `readConsentFromAnyProvider`)? Seems like it?

clients/fides-js/src/lib/initialize.ts
```diff
@@ -144,7 +139,7 @@ export const getInitialCookie = ({ consent, options }: FidesConfig) => {
 /**
  * If saved preferences are detected, immediately initialize from local cache
  */
-export const getInitialFides = ({
+export const getInitialFidesFromConsent = ({
```
NevilleS, line 142: nit: FWIW, the name `getInitialFidesFromConsentCookie` might be more accurate, since this function now mostly just unpacks the `cookie` prop.

clients/fides-js/src/fides.ts
```diff
@@ -145,20 +148,30 @@ async function init(this: FidesGlobal, providedConfig?: FidesConfig) {
...
+  this.experience = config.experience; // pre-fetched experience if available
+  const hasExistingCookie = !isNewFidesCookie(this.cookie);
+  const hasMigratableConsent = !!migratedConsent;
+  if (hasExistingCookie || hasMigratableConsent) {
+    /*
+     * We have enough information to initialize the Fides object before we have a valid experience.
+     * In this case, the experience is less important because the user has already consented to something.
+     * The earlier we can communicate consent to the vendor, the better.
+     */
+    const initialFides = getInitialFidesFromConsent({
```
NevilleS, line 160: OK, yeah, this is clearer; basically you've hoisted the internal `if` out of the `getInitialFides` code and placed it here.

clients/fides-js/src/lib/initialize.ts
gilluminate: yep! ...and only so we could form the `if` statement.

PR #6177, opened 2025-05-28
ENG-600 Respondent Users

JadeCara:
Closes [ENG-600](https://ethyca.atlassian.net/browse/ENG-600?atlOrigin=eyJpIjoiZjgwY2I1YjE0NzIxNDY2YWEyOWY1ZDZkY2JkYWM2MzIiLCJwIjoiaiJ9)

### Description Of Changes

In order to have external users work on manual DSRs we need to create new roles with very limited scopes. This PR introduces two new roles `respondent` and `external_respondent`. This will give us varying degrees of flexibility in the future when handling users who have been assigned very specific tasks who are either internal or external to a company. It allows us to continue to use the `fides_users` table for all users which does come with some risks. To help mitigate this new functionality was also included that does not allow a respondents role, password or email address to be changed. 

We are also introducing an email verification system - similar to privacy center that will allow the user to click a link, and be sent a new verification code which corresponds to the link's access token. The links are associated with new DSR's being created that the user is responsible for processing but a new code will be required each time the link is clicked. There is a 1 hour time out for these 16 character access codes.

Heavily referenced [this](https://ethyca.atlassian.net/wiki/spaces/PM/pages/3697049643/Manual+DSR+Response+UI+-+Threat+Model) threat model while working on these changes 
There is a follow up ticket to ensure these changes are incorporated into the endpoints etc [here](https://ethyca.atlassian.net/browse/ENG-668) 

### Code Changes

* Added additional functionality to the fides_user model which is role based.
* Added `create_respondent` to fides_user model which will verify data based on respondent requirements. 
* Added additional functionality to the permissions model which is role based.
* Added a new class for fides_user_respondent_email_verification. 
* Added a new alembic migration for the above table.
* Added tests for all above and moved to a fides_user dir to make it easier to share fixtures between these tests. 

### Steps to Confirm

1.  These are all back end changes for the moment. All tests should pass and there should be no change to any user related operations at this time.

### Pre-Merge Checklist

* [x] Issue requirements met
* [x] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:

[ENG-600]: https://ethyca.atlassian.net/browse/ENG-600?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/ops/models/test_fidesopsuser.py
galvana, line 1: Nice cleanup of these test files ✨

src/fides/api/models/fides_user_respondent_email_verification.py
```diff
@@ -0,0 +1,141 @@
...
+    This handles two types of verification:
+    1. Access links - long-lived (7 days) for initial access
+    2. Verification codes - short-lived (1 hour) for actual verification
+
+    When an email is sent to an external respondent, a new verification is created with a new access token is created.
+    When a respondent clicks the link in the email, the access token is verified and a verification code is generated.
+    The verification code is sent to the respondent's email address and the respondent is prompted to enter the code.
+    If the code is correct and has not expired, the respondent is considered verified.
+    If the code is incorrect, the number of attempts is incremented and the verification is saved.
+    If the user attemptes to use a link with an expired access token the user is not verified.
```
galvana:
```suggestion
    If the user attempts to use a link with an expired access token the user is not verified.
```

src/fides/api/alembic/migrations/versions/440a5b9a3493_add_fides_user_respondent_email_.py
```diff
@@ -0,0 +1,92 @@
...
+Revision ID: 440a5b9a3493
+Revises: d0cbfec0b2dd
+Create Date: 2025-05-29 15:41:59.521940
+
+"""
+
+import sqlalchemy as sa
+from alembic import op
+from citext import CIText
+from sqlalchemy.dialects import postgresql
```
galvana: Unused import
```suggestion
```

src/fides/api/models/fides_user_respondent_email_verification.py
```diff
@@ -0,0 +1,141 @@
...
+
+    access_token = Column(
+        String, nullable=False, unique=True, index=True
+    )  # Token for the access link
+    access_token_expires_at = Column(DateTime(timezone=True), nullable=False)
+    verification_code = Column(
+        String, nullable=True
+    )  # Optional until access link is used
+    verification_code_expires_at = Column(DateTime(timezone=True), nullable=True)
+    attempts = Column(Integer, nullable=False, server_default="0")
```
galvana: Should we only allow the user a max number of attempts?

src/fides/api/models/fides_user.py
```diff
@@ -23,12 +23,17 @@
 from fides.api.models.audit_log import AuditLog
 
 # Intentionally importing SystemManager here to build the FidesUser.systems relationship
-from fides.api.models.system_manager import SystemManager  # type: ignore[unused-import]
+# from fides.api.models.system_manager import SystemManager  # type: ignore[unused-import]
```
JadeCara:
```suggestion
```

.fides/fides.toml
```diff
@@ -42,6 +42,7 @@ oauth_root_client_id = "fidesadmin"
 oauth_root_client_secret = "fidesadminsecret"
 drp_jwt_secret = "secret"
 env = "prod"
+respondent_verification_max_attempts = 5
```
JadeCara: @daveqnet or @RobertKeyser I am adding an environment variable here for having a max login attempt for respondents and wanted to clear it with the security team!

galvana (approved):
Thanks for making the changes! Looks good

.fides/fides.toml
daveqnet: Hi @JadeCara, from a high-level perspective will this work in roughly the same way as `FIDES__SECURITY__IDENTITY_VERIFICATION_ATTEMPT_LIMIT` or is it quite different?

JadeCara: I guess that depends on if we want to have them be configurable independently. I think it should be fine to go with the existing. Good call out! I will adjust to use that :)

src/fides/api/models/fides_user_respondent_email_verification.py
```diff
@@ -0,0 +1,113 @@
...
+        nullable=False,
+        index=True,
+    )
+
+    access_token = Column(
+        String, nullable=False, unique=True, index=True
+    )  # Token for the access link
+    access_token_expires_at = Column(DateTime(timezone=True), nullable=False)
+    identity_verified_at = Column(DateTime(timezone=True), nullable=True)
+    created_at = Column(DateTime(timezone=True), nullable=False, server_default="now()")
```
galvana: This is inherited from `Base` so we don't need to redefine it here

.fides/db_dataset.yml
```diff
@@ -585,6 +585,38 @@ dataset:
...
+      - system.operations
+    - name: access_token_expires_at
+      data_categories:
+      - system.operations
+    - name: verification_code
+      data_categories:
+      - system.operations
+    - name: verification_code_expires_at
+      data_categories:
+      - system.operations
```
galvana: The fields don't exist on `FidesUserRespondentEmailVerification` anymore

.fides/db_dataset.yml
```diff
@@ -585,6 +585,38 @@ dataset:
...
       data_categories:
       - system.operations
+  - name: fides_user_respondent_email_verification
+    fields:
+    - name: id
+      data_categories:
+      - system.operations
+    - name: username
+      data_categories:
+      - system.operations
```
galvana: Username is also no longer on the model

JadeCara:
```suggestion
```

.fides/db_dataset.yml
```diff
@@ -585,6 +585,35 @@ dataset:
...
+    - name: verification_code
+      data_categories:
+      - system.operations
+    - name: verification_code_expires_at
+      data_categories:
+      - system.operations
+    - name: attempts
+      data_categories:
+      - system.operations
```
JadeCara:
```suggestion
    - name: identity_verified_at
      data_categories:
      - system.operations
```

src/fides/api/models/fides_user_respondent_email_verification.py
JadeCara:
```suggestion
```

PR #6178, opened 2025-05-28
ENG-132 - Task management framework

vcruces:
Closes [<issue>]

### Description Of Changes

### Code Changes

### Steps to Confirm

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:


src/fides/api/models/worker_task.py
```diff
@@ -0,0 +1,90 @@
...
+
+class TaskExecutionLog:
+    """
+    Stores the individual execution logs associated with a WorkerTask.
+    """
+
+    status = Column(
+        EnumColumn(
+            ExecutionLogStatus,
+            native_enum=False,
```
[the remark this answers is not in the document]
galvana: This only happens on the `ctl-not-external` tests because that's where the test the DB reset (`reset_db`)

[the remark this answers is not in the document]
vcruces: But that line was already in `RequestTask` class, I just moved it. Can you think of a reason why it might already be there?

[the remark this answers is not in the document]
vcruces: Searching the history, [I found when the change was made](https://github.com/ethyca/fides/commit/9cae3c22ac999a3b2f9250db8db6918850d302ab#diff-ef7d3e10be42d5a81721075db1647f7a59ad17ab20b273bfcda1edbcbc8ba391L1656). Perhaps the problem is that the enum was never deleted from the database

src/fides/api/models/detection_discovery.py
```diff
@@ -649,3 +645,138 @@ def fetch_staged_resources_by_type_query(
...
+    """
+    Types of tasks that can be executed by a worker.
+    """
+
+    DETECTION = "detection"
+    CLASSIFICATION = "classification"
+    PROMOTION = "promotion"
+
+
+class MonitorTask(WorkerTask, Base):
```
adamsachs: we probably want some more indexes on this table, right? specifically `celery_id`, `monitor_config_id`, since we're going to support APIs that filter on those

src/fides/api/models/detection_discovery.py
adamsachs, line 1: i know up to this point we've thrown a lot into this particular module, but i think that's been more accidental than intentional - i'd suggest we move all the additions here to their own `monitor_task.py` module. i think it also may be time to have a `detection_discovery` package (folder) where both that and this module live

src/fides/api/models/worker_task.py
adamsachs, line 1: nice - this looks like a solid abstraction! i'd like to know what @galvana thinks when he has a chance

src/fides/api/alembic/migrations/versions/4d9b696a3c3b_add_monitor_tasks.py
```diff
@@ -0,0 +1,144 @@
...
+revision = "4d9b696a3c3b"
+down_revision = "c586a56c25e7"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.create_table(
+        "monitortask",
+        sa.Column("id", sa.String(length=255), nullable=False),
```
adamsachs: probably want an index on this `id` PK column too since i think the API uses it as the "index" for the `get_monitor_task_detail` endpoint, right?

src/fides/api/models/privacy_request/request_task.py
```diff
@@ -68,7 +67,7 @@ def create_empty_traversal(cls, connection_key: str) -> TraversalDetails:
         )
 
 
-class RequestTask(Base):
+class RequestTask(WorkerTask, Base):
```
adamsachs, line 71: could be nice to mark a TODO here that we'd like to refactor the model here to store all types of task in a common table that links to tables with specific task attributes?

src/fides/api/models/worker_task.py
```diff
@@ -0,0 +1,90 @@
...
+            self.validate_action_type(action_type)
+        super().__init__(*args, **kwargs)
+
+    @classmethod
+    def validate_action_type(cls, action_type: str) -> None:
+        if action_type not in cls.allowed_action_types():
+            raise ValueError(f"Invalid action_type '{action_type}' for {cls.__name__}")
+
+    @classmethod
+    def allowed_action_types(cls) -> List[str]:
```
adamsachs, line 58: this could use a docstring to explain what's needed from the subclasses!

src/fides/api/models/detection_discovery.py
```diff
@@ -649,3 +645,138 @@ def fetch_staged_resources_by_type_query(
...
+    )
+
+    monitor_task = relationship("MonitorTask", back_populates="execution_logs")
+
+
+def create_monitor_task_with_execution_log(
+    db: Session, monitor_task_data: dict
+) -> MonitorTask:
+    """
+    Creates a monitor task with an execution log.
```
adamsachs: nit: would be good to clarify that `pending` is the starting/default state here 👍

src/fides/api/models/detection_discovery.py
vcruces: You're right, I'll add index=True for `monitor_config_id`. Celery_id is already index=True because it's a unique value

src/fides/api/alembic/migrations/versions/4d9b696a3c3b_add_monitor_tasks.py
vcruces: I think since it's defined as the primary key, the database automatically creates a unique index on it

src/fides/api/models/detection_discovery.py
vcruces: I really like the idea of ​​moving it to another file. And would you leave the same name for the file we already have? Or what would you name it?

src/fides/api/alembic/migrations/versions/4d9b696a3c3b_add_monitor_tasks.py
adamsachs: ah you're right about that, whoops!

src/fides/api/models/detection_discovery.py
adamsachs: up to you, i wouldn't be against renaming it something like `discovery_monitor_core.py` or something else that's better than that (literally the first thing i thought about), but also fine to leave as-is for now/this PR. i leave it up to you! just getting these in another file to start pulling things out into some different files here is a good starting point 👍

PR #6182, opened 2025-05-29
Generic partitioning DSL

galvana:
Closes [<issue>]

### Description Of Changes

### Code Changes

### Steps to Confirm

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:

### DIFF
--- a/src/fides/api/graph/config.py
+++ b/src/fides/api/graph/config.py
@@ -95,6 +95,7 @@ from fides.api.graph.data_type import (
     DataTypeConverter,
     get_data_type_converter,
 )
+from fides.api.schemas.partitioning import TimeBasedPartitioning
 from fides.api.util.collection_util import merge_dicts
 from fides.api.util.querytoken import QueryToken
 
@@ -686,10 +687,23 @@ class Collection(BaseModel):
         if not partitioning:
             return partitioning
 
-        # NOTE: when we deprecate `where_clause` partitioning in favor of a more proper partitioning DSL,
-        # we should be sure to still support the existing `where_clause` partition definition on
-        # any in-progress DSRs so that they can run through to completion.
-        if where_clauses := partitioning.get("where_clauses"):
+        # Keeping the where_clauses support for now even though we have a DSL to allow time for migration
+        where_clauses = partitioning.get("where_clauses")
+        time_based_partitioning = all(
+            field in partitioning for field in ["field", "start", "end", "interval"]
+        )
+
+        if where_clauses and time_based_partitioning:
+            raise ValueError(
+                "Cannot specify both `where_clauses` and time based partitioning (`field`, `start`, `end`, `interval`)"
+            )
+
+        if not where_clauses and not time_based_partitioning:
+            raise ValueError(
+                "Must specify either `where_clauses` or time based partitioning (`field`, `start`, `end`, `interval`)"
+            )
+
+        if where_clauses:
             if not isinstance(where_clauses, List) or not all(
                 isinstance(where_clause, str) for where_clause in where_clauses
             ):
@@ -715,10 +729,15 @@ class Collection(BaseModel):
                             )
                 else:
                     raise ValueError("Unsupported partition clause format")
-            return partitioning
-        raise ValueError(
-            "`where_clauses` must be specified in `partitioning` specification!"
-        )
+        if time_based_partitioning:
+            # Try to parse time-based partitioning
+            TimeBasedPartitioning(
+                field=partitioning["field"],
+                start=partitioning["start"],
+                end=partitioning["end"],
+                interval=partitioning["interval"],
+            )
+        return partitioning
 
 
 class GraphDataset(BaseModel):
--- a/src/fides/api/service/connectors/query_configs/bigquery_query_config.py
+++ b/src/fides/api/service/connectors/query_configs/bigquery_query_config.py
@@ -15,6 +15,7 @@ from fides.api.models.privacy_request import PrivacyRequest
 from fides.api.schemas.namespace_meta.bigquery_namespace_meta import (
     BigQueryNamespaceMeta,
 )
+from fides.api.schemas.partitioning import BigQueryTimeBasedPartitioning
 from fides.api.service.connectors.query_configs.query_config import (
     QueryStringWithoutTuplesOverrideQueryConfig,
 )
@@ -37,7 +38,7 @@ class BigQueryQueryConfig(QueryStringWithoutTuplesOverrideQueryConfig):
 
     @property
     def partitioning(self) -> Optional[Dict]:
-        # Overriden from base implementation to allow for _only_ BQ partitioning, for now
+        # Overridden from base implementation to allow for _only_ BQ partitioning, for now
         return self.node.collection.partitioning
 
     def get_partition_clauses(
@@ -65,7 +66,16 @@ class BigQueryQueryConfig(QueryStringWithoutTuplesOverrideQueryConfig):
         if where_clauses := partition_spec.get("where_clauses"):
             return where_clauses
 
-        # TODO: implement more advanced partitioning support!
+        if all(
+            field in partition_spec for field in ["field", "start", "end", "interval"]
+        ):
+            bigquery_time_based_partitioning = BigQueryTimeBasedPartitioning(
+                field=partition_spec["field"],
+                start=partition_spec["start"],
+                end=partition_spec["end"],
+                interval=partition_spec["interval"],
+            )
+            return bigquery_time_based_partitioning.generate_where_clauses()
 
         raise ValueError(
             "`where_clauses` must be specified in partitioning specification!"
--- a/tests/fixtures/bigquery_fixtures.py
+++ b/tests/fixtures/bigquery_fixtures.py
@@ -6,6 +6,7 @@ from typing import Dict, Generator, List
 from uuid import uuid4
 
 import pytest
+from loguru import logger
 from sqlalchemy.orm import Session
 
 from fides.api.models.connectionconfig import (
@@ -236,10 +237,10 @@ def bigquery_enterprise_test_dataset_config_with_partitioning_meta(
     )
     stackoverflow_posts_partitioned_collection["fides_meta"] = {
         "partitioning": {
-            "where_clauses": [
-                "`creation_date` > TIMESTAMP_SUB(CURRENT_TIMESTAMP(), INTERVAL 1000 DAY) AND `creation_date` <= CURRENT_TIMESTAMP()",
-                "`creation_date` > TIMESTAMP_SUB(CURRENT_TIMESTAMP(), INTERVAL 2000 DAY) AND `creation_date` <= TIMESTAMP_SUB(CURRENT_TIMESTAMP(), INTERVAL 1000 DAY)",
-            ]
+            "field": "creation_date",
+            "start": "NOW() - 2000 DAYS",
+            "end": "NOW()",
+            "interval": "1000 DAYS",
         }
     }
     bigquery_enterprise_dataset["collections"].append(
@@ -320,10 +321,10 @@ def bigquery_example_test_dataset_config_with_namespace_and_partitioning_meta(
     bigquery_dataset["collections"].remove(customer_collection)
     customer_collection["fides_meta"] = {
         "partitioning": {
-            "where_clauses": [
-                "`created` > TIMESTAMP_SUB(CURRENT_TIMESTAMP(), INTERVAL 1000 DAY) AND `created` <= CURRENT_TIMESTAMP()",
-                "`created` > TIMESTAMP_SUB(CURRENT_TIMESTAMP(), INTERVAL 2000 DAY) AND `created` <= TIMESTAMP_SUB(CURRENT_TIMESTAMP(), INTERVAL 1000 DAY)",
-            ]
+            "field": "created",
+            "start": "NOW() - 1000 DAYS",
+            "end": "NOW()",
+            "interval": "500 DAYS",
         }
     }
     bigquery_dataset["collections"].append(customer_collection)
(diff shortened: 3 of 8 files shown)


src/fides/api/graph/config.py
```diff
@@ -683,42 +682,77 @@ def validate_partitioning(
...
-            for partition_clause in where_clauses:
-                if matching := match(
-                    BIGQUERY_PARTITION_CLAUSE_PATTERN, partition_clause
+        # Normalize to a list so the existing validation logic can be reused for
+        # both single-dict and multi-dict inputs.
+        specs: List[Dict[str, Any]]
+        if isinstance(partitioning, dict):
+            specs = [partitioning]
+        elif isinstance(partitioning, list):
+            specs = partitioning
```
erosselli: so here `specs` is either a list or a dict? We could shorten this to

```py
specs: List[Dict[str, Any]] = partitioning if isinstance(partitioning, list) else [partitioning]
```

since the linter was complaining about too many branches, this might help?  Also right now if you pass in something that's not a list or a dict, I think the code will break in ` for spec in specs:` . With this change it will actually make it up to the `raise ValueError("Each partitioning specification must be a dict")` line, which is a much nicer error message

src/fides/api/graph/config.py
```diff
@@ -459,7 +460,7 @@ class Collection(BaseModel):
     grouped_inputs: Set[str] = set()
     data_categories: Set[FidesKey] = set()
     masking_strategy_override: Optional[MaskingStrategyOverride] = None
-    partitioning: Optional[Dict] = None
+    partitioning: Optional[Union[Dict, List[Dict]]] = None
```
erosselli: I don't understand why we wouldn't type this like ` partitioning: Optional[Union[Dict, List[TimeBasedPartitioning]]]` ?  it feels like it might make some of the validations below simpler too

src/fides/api/graph/config.py
```diff
@@ -683,42 +682,77 @@ def validate_partitioning(
...
+
+            if not where_clauses and not time_based_partitioning:
+                raise ValueError(
+                    "Must specify either `where_clauses` or time based partitioning (`field`, `start`, `end`, `interval`)"
+                )
+
+            if where_clauses:
+                if not isinstance(where_clauses, List) or not all(
+                    isinstance(where_clause, str) for where_clause in where_clauses
                 ):
```
erosselli: if we type `partitioning` more specifically so the `dict` part of the union is actually `dict[str, list[str]]`  (or even `dict['where_clauses', list[str]]` Pydantic would do this validation for us right?

src/fides/api/schemas/partitioning/time_based_partitioning.py
```diff
@@ -0,0 +1,829 @@
...
+        return v
+
+    @model_validator(mode="after")
+    def validate_bounds_and_interval(self) -> "TimeBasedPartitioning":
+        """Ensure an interval is supplied when both start and end are provided."""
+        if self.start is not None and self.end is not None and self.interval is None:
+            raise ValueError(
+                "An interval must be provided when both start and end are specified."
+            )
+        return self
```
erosselli: should we also check that either a start or an end were provided?

src/fides/api/schemas/partitioning/time_based_partitioning.py
```diff
@@ -0,0 +1,829 @@
...
+    @classmethod
+    def parse(cls, raw: str) -> "TimeUnit":
+        """Parse raw (possibly lowercase or plural) into a TimeUnit."""
+
+        cleaned = raw.strip().upper().rstrip("S")
+        try:
+            return cls[cleaned]
+        except KeyError as exc:  # pragma: no cover – validated upstream
+            raise ValueError(f"Unsupported time unit: {raw}") from exc
+
```
erosselli: if it's validated upstream, tests should cover it still right? or do you mean in fidesplus? also it should be pretty easy to write a unit test for this :)

src/fides/api/schemas/partitioning/time_based_partitioning.py
```diff
@@ -0,0 +1,829 @@
...
+        if self.interval is None:
+            raise ValueError("Interval must be specified before it can be parsed.")
+
+        # Normalize once to uppercase, then split
+        normalized = str(self.interval).strip().upper()
+        parts = normalized.split()
+
+        if len(parts) != 2 or not parts[0].isdigit():
+            raise ValueError(f"Invalid interval format: {self.interval}")
+
```
erosselli: can we add in-code comments with an example of what each step looks like? I find those make these kinds of string math or date math more readable, e..g

```py
        # Normalize once to uppercase, then split
        # e.g if interval is "500 days", normalized will be "500 days" and parts will be ["500", "DAYS"]
        normalized = str(self.interval).strip().upper()
        parts = normalized.split()
 ```

Comment (erosselli) [reply src/fides/api/schemas/partitioning/time_based_partitioning.py]: however I'm wondering do we even need this validation? we already validate against a regex in `validate_interval` so it should be ok?  

I think we can reuse the same regex to split the interval by changing it to group the digit and the time unit, e.g `"^(\d+)\s+(DAY|DAYS|WEEK|WEEKS|MONTH|MONTHS|YEAR|YEARS)$` and using groups. Ideally define the regex as a constant so we don't duplicate it

File: src/fides/api/schemas/partitioning/time_based_partitioning.py [hunk 7]
```diff
@@ -0,0 +1,829 @@
...
+    def validate_time_expression(cls, value: Optional[str]) -> Optional[str]:
+        """Validate time expressions."""
+        if value is None:
+            return value
+        v = value.strip().upper()
+        patterns = [
+            r"^NOW\(\)$",  # NOW()
+            r"^TODAY\(\)$",  # TODAY()
+            r"^NOW\(\)\s*[+-]\s*\d+\s+(DAY|DAYS|WEEK|WEEKS|MONTH|MONTHS|YEAR|YEARS)$",  # NOW() +/- N unit
+            r"^TODAY\(\)\s*[+-]\s*\d+\s+(DAY|DAYS|WEEK|WEEKS|MONTH|MONTHS|YEAR|YEARS)$",  # TODAY() +/- N unit
```
erosselli: we use these regexes here and also

```
  arithmetic_match = re.match(
            r"^(NOW|TODAY)\(\)\s*([+-])\s*(\d+)\s+(DAY|DAYS|WEEK|WEEKS|MONTH|MONTHS|YEAR|YEARS)$",
            expr,
        )
```

down below, can we maybe share them so they're not duplicated?

src/fides/api/schemas/partitioning/time_based_partitioning.py
```diff
@@ -0,0 +1,829 @@
...
+
+        Supported patterns
+        ------------------
+        1. `NOW() - N DAY|DAYS|WEEK|WEEKS to NOW()`
+        2. `TODAY() - N DAY|DAYS|WEEK|WEEKS to TODAY()`
+        3. Both `start` and `end` are literal `YYYY-MM-DD` strings.
+        """
+        start_str = str(self.start)
+        end_str = str(self.end)
+
```
erosselli: start and end are already strings right? why is this cast needed?

src/fides/api/schemas/partitioning/time_based_partitioning.py
```diff
@@ -0,0 +1,829 @@
...
+            # Determine the relative offset (timedelta) that represents the user
+            # supplied `end` bound so we do not generate slices beyond it.  If the
+            # end expression is simply NOW()/TODAY(), the offset is zero.  If it
+            # is an expression like "NOW() - 500 DAYS" we extract the numeric
+            # offset so we can stop the fixed-length iteration once we reach it.
+
+            end_offset_bound: timedelta = timedelta(0)
+
+            dynamic_end_match = re.match(
+                r"^(NOW\(\)|TODAY\(\))\s*-\s*(\d+)\s+(DAY|DAYS|WEEK|WEEKS)$",
```
erosselli: is this one the same as `offset_pattern` above?

src/fides/api/schemas/partitioning/time_based_partitioning.py
```diff
@@ -0,0 +1,829 @@
...
+                end_str or "",
+            )
+
+            if dynamic_end_match is not None:
+                numeric_val = int(dynamic_end_match.group(2))
+                unit_raw = dynamic_end_match.group(3)
+                end_offset_bound = (
+                    timedelta(weeks=numeric_val)
+                    if unit_raw in ["WEEK", "WEEKS"]
```
erosselli:
```suggestion
                    if TimeUnit .parse(unit_raw) == TimeUnit.WEEK
```

src/fides/api/schemas/partitioning/time_based_partitioning.py
```diff
@@ -0,0 +1,829 @@
...
+            # Dynamic NOW()/TODAY() -> NOW()/TODAY() path
+            if (
+                re.match(r"^NOW\(\)\s*-\s*\d+\s+(MONTH|MONTHS|YEAR|YEARS)$", start_str)
+                and end_str == "NOW()"
+            ) or (
+                re.match(
+                    r"^TODAY\(\)\s*-\s*\d+\s+(MONTH|MONTHS|YEAR|YEARS)$", start_str
+                )
+                and end_str == "TODAY()"
+            ):
```
erosselli: same comment here about reusing the regexes

src/fides/api/service/connectors/query_configs/bigquery_query_config.py
```diff
@@ -36,40 +43,93 @@ class BigQueryQueryConfig(QueryStringWithoutTuplesOverrideQueryConfig):
...
             return []
 
-        if where_clauses := partition_spec.get("where_clauses"):
-            return where_clauses
+        # Normalize to list so the rest of the logic can treat everything the same way.
+        specs: List[Dict] = (
+            [partition_spec]
+            if isinstance(partition_spec, dict)
+            else list(partition_spec)
+        )
```
erosselli: we have similar definition in the schema definition, can we not use that instead ?

src/fides/api/service/connectors/query_configs/bigquery_query_config.py
```diff
@@ -36,40 +43,93 @@ class BigQueryQueryConfig(QueryStringWithoutTuplesOverrideQueryConfig):
...
+                    f"Each partitioning specification must be a dict, got {type(spec)}"
+                )
+            if not TIME_BASED_REQUIRED_KEYS.issubset(spec):
+                raise ValueError(
+                    "Each time-based partitioning spec must contain field, start, end, and interval."
+                )
+            if "where_clauses" in spec:
+                raise ValueError(
+                    "`where_clauses` cannot be combined with time-based partitioning."
+                )
```
erosselli: same question for these checks, they feel similar to the ones in the schema?

tests/fixtures/bigquery_fixtures.py
```diff
@@ -236,10 +237,10 @@ def bigquery_enterprise_test_dataset_config_with_partitioning_meta(
...
     stackoverflow_posts_partitioned_collection["fides_meta"] = {
         "partitioning": {
-            "where_clauses": [
-                "`creation_date` > TIMESTAMP_SUB(CURRENT_TIMESTAMP(), INTERVAL 1000 DAY) AND `creation_date` <= CURRENT_TIMESTAMP()",
-                "`creation_date` > TIMESTAMP_SUB(CURRENT_TIMESTAMP(), INTERVAL 2000 DAY) AND `creation_date` <= TIMESTAMP_SUB(CURRENT_TIMESTAMP(), INTERVAL 1000 DAY)",
-            ]
+            "field": "creation_date",
+            "start": "NOW() - 2000 DAYS",
+            "end": "NOW()",
+            "interval": "1000 DAYS",
```
erosselli: I think we should keep the `where_clauses` tests around for now right? that way we make sure we haven't broken the old functionality.... I think we need to keep the old tests around and check the queries are exactly the same, while still adding new tests for the new partitioning scheme

src/fides/api/graph/config.py
galvana: Agreed! It's a lot cleaner now

src/fides/api/graph/config.py
galvana: It's a lot cleaner now, don't need to do anything here

src/fides/api/graph/config.py
galvana: Same as above

src/fides/api/schemas/partitioning/time_based_partitioning.py
galvana: Pulled this all out to constants

src/fides/api/schemas/partitioning/time_based_partitioning.py
galvana:
```
# At least one bound must be supplied.
if self.start is None and self.end is None:
    raise ValueError("At least one of 'start' or 'end' must be provided.")
```

src/fides/api/schemas/partitioning/time_based_partitioning.py
galvana: I updated this to
```
start_str = self.start or ""
end_str = self.end or ""
```

src/fides/api/schemas/partitioning/time_based_partitioning.py
galvana: Yes, I made both use a constant now

src/fides/api/schemas/partitioning/time_based_partitioning.py
galvana: Good catch!

src/fides/api/service/connectors/query_configs/bigquery_query_config.py
galvana: Yes! Simplifying logic

src/fides/api/service/connectors/query_configs/bigquery_query_config.py
galvana: Simplified this as well

src/fides/api/schemas/partitioning/time_based_partitioning.py
galvana: Good point, I just re-used the regex here

src/fides/api/schemas/partitioning/time_based_partitioning.py
galvana: Added tests!

src/fides/api/graph/config.py
```diff
@@ -654,15 +658,13 @@ def serialize_erase_after(self, erase_after: Set[CollectionAddress]) -> Set[str]
 
     @field_validator("partitioning")
     @classmethod
-    def validate_partitioning(
-        cls, partitioning: Optional[Dict[str, Any]]
-    ) -> Optional[Dict[str, Any]]:
+    def validate_partitioning(  # pylint: disable=too-many-branches,too-many-nested-blocks
```
erosselli:
```suggestion
    def validate_partitioning( 
```

I think we probably don't need the pylint diable comment any more :)

src/fides/api/schemas/partitioning/time_based_partitioning.py
```diff
@@ -0,0 +1,839 @@
...
+from typing import List, Optional, Set, Tuple
+
+from pydantic import Field, field_validator, model_validator
+from sqlalchemy import and_, column, func, text
+from sqlalchemy.sql import ColumnElement
+
+from fides.api.schemas.base_class import FidesSchema
+
+# ------------------------------------------------------------------
+# Regular expression patterns reused across validation helpers.
```
erosselli, line 15: ❤️

src/fides/api/schemas/partitioning/time_based_partitioning.py
```diff
@@ -0,0 +1,839 @@
...
+    # ---------------------------------------------------------------------
+    # Validators
+    # ---------------------------------------------------------------------
+
+    @field_validator("start", "end")
+    @classmethod
+    def validate_time_expression(cls, value: Optional[str]) -> Optional[str]:
+        """Validate time expressions."""
+        if value is None:
+            return value
```
erosselli, line 159: is `None` a valid time expression?

src/fides/api/schemas/partitioning/time_based_partitioning.py
```diff
@@ -0,0 +1,839 @@
...
+        """
+
+        start_str = self.start or ""
+
+        now_match = re.match(
+            r"^NOW\(\)\s*-\s*(\d+)\s+(MONTH|MONTHS|YEAR|YEARS)$", start_str
+        )
+        today_match = re.match(
+            r"^TODAY\(\)\s*-\s*(\d+)\s+(MONTH|MONTHS|YEAR|YEARS)$", start_str
+        )
```
erosselli: just flagging that these ones are missing the update

PR #6183, opened 2025-05-29
ENG-581 Add manual integrations to the integrations page

lucanovera:
### Description Of Changes

Adds manual integrations to the Integrations page. Adjustments to the integrations page to support this new "unconnected" integration. 

### Code Changes

* Adds manual integration info box
* Introduces "features" for integration that controls which tabs are shown
* Introduces "alpha feature flags". Any feature flag that starts with "alpha" will be hidden from the about page but can be enabled from the /settings/about/alpha url.

### Steps to Confirm

1.  Login to admin-ui
2. Go to Integrations page, check that no manual tab exist and manual integrations cannot be added
3. Go to /settings/about-alpha and turn on the "Alpha new manual integration" flag
4. Go to Integrations page, now there should be an option to add a Manual integration.
5. Check the integration has a description and the form only asks for name, description and system
6. Create the integration
7. Go to the integration detail page, check that there are no references to "Test connection" and the only tab available is "Details"

Regression:
8. Check that a normal connected integration like BigQuery can be added and still shows the correct tabs in the integration detail page.

### Pre-Merge Checklist

* [x] Issue requirements met
* [ ] `CHANGELOG.md` updated
  * [x] Code changes under alpha flag, not meant for release yet
* Followup issues:
  * [x] Followup issues created
  * [ ] No followup issues
* Database migrations:
  * [x] No migrations
* Documentation:


clients/admin-ui/src/features/integrations/IntegrationFeatureEnum.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * Enum defining the features that can be enabled for different integration types.
+ * These features control which tabs and functionality are available in the integration detail view.
+ */
+export enum IntegrationFeatureEnum {
```
jpople, line 5: Like this approach a lot, nice idea.

clients/admin-ui/src/pages/settings/about/index.tsx
```diff
@@ -19,10 +19,15 @@ import { FlagControl } from "~/features/common/features/FlagControl";
...
 
-const About: NextPage = () => {
+const About: NextPage<{ showAlphaFeatures: boolean }> = ({
+  showAlphaFeatures = false,
+}) => {
   const features = useFeatures();
   const { flags, override, reset } = useFlags();
 
+  const alphaFlags = FLAG_NAMES.filter((flag) => flag.startsWith("alpha"));
+  const betaFlags = FLAG_NAMES.filter((flag) => !flag.startsWith("alpha"));
```
jpople, line 29: I like the separate-page idea you've gone for here but filtering only on flag name feels a bit sketchy to me.  How much complexity would it add to just keep an entirely separate list of alpha features?

clients/admin-ui/src/flags.json
```diff
@@ -53,5 +53,11 @@
     "development": true,
     "test": true,
     "production": false
+  },
+  "alphaNewManualIntegration": {
+    "description": "New DSR Workflow Manual Integration",
+    "development": true,
```
jpople, line 59: Given we're adding a new page so there's extra friction to be able to enable this, shouldn't it probably be false by default?

lucanovera: It is off for production, which means it's off for all builds (even our testing/rc deployments, etc). The flag is true in development (that would be just when we're npm run dev or turbo run dev), because I saw that we're doing that's the pattern we're using where we develop with all features on.

clients/admin-ui/src/pages/settings/about/index.tsx
lucanovera: Yeah I agree, filtering on names is... not a very strong approach. 
I tried like 2 different approaches where we I added additional properties in the flags.json for alpha features. The problem is when you use the "useFlags" hook all you get back is the name of the flag and a boolean indicating if it's on or not. I couldn't do much else without doing a partial refactor of the features redux slice, which seems risky to do being close to the end of the sprint. 
My thought is that this can be a first version that is cheap and safe to release, and then we can iterate to a better approach if we end up using this flags a lot.

jpople (approved):

PR #6188, opened 2025-06-02
Shared monitor config UI

jpople:
Closes ENG-520

### Description Of Changes

Adds UI for CRUD operations on shared monitor configs.

Screenshot 2025-06-02 at 13 44 38

### Steps to Confirm

1. Navigate to "Integrations"
2. Click "Configurations" in the top-right corner
3. Click "create new"; should be able to enter some values and data categories and configure a shared monitor config
4. While creating, should be able to upload a CSV file like this and have the form populated
```
[abc],user
[def],system
```
6. Should be able to view results in the table and edit/delete shared configs; "Upload CSV" should not be shown on editing
7. Create an integration and go to the "Data discovery" tab
8. Create a monitor or edit an existing one
9. Should see your shared configs as options in the "Shared monitor config" dropdown and be able to apply one to the monitor

### Pre-Merge Checklist

* [x] Issue requirements met
* [x] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:

### DIFF
--- a/clients/admin-ui/src/pages/integrations/[id].tsx
+++ b/clients/admin-ui/src/pages/integrations/[id].tsx
@@ -23,8 +23,8 @@ import DatahubDataSyncTab from "~/features/integrations/configure-scan/DatahubDa
 import ConfigureIntegrationModal from "~/features/integrations/ConfigureIntegrationModal";
 import ConnectionStatusNotice from "~/features/integrations/ConnectionStatusNotice";
 import IntegrationBox from "~/features/integrations/IntegrationBox";
-import SharedConfigModal from "~/features/integrations/SharedConfigModal";
 import { IntegrationFeatureEnum } from "~/features/integrations/IntegrationFeatureEnum";
+import SharedConfigModal from "~/features/integrations/SharedConfigModal";
 import useIntegrationOption from "~/features/integrations/useIntegrationOption";
 import { ConnectionType } from "~/types/api";
 


lucanovera (changes requested):
Nice work, the new modal looks great! Handling an array of N fields is always tricky in a form but it's nicely implemented here. The csv upload is very smooth too. Code changes look good. I did found a couple bugs and have some few UX suggestions:

Bugs:
- When you delete a shared monitor config, the modal switches to it as if you had click edit instead.
- Once you select a config from the Select for a monitor, and then go to edit that monitor again, you'll find the Select is empty instead of showing the config you previously saved.

UX suggestions:
- Move the Configurations button. I think this feature is for a very particular type of client and it applies to monitors only, but there's no context for monitors in the Integrations page. We also have integrations there that support monitors either.
I think the best place for it is in the Monitors table of the Data Discovery tab of an integration For added context the button could be named the same as the modal title "Shared monitor configurations".
- When you create/edit a config and hit save, I think the modal should close or go back to the previous modal with the lists as there's nothing else to do in the page and you already added a nice toast with the success message for feedback.
- In the "Shared monitor configurations" the Create new monitor should be at the header of the table instead of the footer. We usually have add buttons in the headers and this would also prevent having it go too far down in case you have many configs.
- (nitpick) You can use the locale={{ emptyText prop for the Table to make it show a custom text when there are no results, otherwise it shows "No data".

lucanovera (approved):
Just confirmed all the fixed and UI improvements, looking good 💯 

 I'll leave the approve here assuming the test code for empty results will be removed and the cypress test will be fixed.

PR #6199, opened 2025-06-04
Storing large access data externally

galvana:
Closes [ENG-684](https://ethyca.atlassian.net/browse/ENG-684)

### Description Of Changes

This PR introduces an automatic external storage fallback system to handle data that exceeds PostgreSQL's 1GB column limit. The solution centers around a new `EncryptedLargeDataDescriptor` that can be seamlessly applied to SQLAlchemy model columns. This descriptor maintains backward compatibility by storing small data directly in the database as before, while automatically detecting large datasets and transparently offloading them to encrypted external storage through configurable backends (local, S3, or GCS).

```
RequestTask.data_for_erasures: Data size (1,111,650,541 bytes) exceeds threshold (671,088,640 bytes), storing externally
```

**EncryptedLargeDataDescriptor**
- Generic Python descriptor that can be applied to any encrypted SQLAlchemy column
  - Currently applied to `RequestTask.access_results`, `RequestTask.data_for_erasures`, and `PrivacyRequest.filtered_final_upload`
- Automatically determines whether to store data in database or external storage based on size
- Provides transparent getter/setter interface that maintains existing API compatibility
- Handles metadata tracking for externally stored files

**AES GCM Encryption Utilities**
- Leverages existing SQLAlchemy-Utils AES GCM encryption for consistency
- Introduces new memory-efficient chunked processing using Python's cryptography library
- Processes large datasets in batches to avoid memory overflow
- Maintains cross-compatibility between both encryption implementations

**ExternalStorageService**
- Unified interface supporting multiple storage backends (local filesystem, S3, GCS)
- Handles consistent file naming and organization across different storage types
- Manages complete lifecycle: store, retrieve, and cleanup operations
- Provides encrypted file storage with proper key management and metadata tracking

### Additional Code Changes

* Memory-efficient util to estimate the size on-disk for in-memory access results
* No longer cache large results in Redis for DSR 3.0

### Steps to Confirm

1. Check out the branch on https://github.com/ethyca/fidesplus/pull/2232
1. Enable DSR 3.0, put this in your `.env` `FIDES__EXECUTION__USE_DSR_3_0=true`
1. Make sure local storage is set to the default storage config with `html`
1. Create a new system with a `Large Data Test Connector`. The connection test will fail, that's ok

image

5. Submit an access request for `jane@example.com`
6. The encrypted values and final DSR package should be under `fidesplus/fides_uploads`

image

This could take a while, it took ~13 minutes from beginning to end

image

### Pre-Merge Checklist

* [x] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [x] No followup issues
* Database migrations:
  * [x] No migrations
* Documentation:

[ENG-684]: https://ethyca.atlassian.net/browse/ENG-684?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/fides/api/models/encrypted_large_data.py
```diff
@@ -0,0 +1,289 @@
...
+    ExternalDataStorageService,
+)
+from fides.api.util.collection_util import Row
+from fides.api.util.custom_json_encoder import CustomJSONEncoder
+
+# 1GB threshold for external storage
+LARGE_DATA_THRESHOLD_BYTES = 1024 * 1024 * 1024  # 1GB
+
+
+def calculate_data_size(data: List[Row]) -> int:
```
galvana: I ran a quick test that showed this approach is quick and memory-efficient while being within 2% of the actual size.

**Performance Comparison Results:**

**Small Dataset (1K records):**
Accuracy: 100.0% (exact match)
Speed: 1.0x (same speed since both use exact calculation)
Memory: 1.0x (same memory usage)

**Medium Dataset (5K records):**
Accuracy: 98.6% (24,700 bytes difference)
Speed: 11.3x faster
Memory: 9.2x less memory

**Large Dataset (20K records):**
Accuracy: 98.6% (99,747 bytes difference)
Speed: 28.3x faster
Memory: 13.7x less memory

src/fides/api/models/encrypted_large_data.py
```diff
@@ -0,0 +1,328 @@
...
+        """
+        instance_id = getattr(instance, "id", None)
+        if not instance_id:
+            raise ValueError(f"Instance {instance} must have an 'id' attribute")
+
+        timestamp = datetime.utcnow().strftime("%Y%m%d-%H%M%S-%f")
+
+        return f"{self.model_class}/{instance_id}/{self.field_name}/{timestamp}.txt"
+
+    def __get__(self, instance: Any, owner: Type) -> Any:
```
JadeCara: I don't see `owner` being used anywhere in this function - can probably be removed from the signature?

src/fides/api/models/encrypted_large_data.py
JadeCara, line 1: I am not sure if this belongs in models top level? It kind of blends in with the regular models.

src/fides/api/models/privacy_request/privacy_request.py
```diff
@@ -1257,6 +1265,11 @@ def get_consent_results(self) -> Dict[str, int]:
         # DSR 2.0 does not cache the results so nothing to do here
         return {}
 
+    def cleanup_external_storage(self) -> None:
+        """Clean up all external storage files for this privacy request"""
+        # Access the descriptor from the class to call cleanup
+        PrivacyRequest.filtered_final_upload.cleanup(self)
+
```
JadeCara, line 1272: 😎

src/fides/api/models/encrypted_large_data.py
```diff
@@ -0,0 +1,328 @@
...
+    size = calculate_data_size(data)
+    is_large = size > threshold
+
+    if is_large:
+        logger.info(
+            f"Data size ({size:,} bytes) exceeds threshold ({threshold:,} bytes) - using external storage"
+        )
+
+    return is_large
+
```
JadeCara: These top two functions might be useful in other areas as well. I wonder if putting them into a `data_size.py` type util would make them easier to find/use?

src/fides/api/service/external_data_storage.py
JadeCara, line 1: Even more proof we need a proper storage service/interface! 😉

src/fides/api/service/external_data_storage.py
```diff
@@ -0,0 +1,380 @@
...
+                    msg = f"Storage configuration with key '{storage_key}' not found"
+                    logger.error(msg)
+                    raise ExternalDataStorageError(msg)
+            else:
+                storage_config = get_active_default_storage_config(db)
+                if not storage_config:
+                    msg = "No active default storage configuration available for large data"
+                    logger.error(msg)
+                    raise ExternalDataStorageError(msg)
+
```
JadeCara: This function is doing a number of things - this top part might be a good candidate to pull out into its own testable function.

src/fides/api/service/external_data_storage.py
```diff
@@ -0,0 +1,380 @@
...
+                    .first()
+                )
+                if not storage_config:
+                    raise ExternalDataStorageError(
+                        f"Storage configuration with key '{metadata.storage_key}' not found"
+                    )
+            else:
+                storage_config = get_active_default_storage_config(db)
+                if not storage_config:
+                    raise ExternalDataStorageError("No storage configuration found")
```
JadeCara: Same comment as above :)

src/fides/api/service/external_data_storage.py
```diff
@@ -0,0 +1,380 @@
...
+                if not storage_config:
+                    logger.warning(
+                        f"Storage configuration with key '{metadata.storage_key}' not found for cleanup"
+                    )
+                    return
+            else:
+                storage_config = get_active_default_storage_config(db)
+                if not storage_config:
+                    logger.warning("No storage configuration found for cleanup")
+                    return
```
JadeCara: Same comment here!

src/fides/api/models/encrypted_large_data.py
galvana: I moved it under `/models/field_types`

src/fides/api/models/encrypted_large_data.py
galvana: This is one of the methods that are part of the descriptor protocol, we need to leave it even if we don't use it https://docs.python.org/3/reference/datamodel.html#object.__get__

src/fides/api/service/external_data_storage.py
galvana: Good call!

src/fides/api/models/field_types/encrypted_large_data.py
```diff
@@ -0,0 +1,151 @@
...
+    ) -> None:  # noqa: D401 (docstring in orig file)
+        self.name = name
+        self.model_class = owner.__name__
+
+    def _generate_storage_path(self, instance: Any) -> str:
+        instance_id = getattr(instance, "id", None)
+        if not instance_id:
+            raise ValueError(f"Instance {instance} must have an 'id' attribute")
+        timestamp = datetime.utcnow().strftime("%Y%m%d-%H%M%S-%f")
+        return f"{self.model_class}/{instance_id}/{self.field_name}/{timestamp}.txt"
```
galvana, line 48: @JadeCara I wonder if you have any thoughts on this format? I don't like the arbitrary timestamp here. I was thinking more of `{self.model_class}/{instance_id}/{self.field_name}.txt`

JadeCara: I think that tracks as long as you are confident there wont be any chance of accidental overwriting/cleaning etc from concurrent read/writes/cleanup. The time stamp does add a nice buffer in that its very unlikely that two operations will be happening at the same exact millisecond.

PR #6205, opened 2025-06-06
ENG 602 adding manual task model

JadeCara:
Issue  [ENG-602](https://ethyca.atlassian.net/browse/ENG-602)

### Description Of Changes

Adding a new models for manual tasks, manual task references, and manual task logs. These are the foundational models for building to the new ManualTask functionality needed for the ManualDsr project. I have also included the start of the ManualTask Service which deals with service layer implementations needed for the ManualDSR goals.

image

**This PR**

#6208 

#6212 

image

### Code Changes

* `ManualTask` class created with tests
* `ManualTaskReference` class created with tests
* `ManualTaskLogs` class created with tests
* Created a `StatusTransitionMixin` with tests to manage task statuses
* Added Necessary Schemas and enums for Manual Task types with tests
* Created `ManualTaskService` with the initial functionality for assigning users etc with tests

### Steps to Confirm

### Pre-Merge Checklist

* [x] Issue requirements met
* [x] `CHANGELOG.md` updated
* Followup issues:
  * [x] Followup issues created
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:


tests/ops/models/manual_tasks/test_manual_task.py
```diff
@@ -0,0 +1,296 @@
...
+    #         task_id=manual_task.id,
+    #         reference_id="test_config",
+    #         reference_type=ManualTaskReferenceType.manual_task_config,
+    #     )
+    #     db.add(config_ref)
+    #     db.commit()
+
+    #     # Verify configs relationship
+    #     assert len(manual_task.configs) == 1
+    #     assert manual_task.configs[0].id == "test_config"
```
JadeCara:
```suggestion
```

src/fides/api/models/manual_tasks/manual_task.py
```diff
@@ -0,0 +1,124 @@
...
+    # TODO: Add configs relationship when it is implemented
+    # configs = relationship(
+    #     "ManualTaskConfig",
+    #     secondary="manual_task_reference",
+    #     primaryjoin="and_(ManualTask.id == ManualTaskReference.task_id, ManualTaskReference.reference_type == 'manual_task_config')",
+    #     secondaryjoin="ManualTaskConfig.id == ManualTaskReference.reference_id",
+    #     viewonly=True,
+    # )
+    # TODO: Add submissions relationship when it is implemented
+    # submissions = relationship("ManualTaskSubmission", back_populates="task")
```
JadeCara:
```suggestion
```

src/fides/api/models/manual_tasks/manual_task.py
```diff
@@ -0,0 +1,114 @@
...
+        "ManualTaskLog",
+        back_populates="task",
+        primaryjoin="and_(ManualTask.id == ManualTaskLog.task_id)",
+        viewonly=True,
+        order_by="ManualTaskLog.created_at",
+    )
+    # TODO: Add instances relationship when it is implemented
+    # instances = relationship("ManualTaskInstance", back_populates="task")
+    # TODO: Add field_definitions relationship when it is implemented
+    # field_definitions = relationship("ManualTaskConfigField", back_populates="task")
```
JadeCara:
```suggestion
```

src/fides/api/alembic/migrations/versions/5efcdf18438e_add_manual_task_tables.py
```diff
@@ -0,0 +1,160 @@
...
+        sa.Column("parent_entity_id", sa.String(), nullable=False),
+        sa.Column("parent_entity_type", sa.String(), nullable=False),
+        sa.Column("due_date", sa.DateTime(timezone=True), nullable=True),
+        sa.PrimaryKeyConstraint("id"),
+        sa.UniqueConstraint(
+            "parent_entity_id",
+            "parent_entity_type",
+            name="uq_manual_task_parent_entity",
+        ),
+    )
```
erosselli, line 50: so a specific manual task can only have one child?

src/fides/api/alembic/migrations/versions/5efcdf18438e_add_manual_task_tables.py
```diff
@@ -0,0 +1,160 @@
...
+        sa.PrimaryKeyConstraint("id"),
+    )
+
+    # Create indexes for manual_task
+    op.create_index("ix_manual_task_task_type", "manual_task", ["task_type"])
+    op.create_index(
+        "ix_manual_task_parent_entity",
+        "manual_task",
+        ["parent_entity_id", "parent_entity_type"],
+    )
```
erosselli: will we ever want to query something like "manual tasks whose parent is of this type", or will we always query with a specific parent id?

src/fides/api/models/manual_tasks/manual_task.py
```diff
@@ -0,0 +1,110 @@
...
+    ManualTaskType,
+)
+
+
+class ManualTask(Base):
+    """Model for storing manual tasks.
+
+    This model can be used for both privacy request tasks and general tasks.
+    For privacy requests, it replaces the functionality of manual webhooks.
+    For other use cases, it provides a flexible task management system.
```
erosselli, line 23: Thanks for adding the docstring 🙌  Do we have other specific use cases in mind? If so I'd love to understand them . I'm just wondering how the "manual" in the name will apply to other use cases

src/fides/api/models/manual_tasks/manual_task.py
```diff
@@ -0,0 +1,110 @@
...
+
+    There can only be one ManualTask per parent entity.
+    You can create multiple Configs for the same ManualTask.
```
erosselli, line 26: I'm not sure I follow what the parent entity is --- I haven't gotten through all the code yet so it may become clearer as I go , but maybe we can add some more context here as well?

src/fides/api/schemas/manual_tasks/manual_task_schemas.py
```diff
@@ -0,0 +1,75 @@
...
+    paused = "paused"
+    awaiting_input = "awaiting_input"
+
+
+class ManualTaskLogCreate(FidesSchema):
+    """Schema for creating a manual task log entry."""
+
+    model_config = ConfigDict(extra="forbid")
+
+    task_id: str = Field(..., description="ID of the task")
```
erosselli: what's the ellipsis in `Field(..., description="ID of the task")`

src/fides/api/schemas/manual_tasks/manual_task_status.py
```diff
@@ -0,0 +1,146 @@
...
+    completed_at: Optional[datetime]
+    completed_by_id: Optional[str]
+
+    def _get_valid_transitions(self) -> list[StatusType]:
+        """Get valid transitions from the current status.
+
+        Returns:
+            list[StatusType]: List of valid transitions
+        """
+        return StatusType.get_valid_transitions(self.status)
```
erosselli, line 63: shouldn't this be `T` rather than `StatusType`?

src/fides/api/schemas/manual_tasks/manual_task_status.py
```diff
@@ -0,0 +1,146 @@
...
+    This mixin provides methods for managing status transitions and completion tracking.
+    It can be used by any model that needs status management.
+    """
+
+    # These should be overridden by the implementing class
+    status: T
+    completed_at: Optional[datetime]
+    completed_by_id: Optional[str]
+
+    def _get_valid_transitions(self) -> list[StatusType]:
```
erosselli, line 57: same question here, shouldn't return value be `list[T]` ?

src/fides/api/schemas/manual_tasks/manual_task_status.py
```diff
@@ -0,0 +1,146 @@
...
+
+    def _validate_status_transition(self, new_status: T) -> None:
+        """Validate that a status transition is allowed.
+
+        Args:
+            new_status: The new status to transition to
+
+        Raises:
+            ValueError: If the transition is not allowed
+        """
```
erosselli: Do you think we might want to create a `TransitionNotAllowedError`? I feel like it might be useful to be able to catch this error specifically. It could also inherit from `ValueError` if it'd be useful

src/fides/api/schemas/manual_tasks/manual_task_status.py
```diff
@@ -0,0 +1,146 @@
...
+            db: Database session
+            new_status: New status to set
+            user_id: Optional user ID who is making the change
+        """
+        self._validate_status_transition(new_status)
+
+        if new_status == StatusType.completed:
+            self.completed_at = datetime.now(timezone.utc)
+            self.completed_by_id = user_id
+        elif new_status == StatusType.pending:
```
erosselli, line 103: same here, should we be using `T` rather than `StatusType`?

src/fides/api/schemas/manual_tasks/manual_task_status.py
```diff
@@ -0,0 +1,146 @@
...
+
+    @property
+    def is_in_progress(self) -> bool:
+        """Check if in progress."""
+        return self.status == StatusType.in_progress
+
+    @property
+    def is_pending(self) -> bool:
+        """Check if pending."""
+        return self.status == StatusType.pending
```
erosselli, line 151: do we not need to `cast(T, StatusType)` like it's done above?  sorry if these are stupid questions, I haven't done generics in Python in a while 😅

src/fides/service/manual_tasks/manual_task_service.py
```diff
@@ -0,0 +1,142 @@
...
+            task_type: The task type
+
+        Returns:
+            Optional[ManualTask]: The manual task for the connection, if it exists
+        """
+        if not any([task_id, parent_entity_id, parent_entity_type, task_type]):
+            logger.warning("No filters provided to get_task. Returning None.")
+            return None
+
+        stmt = select(ManualTask)  # type: ignore[arg-type]
```
erosselli, line 44: why is the type checker complaining about a select? 😢 

(also using new-style SQLAlchemy syntax 🤩 )

src/fides/service/manual_tasks/manual_task_service.py
```diff
@@ -0,0 +1,142 @@
...
+            stmt = stmt.where(ManualTask.parent_entity_type == parent_entity_type)
+        if task_type:
+            stmt = stmt.where(ManualTask.task_type == task_type)
+        return self.db.execute(stmt).scalar_one_or_none()
+
+    # User Management
+    def assign_users_to_task(
+        self, db: Session, task: ManualTask, user_ids: list[str]
+    ) -> None:
+        """Assign a user to this task. This assumes there is a single user assigned to the task.
```
erosselli: I'm a bit confused about the wording here , should "This assumes there is a single user assigned to the task." be something like "This assumes only a single user can be assigned to the task" ?  Or if not , what is this sentence trying to convey? 

also I'm confused about why we pass in a list of users and not only one , though the method name does say `assign_users`  in plural

src/fides/service/manual_tasks/manual_task_service.py
```diff
@@ -0,0 +1,142 @@
...
+
+        Args:
+            db: Database session
+        """
+        user_ids = list(set(user_ids))
+        if not user_ids:
+            raise ValueError("User ID is required for unassignment")
+
+        unassigned_user_ids = []
+        for ref in task.references:
```
erosselli: do we need to iterate over all references? maybe we can run a `task.references.filter(Reference.reference_type==ManualTaskReferenceType.assigned_user, Reference.reference_id._in(user_ids)` and that would make the code below simpler , without needing all the if statements

tests/ops/models/manual_tasks/test_manual_task.py
erosselli, line 1: rather than creating more new tests under `ops`, which we want to deprecate, can we maybe move these out into `tests/api/models` ?

tests/ops/schemas/manual_tasks/test_manual_task_status.py
```diff
@@ -0,0 +1,138 @@
...
+    def model(self):
+        return TestModel()
+
+    @pytest.fixture
+    def mock_db(self):
+        db = MagicMock()
+        db.add = MagicMock()
+        db.commit = MagicMock()
+        return db
+
```
erosselli: I'm curious, what would be the reason to mock the DB?

tests/ops/service/manual_tasks/test_manual_task_service.py
```diff
@@ -0,0 +1,432 @@
...
+    ManualTaskType,
+)
+from fides.service.manual_tasks.manual_task_service import ManualTaskService
+
+
+@pytest.fixture
+def manual_task(db: Session):
+    task = ManualTask.create(
+        db=db,
+        data={
```
erosselli: do you think we'll want to reuse these fixtures? if so, do you think it would be useful to move them to a `manual_task_fixtures.py` file ?

erosselli (commented):
thanks for the very thorough testing!  I left a lot of questions, mostly cause I don't entirely understand all the models etc etc

src/fides/api/alembic/migrations/versions/5efcdf18438e_add_manual_task_tables.py
JadeCara: There can be only one ManualTask for the parent entity. This is currently how AccessManualWebhooks work with Connection configs and this is sort of the version of that.  

The ManualTask can have multiple configs so its still pretty open ended, but this makes it work more like an integration in that sense.

src/fides/api/alembic/migrations/versions/5efcdf18438e_add_manual_task_tables.py
JadeCara: That is a really interesting question. I am not sure. I can definitely add an index for it! I think it makes sense. I will update to add that index :) 

The id + type was just me being over cautious in case we randomly had some id happen twice on different types - probably unlikely but better safe than sorry!

src/fides/api/models/manual_tasks/manual_task.py
JadeCara: The "manual" in the name means some human will have to take an action. I am not sure specifically where we would need this yet, but building it open ended seemed like a good goal, and was mostly just being careful with naming and not tying it specifically to connection configs while allowing them to be referenced in the model.

src/fides/api/models/manual_tasks/manual_task.py
JadeCara: The parent entity is whatever element is getting a manual task added to it. For the main use case it will be connection configs and the manual task will be an integration (replacing the manual webhook version)

src/fides/api/schemas/manual_tasks/manual_task_schemas.py
JadeCara: The `...` indicates that the field is required for pydantic models. However in looking at this and reading up to make sure I knew what I was doing (heh) I read a little more on how Fields and type checking don't always play nicely together and the suggestion was to move to Annotated fields for that purpose so I am going to a quick update on that here. 

Here is a [link](https://docs.pydantic.dev/latest/concepts/fields/) to the doc :)

src/fides/api/schemas/manual_tasks/manual_task_status.py
JadeCara: OK - had a good think about this and I think using` Generic[T]` is actually just the wrong approach. 
This functionality is supposed to specifically work with the `StatusTypes`. They are only separate so the `Types` themselves can be used other places, not because we might have other types we want to use with this mixin. The generic `T` is adding complexity without providing any real benefit. So I am going to update to just use the `StatusTypes` instead :)

src/fides/api/schemas/manual_tasks/manual_task_status.py
JadeCara: This one is a string comparison so we cannot use the `T` I am going to reimplement this in the service when it comes up that we need to make this update which wont be until `Instances` Thanks for the good call out!

src/fides/api/schemas/manual_tasks/manual_task_status.py
JadeCara: These aren't stupid questions! You are totally right! I am making the update above to handle this all a little more clearly :)

src/fides/service/manual_tasks/manual_task_service.py
JadeCara: This seems to be an issue related to mypy and the version of sqlalchemy we are using not wanting to play nicely. 

This works from a sqlalchemy perspective but does not provide the type hints that mypy wants. 

The newer syntax with proper typing support is available in SQLAlchemy 2.0 and above. SQLAlchemy 2.0 introduced a new typing system and query API that would allow us to write:
```
stmt = select(ManualTask).options(selectinload(ManualTask.references))
```
which would make mypy happy and do the same thing as above.

src/fides/service/manual_tasks/manual_task_service.py
JadeCara: My bad! I updated it to do multiple users and forgot to update the doc string :) Updating now.

src/fides/service/manual_tasks/manual_task_service.py
JadeCara: love it - updating now :)

tests/ops/models/manual_tasks/test_manual_task.py
JadeCara: I will move them around :)

tests/ops/service/manual_tasks/test_manual_task_service.py
JadeCara: I kind of don't. The tasks should be pretty specific to use case - They are also not super hard to implement and keeping the fixture closer to the usage means that you can make updates as necessary without having echo's to other tests.

tests/ops/schemas/manual_tasks/test_manual_task_status.py
JadeCara: The mock allows us to check a couple things in isolation. The `test_model` above is not a real thing and since this is a mixin it should work with any model. This set up allows us to use a fake test model without additional overhead or dependencies. 

The fact the db is also a mock in this case means that we can use the mock functionality like `mock_db.add.assert_called_once()` 

It kind of keeps these tests as true unit tests rather than integration tests. There are other tests for integration on the models that will be using this mixin :)

src/fides/api/schemas/manual_tasks/manual_task_status.py
erosselli: awesome :) we can always make it generic down the road if we need to

src/fides/service/manual_tasks/manual_task_service.py
erosselli: ah got it .. one day we'll update sqlalchemy and life will be good 🥲

tests/ops/service/manual_tasks/test_manual_task_service.py
erosselli: fair enough 😄

src/fides/api/schemas/manual_tasks/manual_task_status.py
```diff
@@ -0,0 +1,151 @@
...
+        return StatusType.get_valid_transitions(self.status)
+
+    def _validate_status_transition(self, new_status: StatusType) -> None:
+        """Validate that a status transition is allowed.
+
+        Args:
+            new_status: The new status to transition to
+
+        Raises:
+            ValueError: If the transition is not allowed
```
erosselli:
```suggestion
        Raises:
            StatusTransitionNotAllowed: If the transition is not allowed
```

erosselli (approved):
Thanks for the explanations and making the changes! remember the changelog entry

PR #6206, opened 2025-06-08
ENG-719:  Assume role for SES

tvandort:
Closes [ENG-719]

### Description Of Changes

Allow SES to assume roles.

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:

[ENG-719]: https://ethyca.atlassian.net/browse/ENG-719?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,9 @@ Changes can also be flagged with a GitHub label for tracking purposes. The URL o
 
 ## [Unreleased](https://github.com/ethyca/fides/compare/2.64.0...main)
 
+### Addded
+- AWS SES notification service now supports assumed roles through environment variable configuration through `FIDES__CREDENTIALS__NOTIFICATIONS__AWS_SES_ASSUME_ROLE_ARN` [#6206](https://github.com/ethyca/fides/pull/6206)
+
 ### Changed
 
 - TCF Banners will no longer resurface on reload after dismissal [#6200](https://github.com/ethyca/fides/pull/6200)


src/fides/service/messaging/aws_ses_service.py
```diff
@@ -86,7 +87,10 @@ def get_ses_client(self) -> SESClient:
         aws_session = get_aws_session(
             auth_method=self.messaging_config_secrets.auth_method.value,
             storage_secrets=storage_secrets,  # type: ignore[arg-type]
-            assume_role_arn=self.messaging_config_secrets.aws_assume_role_arn,
+            assume_role_arn=CONFIG.credentials.get(  # pylint: disable=no-member
+                "notifications", {}
+            ).get("aws_ses_assume_role_arn")
+            or self.messaging_config_secrets.aws_assume_role_arn,
```
erosselli, line 93: this looks ok to me -- I'm just wondering what the order should be here , do we usually "prioritize" the env-set credentials? trying to think of what would follow the principle of least surprise for users configuring this integration

tvandort: I had a similar thought. It's... hard for me to say. For some reason I have this gut feeling that the _most_ expected thing would be for the API to win over other methods.

But the API is less flexible in some cases bc it's a central point of truth. The reason I need to support this is for use cases where Fides is deployed to multiple regions, and this allows us to configure the environment variable granularly per region so it _feels_ more specific in that sense.

Ultimately we chose to let the environment variable to have precedence before: 

https://github.com/ethyca/fides/blob/7ca53a4a0fa92d7344d2ceb8276c4f07f09a228e/src/fides/api/tasks/storage.py#L139-L141

https://github.com/ethyca/fides/blob/7ca53a4a0fa92d7344d2ceb8276c4f07f09a228e/src/fides/api/util/aws_util.py#L53-L54

So I want to keep doing that probably?

erosselli: ah that makes sense -- let's keep it consistent with the S3 configuration

PR #6207, opened 2025-06-09
Handle edge case in staged resource ancestor data migration

adamsachs:
### Description Of Changes

Handles an edge case in the staged resource ancestor creation where we have `children` pointers in our staged resources that refer to URNs of nonexistent `stagedresource`s. This is an unexpected state, but it's one we've seen in practice, so we should guard against it.

Without this update, we'd create `stagedresourceancestor` links that point to the nonexistent `stagedresource` records, and then we get a constraint error when trying to create the FK constraint on `stagedresourceancestor.descendant_urn`.

### Code Changes

- [x] in our data migration: keep track of all existent `stagedresource` records, and only create `stagedresourceancestor` links to existent records
- [x] tweak some logging in our index creation script to be more accurate 

### Steps to Confirm

1. confirmed this handled our edge case locally with a sample dataset that showed the problem initially

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:

### DIFF
--- a/src/fides/api/alembic/migrations/versions/bf713b5a021d_staged_resource_ancestor_link_data_.py
+++ b/src/fides/api/alembic/migrations/versions/bf713b5a021d_staged_resource_ancestor_link_data_.py
@@ -74,6 +74,11 @@ def upgrade():
             children = result.children
             if children:
                 resource_children[urn] = children
+            else:
+                # even if no children, we still add the resource to the map
+                # so that we maintain a record of all resources and can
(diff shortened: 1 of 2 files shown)


vcruces (approved):
LGTM!

PR #6208, opened 2025-06-09
ENG-602 ManualTaskConfig

JadeCara:
Issue [ENG-602](https://ethyca.atlassian.net/browse/ENG-602)

### Description Of Changes

Adding a new config models for manual tasks. These are the foundational models for building to the new ManualTask functionality needed for the ManualDsr project.

image

#6205 

**This PR**

#6212 

image

### Code Changes

* Created ManualTaskConfig class with tests
* Created ManualTaskConfigField class with tests 
* Created new schema file for the above classes
* Created a ManualTaskConfigService with tests.
* Updated areas that should have relationships to configs from the previous PR #6205 

### Steps to Confirm

### Pre-Merge Checklist

* [x] Issue requirements met
* [x] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ Changes can also be flagged with a GitHub label for tracking purposes. The URL o
 
 ### Added
 - Added support for Salesforce custom object monitoring using Helios [#6096](https://github.com/ethyca/fides/pull/6096)
+- Added ManualTaskConfig and ManualTaskConfigField models, foundational for for ManualDSRs [#6208](https://github.com/ethyca/fides/pull/6208) https://github.com/ethyca/fides/labels/db-migration
 
 ### Changed
 - Integrations detail page now includes a setup guide component. [#6096](https://github.com/ethyca/fides/pull/6096)


src/fides/api/alembic/migrations/versions/ba414a58ba90_create_manual_task_config_tables.py
```diff
@@ -0,0 +1,125 @@
...
+        sa.Column("config_type", sa.String(), nullable=False),
+        sa.Column("version", sa.Integer(), nullable=False, server_default="1"),
+        sa.Column("is_current", sa.Boolean(), nullable=False, server_default="true"),
+        sa.ForeignKeyConstraint(["task_id"], ["manual_task.id"], ondelete="CASCADE"),
+        sa.PrimaryKeyConstraint("id"),
+        sa.CheckConstraint(
+            "config_type IN ('access_privacy_request', 'erasure_privacy_request')",
+            name="valid_config_type",
+        ),
+    )
```
erosselli: can you explain why we need this checkConstraint?

src/fides/api/models/manual_tasks/manual_task.py
```diff
@@ -14,6 +14,9 @@
     ManualTaskType,
 )
 
+if TYPE_CHECKING:
+    from fides.api.models.manual_tasks.manual_task_config import ManualTaskConfig
```
erosselli: since these aren't covered by tests (that's ok) , can we add the no cover comment so they don't get included in our coverage report? otherwise we slowly decrease project coverage 😅

src/fides/api/models/manual_tasks/manual_task_config.py
```diff
@@ -0,0 +1,155 @@
...
+    """Model for storing manual task configurations.
+    A single configuration may have many fields of different types.
+    """
+
+    @declared_attr
+    def __tablename__(cls) -> str:
+        return "manual_task_config"
+
+    task_id = Column(String, ForeignKey("manual_task.id"), nullable=False)
+    config_type = Column(String, nullable=False)  # Using ManualTaskConfigurationType
```
erosselli: can you share why we don't want this to be an enum column? do we expect we'll be adding a lot of options to these enums?

src/fides/api/models/manual_tasks/manual_task_config.py
```diff
@@ -0,0 +1,155 @@
...
+            },
+        )
+        return config
+
+    def get_field(self, field_key: str) -> Optional["ManualTaskConfigField"]:
+        """Get a field by its key."""
+        for field in self.field_definitions:
+            if field.field_key == field_key:
+                return field
+        return None
```
erosselli, line 82: maybe we can add a test for getting a field that doesn't exist , so that we cover this line? codecov's saying it's not covered (sometimes it gets it wrong so ignore if we are testing that case)

src/fides/api/schemas/manual_tasks/manual_task_config.py
```diff
@@ -0,0 +1,245 @@
...
+        None,
+        description=(
+            "Regex pattern for validation, for example if the field is an email address, "
+            r"you could use: ^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,}$"
+        ),
+    )
+
+    @field_validator("min_length", "max_length")
+    @classmethod
+    def validate_lengths(cls, v: Optional[int], info: ValidationInfo) -> Optional[int]:
```
erosselli, line 73: there's a couple of uncovered lines in this method -- maybe we can add unit tests for those cases? (again, if codecov is wrong just ignore this comment)

src/fides/api/schemas/manual_tasks/manual_task_config.py
```diff
@@ -0,0 +1,245 @@
...
+
+class ManualTaskTextField(FidesSchema):
+    """Schema for text field definition."""
+
+    model_config = ConfigDict(extra="allow")
+
+    field_key: str = Field(..., description="Unique key for the field")
+    field_type: ManualTaskFieldType = Field(
+        ManualTaskFieldType.text, description="Type of the field"
+    )
```
erosselli: what's the first argument to Field here? also shouldn't the type of `field_type` here be `ManualTaskFieldType.text` rather than `ManualTaskFieldType` ?

src/fides/api/schemas/manual_tasks/manual_task_config.py
```diff
@@ -0,0 +1,245 @@
...
+    field_metadata: ManualTaskAttachmentFieldMetadata = Field(
+        ..., description="Field metadata and configuration"
+    )
+
+
+# Union type for all field types
+ManualTaskField = Union[
+    ManualTaskTextField, ManualTaskCheckboxField, ManualTaskAttachmentField
+]
+
```
erosselli: this is a just a thought so feel free to tell me it's wrong/not useful -- would we benefit from having the three fields inherit from a base class and then use that as the "type " in the classes below? or is the union type better?

src/fides/service/manual_tasks/manual_task_config_service.py
```diff
@@ -0,0 +1,375 @@
...
+            try:
+                field_type = field.get("field_type")
+                if not field_type:
+                    raise ValueError("Invalid field data: field_type is required")
+                if field_type == ManualTaskFieldType.text:
+                    ManualTaskTextField.model_validate(field)
+                elif field_type == ManualTaskFieldType.checkbox:
+                    ManualTaskCheckboxField.model_validate(field)
+                elif field_type == ManualTaskFieldType.attachment:
+                    ManualTaskAttachmentField.model_validate(field)
```
erosselli: I imagine we'll be writing these kinds  of if statements in more than one place.... can we maybe do something like this (if we had a generic `ManualTaskField` base class -- otherwise the method could be in `ManualTaskFieldType` or another place) 
```
field_model = ManualTaskFieldBase.get_field_model_for_type(field_type)
field_model.model_validate(field)
```

src/fides/service/manual_tasks/manual_task_config_service.py
```diff
@@ -0,0 +1,375 @@
...
+            except ValidationError as e:
+                raise ValueError(f"Invalid field data: {str(e)}")
+
+    def _re_create_existing_fields(
+        self,
+        previous_config: ManualTaskConfig,
+        new_config: ManualTaskConfig,
+        modified_field_keys: set[str],
+    ) -> None:
+        """Re-create existing fields that aren't being modified with references to the new config."""
```
erosselli: can we maybe explain in the docstring when this would be needed ?

src/fides/service/manual_tasks/manual_task_config_service.py
```diff
@@ -0,0 +1,375 @@
...
+        Args:
+            new_config: The config to add fields to
+            field_updates: The fields to add or update
+        """
+        self._validate_fields(field_updates)
+
+        for update in field_updates:
+            # Do not recreate if this field is empty - This is a removal
+            if update["field_metadata"] == {}:
+                continue
```
erosselli: it feels like it'll be a bit confusing down the line that the metadata being empty means its a removal (we have a similar check in the method above)... is there any way we could make this more explicit in the code itself rather than with a comment?

src/fides/service/manual_tasks/manual_task_config_service.py
```diff
@@ -0,0 +1,375 @@
...
+                ManualTaskConfigField.field_key == field_key
+            )
+        if version:
+            query = query.filter(ManualTaskConfig.version == version)
+        if config_type:
+            query = query.filter(ManualTaskConfig.config_type == config_type)
+
+        result = query.first()
+        if not result:
+            logger.warning(
```
erosselli: are we logging this at warning level rather than debug because we expect that whenever this method is called, a result should be returned?

src/fides/service/manual_tasks/manual_task_config_service.py
```diff
@@ -0,0 +1,375 @@
...
+        if not config:
+            raise ValueError(f"Config with ID {current_config.id} not found")
+
+        # Create field updates with empty metadata to indicate removal
+        fields_to_remove = [
+            {
+                "field_key": field.field_key,
+                "field_type": field.field_type,
+                "field_metadata": {},
+            }
```
erosselli: maybe we can just explicitly pass in `fields_to_remove` and `fields_to_update` to the create_new_version method, rather than using the emptiness of the metadata attribute

src/fides/api/alembic/migrations/versions/ba414a58ba90_create_manual_task_config_tables.py
JadeCara: We don't! Updating to use enum columns, thanks!

src/fides/api/models/manual_tasks/manual_task_config.py
JadeCara: Nah, I am going to update to use enum column :) Thanks!

src/fides/api/models/manual_tasks/manual_task_config.py
JadeCara: Added!

src/fides/api/schemas/manual_tasks/manual_task_config.py
JadeCara: Added some extra tests across this file and did a little clean up.

src/fides/api/schemas/manual_tasks/manual_task_config.py
JadeCara: I think that was an artifact of the general model that I didnt clean up properly. I have fixed these to be `Literal[ManualTaskFieldType.text, ...` which is more correct.

src/fides/api/schemas/manual_tasks/manual_task_config.py
JadeCara: Nope, love it - putting it in now. :)

src/fides/service/manual_tasks/manual_task_config_service.py
JadeCara: Love it :)

src/fides/service/manual_tasks/manual_task_config_service.py
JadeCara: Yeah - I think that makes more sense. I will update.

src/fides/service/manual_tasks/manual_task_config_service.py
JadeCara: Agreed - going with your comment below and having the remove fields passed in separately

src/fides/service/manual_tasks/manual_task_config_service.py
JadeCara: Good point, updating to debug :)

src/fides/api/alembic/migrations/versions/ba414a58ba90_create_manual_task_config_tables.py
```diff
@@ -0,0 +1,121 @@
...
+            server_default=sa.text("now()"),
+        ),
+        sa.Column(
+            "updated_at",
+            sa.DateTime(timezone=True),
+            nullable=False,
+            server_default=sa.text("now()"),
+        ),
+        sa.Column("task_id", sa.String(), nullable=False),
+        sa.Column("config_type", sa.String(), nullable=False),
```
erosselli, line 38: are enum columns just strings in the migration?

erosselli (approved):
thanks for making the changes! I left a non-blocking question; also remember the changelog entry and to update the down_revision of the migration as needed

src/fides/api/alembic/migrations/versions/ba414a58ba90_create_manual_task_config_tables.py
JadeCara: So there are multiple ways to handle this. 

I have gone for having the EnumColumn defined in the ORM model to:
- Validate that only valid enum values can be stored
- Automatically convert between Python enum objects and their string representations
- Provide type safety at the application level

At the database level, enums are typically stored as strings and these are specifically.
Because the enum validation/constraint is enforced at the application layer through SQLAlchemy's type system, this approach provides more flexibility for future enum value changes without requiring database migrations.

In practice this means the underlying database just sees these as string columns, which is what the migration reflects. This is generally considered a good practice as it keeps the database schema simpler and more flexible while maintaining type safety through the ORM layer. In this case because there may be more enums added having the flexibility to handle that in the model seems preferable. This model is designed to be re-used and having to do a database migration for each new use case is less ideal. For intentionally single use hardened models defining the enums in the migration makes more sense.

PR #6210, opened 2025-06-10
Disable AMD in FidesJS

gilluminate:
Closes [ENG-724]

### Description Of Changes

Fixed AMD module loading conflicts by preventing FidesJS from registering itself as an AMD module when RequireJS or other AMD loaders are present on the page.

### Code Changes

* Added `amd: { define: undefined }` configuration to Rollup UMD output to prevent AMD module registration

### Steps to Confirm

1. Run `npm run build` in `clients/fides-js` to generate bundles
2. Test that FidesJS loads correctly in browser environments without RequireJS conflicts by adding `<script src="https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.6/require.min.js"></script>` near the top of `clients/privacy-center/public/fides-js-demo.html`. Load the demo page with a valid experience and make sure FidesJS still runs correctly without errors (`jsonview.js` will throw an error, but that's not our problem).

### Pre-Merge Checklist

* [x] Issue requirements met
* [x] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [x] No followup issues
* Database migrations:
  * [x] No migrations
* Documentation:

[ENG-724]: https://ethyca.atlassian.net/browse/ENG-724?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/clients/fides-js/rollup.config.mjs
+++ b/clients/fides-js/rollup.config.mjs
@@ -152,6 +152,9 @@ SCRIPTS.forEach(({ name, gzipErrorSizeKb, gzipWarnSizeKb, isExtension }) => {
         name: isExtension ? undefined : "Fides",
         format: isExtension ? undefined : "umd",
         sourcemap: IS_DEV ? "inline" : false,
+        amd: {
+          define: undefined, // prevent the bundle from registering itself as an AMD module, even if an AMD loader (like RequireJS) is present on the page. (FidesJS is not an AMD module.)
+        },
       },
     ],
   };
--- a/clients/fides-js/rollup.config.mjs
+++ b/clients/fides-js/rollup.config.mjs
@@ -150,10 +150,10 @@ SCRIPTS.forEach(({ name, gzipErrorSizeKb, gzipWarnSizeKb, isExtension }) => {
         // Intended for browser <script> tag - defines `Fides` global. Also supports UMD loaders.
         file: `dist/${name}.js`,
         name: isExtension ? undefined : "Fides",
-        format: isExtension ? undefined : "umd",
+        format: isExtension ? "es" : "umd",
         sourcemap: IS_DEV ? "inline" : false,
         amd: {
-          define: undefined, // prevent the bundle from registering itself as an AMD module, even if an AMD loader (like RequireJS) is present on the page. (FidesJS is not an AMD module.)
+          define: undefined, // prevent the bundle from registering itself as an AMD module, even if an AMD loader (like RequireJS) is present on the page. This allows FidesJS to use Rollup's `umd` format to support both `iife` and `cjs` modules, but excludes AMD.
         },
       },
     ],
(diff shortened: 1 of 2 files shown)


clients/fides-js/rollup.config.mjs
```diff
@@ -152,6 +152,9 @@ SCRIPTS.forEach(({ name, gzipErrorSizeKb, gzipWarnSizeKb, isExtension }) => {
         name: isExtension ? undefined : "Fides",
         format: isExtension ? undefined : "umd",
```
NevilleS: nit:
```suggestion
        format: isExtension ? "es" : "umd",
```

This just feels like it's better at communicating what we're doing here

clients/fides-js/rollup.config.mjs
```diff
@@ -152,6 +152,9 @@ SCRIPTS.forEach(({ name, gzipErrorSizeKb, gzipWarnSizeKb, isExtension }) => {
...
         sourcemap: IS_DEV ? "inline" : false,
+        amd: {
+          define: undefined, // prevent the bundle from registering itself as an AMD module, even if an AMD loader (like RequireJS) is present on the page. (FidesJS is not an AMD module.)
+        },
```
NevilleS: Q: is the reason you can't use `format: "iife"` specifically because you still want the `cjs` output format to work? 

If so I think this is OK, but my comment would look like this:
```suggestion
        amd: {
          define: undefined, // prevent the bundle from registering itself as an AMD module, even if an AMD loader (like RequireJS) is present on the page. This allows FidesJS to use Rollup's `umd` format to support both `iife` and `cjs` modules, but excludes AMD.
        },
```

NevilleS (approved):
One question... but this looks to be a reasonable workaround. Thanks

clients/fides-js/rollup.config.mjs
NevilleS: FWIW: the only alternative I see to this would be:
1. Use `format: "iife"` for the `dist/fides.js` export (which is a potentially-breaking change!)
2. Use the `dist/fides.mjs` export for our own repos that want to import Fides, and fight with npm about that....

I can't conclude whether that would work though and I know you tried some of this, so I'll trust what you experimented with!

gilluminate: I did try that and we still had issues that were getting really messy to try and solve. I had it mostly figured out in Admin UI and in Privacy Center, but then ALL Cypress tests were failing because things were missing on the window object. I got to that point after about 8 hours of playing around with it and opted to attempt this solution instead, which seems to be least risky and solves the edge case. It might be worth revisiting at some point when we have more time and resources.

PR #6212, opened 2025-06-10
Eng-762 manual task instances

JadeCara:
Issue [ENG-762](https://ethyca.atlassian.net/browse/ENG-762)

### Description Of Changes

Adding a new config models for manual tasks. These are the foundational models for building to the new ManualTask functionality needed for the ManualDsr project.

image

#6205 

#6208 

**This PR**

image

### Code Changes

* Created ManualTaskInstance class with tests
* Created ManualTaskSubmission class with tests 
* Created a ManualTaskInstanceService with tests.
* Updated areas that should have relationships to instances from the previous PR #6208

### Steps to Confirm

### Pre-Merge Checklist

* [x] Issue requirements met
* [x] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:

[ENG-762]: https://ethyca.atlassian.net/browse/ENG-762?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/fides/service/manual_tasks/manual_task_config_service.py
```diff
@@ -50,37 +60,6 @@ def _get_response_data(self, config: ManualTaskConfig) -> dict[str, Any]:
             "updated_at": config.updated_at,
         }
 
-    def _validate_fields(self, fields: list[dict[str, Any]]) -> None:
```
JadeCara: Moved this to util.py so instances can use it too!

src/fides/service/manual_tasks/manual_task_config_service.py
```diff
@@ -300,60 +343,56 @@ def create_new_version(
...
-                new_config,
-                modified_field_keys,
+                previous_config, new_config, modified_field_keys
             )
 
-        self.db.commit()
+        self.db.flush()
 
-        # Log the version creation
-        ManualTaskLog.create_log(
```
JadeCara: All task logging and error logging is happening through the decorator now.

src/fides/api/alembic/migrations/versions/6a76a1fa4f3f_add_manual_task_instance_table.py
```diff
@@ -0,0 +1,206 @@
...
+    )
+    # Composite index for common query pattern
+    op.create_index(
+        "ix_manual_task_submission_instance_field",
+        "manual_task_submission",
+        ["instance_id", "field_id"],
+    )
+
+    # Add foreign key constraint to manual_task_log.instance_id
+    op.create_foreign_key(
```
JadeCara, line 164: Adding the instances in as a foreign key now that we have them!

src/fides/api/models/manual_tasks/manual_task_config.py
```diff
@@ -13,10 +13,19 @@
...
+)
 
 if TYPE_CHECKING:  # pragma: no cover
     from fides.api.models.manual_tasks.manual_task import ManualTask  # pragma: no cover
+    from fides.api.models.manual_tasks.manual_task_instance import (
+        ManualTaskInstance,  # pragma: no cover
+    )
+    from fides.api.models.manual_tasks.manual_task_instance import (
+        ManualTaskSubmission,  # pragma: no cover
+    )
```
JadeCara, line 28:
```suggestion
        ManualTaskInstance,  # pragma: no cover
        ManualTaskSubmission,  # pragma: no cover
    )
```

.fides/db_dataset.yml
```diff
@@ -1110,6 +1110,52 @@ dataset:
       data_categories: [system.operations]
     - name: task_id
       data_categories: [system.operations]
+    - name: execution_timing
+      data_categories: [system.operations]
```
JadeCara:
```suggestion
```

src/fides/api/models/manual_tasks/manual_task_instance.py
```diff
@@ -0,0 +1,179 @@
...
+    @declared_attr
+    def __tablename__(cls) -> str:
+        """Overriding base class method to set the table name."""
+        return "manual_task_instance"
+
+    # Database columns
+    task_id = Column(String, ForeignKey("manual_task.id"), nullable=False)
+    config_id = Column(String, ForeignKey("manual_task_config.id"), nullable=False)
+    # entity id is the entity that the instance is for (e.g. privacy request)
+    # TODO: Add to schemas
```
vcruces: Is this TODO intentional here?

src/fides/api/models/manual_tasks/manual_task_instance.py
```diff
@@ -0,0 +1,179 @@
...
+    entity_id = Column(String, nullable=False)
+    entity_type = Column(String, nullable=False)
```
vcruces: Is there an enum for this? Is it worth creating?

src/fides/api/models/manual_tasks/manual_task_instance.py
```diff
@@ -0,0 +1,179 @@
...
+    status: StatusType = cast(
+        StatusType, Column(String, nullable=False, default=StatusType.pending)
+    )
+    completed_at: Optional[datetime] = cast(
+        Optional[datetime], Column(DateTime, nullable=True)
+    )
+    completed_by_id: Optional[str] = cast(Optional[str], Column(String, nullable=True))
```
vcruces: Why are these `StatusTransitionMixin` fields rewritten? Is it due to the cast?

src/fides/api/models/manual_tasks/manual_task_log.py
```diff
@@ -26,19 +29,29 @@ def __tablename__(cls) -> str:
     task_id = Column(
         String, ForeignKey("manual_task.id", ondelete="CASCADE"), nullable=False
     )
-    config_id = Column(String, ForeignKey("manual_task_config.id"), nullable=True)
-    instance_id = Column(String, nullable=True)
+    config_id = Column(
+        String, ForeignKey("manual_task_config.id", ondelete="SET NULL"), nullable=True
```
vcruces: Just to be sure: does it make sense to keep the logs even if their associated config or instance has been deleted? Do they provide enough information to be understandable and useful? Or would it be better to delete them in cascade?

src/fides/service/manual_tasks/manual_task_instance_service.py
```diff
@@ -0,0 +1,262 @@
...
+
+    def _create_log_data(
+        self,
+        task_id: str,
+        config_id: str,
+        instance_id: str,
+        details: dict[str, Any],
+    ) -> dict[str, Any]:
+        """Create standard log data structure."""
+        return {
```
vcruces, line 47: Will this be used later to create ManualTaskLogs?

tests/api/models/manual_tasks/test_manual_task_instance.py
```diff
@@ -0,0 +1,397 @@
...
+    assert manual_task_instance.status == StatusType.in_progress
+
+    # Transition to completed
+    manual_task_instance.mark_completed(db, "user1")
+    assert manual_task_instance.status == StatusType.completed
+    assert manual_task_instance.completed_at is not None
+    assert manual_task_instance.completed_by_id == "user1"
+
+
+def test_create_manual_task_submission(
```
vcruces: What do you think about grouping these tests into two classes? One class for `ManualTaskInstance` and another for `ManualTaskSubmission`

tests/service/manual_tasks/conftest.py
```diff
@@ -4,8 +4,12 @@
 from fides.api.models.fides_user import FidesUser
 from fides.api.models.fides_user_permissions import FidesUserPermissions
 from fides.api.models.manual_tasks.manual_task import ManualTask, ManualTaskReference
-from fides.api.models.manual_tasks.manual_task_config import ManualTaskConfig
+from fides.api.models.manual_tasks.manual_task_config import (
+    ManualTaskConfig,
+    ManualTaskConfigField,
```
vcruces: Not hurting anything, but I think this import isn’t used?

tests/service/manual_tasks/test_manual_task_config_service.py
```diff
@@ -14,6 +19,9 @@
 from fides.service.manual_tasks.manual_task_config_service import (
     ManualTaskConfigService,
 )
+from fides.service.manual_tasks.manual_task_instance_service import (
+    ManualTaskInstanceService,
+)
```
vcruces: Same here with imports

src/fides/api/models/manual_tasks/manual_task_instance.py
JadeCara: Oo that is what the TODO is for! Will update that :)

.fides/db_dataset.yml
```diff
@@ -1038,8 +1038,6 @@ dataset:
       data_categories: [system.operations]
     - name: parent_entity_type
       data_categories: [system.operations]
-    - name: due_date
-      data_categories: [system.operations]
```
JadeCara, line 1042: Moved this on to instance - since instances will have due dates not the abstract task.

src/fides/api/models/manual_tasks/manual_task_log.py
JadeCara: You make a good point -  I was thinking of these as audit logs, but I think we are probably covered by the audit logs on the actual privacy request. I am going to update to cascade delete for config and instances as well.

src/fides/service/manual_tasks/manual_task_instance_service.py
JadeCara: It is actually already in use for creating logs!
Anywhere the `with_task_logging` decorator is used it is being returned to the logger for creating the log (ex line 103)

src/fides/api/models/manual_tasks/manual_task_log.py
JadeCara: Updated :)

tests/service/manual_tasks/conftest.py
JadeCara: Cleaned this one and a couple others up :)

src/fides/service/manual_tasks/manual_task_instance_service.py
vcruces: Amazing, I saw the `with_task_logging` code now, thanks!

src/fides/api/models/manual_tasks/manual_task_instance.py
JadeCara: Oh! So basically this is allowing us to separate concerns of the orm - database model still needs the columns defined. The other option would be to explicitly seperate those columns into another class, and use `@declared_attr` to make sure the sqlalchemy picks them up. That being said, the cast is not ideal sooooo - I made some changes 😆 

I updated the mixin and the model to use a Protocol to better define the _what_ using sqlalchemy types so no more cast is needed. This comes with some extra benefits too! 

Type Safety: Protocol ensures correct interface
Reusability: Mixin can be used by any model
Flexibility: Any class with right structure satisfies Protocol
Separation: Interface (Protocol) vs Implementation (Mixin)
Without Protocol: Only explicit inheritance works
With Protocol: Any class with right structure works (structural typing)

PR #6226, opened 2025-06-16
Updates Language-selector to be react-driven, ensuring it closes after selection

eastandwestwind:
Closes https://ethyca.atlassian.net/browse/ENG-721

### Description Of Changes

Updates Language-selector to:

-  automatically close after language selection
-  remove hover-open, and implement click-to-open
-  comply with A11y guidelines, allowing full keyboard control for language selection

### Code Changes

* Update open/close state to be react-driven
* Update CSS accordingly

### Steps to Confirm

1.  Navigate to an experience with multiple languages
2. For both the banner and modal components do the following:
3. Click the language selector, confirm the language modal opens
4. Make a selection, confirm it closes automatically

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:


clients/fides-js/src/components/fides.css
```diff
@@ -1063,16 +1070,6 @@ div#fides-overlay-wrapper .fides-i18n-pseudo-button {
   transform: rotate(180deg);
 }
 
-.fides-i18n-menu:hover .fides-i18n-popover,
-.fides-i18n-menu:focus-within .fides-i18n-popover {
```
eastandwestwind, line 1067: hover / focus is now react-driven. These default CSS properties (when popover is shown) has been moved up to the `.fides-i18n-popover` selector

thabofletcher:
@eastandwestwind - I have tested this on mobile and verified it addresses the issue there - it shows and hides as intended. I'm less familiar with the web, but I did notice that it opens immediately on mouse over, and I'm not sure if that's the same as the behavior was before, but it works fine on web as well. So looks good in testing

Approving the PR (the merge conflict should be easy to resolve) - I did have one question after reviewing

clients/fides-js/src/components/ConsentButtons.tsx
```diff
@@ -45,6 +45,7 @@ export const ConsentButtons = ({
   isGVLLoading,
 }: ConsentButtonProps) => {
   const [isLoadingModal, setIsLoadingModal] = useState<boolean>(false);
+  const [isLanguageMenuOpen, setIsLanguageMenuOpen] = useState(false);
```
thabofletcher: I'm not too familiar with the nuances of TS, is there a significant difference between including `<boolean>` to indicate the type or not?

thabofletcher (approved):

clients/fides-js/src/components/ConsentButtons.tsx
eastandwestwind: nah, I should add, though, to be explicit about types 👍

clients/fides-js/src/components/ConsentButtons.tsx
```diff
@@ -45,6 +45,7 @@ export const ConsentButtons = ({
   isGVLLoading,
 }: ConsentButtonProps) => {
   const [isLoadingModal, setIsLoadingModal] = useState<boolean>(false);
+  const [isLanguageMenuOpen, setIsLanguageMenuOpen] = useState<boolean>(false);
```
gilluminate: Having the state management in this component causes all consent buttons to re-render when the state changes. Is there a reason this isn't handled at the `LanguageSelector` level?

clients/fides-js/src/components/ConsentButtons.tsx
```diff
@@ -118,9 +119,12 @@ export const ConsentButtons = ({
       >
         {includeLanguageSelector && (
           <LanguageSelector
+            id={isInModal ? "fides-i18n-menu-modal" : "fides-i18n-menu-banner"}
```
gilluminate: I don't see these ID's being used anywhere, what purpose do they serve? Are they intended to be for customer styling? If so, I'd recommend using classnames. IMO IDs should never be used for CSS in order to avoid specificity problems (something we get wrong a lot).

clients/fides-js/src/components/LanguageSelector.tsx
```diff
@@ -55,10 +61,28 @@ const LanguageSelector = ({
...
+  const handleToggle = () => {
+    onToggle(!isOpen);
+  };
+
+  const handleMouseEnter = () => {
+    onToggle(true);
+  };
+
+  const handleMouseLeave = () => {
+    onToggle(false);
```
gilluminate: it looks like we're trying to preserve too much of the original CSS hover interaction, while also adding click interaction. IMO we only need one or the other. For example, with this implementation the user hovers which opens the popover then clicks which closes the popover which is unexpected and unnecessarily complex in the UX. Consider removing the hover states altogether.

clients/fides-js/src/components/LanguageSelector.tsx
```diff
@@ -75,7 +99,13 @@ const LanguageSelector = ({
           </MenuItem>
         ))}
       </div>
-      <div className="fides-i18n-pseudo-button">
+      <div
+        className="fides-i18n-pseudo-button"
+        onClick={handleToggle}
+        onKeyPress={handleToggle}
```
gilluminate: `onKeyPress` is deprecated and we should always be using `onKeyDown` instead. This will also fire for ALL keys which is incorrect and should include handing for only enter/space keys. HOWEVER, this pseudo-button was a CSS only solution... I would suggest converting this to a proper `<button>` element which will give you the keyboard interactions for free and all you need is the `onClick` here, no classname and no extra hover handling above.

Additionally, we've lost all accessibility support in the menu itself. With the CSS solution the menu items had focus and the user could use the Tab key to cycle through them. It wasn't the most correct solution either, however, and we should really be supporting up/down arrow key cycling behavior.

clients/fides-js/src/components/fides.css
```diff
@@ -1046,13 +1046,20 @@ div#fides-overlay-wrapper .fides-i18n-pseudo-button {
...
   left: 0;
   bottom: 100%;
-  height: 0px;
-  overflow: hidden;
+  visibility: hidden;
   display: flex;
   flex-direction: column;
   gap: 1px;
   max-height: calc(var(--fides-base-font-size) * 7);
-  transition: height 0.5s;
```
gilluminate, line 1055: not sure why the transition never worked, but if we could get it working that would be a bonus. It's a bit jarring when the menu opens without the animation.

clients/fides-js/src/components/ConsentButtons.tsx
eastandwestwind: This is just used for testing- I'll update to classnames

thabofletcher:
> > Not sure if this is what you are seeing Jason
> 
> @thabofletcher these screenshots appear to be from the TCF Modal, I'm seeing issues on TCF Banner specifically, where the button doesn't open the language selector at all.

That's probably ok from a mobile SDK perspective since we use fides_disable_banner=true to skip the banner in banner_and_modal. But maybe something to discover and address just from an overall correct behavior standpoint?

clients/fides-js/src/components/fides.css
```diff
@@ -558,7 +558,7 @@ div#fides-banner .fides-close-button {
 
 .fides-privacy-policy,
 button.fides-banner-button.fides-banner-button-tertiary,
-div.fides-i18n-pseudo-button {
```
eastandwestwind, line 561: ❗ heads up- I made an intentional decision to change an existing css class.

clients/fides-js/src/components/fides.css
```diff
@@ -1038,39 +1042,38 @@ div.fides-i18n-pseudo-button {
...
-  height: 0px;
-  overflow: hidden;
   display: flex;
   flex-direction: column;
   gap: 1px;
+  background-color: var(--fides-overlay-background-dark-color);
+  border-radius: var(--fides-overlay-component-border-radius);
+  overflow: hidden;
+  max-height: 0;
+  transition: max-height 0.2s ease-in-out;
```
eastandwestwind, line 1060: This fixes a previously broken transition

clients/fides-js/src/components/tcf/fides-tcf.css
```diff
@@ -170,8 +170,10 @@ div#fides-banner-inner-container {
     min-height: unset;
   }
 
-  .fides-tcf-banner-container .fides-i18n-menu {
-    position: unset;
+  .fides-tcf-banner-container
+    .fides-banner-secondary-actions
+    .fides-i18n-menu {
+    position: relative;
```
eastandwestwind, line 174: This was needed because the i18n menu was not showing up at all on mobile TCF

gilluminate (approved):
Fantastic job! Tested on desktop and mobile sized screens and tested for accessibility. All working great. This is a huge improvement for the product! 😄

PR #6227, opened 2025-06-16
ENG-745 Improvements to Redis read-only connection

erosselli:
Closes [ENG-745](https://ethyca.atlassian.net/browse/ENG-745)

### Description Of Changes

Makes the Redis read-only connection more resilient by: 

- Using the writer settings when a read-only setting is not provided. For example, if `read_only_user` is not provided, we will use the `user` value as a fallback. The same will be done for all properties except `read_only_enabled` (which toggles the read-only connection) and `read_only_host`
- Catching all exceptions in the `get_read_only_cache` , and if any exception is raised, we will fall back to the "regular" writer cache 

### Steps to Confirm

Nothing specific since the read replica is used in Fidesplus

1. Tests pass
2. Fides starts up as usual 

### Pre-Merge Checklist

* [ ] Issue requirements met
* [x] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:

[ENG-745]: https://ethyca.atlassian.net/browse/ENG-745?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/fides/config/redis_settings.py
```diff
@@ -77,35 +77,95 @@ class RedisSettings(FidesSettings):
...
+        description="If using TLS encryption rooted with a custom Certificate Authority, set this to the path of the CA certificate. If not provided, the ssl_ca_certs setting will be used.",
     )
 
+    # Computed fields for read-only settings with automatic fallback behavior
+    # If the read-only setting is not provided, we will fall back to the corresponding writer setting.
+    @computed_field  # type: ignore[misc]
+    @property
+    def read_only_port_resolved(self) -> int:
+        """Port for read-only Redis, falls back to writer port if not set."""
+        return self.read_only_port if self.read_only_port is not None else self.port
```
adamsachs: hmm, did you consider using validators to overwrite the existing field values directly? (either field or model validators should allow you to access the value of other fields on the model instance).

i think that approach would be cleaner here, as callers wouldn't have to "know" to leverage the `_resolved` attribute, which seems like a possible developer footgun going forward? perhaps these attributes are not _currently_ access directly (and only exposed via `get_read_only_cache`), but there's no guarantee that will be the case moving forward.

but maybe you already considered this and it's either not possible, or there's a reason you prefer the approach you chose?

to make it clear what i'm suggesting, here's an untested example: 
```suggestion
    @field_validator("read_only_port", mode="before")
    def resolve_read_only_port(cls, v: Optional[int], info: ValidationInfo) -> int:
        """Resolves the read-only port setting by falling back to the `port` value if no `read_only_port` is specified"""
        if v is None:
            return info.data.get("port")
        return v
```

src/fides/api/util/cache.py
```diff
@@ -186,7 +186,9 @@ def _determine_redis_db_index(
         return CONFIG.redis.test_db_index
 
     # 2. Non-test mode
-    return CONFIG.redis.read_only_db_index if read_only else CONFIG.redis.db_index
+    return (
+        CONFIG.redis.read_only_db_index_resolved if read_only else CONFIG.redis.db_index
```
adamsachs: nit: not directly related to this PR but i'd probably take the opportunity to make this part of the docstring a bit more precise/accurate: 

>     2. Non-test mode: return the value already present in `CONFIG.redis.db_index`

should probably read something like: 

>     2. Non-test mode: return the value already present in `CONFIG.redis.db_index` (or CONFIG.redis.read_only_db_index` if accessing the read_only index)

src/fides/config/redis_settings.py
```diff
@@ -130,48 +190,46 @@ def assemble_connection_url(
             # If the whole URL is provided via the config, preference that
             return v
 
+        # For the read-only connection settings, if a setting is not provided, we will
+        # fall back to the corresponding writable setting.
         is_read_only = info.field_name == "read_only_connection_url"
 
+        def get_setting_with_fallback(
```
adamsachs: i _think_ we could also remove this if we used validators?

src/fides/config/redis_settings.py
erosselli: hmm I had tried a version of this but the `Optional[whatever]`  on the fields ended up causing type checker errors in other places that didn't expect possible `None` values.... but maybe with this specific approach I don't need to type the fields with Optional?  I'll try it out

erosselli: the only downside is that we won't know if a particular value was provided or if it's falling back to the writer one, but idk if that would be too useful

adamsachs: oh, yeah i _think_ you can probably get away with making them non-optional with this approach.

in any case, i think i'd probably prefer working around the type-checker errors to having to remember to reference totally separate attributes. it seems less error prone.

tests/ctl/core/config/test_config.py
```diff
@@ -483,82 +483,6 @@ def test_check_required_webserver_config_values_success_from_path() -> None:
     assert check_required_webserver_config_values(config=config) is None
 
 
-class TestBuildingRedisURLs:
```
erosselli, line 486: moved these to the redis-specific file

src/fides/config/redis_settings.py
```diff
@@ -77,35 +77,99 @@ class RedisSettings(FidesSettings):
...
-        default=False,
-        description="Whether the application's connections to the read-only cache should be encrypted using TLS.",
+    read_only_ssl: bool = Field(  # type: ignore[assignment]
+        default=None,
+        description="Whether the application's connections to the read-only cache should be encrypted using TLS. If not provided, the ssl setting will be used.",
     )
-    read_only_ssl_cert_reqs: Optional[str] = Field(
-        default="required",
-        description="If using TLS encryption, set this to 'required' if you wish to enforce the read-only Redis cache to provide a certificate. Note that not all cache providers support this without setting ssl_ca_certs (e.g. AWS Elasticache).",
+    read_only_ssl_cert_reqs: Optional[str] = Field(  # type: ignore[assignment]
```
adamsachs: nit: if this one's actually optional, shouldn't it not need this annotation?

src/fides/config/redis_settings.py
```diff
@@ -77,35 +77,99 @@ class RedisSettings(FidesSettings):
...
+        description="If using TLS encryption rooted with a custom Certificate Authority, set this to the path of the CA certificate. If not provided, the ssl_ca_certs setting will be used.",
     )
 
+    # Field validators for automatic fallback behavior
+    @field_validator("read_only_port", mode="before")
+    @classmethod
+    def resolve_read_only_port(cls, v: Optional[int], info: ValidationInfo) -> int:
+        """Resolves the read-only port setting by falling back to the `port` value if no `read_only_port` is specified."""
+        if v is None:
+            return info.data["port"]
```
adamsachs, line 118: nice! the `[]` rather than `.get` is confident and clear that `port` _must_ be defined at this point, even if it's just with a default 👍

src/fides/config/redis_settings.py
```diff
@@ -77,35 +77,99 @@ class RedisSettings(FidesSettings):
...
+    read_only_port: int = Field(  # type: ignore[assignment]
+        default=None,
+        description="The port at which the read-only Redis cache will be accessible. If not provided, the port setting will be used.",
     )
-    read_only_user: str = Field(
-        default="",
-        description="The user with which to login to the read-only Redis cache.",
+    read_only_user: str = Field(  # type: ignore[assignment]
+        default=None,
+        description="The user with which to login to the read-only Redis cache. If not provided, the user setting will be used.",
```
adamsachs: nit: i'm being a bit of a stickler, but i'd probably make this change with the backticks and similar corresponding changes for other properties just to make the self-documentation here 💯 precise on what you're referring to

```suggestion
        description="The user with which to login to the read-only Redis cache. If not provided, the `user` setting will be used.",
```

adamsachs (approved):
nice job getting this to work with validators, i do think it reads quite a bit more cleanly now and will prevent subtle errors/footguns from arising going forward! and thank you for getting the test coverage in place too, to be extra sure.

only a couple of very nitty-nits that you can feel free to 😒 and ignore

PR #6228, opened 2025-06-16
ENG-602

JadeCara:
Issue [ENG-602](https://ethyca.atlassian.net/browse/ENG-602)

### Description Of Changes

Needed to add some utility functions for better handling logging etc. This is pulled out into its own PR to cut down on the length of #6212 but ultimately is part of getting instances in. 

image

#6205 

#6208 

**You are here **

#6212 

image

### Code Changes

* Added a util file with better logging handling
* Added tests
* Updated the other manual step services with the new configs
* Updated the other service's tests.

### Steps to Confirm

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:

### DIFF
--- a/src/fides/service/manual_tasks/manual_task_config_service.py
+++ b/src/fides/service/manual_tasks/manual_task_config_service.py
@@ -1,171 +1,141 @@
 from typing import Any, Optional
 
 from loguru import logger
-from pydantic import ValidationError
-from sqlalchemy.orm import Session
+from sqlalchemy.orm import Query, Session, selectinload
 
 from fides.api.models.manual_tasks.manual_task import ManualTask
 from fides.api.models.manual_tasks.manual_task_config import (
     ManualTaskConfig,
     ManualTaskConfigField,
 )
-from fides.api.models.manual_tasks.manual_task_log import ManualTaskLog
 from fides.api.schemas.manual_tasks.manual_task_config import (
     ManualTaskConfigResponse,
     ManualTaskConfigurationType,
-    ManualTaskFieldBase,
 )
-from fides.api.schemas.manual_tasks.manual_task_schemas import ManualTaskLogStatus
+from fides.service.manual_tasks.utils import validate_fields, with_task_logging
 
 
 class ManualTaskConfigService:
     def __init__(self, db: Session):
         self.db = db
 
-    # ------- Private Helper Methods -------
+    def _create_log_data(
+        self, task_id: str, config_id: Optional[str], details: dict[str, Any]
+    ) -> dict[str, Any]:
+        """Create standard log data structure."""
+        return {"task_id": task_id, "config_id": config_id, "details": details}
 
-    def _get_fields(self, config: ManualTaskConfig) -> list[dict[str, Any]]:
-        """Get the fields for a config and returns them in the format expected by ManualTaskConfigResponse"""
-        fields = []
-        for field in config.field_definitions:
-            field_data = {
-                "field_key": field.field_key,
-                "field_type": field.field_type,
-                "field_metadata": field.field_metadata,
-            }
-            fields.append(field_data)
-        return fields
+    def _get_base_config_query(self) -> Query:
+        """Get base config query with field definitions loaded."""
+        return self.db.query(ManualTaskConfig).options(
+            selectinload(ManualTaskConfig.field_definitions)
+        )
 
-    def _get_response_data(self, config: ManualTaskConfig) -> dict[str, Any]:
-        """Get the data for a config and returns it in the format expected by ManualTaskConfigResponse"""
-        fields = self._get_fields(config)
-        return {
-            "id": config.id,
-            "task_id": config.task_id,
-            "config_type": config.config_type,
-            "version": config.version,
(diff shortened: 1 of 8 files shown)


src/fides/service/manual_tasks/manual_task_config_service.py
```diff
@@ -1,171 +1,141 @@
...
-            "task_id": config.task_id,
-            "config_type": config.config_type,
-            "version": config.version,
-            "is_current": config.is_current,
-            "fields": fields,
-            "created_at": config.created_at,
-            "updated_at": config.updated_at,
-        }
-
-    def _validate_fields(self, fields: list[dict[str, Any]]) -> None:
```
JadeCara, line 53: Moved this into the `utils.py` file along with the logging utils.

src/fides/service/manual_tasks/manual_task_config_service.py
```diff
@@ -204,216 +173,120 @@ def get_config(
...
-            config_type: The type of config to get
-            field_id: The ID of a field in the config
-            config_id: The ID of the config
-            field_key: The key of a field in the config
-            version: The version number of the config
-
-        Returns:
-            The matching config if found, None otherwise
-        """
+        """Get config by various filters."""
```
JadeCara:
```suggestion
        """Get config by various filters. This is a flexible lookup method that can find configs based on various filters.

        Args:
            task: The task to get the config for
            config_type: The type of config to get
            field_id: The ID of a field in the config
            config_id: The ID of the config
            field_key: The key of a field in the config
            version: The version number of the config
        Returns:
            The matching config if found, None otherwise
            """
```

tests/api/schemas/manual_tasks/test_manual_task_status.py
```diff
@@ -84,10 +88,11 @@ def test_invalid_transition_same_status(self, model, mock_db):
             model.update_status(mock_db, StatusType.pending)
 
     def test_invalid_transition_pending_to_completed(self, model, mock_db):
-        with pytest.raises(
-            StatusTransitionNotAllowed, match="Invalid status transition"
-        ):
-            model.update_status(mock_db, StatusType.completed)
+        model.status = StatusType.pending
```
Vagoasdf, line 91: Shouldn´t this test be renamed, since its no longer an invalid transtion?

tests/service/manual_tasks/test_manual_task_service.py
```diff
@@ -111,128 +116,160 @@ def verify_expected_logs(logs: list[ManualTaskLog], expected_messages: list[str]
...
-    def test_get_task_invalid_id(
-        self,
-        db: Session,
-        manual_task: ManualTask,
-        manual_task_service: ManualTaskService,
-    ):
-        assert manual_task_service.get_task(task_id="invalid-id") is None
-
-    def test_get_task_invalid_parent_entity_type(
+    @pytest.mark.parametrize(
```
Vagoasdf, line 119: +1 For parametrization on these tests 🙌🏽

src/fides/service/manual_tasks/manual_task_service.py
```diff
@@ -43,119 +42,201 @@ def get_task(
...
+        task = self.db.execute(stmt).scalar_one_or_none()
+        if task is None:
+            logger.debug(f"No task found with filters: {filter_desc}")
+            raise ValueError(f"No task found with filters: {filter_desc}")
+        return task
 
-    # User Management
-    def assign_users_to_task(
-        self, db: Session, task: ManualTask, user_ids: list[str]
+    @with_task_logging("Verify user IDs")
```
Vagoasdf: Is this the correct logging for the task? The task logs the non existing users, but this function itself does not verifies, but relies on the incoming array

src/fides/service/manual_tasks/manual_task_service.py
```diff
@@ -43,119 +42,201 @@ def get_task(
...
+    def _non_existent_users(
+        self, non_existent_user_ids: list[str], *, task_id: str
     ) -> None:
-        """Assigns users to this task. We can assign one or more users to a task.
+        """Get users by their IDs.
```
Vagoasdf: The description does not match what the function does

Vagoasdf (approved):

src/fides/service/manual_tasks/manual_task_service.py
JadeCara: Its a good question. The logger util I have either creates a success log or an error log. Since I want any existing users to be added with a success log and any non existant users to create an error log. I split this section out.

If there are no non existant users it will create a success log with the message "Verify User Ids"  <--- I am thinking maybe I update this to "Verify provided User IDs" or similar?
If there are non existant users it will create an error log with the error message.

src/fides/service/manual_tasks/manual_task_service.py
JadeCara: Truth! Fixing :)

PR #6231, opened 2025-06-17
Fix bug in action center pagination

jpople:
Closes ENG-467

### Description Of Changes

Fixes a bug where, after paging forward and then switching tabs, pagination bar would show incorrect range.

### Steps to Confirm

Fix was applied to both the system aggregate and the discovered assets tables, testing is easier on the asset table.
1. Populate some web monitor results (make sure you have at least 26, to have >1 page of results)
2. View the results in the table
3. Go to a page after the first so your pagination bar displays something like "26-50 of X"
4. Click into the "Ignored" tab
5. Pagination bar should show "0-0 of 0"
6. Click back to the "Action required" tab
7. Should be back on the first page

### Pre-Merge Checklist

* [x] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:

### DIFF
--- a/clients/admin-ui/src/pages/data-discovery/action-center/index.tsx
+++ b/clients/admin-ui/src/pages/data-discovery/action-center/index.tsx
@@ -73,7 +73,7 @@ const ActionCenterPage = () => {
 
   useEffect(() => {
     if (data) {
-      setTotalPages(data.total || 1);
+      setTotalPages(data.total ?? 1);
     }
   }, [data, setTotalPages]);
 


clients/admin-ui/src/features/data-discovery-and-detection/action-center/tables/DiscoveredAssetsTable.tsx
```diff
@@ -139,7 +139,7 @@ export const DiscoveredAssetsTable = ({
     if (data) {
       const firstSystemName =
         data.items[0]?.system || systemName || systemId || "";
-      setTotalPages(data.pages || 1);
+      setTotalPages(data.pages ?? 1);
```
jpople, line 142: Quick fix while I'm in here, this was causing it to show "1-0 of 0" instead of "0-0 of 0" for an empty result.

lucanovera: This is a great catch! 💯

lucanovera (approved):
I can confirm the issue is fixed. Thanks for also finding and fixing that bug with the 1-0 page. I noticed that there was one more page that used the || instead of ??, so I pushed that change for consistency. 
Everything working as expected. ship it.

PR #6233, opened 2025-06-17
ENG-744: Adds config for sale_of_data default in Fides Shopify integration

eastandwestwind:
Closes https://ethyca.atlassian.net/browse/ENG-744

### Description Of Changes

Adds config for sale_of_data default in Fides Shopify integration

Follow-up ticket for docs updates- https://ethyca.atlassian.net/browse/ENG-764

### Code Changes

* Update Shopify.ts to handle new config option for `sale_of_data_default`

### Steps to Confirm

1. In `fides-js-components-demo.html`, add the following code after Fides init (L:220):
```
window.Shopify = {
          customerPrivacy: {
            setTrackingConsent: function (consent) {
              // This function is used to test the Shopify integration.
              // It will be called by the Shopify platform script.
              console.log(
                "Shopify customer privacy function called with consent:",
                consent,
              );
            },
          },
        };
        Fides.shopify({ sale_of_data_default: false });
```
2. Nav to `http://localhost:3001/fides-js-components-demo.html`, and open up console
3. Reload page to see console logs, where `sale_of_data` is defaulted to `false`.
Screenshot 2025-06-17 at 5 01 06 PM

4. Now update the code to be `Fides.shopify({ sale_of_data_default: true });`
5. Reload page to see console logs, where `sale_of_data` is defaulted to `true`.
Screenshot 2025-06-17 at 5 03 36 PM

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:

### DIFF
--- a/clients/fides-js/src/integrations/shopify.ts
+++ b/clients/fides-js/src/integrations/shopify.ts
@@ -2,6 +2,10 @@ import { MARKETING_CONSENT_KEYS } from "../lib/consent-constants";
 import { NoticeConsent } from "../lib/consent-types";
 import { processExternalConsentValue } from "../lib/shared-consent-utils";
 
+export interface ShopifyOptions {
+  sale_of_data_default?: boolean;
+}
+
 declare global {
   interface Window {
     Shopify?: {
@@ -48,7 +52,10 @@ type ShopifyConsent = {
   sale_of_data: boolean;
 };
 
-function createShopifyConsent(fidesConsent: NoticeConsent): ShopifyConsent {
+function createShopifyConsent(
+  fidesConsent: NoticeConsent,
+  options?: ShopifyOptions
+): ShopifyConsent {
   const processedConsent = Object.fromEntries(
     Object.entries(fidesConsent).map(([k, v]) => [
       k,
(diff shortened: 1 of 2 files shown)


clients/fides-js/src/integrations/shopify.ts
```diff
@@ -2,6 +2,10 @@ import { MARKETING_CONSENT_KEYS } from "../lib/consent-constants";
 import { NoticeConsent } from "../lib/consent-types";
 import { processExternalConsentValue } from "../lib/shared-consent-utils";
 
+export interface ShopifyOptions {
+  sale_of_data_default?: boolean;
+}
```
gilluminate, line 7: I would suggest making this more scalable and support all/any of the consent values
```suggestion
export interface ShopifyOptions {
  default_consent: Record<string, boolean>
}
```
Then the config setting could look something like 
```
shopify({
  default_consent: {
    sale_of_data: true,
  },
});
```

eastandwestwind: I'm a little torn on this, since this `sale_of_data` [is a special case ](https://shopify.dev/docs/api/customer-privacy#collect-and-register-data-sale-sharing-opt-out) in Shopify, where we _should_ have been treating it as opt-out, but made this mistake in implementation. 

We made the decision to not simply update our code to default this notice to `true` because we currently have customers using it. The other notices should always be opt-in notices

gilluminate: ah got it. Yeah that makes sense in that context. The one-off handling needs a one-off solution. 👍

PR #6238, opened 2025-06-17
get property from fides global if it's not provided

gilluminate:
Closes [<issue>]

### Description Of Changes

### Code Changes

### Steps to Confirm

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:


clients/fides-js/src/lib/preferences.ts
```diff
@@ -152,12 +152,12 @@ export const updateConsentPreferences = async ({
         cookie,
         experience,
         consentMethod,
+        propertyId || window.Fides.config!.propertyId,
```
gilluminate: This method already relies very heavily on the `window.Fides` object, so we can safely assume it's usable as a default.

Kelsey-Ethyca (approved):

clients/fides-js/src/lib/preferences.ts
tvandort: Heh, I had considered doing this but chose not to since I didn't think we'd get it past review. This works for me for the moment though.

Can we use `??` instead of `||` here however?

tvandort: Also since this is tech debt, can we generate an ENG ticket now and comment the ticket # above this line?

gilluminate: Can you tell me why you think `??` is more appropriate?

gilluminate: I consider this entire method to be the tech debt because the props are far too complex for the operations it performs. I want to reverse that, and possibly combine the `updateConsent` below. The thinking is that instead of making a bunch of changes just to get this 1 prop populated, we should rework it all at once. I'll make a ticket for that work and comment at the method level.

tvandort: https://github.com/ethyca/fides/pull/6133#discussion_r2081738815

https://github.com/ethyca/fides/pull/6133#discussion_r2081806660

I'd argue `??` is always more appropriate than `||` when null coalescing as the operator was added to JavaScript in part because the behavior of `||` is often surprising.

In this case we could argue that the type of propertyId will always be string as we _tend_ to prefix `FDS-` and further we _usually_ auto generate these IDs.

That being said if you `POST` to the Property creation endpoint you can actually pass *ANY* ID you want and I don't have full confidence that the value would be treated correct if some user somewhere passed `1` or `0` as their property ID.

Regardless of the last paragraph `||` has surprising behavior and there is little to no advantage to using it over `??` unless we do not pass our JavaScript through some conversion process that makes it compatible with (what are now EXTREMELY) old browsers. Therefore I think we should use the operator that was purpose fit for what we're doing here.

gilluminate: I disagree that `||` is surprising and deprecated. I think each has its own purpose. `||` is a stricter check that I'm intentionally using here to guard against an empty string which `??` would miss. Do you disagree?

tvandort: If we're using `||` _because_ it specifically doesn't include `""` then yes I would be surprised that that choice was the intention after reading the code. If that's the case then I think we should be more expressive / declarative that that's what we are checking for.

```js
const propertyId = ("" ?? window.Fides.config!.propertyId)
// later
{
    propertyId: propertyId.length > 0 ? propertyId : undefined
}
```

gilluminate: I don't even think I'm trying to be clever here, I think that's exactly the right way to use `||`, especially given that the type is already set to `string | undefined` which reduces any other confusion around what might happen there.

tvandort: I'll approve because, ultimately this code hopefully goes away soon. I don't think this reduces confusion. 

`||` gets misused plenty and I don't think _most_ developers are particularly aware that `||` has this behavior. 

We can see that we collective _have_ used it wrong in the privacy center settings code. Therefore I still very strongly think that avoiding `||` as a null coalescing operator is a better path.

If the intent behind using it in a certain location is to specifically allow or not allow `""` I think it would be better to be abundantly clear about it, otherwise I'm going to give every use of `||` in every PR a sideways glance and then have to ponder whether what it lets through is intentional or not.

tvandort (approved):

clients/fides-js/src/lib/preferences.ts
gilluminate: I did make a code change just prior to the approval if you want to give feedback on that. I'll keep that in mind, and maybe we just need an eslint rule that we must intentionally ignore if we're confident about it.

tvandort: I'd love a linting rule like that, would take care of it perfectly.

Rereviewed as well ✅ 

Thank you for bearing with me!

gilluminate: I just checked and we have 225 uses of `||` in our clients directory so adding that rule is going to take a bit of work to clean up 😝

PR #6239, opened 2025-06-17
Use versioneer to set PC/FidesJS version

gilluminate:
### Description Of Changes

Automatically inject `versioneer`-generated version into the Privacy Center and FidesJS build processes. The docker build system replaces a `__RELEASE_VERSION__` placeholder in `next.config.js` with the actual Git-based version during Docker build, ensuring version consistency across the entire application stack. It also writes to the version.json file for FidesJS to pick up and use during the rollup build process.

### Code Changes

* Added Docker build stage to extract version using versioneer
* Modified `built_frontend` stage to replace `__RELEASE_VERSION__` placeholder with actual version as well as write it to the `version.json` file
* Changed hardcoded version in `privacy-center/next.config.js` from "1.2.3" to `__RELEASE_VERSION__` placeholder
* Added `window.Fides.version` property to fides.js bundle for client-side version access
* Added `console.info` to Privacy Center to log the version on every page load.
* Updated smoke test to verify version replacement functionality

### Steps to Confirm

1. Build the Docker image and verify version extraction: `docker build --target version_extractor -t test-version . && docker run --rm test-version cat /version.txt`
2. Build and run the privacy center: `docker build --target prod_pc -t test-pc . && docker run -d -p 3000:3000 test-pc`
4. Load the Privacy Center on http://localhost:3000 and check the console log
5. Load the PC demo page and check the window.Fides.version value

### Pre-Merge Checklist

* [x] Issue requirements met
* [x] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [x] No followup issues
* Database migrations:
  * [x] No migrations
* Documentation:

### DIFF
--- a/clients/build-utils.js
+++ b/clients/build-utils.js
@@ -1,0 +1,49 @@
+const fs = require("fs");
+
+/**
+ * Imports and validates the Fides package version from a JSON file
+ * @param {string} path - Path to the version.json file, defaults to "./version.json"
+ * @returns {string} The package version string, or "unknown" if version cannot be determined
+ * @example
+ * // version.json
+ * {
+ *   "version": "1.2.3"
+ * }
+ *
+ * importFidesPackageVersion() // Returns "1.2.3"
+ * importFidesPackageVersion("invalid/path") // Returns "unknown"
+ */
+const importFidesPackageVersion = (path = "../version.json") => {
+  const errorVersion = "unknown";
+  try {
+    const versionJson = JSON.parse(fs.readFileSync(path, "utf-8"));
+
+    // Validate version file structure and content
+    if (
+      !versionJson?.version ||
+      typeof versionJson.version !== "string" ||
+      versionJson.version.trim() === ""
+    ) {
+      console.warn(
+        `WARNING: Importing Fides package version failed! Invalid version file format or missing version in ${path}`
+      );
+      return errorVersion;
+    }
+
+    return versionJson.version.trim();
+  } catch (error) {
+    console.warn(
+      `WARNING: Importing Fides package version failed! Error when importing version file from ${path}:`,
+      error
+    );
+    return errorVersion;
+  }
+};
+
+// Support both CommonJS and ES module imports
+module.exports = {
+  importFidesPackageVersion,
+};
+
+// Also export as named export for ES modules
+module.exports.importFidesPackageVersion = importFidesPackageVersion;
--- a/clients/fides-js/rollup.config.mjs
+++ b/clients/fides-js/rollup.config.mjs
@@ -12,6 +12,7 @@ import strip from "@rollup/plugin-strip";
 import replace from "@rollup/plugin-replace";
 import fs from "fs";
 import jsxRemoveAttributes from "rollup-plugin-jsx-remove-attributes";
+import { importFidesPackageVersion } from "../build-utils.js";
 
 const NAME = "fides";
 const IS_DEV = process.env.NODE_ENV === "development";
@@ -38,48 +39,6 @@ const preactAliases = {
     { find: "react-dom", replacement: "preact/compat" },
     { find: "react/jsx-runtime", replacement: "preact/jsx-runtime" },
   ],
-};
-
-/**
- * TODO: This is copied from next.config.js in clients/privacy-center. We should refactor this to be shared.
- *
- * Imports and validates the Fides package version from a JSON file
- * @param {string} path - Path to the version.json file, defaults to "../version.json"
- * @returns {string} The package version string, or "unknown" if version cannot be determined
- * @example
- * // version.json
- * {
- *   "version": "1.2.3"
- * }
- *
- * importFidesPackageVersion() // Returns "1.2.3"
- * importFidesPackageVersion("invalid/path") // Returns "unknown"
- */
-const importFidesPackageVersion = (path = "../version.json") => {
-  const errorVersion = "unknown";
-  try {
-    const versionJson = JSON.parse(fs.readFileSync(path, "utf-8"));
-
-    // Validate version file structure and content
-    if (
-      !versionJson?.version ||
-      typeof versionJson.version !== "string" ||
-      versionJson.version.trim() === ""
-    ) {
-      console.warn(
-        `WARNING: Importing Fides package version failed! Invalid version file format or missing version in ${path}`,
-      );
-      return errorVersion;
-    }
-
-    return versionJson.version.trim();
-  } catch (error) {
-    console.warn(
-      `WARNING: Importing Fides package version failed! Error when importing version file from ${path}:`,
-      error,
-    );
-    return errorVersion;
-  }
 };
 
 const fidesScriptPlugins = ({ name, gzipWarnSizeKb, gzipErrorSizeKb }) => [
--- a/clients/privacy-center/next.config.js
+++ b/clients/privacy-center/next.config.js
@@ -1,4 +1,3 @@
-const fs = require("fs");
 const isDebugMode = process.env.FIDES_PRIVACY_CENTER__DEBUG === "true";
 const debugMarker = "=>";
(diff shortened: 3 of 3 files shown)


CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ Changes can also be flagged with a GitHub label for tracking purposes. The URL o
 ### Added
 - Added support for Salesforce custom object monitoring using Helios [#6096](https://github.com/ethyca/fides/pull/6096)
 - Added ManualTaskConfig and ManualTaskConfigField models, foundational for for ManualDSRs [#6208](https://github.com/ethyca/fides/pull/6208) https://github.com/ethyca/fides/labels/db-migration
+- Added `release_version` to the global `window.Fides` object in FidesJS [#6239](https://github.com/ethyca/fides/pull/6239)
```
NevilleS: Nit: we use just `version` on the API (see the `/health` endpoint). I am fundamentally a sucker for consistency so I'd prefer that 👍. I'll repeat this outside the CHANGELOG for clarity too

Dockerfile
```diff
@@ -46,6 +46,27 @@ RUN pip install --no-cache-dir -r optional-requirements.txt
...
+    && rm -rf /var/lib/apt/lists/*
+
+# Copy the source code needed for version extraction
+COPY setup.py setup.cfg versioneer.py ./
+COPY src/ ./src/
+COPY .git/ ./.git/
+
+# Fix git permissions and extract version
+RUN git config --global --add safe.directory /
+RUN python -c "import versioneer; print(versioneer.get_version())" > /version.txt
```
NevilleS: This is clever and I think is pretty much guaranteed to match the version that the `fides` backend build process will arrive at.

That does make me this though... instead of introducing a new image layer, you could do this in the existing `backend` layer and you wouldn't need to install & configure `git`... 

I can't leave a PR suggestion directly in github since it's unedited, but in the Dockerfile#106 you could do this:
```
# This is a required workaround due to: https://github.com/ethyca/fides/issues/2440
RUN git config --global --add safe.directory /fides

# Export the version to a file
RUN python -c "import versioneer; print(versioneer.get_version())" > /fides/version.txt

# Enable detection of running within Docker
ENV RUNNING_IN_DOCKER=true
```

Then later you could copy that `version.txt` from the `backend` base image, instead of your new utility image. Same outcome, but it's guaranteed to use the same dependencies going forward.

Dockerfile
```diff
@@ -123,6 +144,13 @@ COPY clients/ .
 ####################
 FROM frontend AS built_frontend
 
+# Copy the extracted version from the version_extractor stage
+COPY --from=version_extractor /version.txt /tmp/version.txt
```
NevilleS: If you were to follow my suggestion to move version.txt into the `backend` stage, instead this would be:
```suggestion
# Copy the extracted version from the version_extractor stage
COPY --from=backend /fides/version.txt ./version.txt
```

Also, I'd move this into the `frontend` stage above and copy that version.txt file into the `/fides/clients` directory, instead of using `/tmp`. Everything about the docker build is already localized to these working directories, so having a one-off file over in `/tmp` feels like a footgun.

If you do it that way, it'll look something like this:
```
# Build the frontend clients
WORKDIR /fides/clients
COPY clients/package.json clients/package-lock.json ./
COPY clients/fides-js/package.json ./fides-js/package.json
COPY clients/admin-ui/package.json ./admin-ui/package.json
COPY clients/privacy-center/package.json ./privacy-center/package.json
COPY --from=backend /fides/version.txt ./version.txt

RUN npm install
```

Then, each client could reliably know that there was a `clients/version.txt` file they could import into their own builds.

Dockerfile
```diff
@@ -123,6 +144,13 @@ COPY clients/ .
...
+
+# Replace the placeholder version in next.config.js with the actual version
+RUN RELEASE_VERSION=$(cat /tmp/version.txt) && \
+    sed -i "s/__RELEASE_VERSION__/$RELEASE_VERSION/g" /fides/clients/privacy-center/next.config.js
```
NevilleS: nit: This makes sense, but... I dunno, I think it's an important "build" command that I'd expect to happen via `npm run build-privacy-center` instead. So that build command could look for the existence of the `clients/version.txt` file and use it, and you wouldn't have to rely on `sed` as a peculiar side door

Dockerfile
```diff
@@ -123,6 +144,13 @@ COPY clients/ .
...
+
 # Builds and exports admin-ui
 RUN npm run export-admin-ui
 # Builds privacy-center
```
NevilleS, line 134: Awesome - nits aside, this totally gets the version into the privacy center build 👍 

I do think that we should make the `fides-js` build, itself, pick up that version number though. We want both these built outputs to have their own version:
* `privacy-center` image
* `fides.js` built files

clients/privacy-center/pages/api/fides-js.ts
```diff
@@ -398,6 +399,7 @@ export default async function handler(
       : ""
   }
   window.Fides.config = ${fidesConfigJSON};
+  window.Fides.release_version = "${releaseVersion}";
```
NevilleS: nit: I'd prefer it to be `window.Fides.version` 👍 

But also, we shouldn't be injecting that in the privacy-center bundle, it should already be there in the FidesJS file, sorry!

clients/privacy-center/pages/api/fides-js.ts
```diff
@@ -127,6 +127,7 @@ export default async function handler(
 ) {
   const log = createRequestLogger(req);
   const serverSettings = loadServerSettings();
+  const releaseVersion = process.env.version || "unknown"; // set in next.config.js
```
NevilleS: It'd make sense to expose this version number on the Privacy Center somehow too... like, maybe it automatically emits to `console.log` when you load any privacy center page? i.e when I visit https://privacy-fides-nightly-build.fides-staging.ethyca.com/, I can't "get" this version number anywhere

NevilleS (changes requested):
This strategy looks very promising and I'm convinced there's a route here. Lots of suggestions 👍

Dockerfile
NevilleS: Ooh, and one more suggestion...

You should output it as a `version.json` instead with contents `{ "version": "2.63.0" }`. That'll make it easy and clean to import into npm projects without much effort at all.

CHANGELOG.md
gilluminate: The reason I chose `release_version` was to distinguish it from the existing `meta.version` which is the cookie version. It's favoring clarity over consistency. That's my weakly-held strong opinion.

Dockerfile
gilluminate: ~~Last iteration I tried this on, I was setting a ENV variable and reading that in rollup. What do you think about that? I think it's a little cleaner than creating a file.~~
This isn't reliable enough. Sticking with file.

CHANGELOG.md
gilluminate: changed my mind. Went with `version` afterall

Dockerfile
```diff
@@ -135,8 +143,8 @@ FROM node:20-alpine AS prod_pc
 
 WORKDIR /fides/clients
 
-ENV NODE_ENV production
-ENV NEXT_TELEMETRY_DISABLED 1
+ENV NODE_ENV=production
+ENV NEXT_TELEMETRY_DISABLED=1
```
gilluminate, line 145: docker was giving warnings about the deprecated support here, so I cleaned that up

clients/version.json
```diff
@@ -0,0 +1,4 @@
+{
+  "version": "DEV",
+  "comment": "Do not edit this file directly. It will be replaced during the Docker build process by an autogenerated version.json containing the Fides Python package version."
```
gilluminate, line 3: ❤️

NevilleS (approved):
Let's try it out 👍

PR #6257, opened 2025-06-20
Refactor to simplify props passed around

gilluminate:
Closes [ENG-772]

### Description Of Changes

Refactored the fides-js SDK to use React Context for global state management instead of prop drilling. Introduced `FidesGlobalProvider` and `useFidesGlobal` hook to centralize access to the global Fides object. Updated consent handling functions from `updateConsentPreferences` to `updateConsent` with improved parameter structure. Refactored i18n utilities to use pure functions with explicit locale parameters. Centralized `propertyId` management through the `PrivacyExperience` object.

### Code Changes

* Added `FidesGlobalProvider` context and `useFidesGlobal` hook for centralized state management
* Renamed `updateConsentPreferences` to `updateConsent` with new parameter structure using `UpdateConsentOptions`
* Refactored `NoticeOverlay` and `TcfOverlay` components to use context instead of props
* Updated `automaticallyApplyPreferences` to accept `FidesGlobal` object instead of individual parameters
* Refactored i18n utilities (`selectBestNoticeTranslation`, `selectBestExperienceConfigTranslation`) to use explicit locale parameters
* Added `property_id` field to `PrivacyExperience` interface and centralized its usage
* Updated `fidesStringToConsent` to use new `updateConsent` function
* Refactored initialization functions to use props interfaces (`InitializeProps`, `InitOverlayProps`)
* Updated test utilities and mocks to work with new context-based architecture
* Changed `fidesRegionString` return type from `string | null` to `string | undefined` for simplicity and consistency

### Steps to Confirm

1. Verify that consent preferences are properly saved and retrieved using the new context system
2. Test that TCF and non-TCF consent scenarios work correctly with the refactored `updateConsent` function
3. Confirm that i18n translations are properly selected with the new pure function approach
4. Test that GPP string-to-consent functionality works with the updated `updateConsent` integration
5. Verify that all existing functionality continues to work without regression
6. Run the full test suite to ensure all refactored components and utilities pass their tests

### Pre-Merge Checklist

* [x] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [x] No followup issues
* Database migrations:
  * [x] No migrations
* Documentation:

[ENG-772]: https://ethyca.atlassian.net/browse/ENG-772?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


clients/fides-js/src/lib/providers/fides-global-context.tsx
```diff
@@ -0,0 +1,66 @@
...
+    [fidesGlobal, setFidesGlobal],
+  );
+  return (
+    <FidesGlobalContext.Provider value={value}>
+      {children}
+    </FidesGlobalContext.Provider>
+  );
+};
+
+export const useFidesGlobal = () => {
```
eastandwestwind, line 60: Very cool to see this pattern!

eastandwestwind, line 60: This is very cool to see 🌟

clients/fides-js/src/lib/preferences.ts
```diff
@@ -107,9 +107,17 @@ export interface UpdateConsentPreferences {
   servedNoticeHistoryId?: string;
   tcf?: TcfSavePreferences;
   updateCookie?: (oldCookie: FidesCookie) => Promise<FidesCookie>;
-  propertyId?: string;
 }
-// TODO (ENG-772): Refactor to combine updateConsentPreferences and updateConsent an only ask for the global fides object instead of the individual parameters and let the burden of extracting the parameters from the fides object be handled by the method and not the caller.
```
eastandwestwind, line 112: 💯

clients/fides-js/src/components/notices/NoticeOverlay.tsx
```diff
@@ -199,55 +202,52 @@ const NoticeOverlay: FunctionComponent<OverlayProps> = ({
...
-        updateCookie: (oldCookie) =>
-          updateCookieFromNoticePreferences(
-            oldCookie,
-            consentPreferencesToSave,
-          ),
-      }).finally(() => {
+      ).finally(() => {
+        if (window.Fides) {
+          // apply any updates to the fidesGlobal
+          setFidesGlobal(window.Fides as InitializedFidesGlobal);
```
eastandwestwind, line 236: Is there any way we can have this happen as part of `updateConsentPreferences` itself? (https://github.com/ethyca/fides/pull/6257/files#diff-9b3c7a81c5f7d8251528befb21fddb259c1f939368322db0559da8b5aaa7d09aR121)

If I'm understanding correctly, with this refactor, we have to remember to set the FidesGlobal whenever consent preferences change, which feels like something we're going to miss down the line. I'd like to brainstorm if there's any more automated way to manage this?

clients/fides-js/src/components/notices/NoticeOverlay.tsx
```diff
@@ -199,55 +202,52 @@ const NoticeOverlay: FunctionComponent<OverlayProps> = ({
...
+      const noticeConsent: NoticeConsent = {};
+      privacyNoticeItems.forEach((item) => {
+        if (item.notice.consent_mechanism !== ConsentMechanism.NOTICE_ONLY) {
+          noticeConsent[item.notice.notice_key] =
+            enabledPrivacyNoticeKeys.includes(item.notice.notice_key);
+        } else {
+          // always set notice-only notices to true
+          noticeConsent[item.notice.notice_key] = true;
+        }
+      });
```
eastandwestwind, line 221: Seems like we can share this logic with TCFOverlay.tsx and string-to-consent.ts

clients/fides-js/src/lib/gpp/us-notices.ts
```diff
@@ -1,5 +1,8 @@
 /**
  * Helper functions to set the GPP CMP API based on Fides values
+ *
+ * NOTE: Changes to this file should likely also be made in the
+ * GPP encoder API implementation in `fidesplus`.
```
eastandwestwind, line 5: 👍

clients/fides-js/src/lib/initOverlay.ts
```diff
@@ -141,14 +142,8 @@ export const initOverlay = async ({
         // Render the Overlay to the DOM!
         renderOverlay(
           {
-            options,
-            experience,
+            initializedFides,
```
eastandwestwind, line 145: super minimal now, nice!

clients/fides-js/src/lib/initialize.ts
```diff
@@ -191,32 +201,18 @@ export const getInitialFidesFromConsentCookie = ({
...
-   */
-  updateExperience: (
-    props: UpdateExperienceProps,
-  ) => Partial<PrivacyExperience>;
-  overrides?: Partial<FidesOverrides>;
-} & FidesConfig): Promise<Partial<FidesGlobal>> => {
+}: InitializeProps): Promise<Partial<FidesGlobal>> => {
+  const { config } = fides;
+  if (!config) {
+    throw new Error("Fides config should be initialized");
```
eastandwestwind, line 211: When would Fides config not be initialized?

clients/fides-js/src/lib/initialize.ts
```diff
@@ -294,7 +290,15 @@ export const initialize = async ({
       }
 
       /**
-       * Finally, update the "cookie" state to track the user's *current*
+       * If the config has a property_id, we add it to the experience to indicate
+       */
+      if (fides.config?.propertyId) {
+        // eslint-disable-next-line no-param-reassign
+        fides.experience.property_id = fides.config.propertyId;
```
eastandwestwind, line 297: Why would we add this to the experience itself? This feels off, since we don't modify the original `experience` otherwise, right?

clients/fides-js/src/lib/preferences.ts
```diff
@@ -178,9 +185,6 @@ export const updateConsentPreferences = async ({
         cookie,
         experience,
         consentMethod,
-        (propertyId?.length ?? 0) > 0
-          ? propertyId
-          : window.Fides.config!.propertyId,
```
eastandwestwind, line 183: do we not need propertyId anymore when we save preferences?

clients/fides-js/src/lib/preferences.ts
```diff
@@ -219,25 +223,33 @@ export const updateConsentPreferences = async ({
...
-      fides.experience!.non_applicable_privacy_notices?.find((n) => n === key);
+    const nonApplicableNotice = nonApplicablePrivacyNotices.find(
+      (n) => n === key,
+    );
 
-    if (nonApplicableNotice) {
+    if (
+      nonApplicableNotice &&
+      !value &&
+      consentMethod !== ConsentMethod.OT_MIGRATION
```
eastandwestwind, line 246: was this a bug before?

clients/fides-js/src/lib/preferences.ts
```diff
@@ -249,7 +261,8 @@ const validateConsent = (fides: FidesGlobal, consent: NoticeConsent) => {
     if (
       isNoticeOnly &&
       value !== true &&
-      value !== UserConsentPreference.ACKNOWLEDGE
+      value !== UserConsentPreference.ACKNOWLEDGE &&
+      consentMethod !== ConsentMethod.OT_MIGRATION
```
eastandwestwind, line 266: same Q about whether this was a bug? If so, do we anticipate having other consent methods that would bypass this validation check?

clients/fides-js/src/lib/preferences.ts
```diff
@@ -276,44 +289,73 @@ const validateConsent = (fides: FidesGlobal, consent: NoticeConsent) => {
...
-    consent,
-    fidesString,
-    validation = UpdateConsentValidation.THROW,
-  } = options;
-
-  if (!Object.values(UpdateConsentValidation).includes(validation)) {
+  // If neither consent nor fidesString is provided, raise an error
+  if (!consentOptions?.noticeConsent && !consentOptions?.fidesString) {
+    throw new Error("Either consent object or fidesString must be provided");
+  }
```
eastandwestwind: Does this work for TCF, when there are no custom notices? I'm not seeing a `fidesString` passed in from TCF but maybe I'm mistaken

clients/fides-js/src/lib/preferences.ts
```diff
@@ -381,36 +433,50 @@ export const updateConsent = async (
...
-        .privacy_experience_config_history_id;
+      bestExperienceConfigTranslation?.privacy_experience_config_history_id;
   }
 
-  const fidesRegionString =
-    constructFidesRegionString(fides.geolocation) || undefined;
+  const fidesRegionString = constructFidesRegionString(config.geolocation);
+
+  // Generate a new served notice history ID for the served notice
+  // DEFER: This should match with the id used in the notices-served endpoint
```
eastandwestwind, line 469: I thought we had this id already avail?

clients/fides-js/src/lib/tcf/utils.ts
```diff
@@ -484,10 +484,8 @@ export const updateTCFCookie = async (
     tcf_consent: transformTcfPreferencesToCookieKeys(tcf),
     tcf_version_hash: experience.meta?.version_hash,
   };
-  if (consentPreferencesToSave) {
-    result.consent = buildCookieConsentFromConsentPreferences(
```
eastandwestwind, line 488: why don't we need this anymore?

clients/fides-js/src/fides-headless.ts
```diff
@@ -90,6 +90,7 @@ async function init(this: FidesGlobal, providedConfig?: FidesConfig) {
     ...config,
     options: { ...config.options, ...overrides.optionsOverrides },
   };
+  this.config = config;
```
eastandwestwind, line 93: oh man, probably a separate refactor, but I'm sketched out by how we need to manually set items on `this` so that they're properly persisted 😅

clients/fides-js/src/lib/automated-consent.ts
```diff
@@ -1,78 +1,62 @@
...
   const { nc: noticeConsentString } = decodeFidesString(
-    fidesOptions.fidesString || "",
+    options.fidesString || "",
   );
   if (context.globalPrivacyControl) {
     fidesDebugger("GPC is enabled");
   }
   if (noticeConsentString) {
     fidesDebugger("Notice consent string found", noticeConsentString);
+    // fidesString = fidesOptions.fidesString!;
```
eastandwestwind: do we need this line?

clients/fides-js/src/lib/automated-consent.ts
```diff
@@ -84,97 +68,64 @@ export const automaticallyApplyPreferences = async ({
...
-  const privacyExperienceConfigHistoryId =
-    bestTranslation?.privacy_experience_config_history_id;
-
   let gpcApplied = false;
   let noticeConsentApplied = false;
   let migratedConsentApplied = false;
 
-  const consentPreferencesToSave = effectiveExperience.privacy_notices.map(
-    (notice) => {
+  const noticeConsentToSave = experience.privacy_notices.reduce(
```
eastandwestwind: can we add a type to this const?

clients/fides-js/src/lib/automated-consent.ts
```diff
@@ -84,97 +68,64 @@ export const automaticallyApplyPreferences = async ({
     return false;
   }
 
-  /**
-   * Select the "best" translation that should be used for these saved
-   * preferences based on the currently active locale.
-   *
-   * NOTE: This *feels* a bit weird, and would feel cleaner if this was moved
```
eastandwestwind, line 91: 💯 for taking care of this!

clients/fides-js/src/lib/automated-consent.ts
```diff
@@ -84,97 +68,64 @@ export const automaticallyApplyPreferences = async ({
...
+      const isNoticeOnly =
+        notice.consent_mechanism === ConsentMechanism.NOTICE_ONLY;
 
       // First check for migrated consent
       if (hasMigratedConsent && migratedConsent) {
         const preference = migratedConsent[notice.notice_key];
         if (preference !== undefined) {
           migratedConsentApplied = true;
-          const userPreference =
-            typeof preference === "boolean"
```
eastandwestwind, line 118: Wow this file is so much cleaner now! I really like that we let `preferences.ts` handle `transformConsentToFidesUserPreference`, so that much of this duplicate code can be removed from this file 🌟

clients/fides-js/src/components/notices/NoticeOverlay.tsx
gilluminate: The main difficulty here is that there are essentially 2 enviorinments in Fides.js: 1) the initialization, which is pure Typescript, and 2) the UI which is using Preact. This context/hook gets set up as soon as we instantiate Preact and only Preact component files (`.tsx`) can access it. So no, we can't have it happen as part of that utility function and it _only_ has to happen for Preact components. Maybe we could create some shared component that wraps the `updateConsent` util in the Preact side, but that's getting a little abstract and might cause some confusion as well.

clients/fides-js/src/lib/automated-consent.ts
gilluminate: It gets inferred as it gets set and it's a const so it's not really necessary.

clients/fides-js/src/lib/initialize.ts
gilluminate: For all intents and purposes, it will be initialized here, but since the config type says it might be undefined, this is mostly an extra check so that Typescript treats it as defined in all of the uses below this line. Without this then we're left with using `!` to tell TS we're sure it's set, but I try to avoid that since it's prone to bugs.

clients/fides-js/src/lib/initialize.ts
gilluminate: The `experience` we got from the server is for this `propertyId` specifically so in my mind it belongs to the experience. We use the `config.propertyId` to fetch the correct experience. If we only rely on the id on the config object, it's actually prone to bugs since an incorrect or irrelevant property ID on the config _could_ result in an empty `{}` experience. If there's a valid `experience` object, we know the config's property ID is valid, but it no longer belongs to the config, it belongs to the experience. Referencing it from this point forward from the `experience` object is technically more accurate. Hopefully that makes sense.

clients/fides-js/src/lib/preferences.ts
gilluminate: We do, but since it's part of the experience object, and we're already passing in the experience we can just use that. You can see that handling on line 86 above 😄

clients/fides-js/src/lib/preferences.ts
gilluminate: No it wasn't, because we weren't applying this `validateConsent` to the OT migration before, only to the `window.fides.updateConsent()`. Now that we are using it more broadly, I'm handling the nuances of that now.

clients/fides-js/src/lib/preferences.ts
gilluminate: same Answer ;)

clients/fides-js/src/lib/preferences.ts
gilluminate: It's only available as a context on the Preact side. So anything on the non-Preact side gets out of sync with that one. This note was the result of [this slack conversation](https://ethyca.slack.com/archives/CBKK8TS74/p1750712652286549) if you're curious.

clients/fides-js/src/lib/tcf/utils.ts
gilluminate: Because we're now passing in pure `customPurposesConsent` instead of `consentPreferencesToSave` here, which is already formatted correctly for the cookie consent. I suppose you could say that another name for `buildCookieConsentFromConsentPreferences` could have been `buildCustomPurposeConsentFromConsentPreferences`

clients/fides-js/src/lib/preferences.ts
gilluminate: For more context: `updateConsent` here in `preferences.ts` was ONLY used for `window.fides.updateConsent()` prior to this refactor. It was working on that project that sparked this entire concept that if we can be that simple for our external consent management, why are we being so complex for our internal?

clients/fides-js/src/fides-headless.ts
gilluminate: Keep in mind that in this context `this` === `window.Fides` so we're really passing everything to the global.

clients/fides-js/src/lib/preferences.ts
gilluminate: Good catch! It does work right now because we're at least passing in an empty object as `noticeConsent` from the TcfOverlay. However, that's putting the burden on the caller and it's best to also check for `tcf` here. I'll update that. Note: this was originally to alert customers that their configuration was wrong when using `window.Fides.updateConsent()`, which doesn't currently support TCF. For that reason I'm not going to update the message to match. I'll make note of that in the code as well.

clients/fides-js/src/components/notices/NoticeOverlay.tsx
gilluminate: I thought about that, but since TCF uses `notice_key` and the others use `id` it makes it difficult to create a reusable util that isn't more complex to use than repeating these few lines is. Even cursor chokes on it. Maybe we can update TCF to use `id` instead and simplify this. I'll make note of it, but would want to address that separately.

clients/fides-js/src/components/notices/NoticeOverlay.tsx
eastandwestwind: yeah- this makes sense because it's Preact-specific. All good 👍

clients/fides-js/src/components/notices/NoticeOverlay.tsx
eastandwestwind: gotcha- let's leave a todo in the TCF version for this

clients/fides-js/src/lib/initialize.ts
eastandwestwind: gotcha! Feels like something we can eventually get added to the API resp itself to make it easier to work with on FE

clients/fides-js/src/lib/preferences.ts
eastandwestwind: innnteresting, thanks for clarifying!

PR #6287, opened 2025-07-01
Eng 752 web monitor consent

erosselli:
Closes [<issue>]

### Description Of Changes

### Code Changes

### Steps to Confirm

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:


src/fides/api/alembic/migrations/versions/3b0f4905378c_change_asset_with_consent_to_asset_.py
```diff
@@ -0,0 +1,47 @@
...
+        "asset",
+        sa.Column(
+            "consent_status",
+            sa.Enum(
+                name="consentstatus",
+            ),
+            nullable=False,
+        ),
+    )
+    op.drop_column("asset", "with_consent")
```
erosselli: since we're not using the with_consent field yet, I think there's no need for a data migration in either the upgrade or the downgrade operations

src/fides/api/models/connectionconfig.py
```diff
@@ -70,6 +70,8 @@ class ConnectionType(enum.Enum):
     sovrn = "sovrn"
     timescale = "timescale"
     website = "website"
+    # Test connection types, used for testing purposes only
+    test_website = "test_website"
```
erosselli: this is to support a `ConfigurableTestWebsiteMonitor`

src/fides/api/alembic/migrations/versions/d45dec7e541d_add_consent_indexes_to_stagedresource.py
```diff
@@ -0,0 +1,64 @@
...
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    connection = op.get_bind()
+
+    # Check stagedresource table size to decide if we should create indexes immediately
+    stagedresource_count = connection.execute(
+        sa.text("SELECT COUNT(*) FROM stagedresource")
+    ).scalar()
+
+    if stagedresource_count < 1000000:
```
erosselli, line 29: this was a very arbitrary number, I chose it cause it was used for the staged resource ancestor migration but I'm happy to hear why it should be different

adamsachs: hah, the number there was also somewhat arbitrary, but  based on some brief local testing i did, where the index creation was relatively quick (<10s i think) if the # of records was below 1 million.

i honestly don't know how this index creation will perform - i'd imagine it's pretty fast given that the `consent_aggregated` column shouldn't be populated on any records...but i don't know that for sure.

i think it's fine to keep the arbitrary threshold 👍

src/fides/api/models/connectionconfig.py
adamsachs: i'd probably just note this explicitly in the code comments
```suggestion
    # Test connection types, used for testing purposes only
    test_website = "test_website" # used for ConfigurableTesteWebsiteMonitor
```

adamsachs: just checking - does this show up in any UI forms/selectors? it's probably not a huge deal if so, but could be a bit messy...

adamsachs: and another add on - 

this is actually simpler than i was expecting when i took the shortcut of relying on the `ConnctionType.fides` for the `ConfigurableTestMonitor`. it may be worth us just creating a `test_datastore` connection type here and switching the `ConfigurableTestMonitor` to use that.

it's pretty janky that the `ConfigurableTestMonitor` has to use the `fides` connection type...

adamsachs (approved):
things look good on this side! just some minor notes

src/fides/api/models/connectionconfig.py
erosselli: > just checking - does this show up in any UI forms/selectors? it's probably not a huge deal if so, but could be a bit messy

I'll check (I might need to get some help from Jason to know where this is displayed exactly) and make sure it's not.

> this is actually simpler than i was expecting when i took the shortcut of relying on the ConnctionType.fides for the ConfigurableTestMonitor. it may be worth us just creating a test_datastore connection type here and switching the ConfigurableTestMonitor to use that.
 
Is it ok if I ticket it for a follow-up? There's also work in updating fidesplus to use it and also testing, and I think the monitor also has unit tests... it feels like a bit of work and I just want to make sure I have enough time to merge and test this before release cut , while also having other tickets I need to work on  🙂

PR #6292, opened 2025-07-02
ENG-874

tvandort:
Closes [ENG-874]

### Description Of Changes

### Code Changes

### Steps to Confirm

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:

[ENG-874]: https://ethyca.atlassian.net/browse/ENG-874?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/clients/privacy-center/components/HomePage.tsx
+++ b/clients/privacy-center/components/HomePage.tsx
@@ -119,12 +119,12 @@ const HomePage: NextPage = () => {
 
   const content: any = [];
 
-  config.actions.forEach((action) => {
+  config.actions.forEach((action, index) => {
     content.push(
       <PrivacyCard
         key={action.title}
+        index={index}
         title={action.title}
-        policyKey={action.policy_key}
         iconPath={action.icon_path}
         description={action.description}
         onOpen={onPrivacyModalOpen}
--- a/clients/privacy-center/components/PrivacyCard.tsx
+++ b/clients/privacy-center/components/PrivacyCard.tsx
@@ -4,25 +4,25 @@ import Card from "./Card";
 
 type PrivacyCardProps = {
   title: string;
-  policyKey: string;
   iconPath: string;
   description: string;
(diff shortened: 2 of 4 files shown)


clients/privacy-center/components/HomePage.tsx
```diff
@@ -119,12 +119,12 @@ const HomePage: NextPage = () => {
...
   const content: any = [];
 
-  config.actions.forEach((action) => {
+  config.actions.forEach((action, index) => {
     content.push(
       <PrivacyCard
         key={action.title}
+        index={index}
         title={action.title}
-        policyKey={action.policy_key}
```
NevilleS, line 127: OK, these ~8 lines fundamentally show the change here, which is to shift towards using the **action index** instead of the **policy key** as the unique identifier for the "PrivacyCard" action.

I discussed this live with @tvandort and @Kelsey-Ethyca and this makes sense to me. The Fides API doesn't expect unique policy keys per action, and fundamentally is designed to allow a form UI like the privacy center to submit arbitrary fields for a given request.

So it's a long-standing bug with the privacy center that we were assuming `policy_key` would be a unique ID. It's not, we should support multiple actions to use the same policy key, so this makes sense.

tvandort:
@NevilleS @lucanovera One concern I have is this:
image

I have no idea why the type system is listing this type twice. I usually avoid casting where possible but this was the quickest way I could work around this.

lucanovera (approved):
Thanks @tvandort! I think this is the correct fix for the issue, allowing now to use the same policy key in multiple actions. I've reproduced the issue and confirmed the fix to be working. Approved.!

PR #6296, opened 2025-07-02
Adding user ID to ManualTaskLog

galvana:
### Code Changes

* Adds `user_id` to the `manual_task_log` table
* Captures the user ID in a `ContextVar` upon authentication (`extract_token_and_load_client`)

### Steps to Confirm

1.  This is to support https://github.com/ethyca/fidesplus/pull/2319

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:


.fides/db_dataset.yml
```diff
@@ -1072,6 +1072,8 @@ dataset:
       data_categories: [system.operations]
     - name: details
       data_categories: [system.operations]
+    - name: user_id
+      data_categories: [user.unique_id]
```
erosselli, line 1076: is this data category intentional or is it AI? literally all the other fields here just say `system.operations` . tbh I don't even know what this yml is used for

src/fides/api/models/manual_task.py
```diff
@@ -921,10 +928,12 @@ def create_log(
             message: Optional message describing the event
             details: Optional additional details about the event
         """
+
         data = {
             "task_id": task_id,
             "config_id": config_id,
             "instance_id": instance_id,
+            "user_id": user_id or get_request_context().user_id,
```
erosselli: will the request context always have a `user_id`? what about for system-initiated events, does it have a `user_id` property but set to None?

src/fides/api/models/manual_task.py
```diff
@@ -954,12 +964,14 @@ def create_error_log(
...
             The created error log entry
         """
+
         return cls.create_log(
             db=db,
             status=ManualTaskLogStatus.error,
             task_id=task_id,
             config_id=config_id,
             instance_id=instance_id,
+            user_id=user_id or get_request_context().user_id,
```
erosselli: I think it might be nice to define a `get_user_id` method in the class and encapsulate this logic

src/fides/api/request_context.py
```diff
@@ -0,0 +1,66 @@
...
+helpers -> decorators) without having to thread additional parameters through
+all call-sites.
+
+Currently we only capture the authenticated `user_id` but additional fields
+(e.g. correlation_id, locale, feature_flags) can be added in the future by
+expanding the `RequestContext` dataclass.
+
+A `contextvars.ContextVar` is used instead of a module-level global to ensure
+that values are local to the current execution context (async task, thread or
+Celery worker) and therefore safe for concurrent workloads.
```
erosselli, line 13: I haven't worked with contextvars before so idk the specifics of them, but what exactly would be th execution context ? Is this still safe for a single fides webserver that's serving concurrent requests?

.fides/db_dataset.yml
galvana: Yes, this was intentional! This is meant to be the dataset for the Fides system, but I don't know if anyone really uses it. In this case, I set it to `user.unique_id` since it's going to contain `FidesUser.id`s

src/fides/api/models/manual_task.py
galvana: System initiated will be set to `None`

src/fides/api/request_context.py
galvana: Each HTTP request will get its own context. When the request ends the context is discarded. I also found this conversation around this approach, specifically for FastAPI https://github.com/fastapi/fastapi/discussions/8628#discussioncomment-5153107

erosselli: thanks for the clarification!

PR #6301, opened 2025-07-03
ENG-781 Implement require DSR erasure finalization

eastandwestwind:
Closes https://ethyca.atlassian.net/browse/ENG-781 and https://ethyca.atlassian.net/browse/ENG-794

### Description Of Changes

This change introduces a new, optional step in the privacy request lifecycle: manual finalization. When enabled, erasure requests will be paused in a `requires_manual_finalization` state after all automated processing is complete. This gives customers a chance to perform any manual steps or verification before the request is marked as `complete`.

A new API endpoint and UI controls have been added to allow authorized users to trigger the finalization.

This feature is disabled by default and can be enabled by setting `erasure_request_finalization_required` to `True` in the configuration.

### Code Changes

*   **Backend:**
    *   Added `requires_manual_finalization` to `PrivacyRequestStatus` enum.
    *   Added `finalized_at` and `finalized_by` fields to the `PrivacyRequest` model.
    *   Added a new API endpoint `POST /privacy-request/{privacy_request_id}/finalize` to finalize requests.
    *   Updated the request runner service to transition erasure requests to the `requires_manual_finalization` state when the feature is enabled.
    *   Added `erasure_request_finalization_required` setting to `fides.toml`.
    *   Added tests for the new finalization flow.
*   **Frontend:**
    *   Added a "Requires Finalization" status to the privacy request UI.
    *   Added a "Finalize" button to the privacy request details page for requests in the `requires_manual_finalization` state.
    *   `clients/admin-ui/src/features/privacy-requests/hooks/useFinalizePrivacyRequest.ts`: New hook to handle request finalization.
    *   Updated various components to support the new status and action.

### Steps to Confirm

1. I've tested this pinning the latest fides commit to fidesplus. 
2. In Fidesplus local, enable the feature by adding `FIDES__EXECUTION__ERASURE_REQUEST_FINALIZATION_REQUIRED=true` to your env.
3.  Create an erasure privacy request.
4.  Process the request. Observe that it enters the "Requires Finalization" state instead of "Complete".
5.  Navigate to the request's detail page, and click "finalize" underneath "Actions" dropdown. Alternatively, on the table view, click the checkmark directly on the column actions.
6.  Confirm that the request status changes to "Complete".
7.  Confirm a new activity log line item is added for the "finalize" step
8. Set `FIDES__EXECUTION__ERASURE_REQUEST_FINALIZATION_REQUIRED=false` to confirm no regressions in normal erasure DSR

Finalize action on details page:
Screenshot 2025-07-10 at 3 13 23 PM

When messaging is configured:
Screenshot 2025-07-10 at 3 10 32 PM

When messaging is not configured:
Screenshot 2025-07-09 at 7 51 16 PM

Activity log after finalization:
Screenshot 2025-07-14 at 6 14 04 PM

### Pre-Merge Checklist

* [x] Issue requirements met
* [x] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [x] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:


clients/admin-ui/src/features/privacy-requests/FinalizePrivacyRequest.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+  const handleFinalizeClick = () => {
+    handleFinalize(id);
+  };
+
+  return (
+    <Flex vertical>
+      <Text>
+        Finalize the DSR once you&#39;ve reviewed and have completed any manual
+        erasures outside of Fides
```
erosselli: nit: the "you've reviewed and have completed" phrasing threw me off a bit, maybe just "you have reviewed and completed" ?

src/fides/api/alembic/migrations/versions/a7065df4dcf1_add_finalized_fields_to_privacy_request.py
```diff
@@ -0,0 +1,65 @@
...
+        "privacyrequest_fidesuser_id_fkey",
+        "privacyrequest",
+        "fidesuser",
+        ["finalized_by"],
+        ["id"],
+        ondelete="SET NULL",
+    )
+    op.execute(
+        "alter type privacyrequeststatus add value 'requires_manual_finalization'"
+    )
```
erosselli, line 38: so in other migrations we don't do `alter type` , we rename the existing enum, create a new one with the new value, change column type to new enum, and then drop the old one (just noticed you do this in the downgrade code too) ..... idk if this is actually needed or not, we've had a couple different slack conversations about this but wanted to call it out . though if it works , I don't see why we wouldn't do it this way, much easier

src/fides/api/alembic/migrations/versions/a7065df4dcf1_add_finalized_fields_to_privacy_request.py
```diff
@@ -0,0 +1,65 @@
...
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_constraint(
+        "privacyrequest_fidesuser_id_fkey", "privacyrequest", type_="foreignkey"
+    )
+    op.drop_column("privacyrequest", "finalized_by")
```
erosselli, line 47: do you need to specifically drop the constraint , or is dropping the column enough?

src/fides/api/api/v1/endpoints/privacy_request_endpoints.py
```diff
@@ -1862,6 +1863,47 @@ def resume_privacy_request_from_requires_input(
...
+    Finalizes a privacy request, moving it from the 'requires_finalization' state to 'complete'.
+    This is done by re-queueing the request, which will then hit the finalization logic in the
+    request runner service.
+    """
+    privacy_request = get_privacy_request_or_error(db, privacy_request_id)
+
+    if privacy_request.status != PrivacyRequestStatus.requires_manual_finalization:
+        raise HTTPException(
+            status_code=HTTP_400_BAD_REQUEST,
+            detail=f"Privacy request with id '{privacy_request_id}' is not in the requires_manual_finalization state.",
```
erosselli:
```suggestion
            detail=f"Cannot manually finalize privacy request '{privacy_request_id}': status is {privacy_request.status}, not requires_manual_finalization."
```

I think it can be helpful to return current status as part of the error message  -- also I made up privacy_request.status, can't remember if that's the correct name for the field

src/fides/api/api/v1/endpoints/privacy_request_endpoints.py
```diff
@@ -1862,6 +1863,47 @@ def resume_privacy_request_from_requires_input(
...
+    db: Session = Depends(deps.get_db),
+    client: ClientDetail = Security(
+        verify_oauth_client,
+        scopes=[PRIVACY_REQUEST_REVIEW],
+    ),
+) -> PrivacyRequestResponse:
+    """
+    Finalizes a privacy request, moving it from the 'requires_finalization' state to 'complete'.
+    This is done by re-queueing the request, which will then hit the finalization logic in the
+    request runner service.
```
erosselli: can you explain what the finalization logic will do?

src/fides/api/api/v1/endpoints/privacy_request_endpoints.py
```diff
@@ -1862,6 +1863,47 @@ def resume_privacy_request_from_requires_input(
...
+
+    if privacy_request.status != PrivacyRequestStatus.requires_manual_finalization:
+        raise HTTPException(
+            status_code=HTTP_400_BAD_REQUEST,
+            detail=f"Privacy request with id '{privacy_request_id}' is not in the requires_manual_finalization state.",
+        )
+
+    # Set finalized_by and finalized_at here, so the request runner service knows not to
+    # put the request back into the requires_finalization state.
+    privacy_request.finalized_at = datetime.utcnow()
```
erosselli: `datetime.utcnow()` is deprecated so I'm hesitant to add new code that uses it, can we look for a non-deprecated alternative?

src/fides/api/service/privacy_request/request_runner_service.py
```diff
@@ -478,111 +478,124 @@ def run_privacy_request(
...
-                # Access CHECKPOINT
+                # Only run the traversal if the request is not being finalized
                 if (
-                    policy.get_rules_for_action(action_type=ActionType.access)
-                    or policy.get_rules_for_action(action_type=ActionType.erasure)
-                ) and can_run_checkpoint(
-                    request_checkpoint=CurrentStep.access, from_checkpoint=resume_step
+                    privacy_request.status
+                    != PrivacyRequestStatus.requires_manual_finalization
                 ):
```
erosselli: instead of nesting all the code under yet another if, can we just invert the logic and check that if the request requires manual finalization we return / do whatever we have to do in that case?  I know the code for this function is already messy but trying to think of ways to stop it from getting worse 🙂

src/fides/api/service/privacy_request/request_runner_service.py
```diff
@@ -652,58 +665,107 @@ def run_privacy_request(
...
-                        policy,
-                        access_result_urls,
-                        identity_data,
-                        privacy_request.property_id,
+            # Manual Finalize CHECKPOINT
+            if can_run_checkpoint(
+                request_checkpoint=CurrentStep.manual_finalization,
+                from_checkpoint=resume_step,
+            ):
+                # Finally, mark the request as complete and send completion notifications.
```
erosselli: this comment feels out of place here since we're not always marking as complete right ?

src/fides/api/service/privacy_request/request_runner_service.py
```diff
@@ -652,58 +665,107 @@ def run_privacy_request(
...
-                        session,
-                        policy,
-                        access_result_urls,
-                        identity_data,
-                        privacy_request.property_id,
+            # Manual Finalize CHECKPOINT
+            if can_run_checkpoint(
+                request_checkpoint=CurrentStep.manual_finalization,
+                from_checkpoint=resume_step,
+            ):
```
erosselli: I find this a bit confusing since manual finalization can be disabled, how does that factor in? Maybe this CurrentStep should be called `finalization` or something else, cause this seems to imply that it will always be manual?

tests/fixtures/application_fixtures.py
```diff
@@ -4034,3 +4034,21 @@ def comment(db, comment_data):
     comment = Comment.create(db, data=comment_data)
     yield comment
     comment.delete(db)
+
+
+@pytest.fixture(scope="function")
+def erasure_request_finalization_required_true(db):
```
erosselli: nit: usually we call this fixtures `enable_whatever_setting` and `disable_whatever_setting` , can we rename just for consistency / so they're easier to find for others? 
```suggestion
def enable_erasure_request_finalization_required(db):
```

tests/fixtures/application_fixtures.py
```diff
@@ -4034,3 +4034,21 @@ def comment(db, comment_data):
...
+    """Enable erasure finalization via config"""
+    original_value = CONFIG.execution.erasure_request_finalization_required
+    CONFIG.execution.erasure_request_finalization_required = True
+    yield
+    CONFIG.execution.erasure_request_finalization_required = original_value
+
+
+@pytest.fixture(scope="function")
+def erasure_request_finalization_required_false(db):
```
erosselli:
```suggestion
def disable_erasure_request_finalization_required(db):
```

tests/ops/api/v1/endpoints/test_privacy_request_endpoints.py
```diff
@@ -8742,3 +8757,74 @@ def test_send_batch_email_integrations_queues_task(
             == "Email batch job started. This may take a few minutes to complete."
         )
         mock_send_email_batch.assert_called_once()
+
+
+class TestFinalizePrivacyRequest:
```
erosselli: ask: can we move this to a separate test file under a folder tests/ops/api/v1/endpoints/privacy_requests/` folder (and add this file `test_privacy_request_endpoints.py` there too) ? our tests files with 5k+ lines are getting a bit out of hand and also make it hard for cursor to be able to make changes correctly and efficiently since it can't reliably get all the file into its context..

tests/ops/api/v1/endpoints/test_privacy_request_endpoints.py
```diff
@@ -8742,3 +8757,74 @@ def test_send_batch_email_integrations_queues_task(
...
+        payload = {
+            JWE_PAYLOAD_ROLES: user.client.roles,
+            JWE_PAYLOAD_CLIENT_ID: user.client.id,
+            JWE_ISSUED_AT: datetime.now().isoformat(),
+        }
+        # User already has Privacy Request Review scope
+        auth_header = {
+            "Authorization": "Bearer "
+            + generate_jwe(json.dumps(payload), CONFIG.security.app_encryption_key)
+        }
```
erosselli: we have a generate_auth_header fixture for this, can we reuse it instead of defining all this manually?

tests/ops/api/v1/endpoints/test_privacy_request_endpoints.py
```diff
@@ -8742,3 +8757,74 @@ def test_send_batch_email_integrations_queues_task(
...
+        response = api_client.post(url, headers=auth_header)
+        assert response.status_code == 200
+
+        privacy_request_requires_manual_finalization.refresh_from_db(db=db)
+        assert (
+            privacy_request_requires_manual_finalization.status
+            == PrivacyRequestStatus.requires_manual_finalization
+        )
```
erosselli: this is a bit confusing as an assertion cause it seems that manual finalization didn;t work correctly. I know that's cause we're mocking out the queue privacy request but maybe we shouldn't mock it? or if we do mock it, I think we need to make this clear on the test docstring and possibly remove this assertion since it's not really testing anything

erosselli: also if we keep the mock can we add a more thorough E2E test where we call the endpoint without mocking anything?

src/fides/api/alembic/migrations/versions/a7065df4dcf1_add_finalized_fields_to_privacy_request.py
eastandwestwind: Interesting- I do see a couple migrations where we've altered the type to add a value, e.g. `d65e7e921814_add_privacy_request_status.py` or `392992c7733a_add_ga_client_id_as_a_provided_identity_type.py`. But in any case, I've tested both upgrade and downgrade, and it seemed to work fine.

This 2018 postgresql commit seems to have relaxed restrictions on altering type in transaction- https://git.postgresql.org/gitweb/?p=postgresql.git;a=commit;h=212fab9926b2f0f04b0187568e7124b70e8deee5

src/fides/api/alembic/migrations/versions/a7065df4dcf1_add_finalized_fields_to_privacy_request.py
eastandwestwind: I don't think it's necessary but we have used this pattern in other migrations... hmm. I'm open either way

src/fides/api/alembic/migrations/versions/a7065df4dcf1_add_finalized_fields_to_privacy_request.py
erosselli: ah thanks for looking this up, it's good to know!

src/fides/api/service/privacy_request/request_runner_service.py
eastandwestwind: good call 💯

src/fides/api/service/privacy_request/request_runner_service.py
eastandwestwind: That's a good point, I'll rename to `finalization` 👍

tests/ops/api/v1/endpoints/test_privacy_request_endpoints.py
eastandwestwind: Ah now I see we have `generate_auth_header_for_user`, which I'll need to define this specific user 👍

clients/admin-ui/src/features/privacy-requests/FinalizePrivacyRequest.tsx
```diff
@@ -0,0 +1,29 @@
+import { AntButton as Button, AntFlex as Flex, AntText as Text } from "fidesui";
+
+import { useFinalizePrivacyRequest } from "./hooks/useFinalizePrivacyRequest";
+
+type FinalizePrivacyRequestProps = {
+  id: string;
+};
+
+const FinalizePrivacyRequest = ({ id }: FinalizePrivacyRequestProps) => {
```
gilluminate: it doesn't look like this new component is being utilized anywhere. Am I missing something?

clients/admin-ui/src/features/common/api.slice.ts
```diff
@@ -63,6 +63,7 @@ export const baseApi = createApi({
     "Privacy Notice Translations",
     "Privacy Request Attachments",
     "Privacy Request Comments",
+    "Privacy Requests",
```
gilluminate: this new tag type doesn't seem to be used anywhere, is it still necessary?

clients/admin-ui/src/features/common/modals/ConfirmationModal.tsx
```diff
@@ -49,7 +49,7 @@ const ConfirmationModal = ({
     isCentered={isCentered}
   >
     <ModalOverlay />
-    <ModalContent textAlign="center" p={6} data-testid={testId}>
+    <ModalContent textAlign="left" p={6} data-testid={testId}>
```
gilluminate, line 52: ❤️

clients/admin-ui/src/features/privacy-requests/RequestTableActions.tsx
```diff
@@ -126,6 +163,20 @@ export const RequestTableActions = ({
...
+        isOpen={finalizeModal.isOpen}
+        onClose={finalizeModal.onClose}
+        onConfirm={handleFinalizeRequest}
+        title="Finalize privacy request"
+        message={
+          <Text>
+            {activeMessagingProvider?.service_type &&
+            sendRequestCompletionNotification
+              ? 'You are about to finalize this privacy request. The status will be updated to "Complete" and the requesting user will be notified. Would you like to continue?'
+              : 'You are about to finalize this privacy request. The status will be updated to "Complete". Would you like to continue?'}
```
gilluminate: This could be simplified a bit
```suggestion
            `You are about to finalize this privacy request. The status will be updated to "Complete"${sendRequestCompletionNotification ? " and the requesting user will be notified" : ""}. Would you like to continue?`
```

tests/ops/api/v1/endpoints/test_privacy_request_endpoints.py
eastandwestwind: Done! I've removed the mock entirely so this is an e2e test

clients/admin-ui/src/features/privacy-requests/FinalizePrivacyRequest.tsx
eastandwestwind: you're right! This was leftover from previous implementation

src/fides/api/service/privacy_request/request_runner_service.py
```diff
@@ -652,58 +662,107 @@ def run_privacy_request(
...
+                        )
+                        privacy_request.status = (
+                            PrivacyRequestStatus.requires_manual_finalization
+                        )
+                        privacy_request.cache_failed_checkpoint_details(
+                            CurrentStep.finalization
+                        )
+
+                    else:
+                        # Finally, mark the request as complete
```
erosselli: nit: can we `return` at the end of the if block above to avoid this else statement? I think that will read easier

erosselli (approved):
Approved with a small comment . Also let's make sure codecov is passing :)

clients/admin-ui/src/features/common/api.slice.ts
eastandwestwind: nope!

src/fides/api/service/privacy_request/request_runner_service.py
eastandwestwind: done!

erosselli (changes requested):
When running through the testing steps, I got a 500 error on the `/finalize` endpoint when trying to finalize a request as a root user, caused by an `IntegrityError`: `sqlalchemy.exc.IntegrityError: (psycopg2.errors.ForeignKeyViolation) insert or update on table "privacyrequest" violates foreign key constraint "privacyrequest_fidesuser_id_fkey"` 

It seems like we need to handle the root user case, I don't know what we do for the other user FKs like reviewed_by but I assume we have some special handling somewhere, e.g leave it empty if it's the root user? 

After making the change we also need to add a test that checks the root user can successfully call the /finalize endpoint.

PR #6312, opened 2025-07-07
FidesJS: served_notice_history_id consistency

gilluminate:
Closes [ENG-814]

### Description Of Changes

Refactored session management in FidesJS to use a centralized SessionManager singleton that ensures consistent `served_notice_history_id` values across all consent flows within a single session. This improves analytics correlation between the notices-served and privacy-preferences endpoints by using the same session-level identifier for both automated consent (GPC) and manual consent interactions.

### Code Changes

* Created new `SessionManager` class with singleton pattern to manage session-level UUIDs.
* Updated `useNoticesServed` hook to use centralized session manager instead of generating new UUIDs per event.
* Modified `updateConsent` function to use session manager for consistent `served_notice_history_id`.
* Removed `servedNoticeHistoryId` parameter from `updateConsent` function signature and related component prop drilling.
* Updated `NoticeOverlay` and `TcfOverlay` components to remove unused `servedNoticeHistoryId` dependencies.
* Added comprehensive test suite for `SessionManager` covering singleton behavior, session lifecycle, and UUID consistency.
* Added Cypress E2E test to verify consistent `served_notice_history_id` between automated and manual consent flows.

### Steps to Confirm

1. Enable GPC in your browser and load an experience which will trigger the `privacy-preferences` endpoint.
2. Open the modal and Opt in/out.
3. Test that `notices-served` and `privacy-preferences` API calls use the same `served_notice_history_id` in their payload
4. Verify that new sessions (page reloads) generate different `served_notice_history_id` values as expected

### Pre-Merge Checklist

* [x] Issue requirements met
* [x] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [x] No followup issues
* Database migrations:
  * [x] No migrations
* Documentation:

[ENG-814]: https://ethyca.atlassian.net/browse/ENG-814?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/clients/fides-js/src/lib/preferences.ts
+++ b/clients/fides-js/src/lib/preferences.ts
@@ -33,6 +33,7 @@ import {
   FidesEventExtraDetails,
   FidesEventOrigin,
 } from "./events";
+import { fidesLifecycleManager } from "./fides-lifecycle-manager";
 import { decodeFidesString } from "./fides-string";
 import {
   DEFAULT_LOCALE,
@@ -40,7 +41,6 @@ import {
   selectBestExperienceConfigTranslation,
   selectBestNoticeTranslation,
 } from "./i18n";
-import { fidesLifecycleManager } from "./fides-lifecycle-manager";
 import { transformConsentToFidesUserPreference } from "./shared-consent-utils";
 import { TcfSavePreferences } from "./tcf/types";
 


clients/fides-js/src/lib/preferences.ts
```diff
@@ -465,12 +463,8 @@ export const updateConsent = async (
...
   const fidesRegionString = constructFidesRegionString(config.geolocation);
 
-  // Generate a new served notice history ID for the served notice
-  // DEFER: This should match with the id used in the notices-served endpoint
-  if (!servedNoticeHistoryId) {
-    // eslint-disable-next-line no-param-reassign
-    servedNoticeHistoryId = uuidv4();
-  }
+  // Get the session-level served notice history ID for consistency
+  const servedNoticeHistoryId = sessionManager.getServedNoticeHistoryId();
```
eastandwestwind: 🎉

clients/fides-js/src/lib/session-manager.ts
```diff
@@ -0,0 +1,38 @@
+import { v4 as uuidv4 } from "uuid";
+
+class SessionManager {
+  private servedNoticeHistoryId: string | null = null;
+
+  /**
+   * Get the session-level served_notice_history_id.
+   * Creates a new UUID on first access and reuses it for the entire session.
```
eastandwestwind: Nice! I imagine we could use a similar approach in the future for generating the `fides_user_device_id`

eastandwestwind (approved):
Nicely done @gilluminate 💯

clients/fides-js/src/lib/session-manager.ts
gilluminate: As I'm thinking about this comment, I'm now thinking that "session" is probably the wrong wording to use here. This ID will change if the user reloads the page or changes pages, it's really the FidesJS session, not the user's browser session. 🤔

eastandwestwind: ahh gotcha. OK so it's more the FidesJS session, which is just a page load... 

Some naming ideas I got from Claude:

> Page instance - referring to the specific loaded instance of a page
Page context - the runtime environment and state of that particular page
Page lifecycle - encompassing the entire duration from page load to unload
Document session - since each page corresponds to a document in the DOM

So e.g. we could use page-context-manager.ts or something like that

gilluminate: I like lifecycle, but not as big of a fan of "page." I'd rather name it for the script since SPA apps will affect this lifecycle as well. I think Fides Lifecycle is likely the clearest option.

PR #6319, opened 2025-07-09
ENG-898

tvandort:
Closes [ENG-898]

### Description Of Changes

Retry for pulling `custom-fides.css` on failures. 
Do not disable `custom-fides.css` polling on failures that are not 404.

### Code Changes

### Steps to Confirm

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:


clients/privacy-center/pages/api/fides-js.ts
```diff
@@ -460,37 +463,63 @@ async function fetchCustomFidesCss(
...
+            autoRefresh = false;
+            return null;
+          }
+
+          const data = await assetResponse.text();
+          if (!data) {
+            throw new Error("No data returned by the server");
+          }
+          if (!assetResponse.ok) {
+            log.error(
```
thabofletcher: Nit - I'm not sure an error on each attempt is necessary as the final failure after x retry attempts appears to already be thrown - this might be better as a debug log

clients/privacy-center/pages/api/fides-js.ts
```diff
@@ -460,37 +463,63 @@ async function fetchCustomFidesCss(
...
+          }
+
+          return data;
+        },
+        {
+          retries: CUSTOM_CSS_MAX_RETRIES,
+          factor: CUSTOM_CSS_BACKOFF_FACTOR,
+          minTimeout: CUSTOM_CSS_RETRY_MIN_TIMEOUT_MS,
+          onFailedAttempt: (error) => {
+            log.debug(
```
thabofletcher, line 502: Related to above nit - I wasn't able to see my debug logs but if this happens on each attempt its probably sufficient. If it only happens at the end (like the built in pRetry error I am seeing), then it would always show 0 for attempts remaining

This might all be my setup not quite right as well

clients/privacy-center/pages/api/fides-js.ts
```diff
@@ -460,37 +463,63 @@ async function fetchCustomFidesCss(
...
-          response.statusText,
-          data,
-        );
-        throw new Error(`HTTP error occurred. Status: ${response.status}`);
-      }
 
-      if (!data) {
-        throw new Error("No data returned by the server");
+      if (!css) {
+        return null;
```
thabofletcher, line 512: Final nit - Might consider moving the log from line 472 down here or adding a comment - since this logic here is dependent on the above code returning null, it might be nice to call out this dependency as why we are skipping caching. Just tripped me up at first reading it

thabofletcher:
Looks good! 🚀

clients/privacy-center/pages/api/fides-js.ts
tvandort: Discussed in Slack & after configuring debug logging this works as expected.

clients/privacy-center/pages/api/fides-js.ts
tvandort: Great point, updated.

clients/privacy-center/pages/api/fides-js.ts
tvandort: I'm going to add a slightly different log here since I intended the other log that you're mentioning to signal that polling was disabled as a side effect, where as this here is the return result. I agree it _looks_ a bit weird though so I'm hoping the new log makes this more comprehensible.

PR #6330, opened 2025-07-11
Upgrade RollupJS to latest

gilluminate:
Upgrades RollupJS to latest version.

* I've read through their migration docs and all of the potential issues called out are with features we do not use.
* Tests all pass and no issues spotted in manual testing.
* **Trims ~0.7k from each of our build sizes!**

### DIFF
--- a/clients/fides-js/package.json
+++ b/clients/fides-js/package.json
@@ -28,9 +28,8 @@
     }
   },
   "scripts": {
-    "dev": "NODE_ENV=development rollup --watch -c",
-    "build": "NODE_ENV=production rollup -c",
-    "build:windows": "set NODE_ENV=production && rollup -c",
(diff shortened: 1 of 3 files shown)


clients/fides-js/package.json
```diff
@@ -83,7 +82,7 @@
     "jest-environment-jsdom": "^29.7.0",
     "postcss": "^8.4.31",
     "prettier": "^3.3.3",
-    "rollup": "^3.29.5",
+    "rollup": "^4.44.2",
```
eastandwestwind, line 85: Do we have a follow-up to upgrade rollup version in Fidesplus as well? https://github.com/ethyca/fidesplus/blob/main/clients/fides-js-extensions/nytimes/package.json#L31

gilluminate: Great suggestion, I'll create one!

PR #6331, opened 2025-07-11
add `acceptAll` and `rejectAll` to Fides global

gilluminate:
Closes [ENG-934]

### Description Of Changes

Added convenience methods `acceptAll()` and `rejectAll()` to the Fides SDK for programmatic consent management. These methods provide the same functionality as clicking the "Accept All" and "Reject All" buttons in the UI. Refactored existing consent-related tests to use array-based approaches to reduce code duplication and improve maintainability. Added ID attributes to consent buttons so that external scripts (eg. web monitor) don't need to rely on `class` names or `data-testid` which are less reliable.

### Code Changes

* Added `acceptAll()` and `rejectAll()` methods to Fides SDK
* Implemented shared `updateAllConsent()` helper function to avoid code duplication
* Added ID attributes to consent buttons (`fides-accept-all-button`, `fides-reject-all-button`, etc.)
* Refactored Cypress tests to use array-based test cases instead of repetitive individual tests
* Added documentation for new methods

### Steps to Confirm

**Admin UI Consent Setup**

1. Add at least one vendor (`6sense` is a good one to use to test experiences like this)
2. Enable all "available" notices
3. Enable default "US Modal" experience

**Privacy Center Demo Page***

1. Visit the demo page at a location that supports the US Modal experience (eg. `http://localhost:3001/fides-js-demo.html?geolocation=us-ca`)
2. Test the new SDK methods in browser console: `Fides.acceptAll()` and `Fides.rejectAll()`
3. Verify that consent preferences are updated correctly when using the new methods (you'll see them change to/from `true` and `false` in the demo page's Consent JSON section)
4. Confirm that notice-only preferences remain unchanged when using accept/reject all methods (the "essential" notice should always be `true` no matter what)
5. Open the modal by clicking the "Manage preferences" link on the demo page
6. Ensure the toggles are either all "on" or "off" to match the acceptAll and rejectAll notice settings (except for the disabled "essential" toggle which should always be "on")
7. Check that button IDs are properly set on all of the footer buttons in the modal

### Pre-Merge Checklist

* [x] Issue requirements met
* [x] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [x] No followup issues
* Database migrations:
  * [x] No migrations
* Documentation:
  * [x] If there are any new client scopes created as part of the pull request, remember to update public-facing documentation that references our scope registry

[ENG-934]: https://ethyca.atlassian.net/browse/ENG-934?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/clients/fides-js/src/lib/preferences.ts
+++ b/clients/fides-js/src/lib/preferences.ts
@@ -62,7 +62,7 @@ async function savePreferencesApi(
   consentPreferencesToSave?: Array<
     Pick<SaveConsentPreference, "noticeHistoryId" | "consentPreference">
   >,
-  tcf?: TcfSavePreferences,
+  tcfPreferences?: TcfSavePreferences,
   userLocationString?: string,
   servedNoticeHistoryId?: string,
 ) {
@@ -83,7 +83,7 @@ async function savePreferencesApi(
     method: consentMethod,
     served_notice_history_id: servedNoticeHistoryId,
     property_id: experience.property_id,
-    ...(tcf ?? []),
+    ...(tcfPreferences ?? []),
   };
   await patchUserPreference(
     consentMethod,
@@ -104,7 +104,7 @@ interface UpdateConsentPreferencesProps {
   cookie: FidesCookie;
   eventExtraDetails?: FidesEventExtraDetails;
   servedNoticeHistoryId?: string;
-  tcf?: TcfSavePreferences;
+  tcfPreferences?: TcfSavePreferences;
   updateCookie?: (oldCookie: FidesCookie) => Promise<FidesCookie>;
 }
 /**
@@ -128,7 +128,7 @@ export const updateConsentPreferences = async ({
   cookie,
   eventExtraDetails,
   servedNoticeHistoryId,
-  tcf,
+  tcfPreferences,
   updateCookie,
 }: UpdateConsentPreferencesProps) => {
   if (!updateCookie && consentPreferencesToSave) {
@@ -187,7 +187,7 @@ export const updateConsentPreferences = async ({
         consentMethod,
         privacyExperienceConfigHistoryId,
         consentPreferencesToSave,
-        tcf,
+        tcfPreferences,
         userLocationString,
         servedNoticeHistoryId,
       );
@@ -295,7 +295,7 @@ export interface UpdateConsentOptions {
   validation?: UpdateConsentValidation;
   consentMethod?: ConsentMethod;
   eventExtraDetails?: FidesEventExtraDetails;
-  tcf?: TcfSavePreferences;
+  tcfPreferences?: TcfSavePreferences;
   updateCookie?: (oldCookie: FidesCookie) => Promise<FidesCookie>;
 }
 export const updateConsent = async (
@@ -318,7 +318,7 @@ export const updateConsent = async (
   if (
     !consentOptions?.noticeConsent &&
     !consentOptions?.fidesString &&
-    !consentOptions?.tcf
+    !consentOptions?.tcfPreferences
   ) {
     throw new Error("Either consent object or fidesString must be provided");
   }
@@ -343,7 +343,7 @@ export const updateConsent = async (
         origin: FidesEventOrigin.EXTERNAL,
       },
     },
-    tcf,
+    tcfPreferences,
     updateCookie,
   } = consentOptions;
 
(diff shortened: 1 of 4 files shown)


clients/fides-js/docs/interfaces/Fides.md
```diff
@@ -654,3 +654,29 @@ Fides.updateConsent({
   validation: "ignore"
 });
 ```
+
+***
+
+### acceptAll()
```
eastandwestwind, line 660: 💯

clients/fides-js/src/lib/preferences.ts
```diff
@@ -482,3 +482,27 @@ export const updateConsent = async (
     updateCookie,
   });
 };
+
+const updateAllConsent = async (context: FidesGlobal, accept: boolean) => {
```
eastandwestwind, line 491: Can we get some docstrings on these new methods?

clients/fides-js/src/lib/preferences.ts
```diff
@@ -482,3 +482,27 @@ export const updateConsent = async (
...
+  }
+  const privacyNoticeItems = experience.privacy_notices;
+  const consent = privacyNoticeItems?.reduce((acc, n) => {
+    if (n.consent_mechanism !== ConsentMechanism.NOTICE_ONLY) {
+      // eslint-disable-next-line no-param-reassign
+      acc[n.notice_key] = accept;
+    }
+    return acc;
+  }, {} as NoticeConsent);
+  return updateConsent(context, { noticeConsent: consent });
```
eastandwestwind, line 505: I assume we don't need servingComponent or consentMethod here? 

How about a fn to update the fides cookie for TCF? We inject this in within the `TcfOverlay.tsx` component.

clients/fides-js/src/lib/preferences.ts
```diff
@@ -482,3 +482,27 @@ export const updateConsent = async (
     updateCookie,
   });
 };
+
+const updateAllConsent = async (context: FidesGlobal, accept: boolean) => {
+  const { experience } = context;
+  if (!experience) {
+    throw new Error("Experience must be initialized before updating consent");
+  }
+  const privacyNoticeItems = experience.privacy_notices;
```
speaker-ender: Is the `const` assignment necessary here?
I think this could just be the be part of the following line.

clients/fides-js/src/lib/preferences.ts
```diff
@@ -482,3 +482,27 @@ export const updateConsent = async (
     updateCookie,
   });
 };
+
+const updateAllConsent = async (context: FidesGlobal, accept: boolean) => {
+  const { experience } = context;
+  if (!experience) {
```
speaker-ender: Is there a need to extract experience into a new const?
I can see that maybe you are trying to reduce the line length of something like the reducer, but it both abstracts what is being acted upon and creates additional variables.
Would destructuring in the fn parameters itself (something like `async ({ experience, ...rest }, accept: boolean) => {`) make sense here?

clients/fides-js/src/lib/preferences.ts
```diff
@@ -482,3 +482,27 @@ export const updateConsent = async (
...
+    throw new Error("Experience must be initialized before updating consent");
+  }
+  const privacyNoticeItems = experience.privacy_notices;
+  const consent = privacyNoticeItems?.reduce((acc, n) => {
```
speaker-ender: Using `n` as a variable name is a little vague.
Could we either:
1. Name it something like `notice` or `privacyNotice` to be more descriptive of what we are referencing
2. Destructure the values used (by something like `((acc, {consent_mechanism, notice_key}) => { ...`)

speaker-ender (commented):
Just a few small comments but otherwise looks good!

clients/fides-js/src/lib/preferences.ts
gilluminate: Good catch! TCF is going to need some extra love. The `consentMethod` defaults to "script" which is correct in this case. There is no `servingComponent` so it's also correct to leave out.

gilluminate:
Closing this for now until we can find time to get the TCF version working

PR #6337, opened 2025-07-15
Add scope check for d&d menu items

lucanovera:
### Description Of Changes
Fix D&D menu items appearing for users without the required scope.

### Code Changes
* Require DISCOVERY_MONITOR_READ scope to show D&D nav menu items
* Update roles in tests
* Update detection & discovery cypress test to use the owner role

### Steps to Confirm
1.  Login as a internal respondent user
2. Check no Discovery items in the menu appear
3. Login as any other type of user
4. Check Detection & Discovery appears in the menu

### Pre-Merge Checklist

* [x] Issue requirements met
* [x] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [x] No followup issues
* Database migrations:
  * [x] No migrations
* Documentation:


jpople (commented):
Good to add a check for this-- can we handle the case where a user has `discovery_monitor:read` but _not_ `discovery_monitor:update`?  This would allow a user with the former but not the latter to both view the pages and make changes, which seems potentially problematic.

lucanovera:
> Good to add a check for this-- can we handle the case where a user has `discovery_monitor:read` but _not_ `discovery_monitor:update`? This would allow a user with the former but not the latter to both view the pages and make changes, which seems potentially problematic.

Yeah you're right. I'll create a follow up ticket for that so we don't forget. This is a quick fix I made because I saw the new user (internal respondent) is supposed to be very limited but it was seeing that option in the menu.

jpople (approved):

PR #6339, opened 2025-07-15
Adds unordered list as an allowed HTML tag

wadesdev:
Closes ENG-960
### Description Of Changes
* Adds unordered list as an option in Experience text

### Code Changes

* Adds "ul" tag to the allowed_html_tags list and the "style" attribute so that we can margin-left adjust it

### Steps to Confirm

1. Deploy Fides with FIDES_PRIVACY_CENTER__ALLOW_HTML_DESCRIPTION=true
2. Add <ul> tag to the experience description and save
3. Re-open the editor -- the <ul> tag should not have been sanitized (it should still be there)

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:

### DIFF
--- a/src/fides/api/custom_types.py
+++ b/src/fides/api/custom_types.py
@@ -68,9 +68,7 @@ def validate_html_str(val: str) -> str:
         "u",
     }
 
-    ALLOWED_ATTRIBUTES = {
-        "a": {"href", "hreflang", "target"},
-    }
+    ALLOWED_ATTRIBUTES = {"a": {"href", "hreflang", "target"}, "ul": {"margin-left"}}
 
     if val:
         return clean(val, tags=ALLOWED_HTML_TAGS, attributes=ALLOWED_ATTRIBUTES)
--- a/src/fides/api/custom_types.py
+++ b/src/fides/api/custom_types.py
@@ -68,7 +68,7 @@ def validate_html_str(val: str) -> str:
         "u",
     }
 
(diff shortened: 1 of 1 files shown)


src/fides/api/custom_types.py
```diff
@@ -67,9 +68,7 @@ def validate_html_str(val: str) -> str:
         "u",
     }
 
-    ALLOWED_ATTRIBUTES = {
-        "a": {"href", "hreflang", "target"},
-    }
+    ALLOWED_ATTRIBUTES = {"a": {"href", "hreflang", "target"}, "ul": {"style"}}
```
thabofletcher, line 71: Is the allowing of `style` required for this change? One of the primary reasons to filter attributes is to prevent `style` attributes as those are one of the XSS attack vectors.

wadesdev: @thabofletcher from my testing, I couldn't get the bulleted list to actually indent on sub-bullets without adding `style="margin-left: 1em"` -- but I didn't get too creative

thabofletcher: There might be a CSS reset or something causing that to not indent - but we probably want to address that at the privacy experience level rather than custom CSS for a few reason. Not just the security concern but also so everyone gets the benefit of the change

PR #6340, opened 2025-07-15
Optimizing BigQuery deletes

galvana:
Closes [ENG-908](https://ethyca.atlassian.net/browse/ENG-908) and [ENG-994](https://ethyca.atlassian.net/browse/ENG-994)

### Description Of Changes

This PR introduces "batched deletes" for BigQuery collection erasures. This reduces the number of delete statements that need to be executed.

Previously, if an access request returned multiple rows, identical delete statements would be executed for each row. For example:
```
SELECT id, email FROM orders WHERE email = "customer-1@example.com"

id, email
1, customer-1@example.com
2, customer-1@example.com
...
1000, customer-1@example.com
```
This would have resulted in the execution of 1000 identical delete statements
```
DELETE FROM orders WHERE email = "customer-1@example.com"
```

### Code Changes

* Updated `generate_delete()` in `BigQueryQueryConfig` to use the same input data that is used to generate select statements for access requests.

### Steps to Confirm

This is a bit tricky to setup so I recommend reviewing the test cases and helping me spot any missing scenarios. I added tests to `TestBigQueryQueryConfig` and `TestBigQueryConnector`

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:

[ENG-908]: https://ethyca.atlassian.net/browse/ENG-908?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[ENG-994]: https://ethyca.atlassian.net/browse/ENG-994?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/fides/api/service/connectors/base_connector.py
```diff
@@ -82,6 +82,7 @@ def mask_data(
         privacy_request: PrivacyRequest,
         request_task: RequestTask,
         rows: List[Row],
+        input_data: Optional[Dict[str, List[Any]]] = None,
```
galvana, line 85: Updating the `mask_data` method signature to include `input_data`. Previously we were just using the `rows` to generate delete statements but we want to be more efficient and create a single delete statement with the same where clauses as a select statement.

src/fides/api/service/connectors/bigquery_connector.py
```diff
@@ -142,22 +137,42 @@ def mask_data(
...
-        for row in rows:
-            update_or_delete_stmts: List[Executable] = (
-                query_config.generate_masking_stmt(
-                    node, row, policy, privacy_request, client
-                )
-            )
-            if update_or_delete_stmts:
+
+        # Check if we're using DELETE masking strategy
+        if query_config.uses_delete_masking_strategy():
```
galvana, line 192: Instead of calling the generic `generate_masking_stmt` we call `generate_delete` or `generate_update` directly based on the masking strategy. I added this distinction because we don't need to iterate over the rows to create the delete statements.

src/fides/api/service/connectors/dynamodb_connector.py
```diff
@@ -140,8 +140,9 @@ def mask_data(
         privacy_request: PrivacyRequest,
         request_task: RequestTask,
         rows: List[Row],
+        input_data: Optional[Dict[str, List[Any]]] = None,
```
galvana, line 143: Just adding the new param to all these connectors, nothing notable here

src/fides/api/service/connectors/query_configs/bigquery_query_config.py
```diff
@@ -218,37 +219,52 @@ def generate_update(
 
         return [table.update().where(*where_clauses).values(**final_update_map)]
 
-    def generate_delete(self, row: Row, client: Engine) -> List[Delete]:
-        """Returns a List of SQLAlchemy DELETE statements for BigQuery. Does not actually execute the delete statement.
+    def generate_delete(
```
galvana, line 222: Key change, this now uses the same input data that the corresponding access node received to generate the where clauses for the delete statement

src/fides/api/task/execute_request_tasks.py
```diff
@@ -298,17 +298,11 @@ def run_access_node(
...
+                            _build_upstream_access_data(
                                 graph_task.execution_node.input_keys, upstream_results
                             )
                         )
-                        # Pass in access data dependencies in the same order as the input keys.
-                        # If we don't have access data for an upstream node, pass in an empty list
-                        upstream_access_data: List[List[Row]] = [
-                            upstream.get_access_data() if upstream else []
-                            for upstream in ordered_upstream_tasks
-                        ]
```
galvana, line 311: Replacing this block with a single call to `_build_upstream_access_data`

src/fides/api/task/execute_request_tasks.py
```diff
@@ -368,8 +362,24 @@ def run_erasure_node(
...
+                    except Exception as e:
+                        logger.error(
+                            f"Unable to get upstream access data for erasure task {request_task.collection_address}: {e}"
+                        )
+
+                    # Run the main erasure function, passing along the upstream access data.
+                    # The extra data is currently only needed for generating BigQuery delete statements.
+                    erasure_graph_task.erasure_request(
+                        retrieved_data, inputs=upstream_access_data
+                    )
```
galvana, line 382: This is where we pass the new access inputs to `erasure_request`

src/fides/api/service/connectors/bigquery_connector.py
```diff
@@ -142,22 +137,40 @@ def mask_data(
...
```
NevilleS, line 192: nit: Is it possible to move this up a level of abstraction, so that _all_ connectors will check for the `delete` strategy, and throw an "unsupported masking strategy" error if we attempt to use it?

src/fides/api/service/connectors/bigquery_connector.py
```diff
@@ -142,22 +137,40 @@ def mask_data(
...
-                            update_or_delete_stmt
-                        )
-                        update_or_delete_ct = update_or_delete_ct + results.rowcount
+                    for delete_stmt in delete_stmts:
+                        delete_results = connection.execute(delete_stmt)
+                        logger.debug(f"Deleted {delete_results.rowcount} rows")
+                        update_or_delete_ct += delete_results.rowcount
+        else:
+            # For UPDATE operations, process each row individually since each row may have different masked values
+            for row in rows:
```
NevilleS, line 199: This loop is still a potentially large footgun. I'd like us to add a log like:
```suggestion
            logger.debug(f"Starting to generate and execute UPDATE statements for {len(rows)} target rows...")
            for row in rows:
```

(or similar)

src/fides/api/service/connectors/bigquery_connector.py
```diff
@@ -142,22 +137,40 @@ def mask_data(
...
+                update_or_delete_stmts: List[Executable] = query_config.generate_update(
+                    row, policy, privacy_request, client
+                )
+                logger.debug(
+                    f"Generated {len(update_or_delete_stmts)} UPDATE statements"
```
NevilleS, line 204: nit: it'd be nice if logs like this self-reported their status, like "Generated 5 UPDATE statements for target row 2/40"

src/fides/api/task/execute_request_tasks.py
```diff
@@ -487,6 +497,73 @@ def _order_tasks_by_input_key(
...
+    # Get the corresponding access task for the current erasure task
+    access_request_task = (
+        session.query(RequestTask)
+        .filter(
+            RequestTask.privacy_request_id == erasure_request_task.privacy_request_id,
+            RequestTask.collection_address == erasure_request_task.collection_address,
+            RequestTask.action_type == ActionType.access,
+        )
+        .first()
+    )
```
NevilleS, line 532: OK, this part feels the most fragile. Is there any other way to pass this state along, so that we don't have to search for the access task?

NevilleS: ie after the traversal, both the access & erasure tasks are created at the same time, can we pass this context along at that time, or inject a specific task ID so there is no lookup, etc?

src/fides/api/task/graph_task.py
```diff
@@ -645,9 +645,15 @@ def erasure_request(
...
         retrieved_data: List[Row],
         *erasure_prereqs: int,  # TODO Remove when we stop support for DSR 2.0. DSR 3.0 enforces with downstream_tasks.
+        inputs: Optional[
+            List[List[Row]]
+        ] = None,  # Upstream data from corresponding access task
     ) -> int:
         """Run erasure request"""
 
+        if inputs is None:
+            inputs = []
```
NevilleS, line 655: Oh, Python... 😄

tests/ops/service/connectors/test_bigquery_connector.py
```diff
@@ -140,6 +316,122 @@ def test_generate_update_partitioned_table(
...
+            node=execution_node,
+            policy=erasure_policy,
+            privacy_request=PrivacyRequest(),
+            request_task=RequestTask(),
+            rows=[],
+        )
+
+        # Nothing should have been deleted / updated and no SQL should have executed
+        assert update_or_delete_ct == 0
+        assert execute_spy.call_count == 0
```
NevilleS, line 435: Can we assert this results in an error? This should only be possible if something goes wrong with the orchestration of these tasks, right?

tests/ops/service/connectors/test_bigquery_connector.py
```diff
@@ -83,6 +87,178 @@ def execution_node_with_namespace_and_partitioning_meta(
...
+                    )
+                )
+        except Exception as exc:
+            # Log cleanup errors but don't fail the test
+            logger.warning(f"Failed to clean up test data: {exc}")
+
+    @pytest.fixture
+    def multiple_customer_data(
+        self, bigquery_example_test_dataset_config_with_namespace_and_partitioning_meta
+    ):
```
NevilleS, line 194: I would like to see a test that checks how we delete from the `visits_partitioned` table in this test harness.

I'd recommend just making the default fixture have some more entries so that there's ~50 "visits" in there, and that the regular `customer-1@example.com` has at least 10 matching rows.

That should provide ample ground to demonstrate that the batched delete shows a material difference in stmt counts compared to the update.

tests/ops/service/connectors/test_bigquery_connector.py
```diff
@@ -140,6 +316,122 @@ def test_generate_update_partitioned_table(
...
+        connector = BigQueryConnector(dataset_config.connection_config)
+
+        # Set DELETE masking strategy for the customer collection
+        execution_node = execution_node_with_namespace_and_partitioning_meta
+        execution_node.collection.masking_strategy_override = MaskingStrategyOverride(
+            strategy=MaskingStrategies.DELETE
+        )
+
+        erasure_policy.rules[0].targets[0].data_category = "user"
+        erasure_policy.rules[0].targets[0].save(db)
```
NevilleS, line 342: Why? This seems a bit out of place

tests/ops/service/connectors/test_bigquery_connector.py
```diff
@@ -140,6 +316,122 @@ def test_generate_update_partitioned_table(
...
+            privacy_request=PrivacyRequest(),
+            request_task=RequestTask(),
+            rows=customer_data["rows"],
+            input_data={"email": ["customer-3@example.com"], "address_id": [3, 4, 5]},
+        )
+        # Delete all 3 customer rows
+        assert update_or_delete_ct == 3
+
+        # Verify that we only execute one delete per partition, not per row
+        assert execute_spy.call_count == 2
```
NevilleS, line 358: FWIW: this test setup looks really good. The fixtures are ugly and clog the file, but the actual runner is clear and legible.

Can we also have a matching test with the same setup as this one, but *don't* do the DELETE override. Confirm that we see the expected behavior for UPDATEs. This'll be especially helpful if you update the fixtures to include ~10 visit rows, because we'd expect to see 20 update calls.

tests/ops/service/connectors/test_bigquery_connector.py
```diff
@@ -140,6 +316,122 @@ def test_generate_update_partitioned_table(
...
+        # Set DELETE masking strategy for the customer collection
+        execution_node = (
+            execution_node_with_namespace_and_partitioning_meta_and_multiple_identities
+        )
+        execution_node.collection.masking_strategy_override = MaskingStrategyOverride(
+            strategy=MaskingStrategies.DELETE
+        )
+
+        erasure_policy.rules[0].targets[0].data_category = "user"
+        erasure_policy.rules[0].targets[0].save(db)
```
NevilleS, line 383: Same question: ?

tests/ops/task/test_execute_request_tasks.py
```diff
@@ -517,3 +518,108 @@ def test_use_dsr_2_0_override(
...
+            # Should return the mocked upstream data
+            assert result == mock_upstream_data
+            assert isinstance(result, list)
+
+    def test_get_upstream_access_data_missing_access_task(
+        self,
+        db,
+        privacy_request,
+        connection_config,
+    ):
```
NevilleS, line 596: OK, I'm glad this is tested. I'd still prefer if we can remove this possibility, though.

tests/ops/service/connectors/test_bigquery_queryconfig.py
```diff
@@ -239,13 +239,236 @@ def test_generate_delete_stmt(
...
+            },
+        )
+
+        # Should generate ONE delete statement since the where clauses for the delete statement are the same as those for a select statement (access request)
+        assert len(delete_stmts) == 1
+
+        # Check the SQL statement structure - all should have the same structure
+        stmts = set(str(stmt) for stmt in delete_stmts)
+        expected_stmts = {
+            "DELETE FROM `employee` WHERE `employee`.`address_id` IN UNNEST(%(address_id_1:STRING)s) OR `employee`.`email` IN UNNEST(%(email_1:STRING)s)"
```
NevilleS: Why the `UNNEST` here?

NevilleS (commented):
Some requests, a few nits, and a couple comments.

This looks mostly right to me though.

src/fides/api/task/execute_request_tasks.py
```diff
@@ -368,8 +362,24 @@ def run_erasure_node(
...
-                    graph_task.erasure_request(retrieved_data)
+                    upstream_access_data: List[List[Row]] = []
+
+                    try:
+                        upstream_access_data = (
+                            get_upstream_access_data_for_erasure_task(
+                                request_task, session, resources
+                            )
+                        )
+                    except Exception as e:
```
thabofletcher, line 373: I'm not sure if this is off or just my understanding is off - but it seems like if its an actual error condition we should set an error state on the task and stop processing right here. If its expected that we might run into an exception or the data isn't needed (non BigQuery deletes etc), we might be able to handle it without an exception. I'm not familiar enough to know what the case is where we would get an exception and continue on to erasure_request

thabofletcher: Adrian explained this to me in DM - upstream_access_data will be empty but we want to show a log in this case, but erasure_request already handles the empty input collection

PR #6351, opened 2025-07-17
Remove `data-testid` from FidesJS prod builds

gilluminate:
Closes [ENG-968]

### Description Of Changes

Reduces build size and prevents anyone from relying on our test IDs for production functionality (like web monitor)

### Code Changes

* Add `rollup-plugin-jsx-remove-attributes` which defaults to removing `data-testid`
* Updated `jsx` definitions in `tsconfig.json` to use a more modern approach where we no longer need to import `h` from preact and JSX gets treated as JSX. This was necessary to get the `jsxRemoveAttributes` plugin to work properly
* Removed all now-unnecessary `h` imports
* Added `build:test` to preserve `data-testid` when CI builds for Cypress
* Included rendered docs from previous branch

### Steps to Confirm

1. Run privacy center
2. build FidesJS using `npm run dev`
3. Inspect the "Accept all" button and ensure it **does** have `data-testid` attribute
4. build FidesJS using `npm run build:test`
5. Inspect the "Accept all" button and ensure it **does** have `data-testid` attribute
6. build FidesJS using `npm run build`
7. Inspect the "Accept all" button and ensure it **does not** have `data-testid` attribute

### Pre-Merge Checklist

* [x] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [x] No followup issues
* Database migrations:
  * [x] No migrations
* Documentation:


nrxsmith:
I'm likely to start running e2e tests against the various staging environments soon, including the RC environment hopefully at some point. I'll need to take a look and see how we're running fides on those environments. If any of them aren't running dev, that would prevent me from running the tests as they're currently written. So we'd need to come up with something.

Perhaps an environment like RC, though, should really be a prod replica as much as possible, and any tests we run there shouldn't rely on testIds.

gilluminate:
@nrxsmith we don't necessarily need to run those in `dev` mode, but rather just include a `IS_TEST:true` environment variable. Using the new `npm run build:test` instead of `npm run build` should solve for that if we can update those.

PR #6352, opened 2025-07-17
Eng 908 safe mode bq delete

thabofletcher:
Allows for ENG-908 to be verified via logs

### Description Of Changes

Adds a safe_mode configuration and uses it to log destructive SQL statements instead of executing them

### Code Changes

* safe_mode configuration
* safe_mode usage in bq and sql connectors
* autoamted tests for each

### Steps to Confirm

1. Update the configuration using this API call to enable safe mode:
``curl --request PATCH \
  --url 'http://localhost:3000/api/v1/config?api_set=true' \
  --header 'Authorization: Bearer YOUR_TOKEN' \
  --data '
{
  "execution": {
    "safe_mode": false
  }
}'```

2. Verify the config was added
```curl --request GET \
  --url 'http://localhost:3000/api/v1/config?api_set=true' \
  --header 'Authorization: Bearer YOUR_TOKEN'```

3. Make sure you have a BigQuery integration with relevant rows to delete, then make an Erasure request. 
4. Verify that the "SAFE MODE - Would execute" log appears and would delete only the relevant rows

### DIFF
--- a/src/fides/api/service/connectors/bigquery_connector.py
+++ b/src/fides/api/service/connectors/bigquery_connector.py
@@ -9,6 +9,7 @@ from sqlalchemy.sql.elements import TextClause
 
 from fides.api.common_exceptions import ConnectionException
 from fides.api.graph.execution import ExecutionNode
+from fides.api.models.application_config import ApplicationConfig
 from fides.api.models.connectionconfig import ConnectionTestStatus
 from fides.api.models.policy import Policy
 from fides.api.models.privacy_request import PrivacyRequest, RequestTask
@@ -21,6 +22,7 @@ from fides.api.service.connectors.query_configs.bigquery_query_config import (
 from fides.api.service.connectors.query_configs.query_config import SQLQueryConfig
 from fides.api.service.connectors.sql_connector import SQLConnector
 from fides.api.util.collection_util import Row
+from fides.config.config_proxy import ConfigProxy
 
 
 class BigQueryConnector(SQLConnector):
@@ -145,17 +147,28 @@ class BigQueryConnector(SQLConnector):
         update_or_delete_ct = 0
         client = self.client()
 
+        # Check if safe_mode is enabled
+        db: Session = Session.object_session(self.configuration)
+        config_proxy = ConfigProxy(db)
+        safe_mode_enabled = getattr(config_proxy.execution, 'safe_mode', False)
+
         # Check if we're using DELETE masking strategy
         if query_config.uses_delete_masking_strategy():
             # Use batched DELETE for better performance
             delete_stmts = query_config.generate_delete(client, input_data or {})
             logger.debug(f"Generated {len(delete_stmts)} DELETE statements")
             if delete_stmts:
-                with client.connect() as connection:
+                if safe_mode_enabled:
+                    # In safe mode, log the DELETE statements instead of executing them
                     for delete_stmt in delete_stmts:
-                        delete_results = connection.execute(delete_stmt)
-                        logger.debug(f"Deleted {delete_results.rowcount} rows")
-                        update_or_delete_ct += delete_results.rowcount
+                        logger.warning(f"SAFE MODE - Would execute DELETE: {delete_stmt}")
+                else:
+                    # Normal mode - execute the DELETE statements
+                    with client.connect() as connection:
+                        for delete_stmt in delete_stmts:
+                            delete_results = connection.execute(delete_stmt)
+                            logger.debug(f"Deleted {delete_results.rowcount} rows")
+                            update_or_delete_ct += delete_results.rowcount
         else:
             # For UPDATE operations, process each row individually since each row may have different masked values
             for row in rows:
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: src/fides/api/service/connectors/bigquery_connector.py [hunk 1]
```diff
@@ -130,6 +131,41 @@ def test_connection(self) -> Optional[ConnectionTestStatus]:
...
+        Returns:
+            Number of rows affected (0 in safe mode)
+        """
+        if not statements:
+            return 0
+
+        if safe_mode_enabled:
+            # In safe mode, log the statements instead of executing them
+            for stmt in statements:
+                logger.warning(f"SAFE MODE - Would execute SQL: {stmt}")
```
NevilleS, line 156: ❤️

src/fides/api/service/connectors/sql_connector.py
```diff
@@ -184,15 +185,26 @@ def mask_data(
         query_config = self.query_config(node)
         update_ct = 0
         client = self.client()
+
+        # Check if safe_mode is enabled
+        db: Session = Session.object_session(self.configuration)
+        config_proxy = ConfigProxy(db)
+        safe_mode_enabled = getattr(config_proxy.execution, "safe_mode", False)
```
NevilleS: I rarely suggest this (I'm not much of a lover of inheritance), but this does seem to be a scenario where this could be defined on the BaseConnector class and reused for SqlConnector, BigQueryConnector, etc.

src/fides/api/service/connectors/sql_connector.py
```diff
@@ -184,15 +185,26 @@ def mask_data(
...
                 row, policy, privacy_request
             )
             if update_stmt is not None:
-                with client.connect() as connection:
-                    self.set_schema(connection)
-                    results: LegacyCursorResult = connection.execute(update_stmt)
-                    update_ct = update_ct + results.rowcount
+                if safe_mode_enabled:
+                    # In safe mode, log the SQL statement instead of executing it
+                    logger.warning(f"SAFE MODE - Would execute SQL: {update_stmt}")
```
NevilleS, line 212: 👏 for (starting to) add this more generically than just BQ... it's a good pattern for us.

there are other derived connectors though that override mask_data themselves in an unsafe way, though.

NevilleS (commented):
Overall: this looks like what we need

src/fides/api/service/connectors/bigquery_connector.py
```diff
@@ -130,6 +131,41 @@ def test_connection(self) -> Optional[ConnectionTestStatus]:
...
+            return 0
+
+        # Normal mode - execute the statements
+        row_count = 0
+        with client.connect() as connection:
+            for stmt in statements:
+                results = connection.execute(stmt)
+                logger.debug(f"Affected {results.rowcount} rows")
```
thabofletcher, line 164: Subtle change here - in order to share the code effectively in this function I had to change the log where it use to say "Updated" or "Deleted" depending on the calling code's conditional

src/fides/config/utils.py
```diff
@@ -61,6 +61,7 @@ def get_dev_mode() -> bool:
         "task_retry_backoff",
         "require_manual_request_approval",
         "subject_identity_verification_required",
+        "safe_mode",
```
thabofletcher, line 64: This was added to a collection that says (see comment at the top of this config):

# Any updates to this list _need_ to be reviewed by the Ethyca security team

src/fides/api/service/connectors/sql_connector.py
thabofletcher: It looks like most of the other classes are using Depends(get_config_proxy) which would tidy this up and make it more consistent with how the rest of the code gets configs. I'll clean this up and the other instance

thabofletcher: Ok I did end up using inheritance since this was the base class and the other standard patterns didn't apply (CONFIG only  works for static configs and Depends only works on the FastAPI endpoints themselves)

tests/ops/service/connectors/test_bigquery_connector.py
```diff
@@ -525,6 +525,110 @@ def test_retrieve_partitioned_data_with_multiple_identifying_fields(
...
+
+        # In safe mode, no rows are actually affected so count should be 0
+        assert update_or_delete_ct == 0
+
+        # Check that the DELETE statements were logged as warnings instead of executed
+        assert "SAFE MODE - Would execute SQL:" in loguru_caplog.text
+
+        # Clean up: disable safe_mode
+        ApplicationConfig.update_api_set(db, {"execution": {"safe_mode": False}})
+        db.commit()
```
galvana, line 580: Nice! 👌

src/fides/config/utils.py
daveqnet: @thabofletcher this is approved from a security perspective. Exposing the value of `FIDES__EXECUTION__SAFE_MODE` in response to a `GET /api/v1/config` API request has no security impact, unlike, say, the value of `FIDES__DATABASE__PASSWORD` or `FIDES__SECURITY__APP_ENCRYPTION_KEY`.

PR #6356, opened 2025-07-17
[ENG-979] Create new conditional dependencies table

JadeCara:
Closes [ENG-979](https://ethyca.atlassian.net/browse/ENG-979?atlOrigin=eyJpIjoiYWJmMjIyYjU4OWZhNGNmZGE4M2JkZGM5YmZiZGNjYWIiLCJwIjoiaiJ9)

### Description Of Changes

In order to store conditional dependencies in a clean queryable way, we are adding a new table rather than including what could be a deeply nested JSONB column into the ManualTasks.

This PR introduces this table, with migration and adds tests. 

### Code Changes

* DB migration for ManualTaskConditionalDependency
* Add ORM for ManualTaskConditionalDependency
* Add tests for the new table

### Steps to Confirm

1.  Verify All tests pass.

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:


src/fides/api/models/manual_task.py
```diff
@@ -975,3 +993,105 @@ def create_error_log(
...
+        "ManualTaskConditionalDependency",
+        remote_side=[id],
+        back_populates="children",
+        foreign_keys=[parent_id],
+    )
+    children = relationship(
+        "ManualTaskConditionalDependency",
+        back_populates="parent",
+        cascade="all, delete-orphan",
+        foreign_keys=[parent_id],
```
JadeCara:
```suggestion
        foreign_keys=[parent_id],
        passive_delete=True,
```

src/fides/api/task/conditional_dependencies/evaluator.py
```diff
@@ -44,6 +44,7 @@ def _evaluate_leaf_condition(
         self, condition: ConditionLeaf, data: Union[dict, Any]
     ) -> bool:
         """Evaluate a leaf condition against input data"""
+        # Get the actual value from the data
```
JadeCara:
```suggestion
```

src/fides/api/models/manual_task.py
```diff
@@ -241,6 +253,12 @@ def __tablename__(cls) -> str:
         uselist=True,
         viewonly=True,  # No cascade delete - submissions are historical data
     )
+    conditional_dependencies = relationship(
+        "ManualTaskConditionalDependency",
+        back_populates="task",
+        uselist=True,
+        cascade="all, delete-orphan",
+    )
```
JadeCara, line 253:
```suggestion
```

src/fides/api/models/manual_task.py
```diff
@@ -975,3 +987,109 @@ def create_error_log(
...
+    @declared_attr
+    def __tablename__(cls) -> str:
+        """Overriding base class method to set the table name."""
+        return "manual_task_conditional_dependency"
+
+    # redefined here because there's a minor, unintended discrepancy between
+    # this `id` field and that of the `Base` class, which explicitly sets `index=True`.
+    # TODO: we likely should _not_ be setting `index=True` on the `id`
+    # attribute of the `Base` class, as `primary_key=True` already specifies a
+    # primary key constraint, which will implicitly create an index for the field.
```
JadeCara:
```suggestion
```

src/fides/api/models/manual_task.py
```diff
@@ -975,3 +987,106 @@ def create_error_log(
...
+
+    @declared_attr
+    def __tablename__(cls) -> str:
+        """Overriding base class method to set the table name."""
+        return "manual_task_conditional_dependency"
+
+    # redefined here because there's a minor, unintended discrepancy between
+    # this `id` field and that of the `Base` class, which explicitly sets `index=True`.
+    id = Column(String(255), primary_key=True, default=FidesBase.generate_uuid)
+
```
JadeCara:
```suggestion
```

src/fides/api/models/manual_task.py
```diff
@@ -123,6 +128,13 @@ def get_valid_transitions(cls, current_status: "StatusType") -> list["StatusType
         return transitions.get(current_status, [])
 
 
+class ManualTaskConditionalDependencyType(str, Enum):
+    """Enum for manual task conditional dependency types."""
+
+    leaf = "leaf"
+    group = "group"
```
erosselli: can we explain in the docstring / give an example of when you would expect each type? I don't know if I follow, is group just a non-terminal "node" in the dependency graph?

src/fides/api/models/manual_task.py
```diff
@@ -975,3 +994,107 @@ def create_error_log(
             message=message,
             details=details,
         )
+
+
+class ManualTaskConditionalDependency(Base):
```
erosselli: this file is starting to get large, maybe we can move this new model under a `manual_tasks` folder and have something like `conditional_dependency.py` ?

src/fides/api/models/manual_task.py
```diff
@@ -975,3 +994,107 @@ def create_error_log(
...
+
+    @declared_attr
+    def __tablename__(cls) -> str:
+        """Overriding base class method to set the table name."""
+        return "manual_task_conditional_dependency"
+
+    # redefined here because there's a minor, unintended discrepancy between
+    # this `id` field and that of the `Base` class, which explicitly sets `index=True`.
+    # In order for the `id` to be referenced by parent_id, we need to redefine it here.
+    id = Column(String(255), primary_key=True, default=FidesBase.generate_uuid)
```
erosselli: why do we not want `index=True` here?

src/fides/api/models/manual_task.py
```diff
@@ -975,3 +994,107 @@ def create_error_log(
...
+        Index("ix_manual_task_conditional_dependency_sort_order", "sort_order"),
+    )
+
+    # Relationships
+    task = relationship("ManualTask", back_populates="conditional_dependencies")
+    parent = relationship(
+        "ManualTaskConditionalDependency",
+        remote_side=[id],
+        back_populates="children",
+        foreign_keys=[parent_id],
```
erosselli: what is `foreign_keys` doing here? I don't think I've seen it used like this before so just curious, trying to learn more SQLAlchemy

src/fides/api/models/manual_task.py
```diff
@@ -975,3 +994,107 @@ def create_error_log(
...
+    )
+    children = relationship(
+        "ManualTaskConditionalDependency",
+        back_populates="parent",
+        cascade="all, delete-orphan",
+        foreign_keys=[parent_id],
+    )
+
+    def to_condition_leaf(self) -> "ConditionLeaf":
```
erosselli: why are the quotes needed in the return type?

src/fides/api/models/manual_task.py
JadeCara: Updated the doc string! Good call :)

src/fides/api/models/manual_task.py
JadeCara: I think it is because `primary_key=True` already sets `index=True` and that can cause some wierd self reference issues. I am actually redefining because I **am** self referencing here and the function for `remote_side=[id]` does not like the id func that is getting passed from base. I updated the comment to better reflect.

src/fides/api/models/manual_task.py
JadeCara: The `parent` relationship is self referential and sometimes SQLAlchemy can get confused when making a self referencing table. 

This gets used mostly with:
Self-referential relationships (like this one)
Multiple foreign keys to the same table
Complex join conditions where SQLAlchemy can't auto-detect

src/fides/api/models/manual_task.py
JadeCara: Umm usually that would be in the case that Type didnt exist or we are using type checking or something similar. In this case - they are an artifact and don't need to be there! Thank you! good eye.

src/fides/api/models/manual_task.py
JadeCara: I love this plan. 

I have done it... which makes the diff a bit larger 😅

src/fides/api/db/base.py
```diff
@@ -33,12 +33,23 @@
...
 from fides.api.models.manual_task import (
     ManualTask,
+    ManualTaskConditionalDependency,
+    ManualTaskConditionalDependencyType,
     ManualTaskConfig,
     ManualTaskConfigField,
+    ManualTaskConfigurationType,
+    ManualTaskEntityType,
+    ManualTaskExecutionTiming,
+    ManualTaskFieldType,
```
JadeCara:
```suggestion
```

src/fides/api/db/base.py
```diff
@@ -33,12 +33,23 @@
 from fides.api.models.location_regulation_selections import LocationRegulationSelections
 from fides.api.models.manual_task import (
     ManualTask,
+    ManualTaskConditionalDependency,
+    ManualTaskConditionalDependencyType,
```
JadeCara:
```suggestion
```

src/fides/api/db/base.py
```diff
@@ -33,12 +33,23 @@
...
     ManualTaskConfig,
     ManualTaskConfigField,
+    ManualTaskConfigurationType,
+    ManualTaskEntityType,
+    ManualTaskExecutionTiming,
+    ManualTaskFieldType,
     ManualTaskInstance,
     ManualTaskLog,
+    ManualTaskLogStatus,
+    ManualTaskParentEntityType,
```
JadeCara:
```suggestion
```

src/fides/api/db/base.py
```diff
@@ -33,12 +33,23 @@
...
     ManualTaskReference,
+    ManualTaskReferenceType,
```
JadeCara:
```suggestion
```

src/fides/api/db/base.py
```diff
@@ -33,12 +33,23 @@
...
     ManualTaskSubmission,
+    ManualTaskType,
+    StatusType,
```
JadeCara:
```suggestion
```

PR #6387, opened 2025-07-25
Improved API call loading/failure handling

gilluminate:
Closes [ENG-940]

### Description Of Changes

1. Improved compliance by automatically defaulting any property accessed on `Fides.consent` as `false` in case an experience fails to load or is empty. (ie. `Fides.consent.whatever` will return `false`)

2. Improved TCF (Transparency and Consent Framework) error handling and user experience by implementing retry mechanisms for API calls and better loading states. When the full TCF experience fails to load due to network issues or API failures, users now see appropriate loading indicators and error messages, while still being able to accept/reject consent. The system now uses a retry-able fetch hook with exponential back-off for more reliable API communication.

### Code Changes

* Added `useRetryableFetch` hook to `TcfOverlay` with retry logic for fetching full TCF experience
* Created `TcfLoadingErrorMessage` component to display user-friendly error messages when TCF data fails to load
* Added `RecordsListSkeletons` component for skeleton loading states during data fetching
* Removed manual loading state management from `ConsentButtons` that was causing UX issues
* Updated CSS variables and styles to support skeleton loading animations and error states without affecting existing customer styles.
* Enhanced `InfoBox` component to accept additional props for error styling
* Fixed accessibility issues by adding `tabIndex` to scrollable content and improving `aria-checked` values
* Increased TCF bundle size limit from 92KB to 95KB to accommodate new features (increased by 0.4k, now totaling 92.46k)
* Simplified consent button logic by removing `isMinimalTCF` prop dependencies
* Updated type definitions to allow `false` return values from render functions for when we remove the save button on failure.

### Steps to Confirm
1. Navigate to the demo page in the Vercel deploy with TCF consent banner (`/fides-js-demo.html?geolocation=eea`)
2. Block network requests to simulate API failures and verify error messages appear
3. Test that users can still accept/reject consent even when full experience fails to load
4. Verify skeleton loading states appear during data fetching
5. Confirm retry mechanism works by restoring network and observing successful data load

CleanShot 2025-07-25 at 14 10 45@2x

CleanShot 2025-07-25 at 14 10 29@2x

### Pre-Merge Checklist

* [x] Issue requirements met
* [x] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [x] No followup issues
* Database migrations:
  * [x] No migrations
* Documentation:

[ENG-940]: https://ethyca.atlassian.net/browse/ENG-940?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/clients/fides-js/src/components/tcf/TCFBannerSupplemental.tsx
+++ b/clients/fides-js/src/components/tcf/TCFBannerSupplemental.tsx
@@ -31,7 +31,9 @@ export const TCFBannerSupplemental = ({
             </h2>
           </div>
         )}
-      <div className="fides-banner__content">
+      {/* adding tabindex because the content is scrollable so a11y requires it to be focusable */}
+      {/* eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex */}
+      <div className="fides-banner__content" tabIndex={0}>
         <ul className="fides-banner__purpose-list">
           {customPurposes?.map((purpose) => (
             <li key={purpose} className="fides-banner__purpose-item">
--- a/clients/fides-js/src/components/tcf/TCFBannerSupplemental.tsx
+++ b/clients/fides-js/src/components/tcf/TCFBannerSupplemental.tsx
@@ -31,7 +31,8 @@ export const TCFBannerSupplemental = ({
             </h2>
           </div>
         )}
-      {/* adding tabindex because the content is scrollable so a11y requires it to be focusable */}
+      {/* adding tabindex because the content is scrollable so a11y requires it to be focusable
+          see https://accessibilityinsights.io/info-examples/web/scrollable-region-focusable/ */}
       {/* eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex */}
       <div className="fides-banner__content" tabIndex={0}>
         <ul className="fides-banner__purpose-list">
--- a/clients/fides-js/src/components/tcf/TcfOverlay.tsx
+++ b/clients/fides-js/src/components/tcf/TcfOverlay.tsx
@@ -680,63 +680,54 @@ export const TcfOverlay = () => {
           </ConsentBanner>
         );
       }}
-      renderModalContent={
-        !experienceFull
-          ? undefined
-          : () => (
-              <TcfTabs
-                experience={experienceFull}
-                customNotices={privacyNoticesWithBestTranslation}
-                enabledIds={draftIds}
-                onChange={handleToggleChange}
-                activeTabIndex={activeTabIndex}
-                onTabChange={handleTabChange}
+      renderModalContent={() => (
+        <TcfTabs
+          experience={experienceFull || experienceMinimal}
+          customNotices={privacyNoticesWithBestTranslation}
+          enabledIds={draftIds}
+          onChange={handleToggleChange}
+          activeTabIndex={activeTabIndex}
+          onTabChange={handleTabChange}
+        />
+      )}
+      renderModalFooter={({ onClose }) => {
+        const onSave = (consentMethod: ConsentMethod, keys: EnabledIds) => {
+          handleUpdateAllPreferences(consentMethod, keys);
+          onClose();
+        };
+        return (
+          <TcfConsentButtons
+            experience={experienceFull || experienceMinimal}
+            onAcceptAll={() => {
+              handleAcceptAll();
+              onClose();
+            }}
+            onRejectAll={() => {
+              handleRejectAll();
+              onClose();
+            }}
+            renderFirstButton={() => (
+              <Button
+                buttonType={ButtonType.SECONDARY}
+                label={experienceFull ? i18n.t("exp.save_button_label") : ""}
+                onClick={() => {
+                  setTrigger({
+                    type: FidesEventTargetType.BUTTON,
+                    label: i18n.t("exp.save_button_label"),
+                  });
+                  onSave(ConsentMethod.SAVE, draftIds);
+                }}
+                className="fides-save-button"
+                id="fides-save-button"
+                disabled={!experienceFull}
+                loading={!experienceFull}
               />
-            )
-      }
-      renderModalFooter={
-        !experienceFull
-          ? undefined
-          : ({ onClose }) => {
-              const onSave = (
-                consentMethod: ConsentMethod,
-                keys: EnabledIds,
-              ) => {
-                handleUpdateAllPreferences(consentMethod, keys);
-                onClose();
-              };
-              return (
-                <TcfConsentButtons
-                  experience={experienceFull}
-                  onAcceptAll={() => {
-                    handleAcceptAll();
-                    onClose();
-                  }}
-                  onRejectAll={() => {
-                    handleRejectAll();
-                    onClose();
-                  }}
-                  renderFirstButton={() => (
-                    <Button
-                      buttonType={ButtonType.SECONDARY}
-                      label={i18n.t("exp.save_button_label")}
-                      onClick={() => {
-                        setTrigger({
-                          type: FidesEventTargetType.BUTTON,
-                          label: i18n.t("exp.save_button_label"),
-                        });
-                        onSave(ConsentMethod.SAVE, draftIds);
-                      }}
-                      className="fides-save-button"
-                      id="fides-save-button"
-                    />
-                  )}
-                  isInModal
-                  options={options}
-                />
-              );
-            }
-      }
+            )}
+            isInModal
+            options={options}
+          />
+        );
+      }}
     />
   );
 };
--- a/clients/fides-js/src/components/tcf/TcfOverlay.tsx
+++ b/clients/fides-js/src/components/tcf/TcfOverlay.tsx
@@ -34,7 +34,7 @@ import {
   FidesEventDetailsServingComponent,
   FidesEventTargetType,
 } from "../../lib/events";
-import { useNoticesServed } from "../../lib/hooks";
+import { useNoticesServed, useRetryableFetch } from "../../lib/hooks";
 import {
   DEFAULT_LOCALE,
   detectUserLocale,
@@ -79,7 +79,6 @@ import Overlay from "../Overlay";
 import { TCFBannerSupplemental } from "./TCFBannerSupplemental";
 import { TcfConsentButtons } from "./TcfConsentButtons";
 import TcfTabs from "./TcfTabs";
-import { useRetryableFetch } from "../../lib/hooks";
 
 const getAllIds = (
   modelList: TcfModels | Array<PrivacyNoticeWithPreference>,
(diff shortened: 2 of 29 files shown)


clients/fides-js/src/components/tcf/TcfOverlay.tsx
```diff
@@ -199,6 +200,38 @@ export const TcfOverlay = () => {
...
+          const userPrefs = buildUserPrefs(result, cookie);
+          const fullExperience: PrivacyExperience = {
+            ...result,
+            ...userPrefs,
+          };
+          window.Fides.experience = {
+            ...window.Fides.experience,
+            ...fullExperience,
+          };
+          window.Fides.experience.minimal_tcf = false;
```
tvandort: I wonder if there's a more expected pattern for this. It feels surprising to me that a hook would update the window object.

clients/fides-js/src/components/tcf/TcfTabs.tsx
```diff
@@ -37,6 +45,8 @@ const TcfTabs = ({
   ) => void;
   activeTabIndex: number;
   onTabChange: (tabIndex: number) => void;
+  isFullExperienceLoading: boolean;
+  isFullExperienceError: boolean;
```
tvandort: Good to proceed with this but I'd argue this might be better as something like:

`fullExperienceState: 'loading' | 'errored' | 'successful'` or something similar.

clients/fides-js/src/lib/hooks/useRetryableFetch.ts
```diff
@@ -0,0 +1,107 @@
...
+export const useRetryableFetch = <T>({
+  fetcher,
+  validator = () => true,
+  onSuccess,
+  config = {},
+  enabled = true,
+}: UseRetryableFetchOptions<T>): UseRetryableFetchResult<T> => {
+  const [data, setData] = useState<T | undefined>(undefined);
+  const [loading, setLoading] = useState(false);
+  const [error, setError] = useState(false);
```
tvandort: Might be better to combine these into a `fetchState: 'loading' | 'error' | 'success'` too. It seems to me that loading and error are mutually exclusive and this would communicate that relationship through the type system instead of through code.

clients/fides-js/src/lib/initialize.ts
```diff
@@ -408,14 +410,37 @@ export const initialize = async ({
...
     fides.experience?.privacy_notices,
     options.fidesConsentFlagType ?? undefined,
     options.fidesConsentNonApplicableFlagMode ?? undefined,
   );
 
+  const hasPrivacyNotices =
+    !!fides.experience?.non_applicable_privacy_notices ||
+    !!fides.experience?.privacy_notices;
+
+  // Any unknown consent values should return falsy values automatically
```
tvandort, line 424: Checking my understanding:

1. When an unknown consent flag is accessed we return `false`.
2. If an experience does not load that means all consent flags are "unknown".
3. Therefore, when experiences fail to load, this code is in play "by default" to say that the user has not given consent.

clients/fides-js/src/lib/hooks/useRetryableFetch.ts
```diff
@@ -0,0 +1,107 @@
...
+
+  const finalConfig = { ...DEFAULT_CONFIG, ...config };
+
+  const executeWithRetry = async (attempt = 1): Promise<void> => {
```
tvandort, line 55: Because retry functions can be fiddly I think we should add a unit test around this.

lucanovera (approved):
I tested the feature locally, retry works correctly with exponential backoff, error message displays, you can still opt in/out to all. Code changes look good. Approved!
Two minor comments:
- Can we remove the cursor pointer from the skeleton elements? (I could not figure out where that property is coming from lol)
- There's a TODO regarding translations, is there a follow up ticket for that?

clients/fides-js/src/components/tcf/TCFBannerSupplemental.tsx
```diff
@@ -31,7 +31,9 @@ export const TCFBannerSupplemental = ({
             </h2>
           </div>
         )}
-      <div className="fides-banner__content">
+      {/* adding tabindex because the content is scrollable so a11y requires it to be focusable */}
```
tvandort: I'm not familiar with WCAG or specific a11y rules but is there some documentation we can link here? e.g. a link to why one might need to do that? The comment is good but what I'm looking to do is socialize resources for others to learn more if possible.

tvandort (commented):
This is great, thank you for all these improvements!

A couple of thoughts I had: 

- Should the cursor become a pointer over the skeleton items?
- Would we be able to leave the save button but disable it instead of hiding it when the experience fails to load? I'm worried the buttons expanding might cause mis-taps

clients/fides-js/src/components/tcf/TcfOverlay.tsx
gilluminate: ideally this would come from the backend as part of the experience, but since it doesn't we're manually updating. `minimal_tcf` is only set on the minimal experience currently.

clients/fides-js/src/lib/hooks/useRetryableFetch.ts
gilluminate: While I understand where this feedback is coming from, this code is a fairly standard pattern for endpoint related react hooks (RTK uses, Tanstack Query uses, etc) that I don't want to mess with right now.

clients/fides-js/src/lib/initialize.ts
gilluminate: Mostly correct! If an experience does not load, we actually load in the legacy default consent values so this technically sets anything that's not a legacy consent value to false. All non-legacy consent flags are "unknown"

clients/fides-js/src/lib/hooks/useRetryableFetch.ts
tvandort: Tanstack Query internally represents the state of the query the way I'm recommending:
https://github.com/TanStack/query/blob/main/packages/query-core/src/types.ts#L618
https://github.com/TanStack/query/blob/main/packages/query-core/src/query.ts#L580

Managing a state-machine is much easier long term than managing a number of boolean states. 

I'm less familiar with RTK but it looks like the active state is also represented via an enumeration there: https://github.com/reduxjs/redux-toolkit/blob/master/packages/toolkit/src/query/core/apiState.ts#L71

I do accept that both libraries provide properties that allow library users to quickly check `.isLoading` or `.isErrored` but they do not manage their internal state that way. I think it would be reasonable for us to keep an `isLoading` and `isError` as part of the output interface here but I'd generally say we should make it a practice to not have that representation of the data internally.

gilluminate: This is what I'm referring to https://tanstack.com/query/v5/docs/framework/react/reference/useQuery and https://redux-toolkit.js.org/rtk-query/overview#use-hooks-in-components

tvandort: Those flags in these libraries are _derived_ from a state is what I mean. RTK derives that state here for instance: https://github.com/reduxjs/redux-toolkit/blob/master/packages/toolkit/src/query/core/apiState.ts#L108-L116

We should similarly derive that state so that the code is easy to use, while making it easier to update and read internally by adopting the same pattern of state management that RTK and React Query do internally, e.g. with a enumeration that represents the states the query can be in.

clients/fides-js/src/components/tcf/TcfOverlay.tsx
tvandort: Ah yeah... definitely. I guess since we're in a world where we don't get it from the backend what I was looking for is a safer pattern for updating a global variable. There might not be much of a better choice here however.

clients/fides-js/src/lib/hooks/useRetryableFetch.ts
gilluminate: ah, ok. I get what you are saying now. 👍

clients/fides-js/src/components/tcf/TcfOverlay.tsx
gilluminate: taking another look at this. Yes, the code here was copied from a non-hook source and this code likely doesn't belong in the hook. I'll update that. Thanks.

gilluminate: This is happening as a result of the hook, not in the hook. Maybe I'm not understanding what you are getting at. We get the full tcf data back, and we update the global experience that's already in window.Fides with the full experience.

gilluminate: looks like we replied over top of each other. 👍 On the same page though

clients/fides-js/src/lib/hooks/useRetryableFetch.ts
gilluminate: updated!

gilluminate:
@tvandort I've addressed your feedback, please review again?
@lucanovera & @tvandort I've remove the hover states from the skeletons and the save button is now retained but as disabled.

clients/fides-js/src/components/tcf/TcfOverlay.tsx
```diff
@@ -680,63 +696,63 @@ export const TcfOverlay = () => {
...
+            onAcceptAll={() => {
+              handleAcceptAll();
+              onClose();
+            }}
+            onRejectAll={() => {
+              handleRejectAll();
+              onClose();
+            }}
+            renderFirstButton={() => (
+              <Button
```
tvandort, line 727: image

Should the button outline still be black? IMO this makes the button look "focused". Is there a gray we can use here instead?

clients/fides-js/src/lib/initialize.ts
```diff
@@ -408,14 +410,37 @@ export const initialize = async ({
...
-  const consent = applyOverridesToConsent(
+  const consentObject = applyOverridesToConsent(
     fides.cookie.consent,
     fides.experience?.non_applicable_privacy_notices,
     fides.experience?.privacy_notices,
     options.fidesConsentFlagType ?? undefined,
     options.fidesConsentNonApplicableFlagMode ?? undefined,
   );
 
+  const hasPrivacyNotices =
```
tvandort, line 420: After any consent action (save, reject, accept) the consent object goes back to being a POJO. Is that expected?

clients/fides-js/src/components/tcf/TcfOverlay.tsx
gilluminate: This is the "Primary" color, and in our demo's case the primary color is set to our Minos color. It won't necessarily be this color for everyone. (Cookie house is some kind of red currently)

gilluminate: Oh, sorry. I'm trying to move too fast. I missed the word "outline" and thought you were talking about the black buttons... Yeah, I like your suggestion. I'll update the disabled styling.

clients/fides-js/src/lib/initialize.ts
gilluminate: I moved the Proxy logic to be part of the `applyOverridesToConsent` util to avoid this issue. We need to run consent through that every time already anyway.

gilluminate: That wasn't the correct change. `applyOverridesToConsent` is used in places where this causes a problem. I reverted that change and created a new `createConsentProxy` util to use only when we need it.

tvandort (approved):

PR #6396, opened 2025-07-28
ENG-878 Privacy Request cancellation/retry limit

thabofletcher:
Closes [ENG-878]

### Description Of Changes

Addresses known issues with tasks being marked as complete or celery worker tasks failing, but the privacy request still fails. In addition, I have added a retry count to the task requeue logic, since in my manual testing, these could requeue indefinitely

### Code Changes

* Adds a configuration and retry limit for task requeueing on privacy requests
* Cancellation logic for privacy requests when key task IDs are unable to be received from cache or when the retry limit is reached
* Fails privacy requests when interrupted tasks are detected
* Unit tests to verify it all works

### Steps to Confirm

1.  Still trying to find ways to prove these changes, since I am unable to reproduce (old out of memory exceptions are not reproducible). More manual testing required

### Pre-Merge Checklist

* [x] Issue requirements met
* [x] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [x] No followup issues
* Database migrations:
  * [x] No migrations
* Documentation:

[ENG-878]: https://ethyca.atlassian.net/browse/ENG-878?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/fides/api/models/privacy_request/privacy_request.py
```diff
@@ -1009,6 +1009,31 @@ def get_request_task_celery_task_ids(self) -> List[str]:
...
+        # Add all request task IDs
+        request_task_celery_ids = self.get_request_task_celery_task_ids()
+        task_ids.extend(request_task_celery_ids)
+
+        # Revoke all Celery tasks
+        for celery_task_id in task_ids:
+            logger.info("Revoking task {} for request {}", celery_task_id, self.id)
+            try:
+                # Use terminate=False to allow graceful shutdown if already running
+                celery_app.control.revoke(celery_task_id, terminate=False)
```
vcruces: I saw this logic was just moved, so no need to change anything here, just taking the chance to ask a couple of questions to better understand it:

1. I believe revoke accepts a list of task IDs, is there a specific reason to revoke them one by one? Maybe for better logging of failures?
2. Also, with terminate=False, if one of the tasks is already running and ends up completing successfully while the others are revoked, could that leave any inconsistencies behind?

Totally fine if this is already accounted for, just wanted to learn a bit more about how this works

src/fides/api/service/privacy_request/request_service.py
```diff
@@ -393,6 +404,196 @@ def _get_task_ids_from_dsr_queue(
...
+                db,
+                privacy_request,
+                f"Privacy request {privacy_request.id} exceeded max retry attempts "
+                f"({max_retries}), canceling tasks and setting to error state",
+            )
+            # Reset retry count since we're giving up (ignore errors here as it's cleanup)
+            try:
+                reset_privacy_request_retry_count(privacy_request.id)
+            except Exception as reset_exc:
+                logger.debug(
```
vcruces: Minor: Should we consider changing this to `logger.info` instead of `debug`, so we don’t miss it if the retry count reset fails? Some clients may have logging configured to only show INFO and above, so this might get lost otherwise

thabofletcher: That's a good catch, since it still indicates a problem communicating with the cache, I'm going to change it to a warning. And actually since the method itself logs the exception, I'm going to change it over in the `reset_privacy_request_retry_count` method since that is what would actually catch the exception, and remove the exception handling altogether here since that doesnt raise. I think cursor just got a little obsessed with wrapping things in try/except from the other code that was written

src/fides/api/models/privacy_request/privacy_request.py
thabofletcher: Honestly, I hadn't thought this through, since I was just extracting the existing functionality into a function so I could call it independently. I think your points are worth thinking through:

1. I can't imaging that the logging is worth the extra work here - its only called in one other place (from an endpoint to cancel the task), and I can't imagine any side affects from revoking all at once, so I will optimize this by revoking them all at once
2. I think its possible either way to leave inconsistencies - at least with terminate=False the individual tasks will either get completed or not, and then the privacy request would get marked and either requeued or not, and any incomplete tasks would get run again. I think with terminate=True, it would be more likely that the task itself would be marked as failed because it was killed, but then have some partial state that would prevent it from running again

So I think revoking them all together, but keeping terminate=False is the best choice for this code. Let me know if you have any other thoughts!

vcruces: Thanks Thabo! That makes sense!

adamsachs:
OK, i know i'm late to the game here (sorry about that!). these changes look great, but i wanted leave a brief "post-hoc" review here to talk about some potential next steps/ideas on how we could help share code between similar functionality for Helios monitor tasks over in fidesplus:
- the main parallel to look at is between `requeue_interrupted_tasks` [here](https://github.com/ethyca/fides/pull/6396/files#diff-e9b609a6bd089ccdf2444e0784f0f81e6fbcc0380601a583c854ae7a0e420557L398) and `detect_interrupted_tasks` in [fidesplus](https://github.com/ethyca/fidesplus/blob/main/src/fidesplus/api/service/discovery/monitor_tasks_mgmt.py#L47C5-L47C29) used for monitors
- some very low hanging fruit would be for us to use a more properly generic/shared `get_celery_task_ids_from_queue` function - we made [one in fidesplus](https://github.com/ethyca/fidesplus/blob/main/src/fidesplus/api/service/discovery/monitor_tasks_utils.py#L226) for monitor tasks that just doesn't live in the right place yet - i'm imaging we port this over to some shared location here in the `fides` OSS repo and use it across both use cases
- we do want to implement the sort of "retry" behavior in monitor tasks that you have here for DSR tasks: https://ethyca.atlassian.net/browse/ENG-1058
- for monitor tasks, we track the celery task ID as a [column](https://github.com/ethyca/fides/blob/main/src/fides/api/models/detection_discovery/monitor_task.py#L40) on the task record in our app DB, which allows us to reliably [retrieve that ID](https://github.com/ethyca/fidesplus/blob/main/src/fidesplus/api/service/discovery/monitor_tasks_mgmt.py#L113) without having to rely on the redis cache (which is ephemeral!), as is needed here for [privacy requests](https://github.com/ethyca/fides/pull/6396/files#diff-e9b609a6bd089ccdf2444e0784f0f81e6fbcc0380601a583c854ae7a0e420557R363) - @galvana and i have talked about how this would be a nice QOL improvement to the privacy request framework, so that we could more reliably retrieve that task ID.
- generally, as a next iteration for code sharing, i think we could shoot to:
    - abstract `requeue_interrupted_tasks` so that it covers the core logic that is shared between the two code paths, something like: acquiring a redis lock, checking for "in progress" entities (requests/monitor tasks) and retrieving their associated celery task IDs, retrieving queued celery task IDs, checking whether the "in progress" celery task IDs are either queued or 'in flight' in celery, retrying (and eventually canceling) the "in progress" entities as needed (respecting a configurable requeue count as you've done here)
    - we'd need the abstract function to leverage implementation-specific 'plugins' when executing some of those operations, since they'll differ (at least for now) between privacy requests and monitor tasks: checking for in-progress entities, retrieving celery task IDs associated with the core DB entity we're tracking, the actual retry and cancel operations
    - the privacy request codepath also has the additional nuance that it runs a similar sort of flow in the 'embedded' check for individual request tasks, but that seems like it could play well with us abstracting this out - i think it could effectively be a recursive call to the abstract `requeue_interrupted_tasks`?
    - this abstract 'requeue_interrupted_tasks` would need to live in a module that's properly shared between the two code paths (and potentially more code paths that want to leverage it!) - that could be in the [`models/worker_task.py` ORM model module](https://github.com/ethyca/fides/blob/main/src/fides/api/models/worker_task.py) that is already shared between the two code paths, or perhaps a new `service` module that's associated with this generic/abstract model class.


i think as part of https://ethyca.atlassian.net/browse/ENG-1058, when we implement the retry logic for monitor tasks, we could look to make at least some of these abstractions mentioned above. i think we should ticket a separate followup for having the privacy request framework store associated celery task IDs on DB records. there are also probably other further iterations that we can sketch out for code sharing between the two code paths, but i think those will become a bit more obvious as we get started toward the initial iteration outlined above.

what do you all think? hopefully my sketch there isn't too vague/abstract to understand 😬

PR #6403, opened 2025-07-30
Merge existing data categories into updated datasets

galvana:
Closes [ENG-563](https://ethyca.atlassian.net/browse/ENG-563)

### Description Of Changes

Update to preserve data categories from existing SaaS datasets during the connector auto-updates.

### Code Changes

* Added new `merge_dataset_dicts` util to merge existing datasets into new connector template datasets
* Updated `upsert_dataset_config_from_template` to use `merge_dataset_dicts`

### Steps to Confirm

1. Create a SaaS integration
2. Change the data categories for any field in the Saas dataset (can only do this via the API)
3. Upload a new connector template and restart Fides
4. The existing integration should have changes from the new SaaS dataset, but keep the data categories you changed

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:

[ENG-563]: https://ethyca.atlassian.net/browse/ENG-563?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/ops/service/connectors/saas/test_upsert_dataset_config_merge.py
```diff
@@ -0,0 +1,357 @@
...
+        assert new_config.ctl_dataset.name == "New Test Dataset"
+
+        # Basic check that collections exist and have data
+        assert len(new_config.ctl_dataset.collections) > 0
+
+    def test_real_connector_preserves_data_categories_on_template_update(
+        self,
+        db: Session,
+    ):
+        """Test that dataset merge preserves custom data categories when template is updated"""
```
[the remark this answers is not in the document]
galvana: The test suite automatically clears the database between tests. There's no need to add a manual cleanup for each test.

src/fides/api/service/connectors/saas/connector_registry_service.py
```diff
@@ -338,6 +339,194 @@ def create_connection_config_from_template_no_save(
...
+    """
+    if "." in property_path:
+        # Handle nested property with dot notation
+        existing_value = _get_nested_value(existing_item, property_path)
+        if existing_value is not None:
+            _set_nested_value(merged_item, property_path, existing_value)
+    else:
+        # Handle simple property
+        if property_path in existing_item:
+            merged_item[property_path] = existing_item[property_path]
```
Linker44:
```suggestion
      existing_value = pydash.get(existing_item, property_path)
      if existing_value is not None:
            pydash.set_(merged_item, property_path, existing_value)
```

Linker44: we should be able to just use pydash to simplify this logic since it can use dot notation and also handle simple paths.

Linker44:
_i have a question regarding this implementation:_

Wouldn't this prevent ethyca from also updating the data_categories of an integration?

I understand that the desire behaviour is for example if we have a fides-provided template and a customer makes changes to its dataset through our endpoint then upon restart their edited data_categories wouldnt be wiped if a new version of the integration is found.

But what happens if for example a user has an integration set up with our fides provided template, we (the ethyca engineers) notice that some fields should be showing but are not due to wrongly categorized fields. We bump the version and when the server restarts this code would make it so that our changes to the data_categories would be overwritten by the older version of the dataset?

src/fides/api/service/connectors/saas/connector_registry_service.py
```diff
@@ -338,6 +339,194 @@ def create_connection_config_from_template_no_save(
...
+    """
+    Recursively merge nested fields while preserving specified properties.
+
+    Args:
+        merged_field: The field being merged into
+        existing_field: The existing field to preserve from
+        preserved_properties: List of properties to preserve (supports dot notation)
+    """
+    # Handle nested fields - we need to preserve existing nested fields even if template doesn't have them
+    if "fields" in existing_field:
```
Linker44:
```suggestion
    if existing_field.get("fields"):
```
Fields that dont have nested fields store the "fields" value as None, in line 497 trying to iterate through a none value will result in an error.

Linker44: we need to change the merge function logic to take this into account. as there are other lines where this could cause an issue.

ex: line 508

PR #6407, opened 2025-07-30
Add option to FidesJS to show modal by default

jpople:
Closes ENG-995

### Description Of Changes

Adds a boolean option `fides_default_show_modal` that, when enabled for an experience with a modal, opens the modal on page load.  Can be combined with `fides_modal_default_view` on TCF to allow "deep linking" to a specific tab in the TCF modal.

### Steps to Confirm

1. Configure a TCF or banner/modal experience, then view in the demo page.
1. Modal should not be open by default (existing behavior)
1. Pass `fides_default_show_modal=true` as a query param, then reload
1. Modal should be open by default
1. On TCF, pass `fides_default_show_modal=true` with a `fides_modal_default_view` param (e.g. `fides_modal_default_view=%2Ftcf%2Fvendors`)
1. Modal should be open and set to correct tab by default

### Pre-Merge Checklist

* [X] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:


clients/fides-js/src/docs/fides-options.ts
```diff
@@ -297,4 +297,7 @@ export interface FidesOptions {
    * Defaults to `undefined`.
    */
   fides_modal_default_view?: string;
+
+  // Whether to show the consent modal immediately upon initialization
```
gilluminate: Can we update this to `jsdoc` format so it shows up in the Markdown for customers?

clients/fides-js/src/components/Overlay.tsx
```diff
@@ -188,6 +188,12 @@ const Overlay: FunctionComponent<Props> = ({
     };
   }, [experience, handleOpenModal, options.fidesEmbed]);
 
+  useEffect(() => {
+    if (options.fidesDefaultShowModal) {
+      handleOpenModal();
+    }
+  }, [options.fidesDefaultShowModal, handleOpenModal]);
```
gilluminate: what if `options.fidesEmbed` is set to `true`, will that cause an issue here?

gilluminate: ~I don't love that this happens along with rendering the banner, can we use `options.fidesDefaultShowModal` to also `setDisableBanner` to false earlier in the code?~

clients/fides-js/src/lib/consent-constants.ts
```diff
@@ -123,6 +123,12 @@ export const FIDES_OVERRIDE_OPTIONS_VALIDATOR_MAP: FidesOverrideValidatorMap[] =
       overrideKey: "fides_modal_default_view",
       validationRegex: /^\/tcf\/(purposes|features|vendors)$/,
     },
+    {
+      overrideName: "fidesDefaultShowModal",
+      overrideType: "boolean",
+      overrideKey: "fides_default_show_modal",
+      validationRegex: /^(true|false)$/,
+    },
```
gilluminate: Kind of nit-picky, but Tom has taught me the benefit of avoiding boolean as a setting value. Can we change this to be something like `fidesModalDisplay` with values like `"immediate"|"banner_click"` where the default is banner click? If we'd have done this with fides_embed originally, we could just add an option to that instead of creating a whole new override. It becomes much more extensible.

clients/fides-js/src/lib/consent-utils.ts
```diff
@@ -260,6 +261,10 @@ export const shouldResurfaceBanner = (
   if (!isPrivacyExperience(experience)) {
     return false;
   }
+  // Never surface banner if modal is set to show
+  if (options.fidesDefaultShowModal) {
+    return false;
+  }
```
gilluminate: Oh, I see now, this does disable the banner earlier 👍

clients/fides-js/src/components/Overlay.tsx
jpople: @gilluminate This does cause `fides_embed` to render layer 2 in the embed immediately as opposed to showing the banner embedded, is that problematic?  I was under the impression the embed was only used for embedding layer 2 anyway.  If that's okay then this works fine when `fides_embed` is true.

gilluminate: Oh, that's right, I forgot that's how it worked. Sounds like this is working as expected then.

clients/fides-js/src/lib/consent-constants.ts
```diff
@@ -123,6 +123,12 @@ export const FIDES_OVERRIDE_OPTIONS_VALIDATOR_MAP: FidesOverrideValidatorMap[] =
       overrideKey: "fides_modal_default_view",
       validationRegex: /^\/tcf\/(purposes|features|vendors)$/,
     },
+    {
+      overrideName: "fidesModalDisplay",
+      overrideType: "string",
+      overrideKey: "fides_modal_display",
+      validationRegex: /^(immediate|banner_click)$/,
```
gilluminate: sorry, my bad. Can we change "banner_click" to something else? I just realized that some experiences have no banner and it's just the Manage Preferences link on the page. "manage_preferences_click" might be too long? Or maybe just "default" for now?

gilluminate (approved):
Looks good! ship it

PR #6416, opened 2025-08-01
Protect against multiple instantiations of FidesJS and GPP

gilluminate:
Closes [ENG-1089]

### Description Of Changes

Adds protection against multiple FidesJS script loading on the same page, which can cause conflicts and unexpected behavior. By default, FidesJS will now detect if it's already loaded and abort execution with a console.error message. Includes a new configuration option to allow multiple loading for edge cases where customers need this unsupported behavior.

Also updates the GPP extension to prevent duplicate initialization when Fides is already initialized.

### Code Changes

* Added `fides_unsupported_repeated_script_loading` configuration option with "disabled" (default) and "enabled_acknowledge_not_supported" values
* Implemented multiple loading detection check in the fides-js API route that aborts execution if Fides already exists
* Updated GPP extension event listener to check `window.Fides?.initialized` before initializing
* Updated debugger initialization to prevent overwriting existing debugger instance when Fides.init() is used more than once.
* Modified script loading order in demo HTML to load GPP extension before main Fides script so that listeners are in place first.
* Added comprehensive Cypress tests for multiple loading scenarios including script removal/re-addition

### Steps to Confirm

1. Load the demo page at `/fides-js-demo.html` and verify normal functionality
2. Manually inject a second fides.js script tag and confirm error message appears in console
3. Test with `?fides_unsupported_repeated_script_loading=enabled_acknowledge_not_supported` parameter and verify no error when loading multiple scripts
5. Verify GPP does not add multiple listeners any time `Fides.init()` is used in the console. Verify using `getEventListeners(window)` in the console.

### Pre-Merge Checklist

* [x] Issue requirements met
* [x] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [x] No followup issues
* Database migrations:
  * [x] No migrations
* Documentation:

[ENG-1089]: https://ethyca.atlassian.net/browse/ENG-1089?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


clients/privacy-center/pages/api/fides-js.ts
```diff
@@ -392,8 +397,24 @@ export default async function handler(
...
+      typeof window !== "undefined" &&
+      !!window.Fides &&
+      !!window.Fides.options &&
+      window.Fides.options.fidesUnsupportedRepeatedScriptLoading !==
+        "enabled_acknowledge_not_supported"
+    ) {
+      console.error("${multipleLoadingMessage}");
+      return;
+    }
+  `.replace(/(\r\n|\n|\r)/gm, ""); // remove newlines to avoid sourcemap issues!
```
NevilleS: This needs to move into `fides.js` code itself; it's not _guaranteed_ that we always fetch FidesJS from the Privacy Center, we do bundle FidesJS independently in some cases.

NevilleS (changes requested):

clients/privacy-center/pages/api/fides-js.ts
gilluminate: Good reminder! I'll make that change.

clients/fides-js/.eslintignore
```diff
@@ -1,6 +1,6 @@
 node_modules
 dist
-jest.config.js
+**/*.js
```
NevilleS: Seems a *bit* broad... 😁

clients/fides-js/custom-rollup-plugins/prevent-double-load.js
```diff
@@ -0,0 +1,60 @@
+/**
+ * Rollup plugin to prevent double loading of Fides scripts
+ */
+export const preventDoubleLoad = (globalName) => {
+  const multipleLoadingMessage = `${globalName} detected that it was already loaded on this page, aborting execution! See https://www.ethyca.com/docs/dev-docs/js/troubleshooting for more information.`;
+
+  // Common replacement pattern for all UMD wrappers
+  const doubleLoadCheck = `if(typeof ${globalName}!=="undefined" && ${globalName}.options && ${globalName}.options.fidesUnsupportedRepeatedScriptLoading !== "enabled_acknowledge_not_supported"){console.error("${multipleLoadingMessage}");return;}`;
```
NevilleS: I'd recommend changing this check a bit. It currently expects `Fides.options` to exist but that seems unnecessary.

I'd do something more akin to this:
```typescript
if (typeof Fides !== "undefined") {
   // Bad news! Multiple loading detected!
   // In the special case where this is allowed, continue, but otherwise...
   if (Fides.options && Fides.options.fidesUnsupportedRepeatedScriptLoading == "enabled_acknowledge_not_supported") {
     // no-op, continue
  } else {
    console.error("bad!");
  }
}
```

Does this make sense? I am struggling to communicate this

clients/fides-js/custom-rollup-plugins/prevent-double-load.js
```diff
@@ -0,0 +1,60 @@
...
+              chunk.code = replaced;
+              wasReplaced = true;
+              break;
+            }
+          }
+
+          if (!wasReplaced) {
+            console.warn(
+              `[prevent-double-load] Could not find UMD wrapper pattern in ${fileName} to inject double-load prevention code`,
+            );
```
NevilleS: Can this be more explosive? This warning will just scroll by in the build logs if these logs don't match right, but we want this to always match something. Maybe throw an error to blow up the build?

clients/fides-js/custom-rollup-plugins/prevent-double-load.js
```diff
@@ -0,0 +1,60 @@
...
+ * Rollup plugin to prevent double loading of Fides scripts
+ */
+export const preventDoubleLoad = (globalName) => {
+  const multipleLoadingMessage = `${globalName} detected that it was already loaded on this page, aborting execution! See https://www.ethyca.com/docs/dev-docs/js/troubleshooting for more information.`;
+
+  // Common replacement pattern for all UMD wrappers
+  const doubleLoadCheck = `if(typeof ${globalName}!=="undefined" && ${globalName}.options && ${globalName}.options.fidesUnsupportedRepeatedScriptLoading !== "enabled_acknowledge_not_supported"){console.error("${multipleLoadingMessage}");return;}`;
+
+  // Define UMD wrapper patterns to match
+  const umdPatterns = [
```
NevilleS: This is kinda terrifying, but OK

clients/fides-js/custom-rollup-plugins/prevent-double-load.js
```diff
@@ -0,0 +1,60 @@
...
+    const replaced = code.replace(pattern.regex, replacement);
+    return replaced !== code ? replaced : null;
+  };
+
+  return {
+    name: "prevent-double-load",
+    generateBundle(options, bundle) {
+      Object.keys(bundle).forEach((fileName) => {
+        const chunk = bundle[fileName];
+        if (chunk.type === "chunk" && options.format === "umd") {
```
NevilleS: This seems like a subtle footgun for us in the future... if we ever ship a rollup config that doesn't use `umd` but instead switches over to `iife`, for example, this won't inject this check anymore?

clients/fides-js/rollup.config.mjs
```diff
@@ -63,6 +65,7 @@ const fidesScriptPlugins = (stripDebugger = false) => [
     preventAssignment: true,
     include: ["src/lib/init-utils.ts"],
   }),
+  !IS_TEST && preventDoubleLoad(GLOBAL_NAME),
```
NevilleS: This is a clever plugin, but... would `options.intro` work? It seems to potentially do what you want: https://rollupjs.org/configuration-options/#output-intro-output-outro

This would allow you to inject some code into the bundle to run at the very start?

NevilleS (commented):
Some notes!

clients/fides-js/.eslintignore
gilluminate: eh, I'll double check it. I think eslint is always going to have a problem with .js files.

clients/fides-js/custom-rollup-plugins/prevent-double-load.js
gilluminate: yes! Good catch.

clients/fides-js/rollup.config.mjs
gilluminate: I'll try that. Claude did suggest this also, but was strongly recommending a plugin instead. However, it wasn't thinking about the concerning `umdPatterns` at the time (wasn't taking in to account minification). I'll play around with this.

clients/fides-js/custom-rollup-plugins/prevent-double-load.js
gilluminate: I double checked, and our minified `fides.js` is already using optional chaining operators (`foo?.bar`) so I should be safe to just use that here as well.

clients/fides-js/rollup.config.mjs
gilluminate: I'm able to use banner and footer here no problem. I couldn't use intro because it's still inside the UMD wrapper that sets the global.

clients/fides-js/rollup.config.mjs
```diff
@@ -169,12 +172,21 @@ SCRIPTS.forEach(({ name, gzipErrorSizeKb, gzipWarnSizeKb, isExtension }) => {
...
-        name: isExtension ? undefined : "Fides",
+        name: isExtension ? undefined : GLOBAL_NAME,
         format: isExtension ? "es" : "umd",
         sourcemap: IS_DEV ? "inline" : false,
         amd: {
           define: undefined, // prevent the bundle from registering itself as an AMD module, even if an AMD loader (like RequireJS) is present on the page. This allows FidesJS to use Rollup's `umd` format to support both `iife` and `cjs` modules, but excludes AMD.
         },
+        banner: isExtension
+          ? undefined
+          : `(function(){if(typeof ${GLOBAL_NAME}==="undefined" || (${GLOBAL_NAME}.options && ${GLOBAL_NAME}.options.fidesUnsupportedRepeatedScriptLoading === "enabled_acknowledge_not_supported")) {console.log('worked once');`,
```
NevilleS: What's the console.log for?

clients/fides-js/rollup.config.mjs
```diff
@@ -169,12 +172,21 @@ SCRIPTS.forEach(({ name, gzipErrorSizeKb, gzipWarnSizeKb, isExtension }) => {
...
+        footer: isExtension
+          ? undefined
+          : `} else {
+          console.log(${GLOBAL_NAME}.options);
```
NevilleS: What's this console.log for as well?

clients/fides-js/rollup.config.mjs
```diff
@@ -169,12 +172,21 @@ SCRIPTS.forEach(({ name, gzipErrorSizeKb, gzipWarnSizeKb, isExtension }) => {
...
+        name: isExtension ? undefined : GLOBAL_NAME,
         format: isExtension ? "es" : "umd",
         sourcemap: IS_DEV ? "inline" : false,
         amd: {
           define: undefined, // prevent the bundle from registering itself as an AMD module, even if an AMD loader (like RequireJS) is present on the page. This allows FidesJS to use Rollup's `umd` format to support both `iife` and `cjs` modules, but excludes AMD.
         },
+        banner: isExtension
+          ? undefined
+          : `(function(){if(typeof ${GLOBAL_NAME}==="undefined" || (${GLOBAL_NAME}.options && ${GLOBAL_NAME}.options.fidesUnsupportedRepeatedScriptLoading === "enabled_acknowledge_not_supported")) {console.log('worked once');`,
+        footer: isExtension
```
NevilleS: I'd feel _much_ more comfortable if we only added a "banner" instead of using banner/footer to wrap the entire script in an if statement. This seems to work, but it feels really fragile.

A "banner" version of this would be like
```
// FidesJS
// Before loading, check if this script has already loaded
if (...) {
  console.error(...)
  return;
}
```

NevilleS (changes requested):
I'm being picky

clients/fides-js/rollup.config.mjs
gilluminate: d'oh!!

clients/fides-js/rollup.config.mjs
gilluminate: left over debugging. Thanks for catching.

clients/fides-js/rollup.config.mjs
gilluminate: great suggestion

gilluminate: Taking another look at this, I remember why wrapping it is necessary...
the `return` can't be used outside of a function (throws errors otherwise) and if we're not using Privacy Center's `(function(){` wrapper, then that becomes the case. You can't just "exit" javascript like this.

NevilleS: Throw an error?

NevilleS: Since we're doing some pretty clever bundling, I think I prefer the DX of a thrown exception here, instead of a graceful if/else. I don't trust the banner+footer to not surprise us in so d very odd way months from now

gilluminate: makes sense. I'll test that

clients/fides-js/package.json
```diff
@@ -12,16 +12,19 @@
   ],
   "exports": {
     ".": {
+      "module": "./dist/fides.mjs",
```
gilluminate, line 15: needed for Cypress

clients/privacy-center/public/fides-js-components-demo.html
```diff
@@ -282,9 +282,11 @@ <h1>fides.js demo page</h1>
       <script>
         setTimeout(() => {
           const app = document.getElementById("render-embed-here");
-          const container = document.createElement("div");
-          container.id = "fides-embed-container";
-          app.appendChild(container);
+          if (app) {
```
gilluminate, line 285: one of our Cypress tests is overwriting the `<body>` which causes this to error out, so I'm adding a safeguard. Good idea anyway.

clients/privacy-center/cypress/support/commands.ts
```diff
@@ -325,5 +333,15 @@ declare global {
   }
 }
 
+Cypress.Commands.add("handleFidesAlreadyLoadedException", () => {
```
NevilleS: nit:
```suggestion
Cypress.Commands.add("expectFidesAlreadyLoadedException", () => {
```

clients/fides-js/rollup.config.mjs
```diff
@@ -180,13 +180,7 @@ SCRIPTS.forEach(({ name, gzipErrorSizeKb, gzipWarnSizeKb, isExtension }) => {
...
         banner: isExtension
           ? undefined
-          : `(function(){if(typeof ${GLOBAL_NAME}==="undefined" || (${GLOBAL_NAME}.options && ${GLOBAL_NAME}.options.fidesUnsupportedRepeatedScriptLoading === "enabled_acknowledge_not_supported")) {console.log('worked once');`,
-        footer: isExtension
-          ? undefined
-          : `} else {
-          console.log(${GLOBAL_NAME}.options);
-          console.error("${multipleLoadingMessage}");
-        }})()`,
+          : `if(typeof ${GLOBAL_NAME}!=="undefined" && ${GLOBAL_NAME}.options && ${GLOBAL_NAME}.options.fidesUnsupportedRepeatedScriptLoading!=="enabled_acknowledge_not_supported") {throw new Error("${multipleLoadingMessage}");}`,
```
NevilleS: nit: if `Fides.options` is undefined, this won't throw the error and will allow multiple loads. That'd be a bug (Fides.options should be defined whenever Fides is, right?) so it's not horrible, but it's surprising.

You'd have to invert the conditional which is harder to parse, but maybe:
```javascript
if ( typeof Fides !== undefined &&
      !(Fides.options && 
        Fides.options.fidesUnsupportedRepeatedScriptLoading == "enabled_acknowledge_not_supported
        )
  ) { ... }
```

clients/fides-js/rollup.config.mjs
```diff
@@ -180,13 +180,7 @@ SCRIPTS.forEach(({ name, gzipErrorSizeKb, gzipWarnSizeKb, isExtension }) => {
         },
         banner: isExtension
```
NevilleS, line 182: You should add a comment here, for future you.
```suggestion
        // Use the "banner" option to prepend a defensive check into the code to guard against loading FidesJS multiple times on the same page  
        banner: isExtension
```

clients/fides-js/rollup.config.mjs
gilluminate: I got confused here. I thought we wanted to allow if Fides.options was undefined. I'll update.

PR #6421, opened 2025-08-05
DB models for custom taxonomies

galvana:
Closes [ENG-1073](https://ethyca.atlassian.net/browse/ENG-1073)

### Description Of Changes

- Introduces first-class DB models to support custom taxonomies and cross-taxonomy tagging:
  - `Taxonomy`: defines a taxonomy type and its allowed targets via `applies_to`
  - `TaxonomyAllowedUsage`: declares which target types/taxonomies a taxonomy can apply to
  - `TaxonomyElement`: hierarchical elements within a taxonomy (supports parent/child)
  - `TaxonomyUsage`: records applications of one taxonomy’s element to another element
- Adds Alembic migration to create new tables, indexes, FKs, and seed system-managed legacy taxonomies (`data_categories`, `data_uses`, `data_subjects`, `system_groups`).
- Adds a `TaxonomyService` with centralized validation and hierarchy activation/deactivation behavior, and a base handler abstraction with a legacy handler implementation.
- Routes legacy taxonomy endpoints (`data_category`, `data_use`, `data_subject`) through the new `TaxonomyService` to unify validation and reactivation behavior.
- Comprehensive tests for the new models, service behavior, and constraints.

Potential caveats:
- Removing an `applies_to` target that’s currently referenced by `taxonomy_usage` will fail due to FK constraints; usages must be removed first.
- Deleting a parent `TaxonomyElement` with existing children is restricted at the DB level; delete or move children first.
- The service layer currently supports legacy taxonomies only for API operations; CRUD for fully custom taxonomies/elements/usages is not yet exposed via public endpoints.

### Code Changes

- New models:
  - `src/fides/api/models/taxonomy.py` (adds `Taxonomy`, `TaxonomyAllowedUsage`, `TaxonomyElement`, `TaxonomyUsage`)
- Migration:
  - `src/fides/api/alembic/migrations/versions/3baf42d251a6_add_generic_taxonomy_models.py`
- API changes:
  - `src/fides/api/api/v1/endpoints/generic_overrides.py` (create/update for data uses/categories/subjects now use `TaxonomyService`; legacy inline validation removed)
- DI and model registration:
  - `src/fides/api/service/deps.py` (new `get_taxonomy_service`)
  - `src/fides/api/db/base.py` (imports new models)
- New service layer:
  - `src/fides/service/taxonomy/__init__.py`
  - `src/fides/service/taxonomy/handlers/base.py`
  - `src/fides/service/taxonomy/handlers/legacy_handler.py`
  - `src/fides/service/taxonomy/taxonomy_service.py`
  - `src/fides/service/taxonomy/utils.py`
- Dataset updates:
  - `.fides/db_dataset.yml` (alignments for taxonomy-related filtering)
- Tests:
  - `tests/ops/models/test_taxonomy.py`
  - `tests/service/test_taxonomy_service.py`

### Steps to Confirm

1. Run targeted tests for models and service:
   - `docker exec fides bash -c "pytest -q tests/ops/models/test_taxonomy.py"`
   - `docker exec fides bash -c "pytest -q tests/service/test_taxonomy_service.py"`
2. Exercise legacy taxonomy endpoints (requires backend running and auth):
   - POST `/api/v1/data_category` with a new category; expect 201 and generated/validated key behavior.
   - PUT `/api/v1/data_category` to toggle `active` and verify parent/child cascade semantics via subsequent GETs.
   - Repeat for `/api/v1/data_use` and `/api/v1/data_subject`.
3. Database checks after migration:
   - Verify tables exist: `taxonomy`, `taxonomy_allowed_usage`, `taxonomy_element`, `taxonomy_usage`.
   - Confirm seeded rows exist for the legacy taxonomies in `taxonomy`.
4. Confirm FK behavior:
   - Attempt to remove an `applies_to` target that’s in use (should fail).
   - Attempt to delete a `TaxonomyElement` with children (should be restricted).

### Pre-Merge Checklist
- [ ] Issue requirements met
- [ ] All CI pipelines succeeded
- [ ] `CHANGELOG.md` updated
  - [ ] Add a `db-migration` label to the entry
  - [ ] Add a `high-risk` label if you believe the FK constraints or activation/deactivation logic could have performance/regression risk
  - [ ] Updates unreleased work already in Changelog, no new entry necessary
- Followup issues:
  - [ ] Followup issues created
  - [ ] No followup issues
    - Suggested follow-ups:
      - Expose CRUD endpoints for custom taxonomies/elements/usages
      - Document new tables, relationships, and service behaviors
      - UI support for managing custom taxonomies
- Database migrations:
  - [ ] Ensure that your downrev is up to date with the latest revision on `main`
  - [ ] Ensure that your `downgrade()` migration is correct and works
    - Note: Downgrade drops all four new tables and seeded data
  - [ ] No migrations
- Documentation:
  - [ ] Documentation complete, [PR opened in fidesdocs](https://github.com/ethyca/fidesdocs/pulls)
  - [ ] Documentation [issue created in fidesdocs](https://github.com/ethyca/fidesdocs/issues/new/choose)
  - [ ] If any new client scopes were created, update public-facing documentation that references our scope registry
  - [ ] No documentation updates required

[ENG-1073]: https://ethyca.atlassian.net/browse/ENG-1073?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


galvana:
@greptileai

src/fides/api/api/v1/endpoints/generic_overrides.py
```diff
@@ -354,133 +346,7 @@ def clean_datasets(
     )
 
 
-def activate_taxonomy_parents(
```
erosselli, line 357: thanks for moving all these to a service class! 😌

src/fides/api/api/v1/endpoints/generic_overrides.py
```diff
@@ -354,133 +346,7 @@ def clean_datasets(
...
-        if disabled_resource_with_name:
-            data.active = True
-            activate_taxonomy_parents(disabled_resource_with_name, db)
-            return validate_and_update_taxonomy(
-                db, disabled_resource_with_name, validation_schema, data
-            )
-        return validate_and_create_taxonomy(db, model, validation_schema, data)
-
-    return validate_and_create_taxonomy(db, model, validation_schema, data)
+# Legacy validation functions removed - functionality moved to TaxonomyService
```
erosselli: I don't think we need this comment -- it reads like cursor explaining its changes lol

src/fides/api/models/taxonomy.py
```diff
@@ -0,0 +1,238 @@
...
+    SYSTEM = "system"
+    DATASET = "dataset"
+    COLLECTION = "collection"
+    FIELD = "field"
+    PRIVACY_DECLARATION = "privacy_declaration"
+    TAXONOMY = "taxonomy"  # For taxonomy-to-taxonomy relationships
+
+
+class Taxonomy(Base, FidesBase):
+    """The SQL model for taxonomy resources."""
```
erosselli: can we maybe add some info to the docstring about how the out-of-the-box taxonomies relate to this model? so that new devs (or us in 2 years lol) running into this code don't have to search around for an explanation

src/fides/api/models/taxonomy.py
```diff
@@ -0,0 +1,238 @@
...
+
+    id = Column(
+        String(255), nullable=False, index=False, default=FidesBase.generate_uuid
+    )
```
erosselli: usually we inherit the `id` column from `FidesBase` right? is there no issue with re-defining the id column?

src/fides/api/models/taxonomy.py
```diff
@@ -0,0 +1,238 @@
...
+
+    _allowed_usages: RelationshipProperty[List[TaxonomyAllowedUsage]] = relationship(
```
erosselli, line 55: Why do we define this as private? I imagine it's likely that code will do things like `taxonomy.allowed_usages.filter(...)` right?

src/fides/api/models/taxonomy.py
erosselli: also don't we need a `name` column?

src/fides/api/models/taxonomy.py
```diff
@@ -0,0 +1,238 @@
...
+    # is handled through the TaxonomyAllowedUsage model
+    # Updates should be done through the create/update methods
+    applies_to: List[str] = association_proxy("_allowed_usages", "target_type")
+
+    @classmethod
+    def create(
+        cls: Type["Taxonomy"],
+        db: Session,
+        *,
+        data: Dict[str, Any],
```
erosselli, line 78: sometimes I wonder if these should be TypedDicts or something a bit more strictly typed (or even just keyword args for the create method). It'd make it easier to know what keys the `create` method expects without looking at the code

src/fides/api/models/taxonomy.py
```diff
@@ -0,0 +1,238 @@
...
+            TaxonomyAllowedUsage(  # type: ignore
+                source_taxonomy_key=taxonomy.fides_key, target_type=target_type
+            )
+            for target_type in set(applies_to)
+        ]
+        taxonomy._allowed_usages = allowed_usages  # pylint: disable=protected-access
+
+        return cls.persist_obj(db, taxonomy)
+
+    def update(self, db: Session, *, data: Dict[str, Any]) -> "Taxonomy":
```
erosselli, line 99: what's the difference between `update` or `save` ? I think maybe we should consider overriding the `save` method too?

src/fides/api/models/taxonomy.py
```diff
@@ -0,0 +1,238 @@
...
+        taxonomy: Taxonomy = super().create(db=db, data=data, check_name=check_name)
+
+        # Create TaxonomyAllowedUsage objects for each applies_to value
+        allowed_usages = [
+            TaxonomyAllowedUsage(  # type: ignore
+                source_taxonomy_key=taxonomy.fides_key, target_type=target_type
+            )
+            for target_type in set(applies_to)
+        ]
+        taxonomy._allowed_usages = allowed_usages  # pylint: disable=protected-access
```
erosselli: see my comment about the private relationship, we'll end up with this `# pylint: disable=protected-access` everywhere. I don't think it's a good pattern to have here?

src/fides/api/models/taxonomy.py
```diff
@@ -0,0 +1,238 @@
...
+                f"Cannot create taxonomy '{fides_key}'. This is a taxonomy managed by the system."
+            )
+        applies_to = data.pop("applies_to", [])
+
+        # Create the taxonomy
+        taxonomy: Taxonomy = super().create(db=db, data=data, check_name=check_name)
+
+        # Create TaxonomyAllowedUsage objects for each applies_to value
+        allowed_usages = [
+            TaxonomyAllowedUsage(  # type: ignore
```
erosselli: what's up with `# type: ignore` here?

src/fides/api/models/taxonomy.py
```diff
@@ -0,0 +1,238 @@
...
+    )
+
+    source_taxonomy: RelationshipProperty[Taxonomy] = relationship(
+        "Taxonomy",
+        back_populates="_allowed_usages",
+    )
+
+    source_taxonomy_key = Column(
+        String,
+        ForeignKey("taxonomy.fides_key", ondelete="CASCADE"),
```
erosselli, line 179: what's our usual criteria for choosing whether FK constraints are done on the `id` field or on the `fides_key` field?

src/fides/api/models/taxonomy.py
```diff
@@ -0,0 +1,238 @@
...
+        index=True,
+    )
+    target_type = Column(
+        String, primary_key=True
+    )  # Can be "system", "dataset", OR a taxonomy key like "data_categories"
+
+    __table_args__ = (
+        # Simple two-column index for queries
+        Index("ix_allowed_usage_lookup", "source_taxonomy_key", "target_type"),
+    )
```
erosselli: do you need this index given that you've defined `(source_taxonomy_key, target_type)` as a composite PK with `index=True` ?

src/fides/service/taxonomy/handlers/base.py
```diff
@@ -0,0 +1,43 @@
...
+class TaxonomyHandler(ABC):
+    """Abstract handler for taxonomy CRUD operations."""
+
+    def __init__(self, db: Session):
+        self.db = db
+
+    @abstractmethod
+    def get_model(self, taxonomy_type: str) -> Type:
+        pass
+
```
erosselli: I think we should be able to type this more strictly; maybe we can define a `TaxonomyType` union type or a base class that other types inherit from? 

Actually what I'd _really_ like to know is if we can actually use overloads to do something like the following,(ab?)using the fact that a string literal is actually a type:

```py
@overloads
def get_model(self, taxonomy_type: Literal['data_category']) -> DataCategory:
@overloads
def get_model(self, taxonomy_type: Literal['data_use'])->DataUse:
.... other overloads
```

so if you call `get_model('data_category')` the type checker will actually know the return type is `DataCategory`

erosselli: idk if this would be done here or below in the concrete class

src/fides/service/taxonomy/handlers/legacy_handler.py
```diff
@@ -0,0 +1,98 @@
...
+    def create_element(
+        self, taxonomy_type: str, element_data: Dict
+    ) -> Union[DataCategory, DataUse, DataSubject]:
+        model_class = self.get_model(taxonomy_type)
+        element = model_class.create(self.db, data=element_data)
+        return element
+
+    def update_element(
+        self, taxonomy_type: str, fides_key: str, element_data: Dict
+    ) -> Optional[Union[DataCategory, DataUse, DataSubject]]:
```
erosselli, line 60: same comment about using overloading here hehe

src/fides/service/taxonomy/handlers/legacy_handler.py
```diff
@@ -0,0 +1,98 @@
...
+from fides.api.models.sql_models import (  # type:ignore[attr-defined]
+    DataCategory,
+    DataSubject,
+    DataUse,
+)
+
+from .base import TaxonomyHandler
+
+
+class LegacyTaxonomyHandler(TaxonomyHandler):
```
erosselli, line 18: so instead of adding the `taxonomy_type` as an argument to all the methods, wouldn't it make more sense to instantiate a `LegacyTaxonomyHandler` for a specific taxonomy type, i.e `TaxonomyHandler(db, taxonomy_type='data_category')` and then we don't need to provide it each time?

src/fides/service/taxonomy/utils.py
```diff
@@ -0,0 +1,160 @@
...
+from fides.api.models.sql_models import (  # type:ignore[attr-defined]
+    DataCategory,
+    DataSubject,
+    DataUse,
+)
+from fides.api.util.errors import ForbiddenIsDefaultTaxonomyError
+
+
+# Core validation functions for taxonomy operations
+def activate_taxonomy_parents(
```
erosselli, line 16: shouldn't these go in one of the taxonomy handler or service classes? not sure why we would have these in a `utils` file since they do contain quite a bit of important logic

src/fides/api/models/taxonomy.py
galvana: This is fine for the `id` column, I'm just overriding the base definition so it's not treated as a primary key. As for `name`, it's inherited from `FidesBase`.

src/fides/api/models/taxonomy.py
galvana: I did this to encourage the use of `applies_to`. Editing the contents of `applies_to` automatically saves the corresponding rows to the `taxonomy_allowed_usages` table on commit.

src/fides/api/models/taxonomy.py
galvana: Do you mean this as a general statement? As in, in general we should use something more strict for our create methods?

src/fides/api/models/taxonomy.py
galvana: I was getting this error from mypy
```
src/fides/api/models/taxonomy.py:111: error: Unexpected column "source_taxonomy_key" for model
"TaxonomyAllowedUsage"  [misc]
                        TaxonomyAllowedUsage(
                        ^
src/fides/api/models/taxonomy.py:111: error: Unexpected column "target_type" for model
"TaxonomyAllowedUsage"  [misc]
                        TaxonomyAllowedUsage(
                        ^
```
But I just found out you could do this instead (I was missing the `Column[str]` annotations)
```
source_taxonomy_key: Column[str] = Column(
    String,
    ForeignKey("taxonomy.fides_key", ondelete="CASCADE"),
    primary_key=True,
    index=True,
)
target_type: Column[str] = Column(
    String, primary_key=True
)  # Can be "system", "dataset", OR a taxonomy key like "data_categories"
```

src/fides/api/models/taxonomy.py
galvana: I don't think we have general guidance, I usually set `fides_key` as the FK if it's available and it's something user facing. Otherwise, I just stick with ID.

src/fides/api/models/taxonomy.py
galvana: From above:
> I did this to encourage the use of applies_to. Editing the contents of applies_to automatically saves the corresponding rows to the taxonomy_allowed_usages table on commit.

src/fides/api/models/taxonomy.py
galvana: I went with this because it's the most common approach, we usually override `update` instead of `save`. Are you asking what would happen if someone calls `save` directly?

src/fides/api/models/taxonomy.py
galvana: Nice catch! I got rid of the extra index definition

src/fides/service/taxonomy/handlers/base.py
galvana: This is nice! I didn't know about `@overloads` in Python

src/fides/service/taxonomy/handlers/legacy_handler.py
galvana: Good idea 👍

src/fides/service/taxonomy/utils.py
galvana: I have it in `utils` because I don't want this activation to the part of the `TaxonomyService` interface. I'd prefer for a taxonomy element to be activated or de-activated as part of an update.

src/fides/api/api/v1/endpoints/generic_overrides.py
galvana: You right 😄

src/fides/api/models/taxonomy.py
erosselli: okay I totally missed that on my first read. So what's the primary key of this table?

src/fides/service/taxonomy/handlers/base.py
erosselli: I too didn't know they existed but whenever I find myself thinking "In Typescript I could do this way cleaner" I go and look up if there's a way to do it in Python 😆 and there was!

src/fides/api/models/taxonomy.py
erosselli: ah okay! I totally missed the `association_proxy` thing. can we maybe add a comment that explains this then? so if someone goes to use `_allowed_usages` and  wonders "why is this private?" then there's a comment saying "use applies_to instead"

src/fides/api/models/taxonomy.py
erosselli: hah yeah I know we're just following established convention here so feel free to not change, but I did wonder how hard of a change it would be to implement, at least for new models

src/fides/api/models/taxonomy.py
erosselli: yes, I think we have a lot of code that calls save directly so concerned that this is a potential footgun? Also doesn't the `update` method eventually call `save` anyways?

PR #6425, opened 2025-08-06
ENG-1112 and ENG-1109 Location on Privacy Requests

thabofletcher:
Closes [ENG-1109] and [ENG-1112] (combined the PR for easier testing)

### Description Of Changes

Allows the location property to be provided to the following endpoints:

```
POST /api/v1/privacy-request 
POST /api/v1/privacy-request/authenticated 
{
  "location": "US-CA"  // New optional field
}
```

Location is enforce to be ISO 3166 with the EAA exception allowed.

### Code Changes

* Code to accept and validate location on endpoints [ENG-1109]
* Code to return location on privacy requests listing [ENG-1112]
* DB Migration for schema change [ENG-1109]
* Test to make sure it works (both)

### Steps to Confirm

1. Startup Fides and Privacy Center
2. Make an API call like this
```
curl --request POST \
  --url http://localhost:8080/api/v1/privacy-request \
  --data '[
  {
    "identity": {
      "email": "jane@example.com"
    },
    "source": "Privacy Center",
    "location": "US-CA"
  }
]'
```
3. Ensure the response includes the location you sent
4. make an API call to GET all the prtivacy requests (/api/v1/privacy-request), and ensure the reqesut you created has the correct location
5. POST and invalid location, and ensure the response has an error 
```
({
	"detail": [
		{
			"type": "value_error",
			"loc": [
				"body",
				0,
				"location"
			],
			"msg": "Value error, Invalid location format 'US-BLAH'. Must follow ISO 3166 format (e.g., 'US', 'US-CA', 'GB', 'CA-ON', 'EEA')"
		}
	]
})
```

### Pre-Merge Checklist

* [x] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [x] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:
  * [x] Documentation [issue created](https://ethyca.atlassian.net/browse/ENG-1174)

[ENG-1109]: https://ethyca.atlassian.net/browse/ENG-1109?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[ENG-1112]: https://ethyca.atlassian.net/browse/ENG-1112?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,7 +35,12 @@ Changes can also be flagged with a GitHub label for tracking purposes. The URL o
 ### Developer Experience
 - Refactored table column header menu functionality to use column-level menu property instead of custom header cell components [#6481](https://github.com/ethyca/fides/pull/6481)
 - Added new npm task for Admin UI that enables optimized Cypress testing with API server connectivity during local development [#6483](https://github.com/ethyca/fides/pull/6483)
+- Added reusable table system with URL synchronization and Ant Design integration for standardized table behavior and deep linking support [#6447](https://github.com/ethyca/fides/pull/6447)
+- Refactored Action Center tables to use new standardized table state management hooks [#6349](https://github.com/ethyca/fides/pull/6349)
 
+
+### Fixed
+- Handle missing GVL in TCF experience by displaying an error message instead of infinite spinners. [#6472](https://github.com/ethyca/fides/pull/6472)
 
 ## [2.68.0](https://github.com/ethyca/fides/compare/2.67.2...2.68.0)
 
@@ -43,7 +48,6 @@ Changes can also be flagged with a GitHub label for tracking purposes. The URL o
 - Improved logging of underlying errors when raising generic exceptions [#6420](https://github.com/ethyca/fides/pull/6420)
 - Manual Task Graph Tasks now receive data from any nodes referenced by their conditional dependencies. [#6402](https://github.com/ethyca/fides/pull/6402)
 - Added PrivacyCenterSettings to the config. [#6349](https://github.com/ethyca/fides/pull/6439)
-- Added reusable table system with URL synchronization and Ant Design integration for standardized table behavior and deep linking support [#6447](https://github.com/ethyca/fides/pull/6447)
 - Added DSR task conditional operators list types and data type/operator compatibility [#6429](https://github.com/ethyca/fides/pull/6429)
 - Added visual checkmark feedback and screen reader announcements for consent button interactions [#6451](https://github.com/ethyca/fides/pull/6451)
 - Added field to privacy center configuration to support location collection in privacy center actions [#6432](https://github.com/ethyca/fides/pull/6432)
--- a/clients/admin-ui/cypress/e2e/action-center/assets-results.cy.ts
+++ b/clients/admin-ui/cypress/e2e/action-center/assets-results.cy.ts
@@ -246,11 +246,7 @@ describe("Action center Asset Results", () => {
     it("should bulk restore ignored assets", () => {
       // eslint-disable-next-line cypress/no-unnecessary-waiting
       cy.wait(500); // wait for the original router to update
-      cy.getByTestId("asset-state-filter")
-        .contains("Ignored")
-        .click({ force: true });
-      // eslint-disable-next-line cypress/no-unnecessary-waiting
-      cy.wait(500); // wait for the router to update
+      cy.clickAntTab("Ignored");
       cy.getByTestId("asset-state-filter").within(() => {
         cy.get(".ant-menu-item-selected").contains("Ignored").should("exist");
       });
--- a/clients/admin-ui/cypress/e2e/action-center/systems-aggregate.cy.ts
+++ b/clients/admin-ui/cypress/e2e/action-center/systems-aggregate.cy.ts
@@ -1,8 +1,6 @@
 import {
   stubPlus,
-  stubSystemVendors,
   stubTaxonomyEntities,
-  stubVendorList,
   stubWebsiteMonitor,
 } from "cypress/support/stubs";
 
@@ -191,7 +189,7 @@ describe("Action center system aggregate results", () => {
       cy.visit(`${ACTION_CENTER_ROUTE}/${webMonitorKey}#attention-required`);
       cy.location("hash").should("eq", "#attention-required");
 
-      cy.getAntTab("Recent activity").click();
+      cy.clickAntTab("Recent activity");
       cy.location("hash").should("eq", "#recent-activity");
 
       // "recent activity" tab should be read-only
@@ -205,7 +203,7 @@ describe("Action center system aggregate results", () => {
 
       cy.get(".ant-spin-spinning").should("not.exist");
 
-      cy.getAntTab("Ignored").click();
+      cy.clickAntTab("Ignored");
       cy.location("hash").should("eq", "#ignored");
 
       // "ignore" option should not show in bulk actions menu
@@ -219,6 +217,7 @@ describe("Action center system aggregate results", () => {
     it("maintains hash when clicking on a row", () => {
       // no hash (default tab)
       cy.visit(`${ACTION_CENTER_ROUTE}/${webMonitorKey}`);
+      cy.wait("@getSystemAggregateResults");
 
       cy.getAntTableRow("[undefined]").within(() => {
         cy.getByTestId("system-name-link").should(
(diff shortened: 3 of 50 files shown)


src/fides/api/schemas/privacy_request.py
```diff
@@ -18,8 +19,20 @@
...
+# Regex to validate a ISO 3166-2 code:
+# 1. Starts with a 2 letter country code (e.g. "US", "GB") (see ISO 3166-1 alpha-2)
+# 2. (Optional) Ends with a 1-3 alphanumeric character region code (e.g. "CA", "123", "X") (see ISO 3166-2)
+# 3. Country & region codes must be separated by a hyphen (e.g. "US-CA")
+#
+# Fides also supports a special `EEA` geolocation code to denote the European
+# Economic Area; this is not part of ISO 3166-2, but is supported for convenience.
+VALID_ISO_3166_LOCATION_REGEX = re.compile(
+    r"^(?:([a-z]{2})(-[a-z0-9]{1,3})?|(eea))$", re.IGNORECASE
+)
```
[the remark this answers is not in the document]
thabofletcher: I have used this regex to be consistent with the existing front end code in geolocation.ts - please provide suggestions for how this might be better shared between the front end and back end without duplicating code, but unless that is a real possibility, I will probably keep the regex identical for consistency

src/fides/api/schemas/privacy_request.py
```diff
@@ -321,6 +347,7 @@ class PrivacyRequestResponse(FidesSchema):
     custom_privacy_request_fields_approved_by: Optional[str] = None
     custom_privacy_request_fields_approved_at: Optional[datetime] = None
     source: Optional[PrivacyRequestSource] = None
+    location: Optional[str] = None
```
speaker-ender, line 338: Is there a reason we are creating a new field instead of using the existing custom_privacy_requests field?

src/fides/api/api/v1/endpoints/privacy_request_endpoints.py
```diff
@@ -539,6 +540,21 @@ def _filter_privacy_request_queryset(
...
     if external_id:
         query = query.filter(PrivacyRequest.external_id.ilike(f"{external_id}%"))
+    if location:
+        # Support filtering by exact location match or country prefix
+        # e.g., "US" matches both "US" and "US-CA", "US-NY", etc.
+        # "US-CA" matches only "US-CA"
+        if "-" in location:
+            # Exact match for subdivision codes
+            query = query.filter(PrivacyRequest.location == location.upper())
+        else:
```
tvandort: We might want to support both `-` and `_` here. `/api/v1/privacy-experience?region=<region>` supports both and I'd prefer to not have one support both and the other only support one.

src/fides/api/schemas/privacy_request.py
tvandort: I _think_ custom privacy requests fields is for Admin added fields. I think anything official should probably exist outside of that. Unless this is fully optional in that case I can maybe see it in `custom_privacy_requests`.

src/fides/service/privacy_request/privacy_request_service.py
```diff
@@ -71,6 +71,39 @@ def get_privacy_request(self, privacy_request_id: str) -> Optional[PrivacyReques
...
+                # Simple heuristic: if field name contains "required" and "location" and has no value
+                if (
+                    "required" in field_name.lower()
+                    and "location" in field_name.lower()
+                    and (
+                        not hasattr(field_data, "value")
+                        or not field_data.value
+                        or field_data.value == ""
+                    )
+                ):
```
tvandort: Interesting, is there a legacy reason we want to support this heuristic? Is there any way this might over-eagerly apply to existing clients?

src/fides/api/api/v1/endpoints/privacy_request_endpoints.py
tvandort: https://github.com/ethyca/fidesplus/blob/main/src/fidesplus/api/service/privacy_experiences/privacy_experience_service.py#L910-L911

src/fides/api/schemas/privacy_request.py
speaker-ender: This is optional and we are adding to the forms inside the `custom_privacy_request_fields` section of the `config.json`.
If there is some separation between those set in the config vs those set in the admin ui, we should make a clearer distinction.
If not, we should unify that with the options in admin-ui.

src/fides/api/models/privacy_request/privacy_request.py
```diff
@@ -222,6 +222,7 @@ class PrivacyRequest(
     canceled_at = Column(DateTime(timezone=True), nullable=True)
     consent_preferences = Column(MutableList.as_mutable(JSONB), nullable=True)
     source = Column(EnumColumn(PrivacyRequestSource), nullable=True)
+    location = Column(String, nullable=True)
```
vcruces, line 225: Just in case, do you think the location format is something we should validate at the model level?

tests/ops/schemas/test_privacy_request.py
```diff
@@ -52,3 +52,58 @@ def test_invalid_source(self):
                     "source": "Email",
                 }
             )
+
+    def test_valid_location_formats(self):
```
vcruces: For this type of tests it may be better to use `@pytest.mark.parametrize`, and that way a separate test is run for each case and it is simpler to identify in a test run which cases fail

vcruces (commented):
Looks great to me! And really solid test coverage. Just one small question and a suggestion, but nothing required. Great work!

src/fides/api/api/v1/endpoints/privacy_request_endpoints.py
thabofletcher: Thanks, I will make this support both

src/fides/api/schemas/privacy_request.py
thabofletcher: This code was written before we moved the config around, and I though location might be a valuable enough first class citizen to be here. Keep in mind that location is a custom privacy request field in this context, but will soon need to also be a higher level object that can be gathered in order to route to the actions, and a location set in either place would be part of the model here. So I do think location is more than a custom privacy request field

src/fides/service/privacy_request/privacy_request_service.py
thabofletcher: On your first question this looks like AI slop and I apologize, I'll get it to use "required" from the actual configuration to drive the validation. Thanks for seeing this.

On the second question, this only validates when they submit the request, so I don't think it would apply to client's existing requests if they change location to required.

src/fides/api/api/v1/endpoints/privacy_request_endpoints.py
thabofletcher: Let me know what you think of the new implementation here. I asked Cursor to look for the code doing similar things and it only could find the FE, so please let me know if I'm reinventing the wheel!

src/fides/api/models/privacy_request/privacy_request.py
thabofletcher: I've recently updated the pydantic validators to coalesce to strict ISO-3166 2 format - so we could be more opinionated at the model level here. However, I don't typically see that type of validation in the model layer here. Do you think the Pydantic layer is sufficient?

tests/ops/schemas/test_privacy_request.py
thabofletcher: Great, suggestion, thanks! This should be done now

src/fides/service/privacy_request/privacy_request_service.py
tvandort: Mm, I don't think I was clear in what I was asking. By existing clients what I mean is, is it possible that someone has configured a basic dropdown called location already, that we will now be validating suddenly?

src/fides/api/models/privacy_request/privacy_request.py
vcruces: Yes, I think it should be enough in the Pydantic layer

vcruces (approved):

src/fides/service/privacy_request/privacy_request_service.py
thabofletcher: Sorry, I just hadn't pushed up the fix yet, I thought I had it pushed up.

I'm not sure if that validator would run on it before or what it would do, but the new version definitely wont impact it since it will apply based on actual field type.

PR #6441, opened 2025-08-11
ENG-1128

adamsachs:
Closes https://ethyca.atlassian.net/browse/ENG-1128

Done alongside https://github.com/ethyca/fidesplus/pull/2417

### Description Of Changes

Adds a new `FidesBase.sanitize_key` classmethod to sanitize a `key` value based on the `ALLOWED_CHARS` specified in the `base_class` module, which is the criteria used when `save`ing a `OrmWrappedFidesBase`. Specifically, this disallows `.` from the `key` value.

Additionally, this PR includes a significant data migration to migrate any `MonitorConfig.key`s  that currently had a `.` in them - the `.` are replaced with `_`, and new conflicts are accounted for by appending an integer as needed. This migration is needed because, before this change, there was a gap in our validation logic that allowed `MonitorConfig`s to be created with these `key`s, which would cause downstream problems. Since we are now _disallowing_  `MonitorConfig.key`s with `.` in them (see https://github.com/ethyca/fidesplus/pull/2417), we need to update any existing `MonitorConfig.key`s with `.` in them, and the many pointers in our DB to these `key` values -- specifically, `StagedResource.urn`s include a `MonitorConfig.key` as their first component, so we need to update these `urn`s and all their references.

We also now include DB-level validation on the `MonitorConfig.key` column, in addition to the API-level validation added in the fidesplus PR referenced above.

NOTE: there is no downgrade migration specified for the data migration here. If we need to revert the data migration for any reason, that will be a manual effort. We've stored the old -> new monitor config key "mapping" in the `DBCache` DB table as an artifact, in case we do need to walk back our steps manually in any case.

### Code Changes

* [x] Add `FidesBase.sanitize_key` classmethod that sanitizes an input value based on the `ALLOWED_CHARS` that determines valid `FidesBase.key`s
* [x] Database validation for `MonitorConfig.key`s to prevent them from containing `.` in their values 
* [x] (Data) migration to update existing `MonitorConfig.key`s that have `.` in them and their many referrences 

### Steps to Confirm

See fidesplus PR for manual testing steps

### Pre-Merge Checklist

* [x] Issue requirements met
* [x] `CHANGELOG.md` updated
  * [x] Add a https://github.com/ethyca/fides/labels/high-risk label to the entry if your change includes a high-risk change (i.e. potential for performance impact or unexpected regression) that should be flagged
* Followup issues:
  * [ ] Followup issues created
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:

### DIFF
--- a/src/fides/api/alembic/migrations/versions/2f3c1a2d6b10_disallow_dot_in_monitor_key_and_update_refs.py
+++ b/src/fides/api/alembic/migrations/versions/2f3c1a2d6b10_disallow_dot_in_monitor_key_and_update_refs.py
@@ -62,6 +62,10 @@ def _fetch_monitor_key_mapping(conn: Connection) -> Dict[str, str]:
 
 
 def _update_monitorconfig_keys(conn: Connection, mapping: Dict[str, str]) -> None:
+    """
+    Updates `MonitorConfig.key`s based on the mapping of 'old' keys to 'new' keys.
+    """
+
     logger.info("Updating monitorconfig keys...")
     for old, new in mapping.items():
         conn.execute(
@@ -78,6 +82,10 @@ def _update_monitorconfig_keys(conn: Connection, mapping: Dict[str, str]) -> Non
 
 
 def _update_monitorexecution(conn: Connection, mapping: Dict[str, str]) -> None:
+    """
+    Updates `monitorexecution.monitor_config_key` to reference new `MonitorConfig.key`s.
+    """
+
     logger.info("Updating monitorexecution monitor_config_key...")
     for old, new in mapping.items():
         conn.execute(
@@ -96,6 +104,9 @@ def _update_monitorexecution(conn: Connection, mapping: Dict[str, str]) -> None:
 def _update_stagedresource_monitor_config_id(
     conn: Connection, mapping: Dict[str, str]
 ) -> None:
+    """
+    Updates `stagedresource.monitor_config_id` to reference new `MonitorConfig.key`s.
+    """
     logger.info("Updating stagedresource monitor_config_id...")
     for old, new in mapping.items():
         conn.execute(
@@ -111,37 +122,51 @@ def _update_stagedresource_monitor_config_id(
     logger.info("Finished updating stagedresource monitor_config_id.")
 
 
-def _replace_prefix_sql(table: str, column: str) -> str:
-    """
-    Returns SQL snippet to replace leading '<old>.' prefix in a text column with '<new>.' while preserving the rest of the string.
-    Assumes parameters :old and :new are bound.
-    """
-    # substring from old prefix length + 1 (since we want to keep the starting '.')
-    return f"""
-        UPDATE {table}
-        SET {column} = :new || SUBSTRING({column} FROM CHAR_LENGTH(:old) + 1)
-        WHERE {column} LIKE :old_prefix
-    """
-
-
 def _update_stagedresource_urns_and_parent(
     conn: Connection, mapping: Dict[str, str]
 ) -> None:
+    """
+    Updates `stagedresource.urn`s and `stagedresource.parent`s to reference new `MonitorConfig.key`s.
+    """
     logger.info("Updating stagedresource urns and parent...")
     for old, new in mapping.items():
         params = {"old": old, "new": new, "old_prefix": f"{old}.%"}
-        # urn
-        conn.execute(sa.text(_replace_prefix_sql("stagedresource", "urn")), params)
-        # parent
-        conn.execute(sa.text(_replace_prefix_sql("stagedresource", "parent")), params)
+
+        conn.execute(
+            sa.text(
+                f"""
+                UPDATE stagedresource
+                SET urn = :new || SUBSTRING(urn FROM CHAR_LENGTH(:old) + 1)
+                WHERE urn LIKE :old_prefix AND monitor_config_id = :new
+                """
+            ),
+            params,
+        )
+
+        conn.execute(
+            sa.text(
+                f"""
+                UPDATE stagedresource
+                SET parent = :new || SUBSTRING(parent FROM CHAR_LENGTH(:old) + 1)
+                WHERE parent LIKE :old_prefix AND monitor_config_id = :new
+                """
+            ),
+            params,
+        )
     logger.info("Finished updating stagedresource urns and parent.")
 
 
 def _update_stagedresource_children_array(
     conn: Connection, mapping: Dict[str, str]
 ) -> None:
+    """
+    Updates `stagedresource.children` to reference new `MonitorConfig.key`s.
+
+    Elements from `stagedresource.children` are unnested, updated, and re-aggregated in a single query.
+    """
     logger.info("Updating stagedresource children array...")
     for old, new in mapping.items():
+
         conn.execute(
             sa.text(
                 """
@@ -154,8 +179,8 @@ def _update_stagedresource_children_array(
                     ), '{}')
                     FROM unnest(children) AS val
                 )
-                WHERE children IS NOT NULL
-                AND EXISTS (SELECT 1 FROM unnest(children) AS v WHERE v LIKE :old_prefix)
+                WHERE monitor_config_id = :new
+                AND children IS NOT NULL
                 """
             ),
             {"old": old, "new": new, "old_prefix": f"{old}.%"},
@@ -164,11 +189,17 @@ def _update_stagedresource_children_array(
 
 
 def _update_stagedresource_meta_json(conn: Connection, mapping: Dict[str, str]) -> None:
+    """
+    Updates necessary keys in `stagedresource.meta` (`direct_child_urns` and `top_level_field_urn`)
+    to reference new `MonitorConfig.key`s.
+    """
     logger.info("Updating stagedresource meta json...")
 
     for old, new in mapping.items():
         params = {"old": old, "new": new, "old_prefix": f"{old}.%"}
-        # Update direct_child_urns array in meta
+
+        # updates `stagedresource.meta.direct_child_urns` to reference new `MonitorConfig.key`
+        # elements from `stagedresource.meta.direct_child_urns` are unnested, updated, and re-aggregated in a single query.
         conn.execute(
             sa.text(
                 """
@@ -186,12 +217,14 @@ def _update_stagedresource_meta_json(conn: Connection, mapping: Dict[str, str]) 
                     ),
                     true
                 )
-                WHERE meta ? 'direct_child_urns'
+                WHERE monitor_config_id = :new AND meta ? 'direct_child_urns'
                 """
             ),
             params,
         )
-        # Update top_level_field_urn string in meta
+
+        # updates `stagedresource.meta.top_level_field_urn` to reference new `MonitorConfig.key`
+        # we coalesce the value to 'null'::jsonb to ensure a 'JSON null' value is set, not a column-level NULL
         conn.execute(
             sa.text(
                 """
@@ -206,7 +239,7 @@ def _update_stagedresource_meta_json(conn: Connection, mapping: Dict[str, str]) 
                     ), 'null'::jsonb),
                     true
                 )
-                WHERE meta ? 'top_level_field_urn'
+                WHERE monitor_config_id = :new AND meta ? 'top_level_field_urn'
                 """
             ),
             params,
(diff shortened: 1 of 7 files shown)


src/fides/api/db/base_class.py
```diff
@@ -109,6 +114,19 @@ def id_field_path(self) -> str:
...
+    @classmethod
+    def sanitize_key(cls, proposed_key: str) -> str:
+        """
+        Sanitize the key by removing invalid characters.
+
+        Invalid characters are based on the allowed characters defined in this module.
+        Note that this differs slightly from the allowed characters in `sanitize_fides_key`
+        from `core/utils.py`; specifically, `.` are disallowed (converted to `_`) here, while
+        they are allowed in `sanitize_fides_key`. This is because `.` must be allowed in
+        `fides_key`s (), but they are not allowed in `key`s.
```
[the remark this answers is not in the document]
adamsachs: fixed

src/fides/api/alembic/migrations/versions/2f3c1a2d6b10_disallow_dot_in_monitor_key_and_update_refs.py
```diff
@@ -0,0 +1,316 @@
...
+    """
+    Returns SQL snippet to replace leading '<old>.' prefix in a text column with '<new>.' while preserving the rest of the string.
+    Assumes parameters :old and :new are bound.
+    """
+    # substring from old prefix length + 1 (since we want to keep the starting '.')
+    return f"""
+        UPDATE {table}
+        SET {column} = :new || SUBSTRING({column} FROM CHAR_LENGTH(:old) + 1)
+        WHERE {column} LIKE :old_prefix
+    """
```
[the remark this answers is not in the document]
adamsachs: fixed

src/fides/api/models/detection_discovery/core.py
```diff
@@ -332,6 +332,12 @@ def derive_execution_trigger_dict(data: Dict[str, Any]) -> None:
                 cron_trigger_dict["month"] = execution_start_date.month
             data["monitor_execution_trigger"] = cron_trigger_dict
 
+    @validates("key")
+    def validate_key_no_dots(self, _key: str, value: str) -> str:
+        if value and "." in value:
+            raise ValueError('MonitorConfig.key cannot contain "." characters')
```
erosselli, line 345: according to code cov the line for raising the error isn't covered ,can we add a test somewhere that covers it? 🙂

src/fides/api/alembic/migrations/versions/2f3c1a2d6b10_disallow_dot_in_monitor_key_and_update_refs.py
```diff
@@ -0,0 +1,328 @@
...
+            ),
+            {"old": old, "new": new},
+        )
+    logger.info("Finished updating stagedresource monitor_config_id.")
+
+
+def _replace_prefix_sql(table: str, column: str) -> str:
+    """
+    Returns SQL snippet to replace leading '<old>.' prefix in a text column with '<new>.' while preserving the rest of the string.
+    Assumes parameters :old and :new are bound.
```
erosselli: this also assumes `:old_prefix` is bound right?

src/fides/api/alembic/migrations/versions/2f3c1a2d6b10_disallow_dot_in_monitor_key_and_update_refs.py
```diff
@@ -0,0 +1,328 @@
...
+    """
+    # substring from old prefix length + 1 (since we want to keep the starting '.')
+    return f"""
+        UPDATE {table}
+        SET {column} = :new || SUBSTRING({column} FROM CHAR_LENGTH(:old) + 1)
+        WHERE {column} LIKE :old_prefix
+    """
+
+
+def _update_stagedresource_urns_and_parent(
```
erosselli, line 136: how do we expect this to perform in environments that have millions of stagedresource rows? any concerns?

src/fides/api/alembic/migrations/versions/2f3c1a2d6b10_disallow_dot_in_monitor_key_and_update_refs.py
```diff
@@ -0,0 +1,328 @@
...
+        conn.execute(sa.text(_replace_prefix_sql("stagedresource", "parent")), params)
+    logger.info("Finished updating stagedresource urns and parent.")
+
+
+def _update_stagedresource_children_array(
+    conn: Connection, mapping: Dict[str, str]
+) -> None:
+    logger.info("Updating stagedresource children array...")
+    for old, new in mapping.items():
+        conn.execute(
```
erosselli, line 181: can we add a docstring / code comment to explain this query a bit? just trying to leave a trail if we ever end up looking at this code in the future

src/fides/api/alembic/migrations/versions/2f3c1a2d6b10_disallow_dot_in_monitor_key_and_update_refs.py
```diff
@@ -0,0 +1,328 @@
...
+                """
+                INSERT INTO dbcache (id, namespace, cache_key, cache_value, created_at, updated_at)
+                VALUES ('dbc_' || gen_random_uuid(), :namespace, :cache_key, convert_to(:value, 'UTF8')::bytea, now(), now())
+                ON CONFLICT (namespace, cache_key)
+                DO UPDATE SET cache_value = EXCLUDED.cache_value, updated_at = now()
+                """
+            ),
+            {
+                "namespace": "monitor-config-key-mapping",
+                "cache_key": "monitor-config-key-mapping",
```
erosselli: I was thinking we would cache each monitor config as a separate row but this works too. I think any way that we can query for the data in case of emergency works so 👍

src/fides/api/alembic/migrations/versions/2f3c1a2d6b10_disallow_dot_in_monitor_key_and_update_refs.py
```diff
@@ -0,0 +1,328 @@
...
+        # Recreate FKs (mark DEFERRABLE to allow future migrations to defer checks during complex updates)
+        op.create_foreign_key(
+            "fk_staged_resource_ancestor_ancestor",
+            "stagedresourceancestor",
+            "stagedresource",
+            ["ancestor_urn"],
+            ["urn"],
+            ondelete="CASCADE",
+            deferrable=True,
+            initially="DEFERRED",
```
erosselli: a question just cause I'm not sure , `initially="DEFERRED"` defers the check until... when?

src/fides/api/alembic/migrations/versions/2f3c1a2d6b10_disallow_dot_in_monitor_key_and_update_refs.py
```diff
@@ -0,0 +1,328 @@
...
+        logger.info("Finished recreating FKs.")
+
+        logger.info("Finished monitorconfig key 'dot' migration.")
+
+    # Add CHECK constraint to forbid dots in monitorconfig.key going forward
+    op.create_check_constraint(
+        "ck_monitorconfig_key_no_dots",
+        "monitorconfig",
+        "key NOT LIKE '%.%'",
+    )
```
erosselli, line 431: is there a way to define this in the model too so it's defined somewhere declarative rather than only in the migration?

src/fides/api/alembic/migrations/versions/2f3c1a2d6b10_disallow_dot_in_monitor_key_and_update_refs.py
adamsachs: that's right! good spot 👍

src/fides/api/alembic/migrations/versions/2f3c1a2d6b10_disallow_dot_in_monitor_key_and_update_refs.py
vcruces: I think they are validated at the end of the transaction (commit)

src/fides/api/alembic/migrations/versions/2f3c1a2d6b10_disallow_dot_in_monitor_key_and_update_refs.py
adamsachs: yup good call, will do for this and the other queries here that aren't obvious 👍

src/fides/api/alembic/migrations/versions/2f3c1a2d6b10_disallow_dot_in_monitor_key_and_update_refs.py
adamsachs: yeah, my preference is just to condense this as much as possible since it's not really the intended use of this table!

src/fides/api/alembic/migrations/versions/2f3c1a2d6b10_disallow_dot_in_monitor_key_and_update_refs.py
adamsachs: oh, let me dig on that a bit further. full transparency, this was a cursor-addition that i took at face value - the `deferrable` makes sense to me i think (i'll double check that one as well), but i should dig deeper on the `initally="DEFERRED"` 👍

src/fides/api/alembic/migrations/versions/2f3c1a2d6b10_disallow_dot_in_monitor_key_and_update_refs.py
adamsachs: yeah, that's a good call - i think a helpful optimization across the board here would be to leverage the `monitor_config_id` on any updates to `stagedresource`s, since that's an indexed column and it should allow us to effectively limit the scope of our update queries to _not_ scan all stagedresource rows. i can put that in place.

we can't do something similar for `stagedresourceancestor` unfortunately, because it doesn't have a column that scopes a row to its associated `monitor_config_id` (it probably should! that would likely be a helpful optimization for other scenarios as well). the best optimization i can think of there is to update the query mechanism so that we have to issue only a single query (and execute only a single table scan). i'd be open to some other ideas, if you have any?

in practice, we've fortunately (for the moment) got a limited set of fides users with such a large set of staged resources, and i can confirm that those users do not have monitor configurations with `.` in their key, so this migration shouldn't be triggered for them.

so we can do a few things to address the concern here:
1. significantly optimize our stagedresource queries to be scoped by the indexed `monitor_config_id` column
2. update the `stagedresourceancestor` table in one-go, so that we're only executing a single table scan for the entire migration (if needed at all)
3. confirm our high-volume customers do not have monitor configurations with `.` in their key; if they do, set the expectation of migration here that may take some additional time and to account for that in the upgrade deployment.

i'll go ahead and make the code changes now, and will also be sure to coordinate and communicate on the overall migration risk.

erosselli: sounds good!

src/fides/api/alembic/migrations/versions/2f3c1a2d6b10_disallow_dot_in_monitor_key_and_update_refs.py
adamsachs: i think i'm going to remove these annotations generally. i can't feel super confident about any unintended side effects they may have and i'm reluctant to start a new standard here in the middle of a complex migration. 

needing to delete and recreate the FK's within the occasional data migration doesn't seem too burdensome - if we actually wanted the 'deferrable' behavior in our app runtime i'd think about this differently, but since we don't, i'll leave this out.

thanks for calling attention to it!

src/fides/api/models/detection_discovery/core.py
```diff
@@ -172,6 +173,10 @@ class MonitorConfig(Base):
 
     shared_config = relationship(SharedMonitorConfig)
 
+    CheckConstraint(  # type: ignore
+        "key NOT LIKE '%.%'", name="ck_monitorconfig_key_no_dots"
+    )
```
erosselli: does this need to go in `__table_args__` attribute or can it "float" here on the model directly?

erosselli (approved):
Thanks for addressing all the comments. Left a non-blocking question

src/fides/api/models/detection_discovery/core.py
adamsachs: honestly, i was just following existing examples in our codebase, but it does seem like that's the more proper way to define it! i think i'll adjust here.

adamsachs: updated in https://github.com/ethyca/fides/pull/6441/commits/9449053d075ccab5ff956a2ed5f5dcd2cf6a2364

PR #6443, opened 2025-08-11
Configure role assumption in create client.

tvandort:
Closes [ENG-1141]

### Description Of Changes

Configure S3 role assumption in client creation code so that all uses support role assumption.

### Code Changes

### Steps to Confirm

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:

### DIFF
--- a/src/fides/api/util/aws_util.py
+++ b/src/fides/api/util/aws_util.py
@@ -8,6 +8,8 @@ from loguru import logger
 from fides.api.common_exceptions import StorageUploadError
 from fides.api.schemas.storage.storage import AWSAuthMethod, StorageSecrets
 
+from fides.config import CONFIG
+
 
 def get_aws_session(
     auth_method: str,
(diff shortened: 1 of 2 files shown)


src/fides/api/util/aws_util.py
```diff
@@ -95,10 +97,16 @@ def get_s3_client(
     If an `assume_role_arn` is provided, the secrets will be used to
     assume that role and return a Session instantiated with that role.
     """
+
```
adamsachs, line 104: i would callout in the docstring here that we fallback/default to the configured storage credentials.

i don't think we've got any cases yet where we're calling `get_s3_client` outside of the "storage" functionality, but it's not inconceivable that we'd want e.g. an s3 _monitor_ to call this utility function, and it may not rely on these same credentials.

i don't think that hypothetical should prevent us from making this update, but i do think that we should make it clear to any potential future callers of this very broad utility function that they need to be a bit careful about the assumptions it makes on where it's going to look for the `assume_role_arn`...

src/fides/api/util/aws_util.py
```diff
@@ -94,11 +95,21 @@ def get_s3_client(
 
     If an `assume_role_arn` is provided, the secrets will be used to
     assume that role and return a Session instantiated with that role.
+
+    If no `assume_role_arn` is provided, and `aws_s3_assume_role_arn` is
+    configured in the storage config, then that config value will be used
+    to assume that role and return a Session instantiated with that role.
```
adamsachs: this is a bit of a nit but i do think 'storage config' could be confused for a few different things, so is this clearer?
```suggestion
    If no `assume_role_arn` is provided, and `aws_s3_assume_role_arn` is
    configured in the global `credentials.storage` config, then that config value will be used
    to assume that role and return a Session instantiated with that role.
```

PR #6447, opened 2025-08-11
Deep linking tables

gilluminate:
Closes [ENG-831]

### Description Of Changes

Added a comprehensive reusable table system with URL synchronization and Ant Design integration to standardize table behavior across the Admin UI. The system provides deep linking support for pagination, sorting, filtering, and search through URL query parameters. Implemented modular hooks for individual table features that can be used standalone or combined. Migrated existing tables to use the new system and removed deprecated pagination components.

### Code Changes

* Added `useTableState` hook for managing table state with URL synchronization
* Added `useAntTable` hook for Ant Design table integration with selection and bulk actions
* Added standalone `usePagination`, `useSearch`, and `useSorting` hooks with URL sync
* Added comprehensive TypeScript types and interfaces for table state management
* Added NuQS integration for URL state management with custom parsers
* Added extensive test coverage for all new hooks with mocked URL state
* Updated existing `DiscoveredAssetsTable` table to use new hook system (will apply to more after this PR is merged)
* Removed deprecated `CustomPagination` component and related exports
* Added table constants for default page sizes and pagination settings
* Added comprehensive documentation with examples and migration guide`

### Steps to Confirm

1. Visit https://fides-plus-nightly-9g2rf5az7-ethyca.vercel.app/settings/about and enable "Web monitor" and "Asset consent status labels"
2. Visit https://fides-plus-nightly-9g2rf5az7-ethyca.vercel.app/data-discovery/action-center/Test_Monitor_cv75q2qr9bk/%5Bundefined%5D
3. Verify that sorting, filters, search, and pagination are working as expected
5. Verify row selection works across pages and bulk actions function properly
6. Test deep linking by opening the URL with table state parameters in a new tab and confirming the table loads with correct state.

### Pre-Merge Checklist

* [x] Issue requirements met
* [x] `CHANGELOG.md` updated
* Followup issues:
  * [x] Followup issues created
  * [ ] No followup issues
* Database migrations:
  * [x] No migrations
* Documentation:

[ENG-831]: https://ethyca.atlassian.net/browse/ENG-831?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/clients/fidesui/src/hoc/CustomTable.tsx
+++ b/clients/fidesui/src/hoc/CustomTable.tsx
@@ -26,8 +26,9 @@ export const CustomTable = <RecordType = any,>({
 
     return {
       showSizeChanger: true,
-      hideOnSinglePage: pagination.pageSize === pagination.pageSizeOptions?.[0],
-      // Then apply any user-provided config (allows overriding defaults)
+      hideOnSinglePage:
+        pagination.pageSize?.toString() ===
+        pagination.pageSizeOptions?.[0]?.toString(), // Hide pagination if there's only one page
       ...pagination,
     };
   }, [pagination, dataSource]);


clients/admin-ui/cypress/e2e/action-center/assets-results.cy.ts
```diff
@@ -339,6 +339,9 @@ describe("Action center Asset Results", () => {
         );
         cy.location("hash").should("eq", "#attention-required");
 
+        // wait for router
+        // eslint-disable-next-line cypress/no-unnecessary-waiting
+        cy.wait(500);
```
gilluminate, line 352: unrelated to the changes in this PR, just more tab related flaky tests

clients/admin-ui/jest.config.js
```diff
@@ -37,6 +37,7 @@ module.exports = {
     "test-utils.tsx",
   ],
   testEnvironment: "jsdom",
+  extensionsToTreatAsEsm: [".ts", ".tsx"],
```
gilluminate: necessary for mocking Nuqs in tests

clients/admin-ui/__tests__/features/hooks/tables/useAntTable.test.ts
```diff
@@ -0,0 +1,526 @@
...
+    const { tableProps } = result.current;
+
+    const withId = { id: "abc", name: "A" } as Row;
+    const withKey = { key: "k1", name: "B" } as Row;
+
+    if (typeof tableProps.rowKey === "function") {
+      expect(tableProps.rowKey(withId)).toBe("abc");
+      expect(tableProps.rowKey(withKey)).toBe("k1");
+    }
+  });
```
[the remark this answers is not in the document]
gilluminate: fixed

clients/admin-ui/package.json
```diff
@@ -24,7 +24,8 @@
     "openapi:generate": "openapi --input http://localhost:8080/openapi.json --output ./src/types/api --exportCore false --exportServices false --indent 2 && prettier --write .",
     "openapi:generate-dictionary": "openapi --input http://localhost:8081/openapi.json --output ./src/types/dictionary-api --exportCore false --exportServices false --indent 2",
     "start": "next start",
-    "test": "jest --watchAll",
+    "test": "jest",
+    "test:watch": "jest --watchAll",
```
speaker-ender, line 28: Thank you! Watching the tests should def not be the default.

clients/admin-ui/src/features/common/DebouncedSearchInput.tsx
```diff
@@ -12,6 +12,11 @@ export const DebouncedSearchInput = ({
   const INPUT_CHANGE_DELAY = 500;
   const [currentInput, setCurrentInput] = useState(value);
 
+  // Keep internal state in sync with external value prop
+  useEffect(() => {
+    setCurrentInput(value || "");
+  }, [value]);
+
```
speaker-ender, line 19: Adding an effect like this makes me think that this component doesn't need an internal state at all.  
If the internal state should reflect the prop value being passed in, why not just always use that prop directly?

clients/admin-ui/src/features/common/hooks/types.d.ts
```diff
@@ -0,0 +1,111 @@
+/**
```
speaker-ender: I like the idea here of putting some of the types in their own file rather than tied to a component or hook.  
Would it make sense to have these in their own directory?  ie `src/features/common/pagination/types.d.ts` and `src/features/common/sorting/types.d.ts`  
Not the most opinionated on this and think the current directory pattern needs to be evaluated, but I'd like these types to be easily discoverable to promote reuse.

clients/admin-ui/src/features/common/hooks/usePagination.ts
```diff
@@ -0,0 +1,165 @@
...
+  const displayPageSizeOptions = pageSizeOptions ?? DEFAULT_PAGE_SIZES;
+
+  // Create parsers for pagination state
+  // Note: Parsers must be stable across renders for NuQS to work properly
+  const parsers = useMemo(() => {
+    const createdParsers = createPaginationParsers({
+      pageSize: defaultPageSize,
+      pageSizeOptions, // Pass undefined if user didn't provide it
+    });
+    return createdParsers;
```
speaker-ender: nit: can we just return the function result without using an intermediary const?

clients/admin-ui/src/features/common/hooks/usePagination.ts
```diff
@@ -0,0 +1,165 @@
...
+          : pageIndex;
+
+      const updates: PaginationQueryParams = { page: newPageIndex };
+      if (pageSize !== undefined) {
+        updates.size = pageSize;
+      }
+      setQueryState(updates);
+    },
+    [setQueryState, currentState.pageSize],
+  );
```
speaker-ender: I think this is a perfect example of when an effect would make sense to use instead of a function.  
Calling a function that accepts and *requires* a value (`pageIndex`), it would be reasonable to expect it to always do that thing.
Instead, it only sometimes updates `pageIndex` depending on other parameters + the current state.
If you use independent functions (ex. `updatePageIndex`, `updatePageSize`), you can reduce a lot of complexity with the `updatePagination` function and more clearly document the side effects of updating the page size either directly in the updater function or in a `useEffect` hook.

note: I see how this is an issue with the ant pagination component rather than a decision on your end.  
I think this could still be achieved by using the `onChange` to only update the index and `onShowSizeChange` to only update the size.
It's not ideal but optimizations could be made later.

clients/admin-ui/src/features/common/hooks/usePagination.ts
```diff
@@ -0,0 +1,165 @@
...
+      size: defaultPageSize,
+    });
+  };
+
+  // Call onPaginationChange when state changes
+  useEffect(() => {
+    if (onPaginationChange) {
+      onPaginationChange(currentState);
+    }
+  }, [currentState, onPaginationChange]);
```
speaker-ender: Couldn't this be implemented by the component that uses this hook when it's needed?

clients/admin-ui/src/features/common/hooks/usePagination.ts
```diff
@@ -0,0 +1,165 @@
...
+    return createdParsers;
+  }, [defaultPageSize, pageSizeOptions]);
+
+  // Use NuQS for URL state management
+  const [queryState, setQueryState] = useQueryStates(parsers, {
+    history: "push",
+  });
+
+  // Create current state from query state (URL is the single source of truth)
+  const currentState: PaginationState = useMemo(
```
speaker-ender: Was there a specific performance issue that required a useMemo?

clients/admin-ui/src/features/common/hooks/usePagination.ts
```diff
@@ -0,0 +1,165 @@
...
+      current: currentState.pageIndex,
+      pageSize: currentState.pageSize,
+      showSizeChanger,
+      pageSizeOptions: displayPageSizeOptions.map(String),
+      showQuickJumper: true,
+      showTotal: (total: number, range: [number, number]) =>
+        `${range[0]}-${range[1]} of ${total} items`,
+      onChange: updatePagination,
+      onShowSizeChange: updatePagination,
+    },
```
speaker-ender: Not a blocker, but I think tying this hook to component props (esp without strongly typing it) will cause some headaches in the future.  
It would be nice to contain something like this in a mapping/translation function and possibly a separate hook for convenience.

clients/admin-ui/src/features/common/DebouncedSearchInput.tsx
gilluminate: If you used the value prop directly, here's what happens:

1. User types "a" → debouncedOnChange is called but waits 500ms
2. During those 500ms: The input field shows the OLD value because the parent hasn't updated the value prop yet
3. User continues typing "ab" → Input still shows the old value, making it appear broken
4. After 500ms: Parent finally updates, input shows "a"
5. User types "abc" → Input still shows "a" (another 500ms delay)
6. Cycle repeats: Input always lags 500ms behind the user's typing

clients/admin-ui/src/features/common/hooks/types.d.ts
gilluminate: I'd probably sooner just name them `sorting.d.ts` than have a directory structure like that. It's one of the benefits of using the `.d.ts` suffix.

clients/admin-ui/src/features/common/hooks/usePagination.ts
gilluminate: Could do. I typically lean on this pattern for debugging benefits because it's easier to add breakpoints/console logs.

clients/admin-ui/src/features/common/DebouncedSearchInput.tsx
gilluminate: Outside of debounce though, this is just a typical Controlled input pattern.

clients/admin-ui/src/features/common/hooks/usePagination.ts
gilluminate: While I completely agree, Ant is unexpectedly calling the `onShowSizeChange` whenever a basic page change happens, despite what their docs say. So attempting to fix it like this means a page change can't happen because the `onShowSizeChange` subsequently resets the page to 1. I'm going to leave this as is, but add a note about the Ant issue it's solving.

gilluminate: Oh! nevermind, that was something I was causing in useAntTable...still working on this :)

gilluminate: Ok, that's done. Good suggestion!

clients/admin-ui/src/features/common/hooks/usePagination.ts
gilluminate: yeah, totally. This is superfluous. Removed!

clients/admin-ui/src/features/common/hooks/usePagination.ts
gilluminate: removed. good call out

clients/admin-ui/src/features/common/hooks/usePagination.ts
gilluminate: solid point. I'll create a `useAntPagination` similar to `useAntTable`

speaker-ender (approved):
This is looking really good!
Tested the preview and nothing seems broken.
Code is looking clean 👌

PR #6453, opened 2025-08-13
System table migration and system group functionality

jpople:
Closes ENG-1077, ENG-1155, ENG-1110

### Description Of Changes

Migrates system table to Ant Design, changes columns, and adds support for system group features.

### Steps to Confirm

Run a Fidesplus backend on the branch `ENG-1074-system-group-APIs` and enable the alpha flag "System Groups" at /settings/about/alpha.

0. General regression on system table
1. Select systems and add them to a new group using the "create new group" option in the dropdown
2. Select systems and add them to an existing group using the dropdown after doing the above
3. Add an individual system to a group using the cell
4. Filter systems by group or data steward using the filter menu
5. When not using Fidesplus or without the alpha flag, system group features should not be visible

### Pre-Merge Checklist

* [x] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:


clients/fidesui/src/palette/palette.module.scss
```diff
@@ -38,16 +47,25 @@ $colors: (
...
+  "bg-alert": $bg-alert,
+
+  // Taxonomy colors //
+  "bg-taxonomy-white": $bg-default,
+  "bg-taxonomy-red": $bg-error,
+  "bg-taxonomy-orange": $bg-warning,
+  "bg-taxonomy-yellow": $bg-caution,
+  "bg-taxonomy-green": $bg-success,
+  "bg-taxonomy-blue": $bg-info,
+  "bg-taxonomy-purple": $bg-alert,
```
gilluminate: This seems unnecessarily duplicative and doesn't follow a separation of concerns. The color palette naming should be agnostic and not tied to specific use cases. Is there a reason you couldn't just use, for example, `var(--fidesui-bg-orange)` in your `COLOR_VALUES` definition?

clients/admin-ui/cypress/e2e/systems/plus/systems-table-plus.cy.ts
```diff
@@ -0,0 +1,84 @@
...
+  });
+
+  describe("system groups", () => {
+    it("displays system groups column when enabled", () => {
+      cy.findAllByRole("columnheader").contains("Groups").should("exist");
+    });
+
+    it("allows filtering by system groups", () => {
+      cy.intercept("/api/v1/system?**").as("getSystemsByGroup");
+      cy.get(".ant-table-filter-column")
```
gilluminate: We could probably create a new `getAntTableColumn` command for this type of thing

clients/admin-ui/cypress/e2e/systems/systems-table.cy.ts
```diff
@@ -0,0 +1,112 @@
...
+
+  describe("table display and navigation", () => {
+    it("displays systems in a table format", () => {
+      cy.getAntTableRow("fidesctl_system").should("exist");
+      cy.getAntTableRow("demo_analytics_system").should("exist");
+    });
+
+    it("allows clicking on system name to navigate to system details", () => {
+      cy.getAntTableRow("fidesctl_system").within(() => {
+        cy.get("a").click();
```
gilluminate: This feels quite brittle and could start failing if another link is added to the rows. Can we add a data-testid to these links or anything?

clients/admin-ui/cypress/e2e/systems/systems-table.cy.ts
```diff
@@ -0,0 +1,112 @@
...
+    });
+  });
+
+  describe("filtering", () => {
+    it("filters systems by data steward", () => {
+      cy.intercept("/api/v1/system?**").as("getSystemsBySteward");
+      cy.get(".ant-table-filter-column").within(() => {
+        cy.findByRole("button").click({ force: true });
+      });
+      cy.get(".ant-dropdown-menu-item").first().click({ force: true });
```
gilluminate: could `cy.applyTableFilter` be used here?

clients/admin-ui/src/features/system/SystemActionsMenu.tsx
```diff
@@ -0,0 +1,114 @@
...
+      {contextHolder}
+      <Modal
+        open={deleteModal}
+        onCancel={() => deleteModalIsOpen(false)}
+        onOk={handleDelete}
+        okText="Delete"
+        okType="danger"
+        centered
+      >
+        <p>
```
gilluminate: Should `Typography.paragraph` be used here? (Is this getting styled correctly?)

clients/admin-ui/src/features/system/SystemsTable.tsx
```diff
@@ -0,0 +1,444 @@
...
+  AntButton as Button,
+  AntColumnsType as ColumnsType,
+  AntDropdown as Dropdown,
+  AntFlex as Flex,
+  AntMessage as message,
+  AntModal as Modal,
+  AntTable as Table,
+  Icons,
+} from "fidesui";
+import { CustomTypography } from "fidesui/src/hoc";
```
gilluminate: This should be imported from the root "fidesui" with `AntTypography as Typography`

clients/admin-ui/src/features/system/SystemsTable.tsx
```diff
@@ -0,0 +1,444 @@
...
+  const handleFilterChange = (filters: Record<string, any>) => {
+    if (filters.data_steward) {
+      setDataStewardFilter(filters.data_steward[0]);
+    }
+    if (filters.system_groups) {
+      setSystemGroupFilter(filters.system_groups[0]);
+    }
+  };
+
+  const columns: ColumnsType<BasicSystemResponseExtended> = useMemo(() => {
```
gilluminate: lots of fixed widths, which we're trying to avoid/move away from. Can we remove or at least consult Jack on the needs for that?

clients/admin-ui/src/features/system/SystemsTable.tsx
```diff
@@ -0,0 +1,444 @@
...
+        key: group.fides_key,
+        label: group.name,
+        onClick: () => handleBulkAddToGroup(group.fides_key),
+      })) ?? []
+    );
+  }, [allSystemGroups, handleBulkAddToGroup]);
+
+  if (isLoading) {
+    return <TableSkeletonLoader rowHeight={36} numRows={15} />;
+  }
```
gilluminate: Ant has a built in loader we should be using instead of this old version.

clients/admin-ui/src/features/system/system-groups/components/CreateSystemGroupForm.tsx
```diff
@@ -0,0 +1,158 @@
+import { AntButton as Button, AntFlex as Flex } from "fidesui";
+import { CustomTypography } from "fidesui/src/hoc/CustomTypography";
```
gilluminate: We should import from root `AntTypography as Typography` like the other stuff. Note: You can also import individual components if you want `AntTitle as Title` etc.

clients/fidesui/src/palette/palette.module.scss
gilluminate: This looks like it might be because we're trying to use it for user defined values? If that's the case it DEFINITELY doesn't belong in the fidesui palette. There are other ways to handle this, including not using a CSS var at all, custom Admin UI only CSS vars, or even using some CSS specificity to override the defaults. This is the wrong place to handle it though.

clients/admin-ui/src/features/system/system-groups/components/DataUseSelectWithSuggestions.tsx
```diff
@@ -0,0 +1,111 @@
+import {
+  AntDefaultOptionType as DefaultOptionType,
+  AntFlex as Flex,
+  AntSelect as Select,
+  FormControl,
```
gilluminate: Any way to avoid using this Chakra `FormControl`, or is this a necessary evil while we're between Form migration?

clients/package-lock.json
```diff
@@ -27906,6 +27906,21 @@
...
+      "cpu": [
+        "ia32"
+      ],
+      "optional": true,
+      "os": [
+        "win32"
+      ],
+      "engines": {
+        "node": ">= 10"
+      }
```
gilluminate: Should probably remove this from the PR

clients/fidesui/src/palette/palette.module.scss
jpople: It's partially user-defined but also needs to render in the app as tag colors, which is why I put it here-- we're giving a finite limited selection of palette colors to the user and then passing those colors into the tag component so they render in the table.  I can handle this on map in the admin UI but it felt weird to tie, e.g. `red` here to specifically `error`.

clients/admin-ui/src/features/system/SystemsTable.tsx
jpople: Are you referring to the table's `loading` prop or are we using their actual Skeleton component for tables also?

clients/admin-ui/src/features/system/system-groups/components/DataUseSelectWithSuggestions.tsx
jpople: Removed that, good catch.

clients/admin-ui/src/features/system/SystemsTable.tsx
gilluminate: table's `loading` prop

clients/admin-ui/src/features/system/system-groups/components/CreateSystemGroupForm.tsx
```diff
@@ -0,0 +1,157 @@
+import { AntButton as Button, AntFlex as Flex, AntTypography } from "fidesui";
```
gilluminate, line 1:
```suggestion
import { AntButton as Button, AntFlex as Flex, AntTypography as Typography } from "fidesui";
```

clients/admin-ui/src/features/system/useSystemsTable.tsx
```diff
@@ -0,0 +1,396 @@
...
+      },
+      {
+        dataIndex: "system_groups",
+        key: "system_groups",
+        render: (
+          systemGroups: string[] | undefined,
+          record: BasicSystemResponseExtended,
+        ) => (
+          <SystemGroupCell
+            // @ts-ignore - TS doesn't know we filter out undefined
```
gilluminate: I removed this locally just to see what it was doing and see if I could recommend a fix. However, it doesn't seem necessary anymore and looks like it can be removed.
```suggestion
```

clients/admin-ui/src/features/system/SystemActionsMenu.tsx
```diff
@@ -0,0 +1,115 @@
...
+import { useGetAllUsersQuery } from "~/features/user-management";
+import { isErrorResult } from "~/types/errors";
+
+interface SystemActionsMenuProps {
+  selectedRowKeys: React.Key[];
+}
+
+const SystemActionsMenu = ({ selectedRowKeys }: SystemActionsMenuProps) => {
+  const [messageApi, contextHolder] = message.useMessage();
+  const [deleteModal, deleteModalIsOpen] = useState(false);
```
gilluminate:
```suggestion
  const [deleteModalIsOpen, setDeleteModalIsOpen] = useState(false);
```

clients/fidesui/src/palette/palette.module.scss
gilluminate: I would argue that `error` is the name of the color. We're intentionally avoiding names like `red` because they aren't scalable (what about another shade of red we might add later?). I would still recommend just getting rid of these altogether and using the existing vars. If you feel really strongly about it, I would at least change these names from `bg-taxonomy-red` to something like `bg-tags-red` or even `bg-supplemental-red` so that it's reusable.

clients/admin-ui/src/features/system/SystemsTable.tsx
```diff
@@ -0,0 +1,128 @@
...
+              </Modal>
+            </>
+          )}
+          <SystemActionsMenu
+            selectedRowKeys={selectionProps?.selectedRowKeys ?? []}
+          />
+          <Modal
+            open={deleteModalIsOpen}
+            onCancel={() => setDeleteModalIsOpen(false)}
+            onOk={() => handleDelete(selectedSystemForDelete!)}
```
gilluminate: I think we should ban non-null assertion in our eslint, but that's a discussion for another time 😄 
```suggestion
            onOk={() =>
              !!selectedSystemForDelete && handleDelete(selectedSystemForDelete)
            }
```

clients/admin-ui/src/features/system/useSystemsTable.tsx
```diff
@@ -0,0 +1,396 @@
...
+            }
+            allGroups={allSystemGroups!}
+            system={{ ...record, system_groups: systemGroups ?? [] }}
+            columnState={{
+              isWrapped: true,
+              isExpanded: isGroupsExpanded,
+            }}
+          />
+        ),
+        width: 400,
```
gilluminate: Still need to clean up fixed widths on columns as well. It's weird to see 400 width when the only system I have is `6sense` for example.

clients/admin-ui/src/features/common/table/cells/TagExpandableCell.tsx
```diff
@@ -1,6 +1,7 @@
 import {
   AntButton as Button,
   AntFlex as Flex,
+  AntFlexProps,
```
gilluminate:
```suggestion
  AntFlexProps as FlexProps,
```

clients/admin-ui/src/features/system/SystemActionsMenu.tsx
```diff
@@ -0,0 +1,115 @@
+import {
+  AntButton as Button,
+  AntDropdown as Dropdown,
+  AntMessage as message,
+  AntModal as Modal,
+  AntTypography,
```
gilluminate:
```suggestion
  AntTypography as Typography,
```

clients/admin-ui/src/features/system/SystemsTable.tsx
```diff
@@ -0,0 +1,130 @@
+import {
+  AntButton as Button,
+  AntDropdown as Dropdown,
+  AntFlex as Flex,
+  AntModal as Modal,
+  AntTable as Table,
+  AntTypography,
```
gilluminate:
```suggestion
  AntTypography as Typography,
```

clients/admin-ui/src/features/system/useSystemsTable.tsx
```diff
@@ -0,0 +1,400 @@
+import {
+  AntButton as Button,
+  AntColumnsType as ColumnsType,
+  AntDropdown as Dropdown,
+  AntMessage as message,
+  AntTypography,
```
gilluminate:
```suggestion
  AntTypography as Typogrpahy,
```

gilluminate (approved):
Looking great. Tested locally and everything is working as expected. Just a few nit-picks on the import consistency and it's good to go!

PR #6474, opened 2025-08-18
[ENG-1200]

JadeCara:
Closes [ENG-1200](https://ethyca.atlassian.net/browse/ENG-1200?atlOrigin=eyJpIjoiMjg1NDdkYTQ1OWIxNGUxNzhmZDlkZGE4ZjVmZTZmN2QiLCJwIjoiaiJ9)

### Description Of Changes

This PR introduces the basis for a streaming storage service that provides cloud-agnostic interfaces for GCS and S3 storage operations. The new infrastructure enables memory-efficient processing of large datasets with parallel processing, automatic package splitting, and robust error handling.

Updated requirements with a new package - stream_zip. This package handles the headers etc for zip uploads. 

**What Was Added**

🏗️ Complete Streaming Infrastructure
- streaming_storage.py (42KB) - Core streaming logic with package splitting, parallel processing, and memory management
- cloud_storage_client.py - Abstract base class defining cloud storage operations interface
- storage_client_factory.py - Factory pattern for creating cloud-specific storage clients
- schemas.py - Comprehensive Pydantic schemas for configuration and data structures
- dsr_storage.py - Specialized storage handling for DSR (Data Subject Rights) reports

☁️ Cloud Provider Implementations 
- s3/ - Complete AWS S3 implementation with multipart uploads and streaming

**Key Features** 

🚀 Performance & Scalability
- Memory-efficient processing - Limited to ~5.6MB regardless of attachment count
- Automatic package splitting - Intelligent distribution of large datasets across multiple packages

🛡️ Reliability & Error Handling
- Multipart upload support - Robust handling of large files with automatic retry logic
- Exponential backoff - Intelligent retry mechanisms for transient failures
- Graceful degradation - Fallback mechanisms when optimal configurations aren't available

🔧 Developer Experience
- Type-safe interfaces - Full Pydantic validation for all configuration and data structures
- Cloud-agnostic design - Unified interface across different storage providers
- Backward compatibility - Existing integrations continue to work without modification
- Extensible architecture - Easy addition of new cloud providers or storage types

**Technical Architecture**
Core Components:
- CloudStorageClient - Abstract interface defining storage operations
- StorageClientFactory - Factory pattern for cloud-specific client creation
- StreamingStorage - Main orchestrator for data processing and upload

**Data Flow**
```bash
Input Data → Package Splitting → Streaming ZIP Creation → Cloud Upload
        ↓              ↓                                   ↓                    ↓
Validation →   Optimization →     Memory Buffers →        Multipart Upload
```

### Code Changes

* the new streaming dir
* All the tests

### Steps to Confirm

1.  in fides - set the toml to include: 
```bash
[execution]
use_dsr_3_0 = true

[security]
subject_request_download_ui_enabled = true
```
2. Run fidesplus pointed at this branch
3. set up S3 Default storage config with the new `enable_streaming` using `PUT  /api/v1/storage/default`
```bash
 {
  "type": "s3",
    "details": {
      "auth_method": "secret_keys",
      "bucket": "fides-data-test-ethyca-demos",
      "enable_streaming": true
    },
    "format": "json"
  }
```
4. Create a manual Task with attachments - right now all regular DSR attachments are marked as internal use only.
5. Remember to set S3 as the default storage in the config.
6. Create a privacy request - make sure it has some attachments one way or another. 
Once it is completed you should see the logs:
```bash
idesplus-slim  | 2025-08-20 22:06:38.548 | INFO     | fides.api.service.storage.streaming.s3.streaming_s3:upload_to_s3_streaming:90 - Successfully uploaded streaming archive to S3: pri_bb0f3dd6-2874-4dd6-94dd-7a5a1095c53d.json | {'privacy_request_id': 'pri_bb0f3dd6-2874-4dd6-94dd-7a5a1095c53d', 'task_id': 'req_873a0de1-65d6-4cdd-bc1a-b5d4b65fdcaa', 'privacy_request_source': 'Request Manager'}
fidesplus-slim  | 2025-08-20 22:06:38.554 | INFO     | fides.api.service.privacy_request.request_runner_service:upload_access_results:245 - Access package upload successful for privacy request. | {'privacy_request_id': 'pri_bb0f3dd6-2874-4dd6-94dd-7a5a1095c53d', 'task_id': 'req_873a0de1-65d6-4cdd-bc1a-b5d4b65fdcaa', 'privacy_request_source': 'Request Manager', 'time_taken': 1.157496452331543}
fidesplus-slim  | 2025-08-20 22:06:38.586 | INFO     | fides.api.service.privacy_request.request_runner_service:run_privacy_request:712 - Marking privacy request 'pri_bb0f3dd6-2874-4dd6-94dd-7a5a1095c53d' as complete. | {'privacy_request_id': 'pri_bb0f3dd6-2874-4dd6-94dd-7a5a1095c53d', 'task_id': 'req_873a0de1-65d6-4cdd-bc1a-b5d4b65fdcaa', 'privacy_request_source': 'Request Manager'}
```
7. Grab the DSR id and use it to `GET /api/v1/privacy-request/{privacy_request_id}/access-results`, use the presigned url to download. 
You should see the attachments in the zip
Screenshot 2025-08-20 at 4 08 55 PM

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:

[ENG-1200]: https://ethyca.atlassian.net/browse/ENG-1200?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/fides/api/service/storage/streaming/s3/s3_storage_client.py
+++ b/src/fides/api/service/storage/streaming/s3/s3_storage_client.py
@@ -61,6 +61,10 @@ class S3StorageClient(BaseStorageClient):
             ]
         if StorageSecrets.REGION_NAME in self.storage_secrets:
             params["region"] = self.storage_secrets[StorageSecrets.REGION_NAME]
+        if StorageSecrets.AWS_ASSUME_ROLE in self.storage_secrets:
+            params["assume_role_arn"] = self.storage_secrets[
+                StorageSecrets.AWS_ASSUME_ROLE
+            ]
 
         return params
 
@@ -97,7 +101,12 @@ class S3StorageClient(BaseStorageClient):
             else:
                 auth_method = AWSAuthMethod.AUTOMATIC.value
 
-            s3_client = get_s3_client(auth_method, s3_secrets)
+            # Extract assume_role_arn if present for role assumption
+            assume_role_arn = None
+            if StorageSecrets.AWS_ASSUME_ROLE in self.storage_secrets:
+                assume_role_arn = self.storage_secrets[StorageSecrets.AWS_ASSUME_ROLE]
+
+            s3_client = get_s3_client(auth_method, s3_secrets, assume_role_arn)
             return create_presigned_url_for_s3(s3_client, bucket, key, ttl_seconds)
         except Exception as e:
             logger.error(f"Failed to generate S3 presigned URL: {e}")
--- a/src/fides/api/service/storage/streaming/s3/streaming_s3.py
+++ b/src/fides/api/service/storage/streaming/s3/streaming_s3.py
@@ -86,6 +86,67 @@ def upload_to_s3_streaming(
         )
 
 
+def _process_storage_secrets_input(
+    storage_secrets: Union[StorageSecretsS3, dict[StorageSecrets, Any]]  # type: ignore[misc]
+) -> dict[str, Any]:
+    """Process input and convert to string-keyed dictionary."""
+    final_secrets: dict[str, Any] = {}
+
+    if isinstance(storage_secrets, StorageSecretsS3):
+        # Convert StorageSecretsS3 model directly to string keys
+        for key, value in storage_secrets.model_dump().items():
+            if value is not None:
+                final_secrets[key] = value
+    else:
+        # Process dict input, converting enum keys to strings if needed
+        for key, value in (storage_secrets or {}).items():  # type: ignore[assignment]
+            if isinstance(key, str):
+                final_secrets[key] = value
+            elif isinstance(key, StorageSecrets):
+                final_secrets[key.value] = value
+
+    return final_secrets
+
+
+def _validate_aws_credentials(final_secrets: dict[str, Any]) -> None:
+    """Validate that required AWS credentials are present."""
+    has_access_key = (
+        "aws_access_key_id" in final_secrets and final_secrets["aws_access_key_id"]
+    )
+    has_secret_key = (
+        "aws_secret_access_key" in final_secrets
+        and final_secrets["aws_secret_access_key"]
+    )
+
+    # If we have any AWS credentials, we need both for SECRET_KEYS auth
+    if has_access_key or has_secret_key:
+        if not has_access_key:
+            raise ValueError(
+                "Missing required AWS credentials for SECRET_KEYS auth: aws_access_key_id. "
+                "Storage configuration must include valid AWS access key and secret key."
+            )
+        if not has_secret_key:
+            raise ValueError(
+                "Missing required AWS credentials for SECRET_KEYS auth: aws_secret_access_key. "
+                "Storage configuration must include valid AWS access key and secret key."
+            )
+    else:
+        # AUTOMATIC authentication - check if region is provided
+        has_region = "region_name" in final_secrets and final_secrets["region_name"]
+        if not has_region:
+            raise ValueError(
+                "Missing required region_name for AUTOMATIC authentication. "
+                "Storage configuration must include a valid AWS region."
+            )
+
+
+def _set_default_region(final_secrets: dict[str, Any]) -> None:
+    """Set default region if missing."""
+    if "region_name" not in final_secrets or not final_secrets["region_name"]:
+        logger.debug("Setting default region to 'us-east-1'")
+        final_secrets["region_name"] = "us-east-1"
+
+
 def format_secrets(
     storage_secrets: Union[StorageSecretsS3, dict[StorageSecrets, Any]]  # type: ignore[misc]
 ) -> dict[str, Any]:
@@ -95,14 +156,19 @@ def format_secrets(
     This function handles multiple credential formats and processes them through several stages:
     1. Input processing: Accepts either StorageSecretsS3 models (from API) or raw dicts (from database)
     2. Key normalization: Converts all keys to string format for consistency
-    3. Default setting: Sets default region if missing
-    4. Validation: Ensures all required AWS credentials are present
+    3. Validation: Ensures required AWS credentials are present based on auth method
+    4. Default setting: Sets default region if missing (after validation)
     5. Return: Returns string-keyed dict ready for S3StorageClient
 
     Input formats:
     - StorageSecretsS3: Used by storage API endpoints (e.g., put_config_secrets)
     - dict[StorageSecrets, Any]: Used by storage models (e.g., StorageConfig.secrets)
     - dict[str, Any]: Used by database queries (e.g., StorageConfig.get_by)
+
+    Authentication methods:
+    - SECRET_KEYS: Requires aws_access_key_id, aws_secret_access_key, and region_name
+    - AUTOMATIC: Requires only region_name (relies on AWS SDK credential chain)
+    - Role assumption: Can be used with either auth method via assume_role_arn
 
     Output format:
     - dict[str, Any]: Required by S3StorageClient.generate_presigned_url() and other AWS operations
@@ -111,53 +177,18 @@ def format_secrets(
         dict[str, Any]: Credentials with string keys like 'aws_access_key_id'
 
     Raises:
-        ValueError: If required AWS credentials are missing
+        ValueError: If required AWS credentials are missing for the chosen auth method
     """
     logger.debug("format_secrets called with type: {}", type(storage_secrets).__name__)
 
     # Stage 1: Process input and create final format directly
-    # We'll build the final string-keyed dictionary from the start
-    final_secrets: dict[str, Any] = {}
+    final_secrets = _process_storage_secrets_input(storage_secrets)
 
-    if isinstance(storage_secrets, StorageSecretsS3):
-        # Convert StorageSecretsS3 model directly to string keys
-        # This handles input from storage API endpoints (e.g., put_config_secrets)
-        for key, value in storage_secrets.model_dump().items():
-            if value is not None:
-                final_secrets[key] = value
-    else:
-        # Process dict input, converting enum keys to strings if needed
-        # This handles input from StorageConfig.secrets (database storage)
-        for key, value in (storage_secrets or {}).items():  # type: ignore[assignment]
-            if isinstance(key, str):
-                final_secrets[key] = value
-            elif isinstance(key, StorageSecrets):
-                # Convert enum key to string key
-                final_secrets[key.value] = value
+    # Stage 2: Validate required credentials BEFORE setting defaults
+    _validate_aws_credentials(final_secrets)
 
-    # Stage 2: Set default region if missing
-    # AWS S3 operations require a region, so we provide a sensible default
-    # This is needed for S3StorageClient.generate_presigned_url() and other AWS calls
-    if "region_name" not in final_secrets or not final_secrets["region_name"]:
-        logger.debug("Setting default region to 'us-east-1'")
-        final_secrets["region_name"] = "us-east-1"
-
-    # Stage 3: Validate required credentials
-    # Ensure all necessary AWS credentials are present before proceeding
-    # These are required by S3StorageClient and AWS SDK operations
-    required_secrets = ["aws_access_key_id", "aws_secret_access_key", "region_name"]
-
-    missing_secrets = [
-        secret
-        for secret in required_secrets
-        if secret not in final_secrets or not final_secrets[secret]
-    ]
-
-    if missing_secrets:
-        raise ValueError(
-            f"Missing required AWS credentials: {missing_secrets}. "
-            "Storage configuration must include valid AWS access key, secret key, and region."
-        )
+    # Stage 3: Set default region if missing (after validation)
+    _set_default_region(final_secrets)
 
     logger.debug(
         "format_secrets completed successfully with {} keys", len(final_secrets)
--- a/src/fides/api/service/storage/streaming/smart_open_streaming_storage.py
+++ b/src/fides/api/service/storage/streaming/smart_open_streaming_storage.py
@@ -6,6 +6,7 @@ import csv
 import json
 from datetime import datetime
 from io import BytesIO, StringIO
+from itertools import chain
 from typing import Any, Generator, Iterable, Optional, Tuple
 from urllib.parse import urlparse
 
@@ -688,8 +689,6 @@ class SmartOpenStreamingStorage:
         attachment_files_generator = self._create_attachment_files(all_attachments)
 
         # Combine both generators and stream the complete ZIP to storage
-        from itertools import chain
-
         combined_entries = chain(attachment_files_generator, dsr_files_generator)
         with self.storage_client.stream_upload(
             config.bucket_name,
--- a/tests/ops/service/storage/streaming/s3/test_s3_storage_client.py
+++ b/tests/ops/service/storage/streaming/s3/test_s3_storage_client.py
@@ -68,6 +68,37 @@ class TestS3StorageClient:
         assert "secret_key" not in params
         # endpoint_url is no longer supported in the current implementation
 
+    def test_get_transport_params_with_assume_role_arn(self):
+        """Test transport params include assume_role_arn when present."""
+        secrets = {
+            StorageSecrets.AWS_ACCESS_KEY_ID: "test_key",
+            StorageSecrets.AWS_SECRET_ACCESS_KEY: "test_secret",
+            StorageSecrets.REGION_NAME: "us-west-2",
+            StorageSecrets.AWS_ASSUME_ROLE: "arn:aws:iam::123456789012:role/TestRole",
+        }
+        client = S3StorageClient(secrets)
+        params = client.get_transport_params()
+
+        assert params["access_key"] == "test_key"
+        assert params["secret_key"] == "test_secret"
+        assert params["region"] == "us-west-2"
+        assert params["assume_role_arn"] == "arn:aws:iam::123456789012:role/TestRole"
+
+    def test_get_transport_params_without_assume_role_arn(self):
+        """Test transport params don't include assume_role_arn when not present."""
+        secrets = {
+            StorageSecrets.AWS_ACCESS_KEY_ID: "test_key",
+            StorageSecrets.AWS_SECRET_ACCESS_KEY: "test_secret",
+            StorageSecrets.REGION_NAME: "us-west-2",
+        }
+        client = S3StorageClient(secrets)
+        params = client.get_transport_params()
+
+        assert params["access_key"] == "test_key"
+        assert params["secret_key"] == "test_secret"
+        assert params["region"] == "us-west-2"
+        assert "assume_role_arn" not in params
+
     @patch("fides.api.service.storage.streaming.s3.s3_storage_client.get_s3_client")
     @patch(
         "fides.api.service.storage.streaming.s3.s3_storage_client.create_presigned_url_for_s3"
@@ -129,7 +160,7 @@ class TestS3StorageClient:
         assert result == "https://test-url.com"
         # Verify get_s3_client was called with SECRET_KEYS auth method
         mock_get_s3_client.assert_called_once_with(
-            AWSAuthMethod.SECRET_KEYS.value, secrets
+            AWSAuthMethod.SECRET_KEYS.value, secrets, None
         )
 
     @patch("fides.api.service.storage.streaming.s3.s3_storage_client.get_s3_client")
@@ -153,7 +184,7 @@ class TestS3StorageClient:
         assert result == "https://test-url.com"
         # Verify get_s3_client was called with AUTOMATIC auth method
         mock_get_s3_client.assert_called_once_with(
-            AWSAuthMethod.AUTOMATIC.value, secrets
+            AWSAuthMethod.AUTOMATIC.value, secrets, None
         )
 
     @patch("fides.api.service.storage.streaming.s3.s3_storage_client.get_s3_client")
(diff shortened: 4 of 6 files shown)


src/fides/api/util/aws_util.py
```diff
@@ -12,7 +12,7 @@
 
 def get_aws_session(
     auth_method: str,
-    storage_secrets: Optional[Dict[StorageSecrets, Any]],
+    storage_secrets: Optional[dict[StorageSecrets, Any]],
```
JadeCara:
```suggestion
    storage_secrets: Optional[Dict[StorageSecrets, Any]],
```

src/fides/api/util/aws_util.py
```diff
@@ -87,7 +87,7 @@ def get_assumed_role_session(
 
 def get_s3_client(
     auth_method: str,
-    storage_secrets: Optional[Dict[StorageSecrets, Any]],
+    storage_secrets: Optional[StorageSecrets],
```
JadeCara:
```suggestion
    storage_secrets: Optional[Dict[StorageSecrets, Any]],
```

tests/ops/tasks/test_storage.py
```diff
@@ -142,7 +142,7 @@ def test_write_to_in_memory_buffer_invalid_format(self):
     @patch("fides.api.tasks.storage.DsrReportBuilder")
     def test_write_to_in_memory_buffer_html(self, mock_dsr_builder):
         """Test HTML format generation using DsrReportBuilder."""
-        data = {"key": "value"}
+        data = {"collection1": [{"id": 1, "name": "test data"}]}
```
JadeCara:
```suggestion
        data = {"key": "value"}
```

src/fides/api/service/storage/streaming/schemas.py
```diff
@@ -0,0 +1,175 @@
...
+        default=5 * 1024 * 1024,
+        ge=1024 * 1024,
+        description="Zip buffer threshold in bytes (default: 5MB)",
+    )
+    stream_buffer_threshold: int = Field(
+        default=512 * 1024,
+        ge=1024 * 1024,
+        description="Stream buffer threshold in bytes (default: 512KB)",
+    )
+    chunk_size_threshold: int = Field(
```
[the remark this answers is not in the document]
JadeCara: Updated.

.fides/fides.toml
```diff
@@ -42,14 +42,15 @@ oauth_root_client_id = "fidesadmin"
 oauth_root_client_secret = "fidesadminsecret"
 drp_jwt_secret = "secret"
 env = "prod"
+subject_request_download_ui_enabled = true
```
JadeCara:
```suggestion
```

.fides/fides.toml
```diff
@@ -42,14 +42,15 @@ oauth_root_client_id = "fidesadmin"
...
 
 [execution]
 require_manual_request_approval = false
 task_retry_backoff = 1
 subject_identity_verification_required = false
 task_retry_count = 0
 task_retry_delay = 1
-use_dsr_3_0 = false
+use_dsr_3_0 = true
```
JadeCara:
```suggestion
use_dsr_3_0 = false
```

src/fides/api/service/privacy_request/dsr_package/dsr_report_builder.py
```diff
@@ -21,6 +20,9 @@
 HEADER_COLOR = "#FAFAFA"
 BORDER_COLOR = "#E2E8F0"
 
+if TYPE_CHECKING:
+    from fides.api.models.privacy_request import PrivacyRequest  # pragma: no cover
```
erosselli, line 24: is this because of a circular dependency?

src/fides/api/service/storage/streaming/cloud_storage_client.py
```diff
@@ -0,0 +1,82 @@
+# pylint: disable=unnecessary-pass
```
erosselli: FWIW I think if your function has a docstring and no body then you don't need the `pass` statement

src/fides/api/service/storage/streaming/dsr_storage.py
```diff
@@ -0,0 +1,128 @@
...
+            "Completed HTML DSR report multipart upload with {} parts", len(parts)
+        )
+
+    except Exception as e:
+        # Abort multipart upload on failure
+        try:
+            storage_client.abort_multipart_upload(bucket_name, file_key, upload_id)
+        except Exception as abort_error:
+            logger.warning("Failed to abort multipart upload: {}", abort_error)
+        raise e
```
tvandort: Is there an "Aggregate exception" in python so that we can throw both errors in a wrapper? Might be confusing to see one thing logged and then a different exception raised.

Alternatively should we just raise abort_error instead?

src/fides/api/service/storage/streaming/dsr_storage.py
```diff
@@ -0,0 +1,128 @@
...
+from fides.api.service.storage.streaming.util import (
+    AWS_MIN_PART_SIZE,
+    CHUNK_SIZE_THRESHOLD,
+)
+
+if TYPE_CHECKING:
+    from fides.api.models.privacy_request import PrivacyRequest
+
+
+def stream_html_dsr_report_to_storage_multipart(
```
erosselli: nit: it feels like it might make more sense for this function to already receive the `dsr_buffer` as an argument rather than having to generate the DSR report ? I think it keeps the responsibilities more clear  / implements a separation of concerns

src/fides/api/service/storage/streaming/dsr_storage.py
```diff
@@ -0,0 +1,128 @@
...
+    html_content = dsr_buffer.getvalue()
+    content_size = len(html_content)
+
+    # For small files, use single upload to avoid AWS S3 part size validation errors
+    if content_size < AWS_MIN_PART_SIZE:
+        logger.info(
+            "DSR report size {} bytes is below multipart threshold ({} bytes), using single upload",
+            content_size,
+            AWS_MIN_PART_SIZE,
+        )
```
erosselli: since this is AWS-specific, it feels like this logic would be better places within the s3_storage_client class ? We could keep the "common" parts of the logic in this generic method (or even a method in the base class) but call a `upload_dsr_report` method (name is just an example, could be different) that will be client-specific 

I know that essentially this is the only client we support at the moment but I think it would set us up nicely to implement other clients in the future

src/fides/api/service/storage/streaming/dsr_storage.py
```diff
@@ -0,0 +1,128 @@
...
+                    # Extend this chunk to include part of the next chunk
+                    extended_size = min(AWS_MIN_PART_SIZE, content_size - i)
+                    chunk = html_content[i : i + extended_size]
+                    # Adjust the next iteration to account for the overlap
+                    i = i + extended_size - chunk_size
+
+            part = storage_client.upload_part(
+                bucket_name, file_key, upload_id, part_number, chunk
+            )
+            parts.append(part)
```
erosselli: same comment here about all this being AWS-specific

src/fides/api/service/storage/streaming/retry.py
```diff
@@ -0,0 +1,360 @@
+# pylint: disable=unnecessary-pass
```
erosselli: same comment about the pass statements

src/fides/api/service/storage/streaming/retry.py
```diff
@@ -0,0 +1,360 @@
...
+        "gateway timeout",
+        "request timeout",
+        "connection reset",
+        "broken pipe",
+    ]
+
+    return any(pattern in error_str for pattern in transient_patterns)
+
+
+def is_s3_transient_error(error: Exception) -> bool:
```
erosselli: same here, I think it might make more sense to scope this in the s3 storage class . It could just override the base classes `is_transient_error` , add the custom logic and add a call to `super().is_transient_error` as the fallback case

src/fides/api/service/storage/streaming/retry.py
```diff
@@ -0,0 +1,360 @@
...
+    ):
+        self.max_retries = max_retries
+        self.base_delay = base_delay
+        self.max_delay = max_delay
+        self.backoff_factor = backoff_factor
+        self.jitter = jitter
+        self.retry_on_exceptions = retry_on_exceptions or (Exception,)
+
+
+def is_transient_error(error: Exception) -> bool:
```
erosselli, line 45: this could be a method in the base storage class

src/fides/api/service/storage/streaming/s3/s3_schemas.py
```diff
@@ -0,0 +1,300 @@
...
+    bucket: Annotated[
+        str,
+        Field(
+            ..., description="S3 bucket name where the multipart upload was initiated"
+        ),
+    ]
+    key: Annotated[
+        str,
+        Field(..., description="S3 object key (file path) for the multipart upload"),
+    ]
```
erosselli: can we define common fields like `bucket` and `key`, along with their validators, in a base `BaseAWSUploadRequest` model and have all the other schemas inherit from it? there's quite a lot of repetition here , 5 `validate_bucket` and `validate_key` that are all the same

src/fides/api/service/storage/streaming/schemas.py
```diff
@@ -0,0 +1,175 @@
...
+            raise ValueError("Upload ID cannot be empty or whitespace")
+        return v.strip()
+
+    class Config:
+        """Pydantic model configuration"""
+
+        # Allow extra fields from storage providers
+        extra = "allow"
+        # Use case-insensitive field names
+        case_sensitive = False
```
erosselli: nit: we usually use a config dict for this right? not sure which way is the pydantic 2.0-style so feel free to keep this as-is if this is the better/newer approach

src/fides/api/service/storage/streaming/util.py
```diff
@@ -0,0 +1,29 @@
+from typing import Any, Union
+
+from fides.api.schemas.storage.storage import StorageSecrets, StorageSecretsS3
+
+# AWS S3 multipart upload requirements
+AWS_MIN_PART_SIZE = 5 * 1024 * 1024  # 5MB minimum per part (except last)
+ZIP_BUFFER_THRESHOLD = 5 * 1024 * 1024  # 5MB threshold for zip buffer uploads
+CHUNK_SIZE_THRESHOLD = 5 * 1024 * 1024  # 5MB threshold for chunk-based uploads
+LARGE_FILE_THRESHOLD = 25 * 1024 * 1024  # 25MB threshold for large file handling
+
```
erosselli: can we define these in the s3 storage client class, or at least in that same file?

tests/ops/service/storage/streaming/conftest.py
```diff
@@ -0,0 +1,107 @@
...
+
+from fides.api.service.storage.streaming.cloud_storage_client import CloudStorageClient
+from fides.api.service.storage.streaming.schemas import (
+    MultipartUploadResponse,
+    UploadPartResponse,
+)
+
+
+@pytest.fixture
+def mock_storage_client():
```
erosselli: nice!

erosselli (commented):
thanks for adding so many tests! I left some comments and questions. Codecov's reporting 230 uncovered lines which feels like it might be a bit too many

src/fides/api/service/privacy_request/dsr_package/dsr_report_builder.py
JadeCara: It is in this case yes :)

src/fides/api/service/storage/streaming/util.py
JadeCara: I will move the AWS one over for sure - the others I think are general zip things. I can move them to the client or the factory though! and fix the comment string.

src/fides/api/service/storage/streaming/dsr_storage.py
JadeCara: Thanks @tvandort Smart Open is cool looking at adding it - seeing if we can clear out some plumbing code. It will probably be a little bit hybrid-y just because we have some specific dsr package things, but I think we can use it in a lot of places.

.fides/fides.toml
JadeCara:
```suggestion
```

.fides/fides.toml
JadeCara:
```suggestion
use_dsr_3_0 = false
```

.fides/fides.toml
JadeCara:
```suggestion
```

.fides/fides.toml
JadeCara:
```suggestion
use_dsr_3_0 = false
```

src/fides/api/service/storage/storage_uploader_service.py
```diff
@@ -44,22 +45,15 @@ def upload(
     return uploader(db, config, data, privacy_request)
 
 
-def get_extension(resp_format: ResponseFormat) -> str:
+def get_extension(resp_format: ResponseFormat, has_attachments: bool = False) -> str:
```
erosselli: what is `has_attachments` used for ?

src/fides/api/service/storage/storage_uploader_service.py
```diff
@@ -88,13 +82,32 @@ def _s3_uploader(
     data: Dict,
     privacy_request: PrivacyRequest,
 ) -> str:
-    """Constructs necessary info needed for s3 before calling upload"""
+    """Constructs necessary info needed for s3 before calling upload using streaming approach"""
```
erosselli:
```suggestion
    """
    Constructs necessary info needed for s3 before calling upload.
    If `enable_streaming` is configured in the storage config, we use a streaming approach for better memory efficiency. 
    Otherwise we fall back to the traditional upload method.
    """
```

src/fides/api/service/storage/storage_uploader_service.py
```diff
@@ -88,13 +82,32 @@ def _s3_uploader(
...
     bucket_name = config.details[StorageDetails.BUCKET.value]
     auth_method = config.details[StorageDetails.AUTH_METHOD.value]
     document = None
 
+    # Check if streaming is enabled in the storage configuration
+    enable_streaming = config.details.get(StorageDetails.ENABLE_STREAMING.value, False)
+
+    if enable_streaming:
+        # Use streaming upload for better memory efficiency
+        logger.info("Using streaming S3 upload for {}", file_key)
```
erosselli: nit: should these be debug logs? it seems like they might clutter up logging for clients who process a lot of DSRs?

src/fides/api/service/storage/streaming/retry.py
```diff
@@ -0,0 +1,353 @@
...
+
+    # Exponential backoff: base_delay * (backoff_factor ^ attempt)
+    delay = base_delay * (backoff_factor ** (attempt - 1))
+
+    # Cap at maximum delay
+    delay = min(delay, max_delay)
+
+    # Add jitter to prevent multiple retries from synchronizing
+    if jitter:
+        import random
```
erosselli: can we move this to a top-level import?

src/fides/api/service/storage/streaming/s3/streaming_s3.py
```diff
@@ -0,0 +1,204 @@
...
+from fides.api.service.storage.streaming.smart_open_client import SmartOpenStorageClient
+from fides.api.service.storage.streaming.smart_open_streaming_storage import (
+    SmartOpenStreamingStorage,
+)
+
+if TYPE_CHECKING:
+    from fides.api.models.privacy_request import PrivacyRequest
+
+
+@retry_cloud_storage_operation(
```
erosselli, line 21: I was expecting this to use retry_s3_operation, is that method no longer used?

src/fides/api/service/storage/streaming/s3/streaming_s3.py
```diff
@@ -0,0 +1,204 @@
...
+
+    This function now uses smart-open for efficient cloud storage operations while maintaining
+    our DSR-specific business logic for package splitting and attachment processing.
+    """
+    logger.info("Starting smart-open streaming S3 Upload of {}", file_key)
+    logger.info("Storage secrets type: {}", type(storage_secrets))
+    logger.debug("Storage secrets content: {}", storage_secrets)
+
+    if privacy_request is None:
+        raise ValueError("Privacy request must be provided")
```
erosselli: it feels like this check might belong somewhere outside this function? that way we separate concerns between s3 streaming logic and application business logic?

src/fides/api/service/storage/streaming/s3/streaming_s3.py
```diff
@@ -0,0 +1,204 @@
...
+
+    if isinstance(storage_secrets, StorageSecretsS3):
+        logger.info("Processing StorageSecretsS3 object")
```
erosselli: same comment here about log-level, these should probably be debugs.... or do we even need these? they aren't super specific and have no ids or anything so idk how helpful they'd be for debugging anyway

src/fides/api/service/storage/streaming/s3/streaming_s3.py
```diff
@@ -0,0 +1,204 @@
...
+        secrets_dict = update_storage_secrets_s3(storage_secrets)
+    else:
+        logger.info("Processing dict[StorageSecrets, Any] object")
+        # Convert dict[StorageSecrets, Any] to dict[str, Any] for SmartOpenStorageClient
+        secrets_dict = update_storage_secrets_s3(storage_secrets)
```
erosselli: also it seems like in both the if and the else statement we do the same thing?

src/fides/api/service/storage/streaming/s3/streaming_s3.py
```diff
@@ -0,0 +1,204 @@
...
+
+    if document is not None:
+        # Convert to dict[StorageSecrets, Any] for generic_upload_to_s3 compatibility
```
erosselli: can you maybe add a comment explaining what `document` is and how it related to the `generic_upload_to_s3` compatibility?

src/fides/api/service/storage/streaming/s3/streaming_s3.py
```diff
@@ -0,0 +1,204 @@
...
+        raise StorageUploadError(
+            f"Unexpected error during smart-open streaming upload: {e}"
+        )
+
+
+def update_storage_secrets_s3(
+    storage_secrets: Union[StorageSecretsS3, dict[StorageSecrets, Any]]
+) -> dict[str, Any]:
+    """
+    Updates the storage secrets to the expected format for SmartOpenStorageClient.
```
erosselli: do we want to actually update the storage secrets, or would it make more sense to just process them and convert them to the expected format without updating them ?

erosselli: actually I'm not sure I follow what this function is doing

src/fides/api/service/storage/streaming/smart_open_client.py
```diff
@@ -0,0 +1,296 @@
...
+                    # Bytes
+                    fout.write(body)
+                elif isinstance(body, str):
+                    # String
+                    fout.write(body.encode("utf-8"))
+                else:
+                    # Try to convert to string
+                    fout.write(str(body).encode("utf-8"))
+
+            logger.info(f"Successfully uploaded object to {uri}")
```
erosselli: same question here about log level :)

src/fides/api/service/storage/streaming/s3/streaming_s3.py
JadeCara: We have storage secrets passed around a couple of different ways StorageSecretsS3 and dict[StorageSecrets, Any] and this functionality lies irritatingly at the intersection.  I want to take this one out, but I need to verify which version I am receiving and which version needs to go where. I will clean it out though.

src/fides/api/service/storage/storage_uploader_service.py
JadeCara: Cleaned up these methods. Its not really used.

src/fides/api/service/storage/streaming/retry.py
JadeCara: ended up being able to get rid of the s3 versions so we can just use the base ones.

src/fides/api/service/storage/streaming/retry.py
JadeCara: I think it makes more sense in retry. The storage class is about storage and everything in retry is about error detection and handling. Since these are used in retries moving it to the client would also create circular dependencies.

src/fides/api/service/storage/streaming/s3/streaming_s3.py
JadeCara: Yep! Got it cleaned up!

src/fides/api/service/storage/streaming/s3/streaming_s3.py
JadeCara: absolutely agree - updated

src/fides/api/service/storage/streaming/s3/streaming_s3.py
```diff
@@ -0,0 +1,165 @@
...
+) -> dict[str, Any]:
+    """
+    Returns the correct format for the S3StorageClient.
+
+    This function handles multiple credential formats and processes them through several stages:
+    1. Input processing: Accepts either StorageSecretsS3 models (from API) or raw dicts (from database)
+    2. Key normalization: Converts all keys to string format for consistency
+    3. Default setting: Sets default region if missing
+    4. Validation: Ensures all required AWS credentials are present
+    5. Return: Returns string-keyed dict ready for S3StorageClient
```
erosselli: thanks for the detailed docstring!! ❤️

erosselli (approved):
thanks for making all the changes!

src/fides/api/service/storage/streaming/smart_open_client.py
```diff
@@ -0,0 +1,270 @@
...
+        base_delay=2.0,
+        max_delay=30.0,
+    )
+    def put_object(
+        self,
+        bucket: str,
+        key: str,
+        body: Any,
+        content_type: Optional[str] = None,
+        metadata: Optional[dict[str, str]] = None,
```
galvana: Are there plans for the unused `metadata` param? If not, we should get rid of it

src/fides/api/service/storage/streaming/smart_open_streaming_storage.py
```diff
@@ -0,0 +1,999 @@
...
+        with self.storage_client.stream_upload(
+            config.bucket_name,
+            config.file_key,
+            content_type="application/zip",
+        ) as upload_stream:
+            for chunk in stream_zip(attachment_files_generator):
+                upload_stream.write(chunk)
+
+            for chunk in stream_zip(dsr_files_generator):
+                upload_stream.write(chunk)
```
galvana: Each invocation of `stream_zip` will create a separate zip file, so calling this twice will result in two concatenated ZIP files being written to the upload stream. Apparently some clients might treat this as an invalid archive or only read the last ZIP file. It's preferred to `chain` the streams and write them out with a single call to `stream_zip`.
```suggestion
        from itertools import chain

        combined_entries = chain(attachment_files_generator, dsr_files_generator)
        with self.storage_client.stream_upload(
            config.bucket_name,
            config.file_key,
            content_type="application/zip",
        ) as upload_stream:
            for chunk in stream_zip(combined_entries):
                upload_stream.write(chunk)
```

src/fides/api/service/storage/streaming/s3/s3_storage_client.py
```diff
@@ -0,0 +1,95 @@
...
+        Raises:
+            Exception: If presigned URL generation fails
+        """
+        try:
+            # Storage secrets are already in the right format for get_s3_client
+            # get_s3_client expects dict[StorageSecrets, Any] with enum keys
+            s3_secrets = self.storage_secrets
+
+            # Use a default auth method since we only have enum keys now
+            auth_method = AWSAuthMethod.SECRET_KEYS.value
```
galvana: Is it safe to assume the `auth_method` will be secret keys?

galvana (changes requested):

src/fides/api/service/storage/streaming/s3/s3_storage_client.py
JadeCara: Oh no not at all! Thank you :)

src/fides/api/service/storage/streaming/smart_open_client.py
JadeCara: So - That *can* be used for making sure things in the file are in the right order. I think the tools we are using negate the need for that. I will remove it - but making a note here in case we need to revisit at some point!

src/fides/api/service/storage/streaming/smart_open_streaming_storage.py
```diff
@@ -688,15 +688,15 @@ def _handle_html_format_upload(
         attachment_files_generator = self._create_attachment_files(all_attachments)
 
         # Combine both generators and stream the complete ZIP to storage
+        from itertools import chain
```
galvana: We can keep this at the top of the file, I just put it in the snippet as an example

galvana (approved):

PR #6487, opened 2025-08-25
Privacy request redaction patterns

galvana:
Closes [ENG-1230](https://ethyca.atlassian.net/browse/ENG-1230)

### Description Of Changes

This PR implements a new privacy request redaction patterns feature that allows administrators to configure regex patterns for masking sensitive dataset, collection, and field names in DSR (Data Subject Request) package reports. 

The feature supports two redaction approaches:
1. **Global regex patterns** - Configurable patterns that apply across all datasets to match and redact names based on regex rules
2. **Entity-specific configurations** - Individual `fides_meta.redact: name` settings that can be applied to specific datasets, collections, or fields

The feature includes both backend infrastructure for storing and managing redaction patterns, as well as a complete frontend interface for administrators to configure these patterns through the Fides admin UI.

### Code Changes

**Global Patterns:**
* Added new database model `PrivacyRequestRedactionPatterns` as a single-row table to store global regex patterns
* Created database migration to add `privacy_request_redaction_patterns` table
* Implemented API endpoints for getting and updating global redaction patterns with proper OAuth scopes
* Created new admin UI page for managing global privacy request redaction patterns
* Added form interface with dynamic pattern input fields and add/remove functionality for regex patterns
* Added new OAuth scopes: `PRIVACY_REQUEST_REDACTION_PATTERNS_READ` and `PRIVACY_REQUEST_REDACTION_PATTERNS_UPDATE`

**Entity-Specific:**
* Enhanced DSR report builder to support entity-specific `fides_meta.redact: name` configurations
* Added hierarchical key matching for entity-specific redaction (dataset.collection.field format)
* Enhanced type definitions to support entity-specific redaction metadata in `FidesMeta`
* Updated fideslang package dependency to support entity-specific redaction metadata
* Implemented precedence logic where entity-specific configurations override global patterns

### Steps to Confirm

1. Start Fides, Postgres, and MongoDB with `nox -s dev -- postgres mongodb`
2. Run the `python qa redact_annotation_scenario setup` scenario to set up Postgres and MongoDB integrations. These integrations will have `fides_meta.redact: name` on a few collections and fields.
3. Submit a privacy request for `jane@example.com`
4. Open the DSR package and click around, you should see redacted field names in the form (`dataset_*`, `collection_*`, `field_*`)
5. Navigate to `Setting > Privacy Requests` and add `full_name` as an entry
6. Submit another request
7. The first item in the `flights` collection should now have a nested field with a redacted name (`field_1`)

8. Add a regex pattern under the redaction patterns (like `.*ard.*`) and submit another request. Name like "card" or "reward" should be redacted

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:

[ENG-1230]: https://ethyca.atlassian.net/browse/ENG-1230?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


clients/admin-ui/src/features/settings/PrivacyRequestRedactionPatternsPage.tsx
```diff
@@ -0,0 +1,160 @@
...
+        "Failed to update privacy request redaction patterns",
+      );
+    } else {
+      successAlert("Privacy request redaction patterns updated successfully.");
+    }
+  };
+
+  const initialValues: PrivacyRequestRedactionPatternsFormValues = {
+    patterns: (currentPatterns?.patterns || []).map((pattern, index) => ({
+      id: `pattern-${index}-${Date.now()}`,
```
erosselli: is this a pattern we already use , adding the `date.now` call?

clients/admin-ui/src/features/settings/PrivacyRequestRedactionPatternsPage.tsx
```diff
@@ -0,0 +1,160 @@
...
+interface PatternItem {
+  id: string;
+  value: string;
+}
+
+interface PrivacyRequestRedactionPatternsFormValues {
+  patterns: PatternItem[];
+}
+
+const ValidationSchema = Yup.object().shape({
```
erosselli, line 28: I don't really know how `Yup` works so I'm blindly trusting this validation code

clients/admin-ui/src/features/settings/PrivacyRequestRedactionPatternsPage.tsx
```diff
@@ -0,0 +1,160 @@
...
+    useGetPrivacyRequestRedactionPatternsQuery(undefined);
+  const [updatePatterns, { isLoading: isUpdating }] =
+    useUpdatePrivacyRequestRedactionPatternsMutation();
+
+  const handleSubmit = async (
+    values: PrivacyRequestRedactionPatternsFormValues,
+  ) => {
+    // Filter out empty patterns and trim whitespace
+    const cleanedPatterns = values.patterns
+      .filter((pattern) => pattern.value && pattern.value.trim().length > 0)
```
erosselli: is there a way to centralize this validation? since we also have it duplicated on `ValidationSchema` , maybe we can just define it as a `validatePatternNonEmpty` function or similar

clients/admin-ui/src/types/api/models/FidesMeta.ts
```diff
@@ -45,4 +45,8 @@ export type FidesMeta = {
    * Optionally specify a masking strategy override for this field.
    */
   masking_strategy_override?: FieldMaskingStrategyOverride | null;
+  /**
+   * Optionally specify that this entity's name should be redacted in DSR reports. When set to "name", the entity will be redacted using the indexed format (e.g., "dataset_1", "collection_1", "field_1").
+   */
+  redact?: "name" | null;
```
erosselli, line 51: should this use an enum e.g `RedactionStrategy` or similar ? to be consistent with the other fields, e.g  `masking_strategy_override`

src/fides/api/alembic/migrations/versions/f457a75bc0e1_adding_redaction_patterns.py
```diff
@@ -0,0 +1,53 @@
...
+            "updated_at",
+            sa.DateTime(timezone=True),
+            server_default=sa.text("now()"),
+            nullable=True,
+        ),
+        sa.Column("patterns", sa.ARRAY(sa.String()), nullable=False),
+        sa.Column("single_row", sa.Boolean(), nullable=False),
+        sa.PrimaryKeyConstraint("id"),
+        sa.UniqueConstraint("single_row"),
+    )
```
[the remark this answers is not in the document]
erosselli: 🤔 I assume the down revision comment is irrelevant , but what about the constraint comment? is it just making things up?

src/fides/api/models/privacy_request_redaction_patterns.py
```diff
@@ -0,0 +1,94 @@
...
+        nullable=False,
+        unique=True,
+    )  # used to constrain table to one row
+
+    CheckConstraint(
+        "single_row", name="privacy_request_redaction_patterns_single_row_check"
+    )
+    UniqueConstraint(
+        "single_row", name="privacy_request_redaction_patterns_single_row_unique"
+    )
```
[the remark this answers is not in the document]
erosselli: I think this comment is right, it should go in `__table_args__`

src/fides/api/models/privacy_request_redaction_patterns.py
```diff
@@ -0,0 +1,94 @@
...
+
+    @declared_attr
+    def __tablename__(self) -> str:
+        return "privacy_request_redaction_patterns"
+
+    patterns = Column(
+        ARRAY(String),
+        nullable=False,
+        default=[],
+    )  # Array of regex pattern strings
```
erosselli: question: why do we need a single-row table with an array? wouldn't it make more sense to have one row per pattern?

src/fides/api/service/privacy_request/dsr_package/dsr_report_builder.py
```diff
@@ -1,16 +1,25 @@
+# pylint: disable=too-many-nested-blocks,too-many-branches
```
erosselli: 😢

src/fides/api/service/privacy_request/dsr_package/dsr_report_builder.py
```diff
@@ -80,6 +86,19 @@ def __init__(
...
         self.used_filenames: set[str] = set()
 
+        self.redaction_patterns: List[str] = []
+
+        # Set of entity names that should be redacted
+        self.entities_to_redact = set()
+
+        # Load redaction patterns and entities to redact from database
+        db = object_session(privacy_request)
+        if db is not None:
```
erosselli: when would `db` be `None` here?  what does that mean?

src/fides/api/service/privacy_request/dsr_package/dsr_report_builder.py
```diff
@@ -323,6 +379,151 @@ def _get_datasets_from_dsr_data(self) -> dict[str, Any]:
...
+        collection_name: Optional[str] = None,
+    ) -> str:
+        """
+        Helper function to create indexed names for datasets, collections, and fields.
+        Applies redaction if the name matches any of the configured regex patterns OR
+        if there's a fides_meta.redact: name configuration for this entity.
+
+        Args:
+            name_type: The type of name ("dataset", "collection", or "field")
+            name: The original name to be indexed
```
erosselli:
```suggestion
            name: The original name to be redacted
```

should this read "to be redacted" rather than "to be indexed" ?

src/fides/api/service/privacy_request/dsr_package/dsr_report_builder.py
```diff
@@ -323,6 +379,151 @@ def _get_datasets_from_dsr_data(self) -> dict[str, Any]:
...
+        # Fall back to global regex pattern matching
+        if self.redaction_patterns:
+            for pattern in self.redaction_patterns:
+                try:
+                    if re.search(pattern, name, re.IGNORECASE):
+                        return f"{name_type}_{index}"
+                except re.error:
+                    # Skip invalid regex patterns
+                    logger.warning(f"Invalid regex pattern: {pattern}")
+                    continue
```
erosselli: we could remove one level of nesting here 
```suggestion
        for pattern in self.redaction_patterns or []: 
            try:
                if re.search(pattern, name, re.IGNORECASE):
                    return f"{name_type}_{index}"
            except re.error:
                # Skip invalid regex patterns
                logger.warning(f"Invalid regex pattern: {pattern}")
                continue
```

don't commit this suggestion directly cause I messed up the indent I think

src/fides/api/service/privacy_request/dsr_package/dsr_report_builder.py
```diff
@@ -323,6 +379,151 @@ def _get_datasets_from_dsr_data(self) -> dict[str, Any]:
...
+                # Determine the field name to use
+                should_redact = False
+
+                # First check for fides_meta.redact: name configurations
+                if full_path in self.entities_to_redact:
+                    should_redact = True
+
+                # Fall back to global regex pattern matching
+                elif self.redaction_patterns:
+                    for pattern in self.redaction_patterns:
```
erosselli: same here . Also this code is very similar to the one above, maybe we can DRY it up a bit?

src/fides/api/service/privacy_request/dsr_package/dsr_report_builder.py
```diff
@@ -323,6 +379,151 @@ def _get_datasets_from_dsr_data(self) -> dict[str, Any]:
...
+        current_path: str = "",
+        field_index_counter: Optional[dict] = None,
+    ) -> Any:
+        """
+        Recursively redact field names within nested JSON structures based on redaction configurations.
+        """
+        if field_index_counter is None:
+            field_index_counter = {}
+
+        if isinstance(value, dict):
```
erosselli: this part was a bit hard to follow but also I'm not sure how to improve it so maybe it's the best it can be heh

tests/ops/service/privacy_request/test_dsr_report_builder.py
```diff
@@ -1225,3 +1343,1102 @@ def test_performance_large_dataset(self, privacy_request: PrivacyRequest):
         with zipfile.ZipFile(io.BytesIO(report.getvalue())) as zip_file:
             assert "data/dataset1/index.html" in zip_file.namelist()
             assert len([f for f in zip_file.namelist() if "collection" in f]) == 100
+
+
+@pytest.mark.integration_postgres
+class TestGetRedactionEntitiesMap:
```
erosselli: can we move this test class to another test file? this one's getting really big already

src/fides/api/service/privacy_request/dsr_package/dsr_report_builder.py
```diff
@@ -411,3 +627,180 @@ def _map_privacy_request(privacy_request: "PrivacyRequest") -> dict[str, Any]:
...
+                )
+
+    except Exception as exc:
+        # Log error but don't fail, just return empty set
+        logger.warning(f"Error extracting redaction configurations: {exc}")
+
+    return redaction_entities
+
+
+def get_redaction_entities_map_db(db: Session) -> set[str]:
```
erosselli: should this function be on this dsr_report_builder, or should we define it on the model file or somewhere else?

src/fides/api/models/privacy_request_redaction_patterns.py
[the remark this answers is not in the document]
erosselli: also do we need both CheckConstraint and UniqueConstraint? what's the difference here?

clients/admin-ui/src/features/settings/PrivacyRequestRedactionPatternsPage.tsx
galvana: Not needed! That was part of an AI-generated fix, we just needed an unique ID and index is good enough

clients/admin-ui/src/types/api/models/FidesMeta.ts
galvana: In theory, yes, but I'm ok keeping it like this until we run into another use case

src/fides/api/models/privacy_request_redaction_patterns.py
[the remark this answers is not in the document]
galvana: I guess we've never done this correctly 😅 I copied the other usages of this

src/fides/api/service/privacy_request/dsr_package/dsr_report_builder.py
galvana: That would be if the privacy request is not bound to an actual DB session. It's a bit of a hack because I didn't have time to clean up the DSR report builder more. There's a chain of functions that start getting called from the `request_runner_service` before we eventually get to the builder. Ideally, we call something like this from the `request_runner_service`
```
dsr_package_service = DSRPackageService(db)
dsr_package_service.generate_package(db, data, ...)
```
That way we can encapsulate all the steps in the service instead of chaining function calls.
```
1. request_runner_service.upload_and_save_access_results
2. request_runner_service.upload_access_results
3. storage_uploader_service.upload
4. storage_uploader_service._s3_uploader (or _gcs_uploader)
5. streaming_s3.upload_to_s3_streaming
6. SmartOpenStreamingStorage.upload_to_storage_streaming
7. SmartOpenStreamingStorage._handle_html_format_upload
8. DsrReportBuilder.__init__
9. DsrReportBuilder.generate
```

src/fides/api/service/privacy_request/dsr_package/dsr_report_builder.py
galvana: Yes 👍

clients/admin-ui/src/features/settings/PrivacyRequestRedactionPatternsPage.tsx
galvana: Yes, I centralized this

src/fides/api/alembic/migrations/versions/f457a75bc0e1_adding_redaction_patterns.py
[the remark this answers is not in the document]
galvana: Cursor was right about the constraints, I fixed them

src/fides/api/models/privacy_request_redaction_patterns.py
galvana: Maybe? I thought it'd be easier to do it this way 🤷‍♂️

src/fides/api/service/privacy_request/dsr_package/dsr_report_builder.py
galvana: I refactored things, we don't need this anymore!

src/fides/api/service/privacy_request/dsr_package/dsr_report_builder.py
galvana: I cleaned this up a lot and put it into a `DSRDataPreprocessor`. I decoupled the redaction logic from the report generation logic

src/fides/api/service/privacy_request/dsr_package/dsr_report_builder.py
galvana: I moved it to a util for now

src/fides/api/models/privacy_request_redaction_patterns.py
erosselli: from the outside is kind of confusing and makes it a bit harder to reason about; particularly since there's many patterns so one row per pattern would feel more natural?

src/fides/api/api/v1/endpoints/privacy_request_redaction_patterns_endpoints.py
```diff
@@ -0,0 +1,96 @@
...
+    PRIVACY_REQUEST_REDACTION_PATTERNS,
+    status_code=HTTP_200_OK,
+    response_model=PrivacyRequestRedactionPatternsResponse,
+    dependencies=[
+        Security(
+            verify_oauth_client, scopes=[PRIVACY_REQUEST_REDACTION_PATTERNS_UPDATE]
+        )
+    ],
+)
+def update_privacy_request_redaction_patterns(
```
erosselli, line 64: I think we should add a `POST` endpoint so that we can add patterns without replacing existing ones too

src/fides/api/api/v1/endpoints/privacy_request_redaction_patterns_endpoints.py
```diff
@@ -0,0 +1,96 @@
...
+    try:
+        updated = PrivacyRequestRedactionPattern.replace_patterns(
+            db=db, patterns=request.patterns
+        )
+
+        logger.info("Successfully updated privacy request redaction patterns")
+        return PrivacyRequestRedactionPatternsResponse(patterns=updated)
+
+    except Exception as e:
+        logger.error(
```
erosselli: maybe we should use `logger.exception` here so we get a stack trace (if enabled)

src/fides/api/models/privacy_request_redaction_pattern.py
```diff
@@ -0,0 +1,64 @@
...
+    Stores one regex pattern per row for masking dataset, collection, and field names
+    in DSR package reports.
+    """
+
+    @declared_attr
+    def __tablename__(self) -> str:
+        return "privacy_request_redaction_pattern"
+
+    # One pattern per row; unique to prevent duplicates
+    pattern = Column(String, nullable=False, unique=True)
```
erosselli, line 21: ❤️

src/fides/api/models/privacy_request_redaction_pattern.py
```diff
@@ -0,0 +1,64 @@
...
+    def replace_patterns(cls, db: Session, patterns: List[str]) -> List[str]:
+        """
+        Replace the set of stored patterns with the provided list using set reconciliation.
+
+        - Adds patterns that are not present
+        - Removes patterns that are no longer desired
+        - Returns the resulting canonical list in deterministic order
+        """
+        # Normalize: trim whitespace, remove empties and duplicates
+        desired: Set[str] = {p.strip() for p in patterns if p and p.strip()}
```
erosselli, line 46: nit: can we define a method `PrivacyRequestRedactionPattern.normalize(pattern)` that does that? might be a useful utility since we're  calling `strip` in more than one place

src/fides/api/service/privacy_request/dsr_package/dsr_data_preprocessor.py
```diff
@@ -0,0 +1,234 @@
...
+class DSRDataPreprocessor:
+    """
+    Processes DSR data to apply name redaction before report generation.
+    """
+
+    def __init__(self, db: Session):
+        self.db = db
+        self.redaction_patterns: List[str] = []
+        if patterns := PrivacyRequestRedactionPattern.get_patterns(db):
+            self.redaction_patterns = patterns
```
erosselli:
```suggestion
        self.redaction_patterns: PrivacyRequestRedactionPattern.get_patterns(db) or []
```

src/fides/api/service/privacy_request/dsr_package/dsr_data_preprocessor.py
```diff
@@ -0,0 +1,234 @@
...
+        dataset_mapping = self._create_dataset_mapping(dsr_data)
+
+        # Second pass: process data with redaction
+        processed_data = {}
+        collection_indices: Dict[str, int] = {}  # Track per dataset
+
+        for key, rows in dsr_data.items():
+            if key == "attachments":
+                processed_data[key] = rows
+                continue
```
erosselli, line 46: can we add a comment explaining this? also maybe the `"attachments"` string needs to be a constant, so that if at any point we want to rename it in the DSR package, it only needs to be updated once?

src/fides/api/service/privacy_request/dsr_package/dsr_data_preprocessor.py
```diff
@@ -0,0 +1,234 @@
...
+        """Process the DSR data to apply all redaction upfront."""
+        if not self.redaction_patterns and not self.entities_to_redact:
+            return dsr_data
+
+        # First pass: collect and map dataset names
+        dataset_mapping = self._create_dataset_mapping(dsr_data)
+
+        # Second pass: process data with redaction
+        processed_data = {}
+        collection_indices: Dict[str, int] = {}  # Track per dataset
```
erosselli: I'm confused about the `# track per dataset` comment

src/fides/api/service/privacy_request/dsr_package/dsr_data_preprocessor.py
```diff
@@ -0,0 +1,234 @@
...
+                        [
+                            k
+                            for k in collection_indices
+                            if k.startswith(f"{dataset_name}:")
+                        ]
+                    )
+                    + 1
+                )
+
+            collection_index = collection_indices[collection_key]
```
erosselli: instead of doing this `startswith` filtering, can't we have a `collection_indices` dict _per dataset_ , like the comment implies? i.e `collection_index = collection_indices[dataset_name][collection_name]`

src/fides/api/service/privacy_request/dsr_package/dsr_data_preprocessor.py
```diff
@@ -0,0 +1,234 @@
...
+        regular_datasets = [n for n in unique_datasets if n != "dataset"]
+        for index, name in enumerate(regular_datasets, 1):
+            hierarchical_key = self._build_hierarchical_key("dataset", name)
+            if self._should_redact(name, hierarchical_key):
+                mapping[name] = f"dataset_{index}"
+            else:
+                mapping[name] = name  # Keep original name
+
+        # Special "dataset" case comes last
+        if "dataset" in unique_datasets:
```
erosselli: why is this a special case ? also can we make this a constant too ?

src/fides/api/service/privacy_request/dsr_package/dsr_data_preprocessor.py
```diff
@@ -0,0 +1,234 @@
...
+        )
+
+        if self._should_redact(name, hierarchical_key):
+            return f"{name_type}_{index}"
+
+        return name
+
+    def _build_hierarchical_key(
+        self,
+        name_type: str,
```
erosselli: can we type this better, at least using a string literal union ?

src/fides/api/service/privacy_request/dsr_package/utils.py
```diff
@@ -0,0 +1,268 @@
...
+from sqlalchemy import text
+from sqlalchemy.orm import Session
+
+from fides.api.models.datasetconfig import DatasetConfig
+from fides.api.models.privacy_request.privacy_request import PrivacyRequest
+from fides.api.schemas.policy import ActionType
+
+
+# TODO: keeping this for a bit to help with development and testing
+def get_redaction_entities_map(db: Session) -> set[str]:
```
erosselli, line 14: just a reminder so we don't forget this

src/fides/api/service/privacy_request/dsr_package/dsr_data_preprocessor.py
galvana: I added a comment for now. I'll ticket the attachments cleanup https://ethyca.atlassian.net/browse/ENG-1306

src/fides/api/service/privacy_request/dsr_package/dsr_data_preprocessor.py
galvana: Removed, this was a leftover

src/fides/api/service/privacy_request/dsr_package/dsr_data_preprocessor.py
galvana: We don't need this, I removed it

erosselli (approved):
Thanks for making the changes! I tested locally and it's working as expected (with the caveat of the missing UI validation for duplicate patterns; this will be ticketed separately). Don't forget the changelog entry

PR #6489, opened 2025-08-25
privacy center dsr redirect

JadeCara:
Closes [ENG-1201](https://ethyca.atlassian.net/browse/ENG-1201?atlOrigin=eyJpIjoiNDNhMzIzNWY4NjhmNDIxZmIzMTJmYzg3YzhjOTI5YjUiLCJwIjoiaiJ9)

### Description Of Changes

Implemented and thoroughly tested a DSR (Data Subject Request) Package Download Endpoint that provides secure, authenticated access to privacy request results through presigned S3 URLs.

🔧 **Technical Implementation Details**
Storage Type Support: Currently S3-only (GCS explicitly unsupported)
File Handling: Automatically constructs filenames based on privacy request IDs
Error Handling: Comprehensive HTTP status codes (400, 404, 422) with descriptive messages
Rate Limiting: Configurable limits applied to public endpoint access
URL Generation: Proper presigned URL creation with AWS SDK integration

🌐 Privacy Center Integration
Email Integration: Modified completion email logic to use DSR package links instead of direct storage URLs when streaming is enabled
User Experience: Provides clickable links in completion emails that redirect to secure downloads
Configuration: Respects privacy center URL settings for proper link generation

🚀 Business Value
This endpoint enables:
Self-service data access: End users can download their data without admin intervention
Secure file sharing: Presigned URLs eliminate credential exposure risks
Audit trail: All download attempts are logged and rate-limited
Scalability: Handles multiple concurrent requests with proper rate limiting

### Code Changes

*🎯 Core Functionality Implemented
New API Endpoint: /api/v1/privacy-request/{privacy_request_id}/access-package
Public endpoint (no authentication required) for end users to download their data
Generates fresh presigned S3 URLs for secure, time-limited access
Redirects users to the actual file download
S3 Integration:
Supports both manual AWS credentials and automatic authentication methods
Generates presigned URLs with proper expiration and security parameters
Handles file naming conventions (e.g., {privacy_request_id}.zip)
Security Features:
- Rate limiting applied to prevent abuse
- Only completed privacy requests can access results
- Presigned URLs provide secure, temporary access without exposing storage credentials

🧪 Comprehensive Test Coverage
The test suite covers all critical scenarios:
- Request validation: Status checks, missing data handling, invalid IDs
- Storage integration: S3 presigned URL generation, GCS unsupported handling
- Error conditions: Incomplete requests, missing files, rate limiting
- Edge cases: Empty result sets, different storage configurations
- Integration testing: Full redirect flow validation

### Steps to Confirm

1. Ensure your default storage is set to S3 and has the appropriate credentials and auth.
2. add `"enable_streaming": true` and `"enable_access_package_redirect": true` to your storage config details `PUT /api/v1/storage/default`
3. Create a Manual Task integration (to ensure attachments) 
4. Create a DSR and complete the manual step.

Test with the email. 
5. Make sure you have mailgun etc configured. 
6. Create a DSR and complete the manual Step. 
7. Verify the email contains a privacy center url and when you click it the download occurs. 
Screenshot 2025-08-26 at 5 29 59 PM

Screenshot 2025-08-26 at 2 25 06 PM

To test the API you will need the token and the DSR id 

### Pre-Merge Checklist

* [x] Issue requirements met
* [x] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:

[ENG-1201]: https://ethyca.atlassian.net/browse/ENG-1201?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/fides/api/email_templates/templates/privacy_request_complete_access.html
```diff
@@ -23,4 +23,4 @@
             {% endif %}
         </main>
     </body>
-</html>
\ No newline at end of file
+</html>
```
JadeCara:
```suggestion
</html>

```

src/fides/api/email_templates/templates/privacy_request_complete_access.html
```diff
@@ -23,4 +23,5 @@
...
+
```
JadeCara:
```suggestion
</html>
```

clients/privacy-center/app/server-utils/requestLogger.ts
```diff
@@ -4,7 +4,7 @@ import { createLogger } from "./logger";
 
 export const createRequestLogger = (request: NextApiRequest) => {
   const isDevelopment = process.env.NODE_ENV === "development";
-  const id = request.headers["x-request-id"] ?? "null";
```
tvandort: That's surprising, in what situation do we not have a headers dictionary?

src/fides/api/api/v1/endpoints/dsr_package_link.py
```diff
@@ -0,0 +1,156 @@
...
+
+router = APIRouter(tags=["Privacy Center"], prefix=V1_URL_PREFIX)
+
+
+def get_privacy_request_or_error(
+    privacy_request_id: str, db: Session
+) -> PrivacyRequest:
+    """Load the privacy request or throw a 404"""
+    # Validate UUID format to prevent SSRF attacks
+    # Privacy request IDs are in format: pri_uuid where pri is prefix and uuid is UUID v4
```
tvandort: Hm, I'm realizing I may not understand SSRF, can you help me understand how validating the structure of the UUID prevents SSRF?

src/fides/api/api/v1/endpoints/dsr_package_link.py
```diff
@@ -0,0 +1,156 @@
...
+    if not privacy_request_id.startswith("pri_"):
+        raise HTTPException(
+            status_code=HTTP_400_BAD_REQUEST,
+            detail=f"Invalid privacy request ID format: '{privacy_request_id}'. Must start with 'pri_' followed by a valid UUID v4.",
+        )
```
tvandort: Duplicated below, might be worth putting it in a function that throws this exception.

src/fides/api/api/v1/endpoints/dsr_package_link.py
```diff
@@ -0,0 +1,156 @@
...
+
+    try:
+        uuid.UUID(uuid_part, version=4)
+    except ValueError:
+        raise HTTPException(
+            status_code=HTTP_400_BAD_REQUEST,
+            detail=f"Invalid privacy request ID format: '{privacy_request_id}'. Must start with 'pri_' followed by a valid UUID v4.",
+        )
+
+    privacy_request = PrivacyRequest.get(db, object_id=privacy_request_id)
```
tvandort, line 36: Do these have a TTL?

src/fides/api/api/v1/endpoints/dsr_package_link.py
```diff
@@ -0,0 +1,156 @@
...
+    status_code=HTTP_302_FOUND,
+)
+@fides_limiter.limit(CONFIG.security.public_request_rate_limit)
+def get_access_results_urls(
+    privacy_request_id: str,
+    db: Session = Depends(get_db),
+    *,
+    request: Request,  # required for rate limiting
+    response: Response,  # required for rate limiting
+) -> RedirectResponse:
```
tvandort: I might be paranoid but I think we must have a token passed here. We should probably talk to Dave to make sure we're doing this in the must secure way possible.

Trivially if the ID taken here is the same privacy request id exposed to admins then we would now be giving admins full access to packages regardless of any security settings.

tvandort (changes requested):

src/fides/api/api/v1/endpoints/dsr_package_link.py
JadeCara: Thats fair, I will work on something :)

src/fides/api/api/v1/endpoints/dsr_package_link.py
JadeCara: No, they are instant downloads so a few seconds or whatever is probably fine. They get regenerated whenever they hit the link and we dont store them anywhere.

src/fides/api/api/v1/endpoints/dsr_package_link.py
JadeCara: Validating the UUID structure helps prevent SSRF because it prevents user input from ever being treated as a free-form URL or path. It constrains the input into something meaningless on its own until you resolve it in a controlled way.

I mean - we are just running a query - but I always think of [Little Bobby Tables](https://www.explainxkcd.com/wiki/index.php/327:_Exploits_of_a_Mom).

clients/privacy-center/app/server-utils/requestLogger.ts
JadeCara: Honestly - I am a back end engineer 😂 - I have @lucanovera   on here as a reviewer as well. Cursor and I got through this together, but I am assuming there are things I will need to fix.

src/fides/api/api/v1/endpoints/dsr_package_link.py
JadeCara: Ok - I have added a new JWE token generator that is tied to the privacy request id. We do not use this generator/validator for anything else, we do not store the tokens and the token is not logged or visible in any other way. It is included in the link in the email. It is required for getting access to the downloads.

clients/privacy-center/app/server-utils/requestLogger.ts
JadeCara: I think I figured it out!

src/fides/api/api/v1/endpoints/dsr_package_link.py
tvandort: Ah sorry I mean, the actual file in S3. Before the URL used to have a TTL which mean that after that passed everyone lost access to the package. It seems to me that the URL that we provide should not last _indefinitely_. So what I'm asking after is either:

- The link we provide expires after some configurable time
- The S3 file itself gets cleaned up / expired after some time

clients/privacy-center/pages/api/privacy-request/[id]/access-package.ts
```diff
@@ -0,0 +1,306 @@
...
+ *         description: DSR package not found
+ *       500:
+ *         description: Internal server error
+ */
+export default async function handler(
+  req: NextApiRequest,
+  res: NextApiResponse,
+) {
+  // Ensure we have a request ID header for consistent logging
+  if (!req.headers["x-request-id"]) {
```
Comment (lucanovera) [inline clients/privacy-center/pages/api/privacy-request/[id]/access-package.ts]: since this header is being used a few times I think we could make that a const so it's only defined once

src/fides/api/api/v1/endpoints/dsr_package_link.py
JadeCara: > - The link we provide expires after some configurable time
This now times out with the token which uses the CONFIG.security.subject_request_download_link_ttl_seconds (currently defaults to 5 days which matches the message in the email)
> - The S3 file itself gets cleaned up / expired after some time
We currently don't have this feature at all for any of our uploaded things to S3. I think the most straight forward way to handle this though is by configuring lifecycle rules (based on age or tags) - which would be on their side.

src/fides/api/oauth/utils.py
```diff
@@ -160,6 +129,90 @@ def _get_request_task_jwe_or_error(
...
+        raise AuthorizationError(
+            detail="Download token does not grant access to this privacy request"
+        )
+
+    # Verify the token has the required scope
+    required_scope = "privacy-request-access-results:read"
+    if required_scope not in download_token.scopes:
+        raise AuthorizationError(detail="Download token lacks required permissions")
+
+    return download_token
```
tvandort: We should probably document this usage of the token somewhere. I know that there's some interest at Ethyca to move away from JWEs, if we did so this approach would have to be changed.

JadeCara: Thanks for the excellent call out - I will update the doc string and I need to make a followup documentation ticket/PR so I will mention it there.

JadeCara:
> 1. Is there a reason the generated access package only has a .json format? Does this not support the html access package?
That is configured through the storage config. I will do another pass at manual testing with various configs to make sure they all work as expected.

> 2. I see our default messaging says "For security purposes, this secret link will expire in 5 days." so I guess there should be a way to configure the expiration for the aws files?

I added an expiration to the token which can be configured using CONFIG.security.subject_request_download_link_ttl_seconds (currently defaults to 5 days which matches the message in the email)

.fides/fides.toml
```diff
@@ -49,7 +49,7 @@ task_retry_backoff = 1
 subject_identity_verification_required = false
 task_retry_count = 0
 task_retry_delay = 1
-use_dsr_3_0 = false
+use_dsr_3_0 = true
```
JadeCara:
```suggestion
use_dsr_3_0 = false
```

clients/privacy-center/pages/api/privacy-request/[id]/access-package.ts
Comment (JadeCara) [reply clients/privacy-center/pages/api/privacy-request/[id]/access-package.ts]: Updated!

src/fides/api/oauth/utils.py
```diff
@@ -160,6 +133,98 @@ def _get_request_task_jwe_or_error(
...
+
+    Raises:
+        AuthenticationError: If token is invalid or expired
+        AuthorizationError: If token doesn't grant access to the requested privacy request
+    """
+    if not token:
+        raise AuthenticationError(detail="Download token is required")
+
+    # Check if token looks like a JWE (should have 5 parts separated by dots)
+    if not token or token.count(".") != 4:
```
galvana: We already check for token in the previous statement
```suggestion
    if token.count(".") != 4:
```

PR #6518, opened 2025-08-28
Add DSR body to webhook invocations and timestamp

johnewart:
(Opening as a draft for the moment)

Closes ENG-1217

### Description Of Changes

No removal of data or breaking contract changes - only adds new data to web hook payloads; there may be more embedded structures/data that we want to add beyond the policy, custom fields, and the location data that is not in this PR.

### Code Changes

Makes the DSR body available through the webhook payload, which (when applicable) includes custom fields, location data, and policy information.

Additionally, adds a timestamp field to webhook payloads so that the receiving end can make decisions based on execution timestamp (i.e. potentially ignore duplicate invocations, or out-of-order delivery)

### Steps to Confirm

1. None - added tests to validate the presence of data in the payload

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:

### DIFF
--- a/tests/fixtures/application_fixtures.py
+++ b/tests/fixtures/application_fixtures.py
@@ -3,7 +3,7 @@ import uuid
 from copy import deepcopy
 from datetime import datetime, timedelta, timezone
 from io import BytesIO
-from typing import Dict, Generator, List, Optional
+from typing import Dict, Generator, List, Optional, Any
 from unittest import mock
 from uuid import uuid4
 
@@ -118,7 +118,6 @@ logging.getLogger("faker").setLevel(logging.ERROR)
 # disable verbose faker logging
 faker = Faker()
 integration_config = load_toml("tests/ops/integration_test_config.toml")
-
 
 # Unified list of connections to integration dbs specified from fides.api-integration.toml
 
@@ -252,7 +251,7 @@ integration_secrets = {
 @pytest.fixture(scope="session", autouse=True)
 def mock_upload_logic() -> Generator:
     with mock.patch(
-        "fides.api.service.storage.storage_uploader_service.upload_to_s3"
+            "fides.api.service.storage.storage_uploader_service.upload_to_s3"
     ) as _fixture:
         _fixture.return_value = "http://www.data-download-url"
         yield _fixture
@@ -495,7 +494,7 @@ def property_b(db: Session) -> Generator:
 
 
 @pytest.fixture(scope="function")
-def https_connection_config(db: Session) -> Generator:
+def https_connection_config(db: Session) -> Generator[ConnectionConfig, None, None]:
     name = str(uuid4())
     connection_config = ConnectionConfig.create(
         db=db,
@@ -516,7 +515,7 @@ def https_connection_config(db: Session) -> Generator:
 
 @pytest.fixture(scope="function")
 def policy_pre_execution_webhooks(
-    db: Session, https_connection_config, policy
+        db: Session, https_connection_config, policy
 ) -> Generator:
     pre_webhook = PolicyPreWebhook.create(
         db=db,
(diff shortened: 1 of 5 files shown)


tests/ops/models/privacy_request/test_privacy_request.py
```diff
@@ -455,6 +460,97 @@ def test_two_way_validation_issues(
             with pytest.raises(ValidationError):
                 privacy_request.trigger_policy_webhook(webhook)
 
+    def test_payload_contains_expected_fields(
```
galvana, line 463: The `test_payload_contains_expected_fields` test covers all the necessary scenarios but it should be broken down into separate tests for each separate behavior we want to test. This will make it easier to tell where a failure is rather than a single all or nothing test. It's preferred to use Pytest parametrization and to keep each test more focused. So something like this:
```
@pytest.mark.parametrize("webhook_type", ["pre_execution", "post_execution", "pre_approval"])
def test_webhook_payload_contains_timestamp(self, webhook_type):
    # Test just timestamp validation

@pytest.mark.parametrize("webhook_type", ["pre_execution", "post_execution", "pre_approval"])
def test_webhook_payload_includes_custom_fields_when_present(self):
    # Test just custom fields presence

@pytest.mark.parametrize("webhook_type", ["pre_execution", "post_execution", "pre_approval"])
def test_webhook_payload_excludes_custom_fields_when_absent(self):  
    # Test just custom fields absence
```

tests/fixtures/application_fixtures.py
```diff
@@ -495,7 +494,7 @@ def property_b(db: Session) -> Generator:
 
 
 @pytest.fixture(scope="function")
-def https_connection_config(db: Session) -> Generator:
+def https_connection_config(db: Session) -> Generator[ConnectionConfig, None, None]:
```
galvana, line 497: Thanks for the cleanup here 👍

galvana (changes requested):
The main change looks good, just requesting a bit of cleanup on the tests

tests/ops/models/privacy_request/test_privacy_request.py
johnewart: Rather than re-implement test logic, how about 

```
    @pytest.mark.parametrize(
        "webhook_type, field_to_check",
        list(
            itertools.product(
                ["pre_execution", "post_execution", "pre_approval"],
                [
                    "custom_fields",
                    "location",
                    "policy_key",
                    "no_custom_fields",
                    "timestamp",
                ],
            )
        ),
    )
```

and then map to the right webhook and matcher - that way you get marked tests but don't have the problem of test sprawl and duplicated logic for testing the same behavior (I assume you just want to know which of the combinations fail - such as if, for example, the pre-execution logic accidentally deletes the timestamp) 

Doing it this way you get: 


```
tests/ops/models/privacy_request/test_privacy_request.py::TestPrivacyRequestTriggerWebhooks::test_payload_contains_expected_fields[post_execution-no_custom_fields] 
tests/ops/models/privacy_request/test_privacy_request.py::TestPrivacyRequestTriggerWebhooks::test_payload_contains_expected_fields[post_execution-timestamp] 
tests/ops/models/privacy_request/test_privacy_request.py::TestPrivacyRequestTriggerWebhooks::test_payload_contains_expected_fields[pre_approval-custom_fields] 
tests/ops/models/privacy_request/test_privacy_request.py::TestPrivacyRequestTriggerWebhooks::test_payload_contains_expected_fields[pre_approval-location] 
tests/ops/models/privacy_request/test_privacy_request.py::TestPrivacyRequestTriggerWebhooks::test_payload_contains_expected_fields[pre_approval-policy_key] 
tests/ops/models/privacy_request/test_privacy_request.py::TestPrivacyRequestTriggerWebhooks::test_payload_contains_expected_fields[pre_approval-no_custom_fields] 
tests/ops/models/privacy_request/test_privacy_request.py::TestPrivacyRequestTriggerWebhooks::test_payload_contains_expected_fields[pre_approval-timestamp] 
```

galvana: Even better, very clean! 👌

galvana (approved):
Thanks for the test refactor! Looks good to me

PR #6520, opened 2025-08-28
Update add clients to params

JadeCara:
### Description Of Changes

I got a little heavy handed on some util changes, rolling those back. 
I added `client` as a kwarg to the `transport_params` dict.

### Steps to Confirm

1.  Make sure no AWS env vars are set 
2. Follow steps to confirm [here](https://github.com/ethyca/fides/pull/6489) 
### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:


tests/ops/service/storage/streaming/s3/test_s3_storage_client.py
```diff
@@ -42,62 +42,73 @@ def test_build_uri_custom_endpoint(self):
...
         secrets = {
-            StorageSecrets.AWS_ACCESS_KEY_ID: "test_key",
-            StorageSecrets.AWS_SECRET_ACCESS_KEY: "test_secret",
-            StorageSecrets.REGION_NAME: "us-west-2",
+            "client": {
+                StorageSecrets.AWS_ACCESS_KEY_ID: "test_key",
+                StorageSecrets.AWS_SECRET_ACCESS_KEY: "test_secret",
+                StorageSecrets.REGION_NAME: "us-west-2",
+            }
         }
```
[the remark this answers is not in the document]
JadeCara: This is not making storage secrets its creating transport params for smart_open

Kelsey-Ethyca (approved):

JadeCara:
@greptile review the ENTIRE PR

JadeCara:
@greptile review the ENTIRE PR

JadeCara:
@greptile please review

src/fides/api/service/storage/streaming/smart_open_client.py
```diff
@@ -225,16 +224,15 @@ def stream_upload(
...
             Writable file-like object
         """
         uri = self._build_uri(bucket, key)
         transport_params = self._get_transport_params()
-
-        if content_type:
-            transport_params["content_type"] = content_type
+        logger.debug(
+            f"Streaming upload to {uri} with transport params: {transport_params}"
+        )
```
JadeCara:
```suggestion
```

JadeCara:
@greptile please review

JadeCara:
@greptile please reveiw all changes not just most recent

clients/privacy-center/pages/api/privacy-request/[id]/access-package.ts
```diff
@@ -135,24 +138,56 @@ export default async function handler(
...
+        "code" in fetchError.cause &&
+        fetchError.cause.code === "ECONNREFUSED";
+
+      if (isConnectionRefused) {
+        log.error(
+          `Fides API connection refused - service may not be running or is not reachable: ${baseUrl}`,
+          {
+            url,
+            error: "Connection refused (ECONNREFUSED)",
+            suggestion: "Check if Fides API service is running on port 8080",
```
Comment (tvandort) [inline clients/privacy-center/pages/api/privacy-request/[id]/access-package.ts]: Fides won't always be running on 8080.

clients/privacy-center/pages/api/privacy-request/[id]/access-package.ts
```diff
@@ -135,24 +138,56 @@ export default async function handler(
...
+        headers,
+      });
+    } catch (fetchError) {
+      // Check if this is a connection refused error
+      const isConnectionRefused =
+        fetchError instanceof Error &&
+        fetchError.cause &&
+        typeof fetchError.cause === "object" &&
+        "code" in fetchError.cause &&
+        fetchError.cause.code === "ECONNREFUSED";
```
Comment (tvandort) [inline clients/privacy-center/pages/api/privacy-request/[id]/access-package.ts:162]:
```suggestion
      const isConnectionRefused = fetchError?.cause?.code === "ECONNREFUSED;
```

Might be able to simplify this with something like the above if the type checker lets you. But I do question whether we need this sort of error handling at all?

clients/privacy-center/pages/api/privacy-request/[id]/access-package.ts
```diff
@@ -135,24 +138,56 @@ export default async function handler(
...
-    const response = await fetch(url, {
-      method: "GET",
-      headers,
-    });
+    let response: Response;
+    try {
+      response = await fetch(url, {
+        method: "GET",
+        headers,
+      });
```
Comment (tvandort) [inline clients/privacy-center/pages/api/privacy-request/[id]/access-package.ts:154]: Line 149 and lines 192 to 334 could be moved under the try. This will simply the control flow and prevent an unnecessary rethrow.

src/fides/api/oauth/utils.py
```diff
@@ -67,14 +67,17 @@ def is_token_expired(
...
-    return datetime.now() > expiration_time
+    return (datetime.now() - issued_at).total_seconds() / 60.0 > token_duration_minutes
 
 
 def is_callback_token_expired(issued_at: datetime) -> bool:
     """Check if a callback token has expired (24 hours)."""
-    expiration_time = issued_at + timedelta(hours=24)
-    return datetime.now() > expiration_time
+    if not issued_at:
+        return True
```
tvandort, line 76: I might be reading this wrong, if `issued_at` is type of `datetime` when could it be type of `None`? Should the function parameter be optional or maybe unioned with `None`?

src/fides/api/service/storage/storage_uploader_service.py
```diff
@@ -130,18 +124,9 @@ def _s3_uploader(
...
-                list(config.secrets.keys())
-                if isinstance(config.secrets, dict)
-                else "Not a dict"
-            ),
-        )
-        logger.debug("Config secrets type: {}", type(config.secrets))
-    else:
+    if not config.secrets:
         logger.warning("Config secrets is None or empty!")
+        raise StorageUploadError("Config secrets is None or empty!")
```
tvandort: Does this account for secrets being configured in environment variables? I can't remember if the authentication mode of `automatic` still requires a secrets configuration or not.

src/fides/api/service/storage/streaming/s3/s3_storage_client.py
```diff
@@ -7,11 +7,14 @@
...
 from loguru import logger
 
-from fides.api.schemas.storage.storage import AWSAuthMethod, StorageSecrets
+from fides.api.schemas.storage.storage import AWSAuthMethod
 from fides.api.service.storage.s3 import create_presigned_url_for_s3
 from fides.api.service.storage.streaming.base_storage_client import BaseStorageClient
 from fides.api.util.aws_util import get_s3_client
 
+# Type annotation: get_s3_client returns a boto3 S3 client object, not a Session
+# This is what smart-open expects for the "client" transport parameter
```
tvandort: Should we move this comment to where the transport parameters are generated?

src/fides/api/service/storage/streaming/s3/s3_storage_client.py
```diff
@@ -85,28 +145,28 @@ def generate_presigned_url(
...
             # If AWS credentials are present, use SECRET_KEYS, otherwise use AUTOMATIC
             if (
-                StorageSecrets.AWS_ACCESS_KEY_ID in self.storage_secrets
-                and StorageSecrets.AWS_SECRET_ACCESS_KEY in self.storage_secrets
-                and self.storage_secrets[StorageSecrets.AWS_ACCESS_KEY_ID]
-                and self.storage_secrets[StorageSecrets.AWS_SECRET_ACCESS_KEY]
+                "aws_access_key_id" in self.storage_secrets
+                and "aws_secret_access_key" in self.storage_secrets
+                and self.storage_secrets["aws_access_key_id"]
+                and self.storage_secrets["aws_secret_access_key"]
```
tvandort: I feel like using the enum for this was a good call is there a reason we had to use strings for all the attribute accesses instead?

clients/privacy-center/pages/api/privacy-request/[id]/access-package.ts
Comment (tvandort) [reply clients/privacy-center/pages/api/privacy-request/[id]/access-package.ts:162]: I realized what I wrote was unclear, what I mean is do we need to special case `ECONNREFUSED`, I'd assume we could more generically handle most errors that come out of the fetch call.

src/fides/api/service/storage/streaming/s3/s3_storage_client.py
JadeCara: Unfortunately the `aws_util` that this ends up calling is looking specifically for strings - This has been an on going irritation for me for two weeks.
https://github.com/ethyca/fides/blob/89655c433728be7556925862888cdad191f46934/src/fides/api/util/aws_util.py#L39

src/fides/api/service/storage/storage_uploader_service.py
JadeCara: I actually just removed that! Good eye :)

clients/privacy-center/pages/api/privacy-request/[id]/access-package.ts
Comment (JadeCara) [reply clients/privacy-center/pages/api/privacy-request/[id]/access-package.ts]: There it is! I was looking for that. Updated :)

clients/privacy-center/pages/api/privacy-request/[id]/access-package.ts
Comment (JadeCara) [reply clients/privacy-center/pages/api/privacy-request/[id]/access-package.ts:162]: This is what I was using the debug the connectivity issue we were having. Because this is a redirect it was super unclear what was going on. I am happy to simplify though!

Comment (JadeCara) [reply clients/privacy-center/pages/api/privacy-request/[id]/access-package.ts:162]: The simplification doesnt work because cause is not defined :(

src/fides/api/oauth/utils.py
JadeCara: I am not totally sure. This is how it was before my last change, and I was just undoing unnecessary changes.

clients/privacy-center/pages/api/privacy-request/[id]/access-package.ts
Comment (JadeCara) [reply clients/privacy-center/pages/api/privacy-request/[id]/access-package.ts:154]: response is defined on 149 so putting it in side that try makes it error.  I moved the other lines inside, spinning up my local env to test. again

src/fides/api/oauth/utils.py
JadeCara: The change I wanted to revert is in here https://github.com/ethyca/fides/pull/6489/files#diff-0a678a55644204bf9e52c4fae05851e093a384ba9a908a063472260fbdbde509

clients/privacy-center/pages/api/privacy-request/[id]/access-package.ts
Comment (JadeCara) [reply clients/privacy-center/pages/api/privacy-request/[id]/access-package.ts:154]: response is defined on 149 so putting it in side that try makes it error.  I am just waiting on my environment to spin back up so I can test to make sure I didnt break anything.

Comment (tvandort) [reply clients/privacy-center/pages/api/privacy-request/[id]/access-package.ts:154]: I meant to do: 

```
let response = await fetch(url, {
    method: "GET",
    headers,
});
```

clients/privacy-center/pages/api/privacy-request/[id]/access-package.ts
Comment (tvandort) [reply clients/privacy-center/pages/api/privacy-request/[id]/access-package.ts:162]: Ah OK I see what the type casting issue is. 

> Because this is a redirect it was super unclear what was going on.

What was the issue in your env that was causing this? Just trying to figure out if there is some other abstraction we can use here. Specific error handling for one type of node error just feels a bit off to me but maybe my initial impression is wrong.

src/fides/api/oauth/utils.py
erosselli: looking at the original: `issued_at` was `datetime | None`. We should replace it to be `Optional[datetime]` since the `|` can be messy in Python 3.9. 

I think that's what makes the most sense

src/fides/api/service/privacy_request/request_service.py
```diff
@@ -660,8 +661,24 @@ def requeue_interrupted_tasks(self: DatabaseTask) -> None:
...
+                                privacy_request.status
+                                == PrivacyRequestStatus.requires_input
+                            ):
+                                # For requires_input status, don't automatically error the request
+                                # as it's intentionally waiting for user input
+                                logger.warning(
+                                    f"No task ID found for request task {request_task_id} "
+                                    f"(privacy request {privacy_request.id}) in requires_input status - "
+                                    f"keeping request in current status as it may be waiting for manual input"
+                                )
```
erosselli, line 677: this is a warning because we would normally expect a subtask id ?

src/fides/api/service/storage/streaming/dsr_storage.py
```diff
@@ -29,15 +28,12 @@ def stream_dsr_buffer_to_storage(
...
-        content_type: MIME type for the uploaded file (defaults to application/zip)
     """
     # Get the content from the buffer
     content = dsr_buffer.getvalue()
     try:
         # Use smart-open's streaming upload for efficient memory usage
-        with storage_client.stream_upload(
-            bucket_name, file_key, content_type=content_type
-        ) as upload_stream:
+        with storage_client.stream_upload(bucket_name, file_key) as upload_stream:
```
erosselli, line 36: is there a default `content_type` and that's why we're no longer passing it to the library?

src/fides/api/service/storage/streaming/s3/s3_storage_client.py
```diff
@@ -20,17 +20,18 @@ class S3StorageClient(BaseStorageClient):
     generation for the smart-open storage client.
     """
 
-    def __init__(self, storage_secrets: dict[StorageSecrets, Any]):
+    def __init__(self, storage_secrets: dict[str, Any]):
```
erosselli: is str correct here? 🤔

src/fides/api/service/storage/streaming/s3/s3_storage_client.py
```diff
@@ -45,26 +46,84 @@ def build_uri(self, bucket: str, key: str) -> str:
...
+                "aws_access_key_id" in self.storage_secrets
+                and "aws_secret_access_key" in self.storage_secrets
+                and self.storage_secrets["aws_access_key_id"]
+                and self.storage_secrets["aws_secret_access_key"]
+            ):
+                auth_method = AWSAuthMethod.SECRET_KEYS.value
+            else:
+                auth_method = AWSAuthMethod.AUTOMATIC.value
+
+                # For automatic authentication, check if region is available
```
erosselli: don't we store the auth method as part of the storage secrets? in some ways to set storage up, I remember there's an `auth_method` variable that the user can configure... does that not come into play at all here? 

I feel like there might be a way to make a more informed decision rather than just trying to guess the right credential approach?

src/fides/api/service/storage/streaming/s3/s3_storage_client.py
```diff
@@ -45,26 +46,84 @@ def build_uri(self, bucket: str, key: str) -> str:
...
+                    logger.warning(
+                        "No region specified in storage secrets for automatic authentication"
+                        "This may cause credential issues - consider setting a default region"
+                    )
+
+            # Extract assume_role_arn if present
+            assume_role_arn = None
+            if "assume_role_arn" in self.storage_secrets:
+                assume_role_arn = self.storage_secrets["assume_role_arn"]
+                logger.debug(f"Using assume role ARN: {assume_role_arn}")
```
erosselli, line 77: to try to consolidate logging, can we remove this line and add the assume_role_arn if applicable to the debug log on line 90?

src/fides/api/service/storage/streaming/s3/streaming_s3.py
```diff
@@ -41,9 +41,14 @@ def upload_to_s3_streaming(
     This function now uses smart-open for efficient cloud storage operations while maintaining
     our DSR-specific business logic for package splitting and attachment processing.
     """
+    logger.debug(f"upload_to_s3_streaming called with auth_method={auth_method}")
+    logger.debug(f"Storage secrets type: {type(storage_secrets).__name__}")
+
     formatted_secrets = format_secrets(storage_secrets)
+    logger.debug(f"Formatted secrets keys: {list(formatted_secrets.keys())}")
```
erosselli: same here, it seems like this will generate a lo of debug logs. Also I don't think we should log secret keys, even in debug level?  I know we can get less strict with debug logging than info-level and above but a lot of clients do have debug logging enabled in prod so trying to think a bit about log cost here :)

src/fides/api/service/storage/streaming/s3/s3_storage_client.py
tvandort: In the sense that we keep having to type ignore it?

src/fides/api/service/privacy_request/request_service.py
JadeCara: Yeah, I think so. This just popped up recently but was causing the `requires input` status to error when this function was called.

src/fides/api/service/storage/streaming/dsr_storage.py
JadeCara: The content type was something I put in for the previous iteration - smart open does not accept it as an input - it doesnt error, just logs a warning.

src/fides/api/service/storage/streaming/s3/s3_storage_client.py
JadeCara: Dude ... I know ... Its not great. This is the aws_util thing and where do we keep the enums vs changing them into the strings... Since both our `aws_util` and the smart_open api wants them in strings, I just decided to make them strings

src/fides/api/service/storage/streaming/s3/s3_storage_client.py
JadeCara: Yeah I thinks this could be cleaner. Let me make an update and make sure it all still works.

clients/privacy-center/pages/api/privacy-request/[id]/access-package.ts
Comment (tvandort) [reply clients/privacy-center/pages/api/privacy-request/[id]/access-package.ts:154]: I have another alternative that we could use here. Instead of rethrowing on line 190, cause a 500 instead. My main issue is that we're using exceptions for control flow which is hard to follow & refactor.

src/fides/api/service/storage/streaming/s3/s3_storage_client.py
tvandort: What I'm asking is, is there any reason this doesn't work:

```python
StorageSecrets.AWS_ACCESS_KEY_ID.value in self.storage_secrets
and StorageSecrets.AWS_SECRET_ACCESS_KEY.value in self.storage_secrets
and self.storage_secrets[StorageSecrets.AWS_ACCESS_KEY_ID.value]
and self.storage_secrets[StorageSecrets.AWS_SECRET_ACCESS_KEY.value]
```

erosselli: I share the question but haven't dealt with a lot of this AWS code

src/fides/api/oauth/utils.py
galvana: We should update the doc string since the expiration is now configurable

src/fides/api/service/storage/streaming/s3/s3_storage_client.py
tvandort: Also to be clear my main contention is retyping a string in N+1 number of places. If neither the enum nor the enum value works then I'd like if we could reuse variables in the places: e.g. 

```
AWS_ACCESS_KEY_ID="access_key_id"
AWS_SECRET_ACCESS_KEY="secret_access_key"

AWS_ACCESS_KEY_ID in self.storage_secrets
and AWS_SECRET_ACCESS_KEY in self.storage_secrets
and self.storage_secrets[AWS_ACCESS_KEY_ID]
and self.storage_secrets[AWS_SECRET_ACCESS_KEY]
```

src/fides/api/service/storage/streaming/s3/streaming_s3.py
tvandort: I think this is just logging the keys of the dictionary and not the secrets themselves, so it's probably fine?

clients/privacy-center/pages/api/privacy-request/[id]/access-package.ts
Comment (galvana) [reply clients/privacy-center/pages/api/privacy-request/[id]/access-package.ts:162]: I like @tvandort's simplification here. The intention here was to detect either network issues or the server being unavailable and log a friendly message.

galvana (approved):

tvandort (approved):

src/fides/api/service/storage/streaming/s3/s3_storage_client.py
JadeCara: Oof big update but its in :) works and is better solution.

PR #6526, opened 2025-08-29
ENG-681 Re implement pw reset session invalidation

thabofletcher:
Closes [ENG-681]

### Description Of Changes

This is to re-implement ENG-681, which was reverted due to its impact on FidesPlus/unit tests. This has been refactored to not require the changes the the OAuth client test fixture.

### Code Changes

Invalidates exisitng sessions for a user when their password has been reset

### Steps to Confirm

1.  Login as a user
2. Change the pw for that user
3. Watch the session get invalidated

### Pre-Merge Checklist

* [x] Issue requirements met
* [x] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [x] No followup issues
* Database migrations:
  * [x] No migrations
* Documentation:

[ENG-681]: https://ethyca.atlassian.net/browse/ENG-681

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,11 @@ Changes can also be flagged with a GitHub label for tracking purposes. The URL o
 
 ## [Unreleased](https://github.com/ethyca/fides/compare/2.69.0...main)
 
+### Security
+- Changed session invalidation logic to end all sessions for a user when their password has been changed [CVE-2025-57766](https://github.com/ethyca/fides/security/advisories/GHSA-rpw8-82v9-3q87)
+- Added stricter rate limiting to authentication endpoints to mitigate against brute force attacks. [CVE-2025-57815](https://github.com/ethyca/fides/security/advisories/GHSA-7q62-r88r-j5gw)
+- Adds Redis-driven rate limiting across all endpoints [CVE-2025-57816](https://github.com/ethyca/fides/security/advisories/GHSA-fq34-xw6c-fphf)
+
 ### Deprecated
 - DSR 2.0 is deprecated. New requests will be created using DSR 3.0 only. Existing DSR 2.0 requests will continue to process until completion. [#6458](https://github.com/ethyca/fides/pull/6458)
 
@@ -61,11 +66,6 @@ Changes can also be flagged with a GitHub label for tracking purposes. The URL o
 - Fixed bug with non-applicable notices being saved as opted in in Fides.js [#6490](https://github.com/ethyca/fides/pull/6490)
 - Handle missing GVL in TCF experience by displaying an error message instead of infinite spinners. [#6472](https://github.com/ethyca/fides/pull/6472)
 - Prevent edits for assets that have been ignored in the Action Center [#6485](https://github.com/ethyca/fides/pull/6485)
-
-### Security
-- Changed session invalidation logic to end all sessions for a user when their password has been changed [CVE-2025-57766](https://github.com/ethyca/fides/security/advisories/GHSA-rpw8-82v9-3q87)
-- Added stricter rate limiting to authentication endpoints to mitigate against brute force attacks. [CVE-2025-57815](https://github.com/ethyca/fides/security/advisories/GHSA-7q62-r88r-j5gw)
-- Adds Redis-driven rate limiting across all endpoints [CVE-2025-57816](https://github.com/ethyca/fides/security/advisories/GHSA-fq34-xw6c-fphf)
 
 ## [2.68.0](https://github.com/ethyca/fides/compare/2.67.2...2.68.0)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,13 +56,11 @@ Changes can also be flagged with a GitHub label for tracking purposes. The URL o
 
 ### Fixed
 - Fixed bug with non-applicable notices being saved as opted in in Fides.js [#6490](https://github.com/ethyca/fides/pull/6490)
-
-
-### Fixed
 - Handle missing GVL in TCF experience by displaying an error message instead of infinite spinners. [#6472](https://github.com/ethyca/fides/pull/6472)
 - Prevent edits for assets that have been ignored in the Action Center [#6485](https://github.com/ethyca/fides/pull/6485)
 
 ### Security
+- Changed session invalidation logic to end all sessions for a user when their password has been changed [CVE-2025-57766](https://github.com/ethyca/fides/security/advisories/GHSA-rpw8-82v9-3q87)
 - Added stricter rate limiting to authentication endpoints to mitigate against brute force attacks. [CVE-2025-57815](https://github.com/ethyca/fides/security/advisories/GHSA-7q62-r88r-j5gw)
 
 ## [2.68.0](https://github.com/ethyca/fides/compare/2.67.2...2.68.0)
--- a/clients/admin-ui/src/features/user-management/NewPasswordModal.tsx
+++ b/clients/admin-ui/src/features/user-management/NewPasswordModal.tsx
@@ -18,8 +18,9 @@ import { useDispatch } from "react-redux";
 import * as Yup from "yup";
 
 import { useAppSelector } from "~/app/hooks";
-import { LOGIN_ROUTE, STORAGE_ROOT_KEY } from "~/constants";
+import { LOGIN_ROUTE } from "~/constants";
 import { logout, selectUser } from "~/features/auth/auth.slice";
+import { clearAuthAndLogout } from "./logout-helpers";
 import { CustomTextInput } from "~/features/common/form/inputs";
 import { passwordValidation } from "~/features/common/form/validation";
 import { getErrorMessage } from "~/features/common/helpers";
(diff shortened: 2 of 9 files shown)


src/fides/api/models/client.py
```diff
@@ -48,9 +50,11 @@ def __tablename__(self) -> str:
         ARRAY(String), nullable=False, server_default="{}", default=dict
     )
     fides_key = Column(String, index=True, unique=True, nullable=True)
-    user_id = Column(
-        String, ForeignKey(FidesUser.id_field_path), nullable=True, unique=True
-    )
+    user_id = Column(String, ForeignKey("fidesuser.id"), nullable=True, unique=True)
```
thabofletcher: I'm looking at ways of restoring this, fix incoming here

CHANGELOG.md
```diff
@@ -56,13 +56,11 @@ Changes can also be flagged with a GitHub label for tracking purposes. The URL o
...
 ### Fixed
 - Fixed bug with non-applicable notices being saved as opted in in Fides.js [#6490](https://github.com/ethyca/fides/pull/6490)
-
-
-### Fixed
 - Handle missing GVL in TCF experience by displaying an error message instead of infinite spinners. [#6472](https://github.com/ethyca/fides/pull/6472)
 - Prevent edits for assets that have been ignored in the Action Center [#6485](https://github.com/ethyca/fides/pull/6485)
 
 ### Security
+- Changed session invalidation logic to end all sessions for a user when their password has been changed [CVE-2025-57766](https://github.com/ethyca/fides/security/advisories/GHSA-rpw8-82v9-3q87)
```
erosselli: is this going into 2.69.0? if not we should add it to the unreleased section ( or create the 2.69.1 section while we're here )

src/fides/api/api/v1/endpoints/user_endpoints.py
```diff
@@ -208,6 +208,17 @@ def update_user_password(
...
 
+    # Delete the user's associated OAuth client to invalidate all existing sessions
+    if current_user.client:
+        try:
+            current_user.client.delete(db)
+        except Exception as exc:
+            logger.exception(
+                "Unable to delete user client during password reset for user {}: {}",
+                current_user.id,
+                exc,
```
erosselli, line 219: shoudn't we be raising here / returning an erro HTTP after logging the exception? I'd expect the endpoint to return an error if this failed, not an OK response

src/fides/api/api/v1/endpoints/user_endpoints.py
```diff
@@ -236,6 +247,18 @@ def force_update_password(
...
+    # Delete the user's associated OAuth client to invalidate all existing sessions
+    if user.client:
+        try:
+            user.client.delete(db)
+        except Exception as exc:
+            logger.exception(
+                "Unable to delete user client during admin-forced password reset for user {}: {}",
+                user.id,
+                exc,
+            )
```
erosselli, line 260: same question here

src/fides/api/oauth/utils.py
```diff
@@ -80,6 +80,33 @@ def is_callback_token_expired(issued_at: Optional[datetime]) -> bool:
...
+    Any errors accessing related objects are logged and treated as non-invalidating.
+    """
+    try:
+        if client.user is None or client.user.password_reset_at is None:
+            return False
+
+        reset_at = client.user.password_reset_at
+        # DB layer may return a tz-aware timestamp while the JWE issued_at is naive (server time).
+        # Normalize both to naive for a safe, deterministic comparison without raising TypeError.
+        issued_naive = issued_at.replace(tzinfo=None)
```
erosselli: I'm confused about this.... we always change `issued_at` to be with no timestamp, even if the `reset_at` is not naive ? in that case wouldn't we want to keep both `tzinfo` ? 

Also I'm a bit concerned that naively replacing the tz info could lead to not invalidating certain tokens that should be invalidated?  (Also more generally I'm surprised that the JWE one is naive?)

src/fides/api/oauth/utils.py
```diff
@@ -225,7 +252,7 @@ async def get_current_user(
             created_at=datetime.utcnow(),
         )
 
-    return client.user  # type: ignore[attr-defined]
+    return cast(FidesUser, client.user)
```
erosselli, line 250: what does cast do if `client.user` is `None`?

clients/admin-ui/src/features/user-management/NewPasswordModal.tsx
```diff
@@ -52,6 +60,17 @@ const useNewPasswordModal = (id: string) => {
...
+      // Only logout if admin reset their own password
+      if (currentUser?.id === id) {
+        try {
+          localStorage.removeItem(STORAGE_ROOT_KEY);
+        } catch (e) {
+          // no-op
+        }
+        dispatch(logout());
+        router.push(LOGIN_ROUTE);
+      }
```
tvandort: We see almost the same steps in `UpdatePasswordModal` should we extract a function of some sort?

src/fides/api/oauth/utils.py
thabofletcher: I think its essentially the same, still None

```
python - <<'PY'
from typing import cast, Optional

class FidesUser:
    def __init__(self, username: str) -> None:
        self.username = username

client_user: Optional[FidesUser] = None  # Simulate client.user is None

u = cast(FidesUser, client_user)  # cast is a no-op at runtime

print("u is None:", u is None)           # True
print("type(u):", type(u))               # <class 'NoneType'>

PY
```

The real issue here is that the return type on this function is not Optional - when we are apparently running test cases that run into a case where this can be None, which isn't a value its supposed to have since its not Optional

src/fides/api/api/v1/endpoints/user_endpoints.py
thabofletcher: Not necessarily - the password reset didn't fail, just the reset of the user session as a result of the change.

src/fides/api/oauth/utils.py
erosselli: can't we just update the return type instead?

src/fides/api/oauth/utils.py
thabofletcher: The intention is to change both to remove `tzinfo` if it is present - I can make these statements more consistent with one another

I only ran into this issue in test cases, not in real world testing, IIRC reset_at (coming from the test case) was naive and the issued_at was coming from the DB and had tzinfo. So this snippet addressed that test.

I would assume in any real world scenario the tzinfo we are removing is going to be consistent with one another and match but perhaps that is not a guarantee

src/fides/api/oauth/utils.py
```diff
@@ -77,6 +77,26 @@ def is_callback_token_expired(issued_at: datetime) -> bool:
...
+    Return True if the token should be considered invalid due to security events
+    (e.g., user password reset) that occurred after the token was issued.
+
+    Any errors accessing related objects are logged and treated as non-invalidating.
+    """
+    try:
+        if client.user is not None and client.user.password_reset_at is not None:
+            password_reset_at = client.user.password_reset_at
+            if password_reset_at and issued_at < password_reset_at:
+                return True
```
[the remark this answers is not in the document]
thabofletcher: Long since been fixed

src/fides/api/oauth/utils.py
erosselli: I'd say we should make our test cases reflect the real world, not the other way around -- it's fine to code for the edge case but I'd want our tests to reflect real world scenarios

src/fides/api/oauth/utils.py
thabofletcher: We can, but the change is non-trivial and outside the scope of this fix. If this wasn't under pressure to be released tomorrow, we might be able to dig into tht a bit more, but for now can we live with this?

erosselli: ok fair enough

erosselli (approved):
I'm approving, thanks for making the changes. I think we should ticket a follow-up to look at the timezone comparison thing a bit more in depth as it feels like it might be kind of fragile? though not as bad as the original security vuln

tests/ops/api/v1/endpoints/test_user_endpoints.py
```diff
@@ -2063,11 +2192,18 @@ def test_get_system_managed_by_user_not_authenticated(
...
 
     def test_get_system_managed_by_user_wrong_scope(
-        self, api_client: TestClient, generate_auth_header, url
+        self, api_client: TestClient, oauth_client, url
     ):
         # Note that no user is attached to this client so we can't check
         # to see if the user is accessing its own systems
-        auth_header = generate_auth_header(scopes=[PRIVACY_REQUEST_READ])
+        from fides.config import CONFIG
+        from tests.conftest import _generate_auth_header
```
erosselli, line 2200: this is cursor being annoying with the local imports, can we move them top-level? I have an item in my to-do list to add a rule to fides so it doesn't do this

tests/ops/api/v1/endpoints/test_user_endpoints.py
```diff
@@ -1222,6 +1222,72 @@ def test_update_user_password(
...
+        resp_reset = api_client.post(
+            f"{url_no_id}/{application_user.id}/reset-password",
+            headers=auth_header,
+            json={
+                "old_password": str_to_b64_str(OLD_PASSWORD),
+                "new_password": str_to_b64_str(NEW_PASSWORD),
+            },
+        )
+        assert resp_reset.status_code == HTTP_200_OK
+        # Exception is handled and request succeeds; log assertion not required for coverage
```
erosselli, line 1289: log assertion is not required but we could add it at very little cost right ? since it's the only "evidence" that we're hitting the exception code path, and codecov's reporting we aren't ... so it might be a good sanity check to leave in place , you can use the `loguru_caplog` fixture

PR #6536, opened 2025-09-02
updating attachments in dsr builder etc

JadeCara:
Issue found in testing: 
https://ethyca.slack.com/archives/C08SW6SBDKJ/p1756818738554899?thread_ts=1756409420.827519&cid=C08SW6SBDKJ

### Description Of Changes

This PR is tracking down and cleaning up some issues in the streaming uploads for DSR work.

:red_circle: Links in the HTML based DSR ZIP are still to S3 files
:red_circle: If you upload two or more files with the same name you only get one in the access package.
🔶  HTML based DSR still has text "Note: all download links will expire in 7 days." (This text should stay and vary based on the TTL

The issues were because of how the html package was getting the attachments from the datasets. I made updates to the Manual tasks and the dsr builder as well as the streaming storage to make sure the communication across the systems is consistent and that we are handling duplicative names correctly. 

### Code Changes

* Updated the Manual Task to save the attachments data locally for dataset retrieval
* Update the dsr builder to handle local attachments and download attachments 
* Updated the streaming storage to handle duplicate filenames. 
* Added tests

### Steps to Confirm

1. Running on fidesplus: 
2. To confirm you will need to make sure you have set up S3 as the default storage config. 
3. You will need to additionally set the following in your storage config to true
```bash
    "enable_streaming": true,
    "enable_access_package_redirect": true
```
4. You will need to make sure mailgun is set up and configured. 
5. You will need to create at least 2 manual task integrations which require attachments to complete. 
6. Create a Privacy request and upload attachments with the same name (multiple to the same task and at least one to the other task)
7. verify that that they are all present under the correct data sets - run with html to verfiy that the html version shows the correct attachments with the correct data set and they are all available under the attachments tab.

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:

### DIFF
--- a/tests/ops/service/privacy_request/test_dsr_report_builder.py
+++ b/tests/ops/service/privacy_request/test_dsr_report_builder.py
@@ -456,17 +456,20 @@ class TestDsrReportBuilderAttachments(TestDsrReportBuilderBase):
             # Verify that attachments directory exists but contains no actual attachment files
             # (only the index.html file should be present)
             attachment_files = [
-                name for name in zip_file.namelist()
+                name
+                for name in zip_file.namelist()
                 if name.startswith(f"{common_assertions['paths']['attachments_dir']}/")
             ]
             # Should only have the index.html file, no actual attachment files
             assert len(attachment_files) == 1
-            assert attachment_files[0] == f"{common_assertions['paths']['attachments_dir']}/index.html"
+            assert (
+                attachment_files[0]
+                == f"{common_assertions['paths']['attachments_dir']}/index.html"
+            )
 
             # Verify that the attachments index page exists and is accessible
             self.assert_file_in_zip(
-                zip_file,
-                f"{common_assertions['paths']['attachments_dir']}/index.html"
+                zip_file, f"{common_assertions['paths']['attachments_dir']}/index.html"
             )
 
 


src/fides/api/service/storage/streaming/smart_open_streaming_storage.py
```diff
@@ -943,7 +952,30 @@ def _process_attachment_safely(
...
+                else:
+                    dataset_name = "unknown"
+
+            if dataset_name not in self.used_filenames_per_dataset:
+                self.used_filenames_per_dataset[dataset_name] = set()
+
+            unique_filename = get_unique_filename(
+                original_filename, self.used_filenames_per_dataset[dataset_name]
+            )
+            self.used_filenames_per_dataset[dataset_name].add(unique_filename)
```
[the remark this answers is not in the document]
JadeCara: Updated the function - that was supposed to be removed already.

src/fides/api/service/privacy_request/dsr_package/dsr_report_builder.py
```diff
@@ -239,16 +272,40 @@ def _add_collection(
...
             ):
                 # Process attachments and get their URLs
                 attachment_links = self._write_attachment_content(
                     item_data["attachments"],
                     f"data/{dataset_name}/{collection_name}",
+                    dataset_name,
                 )
                 # Add the attachment URLs to the item data
                 item_data["attachments"] = attachment_links
+            else:
```
JadeCara, line 357: I have a clean up in the pipeline for these lines, but I want to make sure it works before pushing it up. - waiting on the local env to spin up.

src/fides/api/service/privacy_request/dsr_package/templates/attachments_index.html
```diff
@@ -14,15 +14,15 @@
...
+            <p class="expiration-notice">Note: All download links will expire in {{ download_link_ttl_days }} days.</p>
             <div class="table table-hover">
                <div class="table-row">
                   <div class="table-cell" style="text-align: left;">File Name</div>
                   <div class="table-cell" style="text-align: left;">Size</div>
                </div>
                {% for name, info in data.items() %}
-               <a href="{{ info.url }}" class="table-row" target="_blank">
-                  <div class="table-cell" style="text-align: left;">{{ name }}</div>
+               <a href="{% if info.url.startswith('http') %}{{ info.url }}{% else %}{{ name }}{% endif %}" class="table-row" target="_blank">
```
tvandort: Is this safe? I don't know enough about jinja(?) to know if I can expect these segments of text to be URI encoded. Specifically I'm concerned that somehow Name could be handled in a way that might allow some sort of injection attack. 

If jinja does not URI encode, I think that we should have a property added to the object here that is the correct URL format instead of trying to heuristically identify which "mode" we're in at this point.

JadeCara: OK - pushing my changes up because they do have a fix for this but arent tested locally yet - this env takes forever to spin up

src/fides/api/service/privacy_request/dsr_package/dsr_report_builder.py
```diff
@@ -195,27 +190,59 @@ def _write_attachment_content(
...
+                unique_filename = self.processed_attachments[attachment_key]
+
+            # Set attachment URL based on streaming mode
+            if self.enable_streaming:
+                depth = directory.count("/") + 1 if directory.startswith("data/") else 0
+                attachment_url = (
+                    "../" * depth + f"attachments/{unique_filename}"
+                    if depth
+                    else unique_filename
+                )
```
tvandort: Thought for the future, we may be able to obviate the need for the calculation by setting a `<base href="../" />` in the document and then references `attachments/{unique_filename}` directly, but for now this totally works. 👍🏻

src/fides/api/service/privacy_request/dsr_package/dsr_report_builder.py
```diff
@@ -70,15 +72,30 @@ def __init__(
...
+
+        # Track attachments by their unique identifier to prevent duplicate processing
+        # Maps (download_url, file_name) -> unique_filename
+        self.processed_attachments: dict[tuple[str, str], str] = {}
 
-        # Track used filenames across all attachments
-        self.used_filenames: set[str] = set()
+    def _get_download_link_ttl_days(self) -> int:
+        """Get the download link TTL in days from the security configuration."""
+        from fides.config import CONFIG
```
tvandort: I'm curious, when in Python do we typical using this "import near the code using it" vs "import at the top of the file"? Fine with this either way, just want to understand better myself.

JadeCara: Ugh! no! its always at the top.

JadeCara: Fixed

src/fides/api/task/manual/manual_task_graph_task.py
```diff
@@ -398,25 +397,28 @@ def _aggregate_submission_data(
...
+                        "download_url": str(url) if url else None,
+                        "file_size": size,
+                    }
+                )
             except Exception as exc:  # pylint: disable=broad-exception-caught
                 logger.warning(
                     f"Error retrieving attachment {attachment.file_name}: {str(exc)}"
                 )
-        return attachment_map or None
+        return attachment_list or None
```
tvandort, line 421: TIL `[] or None` evaluates to `None` 🤯

tvandort (approved):

src/fides/api/service/privacy_request/dsr_package/dsr_report_builder.py
```diff
@@ -195,27 +189,59 @@ def _write_attachment_content(
...
-            if isinstance(file_size, (int, float)):
-                file_size = format_size(float(file_size))
+            file_size = (
+                format_size(float(file_size))
+                if isinstance(file_size, (int, float))
+                else "Unknown"
+            )
+
+            # Get or generate unique filename
+            attachment_key = (download_url, file_name)
```
tvandort: This is fine for now, but I think in the future we really want to be able to not rely on "natural keys" for key-like behavior. We ought to find a way to pass the surrogate `Attachment.id` key here to be used.

Vagoasdf:
During manual testing we found an issue: When we set up multiple files with the same name, only the first one uploaded goes ahead in the final request. 

Here we uploaded two files with the same name "Thisone.pdf" to the two tasks.

Captura de pantalla 2025-09-02 a la(s) 6 35 25 p  m

With the final result only showing one of the two 

Captura de pantalla 2025-09-02 a la(s) 6 34 51 p  m

tvandort:
> During manual testing we found an issue: When we set up multiple files with the same name, only the first one uploaded goes ahead in the final request.
> 
> Here we uploaded two files with the same name "Thisone.pdf" to the two tasks.
> Captura de pantalla 2025-09-02 a la(s) 6 35 25 p  m
> 
> With the final result only showing one of the two
> Captura de pantalla 2025-09-02 a la(s) 6 34 51 p  m

Good catch, thank you!

galvana (approved):

PR #6541, opened 2025-09-04
Decode S3 storage URLs when building object keys for streaming attachments into download packages

NevilleS:
### Description Of Changes

We found that uploading a spreadsheet "Untitled spreadsheet.xlsx" was resulting in a 0 byte file in the download package at CN and created this ticket: https://ethyca.atlassian.net/browse/ENG-1313

We reproduced the same behavior in staging, see some notes here: https://ethyca.atlassian.net/browse/ENG-1313?focusedCommentId=61421

With some experimentation, found that the filename is causing streaming errors. you can see logs of that in datadog in staging: [datadog link](https://app.datadoghq.com/logs?query=%22Failed%20to%20stream%22&agg_m=count&agg_m_source=base&agg_t=count&cols=host%2Cservice&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&from_ts=1756979814893&to_ts=1756994214893&live=true)

Verified the tests failed before change and passed after change. Also verified in manual testing. 

### Code Changes

- Added tests around spaces and special characters in names
- Added some error logic for time outs or read errors for streaming content. 
- Removed some stale artifacts from the code. 
- Updated the code to decode url for special characters and spaces. 

### Steps to Confirm

1.  Running on FidesPlus
2. Set AWS S3 as default storage config with enable streaming and enable redirect set to true.
3. Have a manual task integration set up for attachments
4. Run a DSR and attach a file with a name containing spaces/special chars etc
5. verify the file is present in the redirect download. 
Screenshot 2025-09-04 at 9 48 24 AM

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:

### DIFF
--- a/src/fides/api/service/privacy_request/request_runner_service.py
+++ b/src/fides/api/service/privacy_request/request_runner_service.py
@@ -268,7 +268,7 @@ def upload_access_results(
             connection_key=None,
             dataset_name="Access package upload",
             collection_name=None,
-            message="Access package upload failed for privacy request.",
+            message=f"Access package upload failed for privacy request: {Pii(str(exc))}",
             action_type=ActionType.access,
         )
         privacy_request.status = PrivacyRequestStatus.error
--- a/src/fides/api/service/privacy_request/request_runner_service.py
+++ b/src/fides/api/service/privacy_request/request_runner_service.py
@@ -261,14 +261,14 @@ def upload_access_results(
             "Error uploading subject access data for rule {} on policy {}: {}",
             rule.key,
             policy.key,
-            Pii(str(exc)),
+            str(exc),
         )
         privacy_request.add_error_execution_log(
             session,
             connection_key=None,
             dataset_name="Access package upload",
             collection_name=None,
-            message=f"Access package upload failed for privacy request: {Pii(str(exc))}",
+            message=f"Access package upload failed for privacy request: {str(exc)}",
             action_type=ActionType.access,
         )
         privacy_request.status = PrivacyRequestStatus.error
--- a/src/fides/api/service/storage/streaming/schemas.py
+++ b/src/fides/api/service/storage/streaming/schemas.py
@@ -75,7 +75,7 @@ class StreamingBufferConfig(BaseModel):
         description="Chunk size threshold in bytes (default: 1MB)",
     )
     fail_fast_on_attachment_errors: bool = Field(
-        default=False,
+        default=True,
         description="If True, stop processing on first attachment error. If False, continue with error placeholders.",
     )
     include_error_details: bool = Field(
--- a/src/fides/api/service/storage/streaming/schemas.py
+++ b/src/fides/api/service/storage/streaming/schemas.py
@@ -56,21 +56,6 @@ class StorageUploadConfig(BaseModel):
         return v.strip()
 
 
-class PackageSplitConfig(BaseModel):
-    """Configuration for package splitting operations"""
-
-    max_attachments: int = Field(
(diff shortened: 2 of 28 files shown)


src/fides/api/service/storage/streaming/schemas.py
```diff
@@ -52,21 +56,6 @@ def validate_storage_identifiers(cls, v: str) -> str:
         return v.strip()
 
 
-class PackageSplitConfig(BaseModel):
```
JadeCara, line 55: Artifact no longer used with smart open

src/fides/api/service/storage/streaming/smart_open_streaming_storage.py
```diff
@@ -143,103 +156,6 @@ def _convert_to_stream_zip_format(
                 [content_bytes]
             )
 
-    def build_attachments_list(
```
JadeCara, line 146: Artifact - no longer used with smart open

src/fides/api/service/privacy_request/request_runner_service.py
```diff
@@ -268,7 +268,7 @@ def upload_access_results(
             connection_key=None,
             dataset_name="Access package upload",
             collection_name=None,
-            message="Access package upload failed for privacy request.",
+            message=f"Access package upload failed for privacy request: {Pii(str(exc))}",
```
galvana: Let's get rid of this `Pii` wrapper, it just makes it harder to troubleshoot things

JadeCara: Even though we are using it in the logging just above?

galvana: Yeah, we should get rid of it from the previous log too

JadeCara: Done :)

src/fides/api/service/storage/streaming/smart_open_streaming_storage.py
```diff
@@ -294,23 +210,75 @@ def _create_attachment_content_stream(
...
+
         try:
             with self.storage_client.stream_read(bucket, key) as content_stream:
                 # Stream in chunks instead of reading entire file
-                chunk_count = 0
-                total_bytes = 0
-                while True:
-                    chunk = content_stream.read(self.chunk_size)
+                chunk_count = total_bytes = 0
+                max_chunks = 100000  # Safety limit to prevent infinite loops
```
galvana: This should be calculated instead of hard-coded. It's possible to misconfigure this in a way that we can't download a file of size `LARGE_FILE_THRESHOLD` with 100000 chunks of size `self.chunk_size`

`max_chunks = LARGE_FILE_THRESHOLD // self.chunk_size + 1`

src/fides/api/service/storage/streaming/schemas.py
```diff
@@ -85,6 +74,25 @@ class StreamingBufferConfig(BaseModel):
         ge=1024 * 1024,
         description="Chunk size threshold in bytes (default: 1MB)",
     )
+    fail_fast_on_attachment_errors: bool = Field(
+        default=False,
```
NevilleS: should this default to true? I don't think it's possible to configure this right now right

JadeCara: Yeah - i think that makes sense. Updating.

PR #6542, opened 2025-09-04
Create table to store web monitor scan data

erosselli:
Closes [ENG-1212](https://ethyca.atlassian.net/browse/ENG-1212)

### Description Of Changes

This PR creates a new model and table (`WebMonitorGroupJob`) to store web monitor scan data; for now it'll only store the experiences data obtained in a given scan, but it could be extended to other cases as well. 

I also added two new consent statuses that will be needed for some upcoming web monitor consent work. 

### Steps to Confirm

1.  Run migrations, make sure everything looks ok and tests pass 
2. Downgrade both migrations, make sure everything is still ok 

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:

[ENG-1212]: https://ethyca.atlassian.net/browse/ENG-1212?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/fides/api/models/detection_discovery/web_monitor.py
```diff
@@ -0,0 +1,65 @@
...
+
+from sqlalchemy import Boolean, Column, String, insert, select
+from sqlalchemy.dialects.postgresql import JSONB
+from sqlalchemy.ext.asyncio import AsyncSession
+from sqlalchemy.ext.declarative import declared_attr
+
+from fides.api.db.base_class import Base
+
+
+class WebMonitorGroupJob(Base):
```
erosselli, line 11: I debated with myself a lot trying to land on a name that made sense... not sure I achieved that tbh

tests/api/conftest.py
```diff
@@ -0,0 +1,10 @@
+import pytest
+import requests
+
+
+@pytest.fixture(scope="session")
+def setup_db(api_client, config):
```
erosselli, line 6: needed to run the `api/models` tests on their own

src/fides/api/alembic/migrations/versions/67d0e389b003_update_consentstatus_enum.py
```diff
@@ -0,0 +1,47 @@
...
+    op.execute("ALTER TYPE consentstatus ADD VALUE 'pre_consent'")
+    op.execute("ALTER TYPE consentstatus ADD VALUE 'cmp_error'")
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    # Remove any records that have the new enum values before dropping them
+    op.execute(
+        "UPDATE asset SET consent_status = 'unknown' WHERE consent_status IN ('pre_consent', 'cmp_error')"
```
erosselli, line 30: I'm only updating the `asset` table model here.... but technically we have staged resources that will have this consent status in a key as part of the JSON column . Wasn't sure what the best way to migrate those was, or if I can get away with not doing it ? idk how likely it is that someone will run a web monitor scan and then want to downgrade this?

src/fides/api/models/detection_discovery/web_monitor.py
adamsachs: heh, this isn't bad but my only concern here would be that the `Group` could also be considered a 'group of jobs', i.e. a `JobGroup` 🤯 .

i think i'd either go fully on the side of 'describe the web monitor internals' or the other side of 'abstract away all web monitor internals'...so the former would be something like `WebMonitorTaskGroup`? since that's a bit more precise about what these _actually are_ within the web monitor; the latter would look more like `WebMonitorJob`, since from Fides' POV...that's what ultimately what we care about? 🤷 

idk, i never wanna get too crazy with naming, and i can't say i'm convinced of my suggestions. but i'll leave it as some food for thought!

src/fides/api/models/detection_discovery/web_monitor.py
```diff
@@ -0,0 +1,61 @@
...
+    # The id for the web monitor group on the web monitor service
+    group_id = Column(String, nullable=False, unique=True)
+
+    # The raw experiences data obtained for that group from the web monitor service
+    raw_experiences_data = Column(JSONB, nullable=True)
+    # The processed experiences data, used to calculate an StagedResource's consent status
+    processed_experiences_data = Column(JSONB, nullable=True)
+
+    # Whether this data is from a test_website monitor rather than a real one
+    is_test_run = Column(Boolean, nullable=False, default=False)
```
adamsachs, line 27: 👍

src/fides/api/models/detection_discovery/web_monitor.py
```diff
@@ -0,0 +1,61 @@
...
+
+class WebMonitorGroupJob(Base):
+    """SQL model for storing data returned from a web monitor group job"""
+
+    @declared_attr
+    def __tablename__(self) -> str:
+        return "web_monitor_group_job"
+
+    # The id for the web monitor group on the web monitor service
+    group_id = Column(String, nullable=False, unique=True)
```
adamsachs, line 19: OK, sorry to request a potential scope increase: but something we _may_ want to look at knocking off while we're here - could we link these records up to the `MonitorTask` records that correspond to the Fides execution? 

that's something that's been lingering in the background as a useful [QOL update](https://ethyca.atlassian.net/browse/ENG-212) for web monitor execution is actually persisting (and being able to retrieve via API) the web monitor's `group_id` with the Fides-side execution record. 

as you can imagine, for troubleshooting purposes it'd be great to say "oh, fides is reporting error 'foo' in a web monitor task, looks like it was web monitor group  `x`, let's look up group `x` on the web monitor API/logs to see more about it)..." and in order to do that now, we've got to search through _fides_ logs, which we don't always have access to!

src/fides/api/alembic/migrations/versions/67d0e389b003_update_consentstatus_enum.py
adamsachs: i think we can get away with not doing it. even in the case you suggest, i think it'd be reasonable recourse to request that customers clear out and re-run their web monitor.

src/fides/api/models/detection_discovery/web_monitor.py
erosselli: discussed on slack and there's a couple other missing pieces to be able to do this , so decided against it for the moment to not add more scope creep

PR #6546, opened 2025-09-04
Adds safeguards when encountering rate limit errors upon instantiation

eastandwestwind:
Closes https://ethyca.atlassian.net/browse/ENG-1315

### Description Of Changes

Adds safeguards when encountering rate limit errors upon instantiation. Allows rate_limiting to still function given a Redis pass does not include special chars like `[` or `]`

### Code Changes

* Uses disabled `Limiter` in the cases where either 1. rate limiting is disabled, or 2. we encounter any errors when instantiating the redis-driven `Limiter`

### Steps to Confirm

1.  In `docker_compose.yml`, update `services` -> `redis` -> `command` password to an invalid / unsupported password that contains `[` or `]`, e.g. `command: redis-server --requirepass "passwor[d"`
2. In your `.env`, set the same pass in the `FIDES__REDIS__PASSWORD` var
3. Remove any existing `FIDES__SECURITY__RATE_LIMIT_CLIENT_IP_HEADER` rate limit env var
4. Start up fides using `nox -s dev`
5. Observe fides is able to start up just fine

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:

### DIFF
--- a/src/fides/api/util/rate_limit.py
+++ b/src/fides/api/util/rate_limit.py
@@ -180,15 +180,32 @@ is_rate_limit_enabled = (
     CONFIG.security.rate_limit_client_ip_header is not None
     and CONFIG.security.rate_limit_client_ip_header != ""
 )
-fides_limiter = Limiter(
-    storage_uri=CONFIG.redis.connection_url_unencoded,
-    application_limits=[
-        CONFIG.security.request_rate_limit
-    ],  # Creates ONE shared bucket for all endpoints
+
+disabled_limiter = Limiter(
+    default_limits=[CONFIG.security.request_rate_limit],
     headers_enabled=True,
     key_prefix=CONFIG.security.rate_limit_prefix,
     key_func=safe_rate_limit_key,
     retry_after="http-date",
-    in_memory_fallback_enabled=False,  # Fall back to no rate limiting if Redis unavailable
-    enabled=is_rate_limit_enabled,
+    enabled=False,
 )
+
+try:
+    if is_rate_limit_enabled:
+        fides_limiter = Limiter(
+            storage_uri=CONFIG.redis.connection_url_unencoded,
+            application_limits=[
+                CONFIG.security.request_rate_limit
+            ],  # Creates ONE shared bucket for all endpoints
+            headers_enabled=True,
+            key_prefix=CONFIG.security.rate_limit_prefix,
+            key_func=safe_rate_limit_key,
+            retry_after="http-date",
+            in_memory_fallback_enabled=False,  # Fall back to no rate limiting if Redis unavailable
+            enabled=is_rate_limit_enabled,
+        )
+    else:
+        fides_limiter = disabled_limiter
+except Exception as e:
+    logger.warning("Disabling rate limiting due to Redis configuration issue: {}", e)
+    fides_limiter = disabled_limiter
--- a/src/fides/api/util/rate_limit.py
+++ b/src/fides/api/util/rate_limit.py
@@ -207,5 +207,6 @@ try:
     else:
         fides_limiter = disabled_limiter
 except Exception as e:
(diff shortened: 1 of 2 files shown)


src/fides/api/util/rate_limit.py
```diff
@@ -180,15 +180,32 @@ async def dispatch(self, request: Request, call_next):  # type: ignore
...
+disabled_limiter = Limiter(
+    default_limits=[CONFIG.security.request_rate_limit],
     headers_enabled=True,
     key_prefix=CONFIG.security.rate_limit_prefix,
     key_func=safe_rate_limit_key,
     retry_after="http-date",
-    in_memory_fallback_enabled=False,  # Fall back to no rate limiting if Redis unavailable
-    enabled=is_rate_limit_enabled,
+    enabled=False,
 )
```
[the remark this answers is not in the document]
eastandwestwind: this is fine as it's a placeholder, it's not actually limiting anything since `enabled=False`

src/fides/api/util/rate_limit.py
```diff
@@ -180,15 +180,32 @@ async def dispatch(self, request: Request, call_next):  # type: ignore
...
+            key_func=safe_rate_limit_key,
+            retry_after="http-date",
+            in_memory_fallback_enabled=False,  # Fall back to no rate limiting if Redis unavailable
+            enabled=is_rate_limit_enabled,
+        )
+    else:
+        fides_limiter = disabled_limiter
+except Exception as e:
+    logger.warning("Disabling rate limiting due to Redis configuration issue: {}", e)
+    fides_limiter = disabled_limiter
```
adamsachs: i think we need to be noisier here because this implies that the intention is to configure rate limiting. that is likely a major concern for the user - and a simple warning message could very easily slip by!

i actually think that if `is_rate_limit_enabled=True`, it probably _does_ make sense to fail startup - we can't really rely on someone finding this in logs, even if it's logged with higher severity. 


```suggestion
    logger.exception("Error instantiating rate limiter: {}", e)
    if is_rate_limit_enabled:
        raise e
```

src/fides/api/util/rate_limit.py
```diff
@@ -180,15 +180,33 @@ async def dispatch(self, request: Request, call_next):  # type: ignore
...
+            ],  # Creates ONE shared bucket for all endpoints
+            headers_enabled=True,
+            key_prefix=CONFIG.security.rate_limit_prefix,
+            key_func=safe_rate_limit_key,
+            retry_after="http-date",
+            in_memory_fallback_enabled=False,  # Fall back to no rate limiting if Redis unavailable
+            enabled=is_rate_limit_enabled,
+        )
+    else:
+        fides_limiter = disabled_limiter
```
adamsachs, line 208: nit: i'd probably throw an `INFO` or `DEBUG` log in here just to help with troubleshooting, especially because `is_rate_limit_enabled` is _inferred_ via a setting and not something that is explicitly set (i could imagine us changing the criteria at some point!).  
```suggestion
        logger.info("Fides API rate limiting is not enabled")
        fides_limiter = disabled_limiter
```

eastandwestwind: hmm we already have a log on our main `fides/api/app_setup.py` file when rate limiting is disabled-
```
logger.warning(
            "Rate limiting client IPs is disabled because the FIDES__SECURITY__RATE_LIMIT_CLIENT_IP_HEADER env var is not configured."
        )
```


so this would result in 2 of the same logs. I'd prefer to just have 1 log when rate limiting is disabled

adamsachs: oh whoops, had no idea about that. looks good to me, ignore my comment 👍

PR #6552, opened 2025-09-08
ENG-1334, ENG-1329, ENG-1328

tvandort:
Closes:
- ENG-1334
- ENG-1329
- ENG-1328 

### Description Of Changes

- Request manager respects timezones
- Added Spinner to FideTableV2 for loading feedback
- Request manager persists pagination parameters in URL
- Request manager persists search to URL
- Request manager no longer persists search to local storage
- Refactored CSV download function 
- Clean up Object URLs to prevent memory leaks
- ~Implemented reusable paginator functions that don't need to be rewired per use~

### Code Changes

#### ENG-1334
1. Go to the Request management page
2. Page through results
3. Observe URL change
4. Copy and paste URL or refresh page to ensure you land where you were 

#### ENG-1329
1. Go to the request manager page
1. Use Chrome's Sensors feature (or change your device time to another region)
2. Select a date range. 
3. Look at the network panel and ensure that there is an offset in the timestamp or ensure that requests starting from the morning of the mocked time are visible
4. Download a CSV and ensure the same behavior as above.

##### Search isn't persisted indefinitely
1. Go to request manager page
1. Type a privacy request ID or subject identity
1. Navigate away from the page and back
2. Typed search should disappear
3. Click back search should reappear

##### Search is persisted to URL
1. Go to request manager page
2. Type a privacy request id or subject identity
3. Copy the URL and paste it into another tab
4. observe that the search filter is applied in new tab

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [ ] No followup issues
* Database migrations:

  * [ ] No migrations
* Documentation:


clients/admin-ui/src/features/privacy-requests/hooks/useDownloadPrivacyRequestReport.ts
```diff
@@ -0,0 +1,35 @@
...
+        "A problem occurred while generating your privacy request report.  Please try again.",
+      );
+      toast({ status: "error", description: message });
+    } else {
+      const a = document.createElement("a");
+      const csvBlob = new Blob([result.data], { type: "text/csv" });
+      a.href = window.URL.createObjectURL(csvBlob);
+      a.download = `privacy-request-report.csv`;
+      a.click();
+      a.remove();
```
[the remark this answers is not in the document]
tvandort: Neat, I had no idea that was a function.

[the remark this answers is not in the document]
tvandort: Though, I'm slightly wary of using this. How do I know when the user has completed downloading the file?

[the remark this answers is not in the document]
tvandort: I think it's correct. I'll do some manual testing.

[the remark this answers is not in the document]
tvandort: That being said, I'd love it if we had some sort of better pattern for this, e.g. temp token links on the anchor tag itself.

clients/admin-ui/src/features/common/table/v2/FidesTable.tsx
```diff
@@ -389,125 +392,129 @@ export const FidesTableV2 = <T,>({
...
-    >
-      <Table
-        variant="unstyled"
-        style={{
-          borderCollapse: "separate",
-          borderSpacing: 0,
-          ...columnSizeVars,
-          minWidth: "100%",
-        }}
+    <Spin spinning={loading}>
```
lucanovera, line 395: This change isn't listed in the PR description

clients/admin-ui/src/features/consent-reporting/hooks/useConsentReportingDownload.ts
```diff
@@ -27,9 +27,11 @@ const useConsentReportingDownload = () => {
     } else {
       const a = document.createElement("a");
       const csvBlob = new Blob([result.data], { type: "text/csv" });
-      a.href = window.URL.createObjectURL(csvBlob);
+      const csvUrl = window.URL.createObjectURL(csvBlob);
```
lucanovera, line 30: This change isn't listed in the PR description

clients/admin-ui/src/features/privacy-requests/hooks/useDownloadPrivacyRequestReport.ts
```diff
@@ -0,0 +1,37 @@
+import { useToast } from "fidesui";
+
+import { getErrorMessage } from "~/features/common/helpers";
+
+import { useLazyDownloadPrivacyRequestCsvQuery } from "../privacy-requests.slice";
+
+const useDownloadPrivacyRequestReport = () => {
```
lucanovera, line 7: Nice cleanup of a funcion that shouldn't have been in the slice file.

clients/admin-ui/src/features/common/pagination/PaginationProvider.tsx
```diff
@@ -0,0 +1,64 @@
+import React, { PropsWithChildren, useContext } from "react";
+
+import { useAntPagination } from "./useAntPagination";
+import { usePagination } from "./usePagination";
+
+const PaginationContext = React.createContext<
+  undefined | ReturnType<typeof usePagination>
+>(undefined);
+
+export const usePaginationContext = () => {
```
lucanovera: What is the need for adding a pagination context? It looks like it's adding a lot of complexity and it's not really handling a global state or passing information down many layers of components.

tvandort: The reason I'm doing this is because I don't want to have to rewire the paginator on every page that I use it. I want to just drop the paginator component on the page and have it work.

IMO this will reduce complexity as each page doesn't have to funnel props to a paginator component or pass "paging reset" functions throughout the prop tree. Any component that needs to reset pagination settings in response to some change can easily use the context hook to dispatch commands to the paginator.

clients/admin-ui/src/features/consent-reporting/hooks/useConsentReportingDownload.ts
tvandort: Updated description.

clients/admin-ui/src/features/common/pagination/PaginationProvider.tsx
lucanovera: I get that the pagination props were messy, there were lots of props being passed around. I created the ENG-47 because of that. But I think the problem is in the lack of standarization for the pagination state object, not the state flow itself.
We should have the hook return 1 standard object with the state and the functions to update said state. Then we can have the Table and Pagination components work with that object, something like this:

```
const pagination = usePagination();
return <>
<Table pagination={pagination}>
or
<AntPagination pagination={pagination}>
<>
```

Adding the context adds complexity, it's not drop in if you have to add a new provider to every page where you want to use a pagination component.

tvandort: In my experience nearly every project I've worked on eventually ends up with some component on a page somewhere that wants to reset the pagination state from an arbitrary depth in the component tree. 

I'll reverse the changes so we can move on but ultimately for a cross cutting concern I disagree with the notion that:

- It's better to require every page use the hook and pass a specific prop to the pagination component
- That requiring prop drilling to pass a page reset function to an arbitrary point in the component tree is better than having a hook that extracts it from context
- It's more complex to add a provider to a page than requiring a hook be used and the state passed in a "distant" location down the page in code.

IME the testing experience of being able to more easily mock out the context is also a better separation of concerns.

clients/admin-ui/src/features/common/table/v2/FidesTable.tsx
tvandort: Updated description.

clients/admin-ui/src/features/privacy-requests/RequestTable.tsx
```diff
@@ -113,13 +107,12 @@ export const RequestTable = ({ ...props }: BoxProps): JSX.Element => {
   const handleSort = (columnSort: ColumnSort) => {
     if (!columnSort) {
       dispatch(clearSortKeys());
-      resetPageIndexToDefault();
```
tvandort, line 116: @lucanovera Do you have any insight as to when / where this reset scenario would matter? This reset was what was causing the page and size not to stay in the URL / page.

lucanovera: It looks like it was there to reset the page to the first page when you add/change/remove a column sort. I think it is still relevant and we should leave it. But, this handleSort function should be called when the page load. I would expect it to only be called if there is a change in the sort status.

tvandort: The pagination resets without this line so I'm going to assume it's not longer necessary.

PR #6566, opened 2025-09-09
Async polling

Vagoasdf:
Closes [ENG-570](https://ethyca.atlassian.net/browse/ENG-570)

### Description Of Changes

Adding async polling support for SaaS integrations

### Code Changes

* A new `AsyncDSRStrategy` interface with `async_retrieve_data` and `async_mask_data` methods. The existing async callback logic was refactored to conform to this new interface.
  * `AsyncPollingStrategy` and `AsyncCallbackStrategy`
* `PollingRequestHandler` and `PollingResponseProcessor` to focus on pure request and response processing without business logic.
  * `PollingRequestHandler`
    * Populates and executes SaaS requests
    * Supports request overrides for custom status or result logic
  * `PollingResponseProcessor`
    * Automatically infers the data type from the response (JSON, CSV, XML, attachments)
    * Uses existing storage and attachment services

### Steps to Confirm

1. Two end-to-end workflows are covered in `tests/integration/workflows/test_privacy_request_with_async_callback.py` and `tests/integration/workflows/test_privacy_request_with_async_polling.py`

### Pre-Merge Checklist

* [ ] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:

[ENG-570]: https://ethyca.atlassian.net/browse/ENG-570?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


clients/admin-ui/src/features/privacy-requests/events-and-logs/hooks/usePrivacyRequestComments.ts
```diff
@@ -55,6 +55,7 @@ export const usePrivacyRequestComments = (privacyRequestId: string) => {
           isError: false,
           isSkipped: false,
           isAwaitingInput: false,
+          isPolling: false,
```
galvana, line 58: Most of these front-end changes are related to the new polling status

clients/admin-ui/src/features/privacy-requests/hooks/useDownloadPrivacyRequestResults.ts
```diff
@@ -30,9 +30,14 @@ const useDownloadPrivacyRequestResults = ({
...
-      privacy_request_id: privacyRequest.id,
-    });
+    useGetPrivacyRequestAccessResultsQuery(
+      {
+        privacy_request_id: privacyRequest.id,
+      },
+      {
+        skip: !isCompletedRequest,
+      },
+    );
```
galvana, line 40: This allows the "Download request results" button to be enabled once the request is complete without needing to refresh the page.

noxfiles/ci_nox.py
```diff
@@ -361,14 +361,15 @@ def load_tests(session: nox.Session) -> None:
...
-        "ops-integration",
+        "ops-integration",  # Also runs tests/integration/
         "ops-external-datastores",
         "ops-saas",
     ],
     "tests/service/": ["misc-unit", "misc-integration", "misc-integration-external"],
     "tests/task/": ["misc-unit", "misc-integration", "misc-integration-external"],
     "tests/util/": ["misc-unit", "misc-integration", "misc-integration-external"],
     "tests/qa/": ["misc-unit", "misc-integration", "misc-integration-external"],
+    "tests/integration/": ["ops-integration"],  # Workflow integration tests
```
galvana, line 372: Adding the `tests/integration` directory for end to end user workflows. In this case, I added tests for end-to-end privacy requests with async polling and async callback.

qa/scenarios/create_external_reference.py
galvana, line 1: A QA scenario to create dataset reference integrations. This is to help test SaaS integrations that rely on external integrations without having to do all the setup manually.

src/fides/api/schemas/saas/async_polling_configuration.py
galvana, line 1: New schemas related to async polling
- `PollingStatusRequest` - A standard request with additional `status_path` and `status_completed_value`
- `PollingResultRequest` - A standard request with additional `result_path`

src/fides/api/schemas/saas/strategy_configuration.py
```diff
@@ -178,15 +178,3 @@ class OAuth2ClientCredentialsConfiguration(OAuth2BaseConfiguration):
...
-class PollingAsyncDSRConfiguration(StrategyConfiguration):
-    """
-    Configuration for polling async DSR requests.
-    """
-
-    status_request: SaaSRequest
-    status_path: str
-    status_completed_value: Optional[str] = None
-    result_request: SaaSRequest
-    result_path: str
```
galvana, line 192: Simplified this to
```
class AsyncPollingConfiguration(StrategyConfiguration):
    """
    Simplified configuration for polling async DSR requests.
    The main read request serves as the initial request.
    """

    status_request: PollingStatusRequest
    result_request: PollingResultRequest
```

src/fides/api/service/async_dsr/handlers/polling_request_handler.py
galvana, line 1: Responsible for populating/executing status and result HTTP requests.

src/fides/api/service/async_dsr/handlers/polling_response_handler.py
galvana, line 1: This file is responsible to interpreting different `Response` objects and determining if they should be parsed as rows, or treated as attachments.

src/fides/api/service/async_dsr/strategies/async_dsr_strategy.py
galvana, line 1: The base interface for async polling and async callback. The previously implemented async callback was refactored to fit into this interface.

src/fides/api/service/async_dsr/strategies/async_dsr_strategy_callback.py
galvana, line 1: The `async_retrieve_data` and `async_masking_data` methods each have two phases:

- `initial_async` - The first request to trigger the async flow
- `callback_completion_*` - Completes the request if data has been provided via the callback URL

src/fides/api/service/async_dsr/strategies/async_dsr_strategy_polling.py
galvana, line 1: The `async_retrieve_data` and `async_masking_data` methods each have two phases:

- `initial_async` - The first request to trigger the async flow
- `polling_continuation_*` - Executes the stored sub requests and retrieves the results if the status evaluates to complete. The results aren't stored on the RequestTask until all the sub requests are complete

src/fides/api/service/async_dsr/utils.py
galvana, line 1: Shared utils between the async polling and async callback strategies.

src/fides/api/service/connectors/fides/fides_client.py
galvana, line 1: `fides_client.py` relied on functions that were specified in `request_service.py`. For some reason that I timed out looking into, this caused a cyclic dependency. The fix was to bring in `get_async_client` and `poll_server_for_completion` into this file. These functions weren't used anywhere else except for the tests.

src/fides/api/service/connectors/query_configs/saas_query_config.py
```diff
@@ -140,7 +139,7 @@ def get_erasure_request_by_action(
             )
         return request
 
-    def get_masking_request(self, db: Session) -> Optional[SaaSRequest]:
```
galvana, line 144: The `get_masking_request` function hasn't relied on a DB session for while now. Cleaning this up so we didn't need to pass a session anymore. It was useful to do this for the new code.

src/fides/api/service/connectors/saas_connector.py
galvana, line 1: Simplified `SaaSConnector` by extracting out the async logic into `async_dsr_strategy`.

src/fides/api/service/connectors/saas_connector.py
```diff
@@ -216,19 +218,29 @@ def retrieve_data(
...
+        with get_db() as db:
+            if async_dsr_strategy := _get_async_dsr_strategy(
+                db, request_task, query_config, ActionType.access
+            ):
+                return async_dsr_strategy.async_retrieve_data(
+                    client=self.create_client(),
+                    request_task_id=request_task.id,
+                    query_config=query_config,
+                    input_data=input_data,
+                )
```
galvana, line 243: Using a new DB session for this. I was originally using the DB session that `request_task` was associated with but I kept running into `DetachedInstanceError`.

src/fides/api/service/privacy_request/attachment_handling.py
```diff
@@ -84,12 +84,19 @@ def get_attachments_content(
...
+                bucket_name = attachment.config.details[StorageDetails.BUCKET.value]
+            # For local storage, bucket_name is not needed
+
             yield AttachmentData(
                 file_name=attachment.file_name,
                 file_size=size,
                 download_url=str(url) if url else None,
                 content_type=attachment.content_type,
-                bucket_name=attachment.config.details[StorageDetails.BUCKET.value],
+                bucket_name=bucket_name or "",  # Empty string for local storage
```
galvana, line 99: Bug fix, local storage doesn't have a bucket name

src/fides/api/service/privacy_request/request_runner_service.py
```diff
@@ -922,78 +920,6 @@ def mark_paused_privacy_request_as_expired(privacy_request_id: str) -> None:
     db.close()
 
 
-def _retrieve_child_results(  # pylint: disable=R0911
```
galvana, line 925: Cleaning up unused function.

src/fides/api/service/privacy_request/request_service.py
```diff
@@ -706,7 +666,7 @@ def requeue_interrupted_tasks(self: DatabaseTask) -> None:
 
 
 @celery_app.task(base=DatabaseTask, bind=True)
-def poll_async_tasks_status(self: DatabaseTask) -> None:
+def requeue_polling_tasks(self: DatabaseTask) -> None:
```
galvana, line 680: Responsible for queueing up the polling request tasks. This is picked up by the `retrieve_data` and `mask_data` functions so we don't need to handle state management

src/fides/api/service/saas_request/saas_request_override_factory.py
```diff
@@ -30,6 +33,10 @@ class SaaSRequestType(Enum):
     UPDATE_CONSENT = "update_consent"
     PROCESS_CONSENT_WEBHOOK = "process_consent_webhook"
 
+    # Async polling request types
+    POLLING_STATUS = "polling_status"
+    POLLING_RESULT = "polling_result"
```
galvana, line 38: Adding two new request override types in case we need to register some custom status or result checks that aren't generic enough to be included in the async response handler logic.

src/fides/api/task/execute_request_tasks.py
```diff
@@ -162,7 +162,10 @@ def can_run_task_body(
     if request_task.is_root_task:
         # Shouldn't be possible but adding as a catch-all
         return False
-    if request_task.status != ExecutionLogStatus.pending:
+    if request_task.status not in [
+        ExecutionLogStatus.pending,
+        ExecutionLogStatus.polling,
```
galvana: To allow `polling` tasks to resume execution.

src/fides/api/task/filter_results.py
```diff
@@ -6,10 +6,43 @@
 
 from fides.api.graph.config import CollectionAddress, FieldPath
 from fides.api.graph.graph import DatasetGraph
+from fides.api.service.storage.util import is_attachment_field
 from fides.api.task.manual.manual_task_address import ManualTaskAddress
 from fides.api.util.collection_util import Row
 
 
+def should_skip_data_category_filtering(
```
galvana, line 14: A new helper function to prevent async polling attachments from being removed from the DSR package. We were previously only checking for manual task data but I made this more generic.

src/fides/api/task/graph_task.py
```diff
@@ -416,9 +429,13 @@ def update_status(
...
             if self.request_task.id:
-                # Merge the request task to ensure we keep the changes made
-                # to self.request_task that haven't yet been committed to the database
+                # Merge the request_task into the current session to make it persistent,
+                # then refresh its `async_type` to load the latest state from the
+                # database. This is crucial for async tasks where `async_type` might be
+                # updated by another process, and avoids overwriting local data like
+                # `access_data`.
                 request_task = db.merge(self.request_task)
+                db.refresh(request_task, attribute_names=["async_type"])
```
galvana, line 438: The downside of using a separate session inside of the async strategies is that we need to refresh the `async_type` attribute here. The other data is available but for some reason `async_type` needs this manual refresh. A bit hacky but it works.

src/fides/config/execution_settings.py
galvana, line 1: The configs will need a rework before we merge, these are set low just for development. I'm thinking the intervals should be in hours for the polling since some results can take weeks to come back.

tests/integration/workflows/data/example_async_callback_config.yml
galvana, line 1: Some sample configs and datasets to run end-to-end privacy request tests.

tests/integration/workflows/test_privacy_request_with_async_callback.py
galvana, line 1: We were already covering this in other tests but I wanted to have one that:
- Was in the same location and style as the complementary polling test
- In the new location for end to end user flow tests

tests/integration/workflows/test_privacy_request_with_async_polling.py
```diff
@@ -0,0 +1,184 @@
...
+from fides.api.models.privacy_request.request_task import RequestTaskSubRequest
+from fides.api.models.sql_models import Dataset as CtlDataset
+from fides.api.models.worker_task import ExecutionLogStatus
+from fides.api.schemas.privacy_request import PrivacyRequestStatus
+from fides.api.service.privacy_request.request_service import requeue_polling_tasks
+from tests.conftest import wait_for_privacy_request_status
+from tests.ops.test_helpers.saas_test_utils import MockAuthenticatedClient
+
+
+@pytest.mark.async_dsr
```
galvana, line 27: I added a Pytest marker to be able to run all the Async DSR tests in one run

tests/ops/service/connectors/test_fides_connector.py
```diff
@@ -130,9 +130,7 @@ def test_retrieve_data(
         # make requests to the running webserver which is connected to the application db,
         # but we need them to talk to the test db in pytest
         monkeypatch.setattr(Client, "send", api_client.send)
-        monkeypatch.setattr(
-            request_service, "get_async_client", lambda: async_api_client
-        )
+        monkeypatch.setattr(fides_client, "get_async_client", lambda: async_api_client)
```
galvana, line 133: Updating the patch location after the refactor

tests/ops/test_helpers/saas_test_utils.py
```diff
@@ -30,6 +61,167 @@ def poll_for_existence(
     return return_val
 
 
+class MockAuthenticatedClient(Mock):
```
galvana, line 65: Replicating the `MockAuthenticatedClient` from Fidesplus. Extended it to be able to mock attachment responses.

tests/qa/test_scenarios.py
```diff
@@ -34,6 +34,9 @@ def create_scenario_instance(scenario_class: Type[QATestScenario]) -> QATestScen
 SCENARIOS = discover_scenarios()
 
 
+@pytest.mark.xfail(
+    reason="These tests require too many external services to run. Skipping for now."
+)
```
galvana, line 39: My new QA scenario needs `postgres_example` to run. I'll address this in a later PR

tests/conftest.py
```diff
@@ -813,6 +819,37 @@ def all_tasks_have_run(tasks: Query) -> bool:
             raise DSRThreeTestRunnerTimedOut()
 
 
+def wait_for_privacy_request_status(
```
galvana, line 822: New test helper, waiting for the privacy request to be in a specific status. We might not need it, I think things happen synchronously in the tests?

clients/admin-ui/src/features/privacy-requests/events-and-logs/hooks/usePrivacyRequestComments.ts
Vagoasdf: Its great to have a easy visual indicator if the status!

src/fides/api/service/connectors/saas_connector.py
Vagoasdf: The management of DBs sessions is strange here, I remember getting into similar errors when managing the request data aswell

clients/admin-ui/src/features/privacy-requests/types.ts
```diff
@@ -46,6 +48,7 @@ export const ExecutionLogStatusColors: Record<
   [ExecutionLogStatus.COMPLETE]: undefined,
   [ExecutionLogStatus.PAUSED]: undefined,
   [ExecutionLogStatus.RETRYING]: undefined,
+  [ExecutionLogStatus.POLLING]: undefined,
```
erosselli: I'm confused, we add the warning colors in the scss file but not here? 🤔

noxfiles/setup_tests_nox.py
```diff
@@ -157,7 +157,7 @@ def pytest_ops(
             run_tests=True,
             analytics_opt_out=True,
             datastores=[],
-            pytest_path=OPS_TEST_DIR,
+            pytest_path=f"{OPS_TEST_DIR} tests/integration/",
```
erosselli, line 160: nit: maybe we should make `tests/integration` a constant like all the other dirs here?

src/fides/api/alembic/migrations/versions/4bfbeff34611_add_polling_status.py
```diff
@@ -0,0 +1,36 @@
...
+Create Date: 2025-09-20 23:02:45.550170
+
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "4bfbeff34611"
+down_revision = "7db29f9cd77b"
```
erosselli, line 14: remember to update the down revision before merging :)

src/fides/api/alembic/migrations/versions/4bfbeff34611_add_polling_status.py
```diff
@@ -0,0 +1,36 @@
...
+    op.execute("alter type executionlogstatus rename to executionlogstatus_old")
+    op.execute(
+        "create type executionlogstatus as enum('in_processing', 'pending', 'complete', 'error', 'paused', 'retrying', 'skipped')"
+    )
+    op.execute(
+        (
+            "alter table executionlog alter column status type executionlogstatus using "
+            "status::text::executionlogstatus"
+        )
+    )
```
erosselli: I think before this you need to do something with any rows that have `polling` in their value; if nullable set them to null, otherwise maybe to `pending` or `paused`? I know it's unlikely that we'll need the down migration but since we're writing it, let's write it correctly

src/fides/api/models/privacy_request/request_task.py
```diff
@@ -318,3 +329,90 @@ def request_task_running(self, should_log: bool = False) -> bool:
...
         return task_in_flight
+
+
+class RequestTaskSubRequest(Base):
+    """
+    Model for storing individual sub-request data during the execution of a request task.
+    Supports 1:N relationship - each RequestTask can have multiple sub-requests.
+    Currently used for storing request data for polling tasks.
+    """
+
```
erosselli, line 338: is this because a single polling integration can make different requests? e.g get a user id from an endpoint, go use that user id to get some other data from another endpoint? each endpoint call would be a RequestTaskSubRequest? 

just checking my own understanding here :)

src/fides/api/models/privacy_request/request_task.py
```diff
@@ -183,6 +185,15 @@ class RequestTask(WorkerTask, Base):
         uselist=False,
     )
 
+    # Stores the sub-requests data for async polling tasks
+    sub_requests: RelationshipProperty[AppenderQuery] = relationship(
```
erosselli: what does `AppenderQuery` do here?

src/fides/api/models/privacy_request/request_task.py
```diff
@@ -318,3 +329,90 @@ def request_task_running(self, should_log: bool = False) -> bool:
...
+    )
+
+    request_task = relationship(
+        "RequestTask",
+        back_populates="sub_requests",
+    )
+
+    # Individual sub-request data (e.g., request_id, status, result data)
+    # Additional fields for enhanced sub-request tracking
+    param_values = Column(  # An encrypted JSON String - saved as a dict
```
erosselli, line 362: what does `param_values` mean? 🤔 why not call it something like `subrequest_data` ?

src/fides/api/models/privacy_request/request_task.py
```diff
@@ -318,3 +329,90 @@ def request_task_running(self, should_log: bool = False) -> bool:
...
+
+    def update_status(self, db: Session, status: str) -> None:
+        """Helper method to update the status of this sub-request."""
+        self.status = status
+        self.save(db)
+
+    def cleanup_external_storage(self) -> None:
+        """Clean up all external storage files for this sub-request"""
+        # Access the descriptor from the class to call cleanup
+        RequestTaskSubRequest.access_data.cleanup(self)
```
erosselli, line 408: I'm confused by this , should this not be `self.access_data.cleanup` ?

src/fides/api/models/privacy_request/request_task.py
```diff
@@ -318,3 +329,90 @@ def request_task_running(self, should_log: bool = False) -> bool:
...
+            key=CONFIG.security.app_encryption_key,
+            engine=AesGcmEngine,
+            padding="pkcs5",
+        ),
+    )
+
+    # Use descriptors for automatic external storage handling
+    access_data = EncryptedLargeDataDescriptor(
+        field_name="access_data", empty_default=[]
+    )
```
erosselli, line 389: can you explain how this works? 😅

src/fides/api/schemas/saas/async_polling_configuration.py
```diff
@@ -0,0 +1,79 @@
...
+from fides.api.util.collection_util import Row
+
+
+class SupportedDataType(Enum):
+    """Supported data types for polling async DSR result requests."""
+
+    json = "json"
+    csv = "csv"
+    xml = "xml"
+    attachment = "attachment"
```
erosselli: can you explain attachment in a comment in the code? I would have thought that `json` , `csv`, and `xml` are also attachments so not quite sure what the distinction is here, is it like a catch-all "other" ? (if so maybe we could rename to `other_attachment`)

src/fides/api/schemas/saas/async_polling_configuration.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+@dataclass
+class PollingResult:
```
erosselli: two things: 
1. TIL that `@dataclass` exists
2. Why not use a pydantic schema here instead of `@dataclass` ? to my untrained eyes (cause I just read the  `@dataclass` definition and that's it) they seem to be similar?

src/fides/api/schemas/saas/async_polling_configuration.py
```diff
@@ -0,0 +1,79 @@
...
+    """
+    Flexible result container for async polling operations.
+    Handles both structured data and file attachments.
+    """
+
+    data: Union[List[Row], bytes]
+    result_type: str  # "rows", "attachment", "url"
```
erosselli: any reason why this isn't an enum?

src/fides/api/schemas/saas/saas_config.py
```diff
@@ -115,6 +117,10 @@ class SaaSRequest(BaseModel):
     skip_missing_param_values: Optional[bool] = (
         False  # Skip instead of raising an exception if placeholders can't be populated in body
     )
+    correlation_id_path: Optional[str] = PydanticField(
+        default=None,
+        description="The path to the correlation ID in the response. For use with async polling.",
+    )
```
erosselli, line 123: is "correlation ID" supposed to make sense within the SaaS framework? it doesn't tell me much but also I haven't done a lot in this part of the codebase

src/fides/api/schemas/saas/saas_config.py
```diff
@@ -213,15 +219,15 @@ def validate_request(self) -> "SaaSRequest":
...
+    that is used to format each collection result, and correlation_id_path for async polling.
     """
 
     output: Optional[str] = None
 
     @model_validator(mode="after")
     def validate_request(self) -> "ReadSaaSRequest":
         """Validate that configs related to read requests are set properly"""
-        if not self.request_override:
+        if not self.request_override and not self.async_config:
```
erosselli: should there be some sort of ValueError raised if there's no async_config? I'm slightly confused by this change 😅

src/fides/api/schemas/saas/strategy_configuration.py
erosselli: but where did it go?

src/fides/api/service/async_dsr/handlers/polling_request_handler.py
```diff
@@ -0,0 +1,78 @@
+"""
+Pure HTTP utility for executing polling requests.
+
+This module contains low-level HTTP request execution for async DSR polling,
+with no business logic or orchestration dependencies.
```
erosselli, line 5: ❤️

src/fides/api/service/async_dsr/handlers/polling_request_handler.py
```diff
@@ -0,0 +1,78 @@
...
+    def get_status_response(
+        self,
+        client: AuthenticatedClient,
+        param_values: Dict[str, Any],
+    ) -> Response:
+        """Execute HTTP status request and return raw response."""
+
+        prepared_status_request = map_param_values(
+            "status", "polling request", self.status_request, param_values
+        )
```
erosselli, line 52: can we call this with keyword arguments so it's easier to follow what all of these are doing in `map_param_values` ?

src/fides/api/service/async_dsr/handlers/polling_request_handler.py
```diff
@@ -0,0 +1,78 @@
...
+
+        response: Response = client.send(prepared_status_request)
+
+        if not response.ok:
+            raise PrivacyRequestError(
+                f"Status request failed with status code {response.status_code}: {response.text}"
+            )
```
erosselli, line 59: this is caught somewhere upstream and converted into an error `ExecutionLog` right?

src/fides/api/service/async_dsr/handlers/polling_request_handler.py
```diff
@@ -0,0 +1,78 @@
...
+    def get_result_response(
+        self,
+        client: AuthenticatedClient,
+        param_values: Dict[str, Any],
+    ) -> Response:
+        """Execute HTTP result request and return raw response."""
+        if not self.result_request:
+            raise PrivacyRequestError(
+                "result_request is not configured in the async polling configuration"
+            )
```
erosselli, line 72: so whoever calls `get_result_response` has to make sure they've set `result_request` ? don't we need a similar check for `status_request` in `get_status_response`?

src/fides/api/schemas/saas/async_polling_configuration.py
erosselli: nvm I answered my question by reading further down heh . I still think a code comment here explaining the enum values would be useful :)

src/fides/api/service/async_dsr/handlers/polling_response_handler.py
```diff
@@ -0,0 +1,279 @@
...
+
+# Content type mappings for data type detection
+CONTENT_TYPE_MAP = {
+    "application/json": SupportedDataType.json,
+    "text/csv": SupportedDataType.csv,
+    "application/csv": SupportedDataType.csv,
+}
+
+# File extensions that indicate attachments
+ATTACHMENT_EXTENSIONS = {
```
erosselli, line 37: wondering if there's isn't a library we can use for this kind of thing? this feels like it might be kind of painful to maintain

erosselli: quick google search shows  [mimetypes](https://docs.python.org/3/library/mimetypes.html) as part of the standard library  -- idk if it'd be helpful for this or you think it's overkill but thought I'd bring it up anyways

src/fides/api/service/async_dsr/handlers/polling_response_handler.py
```diff
@@ -0,0 +1,279 @@
...
+            return status_value == status_completed_value
+
+        # Otherwise, check truthiness
+        return bool(status_value)
+
+
+# Private helpers
+
+
+def _infer_data_type(request_path: str, response: Response) -> SupportedDataType:
```
erosselli: can we make these static methods in the `PollingResponseProcessor` class rather than functions here?

src/fides/api/service/async_dsr/handlers/polling_response_handler.py
```diff
@@ -0,0 +1,279 @@
...
+        return SupportedDataType.csv
+
+    # Try parsing as JSON
+    try:
+        response.json()
+        return SupportedDataType.json
+    except (ValueError, TypeError):
+        pass
+
+    return SupportedDataType.attachment  # Preserve unknown types as raw data
```
erosselli: should we have a `SupportedDataType.unknown` to distinguish it from known attachments or is it not needed?

src/fides/api/service/async_dsr/handlers/polling_response_handler.py
```diff
@@ -0,0 +1,279 @@
...
+
+
+def _should_store_as_attachment(
+    response: Response, request_path: str, inferred_type: SupportedDataType
+) -> bool:
+    """Determine if response should be treated as an attachment."""
+    content_type = response.headers.get("content-type", "").lower()
```
erosselli: can we use constants for the header strings? we're using the same raw string in a couple different places

src/fides/api/service/async_dsr/handlers/polling_response_handler.py
```diff
@@ -0,0 +1,279 @@
...
+        return True
+
+    return False
+
+
+def _extract_filename(response: Response, request_url: str) -> str:
+    """Extract filename from response or URL."""
+    # Try Content-Disposition header
+    disposition = response.headers.get("content-disposition", "")
+    match = re.search(r'filename[^;=\n]*=(([\'"]).*?\2|[^;\n]*)', disposition)
```
erosselli: same comment here, can we not get a library to do this? 
also can we leave a comment on what the regex is looking for specifically so it's readable by humans? 😄

src/fides/api/service/async_dsr/handlers/polling_response_handler.py
erosselli: or make them functions in a separate `file_processing_utils.py` file or similar

src/fides/api/service/async_dsr/strategies/async_dsr_strategy_callback.py
```diff
@@ -0,0 +1,219 @@
...
+from fides.api.service.async_dsr.strategies.async_dsr_strategy import AsyncDSRStrategy
+from fides.api.service.async_dsr.utils import AsyncPhase, get_async_phase
+from fides.api.service.connectors.saas.authenticated_client import AuthenticatedClient
+from fides.api.util.collection_util import Row
+
+if TYPE_CHECKING:
+    from fides.api.service.connectors.query_configs.saas_query_config import (
+        SaaSQueryConfig,
+    )
+
```
erosselli: since this is an entirely new file in an entirely new module, can we try to structure it so that there's no circular imports and we don't need `if TYPE_CHECKING` ?

src/fides/api/service/async_dsr/utils.py
```diff
@@ -0,0 +1,130 @@
...
+        f"async_type={request_task.async_type} (polling={async_type_check}), "
+        f"sub_requests_count={sub_requests_count}"
+    )
+
+    return async_type_check and sub_requests_count > 0
+
+
+def is_callback_completion(request_task: RequestTask) -> bool:
+    """Check if this is a completed callback request"""
+    return bool(request_task.callback_succeeded)
```
erosselli, line 55: I'm kind of wondering why all of these aren't instance methods on `RequestTask`

src/fides/api/service/async_dsr/strategies/async_dsr_strategy_callback.py
```diff
@@ -0,0 +1,219 @@
...
+            # Set async_type based on our strategy type
+            request_task.async_type = AsyncTaskType.callback
+            self.session.add(request_task)
+            self.session.commit()
+
+        # For callback strategy, we execute the initial request to trigger the callback
+        # then wait for the external system to call us back via webhook
+        # The actual callback execution and data processing happens in the webhook endpoint
+        # which will eventually call _callback_completion_access when the callback succeeds
+
```
erosselli: I'm a bit confused, isn't `_callback_completion_access` called by `async_retrieve_data`? or does the comment mean the webhook endpoint will call `async_retrieve_data`?

src/fides/api/service/async_dsr/strategies/async_dsr_strategy_callback.py
```diff
@@ -0,0 +1,219 @@
...
+                        f"Executed initial callback request for access task {request_task.id}"
+                    )
+                except ValueError as exc:
+                    if read_request.skip_missing_param_values:
+                        logger.debug(
+                            "Skipping optional access callback request: {}",
+                            exc,
+                        )
+                        continue
+                    raise exc
```
erosselli, line 130: maybe we can check the `path` first, to avoid the extra level of indentation on the rest of the code ? 
```py
if not read_request.path:
      raise whatever
      
...do stuff ...
```

src/fides/api/service/async_dsr/strategies/async_dsr_strategy_callback.py
```diff
@@ -0,0 +1,219 @@
...
+        all_requests.extend(read_requests)
+        rows_updated = 0
+
+        for request in all_requests:
+            if request.async_config and request_task.id:
+                # Set async_type based on our strategy type
+                request_task.async_type = AsyncTaskType.callback
+                self.session.add(request_task)
+                self.session.commit()
+
```
erosselli: I'm confused by this

src/fides/api/service/async_dsr/strategies/async_dsr_strategy_factory.py
```diff
@@ -0,0 +1,72 @@
...
+        try:
+            strategy_config = AsyncPollingConfiguration(**configuration)
+            return strategy_class(session, strategy_config)
+        except ValidationError as e:
+            raise FidesopsValidationError(message=str(e))
+    elif strategy_name == "callback":
+        # Callback strategy only requires session
+        return strategy_class(session)
+    else:
+        raise NoSuchStrategyException(f"Unsupported strategy: {strategy_name}")
```
erosselli: nit: don't really need the else here

src/fides/api/service/async_dsr/strategies/async_dsr_strategy_polling.py
```diff
@@ -0,0 +1,868 @@
+# datetime and timedelta imports would be used in RequestTaskSubRequest helper methods
```
erosselli: 🤔

src/fides/api/service/async_dsr/strategies/async_dsr_strategy_polling.py
```diff
@@ -0,0 +1,868 @@
...
+from fides.api.models.policy import Policy
+from fides.api.models.privacy_request.request_task import (
+    AsyncTaskType,
+    RequestTask,
+    RequestTaskSubRequest,
+)
+from fides.api.models.storage import get_active_default_storage_config
+from fides.api.models.worker_task import ExecutionLogStatus
+
+# ExecutionLogStatus would be used in RequestTaskSubRequest helper methods
```
erosselli: I don't understand these "would be used" comments

src/fides/api/service/async_dsr/strategies/async_dsr_strategy_polling.py
```diff
@@ -0,0 +1,868 @@
...
+    def _get_request_task(self, request_task_id: str) -> RequestTask:
+        """Get request task by ID or raise ValueError if not found."""
+        request_task = (
+            self.session.query(RequestTask)
+            .filter(RequestTask.id == request_task_id)
+            .first()
+        )
+        if not request_task:
+            raise ValueError(f"RequestTask with ID {request_task_id} not found")
+        return request_task
```
erosselli: I think the callback strategy had the same method defined, can we move it up to the base  async DSR strategy class?

src/fides/api/service/async_dsr/strategies/async_dsr_strategy_polling.py
```diff
@@ -0,0 +1,868 @@
...
+        from polling results and creates the necessary database records.
+        """
+        try:
+            # Get active storage config
+            storage_config = get_active_default_storage_config(self.session)
+            if not storage_config:
+                raise PrivacyRequestError("No active storage configuration found")
+
+            # Create attachment record and upload to storage
+            attachment = Attachment.create_and_upload(
```
erosselli: it's so nice how you can just do this and not worry about how it's stored 😄

src/fides/api/service/async_dsr/strategies/async_dsr_strategy_polling.py
erosselli, line 1: this file is a bit long and has a lot of imports which might point to it doing too many things ... maybe we can split some part into a separate file/class ? for example all the attachments utilities

src/fides/api/service/async_dsr/strategies/async_dsr_strategy_polling.py
```diff
@@ -0,0 +1,868 @@
...
+        # Check for status override vs standard HTTP request
+        if self.status_request.request_override:
+            # Handle status override function directly
+            override_function = SaaSRequestOverrideFactory.get_override(
+                self.status_request.request_override,
+                SaaSRequestType.POLLING_STATUS,
+            )
+
+            # Override functions return boolean status directly
+            return cast(
```
erosselli, line 454: so why do we need to cast it? 🤔

src/fides/api/service/connectors/fides/fides_client.py
erosselli: have we checked if they're used in fidesplus?

src/fides/api/service/connectors/query_configs/saas_query_config.py
```diff
@@ -353,7 +352,8 @@ def generate_query(
             ]
 
         param_values[UUID] = str(uuid4())
-        param_values[ISO_8601_DATETIME] = datetime.now().date().isoformat()
+        # Use full ISO-8601 timestamp including time component (UTC, seconds precision)
+        param_values[ISO_8601_DATETIME] = datetime.now().strftime("%Y-%m-%dT%H:%M:%SZ")
```
erosselli, line 370: is this change intentional?

src/fides/api/service/privacy_request/request_service.py
```diff
@@ -680,6 +619,27 @@ def requeue_interrupted_tasks(self: DatabaseTask) -> None:
...
+
+                            if polling_request_task_exists:
+                                # If the polling request task has no cached task ID, it's stuck
+                                logger.warning(
+                                    f"No task ID found for request task {request_task_id} "
+                                    f"(privacy request {privacy_request.id}) Contains polling tasks - "
+                                    f"keeping request in current status as it may be waiting for polling task to complete"
+                                )
+                                should_requeue = False
+                                break
```
erosselli: this method is already very long ,can we maybe split this new code into a little function and call it from here?

src/fides/api/service/saas_request/saas_request_override_factory.py
```diff
@@ -12,6 +12,9 @@
 from fides.api.schemas.saas.shared_schemas import ConsentPropagationStatus
 from fides.api.util.collection_util import Row
 
+if TYPE_CHECKING:
+    from fides.api.schemas.saas.async_polling_configuration import PollingResult
+
```
erosselli, line 17: 🥲

src/fides/api/service/saas_request/saas_request_override_factory.py
```diff
@@ -264,5 +277,57 @@ def validate_process_consent_webhook_function(f: Callable) -> None:
...
+    """
+    sig: Signature = signature(f)
+    if sig.return_annotation is not bool:
+        raise InvalidSaaSRequestOverrideException(
+            "Polling status override function must return a bool"
+        )
+    if len(sig.parameters) < 4:
+        raise InvalidSaaSRequestOverrideException(
+            "Polling status override function must declare at least 4 parameters"
+        )
```
erosselli, line 300: can we check that the first 4 parameters have the correct names? also TIL about `signature` heh

src/fides/api/task/execute_request_tasks.py
erosselli: maybe we can add this to a method in the enum class, e.g `ExecutionLogStatus.get_awaiting_statuses()` or some better name

noxfiles/setup_tests_nox.py
galvana: `OPS_TEST_DIR` is used in a few places so I figured I'd just make a more targeted change

clients/admin-ui/src/features/privacy-requests/types.ts
galvana: Updated to use `"warning"`

src/fides/api/models/privacy_request/request_task.py
galvana: I don't think we need it in this case. We won't have more than a few sub requests so this lazy-loading isn't required
```
sub_requests: "RelationshipProperty[List[RequestTaskSubRequest]]" = relationship(
    "RequestTaskSubRequest",
    back_populates="request_task",
    cascade="all, delete-orphan",
    order_by="RequestTaskSubRequest.created_at",
)
```

src/fides/api/models/privacy_request/request_task.py
galvana: That's exactly it! There will be cases where we need to execute multiple polling requests to get all the data for a single collection. Since these sub-requests will be async they'll need to be tracked separately and we'll aggregate the all the results once they're done

src/fides/api/models/privacy_request/request_task.py
galvana: The `param_values` concept is unique to SaaS integrations, it corresponds to the variables needed to populate an HTTP request. For example here, the `email` is the param value because it's needed to populate the query params.
```
read:
  method: GET
  path: /api/v2/users/search.json
  query_params:
    - name: query
      value: <email>
  param_values:
    - name: email
      identity: email
  data_path: users
  ```

File: src/fides/api/models/privacy_request/request_task.py (hunk 27 above)
Comment (galvana) [reply src/fides/api/models/privacy_request/request_task.py:389]: The `EncryptedLargeDataDescriptor` concept was introduced here https://github.com/ethyca/fides/pull/6199. It checks for the size of the data being written to the column and automatically writes it to external storage if the file is considered "large" (close to ~1GB). This is because of column size limits in Postgres. The read and writes to the column happen normally and the size checks, encryption, uploading, decryption happening automatically.

File: src/fides/api/models/privacy_request/request_task.py (hunk 26 above)
Comment (galvana) [reply src/fides/api/models/privacy_request/request_task.py:408]: It does need `self`. It's a bit indirect but it happens here https://github.com/ethyca/fides/blob/main/src/fides/api/models/field_types/encrypted_large_data.py#L150-L151

File: src/fides/api/schemas/saas/async_polling_configuration.py (hunk 28 above)
Comment (galvana) [reply src/fides/api/schemas/saas/async_polling_configuration.py]: I left some comments and removed `xml`, that was making things a bit confusing
```
class SupportedDataType(Enum):
    """Supported data types for polling async DSR result requests."""

    # Structured data types that can be parsed into rows
    json = "json"  # Parsed into List[Row] from JSON response
    csv = "csv"  # Parsed into List[Row] from CSV response
    # Binary/non-parseable data stored as raw bytes
    attachment = "attachment"  # Binary files (.zip, .pdf, .xml, etc.) stored as bytes
```

File: src/fides/api/schemas/saas/async_polling_configuration.py (hunk 29 above)
Comment (galvana) [reply src/fides/api/schemas/saas/async_polling_configuration.py]: Apparently it's more lightweight than a Pydantic model. I'm going to change it to a Pydantic model so it's consistent with other schemas

File: src/fides/api/schemas/saas/async_polling_configuration.py (hunk 30 above)
Comment (galvana) [reply src/fides/api/schemas/saas/async_polling_configuration.py]: Nope 🙃, I added this
```
class PollingResultType(Enum):
    """Types of results from async polling operations."""

    rows = "rows"  # Structured data parsed into List[Row]
    attachment = "attachment"  # Binary file data stored as bytes
```

File: src/fides/api/schemas/saas/saas_config.py (hunk 31 above)
Comment (galvana) [reply src/fides/api/schemas/saas/saas_config.py:123]: It's something new we added to track requests involved in async polling. For both access and erasures, we fire off an initial request that we then need to check the status of. The initial request will usually return a value that we can use in subsequent status and result requests.

File: src/fides/api/schemas/saas/saas_config.py (hunk 32 above)
Comment (galvana) [reply src/fides/api/schemas/saas/saas_config.py]: I think we can get rid of `not self.sync_config`

File: src/fides/api/schemas/saas/strategy_configuration.py (hunk 4 above)
Comment (galvana) [reply src/fides/api/schemas/saas/strategy_configuration.py:192]: Sorry, it's called `AsyncPollingConfiguration`

File: src/fides/api/service/async_dsr/strategies/async_dsr_strategy_callback.py (hunk 42 above)
Comment (galvana) [reply src/fides/api/service/async_dsr/strategies/async_dsr_strategy_callback.py]: Turns out I didn't need this `TYPE_CHECKING` anymore, no weird imports!

File: src/fides/api/service/async_dsr/strategies/async_dsr_strategy_callback.py (hunk 44 above)
Comment (galvana) [reply src/fides/api/service/async_dsr/strategies/async_dsr_strategy_callback.py]: The comment is a bit confusing, we'll get a value posted to `/request-task/callback` and that will call `queue_request_task -> ... -> async_retrieve_data`

Comment (galvana) [reply src/fides/api/service/async_dsr/strategies/async_dsr_strategy_callback.py]: > For callback strategy, we execute the initial request to trigger the callback
   then wait for the external system to call us back via webhook.
   The callback endpoint logic will then queue the request task for it
   to be executed again with
   AsyncCallbackStrategy.async_retrieve_data -> _callback_completion_access

File: src/fides/api/service/async_dsr/strategies/async_dsr_strategy_callback.py (hunk 45 above)
Comment (galvana) [reply src/fides/api/service/async_dsr/strategies/async_dsr_strategy_callback.py:130]: But if there's no `path` we just don't do anything, the exception is raised outside the for loop

File: src/fides/api/service/async_dsr/strategies/async_dsr_strategy_callback.py (hunk 46 above)
Comment (galvana) [reply src/fides/api/service/async_dsr/strategies/async_dsr_strategy_callback.py]: I cleaned this up, we don't need to commit the `request_task` multiple times

File: src/fides/api/service/async_dsr/strategies/async_dsr_strategy_polling.py (hunk 49 above)
Comment (galvana) [reply src/fides/api/service/async_dsr/strategies/async_dsr_strategy_polling.py]: Cursor comments, I got rid of them

File: src/fides/api/service/async_dsr/strategies/async_dsr_strategy_polling.py (hunk 50 above)
Comment (galvana) [reply src/fides/api/service/async_dsr/strategies/async_dsr_strategy_polling.py]: Good call!

File: src/fides/api/service/async_dsr/strategies/async_dsr_strategy_polling.py (hunk 51 above)
Comment (galvana) [reply src/fides/api/service/async_dsr/strategies/async_dsr_strategy_polling.py]: Agreed! Thank you @JadeCara 🙏

File: src/fides/api/service/async_dsr/strategies/async_dsr_strategy_polling.py (hunk 52 above)
Comment (galvana) [reply src/fides/api/service/async_dsr/strategies/async_dsr_strategy_polling.py:454]: The return type for `override_function` is a union of a few types. Request overrides of type `SaaSRequestType.POLLING_STATUS` will always return a boolean though. We could probably refactor this to improve the type hints but that's out of scope or now

File: src/fides/api/service/async_dsr/utils.py (hunk 43 above)
Comment (galvana) [reply src/fides/api/service/async_dsr/utils.py:55]: I kept them here because this is specific to a SaaS workflow, but we already added `callback_succeeded` and other fields to `RequestTask` 🤔 Not sure what the best approach would be

File: src/fides/api/service/connectors/fides/fides_client.py
Comment (galvana) [reply src/fides/api/service/connectors/fides/fides_client.py:1]: All in Fides, we're good on the Fidesplus side

File: src/fides/api/service/connectors/query_configs/saas_query_config.py (hunk 53 above)
Comment (galvana) [reply src/fides/api/service/connectors/query_configs/saas_query_config.py:370]: Yes! The format for ISO-8601 is `YYYY-MM-DDTHH:MM:SSZ`

File: src/fides/api/service/saas_request/saas_request_override_factory.py (hunk 55 above)
Comment (galvana) [reply src/fides/api/service/saas_request/saas_request_override_factory.py:17]: I know 😢 this one is trickier to untangle, leaving it like this for now

File: src/fides/api/service/saas_request/saas_request_override_factory.py (hunk 56 above)
Comment (galvana) [reply src/fides/api/service/saas_request/saas_request_override_factory.py:300]: We tried this when we first wrote the feature but it was a bit tricky. Adding a todo ticket for this and the type hint cleanup https://ethyca.atlassian.net/browse/ENG-1554

File: src/fides/api/task/execute_request_tasks.py (hunk 11 above)
Comment (galvana) [reply src/fides/api/task/execute_request_tasks.py]: I made it an enum
```
if request_task.status not in RESUMABLE_EXECUTION_LOG_STATUSES:
```

File: src/fides/api/service/async_dsr/handlers/polling_request_handler.py (hunk 35 above)
Comment (galvana) [reply src/fides/api/service/async_dsr/handlers/polling_request_handler.py:59]: That's right! It gets handled by the same logic that catches exceptions from any of the connector's `retrieve_data`, `mask_data` so it shows up in the execution logs.

File: src/fides/api/service/async_dsr/handlers/polling_response_handler.py (hunk 37 above)
Comment (galvana) [reply src/fides/api/service/async_dsr/handlers/polling_response_handler.py:37]: Updated to use mimetypes, thanks!

File: src/fides/api/service/async_dsr/handlers/polling_response_handler.py (hunk 40 above)
Comment (galvana) [reply src/fides/api/service/async_dsr/handlers/polling_response_handler.py]: Using constants for the header strings

File: src/fides/api/service/async_dsr/handlers/polling_response_handler.py (hunk 41 above)
Comment (galvana) [reply src/fides/api/service/async_dsr/handlers/polling_response_handler.py]: Found this in the standard library
```
from email.message import Message
```

File: src/fides/api/service/async_dsr/handlers/polling_response_handler.py (hunk 39 above)
Comment (galvana) [reply src/fides/api/service/async_dsr/handlers/polling_response_handler.py]: I think this is ok for now, we'll just treat anything unknown as binary data

File: src/fides/api/service/async_dsr/handlers/polling_response_handler.py (hunk 38 above)
Comment (galvana) [reply src/fides/api/service/async_dsr/handlers/polling_response_handler.py]: Static methods it is!

File: src/fides/api/service/async_dsr/strategies/async_dsr_strategy_polling.py
Comment (galvana) [reply src/fides/api/service/async_dsr/strategies/async_dsr_strategy_polling.py:1]: I pulled out attachment and sub request functions into
- src/fides/api/service/async_dsr/handlers/polling_attachment_handler.py
- src/fides/api/service/async_dsr/handlers/polling_sub_request_handler.py

File: src/fides/api/service/privacy_request/request_service.py (hunk 54 above)
Comment (galvana) [reply src/fides/api/service/privacy_request/request_service.py]: I pulled out the check into `_has_polling_tasks`

File: src/fides/api/service/async_dsr/handlers/polling_request_handler.py (hunk 36 above)
Comment (galvana) [reply src/fides/api/service/async_dsr/handlers/polling_request_handler.py:72]: Added as similar check for `get_status_response`

File: noxfiles/setup_tests_nox.py (hunk 20 above)
Comment (erosselli) [reply noxfiles/setup_tests_nox.py:160]: I meant like `pytest_path=f"{OPS_TEST_DIR} {INTEGRATION_TEST_DIR}"`

File: src/fides/api/models/privacy_request/request_task.py (hunk 23 above)
Comment (erosselli) [reply src/fides/api/models/privacy_request/request_task.py:338]: is there a way to specify that sub request A's results are needed as input for sub request B? do we have a SaaS graph?

File: src/fides/api/service/async_dsr/utils.py (hunk 43 above)
Comment (erosselli) [reply src/fides/api/service/async_dsr/utils.py:55]: I think that since we have the saas-related fields already, it might make sense to move the methods into RequestTask too. We could prefix them with `saas` or just make sure we add in the docstring something like "Used for SaaS requests"

File: src/fides/api/models/privacy_request/request_task.py (hunk 26 above)
Comment (erosselli) [reply src/fides/api/models/privacy_request/request_task.py:408]: I guess I'm just wondering why it's a static method that receives the instance as a param rather than just having it be an instance method?

File: src/fides/api/alembic/migrations/versions/7db29f9cd77b_create_new_sub_request_table.py [hunk 57]
```diff
@@ -0,0 +1,85 @@
...
+            ),
+            nullable=True,
+        ),
+        sa.Column("rows_masked", sa.Integer(), nullable=True),
+        sa.ForeignKeyConstraint(
+            ["request_task_id"],
+            ["requesttask.id"],
+            name="request_task_sub_request_request_task_id_fkey",
+            ondelete="CASCADE",
+        ),
```
erosselli, line 70: do we want to add an index on `request_task_id` so searches for "all subtasks of a given task" are performant ?

erosselli (approved):
tentatively approving provided CI passes -- we've discussed offline re: manual testing . Left a couple of nit comments

src/fides/api/alembic/migrations/versions/4bfbeff34611_add_polling_status.py
galvana: I gave this a lot of thought and decided to go with this. This way we don't lose any audit info
```
def upgrade():
    # Check if value already exists
    connection = op.get_bind()
    result = connection.execute(
        sa.text(
            "SELECT 1 FROM pg_enum WHERE enumlabel = 'polling' "
            "AND enumtypid = (SELECT oid FROM pg_type WHERE typname = 'executionlogstatus')"
        )
    )

    if not result.fetchone():
        op.execute("ALTER TYPE executionlogstatus ADD VALUE 'polling'")


def downgrade():
    # The 'polling' value will remain in the enum but won't be used by older app versions
    pass
```

PR #6576, opened 2025-09-10
ENG-1373 fix location field union discrimination

thabofletcher:
Closes [ENG-1373]

### Description Of Changes

fixed pydantic's type union discriminator to not add location properties to other custom fields

### Code Changes

* fix and tests to prove it works

### Steps to Confirm

1.  Run admin-ui
2. POST the default config.json to /api/v1/plus/privacy-center-config - prior to the fix, these two fields from the Location object would be added to all custom privacy request fields:
```
"field_type": "location",
"ip_geolocation_hint": false
```

I wouldn't suggest reproducing that, as it could put your config in a bad state. If you want to reset to the empty default to retest you can use:

```
{
	"actions": [
		{
			"title": "Access your data",
			"identity_inputs": {
				"email": "required"
			}
		},
		{
			"policy_key": "default_erasure_policy",
			"title": "Erase your data",
			"identity_inputs": {
				"email": "required"
			}
		}
	]
}
```

### Pre-Merge Checklist

* [x] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [x] No followup issues
* Database migrations:
  * [x] No migrations
* Documentation:

[ENG-1373]: https://ethyca.atlassian.net/browse/ENG-1373?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,12 @@ Changes can also be flagged with a GitHub label for tracking purposes. The URL o
 - https://github.com/ethyca/fides/labels/high-risk: to indicate that a change is a "high-risk" change that could potentially lead to unanticipated regressions or degradations
 - https://github.com/ethyca/fides/labels/db-migration: to indicate that a given change includes a DB migration
 
-## [Unreleased](https://github.com/ethyca/fides/compare/2.69.2...main)
+## [Unreleased](https://github.com/ethyca/fides/compare/2.70.0...main)
+
+### Fixed
+- TCF alert rendering incorrectly during react lifecycle [#6582](https://github.com/ethyca/fides/pull/6582)
+
+## [2.70.0](https://github.com/ethyca/fides/compare/2.69.2...2.70.0)
 
 ### Added
 - Added new `WebMonitorGroupJob` model and table to store web monitor scan data [#6542](https://github.com/ethyca/fides/pull/6542) https://github.com/ethyca/fides/labels/db-migration
@@ -32,6 +37,8 @@ Changes can also be flagged with a GitHub label for tracking purposes. The URL o
 - Request manager screen now shows loader while waiting for sets of results [#6552](https://github.com/ethyca/fides/pull/6552)
 - Request manager screen now persists pagination parameters to URL [#6552](https://github.com/ethyca/fides/pull/6552)
 - Request manager screen no longer persists search to local storage and instead populates URL parameters [#6552](https://github.com/ethyca/fides/pull/6552)
+- Request manager CSV Download now flattens the subject identity object to comma separated columns [#6563](https://github.com/ethyca/fides/pull/6563)
+- Request manager CSV Download now flattens the custom privacy request fields object to comma separated columns [#6563](https://github.com/ethyca/fides/pull/6563)
 
 ### Deprecated
 - DSR 2.0 is deprecated. New requests will be created using DSR 3.0 only. Existing DSR 2.0 requests will continue to process until completion. [#6458](https://github.com/ethyca/fides/pull/6458)
--- a/clients/admin-ui/src/features/datamap/reporting/DatamapReportTable.tsx
+++ b/clients/admin-ui/src/features/datamap/reporting/DatamapReportTable.tsx
@@ -356,20 +356,14 @@ export const DatamapReportTable = () => {
 
   // Get the proper name for dropdown options (no lower-casing)
   const getSystemGroupOptionName = () => {
-    // 1. Priority: Custom column name from user overrides
+    // Priority: Custom column name from user overrides
     const customSystemGroupName =
       columnNameMapOverrides[COLUMN_IDS.SYSTEM_GROUP];
     if (customSystemGroupName) {
       return customSystemGroupName;
     }
 
-    // 2. Priority: Use actual system group taxonomy name if available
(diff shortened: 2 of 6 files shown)


src/fides/api/schemas/privacy_center_config.py
```diff
@@ -86,9 +93,27 @@ def validate_location_field(cls, values: Dict[str, Any]) -> Dict[str, Any]:
...
+    if isinstance(v, dict):
+        field_type = v.get("field_type")
+        if field_type == "location":
+            return "location"
+        return "custom"
+    # For model instances, get field_type attribute
+    field_type = getattr(v, "field_type", None)
+    if field_type == "location":
+        return "location"
+    return "custom"
```
erosselli:
```suggestion
    if isinstance(v, dict):
        field_type = v.get("field_type")
    else:
        # For model instances, get field_type attribute
        field_type = getattr(v, "field_type", None)
      
    if field_type == "location":
        return "location"
    return "custom"
```

erosselli: I don't know if I got the whitespace correctly on the suggestion so I wouldn't commit it directly :)

tests/ops/schemas/test_privacy_center_config.py
```diff
@@ -198,6 +199,121 @@ def test_privacy_center_config_mixed_custom_fields_with_location(self):
...
+
+        # Serialize to dict to check the actual output
+        serialized = config.model_dump(mode="json")
+        fields = serialized["actions"][0]["custom_privacy_request_fields"]
+
+        # Location field should have ip_geolocation_hint
+        location_field = fields["location"]
+        assert "ip_geolocation_hint" in location_field
+        assert location_field["ip_geolocation_hint"] is True
+        assert location_field["field_type"] == "location"
```
erosselli, line 247: any chance we can assert more specifically here, i.e 

```py
assert location_field == { "ip_geoglocation_hint":  True, .... all the other expected key,value pairs .... } 
```

erosselli: cursor loves these types of "structural" assertions like `assert "ip_geolocation_hint" in location_field` before checking actual values e.g `assert location_field["ip_geolocation_hint"] is True` , the first assertion is essentially useless. We have it in our fidesplus backend testing best practices rule but it's not in Fides yet; same for all the other assertions in the test. 

I'm ok with the specific assertions to check the location fields aren;t there like the one below, but generally I think it's more helpful / better tests to assert entire values and not just checking that some key is present /some value is there 

```py
assert (
            "ip_geolocation_hint" not in select_field
        ), f"Select field should not have ip_geolocation_hint: {select_field}"
```

thabofletcher: I'm happy to make that change, but can you help me understand the purpose on these backend testing practices a little more?

As an example, this test's main purpose is about whether the default fields from the Location union are included or not (as opposed to their values) so in/not in seems appropriate. If I switch to a single line assertion I get something more like this:

```
       assert location_field == {
            "label": "Location",
            "required": True,
            "default_value": None,
            "hidden": False,
            "query_param_key": None,
            "field_type": "location",
            "ip_geolocation_hint": True,
        }
```
And we end up asserting very unrelated data and from the purposes of this object, we don't really logically care at all about query_param_key

I would think if we had tests like this with single line assertions and stronger contracts this issue would have been caught by tests so I can see the appeal, but the cost of that practice is much higher maintenance - every time we add a new field to the JSON, you could have tests unrelated to the change failing that need to be updated.

There are pros/cons either way, but unlike the imports cursor keeps pulling down into the test, I'm questioning whether fighting cursor is worth it in this instance? There are some advantages to the one by one assertions it makes and I just want to be sure we aren't fighting it just for personal preference

src/fides/api/schemas/privacy_center_config.py
thabofletcher: Thanks, good change

tests/ops/schemas/test_privacy_center_config.py
erosselli: okay fair enough 🙂 I think it's subjective / there's pros and cons to each approach --  happy to talk more in team retro about asserting entire results vs some specific assertions, I agree it's not needed for ALL tests but I think it's also easy for things to slip through the cracks if you only check certain things (kind of like confirmation bias in a way?) 

however beyond the philosophical debate, I think in these types of assertions the "assert key in result" is not useful if we're checking the value in the line immediately after? 
```
assert "ip_geolocation_hint" in location_field
assert location_field["ip_geolocation_hint"] is True
```

thabofletcher: From a "fewer lines of code" = better perspective I 100% agree and outside of tests I'm always about fewer lines - human readability is paramount. However from a test standpoint I tend to think that it a) doesnt hurt human readability b) is more preferred by cursor and c) arguably provides more accurate output in the test results to the real problem (at least in this test case, since ip_geolocation_hint being missing would indicate a bug in the code but ip_geolocation_hint being False would just meant that something needed to be updated)

PR #6591, opened 2025-09-12
Add Oauth2 support for connection configs

johnewart:
Closes ENG-1216

### Description Of Changes

This adds support for OAuth to HTTPS connections (anywhere they are used); currently it supports client credentials flow, but can be extended if need be to other types. 

### Code Changes

* Database / API schema for OAuth config
* Added an API endpoint to put/patch/delete OAuth configs for connections
* Tests to validate that web hooks (in particular) do the OAuth dance by making sure the token URL is requested and that the bearer token it is given gets used in a subsequent request. 
* Tests to validate that the API endpoints work as expected

### Steps to Confirm

1.  Nothing manual

### Pre-Merge Checklist

* [x] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:


dev-requirements.txt
```diff
@@ -18,10 +18,12 @@ pytest==7.2.2
 requests-mock==1.10.0
 setuptools>=64.0.2
 sqlalchemy-stubs==0.4
+types-oauthlib==3.3.0
 types-paramiko==3.0.0.10
 types-PyYAML==6.0.11
 types-redis==4.3.4
 types-requests
+types-requests_oauthlib==2.0.0
```
galvana: I had to update this locally to get Fides to build, it looks like `types-oauthlib==3.3.0` and `types-requests_oauthlib==2.0.0` are not available
```suggestion
types-oauthlib==3.3.0.20250822
types-paramiko==3.0.0.10
types-PyYAML==6.0.11
types-redis==4.3.4
types-requests
types-requests_oauthlib==2.0.0.20250809
```

src/fides/api/models/connection_oauth_credentials.py
```diff
@@ -0,0 +1,50 @@
...
+    client_credentials = "client_credentials"
+    password = "password"
+    implicit = "implicit"
+
+
+class OAuthConfig(Base):
+    """
+    Stores credentials to connect fidesops to anything that uses OAuth2 for authentication.
+    """
+
```
galvana, line 28: We prefer using underscores in our multi-word table names. Make sure to update the migration file too
```
@declared_attr
    def __tablename__(cls) -> str:
        """Overriding base class method to set the table name."""
        return "oauth_config"
```

src/fides/api/models/privacy_request/privacy_request.py
```diff
@@ -893,9 +893,12 @@ def trigger_pre_approval_webhook(
...
         """
         # temp fix for circular dependency
-        from fides.api.service.connectors import HTTPSConnector, get_connector
+        from fides.api.service.connectors import (
+            HTTPSConnector,
+            get_connector,
+        )
 
-        https_connector: HTTPSConnector = get_connector(webhook.connection_config)  # type: ignore
+        https_connector: Union[HTTPSConnector | HTTPSOAuth2Connector] = get_connector(webhook.connection_config)  # type: ignore
```
galvana, line 959: Python 3.9 will complain about the use of `|`, so let's go with this
```suggestion
        https_connector: Union[HTTPSConnector, HTTPSOAuth2Connector] = get_connector(webhook.connection_config)  # type: ignore
```

galvana: It also looks like `HTTPSOAuth2Connector` isn't defined?

src/fides/api/schemas/connection_configuration/connection_oauth_config.py
```diff
@@ -0,0 +1,47 @@
...
+from fides.api.schemas.base_class import NoValidationSchema
+
+
+class OAuthGrantType(str, Enum):
+    """OAuth2 grant types supported by the system"""
+
+    authorization_code = "authorization_code"
+    client_credentials = "client_credentials"
+    password = "password"
+    implicit = "implicit"
```
galvana: If I understand this correctly,`BackendApplicationClient` would only support the `client_credentials` grant type? If so, can we remove the other grant types so we only accept the supported grant types in the API schemas.
```suggestion
class OAuthGrantType(str, Enum):
    """OAuth2 grant types supported by the system"""

    client_credentials = "client_credentials"
```
    
    
    
    
    
    
```

File: src/fides/api/schemas/connection_configuration/connection_oauth_config.py [hunk 5]
```diff
@@ -0,0 +1,47 @@
...
+
+
+class OAuthConfigResponse(BaseModel):
+    """Schema for OAuth2 configuration used in API responses with masked sensitive fields"""
+
+    grant_type: OAuthGrantType
+    token_url: str
+    scope: Optional[str] = None
+    client_id: str
+    client_secret: str = "**********"  # Always masked in responses
```
galvana, line 36: Nice 👍

galvana (commented):
Looks good overall! Just some minor comments

src/fides/api/schemas/connection_configuration/connection_oauth_config.py
johnewart: Sure, we can remove the ones not currently being used - the oauth library does support them; if other grant types need to be added later, the client creation / OAuth dance logic can be switched based on the type

src/fides/api/models/privacy_request/privacy_request.py
johnewart: Yeah, looks like I missed a few hunks / lines when I pushed it up for draft, I will push them up along with the OAuth grant type cleanup

src/fides/api/models/connection_oauth_credentials.py
johnewart: Updated both the metadata for the class and the migration to use `oauth_config` in the next commit

johnewart:
Going to close and re-open; GitHub's conflict editor went haywire and removed the DSR payload test so it will be easier to read. I will refer to this PR in case someone wants to see the existing comments.

PR #6593, opened 2025-09-12
ENG-1356: Adds support for dynamic vendor disclosure in Fides UI

eastandwestwind:
⚠️ Contains 2 migrations, check downrevs before merging ⚠️ 

Closes https://ethyca.atlassian.net/browse/ENG-1356

Please test alongside Fidesplus PR - https://github.com/ethyca/fidesplus/pull/2515

### Description Of Changes

This pull request adds support for dynamic vendor and cookie disclosure within the Fides UI. This feature allows administrators to configure privacy experiences to show end-users the specific vendors (systems) and cookies (assets) associated with a privacy notice. It also introduces an override mechanism to dynamically disable certain systems from being disclosed.

### Code Changes
*   **Database**:
    *   Adds a database migration to include vendor disclosure configuration on the `ExperienceConfig` model.
    *   Adds a second migration to add a `duration` field to the `Asset` model, intended for storing cookie lifetime.
*   **Backend API**:
    *   Updates the `ExperienceConfig` API model to support the new disclosure fields.
    *   Implements server-side support for a `fidesDisabledSystems` configuration option to filter which systems are returned for a privacy experience.
*   **Admin UI**:
    *   Adds a new section to the Privacy Experience form to configure vendor and cookie disclosure.
    *   Updates the Asset creation/editing modal to include the new `duration` field.
*   **Fides.js Client**:
    *   Adds a `fidesDisabledSystems` override option that can be passed during initialization.
    *   Updates the consent overlay UI to display a list of vendors, and a nested list of cookies under each vendor.
    *   Adds translations for new UI text like "vendors", "cookies", and "back".
*   **Privacy Center**:
    *   Adds support for a `FIDES_DISABLED_SYSTEMS` environment variable to be used as a default for the system override.

### Steps to Confirm

1.  In the Admin UI, navigate to an existing Privacy Experience or create a new one.
2. In the experience configuration, enable the new options for vendor and cookie disclosure.
3.  Ensure a system is set up that has associated assets (cookies) with the new `duration` field populated. The cookie data uses should be as or more specific than one of the enabled privacy notice data uses, e.g. if the data use on a privacy notice is "marketing", the cookie must have "marketing" or more specific like "marketing.advertising"
5.  Load the privacy experience on a test page.
6.  When the consent banner appears, navigate to the vendor disclosure view and confirm that the list of vendors and their associated cookies are displayed as configured.
7.  Test the `fidesDisabledSystems` override by passing system keys to the fides.js initialization options and confirm those systems are excluded from the list, e.g. `http://localhost:3001/fides-js-demo.html?geolocation=us-ca&fides_disabled_systems=google_analytics`. Now the system with the key "google_analytics" should not show up on the Fides UI.

### Pre-Merge Checklist

* [x] Issue requirements met
* [x] `CHANGELOG.md` updated
* Followup issues:
  * [ ] Followup issues created
  * [ ] No followup issues
* Database migrations:
  * [ ] No migrations
* Documentation:

### DIFF
--- a/src/fides/api/alembic/migrations/versions/f108fa05c579_adds_optional_duration_field_to_assets.py
+++ b/src/fides/api/alembic/migrations/versions/f108fa05c579_adds_optional_duration_field_to_assets.py
@@ -1,7 +1,7 @@
 """adds optional duration field to assets
 
 Revision ID: f108fa05c579
-Revises: 918aefc950c9
+Revises: 9caf76161e55
 Create Date: 2025-08-22 12:47:31.374493
 
 """
@@ -11,7 +11,7 @@ from alembic import op
 
 # revision identifiers, used by Alembic.
 revision = "f108fa05c579"
-down_revision = "918aefc950c9"
+down_revision = "9caf76161e55"
 branch_labels = None
 depends_on = None
 


src/fides/api/models/asset.py
```diff
@@ -192,3 +193,12 @@ async def get_by_system_async(
...
         return result.scalars().all()
+
+    # Expose related System attributes for API serialization convenience
+    @property
+    def system_name(self) -> Optional[str]:
+        return self.system.name if getattr(self, "system", None) else None
+
+    @property
+    def system_fides_key(self) -> Optional[str]:
+        return self.system.fides_key if getattr(self, "system", None) else None
```
erosselli: are there any performance concerns to doing this? e.g we'll be running one query per asset to retrieve its system, it might be more efficient to get system info separately and then join it with assets with a proper SQL query. 

If the `system_name` and `system_fides_key` are not used in list / bulk API responses and only in the detail of the asset then that's ok

erosselli (commented):
Backend looks good, I left a question. Also not approving since the bulk of this PR is frontend, consider it backend-approved and needing a frontend review

clients/fides-js/src/components/ConsentContent.tsx
```diff
@@ -20,10 +24,15 @@ const ConsentContent = ({
   renderModalFooter,
   children,
   onVendorPageClick,
+  headerContent,
 }: ConsentContentProps) => {
   const { i18n } = useI18n();
-  const title = i18n.t("exp.title");
-  const description = i18n.t("exp.description");
+  const defaultTitle = i18n.t("exp.title");
```
lucanovera: The logic for what title and description is displayed in split between this ConsentContent.tsx component and the Overlay.tsx. I found it hard to track what's happening. My suggestion would be to make this component simply render the values from headerContent and add the default value to Overlay.tsx. The logic for what to display would then be found exclusively in Overlay.tsx.

clients/fides-js/src/components/ConsentContent.tsx
```diff
@@ -53,7 +62,7 @@ const ConsentContent = ({
               allowHTMLDescription={window.Fides?.options?.allowHTMLDescription}
             />
           </p>
-          {showGpcInfo && (
+          {!headerContent && showGpcInfo && (
```
lucanovera: I think wether the gpc info shows up or not should not be determined by headerContent as it can be an unexpected side effect. We should add a new prop that controls wether it shows or not and that value should depend on isVendorAssetDisclosureView instead. Something like:

renderGpcInfo={!isVendorAssetDisclosureView}

lucanovera (changes requested):
Lots of great work here 💯 . I think this feature really shows how Fides has such an accurate tracking of consent that we can display each provider and even each cookie for the providers.
I tested the feature and found that every major feature is working correctly (yay), the assets are displayed and the exclude by system works. 

I have some comments about my testing and improving the UX:

Testing:
1. When I first switched to the fides/fidesplus branches, I saw a couple errors about the alembic migrations. They would show every time and I was unable to test until I cleared my db and started from scratch.
2. I see GVL vendors do appear in the disclosure, only without Duration, as it's not possible to edit them. Just wanted to make sure this is expected as the testing steps didn't mention that

User Experience:
3. There is no easy way to know if my assets need a Duration or not as there is no indication in the asset table. I can only find out if I try to edit my cookie assets 1 by 1 which isn't a great experience.
4. Initially, I didn't know what the Duration field was for or what was the expected input. It was worsen by the fact the field was required. I suggest you adding a help icon with some info about the expected value and the use.
5. Can the "Automatically delete subdomain cookies" option be part of the "Vendors & Assets" header? I know it's not part of the same feature, but I find it weird we have 2 elements about cookies so close and yet with a divider?
6. If you edit a TCF experience, you see the "Vendors & Assets" header but no options for it. If we do point 5. that would solve it. Otherwise the header shouldn't show if it has no options.
7. I found the "Asset types to disclose" UI to be confusing because all the dropdown options are disabled, the only action possible is to remove the Cookies which is required to be there. I suggest this UI to be hidden until more asset types are supported and only show supported assets in the dropdown. The "only cookie is supported" warning can be added to the help tooltip for Enable vendor disclosure.

Code:
8. Left you some comments in the code about the headerContent & showGpcInfo.
9. Thanks for adding good coverage. All other code changes look good!

src/fides/api/models/asset.py
eastandwestwind: Good call. I've updated the model to eagerly load the system relationship so that these 2 `system_name` and `system_fides_key` props can be accessed without performance issues.

clients/fides-js/src/components/ConsentContent.tsx
eastandwestwind: Great suggestion, I like it!

src/fides/api/models/asset.py
```diff
@@ -190,5 +191,17 @@ async def get_by_system_async(
...
+        # Explicitly eager load the `system` relationship to make this query's
+        # performance predictable and prevent N+1 issues.
+        query = query.options(selectinload(cls.system))  # type: ignore[attr-defined]
         result = await async_session.execute(query)
         return result.scalars().all()
+
+    # Expose related System attributes for API serialization convenience
+    @property
+    def system_name(self) -> Optional[str]:
+        return self.system.name if getattr(self, "system", None) else None
```
erosselli: nit: why do we need `getattr` here ?  we've defined `system` as a relationship on the model, the `system` attribute should always be there right?

src/fides/api/models/privacy_notice.py
```diff
@@ -189,41 +192,95 @@ def default_preference(self) -> UserConsentPreference:
...
+        # Associate the pre-fetched cookies to notices based on their data uses
+        for notice in notices:
+            matching_cookies = [
+                cookie
+                for cookie in all_relevant_cookies
+                if any(
+                    ",".join(cookie.data_uses).lower().startswith(d.lower())
+                    for d in notice.data_uses
+                )
+            ]
```
erosselli: we're iterating over `all_relevant_cookies` (potentially a list with hundreds of items) multiple times, and also ` ",".join(cookie.data_uses).lower().startswith(d.lower())` is hard to read and is essentially duplicating logic that we have in the SQL query in `_get_cookie_filter_for_data_uses` . I think we can re-write this:
 
```py
cookies_by_data_use = {}  # dict of data_use -> list of cookies

for cookie in all_relevant_cookies:
    for data_use in cookie.data_uses: 
        if data_use not in cookies_by_data_use:
             cookies_by_data_use[data_use] = []
        cookies_by_data_use[data_use].append(cookie)

for notice in notices:
    matching_cookies = [*cookies_by_data_use[data_use] for data_use in notice.data_uses]
   # dedup matching cookies since there might be duplicates
```

erosselli: I do wonder if there's a way to do this matching in the DB rather than in-memory... but well, two queries is better than N+1 queries anyway :)

src/fides/api/models/asset.py
eastandwestwind: right- updated!

src/fides/api/models/privacy_notice.py
eastandwestwind: oooh so much better 💯

src/fides/api/models/privacy_notice.py
```diff
@@ -189,42 +192,157 @@ def default_preference(self) -> UserConsentPreference:
...
+                cookie_filter,
+            )
         )
+        if exclude_cookies_from_systems:
+            query = query.outerjoin(System).filter(
+                or_(
+                    Asset.system_id.is_(None),
+                    System.fides_key.not_in(exclude_cookies_from_systems),
+                )
+            )
```
erosselli, line 247: oof this is unfortunate, I didn't realize we had the Asset's system ID but what we get from Fides JS is the fides_key. ok this looks good, thanks for making the change

src/fides/api/models/privacy_notice.py
```diff
@@ -189,42 +192,157 @@ def default_preference(self) -> UserConsentPreference:
...
+
+            # Hierarchical descendants: e.g., "analytics" matches "analytics.reporting"
+            prefix = f"{notice_data_use}."
+            for du_key, cookie_list in cookies_by_data_use.items():
+                if du_key.startswith(prefix):
+                    for cookie in cookie_list:
+                        key = id(cookie)
+                        if key not in seen:
+                            seen.add(key)
+                            ordered_cookies.append(cookie)
```
erosselli: what's the order here? why do we care about ordering? 🤔  if we don't care we could just use a set or dict structure without needing the list as well right?

src/fides/api/models/privacy_notice.py
```diff
@@ -189,42 +192,157 @@ def default_preference(self) -> UserConsentPreference:
...
+
+        return ordered_cookies
+
+    @cached_property
```
erosselli, line 285: let's make sure we know all the places where this `cookies` property is used, to ensure that making it a cached property doesn't accidentally cause stale data to be used elsewhere

erosselli (approved):
backend approved, assuming CI is green :) I did leave one more question

clients/admin-ui/src/features/system/tabs/system-assets/useSystemAssetColumns.tsx
```diff
@@ -120,6 +121,15 @@ const useSystemAssetColumns = ({
     header: "Domain",
   });
 
+  const duration = columnHelper.accessor((row) => row.duration, {
+    id: "duration",
+    cell: (props) =>
+      props.row.original.asset_type?.toLowerCase() === AssetType.COOKIE ? (
```
lucanovera:
```suggestion
      props.row.original.asset_type === AssetType.COOKIE ? (
```
Remove the lowercasing as the AssetType value is "Cookie" and it doesn't work.

src/fides/api/models/privacy_notice.py
eastandwestwind: hm, don't think we should care about ordering at all... will remove this 👍

PR #6594, opened 2025-09-12
Eng 1236 action center in progress view

thabofletcher:
Closes [ENG-1236]

### Description Of Changes

Partial changes to the front end to prove out the functionality of the monitor tasks' activity view

### Code Changes

* ant list view for monitor tasks
* status filtering for monitor tasks
* dismissed state for monitor tasks

### Steps to Confirm

1.  Check the feature flag - on dev and test envs it should be on by default  `/settings/about/alpha`: `Alpha full action center`
2. Verify that the Beta feature flag `Web monitor` is enabled
3. Verify the `Alpha full action center` feature flag turns on and off the Activity tab from within Detection & Discovery Action center

### Pre-Merge Checklist

* [x] Issue requirements met
* [ ] `CHANGELOG.md` updated
* Followup issues:
  * [x] Followup issues created
  * [ ] No followup issues
* Database migrations:
  * [x] No migrations
* Documentation:

[ENG-1236]: https://ethyca.atlassian.net/browse/ENG-1236?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/fides/api/alembic/migrations/versions/e2cda8d6abc3_add_dismissed_in_activity_view_to_.py
```diff
@@ -0,0 +1,31 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # Add the dismissed_in_activity_view column to monitortask table
+    op.add_column(
+        "monitortask",
+        sa.Column(
+            "dismissed_in_activity_view", sa.Boolean(), nullable=True, default=False
```
adamsachs: sorry to make a bit of a nit here and ask you to go back and update this because i know it'll mean you have to make changes downstream - but i'd much prefer we keep this column name more generic, because we very well could (will) end up using this concept of a "dismissed" task in other areas besides just the activity view. the activity view could also easily be renamed "in progress" tomorrow 😄 

while those sort of changes wouldn't _require_ a migration going forward, it'll just make this column feel really stale and it'd be confusing. i'd keep this pretty simple and just say `dismissed`. what do you think?
```suggestion
            "dismissed", sa.Boolean(), nullable=True, default=False
```

adamsachs: another nit is that it reads odd to me to have this both be `nullable=True` but to also set its default to `False`. if we're saying it's nullable, then i don't think we should a default value - the null state would indicate an empty/default state.

but really, i actually think this shouldn't be `nullable=True`. i don't think there's a good justification for making the boolean here have 3 states - the task has either been dismissed or it hasn't been dismissed; and the default state should be `False`, i.e. it hasn't been dismissed :)

```suggestion
            "dismissed", sa.Boolean(), nullable=False, default=False
```

adamsachs (dismissed):
nothing major, but i think the BE pieces here need a bit of cleanup - let me know what you think.

src/fides/api/alembic/migrations/versions/e2cda8d6abc3_add_dismissed_in_activity_view_to_.py
thabofletcher: My mistake, I though t this was applicable to this view only and should have brught this up. If its a generic "dismissed" I can definitely change it but it mill probably have like you said downstream impacts and some other tickets might slip

As far as nullable goes, since this is a new column would the migration take care of adding False to all the existing rows in the table? If we made not nullable=False default=False?

thabofletcher: Move this conve to the other PR where it will be merged

clients/admin-ui/src/features/data-discovery-and-detection/action-center/components/InProgressMonitorTaskItem.tsx
```diff
@@ -0,0 +1,244 @@
...
+  const [dismissMonitorTask, { isLoading: isDismissing }] =
+    useDismissMonitorTaskMutation();
+  const isDismissed = Boolean(task.dismissed_in_activity_view);
+  const canRetry = task.status === "error" && task.action_type !== "detection";
+
+  // Build a minimal monitor key -> icon data map for ConnectionTypeLogo
+  const { data: monitorAgg } = useGetAggregateMonitorResultsQuery({
+    page: 1,
+    size: 1000,
+  });
```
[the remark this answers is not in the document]
thabofletcher: I'm going to clean this up to be a bit cleaner, but realistically we do need to link all the items we have from the API to their icons so the work needs to be done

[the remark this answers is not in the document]
thabofletcher: Added this to TBD in this ticket https://ethyca.atlassian.net/browse/ENG-1241

clients/admin-ui/src/features/data-discovery-and-detection/action-center/components/InProgressMonitorTasksList.tsx
```diff
@@ -0,0 +1,153 @@
...
+                    textAlign: "center",
+                    lineHeight: "1",
+                    fontWeight: "600",
+                  }}
+                >
+                  {activeFilterCount}
+                </span>
+              )}
+            </span>
+            {showFilters ? <Icons.ChevronDown /> : <Icons.ChevronRight />}
```
speaker-ender: Is the `ChevronRight` icon used in the designs somewhere?  
If not, I don't think it's consistent with the rest of the designs.

clients/admin-ui/src/features/data-discovery-and-detection/action-center/components/InProgressMonitorTasksList.tsx
```diff
@@ -0,0 +1,153 @@
...
+                    fontSize: "10px",
+                    minWidth: "16px",
+                    textAlign: "center",
+                    lineHeight: "1",
+                    fontWeight: "600",
+                  }}
+                >
+                  {activeFilterCount}
+                </span>
+              )}
```
speaker-ender: Is there an active count in the designs or requested?  
If it's supposed to be kept, I don't think this is the correct visual or way to style components.  
We try to avoid using `style` attributes whenever possible.

clients/admin-ui/src/features/data-discovery-and-detection/action-center/components/InProgressMonitorTasksList.tsx
```diff
@@ -0,0 +1,153 @@
...
+      <Flex
+        justify="space-between"
+        align="center"
+        style={{ marginBottom: "16px" }}
+      >
+        <DebouncedSearchInput
+          value={searchQuery}
+          onChange={updateSearch}
+          placeholder="Search by monitor name..."
+          style={{ minWidth: "300px" }}
```
speaker-ender: We can set this via tailwind classes.  
I think `min-w-[300px]` is the brute force way of doing it.  
IMO this should be more standardized than it currently is.

clients/admin-ui/src/features/data-discovery-and-detection/action-center/hooks/useInProgressMonitorTasksList.tsx
```diff
@@ -0,0 +1,126 @@
+import { AntEmpty as Empty } from "fidesui";
+import { useCallback, useMemo, useState } from "react";
+
+import { useGetInProgressMonitorTasksQuery } from "../action-center.slice";
+
+export const useInProgressMonitorTasksList = () => {
+  const [pageIndex, setPageIndex] = useState(1);
+  const [pageSize] = useState(20);
+  const [searchQuery, setSearchQuery] = useState("");
```
speaker-ender: We have a `useSearch` hook that will sync this with the url state that we should use here

clients/admin-ui/src/features/data-discovery-and-detection/action-center/hooks/useInProgressMonitorTasksList.tsx
```diff
@@ -0,0 +1,126 @@
+import { AntEmpty as Empty } from "fidesui";
+import { useCallback, useMemo, useState } from "react";
+
+import { useGetInProgressMonitorTasksQuery } from "../action-center.slice";
+
+export const useInProgressMonitorTasksList = () => {
+  const [pageIndex, setPageIndex] = useState(1);
+  const [pageSize] = useState(20);
```
speaker-ender: We have a `usePagination` hook that should make this easier to manage + sync with the url

clients/admin-ui/src/features/data-discovery-and-detection/action-center/components/InProgressMonitorTasksList.tsx
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+  return status
+    .split("_")
+    .map((word) => word.charAt(0).toUpperCase() + word.slice(1))
+    .join(" ");
+};
+
+export const InProgressMonitorTasksList = () => {
+  const [showFilters, setShowFilters] = useState(false);
```
speaker-ender: I think this state can be eliminated if we don't need to adjust the chevron

clients/admin-ui/src/features/data-discovery-and-detection/action-center/action-center.slice.ts
```diff
@@ -319,6 +320,89 @@ const actionCenterApi = baseApi.injectEndpoints({
...
+        items: MonitorTaskInProgressResponse[];
+        total: number;
+        page: number;
+        pages: number;
+        size: number;
+      },
+      SearchQueryParams &
+        PaginationQueryParams & {
+          task_types?: string[];
+          statuses?: string[];
```
speaker-ender: Should these be string literals or enum values?

clients/admin-ui/src/features/data-discovery-and-detection/action-center/components/InProgressMonitorTasksList.tsx
```diff
@@ -0,0 +1,153 @@
...
+        placeholder="Select status..."
+        value={statusFilters}
+        onChange={updateStatusFilters}
+        style={{ width: "100%", marginBottom: "12px" }}
+        options={availableStatuses.map((status) => ({
+          label: formatStatusForDisplay(status),
+          value: status,
+        }))}
+        allowClear
+      />
```
speaker-ender: Wasn't this a checkbox group in the designs?

clients/admin-ui/src/features/data-discovery-and-detection/action-center/hooks/useInProgressMonitorTasksList.tsx
```diff
@@ -0,0 +1,126 @@
...
+  const { data, isLoading, isFetching } = useGetInProgressMonitorTasksQuery({
+    page: pageIndex,
+    size: pageSize,
+    search: searchQuery,
+    statuses: statusFilters.length > 0 ? statusFilters : undefined,
+    return_dismissed: showDismissed,
+  });
+
+  // Use all possible statuses instead of just what's in current data
+  const availableStatuses = allPossibleStatuses;
```
speaker-ender: Is this just to rename the variable?  
This can be done in the return statement if that's really desired.

clients/admin-ui/src/features/datastore-connections/ConnectionTypeLogo.tsx
```diff
@@ -14,31 +15,60 @@ import {
...
+};
+
 type ConnectionTypeLogoProps = {
-  data: string | ConnectionConfigurationResponse | ConnectionSystemTypeMap;
+  data:
+    | string
+    | ConnectionConfigurationResponse
+    | ConnectionSystemTypeMap
+    | MonitorAggregatedResults
+    | LogoSourceLite;
```
speaker-ender: Instead of continuing to type mash (I know you are just following the existing pattern), can we fix this to only accept a subset of properties that matter?  
Each type should have it's own translation function that can be less complex and be tested independently.  
That will help reduced the amount of validation that needs to be written since this is starting to mutate into a generic component (which i think is unnecessary here)

speaker-ender (changes requested):
I haven't pointed out every instance but I think we want most of these changes before merging.  
I'm working through some similar things with my ticket so feel free to reach out to me with questions or if you don't think certain changes should be blockers.

clients/admin-ui/src/features/data-discovery-and-detection/action-center/components/InProgressMonitorTasksList.tsx
thabofletcher: Nope, I had planned to move these later but now is fine to kill it since its so ugly

clients/admin-ui/src/features/data-discovery-and-detection/action-center/components/InProgressMonitorTasksList.tsx
thabofletcher: Wrapped the search input with a min-w-[300px], thanks

clients/admin-ui/src/features/data-discovery-and-detection/action-center/action-center.slice.ts
thabofletcher: Updated the RTK query signature to accept ExecutionLogStatus[], matching the enum we now pass through from the hook

clients/admin-ui/src/features/data-discovery-and-detection/action-center/components/InProgressMonitorTasksList.tsx
thabofletcher: Replaced the multi-select with a checkbox - I know there is more to do here to include status in s theckbox but that will be ahdneld in the follw up ticket

clients/admin-ui/src/features/datastore-connections/ConnectionTypeLogo.tsx
thabofletcher: I believe this was addressed as part of the overall refactor that addressed these issue. Let me know if you see anything more that needs done here!

clients/admin-ui/src/features/data-discovery-and-detection/action-center/components/InProgressMonitorTaskItem.tsx
```diff
@@ -0,0 +1,260 @@
...
+// Helper function to format status names for display
+const formatStatusForDisplay = (status: string): string => {
+  // Special case: "paused" should display as "Awaiting Processing"
+  if (status === "paused") {
+    return "Awaiting Processing";
+  }
+
+  return status
+    .split("_")
+    .map((word) => word.charAt(0).toUpperCase() + word.slice(1))
```
speaker-ender: We have a `capitalize` function in `admin-ui/src/features/common/utils.ts`

clients/admin-ui/src/features/data-discovery-and-detection/action-center/components/InProgressMonitorTaskItem.tsx
```diff
@@ -0,0 +1,260 @@
...
+        error?.data?.detail || error?.message || "Unknown error occurred";
+      toast({
+        status: "error",
+        description: `Failed to dismiss task: ${errorMessage}`,
+      });
+    }
+  };
+
+  return (
+    <div {...props} style={{ width: "100%" }}>
```
speaker-ender:
```suggestion
    <div {...props} className="w-full">
```

clients/admin-ui/src/features/data-discovery-and-detection/action-center/components/InProgressMonitorTaskItem.tsx
```diff
@@ -0,0 +1,260 @@
...
+                  <Button type="link" size="small">
+                    Failure reason
+                  </Button>
+                </Popover>
+              </>
+            )}
+          </Space>
+        </Col>
+        <Col span={4} className="flex items-center justify-end">
+          <Text type="secondary" style={{ fontSize: "12px" }}>
```
speaker-ender:
```suggestion
          <Text type="secondary" size="sm">
```

clients/admin-ui/src/features/data-discovery-and-detection/action-center/components/InProgressMonitorTaskItem.tsx
```diff
@@ -0,0 +1,260 @@
...
+            {monitorGroupLabel}
+          </Text>
+        </Col>
+        <Col span={3} className="flex items-center justify-end">
+          {isInProgress ? (
+            <Spinner size="sm" color="primary" thickness="2px" speed="0.6s" />
+          ) : (
+            <Text type="secondary" style={{ fontSize: "12px" }}>
```
speaker-ender:
```suggestion
            <Text type="secondary" size="sm">
```

clients/admin-ui/src/features/data-discovery-and-detection/action-center/components/InProgressMonitorTasksList.tsx
```diff
@@ -0,0 +1,126 @@
...
+  status: ExecutionLogStatus | string,
+): string => {
+  // Special case: "paused" should display as "Awaiting Processing"
+  if (status === "paused") {
+    return "Awaiting Processing";
+  }
+
+  return status
+    .split("_")
+    .map((word) => word.charAt(0).toUpperCase() + word.slice(1))
```
speaker-ender: We have a capitalize function in `admin-ui/src/features/common/utils.ts`
```suggestion
    .map(capitalize)
```
Will need to add 
`import { capitalize}  from "~/features/common/utils"`
to imports

clients/admin-ui/src/features/data-discovery-and-detection/action-center/hooks/useInProgressMonitorTasksList.tsx
```diff
@@ -0,0 +1,151 @@
...
+          <Empty
+            image={Empty.PRESENTED_IMAGE_SIMPLE}
+            description="No tasks in progress"
+          />
+        ),
+      },
+      pagination: {
+        current: pageIndex,
+        pageSize,
+        total: data?.total || 0,
```
speaker-ender: I think it makes sense to coerce this value if needed at the point of use.
Otherwise, I think there is meaning and context that can be derived from this value not being a number.

clients/admin-ui/src/pages/data-discovery/action-center/index.tsx
```diff
@@ -134,44 +140,68 @@ const ActionCenterPage = () => {
...
+      <Menu
+        aria-label="Action center tabs"
+        mode="horizontal"
+        items={tabs.map((tab) => ({
+          key: tab.hash,
+          label: tab.label,
+        }))}
+        selectedKeys={[activeTab]}
+        onClick={async (menuInfo) => {
+          await onTabChange(menuInfo.key as TopLevelActionCenterTabHash);
```
speaker-ender: Instead of casting, something more like this:
```suggestion
          const menuInfoKey = Object.values((value) => menuInfo.key === value)
          
          if (menuInfoKey) {
              await onTabChange(menuInfo.key as TopLevelActionCenterTabHash);
          }
```
This makes sure that we actually know that the key is valid.  
Using string literals instead of an enum object would reduce the type complexity when checking.  
This is partially a type limitation of from Ant who should really use better types for keys when providing callbacks.

clients/admin-ui/src/features/integrations/SelectableIntegrationBox.tsx
```diff
@@ -67,7 +67,7 @@ const SelectableIntegrationBox = ({
       <Flex justifyContent="space-between" alignItems="flex-start" width="100%">
```
speaker-ender:
```suggestion
      <Flex justifyContent="space-between" alignItems="flex-start" className="w-full">
```

clients/admin-ui/src/features/data-discovery-and-detection/action-center/components/InProgressMonitorTaskItem.tsx
```diff
@@ -0,0 +1,260 @@
...
+
+  return (
+    <div {...props} style={{ width: "100%" }}>
+      <Row gutter={12} className="w-full">
+        <Col span={17} className="align-middle">
+          <Space align="center" size={8} wrap>
+            {logoSource && (
+              <ConnectionTypeLogo data={logoSource} boxSize="24px" />
+            )}
+            <Title level={5} style={{ margin: 0 }}>
```
speaker-ender:
```suggestion
            <Title level={5} className="m-0">
```

clients/admin-ui/src/features/data-discovery-and-detection/action-center/components/InProgressMonitorTaskItem.tsx
```diff
@@ -0,0 +1,260 @@
...
+                  <Button
+                    size="small"
+                    loading={isDismissing}
+                    onClick={handleDismissTask}
+                  >
+                    Dismiss
+                  </Button>
+                )}
+                <Popover
+                  content={<div style={{ maxWidth: 360 }}>{task.message}</div>}
```
speaker-ender:
```suggestion
                  content={<div className="max-w-[360px]">{task.message}</div>}
```

clients/admin-ui/src/features/datastore-connections/add-connection/AddConnection.tsx
```diff
@@ -84,7 +86,9 @@ const AddConnection = () => {
         <Title level={3} style={{ marginBottom: 0 }} className="mt-4">
```
speaker-ender:
```suggestion
        <Title level={3}  className="mt-4 mb-0">
```

speaker-ender (changes requested):
Looking good!  
Just a few small changes.

speaker-ender (approved):

